PR #3, opened 2022-06-08
feat: add HTTP client with exponential backoff policy

miparnisari:
The existing HTTP client tries requests only once.

This new client will ensure that:

- server initialization won't fail if the OIDC keys endpoint is slow to respond
- functional tests won't fail if the dummy OIDC server is slow to start. E.g. https://github.com/openfga/openfga/runs/6839424195?check_suite_focus=true#step:7:28


cmd/openfga/openfga_test.go
```diff
@@ -199,26 +191,12 @@ func TestBuildServerWithOidcAuthentication(t *testing.T) {
...
+	if resp.StatusCode != http.StatusOK || err != nil {
+		t.Fatalf("Failed to start service")
+	}
 
-			return nil
-		},
-		backoffPolicy,
-	)
-	require.NoError(t, err)
+	t.Log("Service is now up")
```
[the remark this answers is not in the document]
craigpastro: I took a look and didn't find anything. Logs may be line oriented, so I don't know if the same styling applies.

pkg/httpclient/httpclient.go
```diff
@@ -0,0 +1,62 @@
+package httpclient
```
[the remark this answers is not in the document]
jon-whit: Can we call this package `retryablehttp` instead of `httpclient`?

pkg/httpclient/httpclient.go
```diff
@@ -0,0 +1,116 @@
...
+	internalClient http.Client
+}
+
+func NewRetryableHTTPClient() *RetryableHTTPClient {
+	return &RetryableHTTPClient{
+		internalClient: http.Client{},
+	}
+}
+
+func (client *RetryableHTTPClient) Do(req *http.Request) (*http.Response, []byte, error) {
```
jon-whit, line 69: Can we stick to the standard library signatures so we don't introduce a whole new (non-standard) library?

https://pkg.go.dev/net/http#Client.Do

pkg/httpclient/httpclient.go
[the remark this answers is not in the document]
craigpastro: Sorry, just seeing this now. Is there any reason we can just use [hashicorp/go-retryablehttp](https://github.com/hashicorp/go-retryablehttp)?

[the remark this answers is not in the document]
miparnisari: @craigpastro its license is MPL-2.0 and [it's forbidden](https://app.fossa.com/policies/licensing/53942)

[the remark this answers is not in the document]
craigpastro: Hmm, that's annoying. By the way, how do we log into Fossa?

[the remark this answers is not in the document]
craigpastro: I did find this: https://github.com/cncf/foundation/issues/138. No resolution as of now though.

pkg/retryablehttp/httpclient.go
```diff
@@ -0,0 +1,75 @@
...
+func (rt *RetryableRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	rt.once.Do(rt.init)
+	return rt.Client.Do(req)
+}
+
+type RetryableHTTPClient struct {
+	internalClient http.Client
+}
+
+func NewRetryableHTTPClient() *RetryableHTTPClient {
```
craigpastro, line 38: Small nit: but do you mind calling this `NewClient`, or something similar, as this is already the `retryablehttp` package?

pkg/retryablehttp/httpclient.go
```diff
@@ -0,0 +1,75 @@
...
+	)
+
+	// All retries failed
+	if err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
```
craigpastro, line 70: Would you mind adding a `Get` convenience function? 
```
func (c * RetryableHTTPClient) Get(url string) (*http.Response, error) {
	req, err := http.NewRequest("GET", url, nil)
	if err != nil {
		return nil, err
	}
	return c.Do(req)
}
```

pkg/retryablehttp/httpclient.go
```diff
@@ -0,0 +1,83 @@
...
+// RoundTrip executes a single HTTP transaction, but does not attempt to read the response, modify it, or close the body.
+func (rt *RetryableRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	rt.once.Do(rt.init)
+	return rt.Client.Do(req)
+}
+
+type RetryableHTTPClient struct {
+	internalClient http.Client
```
craigpastro, line 35: I want to make a retryable client that has its own Transport (for TLS). One way to do this is to use `NewClient` and then set `internalClient`. I am doing that here https://github.com/openfga/openfga/pull/2/files#diff-0b4e82b4c716948a5e14e84e6695d4fd755a5ea47aa0742573dce8fdd2b20bb4R218. In order to do this here I think I will need this to be public. Is that okay?

miparnisari: I added a new constructor

server/authentication/oidc/oidc.go
```diff
@@ -33,8 +36,9 @@ var _ authentication.OidcAuthenticator = (*RemoteOidcAuthenticator)(nil)
 
 func NewRemoteOidcAuthenticator(issuerURL, audience string) (*RemoteOidcAuthenticator, error) {
 	oidc := &RemoteOidcAuthenticator{
-		IssuerURL: issuerURL,
-		Audience:  audience,
+		IssuerURL:  issuerURL,
+		Audience:   audience,
+		httpClient: retryablehttp.RetryableHTTPClient{},
```
elbuo8, line 41: And if I want to inject my own?

miparnisari: ehmm why would you want to?

jon-whit:
@miparnisari After thinking about this a little more. Would it be possible to not introduce a whole new HTTP client library with retries and just use the `backoff.Retry(...)` approach to wait for the mock OIDC server to start up, just like we do with the OpenFGA server `ensureServiceUp`.

I think it'd be nice if we could avoid introducing all of the HTTP client code into the OpenFGA package just to wait for an HTTP server's healthiness. If we're going to introduce it though, it would at least be nice to put it in an `openfga/pkg` repository that has shared packages like this so we can share and reuse it elsewhere. If we're going to introduce it into our software maintenance cycle, we may as well be able to share it and evolve across other projects in the OpenFGA organization on Github.

miparnisari:
> I think it'd be nice if we could avoid introducing all of the HTTP client code into the OpenFGA package just to wait for an HTTP server's healthiness.

That's not the only use case, I also want to retry when fetching the JWKs in the OIDC authenticator.

> it would at least be nice to put it in an `openfga/pkg` repository that has shared packages like this so we can share and reuse it elsewhere

I'm confused, we don't need to put it in a brand new repository to reuse it, no?

jon-whit:
> I'm confused, we don't need to put it in a brand new repository to reuse it, no?

You're right, we don't have to. But for a shared library that expands beyond the scope of just this project it would make sense for it to live elsewhere. An import from openfga just for a shared retry http client is what would seem odd to me. I would expect an import like that to come from a shared repository, not openfga. That's what I meant.

server/authentication/oidc/oidc.go
```diff
@@ -22,6 +23,8 @@ type RemoteOidcAuthenticator struct {
 
 	JwksURI string
 	JWKs    *keyfunc.JWKS
+
+	httpClient retryablehttp.RetryableHTTPClient
```
jon-whit, line 27: This should be `*http.Client` and we can inject it with

`retryClient.StandardClient()`

miparnisari:
> > I'm confused, we don't need to put it in a brand new repository to reuse it, no?
> 
> You're right, we don't have to. But for a shared library that expands beyond the scope of just this project it would make sense for it to live elsewhere. An import from openfga just for a shared retry http client is what would seem odd to me. I would expect an import like that to come from a shared repository, not openfga. That's what I meant.

👍 i think when we have more things that we want to reuse we can create a new repo

server/authentication/oidc/oidc.go
```diff
@@ -138,21 +142,21 @@ func (oidc *RemoteOidcAuthenticator) GetConfiguration() (*authentication.OidcCon
...
-		return nil, errors.Errorf("unexpected status code getting OIDC: %v", res.StatusCode)
-	}
-
 	defer res.Body.Close()
 	body, err := ioutil.ReadAll(res.Body)
 	if err != nil {
 		return nil, errors.Errorf("error reading response body: %v", err)
 	}
 
+	if res.StatusCode != http.StatusOK {
```
elbuo8, line 156: Why was this moved 🤔

miparnisari: Reverted this change :D

elbuo8 (approved):

jon-whit (approved):

PR #4, opened 2022-06-09
ci: Add performance regression check

craigpastro:
## Review Checklist

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,11 +1,11 @@
-name: OpenFGA Main
+name: Main
 on:
   push:
     branches:
       - main
 
 jobs:
-  bench:
+  go-bench:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,59 +1,10 @@
-name: OpenFGA Pull Request
+name: OpenFGA Main
 on:
-  pull_request:
+  push:
     branches:
       - main
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-go@v3
-        with:
-          go-version-file: './go.mod'
-      - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
-        with:
-          version: latest
-
-  unit-tests:
-    runs-on: ubuntu-latest
-    timeout-minutes: 15
-    steps:
-      - uses: actions/checkout@v3
-
-      - uses: actions/setup-go@v3
-        with:
-          go-version-file: './go.mod'
-
-      - id: go-cache-paths
-        run: |
-          echo "::set-output name=go-build::$(go env GOCACHE)"
-          echo "::set-output name=go-mod::$(go env GOMODCACHE)"
-
-      # Cache go build cache, used to speedup go test
-      - uses: actions/cache@v3
-        with:
-          path: ${{ steps.go-cache-paths.outputs.go-build }}
-          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
-
-      # Cache go mod cache, used to speedup builds
-      - uses: actions/cache@v3
-        with:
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_request.yaml
```diff
@@ -53,3 +53,55 @@ jobs:
           files: ./coverageunit.out
           verbose: true
           fail_ci_if_error: false
+
+  bench:
```
miparnisari, line 57: 🪑 ? 🤣

.github/workflows/pull_request.yaml
```diff
@@ -53,3 +53,55 @@ jobs:
...
+
+      - name: Store benchmark result
+        uses: benchmark-action/github-action-benchmark@v1
+        with:
+          # What benchmark tool the output.txt came from
+          tool: 'go'
+          # Where the output from the benchmark tool is stored
+          output-file-path: bench_output.txt
+          # Where the previous data file is stored
+          external-data-json-path: ./cache/benchmark-data.json
```
miparnisari, line 103: Shouldn't this be `./cache/bench_output.txt`?

.github/workflows/pull_request.yaml
craigpastro: It's `go test -bench` so you know 😆

.github/workflows/pull_request.yaml
craigpastro: This file is different and contains the previous runs results. https://github.com/benchmark-action/github-action-benchmark#external-data-json-path-optional

I am kind of wondering how this works between branches? I think we may have to add a `main` workflow.

craigpastro: Okay, updated to only compare on PRs and save on push to main.

.github/workflows/pull_request.yaml
miparnisari: jeez i hate Go authors right now

.github/workflows/pull_request.yaml
```diff
@@ -53,3 +53,58 @@ jobs:
...
+        run: go test ./... -bench=. -run=XXX | tee ${{ github.sha }}_bench_output.txt
+
+      # Download previous benchmark result from cache (if exists)
+      - name: Download previous benchmark data
+        uses: actions/cache@v1
+        with:
+          path: ./cache
+          key: ${{ runner.os }}-benchmark
+
+      - name: Compare benchmark result
```
miparnisari, line 95: compare against what? master?

.github/workflows/pull_request.yaml
miparnisari: > This file is different and contains the previous runs results. https://github.com/benchmark-action/github-action-benchmark#external-data-json-path-optional

Right but in the `main` workflow you are saving the benchmark results to `output-file-path: bench_output.txt`... i don't understand where the json comes from 😕

.github/workflows/pull_request.yaml
craigpastro: Yes. Compare against `./cache/benchmark-data.json` which will be saved on push to main.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,59 @@
+name: OpenFGA Main
```
jon-whit, line 1: Can we just call this "main". The OpenFGA part goes without saying, because this lives in the openfga project.

.github/workflows/pull_request.yaml
jon-whit: My two cents.. We're only saving a few characters. I'd recommend we just be explicit and say `benchmark` or `go-bench` (my preferred recommendation).

.github/workflows/main.yaml
craigpastro: Sure. I'll do the same for "Pull Request". I leave the uppercase because that seems to be the convention, at least https://docs.github.com/en/actions/quickstart.

.github/workflows/pull_request.yaml
craigpastro: That also works for me. Updated. Thanks!

jon-whit (approved):
LGTM. Let's make sure this compares regressions against `main` on subsequent PRs.

PR #15, opened 2022-06-10
chore: add Close to service

craigpastro:
## Review Checklist

### DIFF
--- a/cmd/openfga/openfga.go
+++ b/cmd/openfga/openfga.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openfga/openfga/storage/caching"
 	"github.com/openfga/openfga/storage/memory"
 	"github.com/openfga/openfga/storage/postgres"
+	"go.uber.org/multierr"
 	"go.uber.org/zap"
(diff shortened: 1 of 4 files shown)


cmd/openfga/openfga.go
```diff
@@ -34,11 +35,20 @@ var (
...
-	openFgaServer *server.Server
+	server        *server.Server
 	datastore     storage.OpenFGADatastore
 	authenticator authentication.Authenticator
 }
 
+func (s *service) Close(ctx context.Context) error {
+	s.authenticator.Close()
+
+	return multierr.Combine(
```
jon-whit, line 46: Let's not bring a whole other external package in just for this.

```
if err := s.server.Close(); err != nil {
    return err
}

if err := s.datastore.Close(ctx); err != nil {
    return err
}
```

is fine.

craigpastro: I just realized that server.Close doesn't need an error, so I updated that instead.

jon-whit (approved):

PR #19, opened 2022-06-11
fix: ignore mocks from code coverage

miparnisari:
## Description
ignore mock files (both hand-written and autogenerated) from code coverage

## References
n/a

## Review Checklist


cmd/openfga/openfga_test.go
```diff
@@ -11,8 +11,8 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
-	"github.com/openfga/openfga/mocks"
 	"github.com/openfga/openfga/pkg/logger"
+	"github.com/openfga/openfga/server/authentication/mocks"
```
jon-whit, line 15: Why not leave this under the top-level (shared) mocks package? It will make it easier to skip package coverage on all mocks (generated and manually defined).

miparnisari: I tried to keep the mocks together with the code that they are mocking, but this works too.

miparnisari: Actually.. 

 > you'll only have to ignore the mocks directory

i'm ignoring the `mocks` pattern in directories, so 🤷‍♀️

elbuo8 (approved):

PR #22, opened 2022-06-13
chore: gracefully close HTTP server and gRPC server in Run

jon-whit:
## Description
This PR cleans up the `Close` method of the OpenFGA server in favor of context based closure in the run method. We had both ways of handling server closure and it wasn't clear which one should be used.

The intended usage is, for example,

```
func main() {
    ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
    defer stop()

    var server *server.Server

    go func() {
        if err := server.Run(ctx); err != nil {
            // handle error
        }
    }()

    <-ctx.Done()
}
```

## Review Checklist


craigpastro (dismissed):
LGTM

cmd/openfga/openfga_test.go
```diff
@@ -64,11 +63,12 @@ func TestBuildServerWithPresharedKeyAuthentication(t *testing.T) {
 	require.NoError(t, err, "Failed to build server and/or datastore")
 	defer service.Close(ctx)
 
-	g, ctx := errgroup.WithContext(ctx)
+	ctx, cancel := context.WithCancel(ctx)
```
jon-whit, line 66: This fixes a bug that existed before. `errgroup.WithContext(...)` only cancels the context when the first non-nil error is returned from the Go call or when `g.Wait()` returns. None of these conditions apply, because the context never gets cancelled and thus the server never stops running, and so the Go function never returns.

This changes this behavior and fixes a bug so that tests don't accidentally overlap one another (since they run on the same port).

craigpastro (approved):
LGTM

craigpastro (dismissed):
LGTM

craigpastro (approved):
LGTM

PR #24, opened 2022-06-14
refactor: graph evaluation and dispatch composition

jon-whit:
## Description
These changes introduce a new structure for resolving the relationship graph that is more flexible. These changes include a dispatch interface that allows us to inject behavior into the graph resolution process which will allow us to, for example, dispatch graph subproblems to other nodes in a cluster (if we so choose), integrate a dispatcher with a caching layer in it, and/or  delegate to a concurrent (but local) graph evaluation lookup (like we're simply doing today). We can also combine these dispatchers by chaining them together to implement a combined dispatcher that involves distributing requests to peers using something like consistent hashing in addition to caching.

I thought this implementation is much easier to understand and more elegant to compose than what we're doing today with the `resolutionContext`. Our `resolutionContext` strategy today is hard to understand because of the `rc.fork(..)` and shared state held across goroutines and synchronized by both channels and mutexes. This simplifies the implementation by better leveraging the composition of functions that take write channels, combining them with function chaining, and the use of `errgroup` to bring it all together.

## References
The implementation was largely inspired by the `graph` and `dispatch` packages in `authzed/spicedb`. 

## Review Checklist


jon-whit:
Needs a rebase on main

jon-whit:
[ ] - Wire up the resolution tracer so that we have the check resolution path.
[ ] - Add the metrics back into the graph resolution process.

internal/graph/check.go
```diff
@@ -0,0 +1,485 @@
...
+		return err
+	}
+}
+
+// reduce reduces the check request by breaking it into an appropriate subproblem that
+// is resolved by evaluating the underlying userset rewrite rule.
+func (cc *ConcurrentChecker) reduce(
+	ctx context.Context,
+	req *dispatchpb.DispatchCheckRequest,
+	rewrite *openfgapb.Userset,
```
jon-whit, line 61: I don't love that we've named our userset rewrites (vernacular of the Zanzibar paper) as simply `Userset`. The term `Userset` has become massively overloaded depending on the context. And our definition of a `Userset` rewrite rule is not very recursive or uniform either. It'd be nice to move toward a definition more like the following:

```
message UsersetRewrite {
    oneof rewrite_operation {
        SetOperation union = 1;
        SetOperation intersection = 2;
        SetOperation exclusion = 3;
    }
}

message SetOperation {
    message Child {
        message This {}
        message Nil {}

        oneof child_type {
            This _this = 1;
            ComputedUserset computed_userset = 2;
            TupleToUserset tuple_to_userset = 3;
            UsersetRewrite userset_rewrite = 4;
            Nil _nil = 6;
        }
    }
}
```

internal/graph/check.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+func (cc *ConcurrentChecker) checkDirect(ctx context.Context, req *dispatchpb.DispatchCheckRequest) ReduceableCheckFunc {
+
+	return func(ctx context.Context, resultChan chan<- *dispatchpb.DispatchCheckResponse) error {
+		g := new(errgroup.Group)
+
+		g.Go(func() error {
+
+			tuple, err := cc.datastore.ReadUserTuple(ctx, req.WrappedRequest.StoreId, req.WrappedRequest.TupleKey)
```
jon-whit, line 102: I think it's time for our storage interface to evolve and be refactored a bit. We have code all over the place that yields an `TupleIterator` only to throw away the tuple for the tuple key (e.g. `tuple.GetKey()`). So far I have seen 0 usage of the `tuple.Timestamp`, so I vote we refactor our storage interface in favor of a simpler and more elegant interface that just returns the things we actually need. This way we have less code that is an indirection to the thing that we actually want.

internal/graph/check.go
```diff
@@ -0,0 +1,485 @@
...
+	return func(ctx context.Context, resultChan chan<- *dispatchpb.DispatchCheckResponse) error {
+
+		relation := rw.TupleToUserset.GetTupleset().GetRelation()
+		if relation == "" {
+			relation = req.GetWrappedRequest().GetTupleKey().GetRelation()
+		}
+
+		var requestsToDispatch []ReduceableCheckFunc
+		for {
+			iter, err := cc.datastore.Read(
```
jon-whit, line 207: Our storage interface could also use a little rework in this regard as well. It would be nice if our storage interface took optional filter parameters that you could provide to filter by the different filters that you may want instead of passing in a `TupleKey` with omitted fields to get the filter you desire. This behavior of the storage interface is nuanced and unclear and it would make our code read easier if there was a single read method such as 

```
type RelationshipFilter struct {
    ObjectType string
    OptionalObjectID string
    OptionalRelation string
    OptionalUserFilter string
}

func QueryRelationships(
    ctx context.Context,
    filter *RelationshipFilter,
    opts ...QueryOptionsOption,
) (*RelationshipIterator, error) {...}
```

internal/graph/check.go
```diff
@@ -0,0 +1,500 @@
...
+					WrappedResponse: &openfgapb.CheckResponse{
+						Allowed: false,
+					},
+				}, nil
+			}
+		case sub := <-subtractChan:
+			responseMetadata = combineResponseMetadata(responseMetadata, sub.Metadata)
+
+			if sub.WrappedResponse.Allowed {
+				return &dispatchpb.DispatchCheckResponse{
```
jon-whit, line 469: I plan on cleaning up these responses. Instead of inline instantiating the response objects, I plan on invoking a helper that will make reading this easier.

For example, `checkResponse(metadata, allowed) *dispatchpb.DispatchCheckResponse`

jon-whit:
Performed some preliminary benchmarks and found that these changes lead to additional heap allocations that caused a significant amount of GC pressure. Overall the performance of Check was diminished. Although the composition and readability improved, the performance degradation isn't worth the change.

For now I'm marking this as closed. It was a good exploration nonetheless.

jon-whit:
For reference, the benchmarks that were run are defined below:

```
# after
➜  openfga git:(graph-dispatcher) ✗ go test -cpuprofile cpuprofile.out -memprofile memprofile.out -benchmem -count=5 -run=^$ -bench BenchmarkOpenFGAServer/TestMemoryDatastore/ ./server
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/server
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   22642	     47808 ns/op	   40718 B/op	     641 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   32038	     37316 ns/op	   40575 B/op	     641 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   33388	     36142 ns/op	   40520 B/op	     641 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   35355	     35902 ns/op	   40673 B/op	     641 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   33944	     37152 ns/op	   40806 B/op	     641 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   24348	     45394 ns/op	   46992 B/op	     771 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   28252	     44576 ns/op	   45932 B/op	     771 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   20703	     54264 ns/op	   45917 B/op	     770 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   19468	     55117 ns/op	   45921 B/op	     770 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   19546	     55130 ns/op	   47994 B/op	     770 allocs/op

# before
➜  openfga git:(graph-dispatcher) ✗ go test -cpuprofile cpuprofile.out -memprofile memprofile.out -benchmem -count=5 -run=^$ -bench BenchmarkOpenFGAServer/TestMemoryDatastore/ ./server
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/server
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   52742	     21984 ns/op	    9319 B/op	     186 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   56025	     21274 ns/op	    9332 B/op	     186 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   54915	     21379 ns/op	    9327 B/op	     186 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   56046	     20713 ns/op	    9329 B/op	     186 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkCheckWithoutTrace-10         	   57128	     20606 ns/op	    9333 B/op	     186 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   52317	     22205 ns/op	   11172 B/op	     261 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   53200	     22220 ns/op	   11185 B/op	     261 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   53659	     22236 ns/op	   11156 B/op	     260 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   53799	     21885 ns/op	   11167 B/op	     261 allocs/op
BenchmarkOpenFGAServer/TestMemoryDatastore/BenchmarkWithTrace-10                 	   53848	     22106 ns/op	   11171 B/op	     261 allocs/op
```

The number of heap allocations increased and the operations per second decreased with the changes introduced in this experimental PR. Using pprof to profile the differences, it appeared that the majority of the computation time was being bound to GC because of the heap allocations involved with the reduceable check functions. Although the code is easier to read, it probably doesn't warrant the refactor at this time because of this issue.

PR #31, opened 2022-06-15
chore: introduce "development" and "production" logging

miparnisari:
## Description

image

## References
n/a

## Review Checklist

### DIFF
--- a/cmd/openfga/openfga.go
+++ b/cmd/openfga/openfga.go
@@ -72,6 +72,9 @@ type svcConfig struct {
 	// OIDC authentication
 	AuthOIDCIssuer   string `default:"" split_words:"true"`
 	AuthOIDCAudience string `default:"" split_words:"true"`
+
+	// Logging
+	DevelopmentLogging bool `default:"true" split_words:"true"`
 }
 
 func main() {
@@ -79,15 +82,10 @@ func main() {
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
 
-	logger, err := logger.NewZapLogger()
+	logger, err := buildLogger()
 	if err != nil {
 		log.Fatalf("failed to initialize logger: %v", err)
 	}
-
-	logger.With(
-		zap.String("build.version", version),
-		zap.String("build.commit", commit),
-	)
 
 	service, err := buildService(logger)
 	if err != nil {
@@ -122,14 +120,19 @@ func main() {
 	logger.Info("Server exiting. Goodbye 👋")
 }
 
-func buildService(logger logger.Logger) (*service, error) {
+func getServiceConfig() svcConfig {
 	var config svcConfig
-	var err error
-	var datastore storage.OpenFGADatastore
 
 	if err := envconfig.Process("OPENFGA", &config); err != nil {
-		return nil, fmt.Errorf("failed to process server config: %v", err)
+		log.Fatalf("failed to process server config: %v", err)
 	}
+	return config
+}
+
+func buildService(logger logger.Logger) (*service, error) {
+	var err error
+	var datastore storage.OpenFGADatastore
+	config := getServiceConfig()
 
 	tracer := telemetry.NewNoopTracer()
 	meter := telemetry.NewNoopMeter()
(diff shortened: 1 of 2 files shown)


cmd/openfga/openfga.go
```diff
@@ -72,23 +72,21 @@ type svcConfig struct {
 	// OIDC authentication
 	AuthOIDCIssuer   string `default:"" split_words:"true"`
 	AuthOIDCAudience string `default:"" split_words:"true"`
+
+	// Logging
+	DevelopmentLogging bool `default:"true" split_words:"true"`
```
elbuo8, line 77: Should the default be `true`?

cmd/openfga/openfga.go
```diff
@@ -203,3 +207,28 @@ func buildService(logger logger.Logger) (*service, error) {
 		authenticator: authenticator,
 	}, nil
 }
+
+func buildLogger() (*logger.ZapLogger, error) {
+	var err error
+	var openfgaLogger *logger.ZapLogger
```
elbuo8, line 213: Should the initialization default to the default value of the config and then branch into the non default?

cmd/openfga/openfga.go
```diff
@@ -203,3 +207,28 @@ func buildService(logger logger.Logger) (*service, error) {
 		authenticator: authenticator,
 	}, nil
 }
+
+func buildLogger() (*logger.ZapLogger, error) {
```
jon-whit, line 211: Why not `logger.Logger`? We should return the interface, not the concrete type here. That way we can change the concrete type that we return in the future if we so choose.

cmd/openfga/openfga.go
```diff
@@ -203,3 +207,28 @@ func buildService(logger logger.Logger) (*service, error) {
...
+	var err error
+	var openfgaLogger *logger.ZapLogger
+
+	config := getServiceConfig()
+
+	if config.DevelopmentLogging {
+		openfgaLogger, err = logger.NewDevelopmentLogger()
```
jon-whit, line 218:
```suggestion
		l, err = logger.NewDevelopmentLogger()
```

cmd/openfga/openfga.go
```diff
@@ -203,3 +207,28 @@ func buildService(logger logger.Logger) (*service, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		openfgaLogger, err = logger.NewProductionLogger()
```
jon-whit, line 223:
```suggestion
		l, err = logger.NewProductionLogger()
```

cmd/openfga/openfga.go
miparnisari: [whenever you notice something like that, a wizard did it](https://www.youtube.com/watch?v=sVgVB3qsySQ)

cmd/openfga/openfga.go
miparnisari: Yeah i think so. When people follow the `getting started` section of the readme, they want to see readable logs, and json logs aren't readable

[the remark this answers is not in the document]
miparnisari: I thought it was better to keep it generic. A development logging involves text output, to the console, and with `debug` level (for example). The production logging could mean json, to some file, and with only `warn`...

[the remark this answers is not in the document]
jon-whit: That conflates the two though. You have a config `LogLevel` which can be one of (trace, debug, info, warn, error, fatal, etc..) and you have the log output format `LogFormat` which can e one of `(text, json, etc..)`. It just doesn't make sense to introduce deployment environment terms for log output format. I may want to run text log output in production and I may like to run openfga with json log output in development.

cmd/openfga/openfga.go
```diff
@@ -210,6 +210,8 @@ func buildService(logger logger.Logger) (*service, error) {
 		return nil, fmt.Errorf("failed to initialize authenticator: %v", err)
 	}
 
+	logger.Info(fmt.Sprintf("using '%v' authentication", config.AuthMethod))
```
jon-whit, line 213: If `config.AuthMethod == "none"` can we emit a warning instead of an info statement?

jon-whit (approved):

PR #32, opened 2022-06-16
fix: remove arrows in codeblocks

matthewpereira:
Remove ➜ characters from README.md codeblocks

## Description
* removed styling characters so that codeblocks copy-pasted to terminal wouldn't error
* syntax highlighted codeblocks
* separated commands so that they could be copied and pasted sequentially - documentation should always ask users to do things one step at a time wherever possible.

Before:
Screen Shot 2022-06-16 at 10 20 34 AM

After:
Screen Shot 2022-06-16 at 10 20 13 AM

## Review Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,11 @@ You can install from source using Go modules (make sure `$GOBIN` is on your shel
 
 ```bash
 export PATH=$PATH:$(go env GOBIN)
+```
+
+Then:
+
+```bash
 go install github.com/openfga/openfga/cmd/openfga
 ```
 


matthewpereira:
I'm going to make a separate PR proposing changes to the 'Install from Source' section, since I think it's confusing that two sets of instructions are described sequentially without a heading or break to indicate that building from a cloned project isn't a continuation of installing with Go modules unless you're reading closely.

curfew-marathon (approved):

PR #43, opened 2022-06-16
fix(makefile): warning: no packages being tested depend on matches for pattern

miparnisari:
## Description

https://github.com/golang/go/blob/78b722d8c2f764c3048c6f0344e9ebcd2687813d/src/cmd/go/internal/test/test.go#L852

Fix warning by trimming the last `,` from the `COVERPKG` variable

## References
n/a

## Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .DEFAULT_GOAL := help
-COVERPKG := $(shell eval go list ./... | grep -v mocks | tr '\n' ',')
+COVERPKG := $(shell eval go list ./... | grep -v mocks | tr '\n' ',' | rev | cut -c 2- | rev)
 
 .PHONY: help
 help:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,5 @@
 .DEFAULT_GOAL := help
-COVERPKG := $(shell eval go list ./... | grep -v mocks | tr '\n' ',')
```
jon-whit, line 2: This can be simplified to `go list ./... | grep -v mocks |  paste -sd "," -`

So `$(shell eval go list ./... | grep -v mocks |  paste -sd "," -)` 

This avoids the issue of the two trailing characters in the first place.

Makefile
```diff
@@ -1,5 +1,5 @@
...
+COVERPKG := $(shell eval go list ./... | grep -v mocks | tr '\n' ',' | rev | cut -c 2- | rev)
```
jon-whit, line 2:
```suggestion
COVERPKG := $(shell eval go list ./... | grep -v mocks |  paste -sd "," -)
```

This can be simplified to `go list ./... | grep -v mocks |  paste -sd "," -`

This avoids the issue of the two trailing characters in the first place.

jon-whit (approved):

PR #57, opened 2022-06-20
fix: run test on a different port

craigpastro:
## Description

Running the TLS test on the same port as the server usually runs on sometimes causes an error as it probably tries to interact with a server that is still running from another test. In the future I think we will dockerize the tests and the server will run on a random port, but until then, have the problematic test run on a different port.

## Review Checklist


craigpastro:
fixed in a different way in #58

PR #63, opened 2022-06-21
chore: get config `config.yaml` file and use viper

jon-whit:
## Description
This PR adds structured config to OpenFGA. The changes herein pull config from a `config.yaml` file by looking in the directories `/etc/openfga`, `$HOME/.openfga`, and `.` (the current working directory), in that order. Future work will also wire up the config via CLI flags so that the developer can set config via a file or via CLI flags (or both if they so choose).

## Review Checklist


pkg/cmd/service/service.go
```diff
@@ -20,59 +18,126 @@ import (
...
-	DatastoreEngine               string `default:"memory" split_words:"true"`
-	DatastoreConnectionURI        string `split_words:"true"`
-	DatastoreMaxCacheSize         int    `default:"100000" split_words:"true"`
-	ServiceName                   string `default:"openfga" split_words:"true"`
-	HTTPPort                      int    `default:"8080" split_words:"true"`
-	RPCPort                       int    `default:"8081" split_words:"true"`
-	MaxTuplesPerWrite             int    `default:"100" split_words:"true"`
-	MaxTypesPerAuthorizationModel int    `default:"100" split_words:"true"`
+
+	DatabaseConfig `mapstructure:"database"`
```
craigpastro, line 75: Since these are in the `Config` struct does it make sense to drop the `Config` postfix. Then we can drop the `mapstructure` tag as well.

server/server.go
```diff
@@ -473,7 +472,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	httpServer := &http.Server{
-		Addr: fmt.Sprintf(":%d", s.config.HTTPServer.Addr),
+		Addr: fmt.Sprintf(":%d", s.config.HTTPServer.Port),
```
craigpastro, line 475: Interesting. I guess I never noticed this before. Is `Port` sufficient though? Won't there be some case where a user will want to use the the entire `Addr`?

go.mod
```diff
@@ -57,13 +71,14 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/sirupsen/logrus v1.7.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/spf13/viper v1.12.0
```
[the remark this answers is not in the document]
jon-whit: This may have been placed where it was because of the auto import 🤷 ? I have moved it above.

config-schema.json
```diff
@@ -0,0 +1,127 @@
+{
```
jon-whit, line 1: @craigpastro @miparnisari do you think we should try and autogenerate our default Config struct based on the "default" values in this `config-schema.json` file?

It would make this the source of truth, which would be good for documentation, but it would be a bit more complicated of an integration in this project if we did so.

Personally, I don't mind having to update the `config-schema.json` file along with the changes to the default values (if we make changes to the defaults). At least this now lives alongside the source code, and we can generate doc in openfga.dev off of this file.

pkg/cmd/service/service.go
craigpastro: Oh, I see, you want to embed the configs in this struct. In general I like being more explicit in which case we could have
```
Database DatabaseConfig
```
here, then the configuration stuctures would have clear names. 

In this case, how do you avoid collisions with `Port` and `Enabled` for example?

config-schema.json
craigpastro: I was playing around with something yesterday and found that the grcp backoff package provides a [DefaultConfig](https://pkg.go.dev/google.golang.org/grpc/backoff#pkg-variables), and I thought that was a very cool way to specify a default. Would something like that make integration easier? Or we could even make a `config` package and a `func Default() *Config` function or something like that?

I feel something like this may easily get out of date.

pkg/cmd/service/service.go
jon-whit: The collisions are avoided because the embedded struct is scoped by it's defining type.

For example,

```
type DatabaseConfig struct {
    Port int
}

type OtherConfig struct {
    Port int
}

type Config struct {
    DatabaseConfig
    OtherConfig
}
```

the `Port` field is actually namespaced as `Config.DatabaseConfig.Port` and `Config.OtherConfig.Port`, so there are no collisions.

config-schema.json
jon-whit: Yeah, the `func Default() *Config` would serve the same purpose that the `GetServiceConfig()` func today is serving.

The purpose of the `config-schema.json` is so that our documentation and other sources can generate doc based on the schema. If we wanted to show a default value column in a table we'd want to use the "default" keyword in our json-schema wherever it is applicable to a config with a default value. As you pointed out though, it would be easy for the defaults in `config-schema.json` to get out of date with the actual defaults encoded in the `GetServerConfig()` function. This conundrum is the basis of my question.

We could, within the `GetServiceConfig` function, use the `config-schema.json` to generate the default values. That would be possible, but it's also a little overkill if you ask me, because we'd have to include the json file in our containers and we'd have to map a json-schema to a config struct in Go etc.. It just seems like a lot for a little benefit 🤷

go.mod
```diff
@@ -31,6 +31,21 @@ require (
 	google.golang.org/protobuf v1.28.0
 )
 
+require (
```
craigpastro, line 34: Nit: can we combine this require block with the one below?

pkg/cmd/service/service.go
```diff
@@ -74,13 +109,60 @@ type Config struct {
...
 }
 
-func GetServiceConfig() Config {
-	var config Config
+// GetServiceConfig returns the OpenFGA server configuration based on the values provided in the server's 'config.yaml' file.
+// The 'config.yaml' file is loaded from '/etc/openfga', '$HOME/.openfga', or the current working directory. If no configuration
+// file is present, the default values are returned.
+func GetServiceConfig() (Config, error) {
+
+	config := Config{
```
craigpastro, line 117: Can we make a `func DefaultConfig() Config`  that returns this? I think that would be useful, especially for tests.

jon-whit: 👍 done!

craigpastro: Ah yeah!

go.mod
jon-whit: Fixed! I think this issue actually comes from merge conflicts being resolved slightly wrong. If we resolve a merge conflict but retain multiple require blocks, then we find oursevles in a situation where we have 2+ require blocks in `go.mod`.

Something for us to keep an eye on in the future.

pkg/cmd/service/service.go
miparnisari: vote: option 1

jon-whit: 👍 sounds good. Thanks for the vote!

pkg/cmd/service/service.go
```diff
@@ -21,68 +19,191 @@ import (
...
+	viper.SetConfigName("config")
+	viper.SetConfigType("yaml")
 
-	PlaygroundEnabled bool `default:"true" split_words:"true"`
-	PlaygroundPort    int  `default:"3000" split_words:"true"`
+	configPaths := []string{"/etc/openfga", "$HOME/.openfga", "."}
+	for _, path := range configPaths {
+		viper.AddConfigPath(path)
+	}
+	viper.AutomaticEnv()
```
craigpastro, line 189: Do we still need this?

pkg/cmd/service/service.go
```diff
@@ -21,68 +19,191 @@ import (
...
 
-func GetServiceConfig() Config {
-	var config Config
+		return config, fmt.Errorf("failed to load server config: %w", err)
+	}
 
-	if err := envconfig.Process("OPENFGA", &config); err != nil {
-		log.Fatalf("failed to process server config: %v", err)
+	if err := viper.Unmarshal(&config); err != nil {
+		return config, fmt.Errorf("failed to unmarshal server config: %w", err)
```
craigpastro, line 203: Should we return an empty config if we also return an error?

server/server.go
```diff
@@ -409,6 +408,8 @@ func (s *Server) ListStores(ctx context.Context, req *openfgapb.ListStoresReques
 	ctx, span := s.tracer.Start(ctx, "listStores")
 	defer span.End()
 
+	time.Sleep(2 * time.Second)
```
craigpastro, line 411: This crept in?

server/server.go
```diff
@@ -466,7 +467,7 @@ func (s *Server) Run(ctx context.Context) error {
 	var httpServer *http.Server
 	if s.config.HTTPServer.Enabled {
 		// Set a request timeout.
-		runtime.DefaultContextTimeout = s.config.RequestTimeout
+		runtime.DefaultContextTimeout = 3 * time.Second //s.config.RequestTimeout
```
craigpastro, line 470: This also doesn't look quite right.

pkg/cmd/service/service.go
jon-whit: I usually like to return nil values if an error is encountered, but the signature is currently `(Config, error)`.

I can change it to return a `Config{}, ...`, but what do you think about changing the signature to a pointer return value and just returning `nil`?

pkg/cmd/service/service.go
jon-whit: I guess if you want to provide env variables and they automatically map to the config values then it's ok. In a subsequent PR we will add support for other options (CLI flags and possibly env variables with a better mapping strategy)

server/server.go
jon-whit: Oops. lol

I was testing out the request timeout. Good catch :)

server/server.go
jon-whit: Good call. I was testing things out. Thanks for catching my mistake!

pkg/cmd/service/service.go
craigpastro: That works for me 👍

jon-whit: Ok, I made that change 👍

craigpastro (approved):
Nice 🎉  LGTM

Issue #64, opened 2022-06-21
feat: allow specifying CORS AllowedOrigins


adriantam:
Currently, openfga hard coded CORS Allowed Origins to be []string{"*"}.  It should be possible to allow specifying the allowed origins to be a more restrictive value.

PR #65, opened 2022-06-21
feat: allow specifying CORS allowed origins

adriantam:
## Description
Allow overriding the default CORS allowed origins via variable `OPENFGA_CORS_ALLOWED_ORIGINS`

## References
Close https://github.com/openfga/openfga/issues/64

The corresponding docs change is  https://github.com/openfga/openfga.dev/pull/114

## Review Checklist

### DIFF
--- a/pkg/cmd/service/service_test.go
+++ b/pkg/cmd/service/service_test.go
@@ -9,6 +9,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
@@ -222,6 +223,18 @@ func TestBuildServerWithPresharedKeyAuthentication(t *testing.T) {
 	cancel()
 	require.NoError(t, g.Wait())
 	require.NoError(t, service.Close(ctx))
+}
+
+func TestSettingCORSAllowedOrigins(t *testing.T) {
+	os.Setenv("OPENFGA_AUTH_METHOD", "preshared")
+	os.Setenv("OPENFGA_AUTH_PRESHARED_KEYS", "KEYONE,KEYTWO")
+	os.Setenv("OPENFGA_CORS_ALLOWED_ORIGINS", "http://openfga.dev,http://localhost")
+
+	corsAllowedOrigins := GetServiceConfig().CORSAllowedOrigins
+	expectedCORSAllowedOrigins := []string{"http://openfga.dev", "http://localhost"}
+	if !reflect.DeepEqual(corsAllowedOrigins, expectedCORSAllowedOrigins) {
+		t.Fatalf("Unexpected CORSAllowedOrigin expected %v, actual %v", corsAllowedOrigins, expectedCORSAllowedOrigins)
+	}
 }
 
 func TestBuildServerWithOidcAuthentication(t *testing.T) {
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -475,7 +476,7 @@ func (s *Server) Run(ctx context.Context) error {
 	httpServer := &http.Server{
 		Addr: fmt.Sprintf(":%d", s.config.HTTPServer.Addr),
 		Handler: cors.New(cors.Options{
-			AllowedOrigins:   []string{"*"},
+			AllowedOrigins:   s.config.HTTPServer.CORSAllowedOrigins,
```
craigpastro, line 479: Do you think we should also allow the other options to be configurable as well?

adriantam: `AllowedHeaders` should be configurable.  As for `AllowCredentials` and `AllowedMethods`, we probably don't want to change as they are related to our authentication + provided methods.

pkg/cmd/service/service_test.go
```diff
@@ -224,6 +225,115 @@ func TestBuildServerWithPresharedKeyAuthentication(t *testing.T) {
...
+			req.Header.Set("Origin", test.args.origin)
+			req.Header.Set("Access-Control-Request-Method", "OPTIONS")
+			req.Header.Set("Access-Control-Request-Headers", test.args.header)
+			retryClient := http.Client{}
+			res, err := retryClient.Do(req)
+			require.NoError(t, err, "Failed to execute request")
+
+			origin := res.Header.Get("Access-Control-Allow-Origin")
+			acceptedHeader := res.Header.Get("Access-Control-Allow-Headers")
+			if origin != test.want.origin {
```
craigpastro, line 315: Can use `require.Equal` here, and elsewhere, if you would like. I think it is a bit cleaner.

craigpastro (approved):
LGTM. Left one comment about `require.Equal` if you would like.

craigpastro (approved):

pkg/cmd/service/service_test.go
```diff
@@ -224,6 +225,112 @@ func TestBuildServerWithPresharedKeyAuthentication(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			payload := strings.NewReader(`{"name": "some-store-name"}`)
+			req, err := http.NewRequest("OPTIONS", fmt.Sprintf("%s/stores", openFgaServerURL), payload)
+			require.NoError(t, err, "Failed to construct request")
+			req.Header.Set("content-type", "application/json")
+			req.Header.Set("authorization", "Bearer KEYTWO")
+			req.Header.Set("Origin", test.args.origin)
+			req.Header.Set("Access-Control-Request-Method", "OPTIONS")
```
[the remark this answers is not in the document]
adriantam: No, they are different.  Line 302 is the request method itself.  This one is for the pre-flight to check whether the request method is valid.  You can have
``` 
req, err := http.NewRequest("OPTIONS", fmt.Sprintf("%s/stores", openFgaServerURL), payload)
```
but check for
```
req.Header.Set("Access-Control-Request-Method", "GET")
```

In this case, the client will send a OPTION request checking whether GET is accepted.

PR #68, opened 2022-06-23
feat: embedded playground

matldupont:
Embedded Playground accessing the local OpenFGA instance

## Description
- Serving a local Playground on `localhost` and making use of the hosted `play.fga.dev` version via `<iframe />`
- Added a Playground section to the README
- Added a Table of Contents to the README and refactored for semantic html


adriantam:
Do we need to update https://openfga.dev/docs/getting-started/setup-openfga#configuring-the-server (i.e., through https://github.com/openfga/openfga.dev/blob/main/docs/content/getting-started/setup-openfga.mdx)?

Dockerfile
```diff
@@ -8,4 +8,6 @@ RUN go build -o ./openfga ./cmd/openfga
 FROM alpine as final
 EXPOSE 8080
 COPY --from=builder /app/openfga /app/openfga
-ENTRYPOINT ["/app/openfga"]
\ No newline at end of file
+COPY --from=builder /app/static /app/static
+WORKDIR /app
```
craigpastro, line 12: Do we need this here? And, if so, should we change the entrypoint?

README.md
```diff
@@ -3,29 +3,42 @@
 [![Go Reference](https://pkg.go.dev/badge/github.com/openfga/openfga.svg)](https://pkg.go.dev/github.com/openfga/openfga)
 ![GitHub release (latest SemVer)](https://img.shields.io/github/v/release/openfga/openfga?sort=semver&color=green)
 [![Container Image](https://img.shields.io/github/v/release/openfga/openfga?color=blueviolet&label=container&logo=docker "Container Image")](https://hub.docker.com/r/openfga/openfga/tags)
-![Downloads](https://img.shields.io/github/downloads/openfga/openfga/total.svg?style=flat&color=lightgrey)  <!-- render new line with double space here -->  
+![Downloads](https://img.shields.io/github/downloads/openfga/openfga/total.svg?style=flat&color=lightgrey) <!-- render new line with double space here -->
```
craigpastro, line 6: If we are removing the double space then can remove the comment as well please?

README.md
```diff
@@ -91,17 +104,20 @@ go build cmd/openfga/openfga.go
...
 docker compose up -d postgres
 make run-postgres
 ```
+
 This should start a Postgres container, write the database schema, and start the OpenFGA server.
 
 When you are done you can stop the Postgres container with:
-```bash
+
+````bash
```
craigpastro, line 117:
```suggestion
```bash
```

craigpastro: Also I think we are not based off main, because there should be a closing ``` after the `docker compose down`.

README.md
```diff
@@ -111,7 +127,7 @@ curl -X POST 'localhost:8080/stores' \
 --data-raw '{
     "name": "openfga-demo"
 }'
-```
+````
```
craigpastro, line 130:
```suggestion
```
```
Or does four ticks work? Is this intentional?

File: README.md [hunk 5]
```diff
@@ -124,6 +140,30 @@ If everything is running correctly, you should get a response with information a
 }
 ```
 
+## Playground
+
+OpenFGA includes a playground use to facilitate the creation and visualisation of authorization models and tuples within your running local instance.
+
+Once OpenFGA is running, the playground can be accessed at [http://localhost:3000/playground].
+
+In the event that a port other than the default `:3000` is required, the `OPENFGA_PLAYGROUND_PORT` environment can be set when starting the instance.
```
craigpastro, line 149:
```suggestion
In the event that a port other than the default `:3000` is required, the `OPENFGA_PLAYGROUND_PORT` environment variable can be set when starting the instance.
```

README.md
```diff
@@ -124,6 +140,30 @@ If everything is running correctly, you should get a response with information a
...
+
+eg:
```
craigpastro, line 151: I think this is getting confusing having so many examples of running OpenFGA. I think most people know how to set env vars, so perhaps the example can be dropped? WDYT?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,247 @@
+package main
```
craigpastro, line 1: I think this file has been moved

pkg/cmd/service/service.go
```diff
@@ -65,6 +65,9 @@ type Config struct {
 
 	// Logging. Possible options: text,json
 	LogFormat string `default:"text" split_words:"true"`
+
+	PlaygroundEnabled bool `default:"true" split_words:"true"`
```
craigpastro, line 69: Do we want this on by default? Does it make sense to have it off by default, and add a command to the Makefile such as `run-with-playground` that turns it on.

static/playground/index.html
```diff
@@ -0,0 +1,15 @@
...
+      margin: 0;
+    }
+  </style>
+</head>
+
+<body>
+  <iframe src="https://play.fga.dev/sandbox/?fga_api_host={{.HTTPServerURL}}&fga_api_scheme=http" frameborder="0" marginheight="0" marginwidth="0" width="100%" height="100%" scrolling="auto"></iframe>
+</body>
+
+</html>
```
craigpastro, line 15: Newline?

README.md
matldupont: I assume nothing!

README.md
matldupont: sure!

static/playground/index.html
matldupont: NBD, but can remove it

pkg/cmd/service/service.go
matldupont: @jon-whit can you answer this one?

cmd/openfga/openfga_test.go
matldupont: @jon-whit can you answer this one?

static/playground/index.html
craigpastro: I meant should we add one? It gets rid of that "no newline" mark which always bothers me (but I think I am the only one).

README.md
craigpastro: lol. If you want to keep them can we move the `eg:` to the same line then?

> ... starting the instance. E.g.,

Dockerfile
jon-whit: The binary uses a relative path for the static assets, so it's important that we set our working directory to the same directory where the static assets are located.

README.md
matldupont: Sure!

Dockerfile
```diff
@@ -8,4 +8,6 @@ RUN go build -o ./openfga ./cmd/openfga
...
+ENTRYPOINT ["/app/openfga"]
```
craigpastro, line 13:
```suggestion
ENTRYPOINT ["./openfga", "run"]
```

Dockerfile
craigpastro: I noticed when I tried to run it. 👍

pkg/cmd/service/service.go
craigpastro: I think we will need to update the docker-compose file as well. Expose port 3000 and, if we have this off by default, add an env var.

Dockerfile
```diff
@@ -8,4 +8,6 @@ RUN go build -o ./openfga ./cmd/openfga
 FROM alpine as final
 EXPOSE 8080
 COPY --from=builder /app/openfga /app/openfga
-ENTRYPOINT ["/app/openfga"]
\ No newline at end of file
+COPY --from=builder /app/static /app/static
```
craigpastro, line 11: Can we add `EXPOSE 3000` as well please?

README.md
```diff
@@ -151,6 +164,28 @@ If everything is running correctly, you should get a response with information a
...
+```sh
+OPENFGA_PLAYGROUND_PORT=3001 ./bin/openfga
+```
+
+**Source**
+
+```sh
+OPENFGA_PLAYGROUND_PORT=3001 ./openfga
+```
+
```
craigpastro, line 186: Should we add a docker section with:

```sh
docker run -p 8080:8080 -p 3000:3000 openfga/openfga run
```

Although, I'll mention it again, I think we should simplify our README and the many ways of running and building that we are documenting. I think we should make use of our Makefile and let that be the documentation. What does everyone think?

jon-whit: Personally, I don't love make target bloat such as `make run`, `make run-playground`, `make run-postgres`, `make run-otherA`, etc.. Config should be easy to change and manage, and if it isn't then we should be prioritizing that problem.

I think the Makefile's "North Star" should always be to assist with common, and tedious, developer setup and development tasks. It shouldn't just become a dumping ground for alternate make targets that just set a config value differently before invoking the binary. 

This is obviously my opinion, and I think it's worth talking about more generally, but I think we should strive to discourage make targets that are just aliases for running the server with a particular config value. I think it's ok to ask a developer to change config if they want to run the server with different config.

craigpastro: Actually, I agree with you. What I was thinking more of is that we know have quite a number of ways to run OpenFGA in the readme and, in my opinion, it would be ideal to have (the Docker way of course and then) one easy way such as `make run`, and that would build and execute. If a user needs more details they can take a look at the Makefile. So the Makefile documents the steps to build from source, and we can remove that from the README, where I think it is not necessary.

Once we have an easy to set config (as in your PR) then it makes sense to to trim what we have now in the Makefile. As it is currently, since we don't read from a `.env` file, and finding out what env variables we can set, I think having a bunch of Make targets is not actually that bad a compromise.

pkg/cmd/service/service.go
jon-whit: Done

Dockerfile
jon-whit: Done

README.md
jon-whit: > As it is currently, since we don't read from a .env file, and finding out what env variables we can set, I think having a bunch of Make targets is not actually that bad a compromise.

I agree. I think once we get these config changes we can start moving away from this pattern like you said as well.

> Actually, I agree with you. What I was thinking more of is that we know have quite a number of ways to run OpenFGA in the readme and, in my opinion, it would be ideal to have (the Docker way of course and then) one easy way such as make run, and that would build and execute.

I get what you're saying, and I like the intent behind it a lot. I think the challenges comes from the fact that we have different installation/setup methods, and having a docker target and a `make run` target leaves out the folks that may want to run the pre-compiled binary. I think for the source code installation instructions we could simply focus on the `make run` approach. But we do have

* Docker
* Pre-compiled Binaries
* Building from Source

I think we could probably drop the `docker-compose` instructions and just prioritize the docker instructions, but drop a hint that you can use the docker-compose file included in the repository if you like to use docker-compose 🤷 ?

README.md
```diff
@@ -3,40 +3,53 @@
...
+![Snyk Vulnerabilities for GitHub Repo](https://img.shields.io/snyk/vulnerabilities/github/openfga/openfga?color=orange)
 [![Discord Server](https://img.shields.io/discord/759188666072825867?color=7289da&logo=discord "Discord Server")](https://discord.com/channels/759188666072825867/930524706854031421)
 [![Twitter](https://img.shields.io/twitter/follow/openfga?color=%23179CF0&logo=twitter&style=flat-square "@openfga on Twitter")](https://twitter.com/openfga)
 
 A high-performance and flexible authorization/permission engine built for developers and inspired by [Google Zanzibar](https://research.google/pubs/pub48190/).
 
 OpenFGA is designed to make it easy for developers to model their application permissions and add and integrate fine-grained authorization into their applications.
 
-# Getting Started
+## Table of Contents
```
craigpastro, line 18: I am wondering if we need a ToC. I am not sure how much value it adds, and it seems like another thing that will quickly get out of sync.

Actually, I just found that a ToC is automatically generated. https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/

README.md
```diff
@@ -3,40 +3,53 @@
...
+
+## Getting Started
 
 The following section aims to help you get started quickly. Please look at our official [documentation](https://openfga.dev/) for in-depth information.
 
-## Setup and Installation
+### Setup and Installation
 
-> ℹ️ The following sections setup an OpenFGA server using the default configuration values. These are for rapid development and not for a production environment.
+> ℹ️ The following sections setup an OpenFGA server using the default configuration values. These are for rapid development and not for a production environment. Data written to an OpenFGA instance with default configurations will *not* persist after the service is stopped.
```
craigpastro, line 37: A little more explanation here maybe?
```suggestion
> ℹ️ The following sections setup an OpenFGA server using the default configuration values. These are for rapid development and not for a production environment. Data written to an OpenFGA instance using the default configuration with the memory storage engine will *not* persist after the service is stopped.
```

README.md
jon-whit: Nice 👍 !

jon-whit: Dropped the TOC in favor of the autogenerated one.

jon-whit: Dropped the TOC in favor of the autogenerated one.

jon-whit (approved):

PR #69, opened 2022-06-24
test: use `T.Setenv` to set env vars in tests

Juneezee:
## Description
This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

## References

Reference: https://pkg.go.dev/testing#T.Setenv

## Review Checklist


jon-whit (approved):
Thanks!

PR #70, opened 2022-06-24
chore(ci): only run semgrep, fossa & snyk on branches owned by openfga org

rhamzeh:
These checks will always fail for external contributors, this change only makes them run on branches coming in from within the org, we will have to rely on the weaker webhook integration for outside PRs (these checks will still run on merge to main).

## Review Checklist

### DIFF
--- a/.github/workflows/semgrep.yaml
+++ b/.github/workflows/semgrep.yaml
@@ -1,15 +1,17 @@
 name: Semgrep
 on:
+  pull_request:
   push:
     branches:
       - main
+      - master
 jobs:
   semgrep:
     name: Scan
     runs-on: ubuntu-latest
     container:
       image: returntocorp/semgrep
-    if: (github.actor != 'dependabot[bot]' && github.actor != 'snyk-bot')
+    if: (github.repository_owner == 'openfga' && github.actor != 'dependabot[bot]' && github.actor != 'snyk-bot')
     steps:
     - uses: actions/checkout@v3
     - run: semgrep ci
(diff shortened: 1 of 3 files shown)


.github/workflows/snyk.yaml
```diff
@@ -9,6 +9,7 @@ on:
 jobs:
   security:
     runs-on: ubuntu-latest
+    if: (github.repository_owner == 'openfga' && (github.actor != 'dependabot[bot]' && github.actor != 'snyk-bot'))
```
jon-whit, line 12: Should we run snyk and semgrep even on dependabot PRs? Why did we ignore these actor(s) in the first place?

.github/workflows/semgrep.yaml
```diff
@@ -1,15 +1,17 @@
...
       - main
+      - master
 jobs:
   semgrep:
     name: Scan
     runs-on: ubuntu-latest
     container:
       image: returntocorp/semgrep
-    if: (github.actor != 'dependabot[bot]' && github.actor != 'snyk-bot')
+    if: (github.repository_owner == 'openfga' && (github.actor != 'dependabot[bot]' && github.actor != 'snyk-bot'))
```
craigpastro, line 14: I think the extra parenthesis are not required here. Should we get rid of them?

.github/workflows/snyk.yaml
rhamzeh: At the moment the secrets are not set for dependabot except for FOSSA, we'd need to renable once the secrets are added

craigpastro (approved):
LGTM

PR #75, opened 2022-06-27
fix: make grpc server and http server bind addr configurable

### LINKED ISSUE #74: gRPC server not accessible outside of docker
Currently, the gRPC server is exposed on port 8081, however that is only bound to localhost (and thus only reachable within the docker container itself, and not from the outside).
The docker image itself exposes port 8081, so it seems that it would be the intention that the gRPC server itself is bound to `0.0.0.0:8081` rather than `localhost:8081`.
It might be nice that this is configurable through e.g. an env variable, in case users don't want to expose the gRPC server.

jon-whit:
## Description
These changes allow the developer to run OpenFGA on a configurable `host:port` address, thus allowing the servers bind to a host address other than "localhost" of the running container.

The defaults still remain `:8080` (for the HTTP server) and `:8081` (for the grpc server), so it behaves identically under default conditions, but if you want to bind to a different host and port address you now can.

## References
Closes #74 

## Review Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,9 +6,6 @@ COPY . .
 RUN go build -o ./openfga ./cmd/openfga
 
 FROM alpine as final
-EXPOSE 8081
-EXPOSE 8080
-EXPOSE 3000
 COPY --from=builder /app/openfga /app/openfga
 COPY --from=builder /app/static /app/static
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ COPY . .
 RUN go build -o ./openfga ./cmd/openfga
 
 FROM alpine as final
+EXPOSE 8081
+EXPOSE 8080
+EXPOSE 3000
 COPY --from=builder /app/openfga /app/openfga
 COPY --from=builder /app/static /app/static
 WORKDIR /app
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -80,7 +80,7 @@ func run(_ *cobra.Command, _ []string) {
 					err = tmpl.Execute(w, struct {
 						HTTPServerURL string
 					}{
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -6,9 +6,6 @@ COPY . .
 RUN go build -o ./openfga ./cmd/openfga
 
 FROM alpine as final
-EXPOSE 8081
```
jon-whit, line 9: These aren't needed. Without them `docker run -p 8080:8080 openfga/openfga run` runs just fine.

Since these ports are configurable anyways, it doesn't make sense to statically expose them.

pkg/cmd/run.go
```diff
@@ -80,7 +80,7 @@ func run(_ *cobra.Command, _ []string) {
 					err = tmpl.Execute(w, struct {
 						HTTPServerURL string
 					}{
-						HTTPServerURL: fmt.Sprintf("localhost:%d", config.HTTPPort),
+						HTTPServerURL: config.HTTPAddr,
```
jon-whit, line 83: todo: need to fix this @jon-whit . The default `:8080` doesn't work here obviously, so we'll need to account for that.

Dockerfile
craigpastro: [EXPOSE doesn't actually publish the port](https://docs.docker.com/engine/reference/builder/#expose), but I think it is useful for documentation purposes if by default these are the ports OpenFGA is running on. In that sense, I would prefer to keep them. WDYT?

pkg/cmd/run.go
jon-whit: Alright, I resolved this by waiting for a tcp connection to be established to the HTTP server and then dynamically grabbing the remote host address from the connection. So makes it work regardless of what host:port address you wire the HTTP server up with 👍

elbuo8 (approved):

PR #83, opened 2022-06-28
test: add preliminary functional test coverage

jon-whit:
## Description
This PR starts laying the foundation for functional test coverage of the OpenFGA grpc and HTTP servers. The changes herein work by launching an `openfga/openfga` container (built ahead of time) and running tests against the grpc and HTTP servers by interacting with the APIs via a client connected to the running container.

In the CI/CD pipeline we build the openfga image based on the changes included in the PR and we run the functional test suite against it.

## Review Checklist


cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,402 @@
...
+	require.NoError(t, err)
+
+	return &serverHandle{
+		grpcPort: grpcPort,
+		httpPort: httpPort,
+		cleanup:  stopContainer,
+	}, nil
+}
+
+func randomID(t *testing.T) string {
```
jon-whit, line 154: This probably belongs in our `testutils` package instead of here.

- [ ] todo(jon-whit): move this to the `testutils` package.

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,402 @@
...
+	backoffPolicy.MaxElapsedTime = 30 * time.Second
+
+	err = backoff.Retry(
+		func() error {
+			var err error
+
+			timeoutCtx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+			defer cancel()
+
+			_ = timeoutCtx
```
jon-whit, line 137: - [ ] todo(jon-whit): wire up a health check here.

Makefile
```diff
@@ -47,6 +47,17 @@ unit-test: go-generate ## Run unit tests
 			-count=1 \
 			./...
 
+
+.PHONY: functional-test
+functional-test:
```
jon-whit, line 52: Not sure the best way to approach this. Ideally you could just run `make functional-test` and it would 1) build the docker image against the latest changes and 2) run the functional tests (by passing the `-tags=functional` flag).

If we naively do

```
.PHONY: build-image
build-image: go-generate
	docker build -t openfga/openfga:latest .

.PHONY: functional-test
functional-test: build-image
	go test -v ...
```
then every time we invoke it we will have to go through the docker build process, which doesn't cache the `go build ...` step. This is pretty time costly, and in a CI/CD environment we pre-build the image anyways.

I'm wondering if it's ok to just assume the image is pre-built and if you want to run functional tests locally you'll have to make sure you do so. What do you think 🤷 ?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,404 @@
+//go:build functional
+// +build functional
```
jon-whit, line 2: This is a Go build tag, which means that this file will only get included in the test build if you pass the `-tags=functional` flag to `go test`. The idea I was going for is that, by default, functional tests don't run. You have to specify the build tag to get the functional tests to run.

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,919 @@
...
+
+	require := require.New(t)
+
+	// tester can be shared across tests that aren't impacted
+	// by shared state
+	tester, err := newOpenFGATester(t)
+	require.NoError(err)
+	defer tester.Cleanup()
+
+	t.Run("TestCreateStore", func(t *testing.T) { GRPCCreateStoreTest(t, tester) })
```
jon-whit, line 200: These tests can kind of be broken down into 2 categories: tests for input validation and tests for standard workflows (functional workflows). The former is a great candidate for test tables, while the later becomes a little more challenging for table driven tests because of the preconditions that have to be met in many cases.

What do you think about creating two separate tests for these different categories? Any preference on this structure?

Makefile
craigpastro: Both of these choices are not ideal. Do you think it is not enough to run it against `main`?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,919 @@
...
+	Cleanup() func()
+}
+
+type serverHandle struct {
+	grpcPort string
+	httpPort string
+	cleanup  func()
+}
+
+func (s *serverHandle) GetGRPCPort() string {
```
craigpastro, line 42: Do we need this function? Or can we just make `grpcPort` in `serverHandle` public?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,919 @@
...
+			"3000/tcp": []nat.PortBinding{},
+		},
+	}
+
+	ulid, err := id.NewString()
+	require.NoError(t, err)
+
+	name := fmt.Sprintf("openfga-%s", ulid)
+
+	cont, err := dockerClient.ContainerCreate(context.Background(), &containerCfg, &hostCfg, nil, nil, name)
```
craigpastro, line 92: Should we create a single ctx instead of using `context.Background()` everywhere?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,919 @@
...
+	if !ok || len(m) == 0 {
+		t.Fatalf("failed to get grpc host port mapping from openfga container")
+	}
+	grpcPort := m[0].HostPort
+
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+
+	creds := insecure.NewCredentials()
+	// if transportCredentials != nil {
```
craigpastro, line 139: Shall we remove this?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,919 @@
...
+	require.NoError(t, err)
+	defer conn.Close()
+
+	client := healthv1pb.NewHealthClient(conn)
+
+	backoffPolicy := backoff.NewExponentialBackOff()
+	backoffPolicy.MaxElapsedTime = 30 * time.Second
+
+	err = backoff.Retry(func() error {
+		timeoutCtx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
```
craigpastro, line 163: Do we need this? Or should we just set `MaxElapsedTime` to 2 seconds above?

cmd/openfga/openfga_test.go
jon-whit: Sure!

cmd/openfga/openfga_test.go
jon-whit: I don't want the `Check` to hang until `MaxElapsedTime` has been reached. I want to limit how long the Check can take and then retry the check again in a subsequent retry. If we didn't set the timeout ctx then the single Check could hang until `MaxElapsedTime` elapsed.

cmd/openfga/openfga_test.go
craigpastro: Sounds good to me. Thanks!

Makefile
jon-whit: So I originally added the `build-image` Makefile target and had it as a pre-req for this target, but I really didn't like that because every time you run `make functional-test` it does a `docker build ...` which has all of the docker build output, and it takes longer than it probably needs.

The problem is that the `build-image` target shouldn't probably be run if there aren't any sources that have changed that would affect the built image. So to improve that experience we'd want to only have the `build-image` target run if any of the Go source files have changed or any of the artifacts included in the container.

We could do that, it just requires a good chunk of more work to orchestrate that through the Makefile (but it is what make was made for, haha).

WDYT, should we bite the bullet and do ^^?

jon-whit:
Tests will fail until #105 is fixed. I've added tests to assert the correct behavior here. I'd like #105 to be merged before merging this.

Makefile
miparnisari: > The problem is that the `build-image` target shouldn't probably be run if there aren't any sources that have changed that would affect the built image.

For local development, I will always have to build the image, no? Because presumably I changed something and I want to see if the tests still pass.

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,968 @@
...
+	}
+
+	err = dockerClient.ContainerStart(ctx, cont.ID, types.ContainerStartOptions{})
+	require.NoError(t, err)
+
+	t.Cleanup(stopContainer)
+
+	// spin up a goroutine to survive any test panics or terminations to expire/stop the running container
+	go func() {
+		time.Sleep(2 * time.Minute)
```
miparnisari, line 115: is this the time we expect the tests to take? if so, i think it's a little low... 10 minutes i've seen them take

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,968 @@
...
+
+	require := require.New(t)
+
+	// tester can be shared across tests that aren't impacted
+	// by shared state
+	tester, err := newOpenFGATester(t)
+	require.NoError(err)
+	defer tester.Cleanup()
+
+	t.Run("TestCreateStore", func(t *testing.T) { GRPCCreateStoreTest(t, tester) })
```
miparnisari, line 200: i'm confused as to why we have `GRPCCreateStoreTest` and `HTTPCreateStoreTest` - are we going to duplicate the code? 🤯

jon-whit: They are both part of the API surface, and given that we use middleware that differs between the two I think it's important we test the functional API behavior of both :).

So yeah, we'll test both.

Makefile
jon-whit: @miparnisari 

> For local development, I will always have to build the image, no? Because presumably I changed something and I want to see if the tests still pass.

Most of the time, yes. But not always. If none of the Go files or artifacts that go inside the container have changed, then you don't need to rebuild the image to run functional tests.

> Can you remind me why we did it like this? I forgot :/

Because this gives us the highest fidelity functional test coverage. This approach allows us to test the entrypoint(s) of the container as well as the server code embedded in it. If we only tested the server implementation without spinning up the container, then we could easily foobar the Dockerfile and introduce something unintended. This approach allows us to test the whole integration end to end with very little overhead (you just have to build the container).

cmd/openfga/openfga_test.go
jon-whit: If a test hasn't completed in 2min then we have a problem. Tests should have stricter timeouts by using `context.WithTimeout(...)`.

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,967 @@
...
+	require.Equal(t, healthv1pb.HealthCheckResponse_SERVING, resp.Status)
+
+	_, err = openfgaClient.CreateStore(context.Background(), &openfgapb.CreateStoreRequest{
+		Name: "openfga-demo",
+	})
+	require.Error(t, err)
+
+	s, ok := status.FromError(err)
+	require.True(t, ok)
+	require.Equal(t, codes.Code(openfgapb.AuthErrorCode_bearer_token_missing), s.Code())
```
jon-whit, line 240: This makes no sense to me as a client and as a maintainer of this project, haha 🤣 

What I should be asserting here is the following:

```
require.Equal(t, codes.Unauthenticated, s.Code())
```
We should adhere to the canonical status codes of the protocol we are using. It makes no sense for us to be implementing our own status code protocol ontop of a preexisting one.

server/authn/oidc/oidc.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/MicahParks/keyfunc"
 	"github.com/go-errors/errors"
 	"github.com/golang-jwt/jwt/v4"
-	grpcAuth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
+	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
```
elbuo8, line 14: Why this change?

cmd/openfga/openfga_test.go
```diff
@@ -0,0 +1,967 @@
...
+
+func HTTPCheckTest(t *testing.T) {
+
+}
+
+func HTTPExpandTest(t *testing.T) {
+
+}
+
+func tuple(object, relation, user string) *openfgapb.TupleKey {
```
elbuo8, line 961: I believe a util was recently introduced to this.

.github/workflows/pull_request.yaml
```diff
@@ -54,6 +54,50 @@ jobs:
...
+        - uses: actions/cache@v3
+          with:
+            path: ${{ steps.go-cache-paths.outputs.go-mod }}
+            key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
+
+        # build the openfga/openfga image for the functional tests
+        - uses: docker/build-push-action@v2
+          with:
+            file: Dockerfile
+            push: false
```
elbuo8, line 88: Could you please document this so that it doesn't get randomly flipped?

.github/workflows/pull_request.yaml
```diff
@@ -54,6 +54,50 @@ jobs:
...
+            tags: "openfga/openfga:latest"
```
elbuo8, line 89: Perhaps are more meaningful tag?

cmd/openfga/openfga_test.go
jon-whit: Oops. That's residual. It's in the `tuple` package.

.github/workflows/pull_request.yaml
jon-whit: I switched it to `openfga/openfga:functionaltest`.

.github/workflows/pull_request.yaml
jon-whit: Added a comment to it

server/authn/oidc/oidc.go
elbuo8: For a subsequent PR can we add a linting rule to prevent this from happening and normalize?

.github/workflows/pull_request.yaml
```diff
@@ -54,6 +54,48 @@ jobs:
...
+      runs-on: ubuntu-latest
+      timeout-minutes: 15
+      steps:
+        - uses: actions/checkout@v3
+
+        - uses: actions/setup-go@v3
+          with:
+            go-version-file: './go.mod'
+
+        - id: go-cache-paths
```
elbuo8, line 67: This isn't needed with the other change, remove in a subsequent PR to get this one in @jon-whit

elbuo8 (approved):

PR #86, opened 2022-06-29
feat: add grpc health checks and wire it up with `/healthz`

### LINKED ISSUE #85: support grpc health spec and register HTTP gateway `/healthz` with it
The grpc protocol has a [Health Check Specification](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) that we should be following. We should register the grpc health check endpoints as part of the OpenFGA grpc server, and with the HTTP gateway we can use [runtime.WithHealthzEndpoint](https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/v2@v2.10.3/runtime#WithHealthzEndpoint) to register the `/healthz` endpoint on the HTTP gateway server in a way that delegates to the grpc health check.

This approach will unify our health checks in OpenFGA.

The [grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe) utility can be used to probe the grpc health check endpoint or (if the HTTP gateway is enabled) you can query the `/healthz` endpoint as well.

jon-whit:
## Description
The changes in this PR add grpc health checks to the OpenFGA server in accordance with the [GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md). The grpc health checks are also used to serve the response returned by the `/healthz` endpoint registered on the HTTP gateway server.

The health check endpoints do not require authentication, because requiring authentication would make them challenging to integrate with various health check integrations (e.g. kubernetes health checks, AWS ELB health checks, etc..). For health check probes against the grpc server you can integrate the [`grpc-health-probe`](https://github.com/grpc-ecosystem/grpc-health-probe) tool into your deployment workflow(s). However, if you wish to check server health using the `/healthz` endpoint it is still available to you.

You can test this out two ways:
**grpc**
```
➜ make run
➜ grpcurl -plaintext -d '{"service":"openfga.v1.OpenFGAService"}' localhost:8081 grpc.health.v1.Health/Check
{
  "status": "SERVING"
}
```

**http**
```
➜ make run
➜ curl http://localhost:8080/healthz
{"status":"SERVING"}
```
## References
Closes #85 

[GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md)
[grpc-gateway Health Checks](https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/health_check/#health-check)
[grpc-ecosystem/grpc-health-probe](https://github.com/grpc-ecosystem/grpc-health-probe)

## Review Checklist


pkg/cmd/service/service.go
```diff
@@ -7,14 +7,15 @@ import (
...
+	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
 	"github.com/kelseyhightower/envconfig"
 	"github.com/openfga/openfga/pkg/encoder"
 	"github.com/openfga/openfga/pkg/logger"
 	"github.com/openfga/openfga/pkg/telemetry"
 	"github.com/openfga/openfga/server"
-	"github.com/openfga/openfga/server/authentication"
-	"github.com/openfga/openfga/server/authentication/oidc"
-	"github.com/openfga/openfga/server/authentication/presharedkey"
+	"github.com/openfga/openfga/server/authn"
```
jon-whit, line 16: `authn` is ubiquitous enough to use it in place of the long-hand "authentication". We'll introduce the `authz` package when/if we introduce FGA inside of FGA via an `authz.Authorizer` interface.

pkg/cmd/service/service.go
```diff
@@ -7,14 +7,15 @@ import (
 	"log"
 	"time"
 
+	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
```
jon-whit, line 10: Wired up the `grpc_auth` middleware package in favor of a vanilla grpc interceptor because this allows projects already established in the grpc middleware ecosystem to plug and play their existing `grpc_auth.AuthFunc` with us.

server/server.go
```diff
@@ -426,6 +445,8 @@ func (s *Server) Run(ctx context.Context) error {
 	// nosemgrep: grpc-server-insecure-connection
 	grpcServer := grpc.NewServer(opts...)
 	openfgapb.RegisterOpenFGAServiceServer(grpcServer, s)
+	healthv1pb.RegisterHealthServer(grpcServer, s.healthManager.GetHealthServer())
+	reflection.Register(grpcServer)
```
jon-whit, line 449: I added GRPC Server Reflection here to make it easier to discover the grpc endpoints as a grpc client.

See https://github.com/grpc/grpc/blob/master/doc/server-reflection.md. This is a really nice feature/integration to have for grpc and it made my testing of the functionality contained in this PR easier. That's why I added it.

server/server.go
```diff
@@ -540,3 +574,74 @@ func (s *Server) resolveAuthorizationModelID(ctx context.Context, store, modelID
...
+func (s *serverHealthChecker) Check(ctx context.Context) func() error {
+	return func() error {
+
+		backoffPolicy := backoff.NewExponentialBackOff()
+		backoffPolicy.MaxElapsedTime = 1 * time.Minute
+		ticker := backoff.NewTicker(backoffPolicy)
+		defer ticker.Stop()
+
+		// continuously monitor health status on a ticker interval
+		for {
```
jon-whit, line 622: Implement an exponential backoff/retry to check the server readiness. If we exhaust our retry policy then we return an error and fatal error at the caller level. If we need to we can change this behavior, but it felt right to me at this time.

pkg/cmd/service/service.go
elbuo8: This is good 👍🏽 

But please try to keep PRs within scope, this could have been a separate one.

pkg/cmd/service/service.go
```diff
@@ -45,13 +46,13 @@ type Config struct {
 	// RequestTimeout is a limit on the time a request may take. If the value is 0, then there is no timeout.
 	RequestTimeout time.Duration `default:"0s" split_words:"true"`
 
-	GRPCTLSEnabled  bool   `default:"false" envconfig:"GRPC_TLS_ENABLED"`
-	GRPCTLSCertPath string `envconfig:"GRPC_TLS_CERT_PATH"`
-	GRPCTLSKeyPath  string `envconfig:"GRPC_TLS_KEY_PATH"`
+	GRPCTLSEnabled  bool   `default:"false" envconfig:"OPENFGA_GRPC_TLS_ENABLED"`
```
elbuo8, line 49: Does this play well with the `prefix` in which it's read?

Also why not continue to use `split_words` instead of mixing 🤔

jon-whit: I'm not sure at what point we started using the `envconfig` struct tag in favor of the `split_words`, but I was talking with @craigpastro about this yesterday. I guess the issue with `split_words` is it is sometimes ambiguous where to split on, and `envconfig.Process` doesn't always split it the way we want it, so we've chosen to be explicit for these fields. The splitting typically works on change in upper/lower character case, but since we have the `stylecheck` lint checks our field names end up have less change in character case. 

Another thing I noticed is that if you use the `envconfig` struct tag to specify what env variable to use for the field, then `default` value struct tags get applied to the env variable *before* the prefix is applied. Default values don't work well if you're using a prefix + explicit env variable name. Adding the expected prefix in front of the explicit env variable name fixes the issue.

pkg/cmd/service/service.go
jon-whit: Sounds good.

I felt this approach within the scope of this PR made the most sense because it made the work a little easier for me, and so although it was a tad tangential, it made it easier for me to integrate the authless health checks within the scope of this PR.

elbuo8 (approved):

PR #92, opened 2022-07-07
feat: add basic CLI flags to the `run` command

jon-whit:
## Description
This adds some CLI flags for the `run` command to wire up basic options that you can provide at the command-line. The changes herein wire up the cobra command flags with the config being managed by viper. 

## References
Part of #91 

* Design inspired by the spf13/cobra [User Guide](https://github.com/spf13/cobra/blob/master/user_guide.md)

## Review Checklist
- [ ] I have added documentation for new/changed functionality in this PR or in [openfga.dev](https://github.com/openfga/openfga.dev)
- [X] The correct base branch is being used, if not `main`

Issue #91, opened 2022-07-07
wire up CLI flags in addition to server configuration in `config.yaml`


jon-whit:
The functional tests in #83 launch a container for OpenFGA to test against, and I want to be able to launch the container with specific config but without having to construct a config.yaml in the automation. I'd like to be able to run openfga with, for example,

```
openfga run --authn-method preshared --authn-preshared-keys key1,key2
```
and
```
# enable HTTP TLS only
openfga run --http-tls-enabled --http-tls-cert /some/path --http-tls-key /some/path

# enable grpc TLS only
openfga run --grpc-tls-enabled --grpc-tls-cert /some/path --grpc-tls-key /some/path

# enable both
openfga run \
  --grpc-tls-enabled --grpc-tls-cert /some/path --grpc-tls-key /some/path \
  --http-tls-enabled --http-tls-cert /some/path --http-tls-key /some/path
```

Since we have the CLI it'd be nice to make these flags as part of our CLI in addition to be configurable by `config.yaml`. So config.yaml + CLI flags will work together. We'll have to choose which one should take precedence, but my hunch is that we should just follow viper's default.

> Viper uses the following precedence order. Each item takes precedence over the item below it:
> 
> explicit call to Set
> flag
> env
> config
> key/value store
> default

So flags should take precedence over the config.yaml values.

PR #93, opened 2022-07-08
fix: viper.Unmarshal config even if `config.yaml` is not present

jon-whit:
## Description
I found a small bug that I didn't catch before merging #92 . These changes fix an issue where if you don't have a `config.yaml` present in any one of ["/etc/openfga", "$HOME/.openfga, "."], then the values even from the CLI flags won't get loaded. The problem was that we were returning default config values before calling `viper.Unmarshal`, and so any flags that were binded to viper did not have any influence on the config.

## References
#91 

## Review Checklist


craigpastro (approved):
LGTM

PR #100, opened 2022-07-11
feat: wire up support for env variables

jon-whit:
## Description
The adds support for env variables to the list of optional ways to configure the OpenFGA server. Configuration is loaded according to the default behavior of the spf13/viper package. That is,

> Viper uses the following precedence order. Each item takes precedence over the item below it:
> 
> explicit call to Set
> flag
> env
> config
> key/value store
> default

Each config property is mapped to an env variable by translating the config to all uppercase and replacing any usage of `.` with `_` and attaching the `OPENFGA_` prefix to it. For example, `http.enabled` becomes `OPENFGA_HTTP_ENABLED` and `authn.preshared.keys` becomes `OPENFGA_AUTHN_PRESHARED_KEYS` etc..

## References
https://github.com/spf13/viper#why-viper

## Review Checklist

### DIFF
--- a/pkg/cmd/service/service_test.go
+++ b/pkg/cmd/service/service_test.go
@@ -13,6 +13,8 @@ import (
 	"math/big"
 	"net/http"
 	"os"
+	"path"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -22,6 +24,7 @@ import (
 	"github.com/openfga/openfga/pkg/retryablehttp"
 	"github.com/openfga/openfga/server/authn/mocks"
 	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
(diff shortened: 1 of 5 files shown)


pkg/cmd/service/service_test.go
```diff
@@ -713,3 +725,81 @@ func TestHTTPServerEnabled(t *testing.T) {
 
 	cancel()
 }
+
+func TestDefaultConfig(t *testing.T) {
```
jon-whit, line 729: I added this with the hope that it will make sure what's in our `.config-schema.json` file aligns with the actual server defaults.

craigpastro (approved):
Nice!

miparnisari (approved):

pkg/cmd/service/service_test.go
rhamzeh: Would it make sense to do this in prod instead of redefining the DefaultConfig?

So DefaultConfig gets read from the schema?

Or on build the schema gets written by the default config? Just in case we forget to add new config options to the test

rhamzeh (approved):

pkg/cmd/service/service_test.go
jon-whit: I debated this in https://github.com/openfga/openfga/pull/63#discussion_r903099282

I ended up going with this approach because of the overhead and difficulty of generating a default Go struct from a json-schema, and I felt that the Go struct should be the source of truth and not the other way around (since the Go struct is the definitive entity defining the actual defaults). This approach also avoids having to embed the `.config-schema.json` file in the built container (minor albeit extra dependencies required inside the container). 

Ultimately I could go either way. I'm not opposed to going this approach, and I've already researched what it would take. It just seemed like a lot of overhead for little value 🤷 .

PR #104, opened 2022-07-13
fix: add Playground artifact to the goreleaser build artifacts

jon-whit:
## Description
This fixes the issue with the Playground assets not being included with our build artifacts generated by goreleaser.

## Review Checklist

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -8,7 +8,7 @@ before:
 builds:
   - 
     main: ./cmd/openfga
-    binary: bin/openfga
+    binary: openfga
     env:
       - CGO_ENABLED=0
     goos:
@@ -51,6 +51,8 @@ archives:
       windows: Windows
       386: i386
       amd64: x86_64
+    files:
+      - static/playground
 
 checksum:
   name_template: 'checksums.txt'
--- a/Makefile
+++ b/Makefile
@@ -19,20 +19,20 @@ lint:  ## Lint Go source files
 
 .PHONY: clean
 clean: ## Clean files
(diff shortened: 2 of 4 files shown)


.goreleaser.yaml
```diff
@@ -8,7 +8,7 @@ before:
 builds:
   - 
     main: ./cmd/openfga
-    binary: bin/openfga
```
jon-whit, line 11: Dumping the binary into the `bin/` directory just leads us down a rabbit whole, because the Playground assets are intended to be relative to the `openfga` binary. So to alleviate the scenario where a developer extracts the pre-built binaries and does `cd bin && ./openfga run` (which will error out because the Playground assets are no longer relative to the binary), this helps avoid that issue.

jon-whit: Actually, let me try and rework this altogether so that it's not subject to relative path issues.

craigpastro: Do we need to update the Makefile as well? `make build` drops the binary in the bin folder.

jon-whit: Ah, yes! Good catch.

jon-whit: Ok @craigpastro I've updated it by dropping the `/bin` prefix where applicable.

craigpastro (approved):
LGTM. Do we need doc updates?

jon-whit:
> LGTM. Do we need doc updates?

I don't see any documentation on openfga.dev that references running OpenFGA via binary, so I think the changes to the README herein are sufficient.

PR #106, opened 2022-07-18
chore: bind remaining CLI flags

### LINKED ISSUE #91: wire up CLI flags in addition to server configuration in `config.yaml`

jon-whit:
## Description
This binds the remaining flags for the OpenFGA server.

## References
Closes #91 

## Review Checklist

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -241,6 +241,13 @@ func bindFlags(cmd *cobra.Command) {
 
 	cmd.MarkFlagsRequiredTogether("http-tls-enabled", "http-tls-cert", "http-tls-key")
 
+	cmd.Flags().Duration("http-upstream-timeout", defaultConfig.HTTP.UpstreamRequestTimeout, "the timeout duration for proxying HTTP requests upstream to the grpc endpoint")
+	cmdutil.MustBindPFlag("http.upstreamRequestTimeout", cmd.Flags().Lookup("http-upstream-request-timeout"))
+
+	cmd.Flags().StringSlice("http-cors-allowed-origins", defaultConfig.HTTP.CORSAllowedOrigins, "specifies the CORS allowed origins")
+
+	cmd.Flags().StringSlice("http-cors-allowed-headers", defaultConfig.HTTP.CORSAllowedHeaders, "specifies the CORS allowed headers")
+
 	cmd.Flags().String("authn-method", defaultConfig.Authn.Method, "the authentication method to use")
 	cmdutil.MustBindPFlag("authn.method", cmd.Flags().Lookup("authn-method"))
 	cmd.Flags().StringSlice("authn-preshared-keys", defaultConfig.Authn.Keys, "one or more preshared keys to use for authentication")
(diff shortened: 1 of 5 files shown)


pkg/cmd/run.go
```diff
@@ -262,4 +271,16 @@ func bindFlags(cmd *cobra.Command) {
 
 	cmd.Flags().String("log-format", defaultConfig.Log.Format, "the log format to output logs in")
 	cmdutil.MustBindPFlag("log.format", cmd.Flags().Lookup("log-format"))
+
+	cmd.Flags().Int("max-tuples-per-write", defaultConfig.OpenFGA.MaxTuplesPerWrite, "the maximum allowed number of tuples per Write transaction")
+	cmdutil.MustBindPFlag("maxTuplesPerWrite", cmd.Flags().Lookup("max-tuples-per-write"))
+
+	cmd.Flags().Int("max-types-per-authz-model", defaultConfig.OpenFGA.MaxTypesPerAuthorizationModel, "the maximum allowed number of type definitions per authorization model")
```
adriantam, line 278: Can we use authorization-model instead of authz-model?  Prefer to have 1-1 mapping with the .config-schema.json?  If we want to have authz-model, then update config-schema.json accordingly.

adriantam (changes requested):

pkg/cmd/run.go
craigpastro: In my mind, I always just call it a `model` now because this is an authorization app. My vote would be (if we are voting 😉) to rename them all to `model` 🤪 .

pkg/cmd/run.go
```diff
@@ -241,6 +241,13 @@ func bindFlags(cmd *cobra.Command) {
 
 	cmd.MarkFlagsRequiredTogether("http-tls-enabled", "http-tls-cert", "http-tls-key")
 
+	cmd.Flags().Duration("http-upstream-timeout", defaultConfig.HTTP.UpstreamTimeout, "the timeout duration for proxying HTTP requests upstream to the grpc endpoint")
+	cmdutil.MustBindPFlag("http.upstreamTimeout", cmd.Flags().Lookup("http-upstream-timeout"))
+
+	cmd.Flags().StringSlice("http-cors-allowed-origins", defaultConfig.HTTP.CORSAllowedOrigins, "specifies the CORS allowed origins")
+
```
craigpastro, line 248: Just wondering: these don't need a `cmdutil.MustBindPFlag`?

craigpastro (approved):
LGTM

adriantam (approved):

PR #113, opened 2022-07-25
fix: goroutine leak

craigpastro:
## Review Checklist

### DIFF
--- a/server/commands/check_utils.go
+++ b/server/commands/check_utils.go
@@ -426,7 +426,6 @@ func (t *tupleKeyIterator) next() (*openfgapb.TupleKey, error) {
 }
 
 func (t *tupleKeyIterator) stop() {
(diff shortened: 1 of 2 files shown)


server/commands/check_utils.go
```diff
@@ -426,7 +426,6 @@ func (t *tupleKeyIterator) next() (*openfgapb.TupleKey, error) {
 }
 
 func (t *tupleKeyIterator) stop() {
-	// Close shouldn't error if called multiple times
```
elbuo8, line 429: ?

craigpastro: The function name in the comment was wrong, and it is repeated on the underlying `Stop` function. I don't think it is needed here. I could fix it and put it back though. WDYT?

elbuo8 (approved):

PR #114, opened 2022-07-25
Feature: Reject direct if only indirect relationship allowed

### LINKED ISSUE #73: feat: Reject tuple input with direct relationship if not allowed instead of writing
Currently, a `WriteCommand` (ref: `server/commands/write.go`) with a tupleset that is not valid because of attempting to write a `DirectUserset` (ref: `openfga/v1/authzmodel.proto` in [openfga/api](https://github.com/openfga/api)) is not rejected as invalid.  This tupleset does not cause an invalid authorization check (which is great), but it does create two potential pain points with users:

1. Services that write tuples to an fga store that violate this condition will not receive an error and assume that they provided a semantically valid tuple (they are already syntactically validated with the provided regex patterns); and,
2. fga admins (global or store) may get complaints of authorization checks failing that a service believes should be valid and, upon investigating see the invalid direct relationship, assume there is a deeper or more complex issue than simply violation of the indirect only relationship

Looking through `server/commands/write.go`, we do not currently pull in the `AuthorizationModel` (like in `server/commands/read_authzmodel.go`; though we do have the `StoreId` and `AuthorizationModelId` in the `req` input for `Execute` and should be able to fetch it), but if we did we would have access to the `TypeDefinitions` and should be able to then check if the `TypeDefinition.relations` map contains a `DirectUserset` or not

Screen Shot 2022-06-26 at 2 45 10 PM
.

dblclik:
## Description
[Issue 73](https://github.com/openfga/openfga/issues/73) deals with the current permitted action of writing tuples directly to an indirect-only relationship.  While this does not result in incorrect `/check` authorizations, it does provide a confusing user experience and gives the impression that the authorization model is not being used when writing new tuples.

Documentation has not been updated as this behavior is implied.  If we want to add some mention of this behavior in the docs, let me know and I can add a section for this callout.

## References
Resolves #73 

## Review Checklist
- [x] I have added tests for the new functionality I have added


server/commands/write.go
```diff
@@ -104,3 +139,65 @@ func handleError(err error) error {
...
 	return serverErrors.HandleError("", err)
 }
+
+func isDirectIntersection(nodes *openfgapb.Userset_Intersection, tk *openfgapb.TupleKey) error {
+	isDirect := false
+contolLoop:
+	for _, userset := range nodes.Intersection.Child {
+		switch usType := userset.Userset.(type) {
+		case *openfgapb.Userset_This:
+			fmt.Println(usType.This)
```
craigpastro, line 149: Delete?

server/commands/write.go
```diff
@@ -104,3 +139,65 @@ func handleError(err error) error {
...
+	return nil
+}
+
+func isDirectUnion(nodes *openfgapb.Userset_Union, tk *openfgapb.TupleKey) error {
+	isDirect := false
+contolLoop:
+	for _, userset := range nodes.Union.Child {
+		switch usType := userset.Userset.(type) {
+		case *openfgapb.Userset_This:
+			fmt.Println(usType.This)
```
craigpastro, line 170: Delete?

server/commands/write.go
```diff
@@ -104,3 +139,65 @@ func handleError(err error) error {
...
+}
+
+func isDirectDifference(node *openfgapb.Userset_Difference, tk *openfgapb.TupleKey) error {
+	isDirect := false
+	sets := []*openfgapb.Userset{node.Difference.GetBase(), node.Difference.GetSubtract()}
+contolLoop:
+	for _, userset := range sets {
+		switch usType := userset.Userset.(type) {
+		case *openfgapb.Userset_This:
+			fmt.Println(usType.This)
```
craigpastro, line 191: Delete?

craigpastro (commented):
Thank you so much for your contribution! I had a quick first pass and left some comments. Please let me know what you think.

server/commands/write.go
```diff
@@ -60,9 +61,43 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+			continue
+		case *openfgapb.Userset_ComputedUserset:
+			// if Userset.type is a ComputedUserset then we know it can't be direct
+			err := &tupleUtils.IndirectWriteError{Reason: "Attempting to write directly to an indirect only relationship", TupleKey: tk}
+			if err != nil {
+				return serverErrors.HandleTupleValidateError(err)
+			}
+			continue
+		default:
+			continue // To prevent breaking change, if unsure then continue unexpcted
```
craigpastro, line 99:
```suggestion
			continue // To prevent breaking change, if unsure then continue
```

server/commands/write.go
```diff
@@ -60,9 +61,43 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+			continue
+		case *openfgapb.Userset_Difference:
+			err := isDirectDifference(usType, tk)
+			if err != nil {
+				return serverErrors.HandleTupleValidateError(err)
+			}
+			continue
+		case *openfgapb.Userset_ComputedUserset:
+			// if Userset.type is a ComputedUserset then we know it can't be direct
+			err := &tupleUtils.IndirectWriteError{Reason: "Attempting to write directly to an indirect only relationship", TupleKey: tk}
```
craigpastro, line 93: I think what the linter is saying here is correct. You can just return the error immediately in this case.

server/commands/write.go
```diff
@@ -60,9 +61,43 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	}
 
 	for _, tk := range writes {
-		if _, err := tupleUtils.ValidateTuple(ctx, c.datastore, store, modelID, tk, dbCallsCounter); err != nil {
+		tupleUserset, err := tupleUtils.ValidateTuple(ctx, c.datastore, store, modelID, tk, dbCallsCounter)
+		if err != nil {
 			return serverErrors.HandleTupleValidateError(err)
 		}
+		// TODO: delete the debugging statement
+		c.logger.Info(fmt.Sprintf("%+v\n", tupleUserset))
```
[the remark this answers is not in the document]
dblclik: Thanks, @craigpastro - I forgot to do a final `TODO` search!

server/commands/write.go
```diff
@@ -104,3 +139,65 @@ func handleError(err error) error {
 
 	return serverErrors.HandleError("", err)
 }
+
+func isDirectIntersection(nodes *openfgapb.Userset_Intersection, tk *openfgapb.TupleKey) error {
+	isDirect := false
+contolLoop:
```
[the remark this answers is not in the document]
dblclik: Since Go [break statements](https://go.dev/ref/spec#Break_statements) act on `for`, `switch`, and `loop`, we need a named loop for the `for` loop here so that we can early exit if we find a direct relationship in any of the children

server/commands/write.go
```diff
@@ -60,9 +61,43 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+		if err != nil {
 			return serverErrors.HandleTupleValidateError(err)
 		}
+		// TODO: delete the debugging statement
+		c.logger.Info(fmt.Sprintf("%+v\n", tupleUserset))
+		switch usType := tupleUserset.Userset.(type) {
+		case *openfgapb.Userset_This:
+			continue // no need to check on Direct Relationship
+		case *openfgapb.Userset_Intersection:
+			err := isDirectIntersection(usType, tk)
```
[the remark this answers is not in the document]
dblclik: I agree -- originally I had it doing that, but for some reason switched my approach (probably had a thought, that didn't work out, then forgot to return to original plan).  I've updated the code to use a `bool` return here 👍

server/commands/write.go
dblclik: Good catch -- I tidied this up in my commit!

server/commands/write.go
dblclik: Yep, totally agree -- the `case... return {}` signature now matches throughout this new section

server/commands/write.go
```diff
@@ -60,9 +60,37 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+		if err != nil {
 			return serverErrors.HandleTupleValidateError(err)
 		}
+		switch usType := tupleUserset.Userset.(type) {
+		case *openfgapb.Userset_This:
+			continue // no need to check on Direct Relationship
+		case *openfgapb.Userset_Intersection:
+			isDirect := isDirectIntersection(usType, tk)
+			if !isDirect {
+				return serverErrors.HandleTupleValidateError(&tupleUtils.IndirectWriteError{Reason: "Attempting to write directly to an indirect only relationship", TupleKey: tk})
```
elbuo8, line 73: Seems `Reason` can be a variable.

server/commands/write.go
```diff
@@ -60,9 +60,37 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	}
 
 	for _, tk := range writes {
-		if _, err := tupleUtils.ValidateTuple(ctx, c.datastore, store, modelID, tk, dbCallsCounter); err != nil {
+		tupleUserset, err := tupleUtils.ValidateTuple(ctx, c.datastore, store, modelID, tk, dbCallsCounter)
+		if err != nil {
 			return serverErrors.HandleTupleValidateError(err)
 		}
+		switch usType := tupleUserset.Userset.(type) {
```
elbuo8, line 67: Consider putting this on its own function.

server/commands/write.go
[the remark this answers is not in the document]
craigpastro: Sorry, I just meant that I think it should be spelled `controlLoop`, not `contolLoop`. But actually, now that you are returning a bool here, I think you can just do `return true` on line 143, and remove the break entirely. WDYT?

server/commands/write.go
```diff
@@ -60,9 +60,37 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+		case *openfgapb.Userset_This:
+			continue // no need to check on Direct Relationship
+		case *openfgapb.Userset_Intersection:
+			isDirect := isDirectIntersection(usType, tk)
```
craigpastro, line 71: If you would like you could now remove the extra variable `isDirect` and just write it as
```
if !isDirectIntersection(usType, tk) { ... }
```

server/commands/write.go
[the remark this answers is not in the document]
elbuo8: You are effectively using this to have a single point of exit in the function rather returning within the loop right?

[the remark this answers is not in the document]
dblclik: @craigpastro - 😂  I missed the typo completely... twice!

@craigpastro @elbuo8 - agree, now that the functions return `bool` not `err` these are unnecessary 🙌

server/commands/write.go
dblclik: Ah good point -- I lifted these directly out of the functions that were originally returning bool, but that'll clean it up nicely (I don't foresee bespoke messages, but if we want that in the future all we need to do is add additional variables)

server/commands/write.go
dblclik: Great point -- simplifies it some more, and is in the latest commit 👍

server/commands/write.go
dblclik: @elbuo8 - Thank you for this, I like the idea and have implemented this in the latest commit.  It definitely increased readability and, if ever needed, gives a function for assessing direct/indirect relations for future use

server/commands/write.go
```diff
@@ -104,3 +109,73 @@ func handleError(err error) error {
...
+		// if Userset.type is a ComputedUserset then we know it can't be direct
+		return serverErrors.HandleTupleValidateError(&tupleUtils.IndirectWriteError{Reason: indirectWriteErrorReason, TupleKey: tk})
+	default:
+		return nil
+	}
+	return nil
+}
+
+func isDirectIntersection(nodes *openfgapb.Userset_Intersection, tk *openfgapb.TupleKey) bool {
+	isDirect := false
```
craigpastro, line 138: I think we can get rid of this variable entirely. On line 142 return true, and on line 149 return false. I think that would be even easier to read.

craigpastro (commented):
It's looking great! Thank you for updating so quickly. Sorry, one last comment from me.

server/commands/write.go
dblclik: Another great point -- I kept it in since it was there already, but we've covered all return possibilities and the code is simpler 👍

dblclik: The latest commit has the cleaned up version!

server/test/write.go
```diff
@@ -58,6 +58,196 @@ var writeCommandTests = []writeCommandTest{
...
+							},
+						},
+					},
+				},
+			},
+		},
+		// input
+		request: &openfgapb.WriteRequest{
+			Writes: &openfgapb.TupleKeys{TupleKeys: []*openfgapb.TupleKey{{
+				Object:   "repository:auth0/express-jwt",
```
rhamzeh, line 96:
```suggestion
				Object:   "repository:openfga/openfga",
```

server/errors/errors.go
```diff
@@ -116,6 +116,10 @@ func DuplicateTupleInWrite(tk *openfgapb.TupleKey) error {
 	return status.Error(codes.Code(openfgapb.ErrorCode_cannot_allow_duplicate_tuples_in_one_request), fmt.Sprintf("duplicate tuple in write: user: '%s', relation: '%s', object: '%s'", tk.GetUser(), tk.GetRelation(), tk.GetObject()))
 }
 
+func WriteToIndirectRelationError(reason string, tk *openfgapb.TupleKey) error {
+	return status.Error(codes.Code(openfgapb.ErrorCode_invalid_tuple), fmt.Sprintf("Invalid tuple '%s'. Reason: %s", tk.String(), reason))
```
rhamzeh, line 120: Considering the reason is always: `Attempting to write directly to an indirect only relationship`

Does it make sense to pass it in each case? Couldn't be added here instead?

server/test/write.go
dblclik: Hey @rhamzeh, absolutely!  I copied one of the test cases above, so do you want me to change all of those in the test file I'm currently editing in?

server/errors/errors.go
dblclik: I thought about that, but I wasn't sure if we would/do want to have a unique message for each so I kept it as an input.  Completely open to suggestions here, though, I don't have a strong feeling one way or another

server/test/write.go
rhamzeh: Ah my apologies - let's keep it as is and consistent with the above, and will raise another issue to tackle replacing later.

rhamzeh: Issue being tracked [here](https://github.com/openfga/openfga/issues/123) - resolving this comment. Thanks @dblclik!

craigpastro (approved):
LGTM. Thank you for your contribution!

server/commands/write.go
```diff
@@ -60,9 +60,14 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	}
 
 	for _, tk := range writes {
-		if _, err := tupleUtils.ValidateTuple(ctx, c.datastore, store, modelID, tk, dbCallsCounter); err != nil {
+		tupleUserset, err := tupleUtils.ValidateTuple(ctx, c.datastore, store, modelID, tk, dbCallsCounter)
+		if err != nil {
 			return serverErrors.HandleTupleValidateError(err)
 		}
+		err = validateHasDirectRelationship(tupleUserset, tk)
```
jon-whit, line 67: This comment applies here and elsewhere in the scope of this PR.

Would you mind adding newlines between logically disjoint lines? Collapsing all whitespace between logical blocks makes the code harder to read. I know it's a nit pick, but we've been trying to do that in the project 👍 

I'd appreciate it!

dblclik: @jon-whit - absolutely, thank you for pointing that out 👍 I'll tidy that up this evening!

dblclik:
@jon-whit @rhamzeh - I noticed that Semgrep and Snyk are failing, but I wasn't sure if that is expected if I don't have those accounts (I don't believe they're free offerings, but might be wrong there)

server/commands/write.go
jon-whit: Thank you!

rhamzeh:
> I noticed that Semgrep and Snyk are failing, but I wasn't sure if that is expected if I don't have those accounts (I don't believe they're free offerings, but might be wrong there)

Yeah - we are working on a better solution for those, right now they will fail b/c you don't have the secrets setup. Another fallback FOSSA license check should have ran though. But for the scope of this change, they're not blockers.

They will still run properly and alert us on merge to main.

Thanks for submitting this!

dblclik:
@jon-whit @rhamzeh - after merging main into my branch, I've got a ton of errors showing up in VSC, but `go build ...` and `go test ...` worked just fine.  Almost all of the issues are things like below, but it definitely has me worried I messes something up to see that much red 😅 

Screen Shot 2022-07-28 at 5 26 45 PM

I also noticed that this test (see below) is passing on my local but failing in the GH Action -- any thoughts on why that might be?

Screen Shot 2022-07-28 at 5 32 33 PM

jon-whit (dismissed):

server/commands/write.go
```diff
@@ -104,3 +111,75 @@ func handleError(err error) error {
...
+	case *openfgapb.Userset_Difference:
+		if !isDirectDifference(usType, tk) {
+			return serverErrors.HandleTupleValidateError(&tupleUtils.IndirectWriteError{Reason: indirectWriteErrorReason, TupleKey: tk})
+		}
+
+	case *openfgapb.Userset_ComputedUserset:
+		// if Userset.type is a ComputedUserset then we know it can't be direct
+		return serverErrors.HandleTupleValidateError(&tupleUtils.IndirectWriteError{Reason: indirectWriteErrorReason, TupleKey: tk})
+
+	default:
```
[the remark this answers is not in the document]
dblclik: Ah, @jon-whit - great catch; let me update for that case 🤦

server/commands/write.go
```diff
@@ -104,3 +111,75 @@ func handleError(err error) error {
...
+		return nil
+	}
+
+	return nil
+}
+
+func isDirectIntersection(nodes *openfgapb.Userset_Intersection, tk *openfgapb.TupleKey) bool {
```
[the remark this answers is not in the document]
dblclik: @jon-whit - That's a good point.  I was modeling the cases off of the `switch` in `check` and using the JSON resulting from DSL I write up.  Looking more deeply at how `resolveUsersets` is recursive in `expand`, I see what you mean.

Interestingly,  I've been using the playground's  DSL to JSON to write complex relations in DSL and copy the JSON, but 3+ `or`/`and` result in a list, not a set of nested objects.  Definitely good to know that nesting is valid.

For example:  `define viewer as _editor and -parent and can_share` becomes
```json
          "can_share": {
            "union": {
              "child": [
                {
                  "computedUserset": {
                    "object": "",
                    "relation": "0af-doc_owner"
                  }
                },
                {
                  "computedUserset": {
                    "object": "",
                    "relation": "_editor"
                  }
                },
                {
                  "tupleToUserset": {
                    "tupleset": {
                      "object": "",
                      "relation": "-parent"
                    },
                    "computedUserset": {
                      "object": "",
                      "relation": "0af-doc_owner"
                    }
                  }
                }
              ]
            }
          },
```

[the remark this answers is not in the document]
dblclik: @jon-whit - to clarify, what is not valid (still, I know there's been community ask about it) is the following (where the nested object is a `union` inside of an `intersection`), right?

```json
{
    "type_definitions": [
        {
            "type": "document",
            "relations": {
                "writer": {
                    "this": {}
                },
                "reader": {
                    "intersection": {
                        "child": [
                            {
                                "union": {
                                    "child": [
                                        {
                                            "this": {}
                                        },
                                        {
                                            "computedUserset": {
                                                "relation": "other"
                                            }
                                        }
                                    ]
                                }
                            },
                            {
                                "computedUserset": {
                                    "relation": "writer"
                                }
                            }
                        ]
                    }
                },
                "other": {
                    "this": {}
                }
            }
        }
    ]
}
```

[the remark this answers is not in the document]
jon-whit: @dblclik Any operand (union, exclusion, intersection)  can be nested (any number of times) within another. To determine if `this: {}` is present in any nested rewrite rule you'll have to recursively exhaust the userset rewrite rule tree.

server/commands/write.go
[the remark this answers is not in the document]
dblclik: @jon-whit - latest push has this accounted for 👍

server/commands/write.go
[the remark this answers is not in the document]
dblclik: @jon-whit - That makes sense, and I'll certainly adapt for that 🙏   The DSL parsing grammar does not support mixing operands, does that mean we need to update the parser for it to allow that?  Or, does the engine expect consistent operands as well?  (e.g. nesting is fine, as long as it is consistent, but not inconsistent -- or is this not true?)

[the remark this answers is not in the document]
jon-whit: @dblclik you are right, the DSL parser is limited in it's ability to handle nested (and mixed) operands. This is a known issue and we have plans to address it in the future. Right now the JSON structure is the superset of the two.

Sorry for the confusion. We're a bit out of sync between the DSL and the JSON support.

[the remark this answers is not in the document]
jon-whit: Added an issue in the syntax-transformer repo to track this.

https://github.com/openfga/syntax-transformer/issues/34

miparnisari (approved):

PR #125, opened 2022-07-27
ci: drop auto changelog release notes temporarily until we can resolve

jon-whit:
## Description
The `v0.1.5` CHANGELOG release notes have richer Markdown text (with special characters etc..) that cause issues with bash shell escape sequences. Prior releases didn't have such characters and consequently did not have the same issues. The purpose of this PR is to drop the automatic changelog release notes generator until we can fix this is a more elegant (automated) way. For now this will unblock us from releasing the latest `v0.1.5` release.

## Review Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,20 +24,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Generate Release Changes
-        id: changelog_reader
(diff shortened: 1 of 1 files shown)


rhamzeh (approved):
Let's add an issue for it so we don't forget

jon-whit:
Referencing #126 here

PR #129, opened 2022-07-27
fix: embed assets using go:embed

jon-whit:
## Description
The changes here fix an issue with embedded assets that can't resolve with relative paths between goreleaser released images and images built using the standard `Dockerfile` at the root of the project. This fixes the issue that caused the `v0.1.5` release to not work properly with the embedded assets.

You can test that this works with

```
docker build -t openfga/openfga:testing .
docker run -p 8080:8080 -p 3000:3000 --rm openfga/openfga:testing run

goreleaser release --snapshot --rm-dist
docker run -p 8080:8080 -p 3000:3000 --rm openfga/openfga:v0.1.6-next run

make run
```

All three of the options above should work just fine, which indicates the goreleaser image build should be fixed after we release with these modifications.

## Review Checklist


rhamzeh (approved):

rhamzeh:
Should we add to the tag tests where it can try to run the generated image before publishing?

jon-whit:
> Should we add to the tag tests where it can try to run the generated image before publishing?

Yeah, that's a good idea. We could run a prerelease build in a test job and if that succeeds then proceed with the official release (which will publish the release artifacts). Let's tackle that work in a subsequent PR. I'll add an issue for it.

Issue #132, opened 2022-07-28
Invalid tuple deletes not flagged as invalid tuple error


adriantam:
If I try to delete by writing http://localhost:8080/stores/01G0CECAB4TT41TKN3E7D7J27Q/write with the body

```
{
  "deletes": {
    "tuple_keys": [
      {
        "user": "",
        "relation": "writer",
        "object": "document:2021-budget"
      }
    ]
  }
}
```

It only fails due to tuples do not exist.
```
{
    "code": "write_failed_due_to_invalid_input",
    "message": "cannot delete a tuple which does not exist: user: '', relation: 'writer', object: 'document:2021-budget': invalid write input"
}
```

Note that when I try to write the same tuples (with no user)
```
{
  "writes": {
    "tuple_keys": [
      {
        "user": "",
        "relation": "writer",
        "object": "document:2021-budget"
      }
    ]
  }
}
```

It will fail with
```
{
    "code": "invalid_tuple",
    "message": "Invalid tuple 'object:\"document:2021-budget\" relation:\"writer\"'. Reason: missing user"
}
```

Thus, it looks like we validate against write but not read

PR #135, opened 2022-07-28
fix: need basic syntax validation on deletes

adriantam:
## Description

Even though we should allow deleting tuples where there is no current
relation / authorization model not exist, etc.  We should nevertheless
ensure the tuple itself is valid in syntax (i.e., the user is defined).

## References

- Close https://github.com/openfga/openfga/issues/132

## Review Checklist

### DIFF
--- a/pkg/tuple/validation.go
+++ b/pkg/tuple/validation.go
@@ -54,7 +54,7 @@ func (i *RelationNotFoundError) Error() string {
 // ValidateUser returns whether the user is valid.  If not, return error
 func ValidateUser(tk *openfgapb.TupleKey) error {
 	if !IsValidUser(tk.GetUser()) {
-		return &InvalidTupleError{Reason: "missing user", TupleKey: tk}
+		return &InvalidTupleError{Reason: "the 'user' field must be a non-empty string", TupleKey: tk}
 	}
 	return nil
 }
--- a/pkg/tuple/validation.go
+++ b/pkg/tuple/validation.go
@@ -51,10 +51,18 @@ func (i *RelationNotFoundError) Error() string {
 	return fmt.Sprintf("Relation '%s' not found in type definition '%s' for tuple (%s)", i.Relation, i.TypeName, i.TupleKey.String())
 }
 
+// ValidateUser returns whether the user is valid.  If not, return error
+func ValidateUser(tk *openfgapb.TupleKey) error {
+	if !IsValidUser(tk.GetUser()) {
+		return &InvalidTupleError{Reason: "missing user", TupleKey: tk}
+	}
+	return nil
+}
+
 // ValidateTuple returns whether a *openfgapb.TupleKey is valid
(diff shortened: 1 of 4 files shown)


server/commands/write.go
```diff
@@ -65,6 +65,13 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 		}
 	}
 
+	for _, tk := range deletes {
+		// For delete, we only need to ensure it is well form but no need to validate whether relation exists
+		if err := tupleUtils.ValidateUser(tk); err != nil {
```
jon-whit, line 70: We should be validating the whole tuple. We should just simplify this to:

```
for _, tk := range deletes {
    err := tupleUtils.ValidateTupleKey(tk)
    if err != nil {
        // handle error
    }
}
```

because this only validates the user component.

adriantam: There is no ValidateTupleKey - most of the validation already happened via the gRPC framework.  Only the user needs to be validated separately.  As for ValidateTuple, it validates User + the relations (which in the case of delete, we don't need to).

server/commands/write_test.go
```diff
@@ -89,3 +91,68 @@ func TestValidateWriteTuples(t *testing.T) {
...
+
+	tests := []test{
+		{
+			name:          "nil for deletes and writes",
+			deletes:       nil,
+			writes:        nil,
+			expectedError: serverErrors.InvalidWriteInput,
+		},
+		{
+			name:          "validation on tuples against write",
```
jon-whit, line 130: Can we rename these to a more appropriate name such as "write failure with invalid user".

server/commands/write_test.go
```diff
@@ -89,3 +91,68 @@ func TestValidateWriteTuples(t *testing.T) {
...
+			deletes:       []*openfgapb.TupleKey{},
+			writes:        []*openfgapb.TupleKey{badItem},
+			expectedError: serverErrors.InvalidTuple("missing user", badItem),
+		},
+		{
+			name:          "validation on tuples against delete",
```
jon-whit, line 136: Can we rename this to a more appropriate name such as "delete failure with invalid user".

pkg/tuple/validation.go
```diff
@@ -51,10 +51,18 @@ func (i *RelationNotFoundError) Error() string {
 	return fmt.Sprintf("Relation '%s' not found in type definition '%s' for tuple (%s)", i.Relation, i.TypeName, i.TupleKey.String())
 }
 
+// ValidateUser returns whether the user is valid.  If not, return error
+func ValidateUser(tk *openfgapb.TupleKey) error {
+	if !IsValidUser(tk.GetUser()) {
+		return &InvalidTupleError{Reason: "missing user", TupleKey: tk}
```
jon-whit, line 57: Can we make this more explicit? It's possible they could have provided user but explicitly provided it with an empty string (""). So I think it'd be better to return something like  `"the 'user' field must be a non-empty string"`

jon-whit (approved):

PR #137, opened 2022-07-29
fix: update test data to not include mentions of auth0 and emails

### LINKED ISSUE #123: Test data should not have emails and should use openfga/openfga as a refrence repo
- Users in tests such as [here](https://github.com/openfga/openfga/blob/main/server/test/write.go) should drop the `@auth0.com` email (generally we should avoid using emails in test data)
- Repositories in tests can be switched to `repository:openfga/openfga`?

dblclik:
## Description
Updating test files to not mention Auth0 repos, emails, etc.  Have also changed user identifiers to be `<person>@openfga` instead of keeping a `.com` suffix to imply it is an email.

## References
Resolves #123 

## Review Checklist


jon-whit (approved):
LGTM. Thanks for the contribution!

Issue #133, opened 2022-07-28
is it possible to use playground with auth?


vavsab:
Is it possible to make playground work with enabled auth?
I'm using preshared keys but [playground is not aware of it](https://github.com/openfga/openfga/blob/e07a89823f049be029f5f4f8551d1feb65d649ff/assets/playground/index.html).

I just get missing token exception on every request
![image](https://user-images.githubusercontent.com/10233442/181526525-06f7c480-4d04-4e59-aa08-a8fc7a400575.png)

jon-whit:
👋 @vavsab at this time we do not support the embedded Playground with authentication.

We do have on our roadmap a plan to add an authentication input field (or something similar) to allow you to provide a bearer token in the Playground UI. That work will likely be part of a larger effort to open-source the OpenFGA Playground source code though. We will keep you updated with these developments as they unfold.

jon-whit:
@vavsab we now have support for preshared key authentication in the OpenFGA embedded Playground. If you pull down the main branch it's supported. It will be included in our next most release. I'm closing this now that it's supported. If you find any issues with it, please don't hesitate to reach back out 👍

vavsab:
Thanks a lot 🥰

ldigiuseppe:
Any chance playground will be integrated with OIDC any time soon?

PR #141, opened 2022-08-01
feat: add support for presharedkey auth in the Playground

jon-whit:
## Description
This wires up auth support in the Playground for OpenFGA servers setup with preshared key authentication.

## References
#133 

## Review Checklist


jon-whit:
@rhamzeh @matldupont looks like the preshared auth support in the embedded Playground works mostly (see attached video). The only hiccup is that fetching the predefined model from the default Playground store fails because the local auth key isn't appropriate for the hosted Playground.

Do you guys have some recommendations on how we should resolve that elegantly?

https://user-images.githubusercontent.com/2899204/182228380-a05828ff-a989-4c71-92de-847072d534e3.mov

rhamzeh:
> The only hiccup is that fetching the predefined model from the default Playground store fails because the local auth key isn't appropriate for the hosted Playground.

Is that what's happening? Let's discuss this in an issue in that repo, because the local playground should not be talking to the hosted version.

rhamzeh (approved):
Error not related, should be fixed in Play/SDK repo

PR #144, opened 2022-08-03
fix: add authn error codes

### LINKED ISSUE #105: bug: unauthenticated/unauthorized requests return 500 status code instead of 401/403
When we introduced authn into OpenFGA we didn't introduce the corresponding API error codes for authn related API errors in [openfga/api](https://github.com/openfga/api), so every 401 or 403 (unauthenticated or unauthorized) ends up returning a 500 (internal) error code.

Steps to reproduce (for example):
```
➜ ./openfga run --authn-method preshared --authn-preshared-keys key1

➜ curl --request POST -d '{"name":"openfga-demo"}' http://localhost:8080/stores

{"code":"internal_error","message":"missing bearer token"}

➜ curl --header "Authorization: Bearer key2" --request POST -d '{"name":"openfga-demo"}' http://localhost:8080/stores

{"code":"internal_error","message":"unauthorized"}
```

To fix this we need to introduce authn errors into [openfga/api](https://github.com/openfga/api) and encode authn errors to the appropriate error code herein.

craigpastro:
## References

Closes #105 

## Review Checklist


server/authn/presharedkey/presharedkey.go
```diff
@@ -5,8 +5,15 @@ import (
...
 	grpcAuth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
-
 	"github.com/openfga/openfga/server/authn"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var (
+	ErrMissingBearerToken = status.Error(codes.Code(openfgapb.AuthErrorCode_bearer_token_missing), "missing bearer token")
```
jon-whit, line 15: Can we put these overlapping definitions (here and above in the OIDC authenticator) in the top-level shared package. `authn/authn.go` file 

That way we don't have duplication.

craigpastro: Good call. Thanks! Updated.

server/authn/presharedkey/presharedkey.go
```diff
@@ -5,10 +5,14 @@ import (
...
 	"github.com/go-errors/errors"
 	grpcAuth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
-
 	"github.com/openfga/openfga/server/authn"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
+var errUnauthenticated = status.Error(codes.Code(openfgapb.AuthErrorCode_unauthenticated), "unauthenticated")
```
jon-whit, line 14: This should move up to `authn/authn.go` as well. It's generic enough that it belongs there. Sorry, that's what I meant before, I just wasn't clear.

server/errors/encoded_errors_test.go
```diff
@@ -29,6 +29,69 @@ func TestEncodedError(t *testing.T) {
 			expectedCodeString:     "20",
 			isValidEncodedError:    false,
 		},
+		{
+			_name:                  "auth error: invalid subject",
+			errorCode:              int32(openfgapb.AuthErrorCode_auth_failed_invalid_subject),
+			message:                "error message",
+			expectedHTTPStatusCode: http.StatusUnauthorized,
+			expectedCode:           1001,
```
jon-whit, line 37: What grpc status code do these get mapped to again? We should probably be asserting that behavior as well.

craigpastro: Isn't the idea that these get mapped to the code specified in `errorCode`?

server/authn/presharedkey/presharedkey.go
craigpastro: Got it. Updated. Thanks!

server/errors/encoded_errors_test.go
jon-whit: But we map the code to a canonical HTTP status code, so why don't we do the same for the grpc protocol? It feels weird to me that we deviate from the grpc status code but map to HTTP. If a grpc client wants to integrate with our API, just like an HTTP client would, then they should have error status codes that make sense according to the protocol they are using.

craigpastro: In fact, we convert standard gRPC status codes into our own: https://github.com/openfga/openfga/blob/78045ede7707a4e27c74b1856385c0808d05944c/server/errors/encoded_errors.go#L190

jon-whit: Hmm.. Since this is what we do today let's get this in. But we really need to revisit this. It doesn't make any sense to expose a customer API error code over grpc when the grpc client is going to operate on the canonical codes of the protocol it is using (e.g. grpc). Like, for example, if I am a grpc client and I get a `Unauthenticated` status code back, then I can handle it according to the standards of protocol. But if I get back a custom error code such as 1001, then as a grpc client I don't know what the hell to do with it :) .

server/authn/oidc/oidc.go
```diff
@@ -127,7 +134,7 @@ func (oidc *RemoteOidcAuthenticator) fetchKeys() error {
 func (oidc *RemoteOidcAuthenticator) GetKeys() (*keyfunc.JWKS, error) {
 	jwks, err := keyfunc.Get(oidc.JwksURI, keyfunc.Options{
 		Client:          retryablehttp.New().StandardClient(),
```
miparnisari, line 136:
```suggestion
		Client:          oidc.httpClient,
```

server/errors/encoded_errors_test.go
miparnisari: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

So, for example, if the token has invalid claims, the gRPC status code would be `16 UNAUTHENTICATED`? Would we add the "invalid claims" as a message, too?

server/authn/oidc/oidc.go
elbuo8: Why?

craigpastro: Thank you! Updated.

miparnisari: @elbuo8 it seems wasteful to create a new client every time?

jon-whit (approved):

Issue #101, opened 2022-07-12
feat(listObjects): ListObjects API Proposal


jon-whit:
# ListObjects API Proposal

**Roadmap Issue**: https://github.com/openfga/roadmap/issues/8
**RFC Proposal**: https://github.com/openfga/rfcs/pull/3

PR #145, opened 2022-08-03
feat: implementation of ListObjects and StreamedListObjects API endpoints

jon-whit:
## Description
This adds the implementation of the ListObjects and StreamedListObjects API endpoints to the OpenFGA server as described in the [ListObjects RFC](https://github.com/openfga/rfcs/blob/main/20220714-listObjects-api.md).

## References
https://github.com/openfga/roadmap/issues/8
Closes https://github.com/openfga/openfga/issues/101
https://github.com/openfga/api/pull/21
https://github.com/openfga/rfcs/blob/main/20220714-listObjects-api.md
https://github.com/openfga/rfcs/pull/3

## Review Checklist


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/openfga/openfga
 
 go 1.18
 
+replace go.buf.build/openfga/go/openfga/api => /Users/jonwhit/github/openfga/api/proto
```
jon-whit, line 5: Waiting on https://github.com/openfga/api/pull/21 to be merged before this can be wired up. For now this is pointing at my local module so that it builds against the new API definitions.

server/server.go
```diff
@@ -155,6 +161,216 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+
+						atomic.AddUint32(&results, 1)
+					}
+				}
+
+				return nil
+			}
+
+			g.Go(fn)
+			time.Sleep(4 * time.Second)
```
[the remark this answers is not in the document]
jon-whit: We can remove that. That is residual for testing that the deadline works properly. It was added superfluously to test that behavior.

go.mod
```diff
@@ -39,6 +41,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/docker/distribution v2.8.1+incompatible // indirect
 	github.com/docker/go-units v0.4.0 // indirect
+	github.com/envoyproxy/protoc-gen-validate v0.6.7 // indirect
```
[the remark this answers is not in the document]
jon-whit: `go mod tidy` pulled it in as an indirect

Run `go mod tidy` locally on this branch and it shouldn't prune it.

server/server.go
```diff
@@ -155,6 +161,216 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+	resolvedChan := make(chan struct{})
+
+	var timeoutCtx context.Context
+	var cancel context.CancelFunc
+	if s.config.ListObjectsDeadline != 0 {
+		timeoutCtx, cancel = context.WithTimeout(ctx, s.config.ListObjectsDeadline)
+		defer cancel()
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
jon-whit: Because the `--listObjects-deadline` triggers the parent routine to return as soon as the deadline met.

server/server.go
```diff
@@ -155,6 +161,216 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+			g.Go(fn)
+			time.Sleep(4 * time.Second)
+		}
+
+		err := g.Wait()
+		if err != nil {
+			// todo: handle error
+		}
+
+		resolvedChan <- struct{}{}
```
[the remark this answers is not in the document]
jon-whit: There are 3 cases.

1) The `--listObjects-max-results` limit is reached
2) The `--listObjects-deadline` is reached
3) All objects in the graph are evaluated and the ListObjects processed is fully resolved (before the deadline and max-results limits are reached).

This accounts for the last case.

storage/postgres/postgres.go
```diff
@@ -108,6 +108,40 @@ func (p *Postgres) Close(ctx context.Context) error {
 	return nil
 }
 
+func (p *Postgres) ReadRelationshipTuples(
+	ctx context.Context,
+	filter storage.ReadRelationshipTuplesFilter,
+	opts ...storage.QueryOption,
```
craigpastro, line 114: This seems to be unused?

server/server.go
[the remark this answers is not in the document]
miparnisari: i still don't understand 😆 where are we telling the goroutine to stop doing work if the deadline is met?

server/server.go
```diff
@@ -155,6 +161,216 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+			// todo: handle error
+		}
+
+		close(objectIDsChan)
+	}()
+
+	resolvedChan := make(chan struct{})
+
+	var objectIDs []string
+	go func() {
```
miparnisari, line 253: why do we need this goroutine?

storage/postgres/postgres.go
```diff
@@ -108,6 +108,40 @@ func (p *Postgres) Close(ctx context.Context) error {
...
+) (storage.TupleIterator, error) {
+	ctx, span := p.tracer.Start(ctx, "postgres.QueryRelationshipTuples")
+	defer span.End()
+
+	sql := "SELECT store, object_type, object_id, relation, _user, ulid, inserted_at FROM tuple WHERE store='%v'"
+	args := []interface{}{filter.StoreID}
+
+	if filter.OptionalObjectType != "" {
+		sql += " AND object_type='%v'"
```
craigpastro, line 123: Use squirrel here? But this is nice! If we were not required to use positional arguments like `$1, $2, ...` and could just use `$, $, ...` then this would be as nice as using a query builder. We can't just use `$, $, ...` right? I suppose we could keep track of a counter, but then it would be a bit of a mess and, anyway, there is squirrel.

storage/storage.go
```diff
@@ -43,13 +43,45 @@ type TupleIterator interface {
 type Writes = []*openfgapb.TupleKey
 type Deletes = []*openfgapb.TupleKey
 
+// QueryOptions represent constraints that can be applied to a query to modify the response
+// behavior of the query.
+type QueryOptions struct {
```
craigpastro, line 48: How many options do you envision having? If we only will have `limit` perhaps the extra complexity is not worth it?

server/server.go
```diff
@@ -155,6 +163,247 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+		var objectsFound uint32
+		for {
+			fmt.Println("inside for loop...")
+			time.Sleep(1 * time.Second)
+			if atomic.LoadUint32(&objectsFound) >= s.config.ListObjectsMaxResults {
+				break
+			}
+
+			t, err := iter.Next()
+			if err != nil && err == storage.TupleIteratorDone {
```
craigpastro, line 232: What if the err is not nil and also not a `TupleIteratorDone`?

miparnisari: good point! i need to send the error back to the parent.

pkg/cmd/service/service.go
```diff
@@ -145,12 +151,12 @@ type Config struct {
...
-			MaxTypesPerAuthorizationModel: 100,
-			ChangelogHorizonOffset:        0,
-			ResolveNodeLimit:              25,
-		},
+		MaxTuplesPerWrite:             100,
+		MaxTypesPerAuthorizationModel: 100,
+		ChangelogHorizonOffset:        0,
+		ResolveNodeLimit:              25,
+		ListObjectsDeadline:           0,
+		ListObjectsMaxResults:         math.MaxUint32,
```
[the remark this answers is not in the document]
miparnisari: there is a 3 second timeout on calls but i do think it makes sense to set a numeric limit

server/test/list_objects.go
```diff
@@ -0,0 +1,185 @@
...
+		listObjectsQuery := commands.NewListObjectsQuery(backend, tracer, logger.NewNoopLogger(), telemetry.NewNoopMeter(), defaultListObjectsDeadline, defaultListObjectsMaxResults, defaultResolveNodeLimit)
+		runListObjectsTests(t, ctx, testCases, listObjectsQuery)
+	})
+}
+
+func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObjectsTestCase, listObjectsQuery *commands.ListObjectsQuery) {
+	var res *openfgapb.ListObjectsResponse
+	var err error
+	for _, test := range testCases {
+		res, err = listObjectsQuery.Execute(ctx, test.request)
```
miparnisari, line 104: This file only tests the non-streaming version... still need to find out if there is a way of testing the streaming version.

server/commands/list_objects.go
```diff
@@ -0,0 +1,260 @@
...
+					break
+				} else {
+					errChan <- err
+					close(errChan)
+				}
+			}
+			checkFunction := func() error {
+				_, objectID := tuple.SplitObject(t.Key.Object)
+				if uniqueObjectIdsSet.Has(objectID) {
+					return nil
```
miparnisari, line 98: to avoid unnecessary checks + to ensure resulting array doesn't have duplicates

server/commands/list_objects.go
```diff
@@ -0,0 +1,260 @@
...
+				break
+			}
+
+			t, err := iter.Next()
+			if err != nil {
+				if err == storage.TupleIteratorDone {
+					break
+				} else {
+					errChan <- err
+					close(errChan)
```
miparnisari, line 92: err this probably doesn't belong here but if I put it next to `close(resultsChan)`, it exits through `case genericError := <-errChan` 🤔

server/commands/list_objects.go
```diff
@@ -0,0 +1,261 @@
...
+	case <-resolvedChan:
+		return &openfgapb.ListObjectsResponse{
+			ObjectIds: objectIDs,
+		}, nil
+	case genericError := <-errChan:
+		return nil, genericError
+	}
+}
+
+func (q *ListObjectsQuery) ExecuteStreamed(ctx context.Context, modelID string, req *openfgapb.StreamedListObjectsRequest, srv openfgapb.OpenFGAService_StreamedListObjectsServer) error {
```
miparnisari, line 160: I'm seeing this error

```
2022-08-05T09:58:52.651-0700    ERROR   failed to set grpc header       {"error": "rpc error: code = Internal desc = grpc: failed to fetch the stream from the context context.Background.WithValue(type trace.traceContextKeyType, val <not Stringer>)", "header": "openfga-authorization-model-id"}
```

but it still works

server/server.go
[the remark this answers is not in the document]
miparnisari: Hmmm ok, i wonder why the Checks don't listen for cancellation events 🤔

storage/postgres/postgres.go
miparnisari: i'm not sure what the intention was for this

server/commands/list_objects.go
```diff
@@ -0,0 +1,283 @@
...
+			if err != nil {
+				if err == storage.TupleIteratorDone {
+					break
+				} else {
+					errChan <- err
+				}
+			}
+			checkFunction := func() error {
+				_, objectID := tuple.SplitObject(t.Key.Object)
+				if uniqueObjectIdsSet.Has(objectID) {
```
jon-whit, line 90: Let's just use a DISTINCT in the database query rather than filtering them out here.

storage/postgres/postgres.go
jon-whit: The goal is to use it to provide limits and grouping/aggregate functions. The goal is to have `ReadRelationshipTuples` generically replace all of the specific `ReadUserTuple`, `Read`, and `ReadUsersetTUples` etc.. We don't need to bloat our storage interface with all of those methods.

A single ReadRelationshipTuples method with optional filters and other query options is more elegant IMO.

storage/storage.go
jon-whit: I envision having all of the options we use in our different Read, ReadUserTuple, and ReadUsersetTuples, etc.. I think we mostly use LIMIT and GROUP BY, but we may use a couple others. These are modifiers/constraints for the query in addition to the filters.

server/commands/check.go
```diff
@@ -44,6 +44,14 @@ func NewCheckQuery(datastore storage.OpenFGADatastore, t trace.Tracer, m metric.
 
 // Execute the query in `checkRequest`, returning the response or an error.
 func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckRequest) (*openfgapb.CheckResponse, error) {
+	//if rand.Intn(100) < 10 { // 10% of failure
```
jon-whit, line 47: Mind doing a quick pass to removed commented out code and debug print statements? I've seen a couple in this PR.

pkg/utils/utils.go
```diff
@@ -125,3 +125,27 @@ func (r *ResolutionMetadata) Fork() *ResolutionMetadata {
...
 }
+
+type ConcurrentSet struct {
+	M map[string]bool
+	sync.RWMutex
+}
+
+func NewConcurrentSet() *ConcurrentSet {
+	return &ConcurrentSet{
+		M: make(map[string]bool),
```
jon-whit, line 136: We should try not to use this for ListObjects because it creates contention between goroutines unnecessarily.  It creates read/write lock contention between them when we don't need to.

server/test/list_objects.go
jon-whit: Consume the stream in its entirety and then make your test assertions. That's how we should test the streaming version.

server/commands/list_objects.go
miparnisari: The Postgres query is doing something like

`SELECT store, object_type, object_id, relation, _user, ulid, inserted_at FROM tuple WHERE store='01G7ZXJ1A37XA9KC16QSGZP2H4' AND object_type='document'` 

which should already yield distinct rows, no? Because each `object_type, object_id, relation, _user` combination is unique

Also, there is no guarantee that other data stores can support  `distinct`s

pkg/cmd/run.go
```diff
@@ -299,15 +299,21 @@ func bindFlags(cmd *cobra.Command) {
...
 
-	cmd.Flags().Int("changelog-horizon-offset", defaultConfig.OpenFGA.ChangelogHorizonOffset, "the offset (in minutes) from the current time. Changes that occur after this offset will not be included in the response of ReadChanges")
+	cmd.Flags().Int("changelog-horizon-offset", defaultConfig.ChangelogHorizonOffset, "the offset (in minutes) from the current time. Changes that occur after this offset will not be included in the response of ReadChanges")
 	cmdutil.MustBindPFlag("changelogHorizonOffset", cmd.Flags().Lookup("changelog-horizon-offset"))
 
-	cmd.Flags().Int("resolve-node-limit", int(defaultConfig.OpenFGA.ResolveNodeLimit), "defines how deeply nested an authorization model can be")
+	cmd.Flags().Int("resolve-node-limit", int(defaultConfig.ResolveNodeLimit), "defines how deeply nested an authorization model can be")
 	cmdutil.MustBindPFlag("resolveNodeLimit", cmd.Flags().Lookup("resolve-node-limit"))
+
+	cmd.Flags().Duration("listObjects-deadline", defaultConfig.ListObjectsDeadline, "the timeout deadline for serving ListObjects requests")
```
elbuo8, line 314: Why does it use capital when everything else doesn't?

elbuo8: Does it accept a particular notation because it's a `Duration`? Or it's in `ms`?

pkg/cmd/run.go
```diff
@@ -299,15 +299,21 @@ func bindFlags(cmd *cobra.Command) {
...
+	cmdutil.MustBindPFlag("listObjectsDeadline", cmd.Flags().Lookup("listObjects-deadline"))
+
+	cmd.Flags().Uint32("listObjects-max-results", defaultConfig.ListObjectsMaxResults, "the maximum results to serve for ListObjects requests")
```
elbuo8, line 317:
```suggestion
	cmd.Flags().Uint32("listObjects-max-results", defaultConfig.ListObjectsMaxResults, "the maximum results to serve for ListObjects responses")
```

pkg/cmd/service/service.go
```diff
@@ -145,12 +150,12 @@ type Config struct {
...
-			MaxTuplesPerWrite:             100,
-			MaxTypesPerAuthorizationModel: 100,
-			ChangelogHorizonOffset:        0,
-			ResolveNodeLimit:              25,
-		},
+		MaxTuplesPerWrite:             100,
+		MaxTypesPerAuthorizationModel: 100,
+		ChangelogHorizonOffset:        0,
+		ResolveNodeLimit:              25,
+		ListObjectsDeadline:           3 * time.Second, // we have a 3-second timeout anyway
```
elbuo8, line 157: Where do we have that timeout? 🤔

server/commands/list_objects.go
```diff
@@ -0,0 +1,370 @@
...
+)
+
+type ListObjectsQuery struct {
+	datastore             storage.OpenFGADatastore
+	logger                logger.Logger
+	tracer                trace.Tracer
+	meter                 metric.Meter
+	ListObjectsDeadline   time.Duration
+	ListObjectsMaxResults uint32
+	ResolveNodeLimit      uint32
```
elbuo8, line 31: Any reason why these properties are public?

server/commands/list_objects.go
```diff
@@ -0,0 +1,370 @@
...
+			if atomic.LoadUint32(&objectsFound) >= q.ListObjectsMaxResults {
+				break
+			}
+
+			t, err := iter.Next()
+			if err != nil {
+				if err == storage.TupleIteratorDone {
+					break
+				} else {
+					errChan <- err
```
elbuo8, line 89: I think we are missing a `continue` statement here.

elbuo8: I'd remove the `else` and just have:
```
errChan <- err
continue
```

pkg/cmd/service/service.go
miparnisari: https://github.com/openfga/openfga/blob/a5fb26c2ac203dc8c3c9f89db18cad3205e89a4b/pkg/cmd/service/service.go#L172

server/commands/list_objects.go
```diff
@@ -0,0 +1,370 @@
...
+	iter, err := q.getTupleIterator(ctx, req.StoreId, req.Type)
+	if err != nil {
+		return nil, err
+	}
+	defer iter.Stop()
+
+	g := new(errgroup.Group)
+	g.SetLimit(MaximumConcurrentChecks)
+
+	uniqueObjectIdsSet := utils.NewConcurrentSet()
```
elbuo8, line 62: I'm unsure this should be handled by the service layer. I believe it should be the storage layer. Wether it's supported top level by the database or done in the implementation (which could be done at the iterator level or some other part). 

We must add tests in the storage layer to ensure the implementations are not doing duplicates.

server/commands/list_objects.go
```diff
@@ -0,0 +1,370 @@
...
+	var cancel context.CancelFunc
+	if q.ListObjectsDeadline != 0 {
+		timeoutCtx, cancel = context.WithTimeout(ctx, q.ListObjectsDeadline)
+		defer cancel()
+	}
+
+	objectIDs := make([]string, 0)
+
+	go func() {
+		var objectsFound uint32
```
elbuo8, line 76: What's the reasoning for using `uint32` instead of `int32`?

server/commands/list_objects.go
elbuo8: https://github.com/openfga/openfga/pull/145/files#r939438353

To further add:

> which should already yield distinct rows, no? Because each object_type, object_id, relation, _user combination is unique

But we are looking for unique `object_type` + `object_id` which doesn't have the same constraints and will return duplicates.

server/commands/list_objects.go
```diff
@@ -0,0 +1,370 @@
...
+				}
+
+				return nil
+			}
+
+			g.Go(checkFunction)
+		}
+
+		// iterate over all tuples in the contextual tuples input
+		if req.ContextualTuples != nil {
```
elbuo8, line 125: I'd do these first because it's a pre-defined quantity of routines which (if I remember correctly) is less than the total number of concurrency, so these could be resolved without having to wait on possibly 100 tuples returned on the iterator.

server/commands/list_objects.go
miparnisari: 🤷‍♀️  does it make a difference?

server/commands/list_objects.go
miparnisari: yessss

server/commands/list_objects.go
miparnisari: nope, made private

server/commands/list_objects.go
```diff
@@ -0,0 +1,370 @@
...
+			for _, t := range req.ContextualTuples.TupleKeys {
+				t := t
+				checkFunction := func() error {
```
elbuo8, line 128: Seems like this could be refactored since it's kinda the same function 🤔

miparnisari: I thought about this... i've got duplicate code everywhere, with subtle differences: for example, when checking for tuples in the store (the first `for` loop), the Checks don't need to pass in `req.ContextualTuples`, but in the second one, I do need to pass it 🤦‍♀️ 

And in the streaming version, everything is the same except of how I return results :(

server/commands/list_objects.go
elbuo8: That's why I'm asking the selection. `atomic.LoadUint32(&objectsFound)` as opposed `objectsFound >=`, just making sure that we don't mess up in the future.

server/commands/list_objects.go
```diff
@@ -0,0 +1,372 @@
...
+		StoreID:            storeID,
+		OptionalObjectType: targetObjectType,
+	})
+	if err != nil {
+		return nil, serverErrors.NewInternalError("", err)
+	}
+	return iter, nil
+}
+
+func (q *ListObjectsQuery) validateInput(ctx context.Context, storeID string, targetObjectType string, authModelID string, relation string) error {
```
elbuo8, line 351: Why is this not using https://github.com/openfga/openfga/blob/main/storage/storage.go#L89?

pkg/cmd/service/service.go
craigpastro, line 157:
```suggestion
		ListObjectsDeadline:           3 * time.Second, // there is a 3-second timeout anyway
```

server/test/list_objects.go
```diff
@@ -0,0 +1,192 @@
...
+		if test.expectedError == nil && err != nil {
+			t.Fatalf("[%s] Expected no error but got '%s'", test._name, err)
+		}
+
+		if test.expectedError != nil && err == nil {
+			t.Fatalf("[%s] Expected an error '%s' but got nothing", test._name, test.expectedError)
+		}
+
+		if test.expectedError != nil && err != nil && !strings.Contains(test.expectedError.Error(), err.Error()) {
+			t.Fatalf("[%s] Expected error '%s', actual '%s'", test._name, test.expectedError, err)
```
craigpastro, line 151: I think `require.ErrorIs`, or something very similar, can consolidate all these lines.

server/commands/list_objects.go
miparnisari: > But we are looking for unique `object_type` + `object_id` which doesn't have the same constraints and will return duplicates.

Right. What I meant was that adding DISTINCT to the specific query we have today wouldn't remove duplicates. Unless we write a new query.

server/commands/list_objects.go
miparnisari: Good point, will refactor.

pkg/cmd/run.go
miparnisari: Ahhh. Changed it to `s`.

server/commands/list_objects.go
miparnisari: `atomic.LoadUint32` ensures that we don't mess up the counter from different goroutines

server/commands/list_objects.go
miparnisari: Ok, refactored :)

server/commands/list_objects.go
miparnisari: This has been refactored

pkg/utils/utils.go
miparnisari: Removed

server/commands/list_objects.go
miparnisari: Wrote a new query :P

server/test/list_objects.go
miparnisari: I'm not sure how to instantiate the stream :/ @jon-whit do you mind sending a follow-up PR for these tests? Or if you can point me to an example?

server/commands/list_objects.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resultsChan := make(chan string, q.listObjectsMaxResults)
+	errChan := make(chan error)
+	resolvedChan := make(chan struct{})
+
+	timeoutCtx := ctx
+	var cancel context.CancelFunc
```
elbuo8, line 58: Why declare here if never used again?

server/commands/list_objects.go
```diff
@@ -0,0 +1,232 @@
...
+			return serverErrors.NewInternalError("Timeout exceeded", timeoutCtx.Err())
+		case objectID, ok := <-resultsChan:
+			if ok {
+				err := srv.Send(&openfgapb.StreamedListObjectsResponse{
+					ObjectId: objectID,
+				})
+				if err != nil {
+					return serverErrors.NewInternalError("", err)
+				}
+			} else {
```
elbuo8, line 118: Rather than having an `else`, why not check on the `if` and exit realy?

server/commands/list_objects.go
```diff
@@ -0,0 +1,232 @@
...
+		q.performChecks(timeoutCtx, req.StoreId, req.AuthorizationModelId, req.Type, req.Relation, req.User, req.ContextualTuples, resultsChan, errChan, resolvedChan)
+	}()
+
+	for {
+		select {
+		case <-timeoutCtx.Done():
+			return serverErrors.NewInternalError("Timeout exceeded", timeoutCtx.Err())
+		case objectID, ok := <-resultsChan:
+			if ok {
+				err := srv.Send(&openfgapb.StreamedListObjectsResponse{
```
elbuo8, line 112: `if err:= ... ; err != nil {`

server/commands/list_objects.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+		case genericError, ok := <-errChan:
+			if ok {
+				return serverErrors.NewInternalError("", genericError)
+			}
+		}
+	}
+}
+
+func (q *ListObjectsQuery) getUniqueObjects(ctx context.Context, storeID, targetObjectType string, ctxTuples *openfgapb.ContextualTupleKeys) ([]string, error) {
```
elbuo8, line 130: Do we need `Unique` in the name?

server/commands/list_objects.go
```diff
@@ -0,0 +1,232 @@
...
+		uniqueSet := make(map[string]bool, len(uniqueObjects))
+		for _, o := range uniqueObjects {
+			_, found := uniqueSet[o]
+			if !found {
+				uniqueSet[o] = true
+			}
+		}
+
+		for _, ctxTuple := range ctxTuples.TupleKeys {
+			_, found := uniqueSet[ctxTuple.Object]
```
elbuo8, line 150: I'm guessing it matters the `type` if you are using `targetObjectType` otherwise it's just noise 🤔

server/commands/list_objects.go
```diff
@@ -0,0 +1,232 @@
...
+			return nil
+		}
+
+		g.Go(checkFunction)
+	}
+
+	err = g.Wait()
+	if err != nil {
+		// we don't want to abort everything if one of the checks failed.
+		q.logger.Warn("ListObjects errored: ", logger.Error(err))
```
elbuo8, line 226: If that is the case you should use the error channel within `checkFunction` otherwise in the first error it get's it will exit.

elbuo8: https://github.com/openfga/openfga/pull/145/files#diff-7f35d19e9a3e7af70b152b10bcb55e9906ba5fbece03b60a5a03dbc2d1d9ec1bR209

Should return `nil` but write to the error channel.

server/commands/list_objects.go
miparnisari: If I do `timeoutCtx, cancel :=` i get another error 🤣

server/commands/list_objects.go
miparnisari: image

server/commands/list_objects.go
miparnisari: Excellent point!

server/commands/list_objects.go
miparnisari: No, if I write to the error channel it will have the same effect because of the `case error <- errChan`... 

I'll return nil and just log the error

server/server.go
```diff
@@ -155,6 +157,42 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
 	return server, nil
 }
 
+func (s *Server) ListObjects(ctx context.Context, req *openfgapb.ListObjectsRequest) (*openfgapb.ListObjectsResponse, error) {
+	storeID := req.GetStoreId()
```
elbuo8, line 161: Both methods are missing tracing.

elbuo8 (dismissed):

storage/storage.go
```diff
@@ -43,13 +43,25 @@ type TupleIterator interface {
...
 type TupleBackend interface {
 	// Read the set of tuples associated with `store` and `key`, which may be partially filled. A key must specify at
 	// least one of `Object` or `User` (or both), and may also optionally constrain by relation. The caller must be
 	// careful to close the TupleIterator, either by consuming the entire iterator or by closing it.
 	Read(context.Context, string, *openfgapb.TupleKey) (TupleIterator, error)
 
+	// ListObjectsByType returns all the objects of a specific type.
+	// You can assume that the type has already been validated.
+	// The result can't have duplicate elements.
+	ListObjectsByType(ctx context.Context, filter ListObjectsFilter) ([]string, error)
```
jon-whit, line 63: I think if we're going to do this differently than the proposal described, then at a minimum we should yield an iterator instead of a slice. Yielding an iterator is important here to reduce the memory consumption of downstream components that call this method.

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (q *ListObjectsQuery) validateInput(ctx context.Context, storeID string, targetObjectType string, authModelID string, relation string) error {
+	definition, err := q.datastore.ReadTypeDefinition(ctx, storeID, authModelID, targetObjectType)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return serverErrors.TypeNotFound(targetObjectType)
+		}
+		return err
+	}
+	_, ok := definition.Relations[relation]
```
jon-whit, line 169: We have this same kind of validation elsewhere. Can we please refactor this into a shared component so that it can be used elsewhere?

We use it in `validateAndAuthenticateTuple` and others.

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	go func() {
+		q.performChecks(timeoutCtx, req.StoreId, req.AuthorizationModelId, req.Type, req.Relation, req.User, req.ContextualTuples, resultsChan, errChan, resolvedChan)
+	}()
+
+	for {
+		select {
+		case <-timeoutCtx.Done():
+			return serverErrors.NewInternalError("Timeout exceeded", timeoutCtx.Err())
```
jon-whit, line 109: We shouldn't return an error here. If the timeout was exceeded then we're done. The timeout deadline is merely to control the maximum amount of time to spent streaming/aggregating the results. If we've hit it, then we just return what we've accumulated up to that point.

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	resultsChan := make(chan string, q.listObjectsMaxResults)
+	errChan := make(chan error)
+	resolvedChan := make(chan struct{})
+
+	timeoutCtx := ctx
+	if q.listObjectsDeadlineInSeconds != 0 {
+		var cancel context.CancelFunc
+		timeoutCtx, cancel = context.WithTimeout(ctx, time.Second*time.Duration(q.listObjectsDeadlineInSeconds))
```
jon-whit, line 98: Why do we use an integer just to turn it back into a Duration? Let's just use a duration type.

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+		defer cancel()
+	}
+
+	go func() {
+		q.performChecks(timeoutCtx, req.StoreId, req.AuthorizationModelId, req.Type, req.Relation, req.User, req.ContextualTuples, resultsChan, errChan, resolvedChan)
+	}()
+
+	select {
+	case <-timeoutCtx.Done():
+		return nil, serverErrors.NewInternalError("Timeout exceeded", timeoutCtx.Err())
```
jon-whit, line 70: Same comment as in https://github.com/openfga/openfga/pull/145/files#r941803259

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+const (
+	MaximumConcurrentChecks = 100 // todo(jon-whit): make this configurable, but for now limit to 100 concurrent checks
+)
+
+type ListObjectsQuery struct {
+	datastore                    storage.OpenFGADatastore
+	logger                       logger.Logger
+	tracer                       trace.Tracer
+	meter                        metric.Meter
+	listObjectsDeadlineInSeconds int
```
jon-whit, line 28: Duration instead please?

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+	listObjectsMaxResults        uint32
+	resolveNodeLimit             uint32
+}
+
+// NewListObjectsQuery creates a ListObjectsQuery
+func NewListObjectsQuery(datastore storage.OpenFGADatastore, tracer trace.Tracer, logger logger.Logger, meter metric.Meter, listObjectsDeadlineInSeconds int, listObjectsMaxResults uint32, ResolveNodeLimit uint32) *ListObjectsQuery {
```
jon-whit, line 34: Could we refactor this signature so that it's not so painful on the eyes? When there is more than 3 inputs I think it's helpful to use the Options pattern like so:

```
type ListObjectsQueryOption func(*ListObjectsQuery)

func WithTracer(tracer trace.Tracer) ListObjectsQueryOption {
    return func(q *ListObjectsQuery) {
        q.tracer = tracer
    }
}

func NewListObjectsQuery(opts ...ListObjectsQueryOption) {
    l := &ListObjectsQuery{}
    for _, opt := range opts {
        opt(l)
    }

    return l
}
```

pkg/cmd/service/service.go
```diff
@@ -107,8 +107,24 @@ type PlaygroundConfig struct {
...
+	Addr    string
+}
+
+type Config struct {
+	// If you change any of these settings, please update the documentation at https://github.com/openfga/openfga.dev/blob/main/docs/content/intro/setup-openfga.mdx
+
+	// ListObjectsDeadlineInSeconds defines the maximum amount of time to accumulate ListObjects results
+	// before the server will respond. This is to protect the server from misuse of the
+	// ListObjects endpoints.
+	ListObjectsDeadlineInSeconds int
```
jon-whit, line 122: Why not duration? It's a duration type. If, at the command line, I want to specify `--listObjects-deadline=3ms` then I should be able to do that. Deadlines don't have to be restricted to seconds if I want a strict deadline.

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+				TupleKey: &openfgapb.TupleKey{
+					Object:   object,
+					Relation: relation,
+					User:     user,
+				},
+				ContextualTuples: ctxTuples,
+			})
+			if err != nil {
+				// ignore the error. we don't want to abort everything if one of the checks failed.
+				q.logger.Warn("Check errored: ", logger.Error(err))
```
jon-whit, line 211: Shouldn't this be an error level statement?

server/commands/list_objects.go
```diff
@@ -0,0 +1,234 @@
...
+		return err
+	}
+	_, ok := definition.Relations[relation]
+	if !ok {
+		return serverErrors.UnknownRelationWhenListingObjects(relation, targetObjectType)
+	}
+	return nil
+}
+
+func (q *ListObjectsQuery) performChecks(timeoutCtx context.Context, storeID string, authModelID string, objectType string, relation string, user string, ctxTuples *openfgapb.ContextualTupleKeys, resultsChan chan<- string, errChan chan<- error, resolvedChan chan<- struct{}) {
```
jon-whit, line 176: Huge method signature. Can we please refactor to be easier to read? Maybe consider using the Option pattern or an input struct instead?

storage/storage.go
miparnisari: Ah the reason i don't return a `tupleIterator` is that i don't have tuples to return, just `object_id`s. I could build a brand new iterator just for this, but it felt overkill

server/commands/list_objects.go
miparnisari: I see only one other duplication, in `validateAndAuthenticateTuple` like you said, but the error returned for when the relation is invalid is different because in my PR i don't have a tuple to reference. I'm inclined to keep this as is

storage/storage.go
jon-whit: I think it's important, because the StreamedListObjects API shouldn't have to have a big memory footprint like this. That was one of the whole reasons for the streaming variation, was to only keep iterators/cursors as it evaluates the objects instead of loading them all into memory.

server/test/list_objects.go
jon-whit: See:
https://grpc.io/docs/languages/go/basics/#client-side-streaming-rpc
https://grpc.io/docs/languages/go/basics/#client-side-streaming-rpc

server/commands/list_objects.go
jon-whit: This logic is overlapping though. You're only looking up the `*openfgapb.Userset` value for a specific `(objectType, relation)` pair. If the error messages turns a tuple in one and we only return a objectType error message, here then we should refactor to just return a generic error such as `fmt.Errorf("'%s' undefined for object type '%s'", relation, objectType)`. The code duplication isn't warranted just for a different error message.

This is also common enough functionality that will be useful elsewhere. Every Expand, Check, ListObjects, etc.. will start with validating a relation on a given objectType exists. So it should be shared.

server/commands/list_objects.go
miparnisari: The reasoning here was that it wasn't intuitive how to pass a duration as a configuration value

jon-whit: It follows the standard Go duration syntax and the `cobra` library standard, which is quite common across the Go community though. If I want a `ms` deadline or a `minute` deadline it's nice to be able to use the standard Go syntax and do `--listObjects-deadline=100ms` or `--listObjects-deadline=2m` instead of calculating the seconds equivalent (which doesn't work for fractions if you want to specify something < 1s).

server/commands/list_objects.go
jon-whit: We have it elsewhere and it's cryptic. Method signatures this long are not great for readability. Clean Code standards would highly discourage this long of a method signature. The parameter pattern/option pattern or a input/output struct pattern would be a cleaner code pattern.

Alternatively, if we could just drop the constructor. `ListObjecstQuery` is a publicly exported struct anyways, and the constructor isn't buying us anything. You may as well just construct a `&ListObjectsQuery{...}` with the fields it needs.

server/commands/list_objects.go
miparnisari: Ah you're right it doesn't make sense for the streaming version. Removed it.

For the regular API however, if I insert an artificial delay of 5 seconds in one of the Checks, it doesn't matter what I put here: the whole request aborts at the 3 second mark with "context timeout exceeded" which I think is coming from https://github.com/openfga/openfga/blob/65c5f3bd3d50eb22b2a7fc0f41051eba9631da2e/pkg/cmd/service/service.go#L172

pkg/cmd/service/service.go
miparnisari: changed :)

server/commands/list_objects.go
miparnisari: Dropped the constructor

storage/storage.go
miparnisari: Good point. Implemented the iterators 😄

storage/postgres/iterators.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (t *tupleIterator) Stop() {
+	t.rows.Close()
+}
+
+type PostgresObjectIterator struct {
+	rows pgx.Rows
+}
+
+func (o *PostgresObjectIterator) Next() (string, error) {
```
craigpastro, line 86: We are already in the Postgres package so does objectIterator make more sense? Or, perhaps even more cooler, is there some generic method that would work for both tuples and objects?

storage/postgres/postgres.go
```diff
@@ -108,6 +108,24 @@ func (p *Postgres) Close(ctx context.Context) error {
...
+	ctx context.Context,
+	filter storage.ListObjectsFilter,
+) (storage.ObjectIterator, error) {
+	ctx, span := p.tracer.Start(ctx, "postgres.ListObjectsByType")
+	defer span.End()
+
+	sql := "SELECT DISTINCT object_type, object_id FROM tuple WHERE store='%v' AND object_type='%v'"
+	args := []interface{}{filter.StoreID, filter.ObjectType}
+
+	rows, err := p.pool.Query(ctx, fmt.Sprintf(sql, args...))
```
craigpastro, line 121: I think the use of fmt.Sprintf means that this may be vulnerable to injections. You can simple use positional arts here $1, $2, to remove that use.

server/commands/list_objects.go
```diff
@@ -0,0 +1,271 @@
...
+	defer iter.Stop()
+
+	// iterate over the contextual tuples
+	if input.ctxTuples != nil {
+		for _, t := range input.ctxTuples.TupleKeys {
+			if atomic.LoadUint32(&objectsFound) >= q.ListObjectsMaxResults {
+				break
+			}
+			t := t
+			checkFunction := func() error {
```
elbuo8, line 184: Why can't this be a generic function used across both?

storage/postgres/iterators.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, err
+	}
+	return record.asTuple(), nil
+}
+
+func (t *tupleIterator) Stop() {
+	t.rows.Close()
+}
+
+type PostgresObjectIterator struct {
```
elbuo8, line 82: Bit redundant 🤔

server/test/list_objects.go
```diff
@@ -0,0 +1,192 @@
...
+		listObjectsQuery := commands.NewListObjectsQuery(backend, tracer, logger.NewNoopLogger(), telemetry.NewNoopMeter(), defaultListObjectsDeadline, defaultListObjectsMaxResults, defaultResolveNodeLimit)
+		runListObjectsTests(t, ctx, testCases, listObjectsQuery)
+	})
+}
+
+func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObjectsTestCase, listObjectsQuery *commands.ListObjectsQuery) {
+	var res *openfgapb.ListObjectsResponse
+	var err error
+	for _, test := range testCases {
+		res, err = listObjectsQuery.Execute(ctx, test.request)
```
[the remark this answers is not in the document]
miparnisari: They are wrapped 😄

server/commands/list_objects.go
```diff
@@ -269,3 +221,31 @@ func (q *ListObjectsQuery) performChecks(timeoutCtx context.Context, input *Perf
...
+		TupleKey: &openfgapb.TupleKey{
+			Object:   object,
+			Relation: input.relation,
+			User:     input.user,
+		},
+		ContextualTuples: input.ctxTuples,
+	})
+	if err != nil {
+		// ignore the error. we don't want to abort everything if one of the checks failed.
+		q.Logger.Error("Check errored: ", logger.Error(err))
```
elbuo8, line 241: ErrorWithContext

server/commands/list_objects.go
jon-whit: It is, yeah. We should probably default the UpstreamTimeout to infinity instead of 3sec. I think if `UpstreamTimeout` is more strict that the `ListObjectsDeadline` than this behavior is expected. But I think the default should be to not have an UpstreamTimeout unless it's specified.

miparnisari:
Pending:
- the test for the streaming endpoint
- remove the duplicate validation of the type and relation
- the “failed to set grpc header” error being written on the streaming endpoint
- (a nice to have) checks should listen to cancellation events

server/commands/list_objects.go
miparnisari: 👍 returning partial results now.

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	allResults, _ := q.Meter.SyncInt64().Counter(
+		"openfga.listobjects.allresults",
+		instrument.WithDescription("Number of object IDs returned"),
+		instrument.WithUnit(unit.Dimensionless),
+	)
+	partialResults, _ := q.Meter.SyncInt64().Counter(
+		"openfga.listobjects.partialresults",
+		instrument.WithDescription("Number of object IDs returned after the API timed out"),
```
elbuo8, line 49: I would think we would count the total requests which were partial, not the total results.

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+// Execute the ListObjectsQuery, returning a list of object IDs
+func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjectsRequest) (*openfgapb.ListObjectsResponse, error) {
+	err := q.validateInput(ctx, req.StoreId, req.Type, req.AuthorizationModelId, req.Relation)
+	if err != nil {
+		return nil, err
+	}
+
+	allResults, _ := q.Meter.SyncInt64().Counter(
+		"openfga.listobjects.allresults",
+		instrument.WithDescription("Number of object IDs returned"),
```
elbuo8, line 44: `result_count` perhaps.

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+		TupleKey: &openfgapb.TupleKey{
+			Object:   object,
+			Relation: input.relation,
+			User:     input.user,
+		},
+		ContextualTuples: input.ctxTuples,
+	})
+	if err != nil {
+		// ignore the error. we don't want to abort everything if one of the checks failed.
+		q.Logger.ErrorWithContext(ctx, "Check errored: ", logger.Error(err))
```
elbuo8, line 269:
```suggestion
		q.Logger.ErrorWithContext(ctx, "check_error", logger.Error(err))
```

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/metric/instrument"
+	"go.opentelemetry.io/otel/metric/unit"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/sync/errgroup"
+)
+
+const (
+	MaximumConcurrentChecks = 100 // todo(jon-whit): make this configurable, but for now limit to 100 concurrent checks
```
jon-whit, line 22: I dont' know that this needs to be a publicly exported symbol. It's only relevant within the scope of this package. Let's change it to `maximumConcurrentCheck`.

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+
+// Execute the ListObjectsQuery, returning a list of object IDs
+func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjectsRequest) (*openfgapb.ListObjectsResponse, error) {
+	err := q.validateInput(ctx, req.StoreId, req.Type, req.AuthorizationModelId, req.Relation)
+	if err != nil {
+		return nil, err
+	}
+
+	allResults, _ := q.Meter.SyncInt64().Counter(
+		"openfga.listobjects.allresults",
```
jon-whit, line 43: The `openfga.` prefix should be redundant. We should be setting a service name prefix when we instantiate our metrics collector anyways.

Can we use camelCase on `listObjects`?

And instead of `allresults` can we name this something more like `objectCount` or something like that?

server/commands/list_objects.go
jon-whit: Agreed. If we're going to track a metric like this, let's count the requests that returned because of the `listObjects-deadline` being met. Though, what use do we have for the metric at this time? If we don't have a strong reason to introduce a metric to track some defined SLO, then I'd say we don't introduce a new metric at all. We can always add it later if we have a need for it then.

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+func (q *ListObjectsQuery) getUniqueObjects(ctx context.Context, storeID, targetObjectType string, ctxTuples *openfgapb.ContextualTupleKeys) (storage.ObjectIterator, error) {
+	iter, err := q.Datastore.ListObjectsByType(ctx, storage.ListObjectsFilter{
+		StoreID:    storeID,
+		ObjectType: targetObjectType,
+	})
+	if err != nil {
+		return nil, serverErrors.NewInternalError("", err)
+	}
+
+	return iter, nil
```
jon-whit, line 165: Shouldn't this method return an iterator that yields unique objects by combining the iterator returned from the database and the contextual tuples?

It would be nice if the `ObjectIterator` returned here iterates over both so that iteration over the two different relationship tuple streams doesn't have to happen independently in the caller.

server/commands/list_objects.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		errChan <- err
+		close(errChan)
+		return
+	}
+	defer iter.Stop()
+
+	// iterate over the contextual tuples
+	if input.ctxTuples != nil {
+		for _, t := range input.ctxTuples.TupleKeys {
```
jon-whit, line 207: See https://github.com/openfga/openfga/pull/145/files#r943606933

We abstract this away in the `getUniqueObjects` method, especially if we're going to pass the contextual tuples to that function.

server/errors/errors.go
```diff
@@ -108,10 +108,14 @@ func InvalidObjectFormat(tuple *openfgapb.TupleKey) error {
 	return status.Error(codes.Code(openfgapb.ErrorCode_invalid_object_format), fmt.Sprintf("Invalid object format for tuple '%s'", tuple.String()))
 }
 
-func UnknownRelation(relation string) error {
+func UnknownRelationWhenWritingAuthzModel(relation string) error {
```
jon-whit, line 111: Why are we creating two different errors for the same error? I think we should leave this as is but refactor the name to `ErrUnknownRelation`. Any RPC that accepts a `relation` field as an input argument will return this error if the relation they provide is undefined for the authorization model that the request is served against.

server/server.go
```diff
@@ -155,6 +157,67 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+	targetObjectType := req.GetType()
```
jon-whit, line 162: Mind adding a newline here?

storage/memory/memory.go
```diff
@@ -57,6 +57,21 @@ func (s *staticIterator) Next() (*openfgapb.Tuple, error) {
 
 func (s *staticIterator) Stop() {}
 
+type StaticObjectIterator struct {
```
jon-whit, line 60: Let's just move this into the `storage.go` file. It could be useful outside of the scope of the storage/memory package.

server/test/list_objects.go
```diff
@@ -0,0 +1,203 @@
...
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+const (
+	defaultListObjectsDeadline   = 5 * time.Second
+	defaultListObjectsMaxResults = 5
+)
+
+type listObjectsTestCase struct {
+	_name          string
```
jon-whit, line 29: No need for the `_` prefix here. `name` will do

server/test/list_objects.go
```diff
@@ -0,0 +1,203 @@
...
+	request        *openfgapb.ListObjectsRequest
+	expectedError  error
+	expectedResult []string //all the results. the server may return less
+}
+
+var tKAllAdminsRepo6 = &openfgapb.TupleKey{
```
jon-whit, line 35: Let's add a helper/utility function somewhere in `pkg` to streamline these instantiations. 

```
func tuplekey(object, relation, user string) *openfgav1.TupleKey {
	return &openfgav1.TupleKey{
		Object:   object,
		Relation: relation,
		User:     user,
	}
}
```

then these simply become, for example:

```
var tkAllAdminsRepo6 = tuplekey("repo:6", "admin", "*")
```

server/errors/errors.go
miparnisari: We also have `RelationNotFound` 😆  They are all the same in meaning but the error message is tailored to each specific context:

`RelationNotFound` -> checks, writes, reads, expands; a tuple is involved
`UnknownRelationWhenWritingAuthzModel` -> writeAuthModel; only relation involved
`UnknownRelationWhenListingObjects` -> listObjects; relation and type involved

server/commands/list_objects.go
miparnisari: I had thought of two ways we could do this:

1. Have the storage implementations append the contextual tuples to the iterator they return. This feels like a violation of SRP, the storage implementations shouldn't care about the contextual tuples
2. Add a method in the `ObjectIterator` interface to append the contextual tuples. This also feels wrong... iterator interfaces i've seen only allow you to return elements, not add to them

but I didn't like either, so that's why I kept it as is

storage/memory/memory.go
miparnisari: The static tuple iterator has existed for a while and it's in `memory.go`, too. I don't see why we should move this one if that one hasn't moved

server/commands/list_objects.go
miparnisari: > The `openfga.` prefix should be redundant.

I kept it the same as the other metrics 🤷‍♀️ the prefix makes it faster to see all metrics available

server/commands/list_objects.go
miparnisari: > what use do we have for the metric at this time

satisfy my own curiosity of how many requests terminate on time 😜

server/errors/errors.go
jon-whit: That's way too overloaded. Let's simplify here. If the relation was not found, regardless of the variation, then lets just return  `ErrRelationNotFound`. Bloating our public API error definitions isn't really buying us anything here, and it's making us define and map more "types" of errors.

server/commands/list_objects.go
jon-whit: I'm not debating whether a prefix is helpful or not. Prefixes are definitely valuable. I'm saying that the prefix should be included in all metrics under the metric collected, because we should be setting ServiceName at the collector level and not per metric.

All I'm saying is that this should be refactored so that we don't have to prefix every metric with the service name. That should be a top-level thing.

server/commands/list_objects.go
jon-whit: > satisfy my own curiosity
That's not a very strong reason to track a metric though 😄 . It has no relationship with any SLOs we publicly declare as being important for a developer to monitor OpenFGA health/performance.

If we have an initiative to track this because it's relevant to a broader monitoring objective, then let's do it. Otherwise let's leave it out until it has a stronger purpose.

server/commands/list_objects.go
jon-whit: Can I suggest that we create a `CombinedIterator` in the storage package that takes two iterators and combines them into a single iterator then? We could use this in multiple places. I may recommend we introduce generics a bit to help us make this more elegant. Something like:

```
package storage

type Iterator[T any] interface {
    Next() (T, error)
    Stop()
}

var tupleIterator Iterator[*openfgapb.Tuple]
var objectIterator Iterator[*openfgapb.Object] // see https://github.com/openfga/api/pull/24

func CombinedIterator[T any](iter1, iter2 Iterator[T]) Iterator[T] {
    ...
}

objIter := CombinedIterator[*openfgapb.Object](dbIter, ctxTupleIter)
tupleIter := CombinedIterator[*openfgapb.Tuple](dbIter, ctxTupleIter)
```

I don't think creating a combined iterator at the storage layer is breaking any SRP principles. It's just abstracting away the idea that the results from the database are augmented with the context passed in.

We now have code merging contextual info with database results in multiple places, and that's what I think could be improved.

If we want to do this work in a subsequent PR that's fine, but I think it should be done. If we're not going to do it here to make this code more streamlined and elegant, then let's at least add an issue for this work and follow up on it soon 👍 .

server/commands/list_objects.go
miparnisari: how are we going to find out if our deadline timeout is too low if we don't emit this metric? 

also metrics aren't always about SLOs, if i'm oncall i want to know this kind of things

server/commands/list_objects.go
miparnisari: I agree that it's a good idea, I hadn't thought of a combined iterator. Let's do it in a separate PR

elbuo8 (approved):

server/commands/list_objects.go
miparnisari: Ahhh ok. Separate PR :)

PR #146, opened 2022-08-03
fix: support Check on userset types of users

jon-whit:
## Description
This fixes Checks so that they support evaluation of relationships on usersets. For example, `Check(folder:folder1, viewer, group:engineering#member)`. Prior to this PR checks of this form would always return false, because we did not support checks on userset types.

## Review Checklist


craigpastro (commented):
I think contextual tuples may need some changes. And can we add tests to check as well please?

jon-whit:
> I think contextual tuples may need some changes. And can we add tests to check as well please?

@craigpastro it seems to work fine with ContextualTuples. I added two tests in `server/test/check` for these cases. Let me know what you think!

storage/postgres/postgres.go
```diff
@@ -236,8 +236,9 @@ func (p *Postgres) ReadUserTuple(ctx context.Context, store string, tupleKey *op
 	defer span.End()
 
 	objectType, objectID := tupleUtils.SplitObject(tupleKey.GetObject())
-	row := p.pool.QueryRow(ctx, `SELECT object_type, object_id, relation, _user FROM tuple WHERE store = $1 AND object_type = $2 AND object_id = $3 AND relation = $4 AND _user = $5 AND user_type = $6`,
-		store, objectType, objectID, tupleKey.GetRelation(), tupleKey.GetUser(), tupleUtils.User)
+	row := p.pool.QueryRow(ctx, `SELECT object_type, object_id, relation, _user FROM tuple WHERE store = $1 AND object_type = $2 AND object_id = $3 AND relation = $4 AND _user = $5`,
```
elbuo8, line 239: Usertype should still be provided but extracted from the tupleKey. This way we guarantee we land in the correct Index https://github.com/openfga/openfga/blob/b7d987f12a5e8f093a7eefa52bfddea349c9f966/assets/migrations/postgres/001_initialize_schema.sql#L14-L16

jon-whit: Good call. I've updated it to use the `user_type` column as a filter. Let me know how that looks.

elbuo8 (approved):

PR #147, opened 2022-08-04
feat: use query builder for Postgres queries

craigpastro:
Use a query builder for some of our more "complicated" Postgres queries.

## Review Checklist


storage/postgres/utils.go
```diff
@@ -75,17 +75,20 @@ func (t *tupleIterator) toArray(opts storage.PaginationOptions) ([]*openfgapb.Tu
...
+	tupleRecord, err := t.next()
+	if err != nil {
+		if errors.Is(err, storage.TupleIteratorDone) {
+			return res, nil, nil
+		}
+		return nil, nil, err
 	}
 
-	contToken, err := json.Marshal(newContToken(ulid, ""))
+	contToken, err := json.Marshal(newContToken(tupleRecord.ulid, ""))
```
craigpastro, line 91: Moved picking up the ulid to here so that all our queries look the same. On one of our queries we had `>` instead of `>=` because of this discrepancy. This makes it a bit clearer.

jon-whit (approved):

PR #149, opened 2022-08-06
fix: remove restrictions on read

craigpastro:
Remove artificial restrictions on read.

Are these necessary? I could see having one on objectType perhaps, but the other one not so much.

## Review Checklist


server/commands/read.go
```diff
@@ -76,15 +76,7 @@ func (q *ReadQuery) validateAndAuthenticateTupleset(ctx context.Context, store,
 	ctx, span := q.tracer.Start(ctx, "validateAndAuthenticateTupleset")
 	defer span.End()
 
-	objectType, objectID := tupleUtils.SplitObject(tupleKey.GetObject())
```
jon-whit, line 79: I like this change 👍 However, if we're going to remove these restrictions let's make sure we add tests to test that we can Read with different filter combinations. We just removed tests below, but we should also add tests to make sure this new behavior is tested.

craigpastro:
Let's do this when we update Read.

PR #164, opened 2022-08-12
docs: add `v0.2.0` release details to the CHANGELOG

jon-whit:
## Description
Add the `v0.2.0` release details to the changelog.

## Review Checklist


jon-whit:
Working on adding the appropriate documentation to openfga/openfga.dev for this new feature. I'll cross-reference with this.

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.2.0] - 2022-08-12
+### Added
+* [ListObjects API](https://openfga.dev/api/service#/Relationship%20Queries/ListObjects)
+
+  The ListObjects API provides a way to list all of the objects (of a particular type) that a user has a relationship with. It provides a solution to the [Search with Permissions (Option 3)](https://openfga.dev/docs/interacting/search-with-permissions#option-3-build-a-list-of-ids-then-search) use case for access-aware filtering on smaller object collections.
+
```
rhamzeh, line 14:
```suggestion
Note that that the current implementation has very high latency, and adds some configuration options to limit the response.

Warning: If the configuration options to limit the response are passed, the caller will only get partial data - pagination is not possible on this endpoint.

This change implements the [ListObjects RFC](https://github.com/openfga/rfcs/blob/main/20220714-listObjects-api.md)
```

rhamzeh: [TODO: Link to PR introducing the config options]
[TODO: Add a sample call like the one below]
```shell
$ curl -X POST ${FGA_SERVER_URI}/stores/${FGA_STORE_ID}/list-objects \
  -H 'Content-Type: application/json'
  -d '{"user":"anne","type":"document","relation":"can_view"}'

$ {"object_ids":["object_id_1","object_id_2"]}
```

rhamzeh (changes requested):

CHANGELOG.md
jon-whit: I've update the section based on this guidance. Let me know what you think about the new changes.

CHANGELOG.md
```diff
@@ -6,6 +6,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  -d '{"user":"anne", "type":"document", "relation":"can_view"}'
+
+  {"object_ids":["object_id_1", "object_id_2"]}
+  ```
+
+  The ListObjects API provides a way to list all of the objects (of a particular type) that a user has a relationship with. It provides a solution to the [Search with Permissions (Option 3)](https://openfga.dev/docs/interacting/search-with-permissions#option-3-build-a-list-of-ids-then-search) use case for access-aware filtering on smaller object collections. It implements the [ListObjects RFC](https://github.com/openfga/rfcs/blob/main/20220714-listObjects-api.md).
+
+  This addition brings with it two new server configuration options `--listObjects-deadline` and `--listObjects-max-results`. These configurations help protect the server from excessively long lived and large responses.
+
+  > ⚠️ If `--listObjects-deadline` or `--listObjects-max-results` are provided, the endpoint may only return a subset of the data. For example, if you limit the deadline to 1s and the results would take 2s to stream back, then you'll only get half of the results. If the response would contain 2 results but you limit the max results to 1, then you'll only get the one.
```
miparnisari, line 25: > For example, if you limit the deadline to 1s and the results would take 2s to stream back, then you'll only get half of the results.

This is a bit misleading.. you can also get zero, no?

jon-whit: I can change the sentence to "may only get back half of the results (if any results are available)"

How does that sound?

miparnisari (dismissed):

rhamzeh (dismissed):

CHANGELOG.md
matthewpereira: > [TODO: Add a sample call like the one below]

@rhamzeh how do you feel about limiting sample API calls and other info like that to the openfga.dev documentation site?

rhamzeh (approved):

CHANGELOG.md
rhamzeh: I personally think CHangelogs should include just enough context for a developer to be able to pick it up without having to go to the website, especially for:
- breaking changes
- bigger features such as this one

Some of the reasons why I like to include code snippets:
- I can easily grok what the API looks like
- If a change happens in the future, but a developer is using an earlier version, they do not have to wonder what the API looked like at that point in time, they can just quickly skim the changelog to find out

And in general I haven't heard people ever complain about having too much info in the Changelog :)

PR #172, opened 2022-08-16
refactor: make Iterator a generic and rework tuple iter composition

jon-whit:
## Description
This refactors storage iterators by leveraging generics and refactors the composition of ContextualTuple iteration so that it is composed with regular iteration yielded from the storage layer.

## Review Checklist


internal/middleware/http/handler_test.go
```diff
@@ -2,7 +2,7 @@ package http
 
 import (
 	"context"
-	"io/ioutil"
+	"io"
```
jon-whit, line 5: `iotuil` is being deprecated so this was throwing linter warnings

pkg/tuple/validation.go
```diff
@@ -58,39 +54,3 @@ func ValidateUser(tk *openfgapb.TupleKey) error {
 	}
 	return nil
 }
-
-// ValidateTuple returns whether a *openfgapb.TupleKey is valid
-func ValidateTuple(ctx context.Context, backend storage.TypeDefinitionReadBackend, store, authorizationModelID string, tk *openfgapb.TupleKey, dbCallsCounter utils.DBCallCounter) (*openfgapb.Userset, error) {
```
jon-whit, line 63: I moved these because of a cyclical import issues with tuple --> storage --> tuple. This validation that depends on a reference to a storage interface probably doesn't belong in public utility methods like this, so I moved them to a validation package in the server parent package. If we prefer them living somewhere other than where I put them we can move them. But they shouldn't exist here.

server/test/list_objects.go
```diff
@@ -59,18 +58,18 @@ func TestListObjects(t *testing.T, datastore storage.OpenFGADatastore) {
...
-				expectedResult: []string{"1", "2", "3", "4", "6"},
-				expectedError:  nil,
-			},
-			{
-				name:           "performs correct checks",
-				request:        newListObjectsRequest(store, "repo", "admin", "bob", modelID, nil),
-				expectedResult: []string{"2", "6"},
-				expectedError:  nil,
-			},
+			// {
```
jon-whit, line 61: Oops. Uncommenting these.

tools/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/openfga/openfga/tooling
 go 1.18
 
 require (
-	github.com/golang/mock v1.6.0
+	github.com/golang/mock v1.7.0-rc.1
```
jon-whit, line 6: This adds support to `mockgen` for generics, which was required for this PR because we now generic storage interface mocks that depend on generics.

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+			if errors.Is(err, IteratorDone) {
+				break
+			}
+			return nil, err
+		}
+
+		objectMap[tuple.ObjectKey(val)] = val
+	}
+
+	iter := &channelledStaticIterator[*openfgapb.Object]{
```
jon-whit, line 81: We use a channelledStaticIterator here instead of additional allocations for `[]*openfgapb.Object`. The overhead for a buffered channel that yields the values from the map is lower cost than building a new slice of the objects and then returning a static iterator over the new slice.

server/commands/list_objects.go
```diff
@@ -172,33 +172,26 @@ func (q *ListObjectsQuery) performChecks(timeoutCtx context.Context, input *Perf
...
-			}
-			t := t
-			checkFunction := func() error {
-				return q.internalCheck(timeoutCtx, t.Object, input, objectsFound, resultsChan)
-			}
+	iter2 := storage.NewTupleKeyObjectIterator(input.ctxTuples.GetTupleKeys())
 
-			g.Go(checkFunction)
-		}
+	iter, err := storage.UniqueObjectIterator(iter1, iter2)
```
elbuo8, line 183: Unsure this should be here. It makes all storage implementations keep a shadow copy instead those that cannot support it internally. I'm unsure if the case of contextual tuples makes it worth that.

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+
+type combinedIterator[T any] struct {
+	iter1, iter2 Iterator[T]
+}
+
+func (c *combinedIterator[T]) Next() (T, error) {
+	val, err := c.iter1.Next()
+	if err != nil {
+		if !errors.Is(err, IteratorDone) {
+			return val, err
```
elbuo8, line 100:
```suggestion
			return nil, err
```

elbuo8: Ditto the 2 below.

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+
+func (c *channelledStaticIterator[T]) Next() (T, error) {
+	if c.iteratorCh != nil && atomic.LoadUint64(&c.index) < c.size {
+		val := <-c.iteratorCh
+		atomic.AddUint64(&c.index, 1)
+		return val, nil
+	}
+
+	var val T
+	return val, IteratorDone
```
miparnisari, line 189: why not `return nil, IteratorDone`?

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+}
+
+type staticIterator[T any] struct {
+	items []T
 }
 
+func (s *staticIterator[T]) Next() (T, error) {
+	var val T
+	if len(s.items) == 0 {
+		return val, IteratorDone
```
miparnisari, line 204: same here?

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+// returns an IteratorDone error.
+type TupleKeyIterator = Iterator[*openfgapb.TupleKey]
+
+// UniqueObjectIterator takes two ObjectIterators and yields an ObjectIterator that returns only distinct objects with the
+// duplicates removed.
+func UniqueObjectIterator(iter1, iter2 ObjectIterator) (ObjectIterator, error) {
+	objectMap := map[string]*openfgapb.Object{}
+
+	for {
+		val, err := iter1.Next()
```
elbuo8, line 58: This will drain the whole iterator even in cases when it's not needed. While ListObjects it's the only current consumer, would rather not have this blow up later on.

storage/storage.go
miparnisari: ah never mind..

storage/storage.go
miparnisari: never mind

storage/storage.go
elbuo8: We are already doing more allocations than we need here by retaining the objects in the maps.

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+	var val T
+	return val, IteratorDone
+}
+
+func (c *channelledStaticIterator[T]) Stop() {
+	close(c.iteratorCh)
+	c.iteratorCh = nil
+}
+
+type staticIterator[T any] struct {
```
elbuo8, line 197: :+1:

storage/storage.go
```diff
@@ -31,22 +32,211 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+	}
+
+	return iter, nil
+}
+
+type combinedIterator[T any] struct {
+	iter1, iter2 Iterator[T]
+}
+
+func (c *combinedIterator[T]) Next() (T, error) {
```
elbuo8, line 96: 👍🏽 this def simplifies managing contextual tuples

storage/storage.go
jon-whit: The generic type `T` cannot take the value `nil`. So this is necessary.

storage/storage.go
jon-whit: That's necessary is it not? You cannot accumulate the unique objects without draining the two iterators in totality.

Since the `UniqueObjectIterator` is only consumed for ListObjects I'd be happy to move this as a utility/helper to the `commands/list_objects.go` source file, but I think it's ok here is it not? I can add doc to the godoc string that makes it more clear that this drains boths iterators, returning the unique objects between the two of them.

storage/storage.go
jon-whit: But those allocations are separate from the ones I was referring to with this optimization. The allocations to the `objectMap` are necessary to build the distinct set between the two iterators. I'm not sure how we'd eliminate those allocations unless we had a clever iterator that somehow new what it already yielded without storing anything in memory. 

I just benchmarked the difference between this and doing a copy of the values in the map into a slice and then returning a static iterator over the slice and the it's actually quite a bit faster. So the map value copy and then static iterator is better than this. I'll just drop the `channelledStaticIterator` then 👍

jon-whit: Turns out the memory copy to a slice has less overhead than the bandwidth of the channel send/receive. So it's better overall.

server/commands/list_objects.go
jon-whit: Sorry, I'm not sure I totally follow what you're saying? Don't we have iterators for every storage adapter that we support? If so, how come we can't compose them as a UniqueObjectIterator for distinct objects?

storage/memory/memory.go
```diff
@@ -48,7 +48,7 @@ func match(key *openfgapb.TupleKey, target *openfgapb.TupleKey) bool {
 
 func (s *staticIterator) Next() (*openfgapb.Tuple, error) {
```
jon-whit, line 49: I think we can get rid of this iterator in favor of the `storage.staticIterator[*openfgapb.Tuple]` instead. More specifically, `storage.NewStaticTupleIterator(tuples)` is effectively the same thing. I don't see us use the `continuationToken` in the iterator anyways. Can someone confirm for me?

storage/storage.go
miparnisari: > You cannot accumulate the unique objects without draining the two iterators in totality.

The iterator can keep a map of unique objects and on Next() you could check if the object you are about to return has been added to it. If not, you return it and add it to the map.

jon-whit: You still have to iterate over all of the objects though. I'm confused how this helps the issue? You can't return prematurely from the iterator because you haven't seen the objects the iterator may yield until you've iterated over them.

What am I missing here?

miparnisari: > You still have to iterate over all of the objects though. 

If the customer cancels the request midway, in your current code you would have unnecessarily iterated over everything

jon-whit: That's the point of `Stop()` is it not? If the request is cancelled midway then the caller calling this would have called `iter1.Stop()` and this would return.

elbuo8: `Stop` is never happening here. This will drain iter1 and iter2 the moment this third iterator is created. The iterators stop being lazy.

jon-whit: Alright @elbuo8 @miparnisari

I refactored this to lazily yield unique objects from the two iterators. I think this addresses your concern. Let me know what you think 👍 ! Sorry for the misunderstanding. It makes a lot more sense now what you guys were trying to point out. I just wasn't processing it the same way you were.

storage/memory/memory.go
elbuo8: >  I don't see us use the continuationToken in the iterator anyways.

The token is used _before_ creating the iterator. So yes, effectively you could replace it with the `NewStaticTupleIterator`. But let's do that in a subsequent PR.

storage/storage.go
```diff
@@ -31,22 +32,202 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
+}
+
+var _ ObjectIterator = (*staticObjectIterator)(nil)
+
+// NewStaticObjectIterator returns an ObjectIterator that iterates over the provided slice of objects.
+func NewStaticObjectIterator(objects []*openfgapb.Object) ObjectIterator {
+	return &staticObjectIterator{objects}
+}
+
+func (s *staticObjectIterator) Next() (*openfgapb.Object, error) {
```
elbuo8, line 217: Why can't this be implemented with the `staticIterator`?

pkg/id/id.go
```diff
@@ -38,6 +38,16 @@ func NewStringFromTime(time time.Time) (string, error) {
 	return id.value.String(), nil
 }
 
+// MustNewString returns an ID as a string and panics if one cannot be generated.
+func MustNewString() string {
```
elbuo8, line 42: Please don't add methods that `panic`. The benefit of this is saving a line in a test, not worth it.

storage/storage.go
jon-whit: It totally can. Good catch! That was remnant before the generic staticIterator that I added.

elbuo8 (approved):

PR #173, opened 2022-08-17
chore: normalize RelationNotFound errors

miparnisari:


server/errors/errors.go
```diff
@@ -74,8 +74,15 @@ func TypeNotFound(t string) error {
 	return status.Error(codes.Code(openfgapb.ErrorCode_type_not_found), fmt.Sprintf("Type '%s' not found", t))
 }
 
-func RelationNotFound(relation string, typeDefinition string, tuple *openfgapb.TupleKey) error {
-	return status.Error(codes.Code(openfgapb.ErrorCode_relation_not_found), fmt.Sprintf("Relation '%s' not found in type definition '%s' for tuple (%s)", relation, typeDefinition, tuple.String()))
+func RelationNotFound(relation string, typeName string, tuple *openfgapb.TupleKey) error {
```
jon-whit, line 77: You have the `tuple` so why do you need the other parameters?

relation == tuple.GetRelation()
and the typeName comes from the tuple object.

miparnisari: For ListObjects, I have `relation` and `typeName`.. if I were to make a `tuple` object, `tuple.user` wouldn't apply and I wouldn't have the full `tuple.object` (there would be no ID)

server/errors/errors.go
```diff
@@ -74,8 +74,15 @@ func TypeNotFound(t string) error {
...
+	msg := fmt.Sprintf("Authorization model contains an unknown relation '%s'", relation)
+	if tuple != nil {
+		msg = fmt.Sprintf("Unknown relation '%s' for type '%s' and tuple %s", relation, typeName, tuple.String())
+	}
+	if typeName != "" {
```
elbuo8, line 82: `else if`

elbuo8 (approved):

PR #181, opened 2022-08-19
fix: respect the value of ListObjectsMaxResults

miparnisari:
Pointers are hard eh? :)


jon-whit:
Tricky tricky. When we refactored this we introduce that tricky bugger.

Good catch 🚀 !

jon-whit (approved):

PR #184, opened 2022-08-22
fix: db_reads log when calling Check API

miparnisari:
When we call List Objects API, the logs say "check", because internally List Objects is calling Check.

```
2022-08-19T20:01:41.068Z	INFO	db_stats	{"method": "Check", "reads": 7, "writes": 0}
```

### DIFF
--- a/server/commands/check.go
+++ b/server/commands/check.go
@@ -29,16 +29,18 @@ type CheckQuery struct {
 	meter            metric.Meter
 	datastore        storage.OpenFGADatastore
 	resolveNodeLimit uint32
+	parentQuery      string
 }
 
 // NewCheckQuery creates a CheckQuery with specified `tupleBackend` and `typeDefinitionReadBackend` to use for storage
-func NewCheckQuery(datastore storage.OpenFGADatastore, t trace.Tracer, m metric.Meter, l logger.Logger, resolveNodeLimit uint32) *CheckQuery {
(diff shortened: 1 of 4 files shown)


server/commands/check.go
```diff
@@ -70,12 +72,12 @@ func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckReques
 
 	userset, err := query.getTypeDefinitionRelationUsersets(ctx, rc)
 	if err != nil {
-		utils.LogDBStats(ctx, query.logger, "Check", rc.metadata.GetReadCalls(), 0)
+		utils.LogDBStats(ctx, query.logger, query.parentQuery, rc.metadata.GetReadCalls(), 0)
```
elbuo8, line 75: The `method` data isn't available in the `ctx` by any chance?

elbuo8:
While the `method` is arguably wrong (it's Check itself that produced the metric) you can correlate that the parent of the Check came from ListObjects via the `requestID` field. 

With the iteration work being discussed for ListObjects I don't think there is the need to adjust this immediately.

jon-whit:
> While the `method` is arguably wrong (it's Check itself that produced the metric) you can correlate that the parent of the Check came from ListObjects via the `requestID` field.
> 
> With the iteration work being discussed for ListObjects I don't think there is the need to adjust this immediately.

I agree with this. To me, this is a matter of distributed tracing with correlations between ListObjects --> Check. The DB Read metrics are still tightly bound to the Check API, since Check is the delegate in this case anyways. In the future when ListObjects is implemented using reverse expansion, then we'll have read metrics unique to the ListObjects endpoint.

PR #186, opened 2022-08-24
chore: simplify health check

miparnisari:
Removing the health manager because we only have one service for now.


server/server.go
```diff
@@ -486,6 +481,34 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
...
+type AuthenticationOverride struct{}
+
+var _ grpc_auth.ServiceAuthFuncOverride = (*AuthenticationOverride)(nil)
+
+// AuthFuncOverride skips middleware/AuthFunc
+func (m AuthenticationOverride) AuthFuncOverride(ctx context.Context, fullMethodName string) (context.Context, error) {
+	return ctx, nil
+}
+
+type openfgHealthServer struct {
```
elbuo8, line 493:
```suggestion
type openfgaHealthServer struct {
```

server/server.go
```diff
@@ -486,6 +481,34 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
...
+func (o openfgHealthServer) Check(ctx context.Context, request *healthv1pb.HealthCheckRequest) (*healthv1pb.HealthCheckResponse, error) {
+	_, err := o.Server.IsReady(ctx)
+	if err != nil {
+		return &healthv1pb.HealthCheckResponse{Status: healthv1pb.HealthCheckResponse_NOT_SERVING}, err
+	} else {
+		return &healthv1pb.HealthCheckResponse{Status: healthv1pb.HealthCheckResponse_SERVING}, nil
+	}
+}
+
+func (o openfgHealthServer) Watch(request *healthv1pb.HealthCheckRequest, server healthv1pb.Health_WatchServer) error {
```
elbuo8, line 508: ?

miparnisari: Watch is for a streaming response. https://pkg.go.dev/google.golang.org/grpc/health/grpc_health_v1@v1.48.0#pkg-variables

I don't think we need this.

elbuo8 (approved):

PR #187, opened 2022-08-25
fix: read env vars for migrate command

### LINKED ISSUE #163: Migrate does not respect the same configuration methods as openfga
## To Reproduce
```shell
docker run --env OPENFGA_DATASTORE_ENGINE=postgres --env OPENFGA_DATASTORE_URI=postgres://postgres:postgres@localhost:5432/openfga openfga/openfga:v0.1.7 migrate
```

## Expected Result(s)
- An error connecting to postgres if postgres is not running or the uri is invalid
- A successful migration

## Actual Result
- Complaint about datastore params not being passed
Screen Shot 2022-08-12 at 3 22 49 PM

craigpastro:
## Description

Resolves #163.

This is the opposite of DRY. However, I think it is okay here. Since version does not belong in the app config, I don't think the migrate command should be reading values from the app `config.yaml`. Hence, I didn't refactor the flags from run to also apply here.

Also OPENFGA_VERSION is not a good name for the migrate version env var. But it is the appropriate name for the flag. So... 🤷 

## Review Checklist


jon-whit (approved):
Looks good to me 👍

rhamzeh:
> Since version does not belong in the app config, I don't think the migrate command should be reading values from the app config.yaml

Can we at least read the other common variables from config? It would be very confusing if something was modified there but not here.

craigpastro:
> Can we at least read the other common variables from config? It would be very confusing if something was modified there but not here.

I can't imagine what other config variables would apply here. We could separate the datastore flags and apply these both to run and migrate. This will prevent them changing independently.

> These are very vague terms, what are they referencing? OpenFGA version? The tag? The commit hash?

My thought with the cli interface is that each command has its own flags/env vars and they are not related to each other. I wasn't expected flags that work for run to work for migrate unless explicitly mentioned that they do so. In that case, I think version makes sense here because it is related only to the migrate command. I am curious what other people think.

That said, I am not opposed to using `db-schema-version` if you think it adds clarity. It just conflicts with `datastore` which we are using everywhere else, and `datastore-schema-version` is getting pretty long... but it is also okay.

PR #188, opened 2022-08-25
feat: add backoff when connecting to Postgres

craigpastro:
## Description

Don't fail if we don't manage to connect to Postgres right away. Try again. This means that we can start OpenFGA and then start Postgres. It is a bit kinder to developers.

I read about it here and thought it was a nice idea. https://medium.com/@kelseyhightower/12-fractured-apps-1080c73d481c

Here is what happens when I start Postgres after OpenFGA:

Screen Shot 2022-08-25 at 15 12 04

## Review Checklist


pkg/cmd/run.go
```diff
@@ -40,9 +40,6 @@ func NewRunCommand() *cobra.Command {
 }
 
 func run(_ *cobra.Command, _ []string) {
-	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
craigpastro, line 43: This needed to move after the initialization of the database. Otherwise, I was unable to ctrl-c out of the "waiting for Postgres" loop until the backoff expired.

storage/postgres/postgres.go
```diff
@@ -97,6 +85,22 @@ func NewPostgresDatastore(uri string, opts ...PostgresOption) (*Postgres, error)
...
+	var pool *pgxpool.Pool
+	err := backoff.Retry(func() error {
+		var err error
+		pool, err = pgxpool.Connect(context.Background(), uri)
+		if err != nil {
+			p.logger.Info("waiting for Postgres")
+			return err
+		}
+		return nil
+	}, backoff.NewExponentialBackOff())
```
jon-whit, line 97: The default here uses an extremely long MaxElapsedTime setting (like 13min or something like that). We should create an exponential backoff policy with a tighter max elapsed time so that our readiness/liveness checks up front fail quicker if we can't connect to the database in time.

craigpastro: Good point. Thanks! How does five minutes sound?

jon-whit: I was thinking more like 30s - 1min max. Readiness checks in Cloud environments like Kubernetes are usually on the order of no more than a 2min timeout window. Setting a little tighter timeout here will make readiness a little more lively, which will help more quickly identify deployments that are wired up to a database that is unreachable.

craigpastro: Okay. Let's try 1 minute ;)

storage/postgres/postgres.go
```diff
@@ -97,6 +85,26 @@ func NewPostgresDatastore(uri string, opts ...PostgresOption) (*Postgres, error)
...
 
+	policy := backoff.NewExponentialBackOff()
+	policy.MaxElapsedTime = 1 * time.Minute
+	var pool *pgxpool.Pool
+	attempt := 1
+	err := backoff.Retry(func() error {
+		var err error
+		pool, err = pgxpool.Connect(context.Background(), uri)
+		if err != nil {
+			p.logger.Info(fmt.Sprintf("failed to initialize Postgres connection: (attempt #%d)", attempt))
```
elbuo8, line 96: Use fields to indicate count. Keep the message brief as it was.

storage/postgres/postgres.go
```diff
@@ -97,6 +85,26 @@ func NewPostgresDatastore(uri string, opts ...PostgresOption) (*Postgres, error)
 		p.maxTypesInTypeDefinition = defaultMaxTypesInDefinition
 	}
 
+	policy := backoff.NewExponentialBackOff()
+	policy.MaxElapsedTime = 1 * time.Minute
+	var pool *pgxpool.Pool
+	attempt := 1
+	err := backoff.Retry(func() error {
```
elbuo8, line 92: Should we stop attempting at some point?

storage/postgres/postgres.go
```diff
@@ -62,20 +63,7 @@ func WithTracer(t trace.Tracer) PostgresOption {
 }
 
 func NewPostgresDatastore(uri string, opts ...PostgresOption) (*Postgres, error) {
-
-	pgxConfig, err := pgxpool.ParseConfig(uri)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse config from uri: %v", err)
-	}
-
-	dbpool, err := pgxpool.ConnectConfig(context.Background(), pgxConfig)
```
elbuo8, line 71: Why not continue using this method to connect?

storage/postgres/postgres.go
craigpastro: 👍 Thanks! Updated.

storage/postgres/postgres.go
craigpastro: `Connect` does exactly these 5 lines under the hood:
```
func Connect(ctx context.Context, connString string) (*Pool, error) {
	config, err := ParseConfig(connString)
	if err != nil {
		return nil, err
	}

	return ConnectConfig(ctx, config)
}
```

elbuo8 (approved):

PR #192, opened 2022-08-26
fix: make service tests less flaky

craigpastro:
## Description

Make service tests less flaky by:
- Using retryablehttp client everywhere
- Turning off HTTP server when not needed
- Hit the actually HTTP health endpoint when the test requires it

## Review Checklist


pkg/cmd/service/service_test.go
```diff
@@ -612,7 +614,8 @@ func TestHTTPServingTLS(t *testing.T) {
 }
 
 func TestGRPCServingTLS(t *testing.T) {
-	logger := logger.NewNoopLogger()
+	logger, err := logger.NewTextLogger()
```
elbuo8, line 617: ?

pkg/cmd/service/service_test.go
```diff
@@ -562,6 +562,9 @@ func TestHTTPServingTLS(t *testing.T) {
 
 		ensureServiceUp(t, nil)
 
+		_, err = retryablehttp.NewClient().Get("http://localhost:8080/healthz")
```
elbuo8, line 565: This should be happening within `ensureServiceUp` otherwise you'd be copy pasting everywhere.

pkg/cmd/service/service_test.go
```diff
@@ -717,7 +722,7 @@ func TestHTTPServerEnabled(t *testing.T) {
 
 	ensureServiceUp(t, nil)
 
-	resp, err := http.Get("http://localhost:8080/healthz")
+	resp, err := retryablehttp.NewClient().Get("http://localhost:8080/healthz")
```
elbuo8, line 725: Can we move this within `ensureServiceUp`?

pkg/cmd/service/service_test.go
```diff
@@ -671,8 +676,8 @@ func TestGRPCServingTLS(t *testing.T) {
 		ensureServiceUp(t, creds)
 
 		cancel()
-		require.NoError(t, g.Wait())
 		require.NoError(t, service.Close(ctx))
+		require.NoError(t, g.Wait())
```
elbuo8, line 680: 👍🏽 good catch.

elbuo8: Come to think of it, why are we `canceling` if we are actually gracefully closing right after? Seems a bit of a mix up there.

jon-whit (changes requested):
Let's talk more about these before merging. I think the number 1 issue with flaky tests is largely due to tests overlapping on the same port. At least this is true in my experience with this project so far. In `cmd/service/service_test.go` we spin up servers using the same port mappings. We should be auto assigning a port using an address such as `:0` or `127.0.0.1:` (https://pkg.go.dev/net#Listen).

I think we should refactor our service tests to auto assign a port dynamically and run the server using that port. This will require some refactoring a bit to make that composition neat and tidy in our tests, but I think it is well worth it.

elbuo8:
@jon-whit agreed. Particularly if want them to run concurrently. But don't think these changes should be blocked until that lands.

pkg/cmd/service/service_test.go
craigpastro: Ah, was using for debugging. Reverted. Thanks.

pkg/cmd/service/service_test.go
craigpastro: Moved to `ensureServiceUp`. Thanks!

pkg/cmd/service/service_test.go
craigpastro: Done. Thanks!

pkg/cmd/service/service_test.go
craigpastro: `canlceling` is the thing that is triggering the graceful closure. Server.Close() closes database and auth connections only. It's a bit ... odd maybe?

pkg/cmd/service/service_test.go
```diff
@@ -48,7 +48,7 @@ const (
 	openFGAGRPCAddr   = "localhost:8081"
 )
 
-func ensureServiceUp(t *testing.T, transportCredentials credentials.TransportCredentials) {
+func ensureServiceUp(t *testing.T, transportCredentials credentials.TransportCredentials, useHTTPToo bool) {
```
elbuo8, line 51: I'd always spin it up so we don't need to have the context. With the introduction of the healthcheck we ensure we don't run into a race condition. Plus the upcoming work for having random ports assigned should help too.

pkg/cmd/service/service_test.go
```diff
@@ -93,6 +93,11 @@ func ensureServiceUp(t *testing.T, transportCredentials credentials.TransportCre
 		return nil
 	}, policy)
 	require.NoError(t, err)
+
+	if useHTTPToo {
+		_, err = retryablehttp.NewClient().Get("http://localhost:8080/healthz")
```
elbuo8, line 98: `openFGAServerAddr` has this config rather than hardcoding. That way when this gets refactored we don't have to look it up.

Might be worth also putting it in the `backoff` function after the GRPC healthcheck.

pkg/cmd/service/service_test.go
```diff
@@ -93,6 +93,11 @@ func ensureServiceUp(t *testing.T, transportCredentials credentials.TransportCre
 		return nil
 	}, policy)
 	require.NoError(t, err)
+
+	if useHTTPToo {
```
miparnisari, line 97:
```suggestion
	if checkHttp {
```

?

pkg/cmd/service/service_test.go
jon-whit: Wait, why do we have the HTTP option instead of just using the single RPC health check? I feel like this is extra code that isn't needed.

elbuo8: The RPC gateway might boot up while the HTTP listener might not. We want to ensure both are up.

craigpastro: What Yamil said. I suppose this would be fixed with retryable client everywhere.

Another problem however, was some tests were failing because we could cancel the context when the HTTP server was in the middle of starting up resulting in an error. This is probably fixed by disabling the HTTP server in such tests.

In any case, I think this adds robustness.

pkg/cmd/service/service_test.go
```diff
@@ -34,21 +34,27 @@ import (
...
 		panic(err)
 	}
-}
 
-const (
-	openFGAServerAddr = "http://localhost:8080"
-	openFGAGRPCAddr   = "localhost:8081"
-)
+	config := DefaultConfig()
+	openFGAServerAddr = fmt.Sprintf("http://localhost%s", config.HTTP.Addr)
```
jon-whit, line 51: `config.HTTP.Addr` is the addr. For example if I set `config.HTTP.Addr: "127.0.0.1:8082"` then the HTTP server will only bind on host address 127.0.0.1 and port 8082. 

So the `localhost` component here isn't quite right.

jon-whit: 🤔 but you're using the default config above which will be like `:8080`. This just feels weird. What's a better way to do it? I'm thinking..

pkg/cmd/service/service_test.go
```diff
@@ -208,32 +219,24 @@ type authTest struct {
 }
 
 func TestBuildServiceWithNoAuth(t *testing.T) {
-	config, err := GetServiceConfig()
-	require.NoError(t, err)
+	config := DefaultConfig()
```
jon-whit, line 222: The reason we used `GetServiceConfig` was that it is the config entrypoint for the service. It is supposed to call `DefaultConfig` and merge other user provided configs ontop of it. 

I think that we actually get better test coverage and test fidelity if we leave this as it was before. It also gives us the apportunity to use `os.SetEnv(...)` to test that our env variables are wired up in our service properly and that `GetServiceConfig` respects it. So I think I'd prefer to leave it even though it is one extra line of code for the `require.NoError(...)`.

pkg/cmd/service/service_test.go
craigpastro: Use netip.AddrPort and make it explicit everywhere? I can do that in a following PR.

jon-whit: I guess I'll need to do something like this with the changes I'm making for the random port. I'll include that in that PR.

pkg/cmd/service/service_test.go
```diff
@@ -93,6 +99,11 @@ func ensureServiceUp(t *testing.T, transportCredentials credentials.TransportCre
 		return nil
 	}, policy)
 	require.NoError(t, err)
+
+	if useHTTPToo {
```
jon-whit, line 103: Can we rename this to `httpHealthCheck`?

jon-whit (dismissed):

pkg/cmd/service/service_test.go
craigpastro: Done. Thanks!

pkg/cmd/service/service_test.go
craigpastro: Renamed. Thank you!

pkg/cmd/service/service_test.go
craigpastro: It makes sense. I thought that if we need it in a particular test we could use it, but I see what you are saying too. Reverted. Thanks!

jon-whit (approved):

PR #193, opened 2022-08-27
chore: log internal errors with more details

miparnisari:
Log all errors, and especially log internal server errors.


server/server.go
```diff
@@ -137,11 +137,11 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
 		defaultServeMuxOpts: []runtime.ServeMuxOption{
 			runtime.WithForwardResponseOption(httpmiddleware.HTTPResponseModifier),
 			runtime.WithErrorHandler(func(c context.Context, sr *runtime.ServeMux, mm runtime.Marshaler, w http.ResponseWriter, r *http.Request, e error) {
-				actualCode := serverErrors.ConvertToEncodedErrorCode(status.Convert(e))
-				if serverErrors.IsValidEncodedError(actualCode) {
-					dependencies.Logger.ErrorWithContext(c, "grpc error", logger.Error(e), logger.String("request_url", r.URL.String()))
+				dependencies.Logger.ErrorWithContext(c, "grpc error", logger.Error(e), logger.String("request_url", r.URL.String()))
```
craigpastro, line 140: Would this make filtering a bit nicer?
```suggestion
				dependencies.Logger.ErrorWithContext(c, "grpc_error", logger.Error(e), logger.String("request_url", r.URL.String()))
```

server/server.go
```diff
@@ -137,11 +137,11 @@ func New(dependencies *Dependencies, config *Config) (*Server, error) {
...
+				if internalError, ok := e.(serverErrors.InternalError); ok {
+					dependencies.Logger.ErrorWithContext(c, "internal error", logger.Error(internalError.Internal()))
```
craigpastro, line 142: Similarly here?
```suggestion
					dependencies.Logger.ErrorWithContext(c, "internal_error", logger.Error(internalError.Internal()))
```
Or should this also be a `grpc_error` so we can filter all errors and have an extra field like `"error_type": "internal"` or something?

elbuo8: I'm actually wondering if we can perform a single log. Building a `fields` array that gets extended if it's an `InternalError` and with a single log you can see everything.

elbuo8 (approved):

craigpastro (approved):
LGTM

PR #194, opened 2022-08-27
fix: create store doesn't return times

craigpastro:
## Description

In Postgres, when creating a store we don't return the `created_at` and `updated_at` fields, so the response looks like:
```
{
    "id": "01GBECXZ9HX7ASTK57GRFAN0JY",
    "name": "ppop",
    "created_at": null,
    "updated_at": null,
}
```
This PR ensures we do return these fields.

Alternatively, we can just leave them null and perhaps exclude null values from the response? WDYT?

## Review Checklist


craigpastro:
Unit tests always fail! Need to merge https://github.com/openfga/openfga/pull/192 first, which hopefully helps.

storage/postgres/postgres.go
```diff
@@ -467,7 +467,12 @@ func (p *Postgres) CreateStore(ctx context.Context, store *openfgapb.Store) (*op
 		return nil, handlePostgresError(err)
 	}
 
-	return store, nil
+	createdStore, err := p.GetStore(ctx, store.Id)
+	if err != nil {
+		return nil, err
+	}
+
+	return createdStore, nil
```
elbuo8, line 475: Could we add a test to prevent this level of inconsistency in the future?

storage/postgres/postgres.go
```diff
@@ -467,7 +467,12 @@ func (p *Postgres) CreateStore(ctx context.Context, store *openfgapb.Store) (*op
 		return nil, handlePostgresError(err)
 	}
 
-	return store, nil
+	createdStore, err := p.GetStore(ctx, store.Id)
```
jon-whit, line 470: My preference here is to not call a storage method inside another storage method. I don't think adding the timestamp metadata in the response of `CreateStore` is largely that helpful anyways.

I'm curious if we're ok just returning the `id` and `name` and if the user wants other metadata about the store after it's created, then they can call the `GetStore` RPC. I think simpler is better here.

jon-whit (changes requested):
I left a comment about either using the `RETURNING` keyword or simplifying the response behavior of the `CreateStore` API altogether. Let me know what you think.

storage/postgres/postgres.go
elbuo8: > the only problem I see with this approach is that other storage implementations aren't going to have this ability (such as dynamodb).

It would be the responsibility of the storage implementation to match the spec. 

> That being said, if we're going to return calculated metadata from the storage layer, let's use RETURNING here instead.

Agreed, specially since its PSQL's responsibility to generate it and there is no need for the roundtrip of operations.

storage/postgres/postgres.go
craigpastro: Test added. Thank you!

jon-whit (changes requested):
Left one last comment. Other than that it looks great!

storage/postgres/postgres.go
jon-whit: What I don't like about adding this metadata to the storage layer is that it fundamentally will introduce clock skew issues into the system.

To implement this storage method for Dynamodb we'll have to do something like

```
var store *openfgapb.Store
store.CreatedAt = time.Now()
dynamodb.PutItem(store)

return store
```
The `time.Now` piece isn't a serially ordered timestamps as viewed from the txns against the database. In postgres you can use the [pg_xact_commit_timestamp(xid)](https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-COMMIT-TIMESTAMP) to get an accurate txn commit timestamp, but not all databases have that functionality.

This approach just adds complexities that aren't really warranted IMO.

server/test/create_store.go
```diff
@@ -4,68 +4,43 @@ import (
...
 
-			if actualResponse == nil {
-				t.Error("Expected non nil response, got nil")
-			} else {
-				if diff := cmp.Diff(actualResponse, test.response, ignoreStateOpts, ignoreStoreFields, cmpopts.EquateEmpty()); diff != "" {
-					t.Errorf("[%s] store mismatch (-got +want):\n%s", test._name, diff)
-				}
-			}
+			require.Equal(t, test.request.Name, resp.Name)
+			require.NotEmpty(t, resp.Id)
```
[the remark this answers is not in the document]
craigpastro: I thought about that, and then forgot about it. I think I was trying to do too many things at once. Thanks for the review!

jon-whit (approved):

storage/postgres/postgres.go
craigpastro: After giving it some thought, I agree. It would be nice to simplify our apis at some point.

PR #195, opened 2022-08-27
chore: default to Postgres in docker compose

craigpastro:
## Description

It is a bit confusing that we have Postgres in the docker compose file, but don't use it when we `docker compose up`. So let's use it. With these changes we start a Postgres container, start OpenFGA using that container, and perform the migration.

## Review Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -17,11 +17,22 @@ services:
       retries: 5
 
   openfga:
+    depends_on:
+      - postgres
     image: openfga/openfga:latest
     container_name: openfga
-    command: run
+    command: run --datastore-engine postgres --datastore-uri 'postgres://postgres:password@postgres:5432/postgres?sslmode=disable'
     networks:
       - default
     ports:
       - "8080:8080"
       - "3000:3000"
+
+  migrate:
+    depends_on:
+      - openfga
+    image: openfga/openfga:latest
+    container_name: migrate
+    command: migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@postgres:5432/postgres?sslmode=disable'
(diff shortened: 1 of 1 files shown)


jon-whit (approved):

docker-compose.yaml
```diff
@@ -17,11 +17,22 @@ services:
...
       - default
     ports:
       - "8080:8080"
-      - "3000:3000"
\ No newline at end of file
+      - "3000:3000"
+
+  migrate:
+    depends_on:
+      - openfga
```
elbuo8, line 33: This order of operations is weird to me. The DB should be migrated first and then boot up the service. I know it happens one right after the other right now, but as this compose file matures, health checks get added and other systems the sequence should be proper.

craigpastro: Yeah. The reason why this order is that immediately after the postgres container starts up you can't actually always connect, so migrate would fail. Starting openfga gives it enough time I guess. I am not actually sure that it guarantees that we can connect now that I think about it. Maybe we need to add the healthcheck to the openfga part as well?

elbuo8: The solution would be to implement health check on the psql container so migrate runs when it's ready. Otherwise it's a race condition that for now it's passing :)

As for openfga having a healthcheck, yes, that would be good too.

craigpastro: What may be good is adding a backoff to migrate as well to make sure we can connect. Does that seem reasonable?

elbuo8: A backoff is just left for luck. Having the health check solves that.
https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

PR #197, opened 2022-08-28
fix: validate type and relation in user values that are usersets

### LINKED ISSUE #189: bug: validate type and relations in userset values
Given the following authorization model:

```
type document
  relations
    define viewer as self
```

I can write the following relationship tuple `Write("document:doc1", "viewer", "group:engineering#member")`.

I shouldn't be able to do this because the `group` type does not exist in the authorization model.

jon-whit:
## Description
This fixes a bug that allows clients to write tuples with a user value that references an undefined type or relation on a type. See #189.

## References
Fixes #189 

## Review Checklist


server/test/check.go
```diff
@@ -145,18 +145,8 @@ var checkQueryTests = []checkQueryTest{
...
-							Relation: "writer",
-						},
-					}},
-				"writer": {
-					Userset: &openfgapb.Userset_ComputedUserset{
-						ComputedUserset: &openfgapb.ObjectRelation{
-							Relation: "reader",
-						},
-					}},
+				"reader": ComputedUserset("", "writer"),
```
jon-whit, line 148: Refactored these so they are easier to read and take up way less space. The helpers are in `server/test/helpers.go`

server/test/check.go
```diff
@@ -1251,6 +1012,42 @@ var checkQueryTests = []checkQueryTest{
 			Allowed: true,
 		},
 	},
+	{
```
jon-whit, line 1015: These are the tests for the new functionality that validates the type and relation if the user is a userset.

server/test/check.go
miparnisari: i'm not against these changes but they belong to a different PR :P

jon-whit: There are a lot more of these changes to make, and I plan on saving those for a future PR. This just started that work and made it a little easier for me to write and read the new tests I was writing, so I went ahead and did it just for `check.go`.

server/test/write.go
```diff
@@ -1003,9 +900,46 @@ var writeCommandTests = []writeCommandTest{
 				tuple.NewTupleKey("document:budget", "reader", "group:abc#member"),
 			}},
 		},
-		err: serverErrors.InvalidTuple("User 'group:abc#member' is not allowed to have relation reader with document:budget",
-			tuple.NewTupleKey("document:budget", "reader", "group:abc#member"),
-		),
+		err: serverErrors.RelationNotFound("member", "group", tuple.NewTupleKey("document:budget", "reader", "group:abc#member")),
+	},
+	{
```
miparnisari, line 905: The test existed - i just updated the error

server/test/write.go
```diff
@@ -1003,9 +900,46 @@ var writeCommandTests = []writeCommandTest{
...
+					},
+				},
+			},
+		},
+		request: &openfgapb.WriteRequest{
+			Writes: &openfgapb.TupleKeys{TupleKeys: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("document:budget", "reader", "notallowed:abc#member"),
+			}},
+		},
+		err: serverErrors.TypeNotFound("notallowed"),
```
miparnisari, line 942: New test

server/test/write.go
```diff
@@ -1096,6 +1030,9 @@ var writeCommandTests = []writeCommandTest{
 				},
 				{
 					Type: "group",
+					Relations: map[string]*openfgapb.Userset{
+						"member": typesystem.This(),
+					},
```
miparnisari, line 1035: The new change broke this test because the model was not properly defined 😅

server/validation/validation.go
```diff
@@ -43,5 +43,27 @@ func ValidateObjectsRelations(ctx context.Context, backend storage.TypeDefinitio
 	if !ok {
 		return nil, &tuple.RelationNotFoundError{Relation: t.GetRelation(), TypeName: ns.GetType(), TupleKey: t}
 	}
+
+	// if the user is a userset, validate the type and relation on it as well
+	user := t.GetUser()
```
craigpastro, line 48: Should this part be in the `ValidateTuple` function as this function is called `ValidateObjectsRelations`?

Also, what if the user is just an object, like `folder:X`, and not an ObjectRelation? Should we be validating that the type `folder` exists? Or is this done elsewhere?

miparnisari: > Should this part be in the ValidateTuple function as this function is called ValidateObjectsRelations?

Good point! Yes! `ValidateObjectsRelations` is being called in Expand, which doesn't take a `user` param at all, so it's pointless to call that code. I moved it to its own function now.

> what if the user is just an object, like folder:X, and not an ObjectRelation? Should we be validating that the type folder exists? Or is this done elsewhere?

If the model is 1.1, this is done in https://github.com/openfga/openfga/blob/776e80505e8d184b2286acc8268d8d74f36a9984/server/commands/write.go#L161

miparnisari: We really need https://github.com/openfga/openfga/pull/341. Now i don't know whether https://github.com/openfga/openfga/issues/189 applies to 1.0 models, 1.1 models, or both...

craigpastro: Ah, yes, good point. I forgot that we allow user_ids to be anything in the 1.0 world. oh my.

craigpastro (approved):
LGTM

server/test/write.go
```diff
@@ -795,54 +724,72 @@ var writeCommandTests = []writeCommandTest{
...
-					Relation: "owner",
-					User:     "github|yenkel@openfga",
+					Type: "document",
+					Relations: map[string]*openfgapb.Userset{
+						"viewer": typesystem.This(),
+					},
 				},
+			},
+		},
+		tuples: []*openfgapb.TupleKey{},
```
jon-whit, line 759: Omit since it's not needed for this test case.

server/test/write.go
```diff
@@ -795,54 +724,72 @@ var writeCommandTests = []writeCommandTest{
...
-					User:     "team:openfga/platform#member",
+					Type: "document",
+					Relations: map[string]*openfgapb.Userset{
+						"viewer": typesystem.This(),
+					},
 				},
-			}},
+			},
+		},
+		tuples: []*openfgapb.TupleKey{},
```
jon-whit, line 784: Omit since it's not needed for this test case.

server/test/write.go
```diff
@@ -967,7 +914,7 @@ var writeCommandTests = []writeCommandTest{
 		),
 	},
 	{
-		_name: "Write fails if user field is a userset that is not allowed by the authorization model (which only allows group:...)",
+		_name: "1.1_Execute_fails_if_relation_in_userset_value_was_not_found",
```
jon-whit, line 917: Mind changing the `document#reader` rewrite to `typesystem.This()` below?

server/test/write.go
```diff
@@ -1003,9 +950,46 @@ var writeCommandTests = []writeCommandTest{
...
+				{
+					Type: "user",
+				},
+				{
+					Type: "group",
+				},
+				{
+					Type: "document",
+					Relations: map[string]*openfgapb.Userset{
+						"reader": {Userset: &openfgapb.Userset_This{}},
```
jon-whit, line 971: `typesystem.This()` please

server/test/write.go
```diff
@@ -1459,10 +1450,9 @@ func TestWriteCommand(t *testing.T, datastore storage.OpenFGADatastore) {
 
 			require.ErrorIs(gotErr, test.err)
 
+			// todo: validate the response bodies are equivalent
```
jon-whit, line 1453: Add an issue for this please.

server/validation/validation.go
```diff
@@ -43,5 +46,30 @@ func ValidateObjectsRelations(ctx context.Context, backend storage.TypeDefinitio
 	if !ok {
 		return nil, &tuple.RelationNotFoundError{Relation: t.GetRelation(), TypeName: ns.GetType(), TupleKey: t}
 	}
+
 	return userset, nil
 }
+
+// ValidateUserIfItsAUserset returns an error if the t.User is a userset and its type/relation don't exist in the modelID
+func ValidateUserIfItsAUserset(ctx context.Context, backend storage.TypeDefinitionReadBackend, store string, modelID string, t *openfgapb.TupleKey) error {
```
jon-whit, line 54: Don't take a `TypeDefinitionReadBackend`. Let's pass either the model OR the typesystem.

server/validation/validation.go
```diff
@@ -15,6 +15,9 @@ func ValidateTuple(ctx context.Context, backend storage.TypeDefinitionReadBacken
 	if err := tuple.ValidateUser(tk); err != nil {
 		return nil, err
 	}
+	if err := ValidateUserIfItsAUserset(ctx, backend, store, authorizationModelID, tk); err != nil {
```
jon-whit, line 18: Please don't delegate this obscurely like it is doing right now. It's not obvious the top level path this is taking whatsoever. Instead, I recommend we make this top-level code path more explicit with

```
if tuple.IsObjectRelation(tk.GetUser()) {
   ... // validation logic
}
```

this way the code is more inline and obvious what validations are taking place in the event the user field is an ObjectRelation (e.g. userset).

server/validation/validation.go
miparnisari: I could do this refactor, so that both `ValidateObjectsRelations` and `ValidateUserIfItsAUserset` take in a typesystem, but aren't you going to change that code anyway after https://github.com/openfga/openfga/pull/341 ?

server/test/write.go
miparnisari: fixed it inline 😆

server/validation/validation.go
jon-whit: Yeah, I guess so.

miparnisari (approved):

PR #198, opened 2022-08-29
chore: minor Health server touch ups

jon-whit:
## Description
This changes our Health check implementation slightly so that it is inline with the grpc Health Check protocol.

> The server should register all the services manually and set the individual status, including an empty service name and its status. For each request received, if the service name can be found in the registry, a response must be sent back with an OK status and the status field should be set to SERVING or NOT_SERVING accordingly. If the service name is not registered, the server returns a NOT_FOUND GRPC status.
> 
> The server should use an empty string as the key for server's overall health status, so that a client not interested in a specific service can query the server's status with an empty request. 

More specifically, we only return a status if the request is made with a service field that is the empty string "" or the service name matching the OpenFGA service (`openfga.v1.OpenFGAService`).

## References
https://github.com/grpc/grpc/blob/master/doc/health-checking.md

## Review Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -481,34 +481,6 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
 	return s.datastore.IsReady(ctx)
 }
 
-type AuthenticationOverride struct{}
-
-var _ grpc_auth.ServiceAuthFuncOverride = (*AuthenticationOverride)(nil)
-
-// AuthFuncOverride skips middleware/AuthFunc
-func (m AuthenticationOverride) AuthFuncOverride(ctx context.Context, fullMethodName string) (context.Context, error) {
-	return ctx, nil
-}
-
-type openfgaHealthServer struct {
-	healthv1pb.UnimplementedHealthServer
-	Server *Server
-	AuthenticationOverride
-}
-
-func (o openfgaHealthServer) Check(ctx context.Context, request *healthv1pb.HealthCheckRequest) (*healthv1pb.HealthCheckResponse, error) {
-	_, err := o.Server.IsReady(ctx)
-	if err != nil {
-		return &healthv1pb.HealthCheckResponse{Status: healthv1pb.HealthCheckResponse_NOT_SERVING}, err
-	} else {
-		return &healthv1pb.HealthCheckResponse{Status: healthv1pb.HealthCheckResponse_SERVING}, nil
-	}
-}
-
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -481,34 +481,6 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
 	return s.datastore.IsReady(ctx)
 }
 
-type AuthenticationOverride struct{}
```
jon-whit, line 484: Moved this section down to the bottom so it doesn't intercept the other server method declarations.

server/server.go
```diff
@@ -662,3 +634,43 @@ func (s *Server) resolveAuthorizationModelID(ctx context.Context, store, modelID
 
 	return modelID, nil
 }
+
+// HealthChecker defines an interface that services can implement for server health checks.
+type HealthChecker interface {
```
jon-whit, line 639: Introduced this so that it's more injectable and more easily testable :)

server/server.go
```diff
@@ -662,3 +634,43 @@ func (s *Server) resolveAuthorizationModelID(ctx context.Context, store, modelID
...
+
+type openfgaHealthServer struct {
+	healthv1pb.UnimplementedHealthServer
+	HealthChecker
+}
+
+var _ grpc_auth.ServiceAuthFuncOverride = (*openfgaHealthServer)(nil)
+
+// AuthFuncOverride implements the grpc_auth.ServiceAuthFuncOverride interface by bypassing authn middleware.
+func (o *openfgaHealthServer) AuthFuncOverride(ctx context.Context, fullMethodName string) (context.Context, error) {
```
jon-whit, line 651: Attached this directly to the `openfgaHealthServer` instead of the intermediate type composition.

server/server.go
```diff
@@ -662,3 +634,43 @@ func (s *Server) resolveAuthorizationModelID(ctx context.Context, store, modelID
...
+		}
+
+		if !ready {
+			return &healthv1pb.HealthCheckResponse{Status: healthv1pb.HealthCheckResponse_NOT_SERVING}, nil
+		}
+
+		return &healthv1pb.HealthCheckResponse{Status: healthv1pb.HealthCheckResponse_SERVING}, nil
+	}
+
+	return nil, status.Errorf(codes.NotFound, "service '%s' is not registered with the Health server", service)
```
jon-whit, line 671: According to the grpc Health protocol

> If the service name is not registered, the server returns a NOT_FOUND GRPC status.

PR #206, opened 2022-08-30
chore: assign random port to openfga service in tests

jon-whit:
## Description
This fixes flaky tests by auto assigning a random port to the OpenFGA service when spinning the service up in the `pkg/cmd/service/service_test.go` test suite.

## Review Checklist


pkg/cmd/service/service_test.go
```diff
@@ -48,14 +43,10 @@ func TestMain(m *testing.M) {
...
 	}
 
-	config := DefaultConfig()
-	openFGAServerAddr = fmt.Sprintf("http://localhost%s", config.HTTP.Addr)
-	openFGAGRPCAddr = fmt.Sprintf("localhost%s", config.GRPC.Addr)
-
 	os.Exit(m.Run())
 }
 
-func ensureServiceUp(t *testing.T, transportCredentials credentials.TransportCredentials, httpHealthCheck bool) {
```
jon-whit, line 58: I'd like to refactor this composition at some point. This is starting to get rough. I'm going to save that minor refactor for a later PR though. The main goal is to try and fix flaky tests for now.

pkg/cmd/service/service_test.go
```diff
@@ -311,7 +302,7 @@ func TestBuildServiceWithPresharedKeyAuthentication(t *testing.T) {
 }
 
 func TestHTTPServerWithCORS(t *testing.T) {
-	config, err := GetServiceConfig()
+	config, err := DefaultConfigWithRandomPorts()
```
jon-whit, line 305: @craigpastro haha I changed my mind on this. I like using `GetServiceConfig` because it's higher fidelity and the actual path our server uses on startup to get config, but we'd always be overrding the GRPC and HTTP addresses based on the auto assigned port, and so since we don't use env variables or anything in our tests today I decided I changed my mind on this point.

Sorry for the confusion :)

pkg/cmd/service/service.go
```diff
@@ -147,6 +148,29 @@ type Config struct {
 	Profiler   ProfilerConfig
 }
 
+func DefaultConfigWithRandomPorts() (*Config, error) {
```
craigpastro, line 151: Is this more of a test utility, rather than something that should be in a non-test file?

pkg/cmd/service/service.go
```diff
@@ -147,6 +148,29 @@ type Config struct {
...
+	httpPort := l.Addr().(*net.TCPAddr).Port
+
+	l, err = net.Listen("tcp", "")
+	if err != nil {
+		return nil, err
+	}
+	defer l.Close()
+	grpcPort := l.Addr().(*net.TCPAddr).Port
+
+	config.GRPC.Addr = fmt.Sprintf(":%d", grpcPort)
```
craigpastro, line 168: At this point do you think it is worth explicitly using [`netip.AddrPort`](https://pkg.go.dev/net/netip#AddrPort) for the type of `config.GRPC.Addr` (and also the HTTP version). I would prefer having a typed address rather than a string.

pkg/cmd/service/service.go
```diff
@@ -227,6 +251,8 @@ func GetServiceConfig() (*Config, error) {
 
 type service struct {
 	server        *server.Server
+	grpcPort      string
```
craigpastro, line 254: Similary here. Can we use grcpAddrPort of type netip.AddrPort?

pkg/cmd/service/service_test.go
craigpastro: lol 👍

pkg/cmd/service/service.go
jon-whit: It could still be useful outside of the scope of tests. If you want to pass config to OpenFGA that uses random ports then this works for you. That's why I put it there.

pkg/cmd/service/service.go
jon-whit: Probably, yes. Let me look into that approach as an alternative and see how it turns out 👍

pkg/cmd/service/service_test.go
```diff
@@ -417,12 +408,12 @@ func TestHTTPServerWithCORS(t *testing.T) {
...
 	const localOIDCServerURL = "http://localhost:8083"
 
-	config, err := GetServiceConfig()
+	config, err := DefaultConfigWithRandomPorts()
 	require.NoError(t, err)
 
 	config.Authn.Method = "oidc"
 	config.Authn.AuthnOIDCConfig = &AuthnOIDCConfig{
-		Audience: openFGAServerAddr,
+		Audience: "openfga.dev",
```
miparnisari, line 416: this doesn't change anything right?

pkg/cmd/service/service.go
craigpastro: 👍

pkg/cmd/service/service_test.go
jon-whit: openFGAServerAddr isn't static anymore and this happens before the server starts up, because this config is a dependency of the server. So I just hardcoded this since it's just an audience string anyways.

jon-whit:
We can't run tests in parallel yet because our server code has race conditions on shared globals with packages we depend on :( .

https://github.com/openfga/openfga/blob/main/server/server.go#L150
https://github.com/openfga/openfga/blob/main/server/server.go#L528

We need to refactor to avoid the races on these. Ideally `New()` is a singleton depending on no shared globals. `Run()`, ideally, also shouldn't depend on shared variables across packages.

craigpastro (approved):
LGTM

PR #211, opened 2022-09-01
refactor: validation for writing auth models

craigpastro:
## Description

Simplify the validation for auth models. Previously the code was a bit hard to understand... and I was the one who wrote it.

This will make it much easier to update to the new version and also to add the checks on relation types.

## Review Checklist


server/commands/write_authzmodel.go
```diff
@@ -50,115 +55,104 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
...
+	return nil
+}
 
-		targets, err := validateTypeDefinition(td)
-		if err != nil {
-			return err
-		}
-		tupleToUsersetTargets = append(tupleToUsersetTargets, targets...)
-	}
+func containsDuplicateType(tds []*openfgapb.TypeDefinition) bool {
```
miparnisari, line 70:
```suggestion
func containsDuplicateTypes(tds []*openfgapb.TypeDefinition) bool {
```

server/errors/errors.go
```diff
@@ -84,8 +84,8 @@ func RelationNotFound(relation string, typeName string, tuple *openfgapb.TupleKe
 	return status.Error(codes.Code(openfgapb.ErrorCode_relation_not_found), msg)
 }
 
-func EmptyRelationDefinition(typeDefinition, relation string) error {
-	return status.Error(codes.Code(openfgapb.ErrorCode_empty_relation_definition), fmt.Sprintf("Type definition '%s' contains an empty relation definition '%s'", typeDefinition, relation))
```
miparnisari, line 88: This is a user-facing error. I feel like the existing message is clearer. https://auth0.slack.com/archives/C033LQD4S11/p1662068452184499?thread_ts=1662066590.039589&cid=C033LQD4S11

server/commands/write_authzmodel.go
craigpastro: Fixed. Thank you!

server/errors/errors.go
craigpastro: How about now? Previously it was kind of wrong. Do you have any suggestions? It is in the case a user has something like
```
type doc
  relations
    define owner  // and nothing else
```

miparnisari: My nitpick is that we don't mention the word `rewrite` anywhere in our documentation.

How about `the definition of relation R on type T is invalid` ?

craigpastro: Oh yeah, that's much better. Thanks! Updated.

miparnisari (approved):

PR #223, opened 2022-09-10
feat: add validations for 1.1 authorization models

craigpastro:
## Description

Needs https://github.com/openfga/api/pull/27 to be merged first.

## Review Checklist


pkg/cmd/service/service_test.go
```diff
@@ -590,8 +590,6 @@ func TestHTTPServingTLS(t *testing.T) {
 			KeyPath:  certsAndKeys.serverKeyFile,
 		}
 
-		fmt.Println(config)
```
elbuo8, line 593: ?

craigpastro: I think I added this in on some PR way back when and now I am trying to sneak the deletion of it in this PR.

server/commands/write_authzmodel.go
```diff
@@ -29,14 +31,23 @@ func NewWriteAuthorizationModelCommand(
...
+
 	typeDefinitions := req.GetTypeDefinitions().GetTypeDefinitions()
 
 	// Until this is solved: https://github.com/envoyproxy/protoc-gen-validate/issues/74
 	if len(typeDefinitions) > w.backend.MaxTypesInTypeDefinition() {
 		return nil, serverErrors.ExceededEntityLimit("type definitions in an authorization model", w.backend.MaxTypesInTypeDefinition())
 	}
 
-	if err := validateAuthorizationModel(typeDefinitions); err != nil {
+	if err := validateAuthorizationModel(req.GetSchemaVersion(), typeDefinitions); err != nil {
```
elbuo8, line 50: `schemaVersion` is already defined above.

craigpastro: Oops. Thank you! Fixed.

server/commands/write_authzmodel.go
```diff
@@ -163,3 +185,83 @@ func isUsersetRewriteValid(allRelations map[string]struct{}, relationsOnType map
...
+	for _, td := range tds {
+		for _relation, _ := range td.GetRelations() {
+			metadata, ok := td.GetMetadata().GetRelations()[_relation]
+			if ok {
+				for _, userType := range metadata.GetDirectlyRelatedUserTypes() {
+					objectType := userType.GetType()
+					relation := userType.GetRelation()
+
+					if _, ok := typeToRelations[objectType]; !ok {
+						return serverErrors.InvalidRelationType(td.GetType(), _relation, objectType, relation)
```
miparnisari, line 211: i think these two errors should be different, because at this point, `relation` can be empty

server/test/write_authzmodel.go
```diff
@@ -337,28 +470,78 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
...
+}
+
+func TestWriteAuthorizationModelWithRelationalTypes(t *testing.T, datastore storage.OpenFGADatastore) {
+	var tests = []struct {
+		name    string
+		request *openfgapb.WriteAuthorizationModelRequest
+		err     error
+	}{
+		{
+			name: "succeeds",
```
miparnisari, line 526:
```suggestion
			name: "succeeds when DirectlyRelatedUserTypes has only a type",
```

server/commands/write_authzmodel.go
craigpastro: It is the type of the relation that is invalid so I think this makes sense. Suppose the type is `user` and then:

1. Without relation the error is "the type "user" is invalid on ..."
2. With a relation, say `member`, the error is "the type "user#member` is invalid on ..."

WDYT?

server/test/write_authzmodel.go
```diff
@@ -367,37 +550,141 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
 					},
 				},
 			},
-			err: errors.RelationNotFound("owner", "", nil),
 		},
 		{
-			_name: "ExecuteWriteFailsIfTupleToUsersetReferencesUnknownRelation",
+			name: "succeeds part 2",
```
miparnisari, line 555:
```suggestion
			name: "succeeds when DirectlyRelatedUserTypes has a type and a relation",
```

server/test/write_authzmodel.go
```diff
@@ -494,38 +796,73 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
...
+			err: errors.InvalidRelationType("document", "reader", "user", ""),
 		},
 	}
+
+	ctx := context.Background()
+	logger := logger.NewNoopLogger()
+
 	for _, test := range tests {
-		t.Run(test._name, func(t *testing.T) {
+		t.Run(test.name, func(t *testing.T) {
```
miparnisari, line 865: please send a different PR with this change 😬

server/commands/write_authzmodel.go
miparnisari: You mean that you would use `InvalidRelationType` in both but format the error nicely? If so, 👍

server/commands/write_authzmodel.go
```diff
@@ -163,3 +185,83 @@ func isUsersetRewriteValid(allRelations map[string]struct{}, relationsOnType map
...
+			typeToRelations[typeName][relationName] = struct{}{}
+		}
+	}
+
+	// Here we are checking that every type on a relation is valid. This means:
+	// 1. If it is a type (e.g. user) then this type is a type in the model
+	// 2. If it is a type#relation then this type with this relation is in the model
+	for _, td := range tds {
+		for _relation, _ := range td.GetRelations() {
+			metadata, ok := td.GetMetadata().GetRelations()[_relation]
```
miparnisari, line 204: if people give us some relations with types defined but others with nothing (i.e. `td.GetMetadata() == nil`), we should throw an error

server/test/write_authzmodel.go
```diff
@@ -367,37 +550,141 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
...
+				StoreId:       storeID,
+				TypeDefinitions: []*openfgapb.TypeDefinition{
+					{
+						Type: "document",
+						Relations: map[string]*openfgapb.Userset{
+							"reader": {Userset: &openfgapb.Userset_This{}},
+						},
+						Metadata: &openfgapb.Metadata{
+							Relations: map[string]*openfgapb.RelationalMetadata{
 								"reader": {
```
miparnisari, line 650: can you add a test for the case where the auth model has 2 types but only one has defined metadata? (we should not allow it)

miparnisari, line 650: what happens if the relation `owner` exists but is not defined in the `metadata` map?

server/commands/write_authzmodel.go
craigpastro: Yes. In this case, if so, then `ok` should be false.

server/commands/write_authzmodel.go
```diff
@@ -60,7 +65,13 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
...
 //     a. Be valid relations on the same type in the authorization model (in cases of computedUserset)
 //     b. Be valid relations on another existing type (in cases of tupleToUserset)
-func validateAuthorizationModel(tds []*openfgapb.TypeDefinition) error {
+//
+// If it is a 1.1 model (with types on relations), additionally:
+//  3. Every type on a relation must be a valid type:
+//     a. For a type (e.g. user) this means checking that this type is in the model
+//     b. For a type#relation this means checking that this type with this relation is in the model
+//  4. Check that a relation is assignable if and only if it has a non-zero list of types
+func validateAuthorizationModel(schemaVersion typesystem.SchemaVersion, tds []*openfgapb.TypeDefinition) error {
```
jon-whit, line 74: Let's put this and similar methods under the `typesystem` package since it is logic specific to the validation of types in an authorization model.

craigpastro: I love this idea, but do you mind if I do it in a following PR? I have one PR to follow this which will introduce the `typesystem` package and get the schema version into the datastore and that has already blown up like crazy.

server/test/write_authzmodel.go
craigpastro: Updated. Thanks!

server/test/write_authzmodel.go
craigpastro: Updated. Thanks!

server/test/write_authzmodel.go
craigpastro: https://github.com/openfga/openfga/pull/238

server/test/check.go
```diff
@@ -1400,7 +1400,7 @@ func TestCheckQuery(t *testing.T, datastore storage.OpenFGADatastore) {
 	data, err := os.ReadFile(gitHubTestDataFile)
 	require.NoError(t, err)
 
-	var gitHubTypeDefinitions openfgapb.TypeDefinitions
+	var gitHubTypeDefinitions openfgapb.WriteAuthorizationModelRequest
```
jon-whit, line 1403: The json data contained in `./testdata/github.json` encodes one or more type definitions, but it doesn't encode a `WriteAuthorizationModelRequest`.

This should remain as is should it not?

server/commands/write_authzmodel.go
jon-whit: I'd prefer we do it here since it is technically part of the scope of this work. This PR introduces the `typesystem` package and since this method is a static method, there shouldn't be any reason why we can't just quickly copy/paste it to that package and then reference it. We're already importing the `typesystem` package in this code, so we may as well move this to where it belongs now. I don't see a need to postpone that for another PR.

pkg/typesystem/typesystem.go
```diff
@@ -0,0 +1,31 @@
...
+
+type SchemaVersion int
+
+const (
+	SchemaVersionUnspecified SchemaVersion = iota
+	SchemaVersion1_0
+	SchemaVersion1_1
+)
+
+func NewSchemaVersion(s string) SchemaVersion {
```
jon-whit, line 11: Please add a `typesystem_test.go` test file and test this and the `String()` method below.

server/test/check.go
craigpastro: We don't have an `openfgapb.TypeDefinitions` anymore, which is why this changed. I imagine in the future when we update these models to have types we will need to include the schema version as well.

server/test/server.go
```diff
@@ -62,11 +62,8 @@ func RunQueryTests(t *testing.T, ds storage.OpenFGADatastore) {
 
 func RunCommandTests(t *testing.T, ds storage.OpenFGADatastore) {
 	t.Run("TestWriteCommand", func(t *testing.T) { TestWriteCommand(t, ds) })
-
-	t.Run("TestWriteAuthorizationModel",
-		func(t *testing.T) { TestWriteAuthorizationModel(t, ds) },
-	)
-
+	t.Run("TestWriteAuthorizationModel", func(t *testing.T) { TestWriteAuthorizationModel(t, ds) })
+	t.Run("TestWriteAuthorizationModelWithRelationalTypes", func(t *testing.T) { TestWriteAuthorizationModelWithRelationalTypes(t, ds) })
```
jon-whit, line 66: Can we refactor this name to `TestWriteAuthorizationModelWithTypeInfo` and the same goes for the actual function name getting called in `server/test/write_authzmodel.go`

server/commands/write_authzmodel.go
```diff
@@ -29,14 +30,18 @@ func NewWriteAuthorizationModelCommand(
 
 // Execute the command using the supplied request.
 func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openfgapb.WriteAuthorizationModelRequest) (*openfgapb.WriteAuthorizationModelResponse, error) {
-	typeDefinitions := req.GetTypeDefinitions().GetTypeDefinitions()
+	schemaVersion := typesystem.NewSchemaVersion(req.GetSchemaVersion())
+	if schemaVersion == typesystem.SchemaVersionUnspecified {
```
jon-whit, line 34: Existing clients will not be providing a `schema_version` field in their requests, and thus this will default to `SchemaVersionUnspecified`, which will lead to this error every time.

We need to account for this to make sure we don't break people's existing integrations. What are your thoughts?

jon-whit: Nevermind. We return 1.0 on "". Sorry

server/commands/write_authzmodel.go
jon-whit, line 74: Mind changing this to take an `AuthorizationModel` instead of type definitions?

jon-whit: Also, do you mind changing this to take an `AuthorizationModel` instead of type definitions?

server/commands/write_authzmodel.go
```diff
@@ -163,3 +180,83 @@ func isUsersetRewriteValid(allRelations map[string]struct{}, relationsOnType map
 
 	return nil
 }
+
+func areRelationalTypesValid(tds []*openfgapb.TypeDefinition) error {
+	typeToRelations := map[string]map[string]struct{}{}
```
jon-whit, line 185: There should be a helper method for this in the `typesystem` package.

I think we should have something like

```
package typesystem

type Typesystem struct {
    model *openfgapb.AuthorizationModel
}

func New(model *openfga.AuthorizationModel) *Typesystem {
    ....
}

func (t *Typesystem) GetTypeRelationsMap() map[string]map[string]Relation {
    // logic here
}
```

server/commands/write_authzmodel.go
jon-whit: If `ok` is false this indicates the relation had no type information associated with it, correct?

e.g.

```
type user

type document
  relations
    define reader: [user]
    define can_read: reader
```

Is that correct?

server/commands/write_authzmodel.go
```diff
@@ -163,3 +180,83 @@ func isUsersetRewriteValid(allRelations map[string]struct{}, relationsOnType map
...
+					return serverErrors.NonassignableRelationHasAType(td.GetType(), relation)
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func isAssignable(rewrite *openfgapb.Userset) bool {
```
jon-whit, line 238: This should also be a helper in the `typesystem` package.

```
package typesystem

type Typesystem struct {...}

func (t *Typesystem) IsAssignableRelation(objectType, relation string) (bool, error) {
   // logic here
}
```

server/commands/write_authzmodel.go
```diff
@@ -163,3 +180,83 @@ func isUsersetRewriteValid(allRelations map[string]struct{}, relationsOnType map
...
+	switch rw := rewrite.GetUserset().(type) {
+	case *openfgapb.Userset_This:
+		return true
+	case *openfgapb.Userset_Union:
```
jon-whit, line 242: If we used `UsersetRewrite_SetOp` here instead of individual `Union`, `Intersection`, `Difference` this would collapse into

```
case *openfgapb.UsersetRewrite_SetOp:
    for _, child := rw.GetSetOperation().GetChild() {
       if isAssignable(child) { return true }
    }
```
Another +1 opportunity for that refactor at some point 😄

server/test/write_authzmodel.go
```diff
@@ -362,28 +491,78 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
...
+}
+
+func TestWriteAuthorizationModelWithRelationalTypes(t *testing.T, datastore storage.OpenFGADatastore) {
+	var tests = []struct {
+		name    string
+		request *openfgapb.WriteAuthorizationModelRequest
+		err     error
+	}{
+		{
+			name: "succeeds on a valid model with a type type",
```
jon-whit, line 547: with a type type?

server/test/write_authzmodel.go
```diff
@@ -362,28 +491,78 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
...
+						Type: "document",
+						Relations: map[string]*openfgapb.Userset{
+							"reader": {Userset: &openfgapb.Userset_This{}},
+						},
+						Metadata: &openfgapb.Metadata{
+							Relations: map[string]*openfgapb.RelationMetadata{
+								"reader": {
+									DirectlyRelatedUserTypes: []*openfgapb.RelationReference{
+										{
+											Type: "user",
```
jon-whit, line 565: I've been using help methods for things like this to shrink it from 3 lines to 1.

e.g. `relationref("user", "")`
or `relationref("group", "member")`

server/test/write_authzmodel.go
```diff
@@ -392,37 +571,107 @@ func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastor
...
 										},
 									},
 								},
+								"member": {
+									DirectlyRelatedUserTypes: []*openfgapb.RelationReference{
+										{
+											Type: "user",
```
jon-whit, line 602: Sorry, did you mean to have `relationref("group", "member")`. 

The test is called "succeeds on a valid model with a type and type#relation type" but I don't see a `type#relation` reference.

I'm assuming you're testing

```
type user

type group
  relations
    define admin: [user]
```
Is that correct?

Also, as a general note, it would be really cool if we could provide the DSL syntax as a string in these tests for validation 😄 That would be so nice!

@rhamzeh this is another opportunity why I feel it would be good for us to leverage WASM with the DSL transformer. If we could pass in a DSL string in our tests and run it through a DSL transformer wasm module loaded in Go which outputs the protobuf model, then our tests like this could be easier to read and understand.

server/commands/write_authzmodel.go
craigpastro: Correct. Or for some reason the JSON doesn't contain it.

server/commands/write_authzmodel.go
craigpastro: Alright. Let's do it 🚀

server/test/write_authzmodel.go
jon-whit: It would become:

```
tests := []struct{
  ...
  model string
}{
  {
    ...
    model: `
        type user

        type document
          relations
            define reader: [user]
    `
  },
}
```
instead of writing out the protobuf structs everywhere.

server/test/write_authzmodel.go
```diff
@@ -7,24 +7,16 @@ import (
...
-func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastore) {
-	type writeAuthorizationModelTestSettings struct {
-		name    string
-		request *openfgapb.WriteAuthorizationModelRequest
-		err     error
-	}
-
-	ctx := context.Background()
-	logger := logger.NewNoopLogger()
+const storeID = "01GCD5SAJVC2Y38MJBNJ9QVZ60"
```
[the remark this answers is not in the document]
craigpastro: Good idea. Thanks. Updated in a new PR and will update here if we get it in!

server/test/server.go
craigpastro: Done. Thanks!

server/test/write_authzmodel.go
craigpastro: Yeah. haha. I'll update it to read "with a objectType type".

server/test/write_authzmodel.go
craigpastro: If you look a bit more down the `document` type relations have `group#member` in them.

server/test/write_authzmodel.go
```diff
@@ -7,24 +7,16 @@ import (
...
 
+func TestWriteAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastore) {
```
jon-whit, line 19: Can we add a test that uses TupleToUserset to verify the `tupleset` relation validates against a defined relation as well?

server/commands/write_authzmodel.go
jon-whit: Can we add some more tests for this that make sure behavior under intersection, union, and exclusion are working as expected? I didn't see any rewrites that involved union, intersection, or exclusion in the newly added tests for this functionality.

server/test/write_authzmodel.go
craigpastro: Doesn't `succeeds part II` have such?

server/commands/write_authzmodel.go
jon-whit: If it doesn't contain it then we assume it is not assignable, and if it references `self` then the logic below (determining it's assignability) would fail yeah?

server/test/write_authzmodel.go
jon-whit: Yeah, probably `typesystem`. I think that's a decent home, because it's referring to relations on a specific object type.

server/test/write_authzmodel.go
jon-whit: Yeah, sorry I missed that. That's why I deleted the first part of my comment 😄 . Sorry about that. The comment about the DSL still applies though. That'd be cool to do at a later time.

server/test/write_authzmodel.go
jon-whit: Sorry, I put this on the wrong test. I meant for the tests with type restrictions.

I don't see a test that involves union, intersection, exclusion or tuple_to_userset with assignable vs non-assignable types.

jon-whit: Ideally these kind of tests would belong on the helper methods in the `typesystem` package though like we discussed. That way it's more clear what paths of these critical helpers have been fully tested.

pkg/typesystem/typesystem.go
craigpastro: Tests added. Thanks!

server/commands/write_authzmodel.go
craigpastro: Thanks! As discussed, let's refactor in a follow up PR.

server/commands/write_authzmodel.go
craigpastro: Tests added. Please take a look!

server/test/write_authzmodel.go
craigpastro: Again, in a follow up PR. Is this `typesystem_testutils` stuff? How do we arrange all this?

pkg/typesystem/typesystem.go
```diff
@@ -0,0 +1,251 @@
...
+		err = isUsersetRewriteValid(allRelations, relationsOnType, objectType, relation, t.Difference.Subtract)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func areRelationalTypesValid(tds []*openfgapb.TypeDefinition) error {
```
jon-whit, line 173: Can we please refactor this so that it's easier to understand and read?

I think it makes more sense for this code to be composed like so:

```
func (t *Typesystem) GetRelation(
    objectType, relation string,
) (*openfgapb.Relation, error) {
    // todo: implement me
}

func (t *Typesystem) DirectlyRelatedUserTypes(
    objectType, relation string,
) ([]*openfgapb.RelationReference, error) {
	// todo: implement me
}

func (t *Typesystem) IsDirectlyAssignable(relation *openfgapb.Relation) (bool, error) {
    return containsSelf(relation.GetRewrite())
}

func containsSelf(rewrite *openfgapb.Userset) bool {
    // same code as isAssignable in this PR
}

func (t *Typesystem) ValidateRelationTypeRestrictions(objectType, relation string) error {
    relation, err := t.GetRelation(objectType, relation)
    if err != nil {
        // return an error because objectType or relation is undefined
    }

    relatedTypes := relation.GetTypeInfo().GetDirectlyRelatedUserTypes()

    assignable := t.IsDirectlyAssignable(relation)
    if assignable && len(relatedTypes) < 1 {
        // return an error since there were no type restrictions specified
    }

    if !assignable && len(relatedTypes) != 0 {
        // return an error since there were type restrictions on an unassignable type
    }

    for _, relatedType := range relatedTypes {
        relatedObjType := relatedType.GetObjectType()
        relatedRelation := relatedType.GetRelation()
        
        if relatedRelation != "" {
            _, err := t.GetRelation(relatedObjType, relatedRelation))
            if err != nil {
                // related object type or relation does not exist, return appropriate err
            }

            continue
        }

        _, err := t.GetTypeDefinition(relatedObjType)
        if err != nil {
            // related type does not exist, return appropriate err
        }		
    }

    return nil
}
```

The way it is written now is quite hard to read because you have to cut through so much boilerplate map code to understand the primary important pieces.

jon-whit (changes requested):
Left one last comment about code composition and readability. Let's refactor the code a bit within the scope of this PR so that it leverages more reusable code pieces and makes the code easier to read, and then I think we're good.

Thanks for adding the tests!

pkg/typesystem/typesystem.go
jon-whit: Related to https://github.com/openfga/api/pull/33 as well. Makes this composition possible.

server/test/write_authzmodel.go
jon-whit: I have a follow up PR that will add this. I'll submit that once we merge this 👍

server/commands/write_authzmodel.go
jon-whit: I left a comment here:
https://github.com/openfga/openfga/pull/223#discussion_r974494567

I don't think this work should get put off, because it's part of the review of this code. I feel we should make the composition of this code more readable before merging, because it makes the code easier to read and maintain and also lays the foundation for further developments to make use of more reusable software pieces related to type definitions.

The last remaining bit that I feel should be done is just a slight refactor like I recommended below. I think once we make it a little more readable through better composition, then we're good to merge 👍

pkg/typesystem/typesystem.go
```diff
@@ -0,0 +1,251 @@
...
+func (m *AuthorizationModel) ToProto() *openfgapb.AuthorizationModel {
+	return &openfgapb.AuthorizationModel{
+		Id:              m.ID,
+		SchemaVersion:   m.Version.String(),
+		TypeDefinitions: m.TypeDefinitions,
+	}
+}
+
+// Validate validates the model according to the following rules:
+//  1. Do not allow duplicate types (or duplication relations but that is inherent in the map structure)
```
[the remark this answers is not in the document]
craigpastro: Thank you! Fixed.

server/commands/write_authzmodel.go
```diff
@@ -29,137 +30,37 @@ func NewWriteAuthorizationModelCommand(
...
-		err := isUsersetRewriteValid(allRelations, relationsOnType, objectType, relation, t.Difference.Base)
-		if err != nil {
-			return err
-		}
-
-		err = isUsersetRewriteValid(allRelations, relationsOnType, objectType, relation, t.Difference.Subtract)
-		if err != nil {
-			return err
-		}
+	utils.LogDBStats(ctx, w.logger, "WriteAuthzModel", 0, 1)
```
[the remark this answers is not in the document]
craigpastro: I don't think I changed the location of this function. I believe that it is where it is since even if the write fails this log will be sent. I am not sure I agree with that, but 🤷

server/commands/write_authzmodel.go
```diff
@@ -29,137 +30,42 @@ func NewWriteAuthorizationModelCommand(
...
+	if err := w.backend.WriteAuthorizationModel(ctx, req.GetStoreId(), id, typeSystem.GetTypeDefinitions()); err != nil {
```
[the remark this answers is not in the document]
craigpastro: I want to do this in a follow up PR as it is quite a lot and it would be nice to clean up the interfaces as
```
WriteAuthorizationModel(ctx, req.GetStoreId(), id, schemaVersion, typeSystem.GetTypeDefinitions())
```
is getting a bit long. I had such a follow up PR before but it will need a lot of ❤️  now.

jon-whit (approved):
LGTM for now. We can introduce a follow up PR to handle some minor improvements to the typesystem package and also store the schema version etc..

PR #210, opened 2022-09-01
Add support for MySQL datastore backend.

MidasLamb:
## Description
Add support for MySQL as a datastore backend.

## References
https://discordapp.com/channels/759188666072825867/930524706854031421/1004733828381810698

Documentation PR: https://github.com/openfga/openfga.dev/pull/198

## Review Checklist
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #227, opened 2022-09-12
refactor: add start-postgres-container to the makefile

craigpastro:
## Description

Now that we are adding [other storage engines](https://github.com/openfga/openfga/pull/210), I think it would be useful to have commands in the Makefile that help with the local development of OpenFGA for these engines. Previously to start Postgres we could rely on the docker-compose file, but I don't suppose this will be the case with MySQL. So to keep things consistent, this PR adds a `start-postgres-container` command to the Makefile which can be used with the `migrate-postgres` and `run-postgres` commands. I envision we will have similar commands for MySQL.

## Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,11 +37,11 @@ start-postgres-container:
 .PHONY: migrate-postgres
 migrate-postgres: build
 	# nosemgrep: detected-username-and-password-in-uri
-	./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres?sslmode=disable'
+	./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres'
 
 .PHONY: run-postgres
 run-postgres: build
-	./openfga run --datastore-engine postgres --datastore-uri postgres://postgres:password@localhost:5432/postgres?sslmode=disable
+	./openfga run --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres'
 
 .PHONY: go-generate
 go-generate: install-tools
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,8 +70,3 @@ functional-test:
 .PHONY: bench
 bench: go-generate
 	go test ./... -bench=. -run=XXX -benchmem
-
-.PHONY: initialize-db
-initialize-db: go-generate
```
craigpastro, line 70: As far as I can tell this is no longer needed.

Makefile
```diff
@@ -29,14 +29,19 @@ build: ## Build/compile the OpenFGA service
 run: build ## Run the OpenFGA server with in-memory storage
 	./openfga run
 
+
+.PHONY: start-postgres-container
+start-postgres-container:
+	docker run -d --name postgres -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password postgres:14
```
miparnisari, line 35: can we add an echo here that shows the connection string?

Makefile
miparnisari: because the migrate command creates the tables, no?

craigpastro: Yep. I don't think we used this for anything else right?

Makefile
craigpastro: Okay. Now it will look like:
```
docker run -d --name postgres -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password postgres:14
6200b45f0ad4ec464490becb91faa18c1c8dcf5dc6a94e33c3bae06b642f4897
> use 'postgres://postgres:password@localhost:5432/postgres' to connect to postgres
```

Makefile
miparnisari: No

jon-whit (approved):
Looks good to me. The commands seem to work fine. The only source of feedback I have is the make outputs are pretty chatty.

craigpastro:
I'm happy to make them less chatty. Anything in particular?

PR #232, opened 2022-09-13
feat: validate types when adding new tuples

miparnisari:
On every write, and if the authorization model has `schemaVersion == 1.1`, we will validate that the `user` part of the tuple satisfies the type definitions on the authorization model.

⚠️ Needs https://github.com/openfga/openfga/pull/247 to be merged so that if `schemaVersion == 1.1` and the auth model is incorrect, we throw an error.

## References
- https://github.com/openfga/rfcs/pull/7
- https://github.com/openfga/api/pull/27

## Review Checklist


server/commands/write.go
```diff
@@ -87,6 +90,63 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	return nil
 }
 
+// validateTypes makes sure that when writing a tuple, the types are compatible.
+// 1. If the tuple is of the form (person:bob, reader, doc:budget), then the type "doc", relation "reader" allows type "person".
+// 2. If the tuple is of the form (group:abc#member, reader, doc:budget), then the type "doc", relation "reader" must allow type "group", relation "member".
+// 3. If the tuple is of type (*, reader, doc:budget), we allow it only if the type "doc" relation "reader" has "allow public" = true
```
craigpastro, line 96: Is this case still true?

miparnisari: 😬 no

server/commands/write.go
```diff
@@ -87,6 +90,63 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+		if errors.Is(err, storage.ErrNotFound) {
+			return &tupleUtils.TypeNotFoundError{TypeName: objectType}
+		}
+		return err
+	}
+	if typeDefinition.Metadata == nil {
+		// authorization model is old and does not have type information
+		return nil
+	}
+	relationInformation := typeDefinition.Metadata.Relations[tk.Relation]
```
craigpastro, line 112: I am not sure, but I think you may have to worry if relations is nil? Also if the map contains `tk.Relation`. Perhaps something like 
```
relationInformation, ok := typeDefinition.GetMetadata().GetRelations()[tk.Relation]
```
may help. The `Gets` handle the nils.

miparnisari: > The Gets handle the nils.

TIL! Thank you.

server/commands/write.go
```diff
@@ -87,6 +90,58 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+
+	dbCallsCounter.AddReadCall()
+	typeDefinition, err := c.datastore.ReadTypeDefinition(ctx, store, authorizationModelID, objectType)
+	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return &tupleUtils.TypeNotFoundError{TypeName: objectType}
+		}
+		return err
+	}
+	relations := typeDefinition.GetMetadata().GetRelations()
```
jon-whit, line 108: Let's hold off on this PR a bit until the `typesystem` package gets merged in from the other PRs. A lot of this code can be reduced to helpers that will be in that package.

server/commands/write.go
```diff
@@ -87,6 +90,58 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+	if relations == nil {
+		// authorization model is old or invalid and does not have type information
+		return nil
```
jon-whit, line 111: We should be handling the error case where it's invalid differently, right?

For example, if we're using schema_version 1.0 then relations will be nil and we can return nil safely, but if schema_version is 1.1 and relations is nill, then that's a problem. Let's make sure we cover that non-nil error case here.

server/test/write.go
```diff
@@ -28,7 +28,7 @@ type writeCommandTest struct {
 var tk = &openfgapb.TupleKey{
```
jon-whit, line 28: Refactor this to `tuple.NewTupleKey("repository:openfga/openfga", "administrator", "user:github|alice@openfga")`

server/test/write.go
```diff
@@ -43,6 +43,10 @@ var writeCommandTests = []writeCommandTest{
 		_name: "ExecuteWithSameTupleInWritesReturnsError",
 		// state
 		typeDefinitions: []*openfgapb.TypeDefinition{
+			{
+				Type:      "user",
```
jon-whit, line 47: Shouldn't all of the prior tests still work without adding an explicit user type? I think we should be adding tests asserting new behavior for writes to models under the new schema_version 1.1. The existing tests are under schema_version 1.0 and should remain as is, yeah?

server/commands/write.go
```diff
@@ -67,8 +68,10 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 		}
 
 		// Validate that we are not trying to write to an indirect-only relationship
-		err = validateHasDirectRelationship(tupleUserset, tk)
-		if err != nil {
+		if err := validateHasDirectRelationship(tupleUserset, tk); err != nil {
```
jon-whit, line 71: Under the new `typesystem` package there should be a method called `IsDirectlyAssignable(objectType, relation string)` which will return 'true' if the relation is assignable and 'false' otherwise.

We should use that here instead.

server/test/write.go
miparnisari: You're right! When I wrote the original commit i didn't have the schema version available. 

Added it in recent commits and went with option b. The existing tests target 1.0, the ones i added target 1.1.

server/commands/write.go
miparnisari: Cool, updated.

pkg/typesystem/typesystem.go
```diff
@@ -252,25 +252,25 @@ func (t *TypeSystem) validateRelationTypeRestrictions() error {
 	return nil
 }
 
-func isAssignable(rewrite *openfgapb.Userset) bool {
+func IsAssignable(rewrite *openfgapb.Userset) bool {
```
jon-whit, line 255: Can we please refactor this so that it's not associated with a rewrite but instead a relation itself.

e.g.

```
func (t *TypeSystem) IsDirectlyAssignable(relation *openfgapb.Relation) bool {
    rewrite := relation.GetUsersetRewrite()

    return containsSelf(rewrite)
}

func containsSelf(rewrite *openfgapb.Userset) bool {
    // copy/paste the current logic here
}
```
It's weird to me that we have an IsAssignable method that applies to a rewrite but not a Relation. Rewrites aren't assignable or not, relations are.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+			{
+				Object:   "org:openfga",
+				Relation: "owner",
+				User:     "impossible:1",
+			},
+		},
+		typeDefinitions: []*openfgapb.TypeDefinition{
+			{
+				Type:      "user",
+				Relations: map[string]*openfgapb.Userset{},
```
jon-whit, line 769: Let's omit these empty relation definitions.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+			},
+			},
+		},
+	},
+	{
+		_name: "Write fails if user field contains a type that does not exist",
+		// state
+		schemaVersion: typesystem.SchemaVersion1_1,
+		typeDefinitions: []*openfgapb.TypeDefinition{
+			{
```
jon-whit, line 800: Missing the typedef for `{Type: "user"}` here.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+				Type:      "user",
+				Relations: map[string]*openfgapb.Userset{},
+			},
+			{
+				Type: "group",
+				Relations: map[string]*openfgapb.Userset{
+					"member": {Userset: &openfgapb.Userset_This{}},
+				},
```
jon-whit, line 838: We're missing the type info for `group#member` here. Let's add it such that group#member is related to a user.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		request: &openfgapb.WriteRequest{
+			Writes: &openfgapb.TupleKeys{TupleKeys: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("document:budget", "reader", "user:abc"),
```
jon-whit, line 860: Can we please add a test case that makes sure the write with a user value such as `group:eng#member` is also invalid here. Since `reader` is restricted to `group`, a write such as `document:budget, reader, group:eng#member` is also invalid.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+			}},
+		},
+		err: serverErrors.InvalidTuple("Object of type user is not allowed to have relation reader with document:budget",
+			tuple.NewTupleKey("document:budget", "reader", "user:abc"),
+		),
+	},
+	{
+		_name: "Write succeeds if user field contains a type that is allowed by the authorization model (which only allows user:...)",
```
jon-whit, line 868: Can we also add a test case to make sure that Writes have to validate against any one of the directly related user types? All of the tests we have here only define 1 DirectlyRelatedUserTypes, and so it would be nice to add a test with 2 writes that succeeds because both of the writes validate against one of the two DirectlyRelatedUserTypes.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+		// state
+		schemaVersion: typesystem.SchemaVersion1_1,
+		typeDefinitions: []*openfgapb.TypeDefinition{
+			{
+				Type:      "user",
+				Relations: map[string]*openfgapb.Userset{},
+			},
+			{
+				Type: "group",
+				Relations: map[string]*openfgapb.Userset{
```
jon-whit, line 994: This isn't needed as we don't define any relations on it.

server/test/write.go
```diff
@@ -733,6 +752,335 @@ var writeCommandTests = []writeCommandTest{
...
+					},
+				},
+			},
+		},
+		request: &openfgapb.WriteRequest{
+			Writes: &openfgapb.TupleKeys{TupleKeys: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("document:budget", "reader", "*"),
+			}},
+		},
+		err: serverErrors.InvalidTuple("User=* is not allowed to have relation reader with document:budget",
```
jon-whit, line 1061: I think it's probably better to just say "The wildcard '*' is an invalid user value for the 'document' relation 'reader'"

server/commands/write.go
```diff
@@ -60,15 +66,35 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 		return serverErrors.InvalidWriteInput
 	}
 
+	var authModel *openfgapb.AuthorizationModel = nil
```
jon-whit, line 69:
```suggestion
	var authModel *openfgapb.AuthorizationModel
```
for short

server/commands/write.go
```diff
@@ -60,15 +66,35 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
 		}
 
 		// Validate that we are not trying to write to an indirect-only relationship
-		err = validateHasDirectRelationship(tupleUserset, tk)
-		if err != nil {
+		if !typesystem.IsAssignable(tupleUserset) {
+			return serverErrors.HandleTupleValidateError(&tupleUtils.IndirectWriteError{Reason: IndirectWriteErrorReason, TupleKey: tk})
+		}
+
+		if err := c.validateTypesForTuple(authModel, schemaVersion, tk, dbCallsCounter); err != nil {
```
jon-whit, line 97: Why do we need to pass `schemaVersion` if we're passing the `authModel`? The authModel contains the schema version. I vote we consolidate here.

server/commands/write.go
```diff
@@ -87,6 +113,72 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	return nil
 }
 
+// validateTypesForTuple makes sure that when writing a tuple, the types are compatible.
+// 1. If the tuple is of the form (person:bob, reader, doc:budget), then the type "doc", relation "reader" allows type "person".
+// 2. If the tuple is of the form (group:abc#member, reader, doc:budget), then the type "doc", relation "reader" must allow type "group", relation "member".
+// 3. If the tuple is of the form (*, reader, doc:budget), we allow it only if the type "doc" relation "reader" allows at least one type (with no relation)
+func (c *WriteCommand) validateTypesForTuple(authModel *openfgapb.AuthorizationModel, schemaVersion typesystem.SchemaVersion, tk *openfgapb.TupleKey, dbCallsCounter utils.DBCallCounter) error {
```
jon-whit, line 120: This is a static method. No need for it to be attached to an instance of `*WriteCommand`.

server/commands/write.go
```diff
@@ -87,6 +113,72 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	return nil
 }
 
+// validateTypesForTuple makes sure that when writing a tuple, the types are compatible.
+// 1. If the tuple is of the form (person:bob, reader, doc:budget), then the type "doc", relation "reader" allows type "person".
```
jon-whit, line 117: (user=person:bob, relation=reader, object=doc:budget)

server/commands/write.go
```diff
@@ -87,6 +113,72 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+// 3. If the tuple is of the form (*, reader, doc:budget), we allow it only if the type "doc" relation "reader" allows at least one type (with no relation)
+func (c *WriteCommand) validateTypesForTuple(authModel *openfgapb.AuthorizationModel, schemaVersion typesystem.SchemaVersion, tk *openfgapb.TupleKey, dbCallsCounter utils.DBCallCounter) error {
+	objectType, _ := tupleUtils.SplitObject(tk.GetObject())    // e.g. "doc"
+	userType, userID := tupleUtils.SplitObject(tk.GetUser())   // e.g. (person, bob) or (group, abc#member) or ("", *)
+	_, userRel := tupleUtils.SplitObjectRelation(tk.GetUser()) // e.g. (person:bob, "") or (group:abc, member) or (*, "")
+
+	// find the type information
+	var typeDefinitionForObject *openfgapb.TypeDefinition
+	var typeDefinitionForUser *openfgapb.TypeDefinition
+	for _, td := range authModel.GetTypeDefinitions() {
```
jon-whit, line 128: Please use the typesystem package here instead. You should be able to do:

```
relation, err := typesys.Relation(objectType, tk.GetRelation())
```
and then use the relation
```
relation.GetTypeInfo().GetDirectlyRelatedUserTypes()
// etc..
```

server/commands/write.go
craigpastro: Or let's agree to use the `object#relation@user` notation from the paper everywhere to keep things shorter 😬

server/commands/write.go
```diff
@@ -87,6 +113,72 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+		if td.GetType() == userType {
+			typeDefinitionForUser = td
+		}
+	}
+
+	relationsForObject := typeDefinitionForObject.GetMetadata().GetRelations()
+	if relationsForObject == nil {
+		if schemaVersion.String() == "1.1" {
+			// if we get here, there's a bug in the validation of WriteAuthorizationModel API
+			return serverErrors.NewInternalError("", errors.New("invalid authorization model"))
```
craigpastro, line 141: This error seems public facing so maybe:
```suggestion
             msg := "invalid authorization model"
			return serverErrors.NewInternalError(msg, errors.New(msg))
```

server/commands/write.go
```diff
@@ -87,6 +113,72 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
+			typeDefinitionForObject = td
+		}
+		if td.GetType() == userType {
+			typeDefinitionForUser = td
+		}
+	}
+
+	relationsForObject := typeDefinitionForObject.GetMetadata().GetRelations()
+	if relationsForObject == nil {
+		if schemaVersion.String() == "1.1" {
```
craigpastro, line 139: You could compare it to the enum if you would like.
```suggestion
		if schemaVersion == typesystem.SchemaVersion1_1 {
```

pkg/typesystem/typesystem.go
miparnisari: I can do that, but then I need to rollback my use of `typesystem.IsAssignable` in Write like you suggested, because I don't have the relation readily available there.

server/commands/write.go
miparnisari: I added a `log.warn` in this method, so I need this :(

server/commands/write.go
miparnisari: Haha @craigpastro i don't want to introduce a brand new notation :P

server/test/write.go
miparnisari: 🤔 sorry I don't understand this

server/test/write.go
miparnisari: Hmmm i'm hesitant to say "wildcard". https://docs.fga.dev/modeling/basics/public-access#02-check-that-the-relationship-exists

server/test/write.go
jon-whit: I just mean let's drop this, because it shouldn't be needed. If we omit the map then it will be nil, and that should be fine.

server/test/write.go
jon-whit: Why are you hesitant to say it? We refer to it in that doc page multiple times.
Screen Shot 2022-09-26 at 9 53 07 AM

server/commands/write.go
miparnisari: I'm confused.. i don't want to return an error 🤔 I want to continue execution assuming a 1.0 model.

server/test/write.go
miparnisari: Because the doc says "it is not a wildcard" 😅

server/test/write.go
miparnisari: Ahh okay. I was confused because I thought this should fail validation, but i realized i am not calling `commands.WriteAuthorizationModel` but `storage.WriteAuthorizationModel`

server/test/write.go
jon-whit: But the heading says "Wildcard Syntax Usage" 🤣

server/commands/write_test.go
```diff
@@ -142,6 +129,21 @@ func TestValidateWriteTuples(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			tracer := telemetry.NewNoopTracer()
+			logger := logger.NewNoopLogger()
+			dbCounter := utils.NewDBCallCounter()
```
craigpastro, line 134: Is this actually used anywhere?

craigpastro (dismissed):
LGTM

jon-whit (approved):

server/test/write.go
miparnisari: Following up this discussion: https://auth0.slack.com/archives/C033LQD4S11/p1664223016524409?thread_ts=1663955675.635259&cid=C033LQD4S11

PR #234, opened 2022-09-14
fix: skip malformed tuples involving tuple to userset definitions

jon-whit:
## Description
This fixes the bug reported in https://discord.com/channels/759188666072825867/920359187949707305/1019251304251523112

## References
(https://discord.com/channels/759188666072825867/920359187949707305/1019251304251523112)

## Review Checklist

### DIFF
--- a/server/commands/check.go
+++ b/server/commands/check.go
@@ -425,11 +425,12 @@ func (query *CheckQuery) resolveTupleToUserset(ctx context.Context, rc *resoluti
 			break // the user was resolved already, avoid launching extra lookups
 		}
 
-		if !tupleUtils.IsValidObject(tuple.GetUser()) && !tupleUtils.IsObjectRelation(tuple.GetUser()) {
+		userObj, userRel := tupleUtils.SplitObjectRelation(tuple.GetUser())
+
+		if !tupleUtils.IsValidObject(userObj) {
 			continue // TupleToUserset tuplesets should be of the form 'objectType:id' or 'objectType:id#relation' but are not guaranteed to be because it is neither a user or userset
 		}
 
-		userObj, userRel := tupleUtils.SplitObjectRelation(tuple.GetUser())
 		usersetRel := node.TupleToUserset.GetComputedUserset().GetRelation()
 
 		// userRel may be empty, and in this case we set it to usersetRel.
--- a/server/commands/check.go
+++ b/server/commands/check.go
@@ -425,6 +425,10 @@ func (query *CheckQuery) resolveTupleToUserset(ctx context.Context, rc *resoluti
 			break // the user was resolved already, avoid launching extra lookups
 		}
 
+		if !tupleUtils.IsValidObject(tuple.GetUser()) && !tupleUtils.IsObjectRelation(tuple.GetUser()) {
+			continue // TupleToUserset tuplesets should be of the form 'objectType:id' or 'objectType:id#relation' but are not guaranteed to be because it is neither a user or userset
+		}
+
 		userObj, userRel := tupleUtils.SplitObjectRelation(tuple.GetUser())
 		usersetRel := node.TupleToUserset.GetComputedUserset().GetRelation()
 
--- a/server/test/check.go
+++ b/server/test/check.go
@@ -1252,10 +1252,72 @@ var checkQueryTests = []checkQueryTest{
 		},
 	},
 	{
+		name:             "CheckUsersetAsUser_WithContextualTuples",
+		resolveNodeLimit: defaultResolveNodeLimit,
+		request: &openfgapb.CheckRequest{
+			TupleKey: tuple.NewTupleKey("team:iam", "member", "org:openfga#member"),
+			ContextualTuples: &openfgapb.ContextualTupleKeys{
+				TupleKeys: []*openfgapb.TupleKey{
+					tuple.NewTupleKey("team:iam", "member", "team:engineering#member"),
+					tuple.NewTupleKey("team:engineering", "member", "org:openfga#member"),
+				},
+			},
+		},
+		typeDefinitions: []*openfgapb.TypeDefinition{
+			{
+				Type: "team",
+				Relations: map[string]*openfgapb.Userset{
+					"member": {Userset: &openfgapb.Userset_This{}},
+				},
+			},
+			{
+				Type: "org",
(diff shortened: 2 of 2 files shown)


server/commands/check.go
```diff
@@ -425,6 +425,10 @@ func (query *CheckQuery) resolveTupleToUserset(ctx context.Context, rc *resoluti
 			break // the user was resolved already, avoid launching extra lookups
 		}
 
+		if !tupleUtils.IsValidObject(tuple.GetUser()) && !tupleUtils.IsObjectRelation(tuple.GetUser()) {
```
rhamzeh, line 428: You should use `IsValidUser` here, otherwise you are ignoring valid relationship tuples, namely `user_id` and `*`

server/test/check.go
```diff
@@ -1251,6 +1251,45 @@ var checkQueryTests = []checkQueryTest{
...
+				Type: "folder",
+				Relations: map[string]*openfgapb.Userset{
+					"viewer": {
+						Userset: &openfgapb.Userset_This{},
+					},
+				},
+			},
+		},
+		tuples: []*openfgapb.TupleKey{
+			tuple.NewTupleKey("document:doc1", "parent", "folder1"), // folder1 isn't an object or userset
```
rhamzeh, line 1287: > folder1 isn't an object or userset

But it is still valid, a user is able to write:
- `(user=anne, relation=viewer, object=folder:1)`
- `(user=folder:1, relation=parent, object=document:1)`

And expect `check(user=anne, relation=viewer, object=document:1)` to return `allowed: true`

server/commands/check.go
rhamzeh: What you need to be checking is that:
1) the user is of the correct user format
2) the user, if it has a type, is in the model
2) the user, if it has a relation, is in type definition of its type
2) the object, if it has a type, is in the model

rhamzeh (changes requested):

server/test/check.go
jon-whit: Right, but this isn't that. This is fixing

* `(user=anne, relation=viewer, object=folder:1)
* `(user=blah, relation=parent, object=document:doc1)`
which today returns an error, because the `parent` lookup on `blah` can't resolve because it is not associated with any object type, and thus an authz model cannot be loaded for it.

server/commands/check.go
jon-whit: I'm not sure it makes sense for a `*` or `user_id` value to be acceptable for a tupleset relationship. A tupleset relationship should always be referencing an object, and not a user_id or `*`. Otherwise you can't resolve the computed_userset on the tupleset relation (e.g. `parent`).

server/test/check.go
rhamzeh: In that case:
* is valid as it means objects of all types

```
(user=anne, relation=viewer, object=folder:1)
(user=*, relation=parent, object=document:doc1)
```
should return true as `*` is valid.

Re ignoring the user id here, that's cool, but can you update the test, because what it is testing is a very specific case that may or may not be indicative.

Write:
- `(user=anne, relation=viewer, object=folder:1)`
- `(user=folder:1, relation=parent, object=document:1)`
`check(user=anne, relation=viewer, object=document:1) // allowed=true`

Write:
- `(user=anne, relation=viewer, object=folder:1)`
- `(user=*, relation=parent, object=document:1)`
`check(user=anne, relation=viewer, object=document:1) // allowed=true`

Write:
- `(user=anne, relation=viewer, object=folder:1)`
- `(user=folder:1, relation=parent, object=document:1)`
- `(user=folder1, relation=parent, object=document:1)`
`check(user=anne, relation=viewer, object=document:1) // allowed=true`

Write:
- `(user=anne, relation=viewer, object=folder:1)`
- `(user=folder1, relation=parent, object=document:1)`
`check(user=anne, relation=viewer, object=document:1) // allowed=false`

server/commands/check.go
craigpastro: We want to ignore `user_id` here. Validation of stuff is done elsewhere. This is used to ensure that in a tuple to userset rewrite that we are only following usersets (and possibly `*`, but I haven't thought that far yet. It would be good to add a test of that case if we don't have it).

elbuo8: This gate should happen after:
```go
userObj, userRel := tupleUtils.SplitObjectRelation(tuple.GetUser())
```
then
```go
if !tupleUtils.IsValidObject(userObj) {
    continue
}
```

The relation is already being handled/filtered appropriately.

jon-whit: Consider the following model and tuples:

```
type group
  relations
    define viewer as self
    define member as self

type document
  relations
    define parent as self
```

```
1. document:doc1, parent, group:eng#member
2. document:doc2, parent, folder:folder1
3. document:doc3, parent, myid
4. group:eng, viewer, jon
```

If the `parent` relationship of a document is defined as a tupleset relation, then only tuple 2 is valid (technically speaking). However, today you can write all three of these tuples. 

I think it's dangerous to allow tuple 1, because then what will happen here is the parent lookup will happen on `group:eng`. If `group:eng` had a viewer (e.g. jon) then that user will have been granted indirect access to the document unintentionally.

If we do what you're suggesting @elbuo8 we'll have
```
userObj, userRel := tupleUtils.SplitObjectRelation(tuple.GetUser())
if !tupleUtils.IsValidObject(userObject) {
  ...
}
// throw away userRel
// so userObject=group:eng & userObject=folder:folder1
```
The tupleset relation `parent` will be looked up on both of those, which will lead to the odd behavior above.

If we feel this is appropriate, then so be it. But that seems quite fragile and broken. @aaguiarz any input here.

craigpastro: I don't think @elbuo8 said to throw away the `userRel` above. Just ensure that `userObj` is indeed an object.

elbuo8: @craigpastro correct.

server/test/check.go
jon-whit: For the last test case you enumerated I'm going to save that work for the PR that address #235 . I'll submit a PR to address that work next.

I've added tests for the other cases that don't already have tests for.

elbuo8 (approved):

craigpastro (approved):
LGTM

PR #237, opened 2022-09-15
Update docker-compose.yaml

trondhindenes:
update docker-compose.yaml to it matches the one in https://github.com/trondhindenes/openfga.dev/blob/main/docs/content/getting-started/setup-openfga.mdx

## Description
Makes docker-compose.yaml match the one in the getting started guide (https://github.com/trondhindenes/openfga.dev/blob/main/docs/content/getting-started/setup-openfga.mdx)

## References
- https://github.com/openfga/openfga.dev/pull/205

## Review Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -12,6 +12,8 @@ services:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=password
       - POSTGRES_DB=openfga
+    ports:
+      - "5432:5432"
     healthcheck:
       test: [ "CMD-SHELL", "pg_isready -U postgres" ]
       interval: 5s
@@ -46,8 +48,6 @@ services:
       - OPENFGA_LOG_FORMAT=json
     command: run
     depends_on:
-      db:
-        condition: service_started
       migrate:
         condition: service_completed_successfully
     networks:
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,40 +1,61 @@
 version: '3.8'
+
+networks:
+  openfga:
+
 services:
-  postgres:
+  db:
     image: postgres:14
-    container_name: postgres
     networks:
-      - default
-    ports:
-      - "5432:5432"
+      - openfga
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=password
+      - POSTGRES_DB=openfga
     healthcheck:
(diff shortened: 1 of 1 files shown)


rhamzeh (approved):

jon-whit (changes requested):
Why do we need the guard on `db.condition.service_started`? The `migrate` guard already covers our bases does it not?

If the migration service has run, then the postgres service is already healthy.

docker-compose.yaml
```diff
@@ -1,40 +1,61 @@
...
-    container_name: postgres
     networks:
-      - default
-    ports:
-      - "5432:5432"
+      - openfga
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=password
+      - POSTGRES_DB=openfga
```
craigpastro, line 14: If this is changed, this postgres container will no longer work with `make run-postgres`. I think it is nice to be able to `docker compose up -d postgres` and then `make run-postgres`.

craigpastro (changes requested):

docker-compose.yaml
```diff
@@ -1,40 +1,61 @@
...
     healthcheck:
       test: [ "CMD-SHELL", "pg_isready -U postgres" ]
       interval: 5s
       timeout: 5s
       retries: 5
 
+  pause:
```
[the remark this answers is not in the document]
trondhindenes: In my experience it's not completely robust. `pause` is definetely a hack, but in my experience a needed one.

docker-compose.yaml
trondhindenes: I'm not sure what that means. Was the default db name of `postgres` used in the past?

trondhindenes:
this PR has been stuck for a bit, I'm not quite sure what to do. It started with me just updating the corresponding compose-file in docs to a working example (https://github.com/openfga/openfga.dev/pull/205) and I was asked in the comment to make the same changes in this repo (which is the current PR). So I did that but it feels like a bit of can of worms. I guess the compose-file in docs and the one in this repo should be in sync, which they currently aren't. I also understand that changing the compose-file in this repo has other implications. So I guess I'm asking:

- should the compose-file in the docs repo and this one be kept in sync at all?
- if so, it's probably better if the docs one replicates this one instead of the other way around

PR #241, opened 2022-09-16
chore: add ReadStartingWithUser storage

jon-whit:
## Description
This adds the `ReverseReadTuples` stub to the storage interface to support reverse queries for reverse expansion use cases. This will help us improve the implementation of the ListObjects API by reverse expanding tuples to reveal objects the user has access to.

## Review Checklist


storage/test/tuples.go
```diff
@@ -457,3 +458,59 @@ func TuplePaginationOptionsTest(t *testing.T, datastore storage.OpenFGADatastore
 		}
 	})
 }
+
+func ReverseReadTuplesTest(t *testing.T, datastore storage.OpenFGADatastore) {
+
+	require := require.New(t)
```
jon-whit, line 464: This is a stupid simple test. We may want to add more, but this outlines the gist of it.

jon-whit:
Tests will fail until it is implemented. Feel free to help with the implementation for the memory and postgres storage adapters 👍

storage/storage.go
```diff
@@ -230,30 +230,75 @@ type TupleBackend interface {
...
+	) (*openfgapb.Tuple, error)
+
+	ReadUsersetTuples(
+		ctx context.Context,
+		store string,
+		tk *openfgapb.TupleKey,
+	) (TupleIterator, error)
+
+	// ReverseReadTuples performs a reverse read of relationship tuples starting at one or
+	// more user(s) or userset(s) and filtered by object type and relation.
```
miparnisari, line 268: The definition of Read API:

> // Read the set of tuples associated with `store` and `key`, which may be partially filled. A key must specify at
>	// least one of `Object` or `User` (or both), and may also optionally constrain by relation.

They both look very similar. 

Why is this one called "reverse"?

storage/test/tuples.go
```diff
@@ -457,3 +458,59 @@ func TuplePaginationOptionsTest(t *testing.T, datastore storage.OpenFGADatastore
...
+	require.NoError(err)
+
+	err = datastore.Write(
+		ctx,
+		storeID,
+		nil,
+		[]*openfgapb.TupleKey{
+			tuple.NewTupleKey("document:doc1", "viewer", "user:jon"),
+			tuple.NewTupleKey("document:doc2", "viewer", "group:eng#member"),
+			tuple.NewTupleKey("document:doc3", "editor", "user:jon"),
```
miparnisari, line 477: can we add another tuple here with a type that is not `document`?

storage/storage.go
jon-whit: As far as calling it "reverse" I don't think we need to keep that nomenclature if we don't want. But the query is really only intended for query relationship tuples starting at the user/userset and working backwards. It's meant to be indexed as such.

My understanding is that `Read` reads either all tuples for a single object (optionally constrained by a relation) or it reads all objects of any type for a user (optionally constrained by relation).

The goal here is to read all objects that a user or userset has a specific relation with without filtering by object_id whatsoever. As a simple query example,

`SELECT object_type, object_id FROM tuples WHERE store='mystore AND object_type='document' AND relation='viewer' AND user_type=X AND user=Y`

storage/test/tuples.go
jon-whit: Yeah, good point.

jon-whit: Updated.

jon-whit:
@miparnisari do you mind renaming this signature based on what we discussed yesterday? I agree that a name such as `ReadStartingWithUser` probably is more logical at this point (until we refactor our storage interface to be more concise).

storage/postgres/postgres.go
```diff
@@ -274,6 +275,28 @@ func (p *Postgres) ReadUsersetTuples(ctx context.Context, store string, tupleKey
 	return &tupleIterator{rows: rows}, nil
 }
 
+func (p *Postgres) ReadStartingWithUser(ctx context.Context, store string, opts storage.ReadStartingWithUserFilter) (storage.TupleIterator, error) {
+	ctx, span := p.tracer.Start(ctx, "postgres.ReadStartingWithUser")
+	defer span.End()
+
+	stmt := "SELECT store, object_type, object_id, relation, _user, ulid, inserted_at FROM tuple WHERE store = $1 AND object_type = $2 AND relation = $3 AND _user = any($4)"
```
miparnisari, line 282: https://github.com/jackc/pgx/issues/334

miparnisari: Also, this query is using the primary key:

```
Index Scan using tuple_pkey on tuple  (cost=0.15..8.17 rows=1 width=200)
"  Index Cond: ((store = '01GCJB1DJ88RJ1KA5TMG5PJXQ4'::text) AND (object_type = 'document'::text) AND (relation = 'owner'::text))"
"  Filter: (_user = ANY ('{maria,jon}'::text[]))"
```

storage/memory/memory.go
```diff
@@ -372,6 +372,40 @@ func (s *MemoryBackend) ReadUsersetTuples(ctx context.Context, store string, key
...
+	defer span.End()
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	var matches []*openfgapb.Tuple
+	for _, t := range s.tuples[store] {
+		if tupleUtils.GetType(t.Key.GetObject()) != filter.ObjectType {
+			continue
+		}
```
jon-whit, line 390: Can we add some whitespace between these separate blocks?

storage/postgres/postgres.go
```diff
@@ -275,6 +276,28 @@ func (p *Postgres) ReadUsersetTuples(ctx context.Context, store string, tupleKey
...
+	var targetUsersArg []string
+	for _, u := range opts.UserFilter {
+		targetUser := u.GetObject()
+		if u.GetRelation() != "" {
+			targetUser = strings.Join([]string{u.GetObject(), u.GetRelation()}, "#")
```
jon-whit, line 288: `fmt.Sprintf("%s#%s", ...)` makes it a little more obvious IMO

storage/test/tuples.go
```diff
@@ -457,3 +458,199 @@ func TuplePaginationOptionsTest(t *testing.T, datastore storage.OpenFGADatastore
...
+					break
+				}
+
+				require.Fail(err.Error())
+			}
+
+			objects = append(objects, tp.GetKey().GetObject())
+		}
+
+		require.Equal([]string{"document:doc1", "document:doc2"}, objects)
```
jon-whit, line 519: We don't necessarily guarantee order, so we should sort the two slices (expected and actual). This should be done here and elsewhere.

storage/test/tuples.go
```diff
@@ -457,3 +458,199 @@ func TuplePaginationOptionsTest(t *testing.T, datastore storage.OpenFGADatastore
...
+						Relation: "member",
+					},
+				},
+			},
+		)
+		require.NoError(err)
+
+		var objects []string
+		for {
+			tp, err := iter.Next()
```
jon-whit, line 507: This iterator code is used here and below. We should create a helper so we avoid code duplication.

storage/test/tuples.go
miparnisari: Good point!

jon-whit (commented):
LGTM! I can't approve because I created it though 😄

PR #247, opened 2022-09-21
chore: store the schema version

craigpastro:
## Description

Store the schema version in our datastore.

One thing I don't like there is that our signatures have grown a bit unwieldy. Do we prefer to combine the `version SchemaVersion, typeDefinitions []*openfgapb.TypeDefinition` into a `typeSystem *typesystem.Typesystem`. There is a bit of subtlety here as a `TypeSystem` contains a `map[string]*openfgapb.TypeDefinition` rather than an `[]*openfgapb.TypeDefinition`.

You know, we could even add the model id to the `TypeSystem`. Then, in checks, we simply read the entire model rather than a type at a time. WDYT?

## Review Checklist

### DIFF
--- a/assets/migrations/postgres/002_add_authorization_model_version.sql
+++ b/assets/migrations/postgres/002_add_authorization_model_version.sql
@@ -1,5 +1,5 @@
 -- +goose Up
-ALTER TABLE authorization_model ADD COLUMN schema_version INT DEFAULT 1;
+ALTER TABLE authorization_model ADD COLUMN schema_version INT NOT NULL DEFAULT 1;
 
 -- +goose Down
 ALTER TABLE authorization_model DROP COLUMN schema_version;
--- a/assets/migrations/postgres/002_add_authorization_model_version.sql
+++ b/assets/migrations/postgres/002_add_authorization_model_version.sql
@@ -1,0 +1,5 @@
+-- +goose Up
+ALTER TABLE authorization_model ADD COLUMN version INT DEFAULT 1;
+
+-- +goose Down
+ALTER TABLE authorization_model DROP COLUMN version;
--- a/assets/migrations/postgres/002_add_authorization_model_version.sql
+++ b/assets/migrations/postgres/002_add_authorization_model_version.sql
@@ -1,5 +1,5 @@
 -- +goose Up
-ALTER TABLE authorization_model ADD COLUMN version INT DEFAULT 1;
+ALTER TABLE authorization_model ADD COLUMN schema_version INT DEFAULT 1;
 
 -- +goose Down
-ALTER TABLE authorization_model DROP COLUMN version;
+ALTER TABLE authorization_model DROP COLUMN schema_version;
--- a/server/commands/write_authzmodel.go
+++ b/server/commands/write_authzmodel.go
@@ -35,12 +35,12 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 		return nil, serverErrors.ExceededEntityLimit("type definitions in an authorization model", w.backend.MaxTypesInTypeDefinition())
 	}
 
-	schemaVersion, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
+	version, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
 	if err != nil {
 		return nil, serverErrors.UnsupportedSchemaVersion
 	}
 
-	typeSystem := typesystem.NewTypeSystem(schemaVersion, req.GetTypeDefinitions())
+	typeSystem := typesystem.NewTypeSystem(version, req.GetTypeDefinitions())
 	if len(typeSystem.TypeDefinitions) != len(req.GetTypeDefinitions()) {
 		return nil, serverErrors.CannotAllowDuplicateTypesInOneRequest
 	}
@@ -55,7 +55,7 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 	}
 
 	utils.LogDBStats(ctx, w.logger, "WriteAuthzModel", 0, 1)
-	if err := w.backend.WriteAuthorizationModel(ctx, req.GetStoreId(), id, typeSystem.GetTypeDefinitions()); err != nil {
+	if err := w.backend.WriteAuthorizationModel(ctx, req.GetStoreId(), id, version, typeSystem.GetTypeDefinitions()); err != nil {
 		return nil, serverErrors.NewInternalError("Error writing authorization model configuration", err)
 	}
 
--- a/server/commands/write_authzmodel.go
+++ b/server/commands/write_authzmodel.go
@@ -35,12 +35,12 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 		return nil, serverErrors.ExceededEntityLimit("type definitions in an authorization model", w.backend.MaxTypesInTypeDefinition())
 	}
 
-	version, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
+	schemaVersion, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
 	if err != nil {
 		return nil, serverErrors.UnsupportedSchemaVersion
 	}
 
-	typeSystem := typesystem.NewTypeSystem(version, req.GetTypeDefinitions())
+	typeSystem := typesystem.NewTypeSystem(schemaVersion, req.GetTypeDefinitions())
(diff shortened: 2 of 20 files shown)


assets/migrations/postgres/002_add_authorization_model_version.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+ALTER TABLE authorization_model ADD COLUMN version INT DEFAULT 1;
```
jon-whit, line 2: schema_version instead ok?

server/commands/write_authzmodel.go
```diff
@@ -35,12 +35,12 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 		return nil, serverErrors.ExceededEntityLimit("type definitions in an authorization model", w.backend.MaxTypesInTypeDefinition())
 	}
 
-	schemaVersion, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
+	version, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
```
miparnisari, line 38: why this change? shcemaVersion is clear

storage/test/authz_models.go
```diff
@@ -199,7 +200,7 @@ func ReadTypeDefinitionTest(t *testing.T, datastore storage.OpenFGADatastore) {
 			},
 		}
 
-		err := datastore.WriteAuthorizationModel(ctx, store, modelID, []*openfgapb.TypeDefinition{expectedTypeDef})
+		err := datastore.WriteAuthorizationModel(ctx, store, modelID, typesystem.SchemaVersion1_0, []*openfgapb.TypeDefinition{expectedTypeDef})
```
miparnisari, line 203: is there a test that attempts to write a model with version 1_1?

server/commands/write_authzmodel.go
craigpastro: The function call lines are getting so long and I thought is was pretty clear because in the write auth model command, but I'll revert it for now.

assets/migrations/postgres/002_add_authorization_model_version.sql
craigpastro: Updated!

server/commands/write_authzmodel.go
```diff
@@ -55,7 +55,7 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 	}
 
 	utils.LogDBStats(ctx, w.logger, "WriteAuthzModel", 0, 1)
-	if err := w.backend.WriteAuthorizationModel(ctx, req.GetStoreId(), id, typeSystem.GetTypeDefinitions()); err != nil {
```
jon-whit, line 58: Can we just change this signature to write the AuthorizationModel itself? That'd be quite a bit nicer.

```
func ... WriteAuthorizationModel(ctx context.Context, model *openfgapb.AuthorizationModel) error {
    ...
}
```

server/test/write_authzmodel.go
```diff
@@ -15,6 +15,7 @@ import (
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
 
+// TODO: update these to SchemaVersion1_1 models
```
jon-whit, line 18: Can we just add an issue for this with a link to this line number and remove this comment here? That way we have some book keeping to come back around to. I'll forget about this TODO comment if it's not an issue :)

storage/postgres/postgres.go
```diff
@@ -325,7 +326,15 @@ func (p *Postgres) ReadAuthorizationModel(ctx context.Context, store string, mod
 		return nil, storage.ErrNotFound
 	}
 
+	// Update the schema version lazily
+	if version == typesystem.SchemaVersionUnspecified {
+		version = typesystem.SchemaVersion1_0
+		// Don't worry if we error, we'll update it lazily next time
+		_, _ = p.pool.Exec(ctx, "UPDATE authorization_model SET schema_version = $1 WHERE store = $2 AND authorization_model_id = $3", version, store, modelID)
```
jon-whit, line 333: We should still log the error in case we screwed something up in the statement itself, because it's possible this could always error and we'd never know about it.

storage/test/authz_models.go
jon-whit: Yeah, let's add a test to write and read a model v1.0 and model v1.1 so that we can assert the deserialized schema_version string get's cast to the appropriate enum.

server/test/read_authzmodel.go
```diff
@@ -2,18 +2,96 @@ package test
...
+					Relations: map[string]*openfgapb.Userset{
+						"reader": {
+							Userset: &openfgapb.Userset_This{},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:          "write and read an empty 1.1 model",
```
jon-whit, line 45: What do we mean by "empty 1.1 model"? Should this just be named "write and read a 1.1 model"?

server/test/read_authzmodel.go
```diff
@@ -2,18 +2,96 @@ package test
...
+	"github.com/openfga/openfga/pkg/typesystem"
 	"github.com/openfga/openfga/server/commands"
 	serverErrors "github.com/openfga/openfga/server/errors"
 	"github.com/openfga/openfga/storage"
 	"github.com/stretchr/testify/require"
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
 
+func TestSuccessfulReadAuthorizationModelQuery(t *testing.T, datastore storage.OpenFGADatastore) {
+	var tests = []struct {
```
jon-whit, line 18: I may have missed it, but if we don't have a test that writes a model without a version, can we test for that? 

Also, we should inject a row into the models table that has an empty schema_version string and make sure the read returns a v1.0 schema version.

These two tests will make sure we cover our bases.

server/test/read_authzmodel.go
craigpastro: Yes, that is a typo. Thanks. Fixed.

storage/postgres/postgres.go
craigpastro: Good point. Thanks. Updated.

server/test/read_authzmodel.go
craigpastro: We can't do the first one because of the non null constraint.

server/commands/write_authzmodel.go
craigpastro: I thought about that, but we lose the typing on `SchemaVersion` if we do that as the schema version in openfgapb.AuthorizationModel is a string.

server/test/read_authzmodel.go
craigpastro: Test added for the second case. It is Postgres specific.

storage/postgres/postgres.go
```diff
@@ -325,7 +326,18 @@ func (p *Postgres) ReadAuthorizationModel(ctx context.Context, store string, mod
 		return nil, storage.ErrNotFound
 	}
 
+	// Update the schema version lazily if it is not a valid typesystem.SchemaVersion.
+	if version != typesystem.SchemaVersion1_0 && version != typesystem.SchemaVersion1_1 {
+		version = typesystem.SchemaVersion1_0
+		_, err = p.pool.Exec(ctx, "UPDATE authorization_model SET schema_version = $1 WHERE store = $2 AND authorization_model_id = $3", version, store, modelID)
+		if err != nil {
+			// Don't worry if we error, we'll update it lazily next time, but let's log:
+			p.logger.Warn("failed to lazily update schema version", zap.String("store", store), zap.String("authorization_model_id", modelID))
```
jon-whit, line 335: Let's set this at the Error level, instead of warning. This shouldn't fail, and if it does we should know about it.

server/commands/write_authzmodel.go
jon-whit: Take the `*openfgapb.AuthorizationModel` and convert to the underlying serialized type (the enum) inside of the storage layer. I believe this is a decent approach since we're storing a string in a enum specific way within the storage layer.

storage/postgres/postgres.go
craigpastro: That's pretty serious. Won't that wake up the on-call primary? Is there a way to log an error and not wake people up?

server/commands/write_authzmodel.go
craigpastro: I like this idea and I am all for it, but there are millions of changes due to tests. I can do this now, or in a focused follow-up PR. WDYT?

miparnisari: refactor later ™️

miparnisari (approved):

jon-whit (approved):

PR #249, opened 2022-09-22
chore: add `graph` package stub to be implemented for reverse expand

jon-whit:
## Description
This PR adds signature stubs and some tests for a new `graph` package which will be used to introspect paths of connectivity between types that are connected with one another through one or more direct or indirect paths in the graph of relationships.

The graph package will be used in our implementation of reverse expansion which we'll use to optimize our [ListObjects API](https://openfga.dev/api/service#/Relationship%20Queries/ListObjects). 

A `RelationshipIngress` defines a relationship reference (e.g. `objectType#relation` pair) that is either directly or indirectly connected to another relationship reference. That is, it defines an incoming edge or possible path of connectivity between a source object reference and a target user reference.

The method `RelationshipIngresses` computes all of the relationship references that can be visited between a source object reference and a target user reference. For example, given the following authorization model:

```
type user

type group
  relations

type document
  relations
    define owner: [user]
    define editor: [user, group#member] as self or owner
    define viewer: [user, group#member] as self or editor
```
then `RelationshipIngresses(document#viewer, user) --> [document#viewer, document#editor, document#owner, group#member]` which are the paths of connectivity between the source document#viewer relationship and the target user object type.

We'll use these paths of connectivity to guide our reverse expansion process by pruning unreachable relationship paths as we expand in reverse.

## Review Checklist


internal/graph/graph_test.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			typesys := typesystem.NewTypeSystem(typesystem.SchemaVersion1_1, test.model.TypeDefinitions)
+
+			objGraph := BuildConnectedObjectGraph(typesys)
+
+			ingresses, err := objGraph.RelationshipIngresses(test.targetUserRef, test.targetObjectRef)
```
miparnisari, line 340:
```suggestion
			ingresses, err := objGraph.RelationshipIngresses(test.target, test.source)
```

internal/graph/graph_test.go
```diff
@@ -0,0 +1,346 @@
...
+	tests := []struct {
+		name              string
+		model             *openfgapb.AuthorizationModel
+		targetUserRef     *openfgapb.RelationReference
+		targetObjectRef   *openfgapb.RelationReference
+		expectedIngresses []*RelationshipIngress
+		expectedError     error
+	}{
+		{
+			name: "Direct ingress through ComputedUserset with multiple type restrictions",
```
miparnisari, line 22: https://dot-to-ascii.ggerganov.com/

input

```
digraph {
    
    "document editor" -> "user"
    "document viewer" -> "document editor"
    "document viewer" -> "user"
}
```

output

```
+-----------------+
| document viewer | -+
+-----------------+  |
  |                  |
  |                  |
  v                  |
+-----------------+  |
| document editor |  |
+-----------------+  |
  |                  |
  |                  |
  v                  |
+-----------------+  |
|      user       | <+
+-----------------+
```

it would be nice to have this as comment for each test

jon-whit: dot graphs are cool, and this makes it even easier to use in our document and/or comments in code.

What we really should do is be able to generate a digraph structure from an authorization model.

e.g.

```
type user

type group
  relations

type document
  relations
    define editor: [user, group#member]
    define viewer: [user] as editor
```
then in our graph package we can parse the model and export a D
```
package graph

import (
  ...
  "github.com/emicklei/dot"
)

func ToDigraph(model *openfgapb.AuthorizationModel) (*dot.Graph, error) {
    digraph := dot.NewGraph(dot.Directed)
    // todo: construct the digraph from the model

    return digraph, nil
}
```

Then we can use the Digraph structure and print it out and run it through https://dot-to-ascii.ggerganov.com/ like you did 😄 . That would be awesome 👍 !

internal/graph/graph_test.go
```diff
@@ -0,0 +1,546 @@
...
+						},
+					},
+				},
+			},
+			targetUserRef:     typesystem.RelationReference("user", ""),
+			targetObjectRef:   typesystem.RelationReference("team", "member"),
+			expectedIngresses: []*RelationshipIngress{},
+		},
+		{
+			name: "",
```
jon-whit, line 333: Not sure the best name for this? Any recommendations 😄 ?

internal/graph/graph_test.go
```diff
@@ -0,0 +1,546 @@
...
+					Ingress: typesystem.RelationReference("folder", "viewer"),
+				},
+				{
+					Type:    DirectIngress,
+					Ingress: typesystem.RelationReference("group", "member"),
+				},
+			},
+		},
+		{
+			name: "",
```
jon-whit, line 402: Not sure the best name for this? Any recommendations 😄 ?

internal/graph/graph_test.go
```diff
@@ -0,0 +1,546 @@
...
+			targetObjectRef: typesystem.RelationReference("document", "viewer"),
+			expectedIngresses: []*RelationshipIngress{
+				{
+					Type:    DirectIngress,
+					Ingress: typesystem.RelationReference("folder", "viewer"),
+				},
+			},
+		},
+		{
+			name: "",
```
jon-whit, line 467: Not sure the best name for this? Any recommendations 😄 ?

internal/graph/graph.go
```diff
@@ -0,0 +1,209 @@
...
+// 2. If computed userset, say `define viewer as writer`, then recurse on `document#writer, folder#viewer`.
+// 3. If tuple-to-userset, say, viewer from parent. Go to parent and find its related types.
+//  1. If parent's types includes `folder` type, and `folder` contains `viewer` relation then this is exactly a ttu rewrite.
+//  2. Otherwise, suppose the types contains `objectType` which has a relation `viewer`, then recurse on `objectType#viewer, folder#viewer`
+func (g *ConnectedObjectGraph) RelationshipIngresses(target *openfgapb.RelationReference, source *openfgapb.RelationReference) ([]*RelationshipIngress, error) {
+	return g.relationshipIngressesHelper(target, source, map[string]struct{}{})
+}
+
+func (g *ConnectedObjectGraph) relationshipIngressesHelper(target *openfgapb.RelationReference, source *openfgapb.RelationReference, visited map[string]struct{}) ([]*RelationshipIngress, error) {
+	key := createKey(target.GetType(), target.GetRelation())
```
jon-whit, line 80: No need to call a function just to call a function. If the function were doing more I'd say it may be warranted.

Let's just do `fmt.Sprintf("%s#%s", target.GetType(), target.GetRelation())` here instead.

internal/graph/graph.go
```diff
@@ -0,0 +1,209 @@
...
+	if _, ok := visited[key]; ok {
+		// We've already visited the target so no need to do so again.
+		return nil, nil
+	}
+	visited[key] = struct{}{}
+
+	relation, ok := g.typesystem.GetRelation(target.GetType(), target.GetRelation())
+	if !ok {
+		return nil, ErrTargetError
```
jon-whit, line 89: It'd be nice if we returned a more specific error indicating the target object type and/or relation that didn't exist. That way we can track it down more easily if we ever encounter this error.

internal/graph/graph.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	relatedUserTypes := relation.GetTypeInfo().GetDirectlyRelatedUserTypes()
+
+	var res []*RelationshipIngress
+
+	if contains(relatedUserTypes, source) {
```
jon-whit, line 95: This should be a helper in the typesystem package.  For example,

```
func (t *TypeSystem) IsDirectlyRelated(target *openfgapb.RelationReference, source *openfgapb.RelationReference) bool {...}
```

then this code could read a little easier with:

```
for _, targetObjRef := range targetObjRelatedTypes {
    if ok := typesys.IsDirectlyRelated(targetObjRef, sourceUserRef); ok {
        res = append(res, ...)
    }
}
```

internal/graph/graph.go
```diff
@@ -0,0 +1,209 @@
...
+							Type:     relatedUserType.GetType(),
+							Relation: relatedUserType.GetRelation(),
+						},
+					})
+				}
+			}
+		}
+	}
+
+	ingresses, err := g.relationshipIngressesResolveRewrite(target, source, relation.GetRewrite(), visited)
```
jon-whit, line 121: The block of code above this line is specific to the `case *openfgapb.Userset_This:` case in the switch statement in the helper function that takes the rewrite.

Can we refactor slightly so that case is not treated differently than the other cases in the switch statement. That will make the code more uniform.

internal/graph/graph.go
```diff
@@ -0,0 +1,209 @@
...
+// 1. If `folder#viewer` is a relational type then add `folder#viewer, direct` to the result.
+// 2. If computed userset, say `define viewer as writer`, then recurse on `document#writer, folder#viewer`.
+// 3. If tuple-to-userset, say, viewer from parent. Go to parent and find its related types.
+//  1. If parent's types includes `folder` type, and `folder` contains `viewer` relation then this is exactly a ttu rewrite.
+//  2. Otherwise, suppose the types contains `objectType` which has a relation `viewer`, then recurse on `objectType#viewer, folder#viewer`
+func (g *ConnectedObjectGraph) RelationshipIngresses(target *openfgapb.RelationReference, source *openfgapb.RelationReference) ([]*RelationshipIngress, error) {
+	return g.relationshipIngressesHelper(target, source, map[string]struct{}{})
+}
+
+func (g *ConnectedObjectGraph) relationshipIngressesHelper(target *openfgapb.RelationReference, source *openfgapb.RelationReference, visited map[string]struct{}) ([]*RelationshipIngress, error) {
```
jon-whit, line 79: mind if we rename this to simply `func (g *ConnectedObjectGraph) ingresses(...)` and the one below to simply `ingressesWithRewrite(...)`

internal/graph/graph.go
```diff
@@ -0,0 +1,209 @@
...
+func BuildConnectedObjectGraph(typesystem *typesystem.TypeSystem) *ConnectedObjectGraph {
+	return &ConnectedObjectGraph{
+		typesystem: typesystem,
+	}
+}
+
+// RelationshipIngresses computes the incoming edges (ingresses) that are possible between the source object and
+// relation and the target user (user or userset).
+//
+// To look up Ingresses(`document#viewer`, `source`), where `source` is an object type with no relation, find the rewrites and types of viewer in the document type:
```
miparnisari, line 63: "// To look up Ingresses(`document#viewer`, `source`).."

on what model?

internal/graph/graph.go
miparnisari: +1

internal/graph/graph_test.go
```diff
@@ -0,0 +1,647 @@
...
+			typesys := typesystem.New(test.model)
+
+			g := BuildConnectedObjectGraph(typesys)
+
+			ingresses, err := g.RelationshipIngresses(test.target, test.source)
+			require.NoError(t, err)
+
+			//for _, i := range ingresses {
+			//	fmt.Println(">>>", i)
+			//}
```
miparnisari, line 642: remove

internal/graph/graph_test.go
miparnisari: it's hard to come up with a name unless you understand the model, and in its current form the model is hard to understand... @craigpastro can you put a comment on each test with the DSL representation?

internal/graph/graph_test.go
```diff
@@ -0,0 +1,681 @@
...
+			source: typesystem.RelationReference("user", ""),
+			expected: []*RelationshipIngress{
+				{
+					Type:    DirectIngress,
+					Ingress: typesystem.RelationReference("folder", "viewer"),
+				},
+			},
+		},
+		{
+			name: "Nested group membership returns only top-level ingress",
```
jon-whit, line 630: @craigpastro I added this test just now. We're currently failing this test.

```
type user

type group
  relations
```

`RelationshipIngresses(group#member, user)` currently returns `[group#member, group#member]`, but it should only return one occurrence of `group#member`.

internal/graph/graph.go
craigpastro: Good point. Done. Thanks!

internal/graph/graph.go
craigpastro: I can do this indeed.

What do you recommend in this situation. I can make a function which returns an error, but then we lose any ability to match against it (except for the string representation). Or we can create an error struct which we can match against. Both of these are not the greatest, which is why I try to avoid them both, but, yeah, it doesn't provide much context to the user.

internal/graph/graph.go
craigpastro: This was useful. Thank you! Done.

internal/graph/graph.go
craigpastro: Moved. Thank you!

internal/graph/graph.go
craigpastro: Great minds 😆  Before I even saw this comment I renamed it to `findIngresses` and `findIngressesWithRewrite`. Is that okay?

internal/graph/graph_test.go
craigpastro: Whoops. Thank you! Fixed.

internal/graph/graph_test.go
craigpastro: Squashed. Thanks!

internal/graph/graph_test.go
jon-whit: Tracking this here https://github.com/openfga/openfga/issues/277

internal/graph/graph.go
jon-whit: I would recommend we have an `ErrTypeNotFound` and `ErrRelationNotFound` error in the `typesystem` package and then this code would

```
relation, err := g.typesystem.GetRelation(...)
if err != nil {
    if err == ErrTypeNotFound {
        return nil, errors.Wrap(err, fmt.Sprintf("'%s' undefined", target.GetType()))
    }

    if err == ErrRelationNotFound {
        return nil, errors.Wrap(err, fmt.Sprintf("'%s' undefined", target.GetRelation())
    }
}
```

jon-whit: Let's create an issue for this and come back around to it in another PR after this.

craigpastro (dismissed):

internal/graph/graph_test.go
craigpastro: Test1!

craigpastro (approved):

PR #251, opened 2022-09-23
refactor: write auth models to take an auth model

craigpastro:
## Description

Simplify the interface to writeAuthorizationModel to simply take an authorization model. 

Also refactor the check tests by splitting them into separate test functions. One function was trying to do too much. Use github model, don't use github model, write tuples, don't write tuples, use contextual tuples, don't use contextual tuples. omg.

Also found that we were not even running one of the check tests 😱 

## Review Checklist


storage/caching/caching_test.go
```diff
@@ -68,7 +70,7 @@ func TestReadTypeDefinition(t *testing.T) {
 				continue
 			}
 
-			cacheKey := strings.Join([]string{test.store, modelID, test.name}, Separator)
+			cacheKey := strings.Join([]string{test.store, test.model.Id, test.name}, Separator)
```
jon-whit, line 73: Can we join this with `fmt.Sprintf(...)` instead? It's less obvious the resulting string using `strings.Join(..)`.

storage/test/authz_models.go
```diff
@@ -49,50 +49,57 @@ func TestWriteAndReadAuthorizationModel(t *testing.T, datastore storage.OpenFGAD
 		),
 	}
 
-	if diff := cmp.Diff(expectedModel, model.TypeDefinitions, cmpOpts...); diff != "" {
+	if diff := cmp.Diff(got.TypeDefinitions, model.TypeDefinitions, cmpOpts...); diff != "" {
```
jon-whit, line 52: Let's just compare the returned model against the provided model in their entirety, instead of checking only the type defs.

Every field of `got` and `model` should be equal, and so we may as well compare them directly. It will increase our test confidence by doing this as well, because we'll be making sure the fields match in their entirety.

storage/caching/caching_test.go
craigpastro: I can do this, but it is quite a bit worse honestly:
```
cacheKey := fmt.Sprintf("%s%s%s%s%s", test.store, separator, test.model.Id, separator, test.name)
```
You still like it?

storage/test/authz_models.go
craigpastro: Good call. Updated. Thanks!

storage/caching/caching_test.go
jon-whit: Oh, sorry I don't think I was clear. Variablizing the separator and using strings.Join just adds indirection and makes it hard to understand the resulting key.

I was proposing `fmt.Sprintf("%s:%s:%s", test.store, test.model.Id, test.name)`, which is better IMO to understand.

craigpastro: Updated to remove the separator indirection as discussed offline:
```
cacheKey := fmt.Sprintf("%s:%s:%s", test.store, test.model.Id, test.name)
```

jon-whit (approved):

PR #253, opened 2022-09-26
chore: reorganize tests

miparnisari:
- Move `ListObjectsByTypeTest` from one file to another


storage/test/authz_models.go
```diff
@@ -15,7 +15,7 @@ import (
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
 
-func TestWriteAndReadAuthorizationModel(t *testing.T, datastore storage.OpenFGADatastore) {
+func WriteAndReadAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastore) {
```
craigpastro, line 18: All of our other tests start with `Test`. Do we want to change this? I actually don't mind because it would make my IDE happy, and what we have now doesn't quite fit into the Go `TestFoo(t *testing.T)` style. But we should be consistent I guess.

miparnisari: The format i'm establishing is 

```
t.Run("Test....", func(t *testing.T) { Test....(t, ds) })
```

image

craigpastro: Oh, I see. I like it.

craigpastro (approved):
LGTM

storage/test/authz_models.go
jon-whit: I'm not sure I understand. Your IDE shouldn't recognize these anyways because they aren't standard Go tests. A Go test has signature `func TestMyFunc(t *testing.T)`

However, since we pass a storage interface, these don't match the Go test signature, and so naming then `TestMyFunc` or `MyFuncTest` is irrelevant is it not? The reason I established the pattern of `MyFuncTest` is to that it was not to be confused with the actual Go test that invokes it at the top level.

Does this change make your IDE discover them better somehow? I'm not sure I understand.

craigpastro: If it starts with Test and is not of the form `TestFoo(t *testing.T)` and is in a package called test, GoLand doesn't like it. But, yes, I agree, if it is not a test that is invoked at the top level it is nicer to not call it TestSomething.

miparnisari: > so naming them TestMyFunc or MyFuncTest is irrelevant is it not?

correct

PR #256, opened 2022-09-28
fix streaming list objects API returning more objects than it should

miparnisari:
When we read the value of `objectsFound` and then increment it, we need to put a lock around this entire block of code, otherwise a data race can occur.

Alternatively, read and increment `objectsFound` in one operation :D

### DIFF
--- a/server/test/list_objects.go
+++ b/server/test/list_objects.go
@@ -187,7 +187,7 @@ func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObje
 			require.ErrorIs(t, err, test.expectedError)
 
 			if len(streamedObjectIds) > defaultListObjectsMaxResults {
-				t.Errorf("expected a maximum of %d results but got %d:", defaultListObjectsMaxResults, len(streamedObjectIds))
+				t.Errorf("expected a maximum of %d results but got %d", defaultListObjectsMaxResults, len(streamedObjectIds))
 			}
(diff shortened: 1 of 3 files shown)


server/test/list_objects.go
```diff
@@ -187,7 +187,7 @@ func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObje
 			require.ErrorIs(t, err, test.expectedError)
 
 			if len(streamedObjectIds) > defaultListObjectsMaxResults {
-				t.Errorf("expected a maximum of %d results but got %d:", defaultListObjectsMaxResults, len(streamedObjectIds))
+				t.Errorf("expected a maximum of %d results but got %d", defaultListObjectsMaxResults, len(streamedObjectIds))
```
jon-whit, line 190: Can we please make this a one liner using `require.LessOrEqual(len(streamedObjectIds), defaultListObjectsMaxResults)`?

server/test/list_objects.go
```diff
@@ -187,7 +187,7 @@ func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObje
...
 			}
 			if !subset(streamedObjectIds, test.expectedResult) {
```
jon-whit, line 192: Can we please use `require.Subset` here instead? No need to maintain the extra `subset` method here.

craigpastro (approved):
LGTM

PR #257, opened 2022-09-28
refactor: store version as a string

craigpastro:
## Description

- Store auth model version as a string in storage
- Validate is now a standalone function
- typesystem.New now just takes an AuthorizationModel
- ReadAuthorizationModel added to cache backend

## Review Checklist


server/errors/errors.go
```diff
@@ -16,19 +16,16 @@ const InternalServerErrorMsg = "Internal Server Error"
 
 var (
 	// AuthorizationModelResolutionTooComplex is used to avoid stack overflows
-	AuthorizationModelResolutionTooComplex     = status.Error(codes.Code(openfgapb.ErrorCode_authorization_model_resolution_too_complex), "Authorization Model resolution required too many rewrite rules to be resolved. Check your authorization model for infinite recursion or too much nesting")
```
craigpastro, line 19: Formatting due to removing unused errors

pkg/typesystem/typesystem.go
```diff
@@ -108,45 +81,75 @@ func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relati
...
+// If it is a 1.1 authorization model (with types on relations), then additionally validate the type system
 // according to the following rules:
 //  3. Every type on a relation must be a valid type:
 //     a. For a type (e.g. user) this means checking that this type is in the TypeSystem
 //     b. For a type#relation this means checking that this type with this relation is in the TypeSystem
 //  4. Check that a relation is assignable if and only if it has a non-zero list of types
-func (t *TypeSystem) Validate() error {
-	if err := t.validateRelationRewrites(); err != nil {
-		return err
+func Validate(model *openfgapb.AuthorizationModel) (*openfgapb.AuthorizationModel, error) {
```
miparnisari, line 99: This is a weird signature. I wouldn't expect a `Validate` function to return the thing it is validating with some fields changed.

pkg/typesystem/typesystem.go
```diff
@@ -108,45 +81,75 @@ func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relati
...
+	default:
+		return nil, ErrInvalidSchemaVersion
 	}
 
-	if t.Version == SchemaVersion1_1 {
-		if err := t.validateRelationTypeRestrictions(); err != nil {
-			return err
+	if containsDuplicateType(model) {
+		return nil, ErrDuplicateTypes
+	}
```
miparnisari, line 110:
```suggestion
	if err := containsDuplicateType(model); err != nil {
		return nil, err
	}
```

pkg/typesystem/typesystem.go
craigpastro: Yeah, this is a good point. I'll refactor this. Thanks!

craigpastro: Updated!

pkg/typesystem/typesystem.go
craigpastro: `containsDuplicateType` returns a boolean

miparnisari: I know, but to be consistent with the other validation methods, it could return an error

pkg/typesystem/typesystem.go
```diff
@@ -7,53 +7,23 @@ import (
...
-		return "1.1"
-	default:
-		return "unspecified"
-	}
-}
+var ErrDuplicateTypes = errors.New("an authorization model cannot contain duplicate types")
 
 type TypeSystem struct {
-	Version         SchemaVersion
+	Version         string
```
jon-whit, line 13: This and `TypeDefinitions` don't need to be publicly exported. We want to encourage usage of the methods themselves instead of indexing into the fields of the TypeSystem.

pkg/typesystem/typesystem.go
```diff
@@ -108,45 +78,65 @@ func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relati
...
 //     b. Be valid relations on another existing type (in cases of tupleToUserset)
-//  2. Do not allow duplicate types or duplicate relations (but that is inherent in the map structure so nothing to
-//     actually check)
+//  2. Do not allow duplicate types or duplicate relations (only need to check types as relations are
+//     in a map so cannot contain duplicates)
 //
-// If it is a SchemaVersion1_1 type system (with types on relations), then additionally validate the type system
+// If it is a 1.1 authorization model (with types on relations), then additionally validate the type system
 // according to the following rules:
 //  3. Every type on a relation must be a valid type:
```
jon-whit, line 90: Every type restriction on a relation must be a valid type

server/commands/write_authzmodel.go
```diff
@@ -30,40 +30,37 @@ func NewWriteAuthorizationModelCommand(
...
 	// Until this is solved: https://github.com/envoyproxy/protoc-gen-validate/issues/74
-	if len(typeDefinitions) > w.backend.MaxTypesInTypeDefinition() {
+	if len(req.GetTypeDefinitions()) > w.backend.MaxTypesInTypeDefinition() {
 		return nil, serverErrors.ExceededEntityLimit("type definitions in an authorization model", w.backend.MaxTypesInTypeDefinition())
 	}
 
-	schemaVersion, err := typesystem.NewSchemaVersion(req.GetSchemaVersion())
-	if err != nil {
+	switch req.SchemaVersion {
+	case "", "1.0":
```
jon-whit, line 39: Can we put this validation logic as part of the logic of `typesystem.Validate(model)` instead? 

Makes our separation of concerns more well organized.

storage/test/authz_models.go
```diff
@@ -15,48 +15,40 @@ import (
...
-						},
-					},
-				},
-			},
-		},
-	}
+	t.Run("write, then read, succeeds", func(t *testing.T) {
+		model := &openfgapb.AuthorizationModel{
+			Id:              id.Must(id.New()).String(),
+			SchemaVersion:   "1.0",
```
jon-whit, line 23: One thing we may consider doing is making these const in the `typesystem` package.

```
package typesystem

const (
	SchemaVersion_1_0 = "1.0"
	SchemaVersion_1_1 = "1.1"
)

```

storage/test/authz_models.go
```diff
@@ -15,48 +15,40 @@ import (
...
-	cmpOpts := []cmp.Option{
-		cmpopts.IgnoreUnexported(
-			openfgapb.AuthorizationModel{},
-			openfgapb.TypeDefinition{},
-			openfgapb.Userset{},
-			openfgapb.Userset_This{},
-			openfgapb.DirectUserset{},
-		),
-	}
+		cmpOpts := []cmp.Option{
```
jon-whit, line 33: We use these same compare options in multiple places. Can we just define this as a global in the `test/storage.go` . We can add every type that we want the unexported fields to ignore to the list and reuse it everywhere we compare protobuf structs.

storage/postgres/postgres_test.go
```diff
@@ -32,7 +31,7 @@ func TestReadAuthorizationModelPostgresSpecificCases(t *testing.T) {
 	ctx := context.Background()
 	store := "store"
 	modelID := "foo"
-	schemaVersion := 4 // which is not an enum value... yet.
+	schemaVersion := "7.8"
 
 	bytes, err := proto.Marshal(&openfgapb.TypeDefinition{Type: "document"})
```
jon-whit, line 36: We use `protojson.Marshal` in some places and this uses `proto.Marshal` in others. We should confirm they are identical and strive to use only one. We can save that for a future PR, but let's make sure and add an issue to verify and refactor.

server/commands/write.go
```diff
@@ -117,20 +117,14 @@ func (c *WriteCommand) validateTypesForTuple(authModel *openfgapb.AuthorizationM
...
-	ts := typesystem.NewTypeSystem(sv, authModel.TypeDefinitions)
+	ts := typesystem.New(authModel)
 
 	typeDefinitionForObject := ts.TypeDefinitions[objectType]
 	typeDefinitionForUser := ts.TypeDefinitions[userType]
 
 	relationsForObject := typeDefinitionForObject.GetMetadata().GetRelations()
 	if relationsForObject == nil {
-		if authModel.SchemaVersion == "1.1" {
+		if ts.Version == "1.1" {
```
jon-whit, line 127: `ts.GetSchemaVersion()` let's strive not to access direct fields of the Typesystem and use the helpers methods instead. That way we can avoid people indexing into the typedefs etc..

pkg/typesystem/typesystem.go
```diff
@@ -108,45 +78,65 @@ func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relati
...
-//     a. Be valid relations on the same type in the authorization typeSystem (in cases of computedUserset)
+//     a. Be valid relations on the same type in the authorization model (in cases of computedUserset)
 //     b. Be valid relations on another existing type (in cases of tupleToUserset)
-//  2. Do not allow duplicate types or duplicate relations (but that is inherent in the map structure so nothing to
-//     actually check)
+//  2. Do not allow duplicate types or duplicate relations (only need to check types as relations are
+//     in a map so cannot contain duplicates)
 //
-// If it is a SchemaVersion1_1 type system (with types on relations), then additionally validate the type system
+// If it is a 1.1 authorization model (with types on relations), then additionally validate the type system
```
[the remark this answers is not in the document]
craigpastro: Done. Thanks!

pkg/typesystem/typesystem.go
craigpastro: Done. Thanks!

server/commands/write_authzmodel.go
craigpastro: Done. Thanks!

storage/test/authz_models.go
craigpastro: Oh yeah. I didn't see that. Updated. Thanks!

server/commands/write.go
craigpastro: Done. Thanks!

storage/postgres/postgres_test.go
craigpastro: Addressed. Thanks!

storage/test/authz_models.go
craigpastro: Good idea. Done. Thanks!

pkg/typesystem/typesystem.go
craigpastro: I am not 100% sure I agree, but let's try it out. Done. Thanks for the comment!

pkg/typesystem/typesystem.go
```diff
@@ -7,72 +7,54 @@ import (
...
 }
 
-// TODO: In a future PR refactor WriteAuthorizationModel to take a map[string]*openfgapb.TypeDefinition
-// and get rid of this method
-func (t *TypeSystem) GetTypeDefinitions() []*openfgapb.TypeDefinition {
-	tds := make([]*openfgapb.TypeDefinition, 0, len(t.TypeDefinitions))
-	for _, td := range t.TypeDefinitions {
-		tds = append(tds, td)
-	}
+func (t *TypeSystem) GetVersion() string {
```
jon-whit, line 39: To not confuse this method with returning a version of the typesystem, could we name it schema_version? Schema version refers to the version of the model's schema whereas version is so generic I don't know if it refers to the type system's version or the model's version etc..

pkg/typesystem/typesystem.go
```diff
@@ -7,72 +7,54 @@ import (
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
 
-type SchemaVersion int32
-
 const (
-	SchemaVersionUnspecified SchemaVersion = 0
-	SchemaVersion1_0         SchemaVersion = 1
-	SchemaVersion1_1         SchemaVersion = 2
+	SchemaVersion10 = "1.0"
```
jon-whit, line 11: SchemaVersion # 10? Or 1.0?

I usually see Go separate with underscores in these scenarios. Underscores represent ".", since "." is not an acceptable characters for variable names.

craigpastro: GoLand hated it. It said "Use Camel case instead of snake case", but yeah, having the underscore is better. Fixed. Thanks.

pkg/typesystem/typesystem.go
craigpastro: Good point. Updated. Thanks!

assets/migrations/mysql/002_add_authorization_model_version.sql
```diff
@@ -1,5 +1,5 @@
 -- +goose Up
-ALTER TABLE authorization_model ADD COLUMN schema_version INT NOT NULL DEFAULT 1;
+ALTER TABLE authorization_model ADD COLUMN schema_version VARCHAR(5) NOT NULL DEFAULT '1.0';
```
jon-whit, line 2: VARCHAR(5) doesn't give us a lot of room for growth. Is it worth making it a little more generous?

jon-whit (dismissed):

jon-whit (approved):

PR #276, opened 2022-10-11
feat: allow migrations to also read the config file

MidasLamb:
## Description

Allow the migrations to use the same config files as the running server. This way it allows re-using the same file for both which is a nice QoL improvement in case you're using the files.

## Review Checklist
- [ ] ~I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).~ Not possible for a pr coming from company fork.
- [ ] ~I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]~ Don't think this is necessary as it's already implied in the docs.

### DIFF
--- a/pkg/cmd/migrate.go
+++ b/pkg/cmd/migrate.go
@@ -29,6 +29,14 @@ func NewMigrateCommand() *cobra.Command {
 		RunE:  runMigration,
 	}
 
+	viper.SetConfigName("config")
+	viper.SetConfigType("yaml")
+
+	configPaths := []string{"/etc/openfga", "$HOME/.openfga", "."}
+	for _, path := range configPaths {
+		viper.AddConfigPath(path)
+	}
+
 	viper.SetEnvPrefix("OPENFGA")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	viper.AutomaticEnv()


pkg/cmd/migrate.go
```diff
@@ -29,6 +29,14 @@ func NewMigrateCommand() *cobra.Command {
 		RunE:  runMigration,
 	}
 
+	viper.SetConfigName("config")
```
jon-whit, line 32: I'd prefer we do this in way that shares this logic.  For example, something like

```
package config

func SetupConfig(v *viper.Viper) {
    ...
}
```
and then use `SetupConfig` in the `run` and the `migrate` CLI targets.

MidasLamb: I found the function you linked, but potential problem I saw with using that function is that it returns a specific `Config` struct, and I wasn't sure if the two commands should deal with the exact same config, or to consider the configurations merely as overlapping.
In other words: could it be confusing that in `run` (and potentially other commands in the future), the config has an option `migrateToVersionAndStop` or `DoAllMigrationsInOneBigTransaction`, because it is needed in the `migrate` command

jon-whit (approved):

craigpastro (approved):
LGTM

craigpastro:
Thank you for the contribution again!

PR #279, opened 2022-10-12
feat: optimized ListObjects with reverse expansion

jon-whit:
## Description
This PR introduces a new internal API called the `ConnectedObjects` API. This API is used to reverse expand relationship tuples between some source user and some target `object#relation`. We use this ConnectedObjects API in an improved version of ListObjects for any models other than those exhibiting intersection or exclusion rewrites (support for improved intersection and exclusion coming soon).

## References
https://github.com/openfga/roadmap/issues/8

## Review Checklist

### DIFF
--- a/pkg/typesystem/typesystem.go
+++ b/pkg/typesystem/typesystem.go
@@ -14,8 +14,12 @@ const (
 )
 
 var (
-	ErrDuplicateTypes       = errors.New("an authorization model cannot contain duplicate types")
-	ErrInvalidSchemaVersion = errors.New("invalid schema version")
+	ErrDuplicateTypes        = errors.New("an authorization model cannot contain duplicate types")
+	ErrInvalidSchemaVersion  = errors.New("invalid schema version")
+	ErrInvalidModel          = errors.New("invalid authorization model encountered")
+	ErrRelationUndefined     = errors.New("undefined relation")
+	ErrObjectTypeUndefined   = errors.New("undefined object type")
+	ErrInvalidUsersetRewrite = errors.New("invalid userset rewrite definition")
 )
 
 func RelationReference(objectType, relation string) *openfgapb.RelationReference {
@@ -121,10 +125,13 @@ func (t *TypeSystem) GetTypeDefinition(objectType string) (*openfgapb.TypeDefini
 	return nil, false
 }
 
-func (t *TypeSystem) GetRelations(objectType string) (map[string]*openfgapb.Relation, bool) {
+func (t *TypeSystem) GetRelations(objectType string) (map[string]*openfgapb.Relation, error) {
 	td, ok := t.typeDefinitions[objectType]
 	if !ok {
-		return nil, false
+		return nil, &ObjectTypeUndefinedError{
+			ObjectType: objectType,
+			Err:        ErrObjectTypeUndefined,
+		}
 	}
 
 	relations := map[string]*openfgapb.Relation{}
@@ -143,42 +150,231 @@ func (t *TypeSystem) GetRelations(objectType string) (map[string]*openfgapb.Rela
 		relations[relation] = r
 	}
 
-	return relations, true
+	return relations, nil
 }
 
-func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relation, bool) {
-	relations, ok := t.GetRelations(objectType)
-	if !ok {
-		return nil, false
+func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relation, error) {
+	relations, err := t.GetRelations(objectType)
+	if err != nil {
+		return nil, err
 	}
 
 	r, ok := relations[relation]
 	if !ok {
-		return nil, false
-	}
-
-	return r, true
-}
-
-func (t *TypeSystem) GetDirectlyRelatedUserTypes(objectType, relation string) []*openfgapb.RelationReference {
-	if r, ok := t.GetRelation(objectType, relation); ok {
-		return r.GetTypeInfo().GetDirectlyRelatedUserTypes()
-	}
-
-	return nil
-}
-
-// IsDirectlyRelated determines whether the type of the target RelationReference contains the source RelationReference.
-func (t *TypeSystem) IsDirectlyRelated(target *openfgapb.RelationReference, source *openfgapb.RelationReference) bool {
-	if relation, ok := t.GetRelation(target.GetType(), target.GetRelation()); ok {
-		for _, relationReference := range relation.GetTypeInfo().GetDirectlyRelatedUserTypes() {
-			if source.GetType() == relationReference.GetType() && source.GetRelation() == relationReference.GetRelation() {
-				return true
-			}
+		return nil, &RelationUndefinedError{
+			ObjectType: objectType,
+			Relation:   relation,
+			Err:        ErrRelationUndefined,
 		}
 	}
 
-	return false
+	return r, nil
+}
+
+func (t *TypeSystem) GetDirectlyRelatedUserTypes(objectType, relation string) ([]*openfgapb.RelationReference, error) {
+
+	r, err := t.GetRelation(objectType, relation)
+	if err != nil {
+		return nil, err
+	}
+
+	return r.GetTypeInfo().GetDirectlyRelatedUserTypes(), nil
+}
+
+// IsDirectlyRelated determines whether the type of the target RelationReference contains the source RelationReference.
+func (t *TypeSystem) IsDirectlyRelated(target *openfgapb.RelationReference, source *openfgapb.RelationReference) (bool, error) {
+
+	relation, err := t.GetRelation(target.GetType(), target.GetRelation())
+	if err != nil {
+		return false, err
+	}
+
+	for _, relationReference := range relation.GetTypeInfo().GetDirectlyRelatedUserTypes() {
+		if source.GetType() == relationReference.GetType() && source.GetRelation() == relationReference.GetRelation() {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (t *TypeSystem) HasTypeInfo(objectType, relation string) (bool, error) {
+	r, err := t.GetRelation(objectType, relation)
+	if err != nil {
+		return false, err
+	}
+
+	if t.GetSchemaVersion() == SchemaVersion1_1 && r.GetTypeInfo() != nil {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// RelationInvolvesIntersection returns true if the provided relation's userset rewrite
+// is defined by one or more direct or indirect intersections.
+func (t *TypeSystem) RelationInvolvesIntersection(objectType, relation string) (bool, error) {
+
+	rel, err := t.GetRelation(objectType, relation)
+	if err != nil {
+		return false, err
+	}
+
+	rewrite := rel.GetRewrite()
+
+	switch rw := rewrite.GetUserset().(type) {
+	case *openfgapb.Userset_This:
+		return false, nil
+	case *openfgapb.Userset_ComputedUserset:
+		rewrittenRelation := rw.ComputedUserset.GetRelation()
+		rewritten, err := t.GetRelation(objectType, rewrittenRelation)
+		if err != nil {
+			return false, err
+		}
+
+		return t.RelationInvolvesIntersection(objectType, rewritten.GetName())
+	case *openfgapb.Userset_TupleToUserset:
+		tupleset := rw.TupleToUserset.GetTupleset().GetRelation()
+		rewrittenRelation := rw.TupleToUserset.ComputedUserset.GetRelation()
+
+		tuplesetRel, err := t.GetRelation(objectType, tupleset)
+		if err != nil {
+			return false, err
+		}
+
+		directlyRelatedTypes := tuplesetRel.GetTypeInfo().GetDirectlyRelatedUserTypes()
+		for _, relatedType := range directlyRelatedTypes {
+			// must be of the form 'objectType' by this point since we disallow `tupleset` relations of the form `objectType:id#relation`
+			r := relatedType.GetRelation()
+			if r != "" {
+				return false, fmt.Errorf(
+					"invalid type restriction '%s#%s' specified on tupleset relation '%s#%s': %w",
+					relatedType.GetType(),
+					relatedType.GetRelation(),
+					objectType,
+					tupleset,
+					ErrInvalidModel,
+				)
+			}
+
+			rel, err := t.GetRelation(relatedType.GetType(), rewrittenRelation)
+			if err != nil {
+				if errors.Is(err, ErrObjectTypeUndefined) || errors.Is(err, ErrRelationUndefined) {
+					continue
+				}
+
+				return false, err
+			}
+
+			containsIntersection, err := t.RelationInvolvesIntersection(relatedType.GetType(), rel.GetName())
+			if err != nil {
+				return false, err
+			}
+
+			if containsIntersection {
+				return true, nil
+			}
+		}
+
+	case *openfgapb.Userset_Intersection:
+		return true, nil
+	case *openfgapb.Userset_Union:
+		for _, child := range rw.Union.GetChild() {
+			if RewriteContainsIntersection(child) {
+				return true, nil
+			}
+		}
+	case *openfgapb.Userset_Difference:
+		difference := rw.Difference
+		if RewriteContainsIntersection(difference.GetBase()) || RewriteContainsIntersection(difference.GetSubtract()) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+// RelationInvolvesExclusion returns true if the provided relation's userset rewrite
+// is defined by one or more direct or indirect exclusions.
+func (t *TypeSystem) RelationInvolvesExclusion(objectType, relation string) (bool, error) {
+	rel, err := t.GetRelation(objectType, relation)
+	if err != nil {
+		return false, err
+	}
+
+	rewrite := rel.GetRewrite()
+
+	switch rw := rewrite.GetUserset().(type) {
+	case *openfgapb.Userset_This:
+		return false, nil
+	case *openfgapb.Userset_ComputedUserset:
+		rewrittenRelation := rw.ComputedUserset.GetRelation()
+		rewritten, err := t.GetRelation(objectType, rewrittenRelation)
+		if err != nil {
+			return false, err
+		}
+
+		return t.RelationInvolvesExclusion(objectType, rewritten.GetName())
+	case *openfgapb.Userset_TupleToUserset:
+		tupleset := rw.TupleToUserset.GetTupleset().GetRelation()
+		rewrittenRelation := rw.TupleToUserset.ComputedUserset.GetRelation()
+
+		tuplesetRel, err := t.GetRelation(objectType, tupleset)
+		if err != nil {
+			return false, err
+		}
+
+		directlyRelatedTypes := tuplesetRel.GetTypeInfo().GetDirectlyRelatedUserTypes()
+		for _, relatedType := range directlyRelatedTypes {
+			// must be of the form 'objectType' by this point since we disallow `tupleset` relations of the form `objectType:id#relation`
+			r := relatedType.GetRelation()
+			if r != "" {
+				return false, fmt.Errorf(
+					"invalid type restriction '%s#%s' specified on tupleset relation '%s#%s': %w",
+					relatedType.GetType(),
+					relatedType.GetRelation(),
+					objectType,
+					tupleset,
+					ErrInvalidModel,
+				)
+			}
+
+			rel, err := t.GetRelation(relatedType.GetType(), rewrittenRelation)
+			if err != nil {
+				if errors.Is(err, ErrObjectTypeUndefined) || errors.Is(err, ErrRelationUndefined) {
+					continue
+				}
+
+				return false, err
+			}
+
+			containsExclusion, err := t.RelationInvolvesExclusion(relatedType.GetType(), rel.GetName())
+			if err != nil {
+				return false, err
+			}
+
+			if containsExclusion {
+				return true, nil
+			}
+		}
+
+	case *openfgapb.Userset_Intersection:
+		for _, child := range rw.Intersection.GetChild() {
+			if RewriteContainsExclusion(child) {
+				return true, nil
+			}
+		}
+	case *openfgapb.Userset_Union:
+		for _, child := range rw.Union.GetChild() {
+			if RewriteContainsExclusion(child) {
+				return true, nil
+			}
+		}
+	case *openfgapb.Userset_Difference:
+		return true, nil
+	}
+
+	return false, nil
 }
 
 // Validate validates an *openfgapb.AuthorizationModel according to the following rules:
@@ -287,24 +483,24 @@ func isUsersetRewriteValid(
 	rewrite *openfgapb.Userset,
 ) error {
 	if rewrite.GetUserset() == nil {
-		return InvalidRelationError(objectType, relation)
+		return &InvalidRelationError{ObjectType: objectType, Relation: relation, Cause: ErrInvalidUsersetRewrite}
 	}
 
 	switch t := rewrite.GetUserset().(type) {
 	case *openfgapb.Userset_ComputedUserset:
 		computedUserset := t.ComputedUserset.GetRelation()
 		if computedUserset == relation {
-			return InvalidRelationError(objectType, relation)
+			return &InvalidRelationError{ObjectType: objectType, Relation: relation, Cause: ErrInvalidUsersetRewrite}
 		}
 		if _, ok := relationsOnType[computedUserset]; !ok {
-			return RelationDoesNotExistError(objectType, computedUserset)
+			return &RelationUndefinedError{ObjectType: objectType, Relation: computedUserset, Err: ErrRelationUndefined}
 		}
 	case *openfgapb.Userset_TupleToUserset:
 		tupleset := t.TupleToUserset.GetTupleset().GetRelation()
 
 		tuplesetRelation, ok := relationsOnType[tupleset]
 		if !ok {
-			return RelationDoesNotExistError(objectType, tupleset)
+			return &RelationUndefinedError{ObjectType: objectType, Relation: tupleset, Err: ErrRelationUndefined}
 		}
 
 		// tupleset relations must only be direct relationships, no rewrites
@@ -316,7 +512,7 @@ func isUsersetRewriteValid(
 
 		computedUserset := t.TupleToUserset.GetComputedUserset().GetRelation()
 		if _, ok := allRelations[computedUserset]; !ok {
-			return RelationDoesNotExistError("", computedUserset)
+			return &RelationUndefinedError{ObjectType: "", Relation: computedUserset, Err: ErrRelationUndefined}
 		}
 	case *openfgapb.Userset_Union:
 		for _, child := range t.Union.GetChild() {
@@ -351,9 +547,9 @@ func validateRelationTypeRestrictions(model *openfgapb.AuthorizationModel) error
 	t := New(model)
 
 	for objectType := range t.typeDefinitions {
-		relations, ok := t.GetRelations(objectType)
-		if !ok {
-			return InvalidRelationError(objectType, "")
+		relations, err := t.GetRelations(objectType)
+		if err != nil {
+			return err
 		}
 
 		for name, relation := range relations {
@@ -372,12 +568,12 @@ func validateRelationTypeRestrictions(model *openfgapb.AuthorizationModel) error
 				relatedObjectType := related.GetType()
 				relatedRelation := related.GetRelation()
 
-				if _, ok := t.GetRelations(relatedObjectType); !ok {
+				if _, err := t.GetRelations(relatedObjectType); err != nil {
 					return InvalidRelationTypeError(objectType, name, relatedObjectType, relatedRelation)
 				}
 
 				if relatedRelation != "" {
-					if _, ok := t.GetRelation(relatedObjectType, relatedRelation); !ok {
+					if _, err := t.GetRelation(relatedObjectType, relatedRelation); err != nil {
 						return InvalidRelationTypeError(objectType, name, relatedObjectType, relatedRelation)
 					}
 				}
@@ -391,28 +587,30 @@ func validateRelationTypeRestrictions(model *openfgapb.AuthorizationModel) error
 func (t *TypeSystem) IsDirectlyAssignable(relation *openfgapb.Relation) bool {
 	rewrite := relation.GetRewrite()
 
-	return ContainsSelf(rewrite)
+	return RewriteContainsSelf(rewrite)
 }
 
-func ContainsSelf(rewrite *openfgapb.Userset) bool {
+// RewriteContainsSelf returns true if the provided userset rewrite
+// is defined by one or more self referencing definitions.
+func RewriteContainsSelf(rewrite *openfgapb.Userset) bool {
 	switch rw := rewrite.GetUserset().(type) {
 	case *openfgapb.Userset_This:
 		return true
 	case *openfgapb.Userset_Union:
 		for _, child := range rw.Union.GetChild() {
-			if ContainsSelf(child) {
+			if RewriteContainsSelf(child) {
 				return true
 			}
 		}
 	case *openfgapb.Userset_Intersection:
 		for _, child := range rw.Intersection.GetChild() {
-			if ContainsSelf(child) {
+			if RewriteContainsSelf(child) {
 				return true
 			}
 		}
 	case *openfgapb.Userset_Difference:
 		difference := rw.Difference
-		if ContainsSelf(difference.GetBase()) || ContainsSelf(difference.GetSubtract()) {
+		if RewriteContainsSelf(difference.GetBase()) || RewriteContainsSelf(difference.GetSubtract()) {
 			return true
 		}
 	}
@@ -420,22 +618,95 @@ func ContainsSelf(rewrite *openfgapb.Userset) bool {
 	return false
 }
 
-func InvalidRelationError(objectType, relation string) error {
-	return fmt.Errorf("the definition of relation '%s' in object type '%s' is invalid", relation, objectType)
-}
-
-func ObjectTypeDoesNotExistError(objectType string) error {
-	return fmt.Errorf("object type '%s' does not exist", objectType)
-}
-
-// RelationDoesNotExistError may have an empty objectType, but must have a relation
-// (otherwise the error won't make much sense).
-func RelationDoesNotExistError(objectType, relation string) error {
-	msg := fmt.Sprintf("relation '%s'", relation)
-	if objectType != "" {
-		msg = fmt.Sprintf("%s in object type '%s'", msg, objectType)
+// RewriteContainsIntersection returns true if the provided userset rewrite
+// is defined by one or more direct or indirect intersections.
+func RewriteContainsIntersection(rewrite *openfgapb.Userset) bool {
+	switch rw := rewrite.GetUserset().(type) {
+	case *openfgapb.Userset_Intersection:
+		return true
+	case *openfgapb.Userset_Union:
+		for _, child := range rw.Union.GetChild() {
+			if RewriteContainsIntersection(child) {
+				return true
+			}
+		}
+	case *openfgapb.Userset_Difference:
+		difference := rw.Difference
+		if RewriteContainsIntersection(difference.GetBase()) || RewriteContainsIntersection(difference.GetSubtract()) {
+			return true
+		}
 	}
-	return fmt.Errorf("%s does not exist", msg)
+
+	return false
+}
+
+// RewriteContainsExclusion returns true if the provided userset rewrite
+// is defined by one or more direct or indirect exclusions.
+func RewriteContainsExclusion(rewrite *openfgapb.Userset) bool {
+	switch rw := rewrite.GetUserset().(type) {
+	case *openfgapb.Userset_Intersection:
+		for _, child := range rw.Intersection.GetChild() {
+			if RewriteContainsExclusion(child) {
+				return true
+			}
+		}
+	case *openfgapb.Userset_Union:
+		for _, child := range rw.Union.GetChild() {
+			if RewriteContainsExclusion(child) {
+				return true
+			}
+		}
+	case *openfgapb.Userset_Difference:
+		return true
+	}
+
+	return false
+}
+
+type InvalidRelationError struct {
+	ObjectType string
+	Relation   string
+	Cause      error
+}
+
+func (e *InvalidRelationError) Error() string {
+	return fmt.Sprintf("the definition of relation '%s' in object type '%s' is invalid", e.Relation, e.ObjectType)
+}
+
+func (e *InvalidRelationError) Unwrap() error {
+	return e.Cause
+}
+
+type ObjectTypeUndefinedError struct {
+	ObjectType string
+	Err        error
+}
+
+func (e *ObjectTypeUndefinedError) Error() string {
+	return fmt.Sprintf("'%s' is an undefined object type", e.ObjectType)
+}
+
+func (e *ObjectTypeUndefinedError) Unwrap() error {
+	return e.Err
+}
+
+type RelationUndefinedError struct {
+	ObjectType string
+	Relation   string
+	Err        error
+}
+
+func (e *RelationUndefinedError) Error() string {
+
+	if e.ObjectType != "" {
+		return fmt.Sprintf("'%s#%s' relation is undefiend", e.ObjectType, e.Relation)
+	}
+
+	return fmt.Sprintf("'%s' relation is undefined", e.Relation)
+}
+
+func (e *RelationUndefinedError) Unwrap() error {
+	return e.Err
 }
 
 func AssignableRelationError(objectType, relation string) error {
--- a/server/commands/connected_objects.go
+++ b/server/commands/connected_objects.go
@@ -1,0 +1,350 @@
+package commands
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
+
+	serverErrors "github.com/openfga/openfga/server/errors"
+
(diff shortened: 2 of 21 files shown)


assets/migrations/postgres/003_reverse_lookup_index.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+CREATE INDEX idx_reverse_lookup_user on tuple (store, object_type, relation, _user);
```
jon-whit, line 2: This index is needed to ensure the `ReadStartingWithUser` storage method produces index scans on larger collections of relationship tuples instead of sequence scans.

pkg/typesystem/typesystem.go
```diff
@@ -180,6 +180,31 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgapb.RelationReference, sour
 	return false
 }
 
+func (t *TypeSystem) HasTypeInfo(objectType, relation string) (bool, bool) {
```
jon-whit, line 183: I don't love this function signature returning (bool, bool), but I kept the same pattern as we are using elsewhere. I'd like to revisit this in a subsequent PR and changes these back to appropriate errors.

So

```
func (t *TypeSystem) HasTypeInfo(objectType, relation string) (bool, error) {...}
```

WDYT?

server/commands/list_objects.go
```diff
@@ -94,13 +152,49 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
...
+	}
+
+	relation, ok := typesys.GetRelation(targetObjectType, targetRelation)
+	if !ok {
+		return serverErrors.RelationNotFound(targetRelation, targetObjectType, nil)
+	}
+
+	hasTypeInfo, ok := typesys.HasTypeInfo(targetObjectType, targetRelation)
+	if !ok {
+		// todo(jon-whit): handle error
```
jon-whit, line 190: 🤔 wondering the best course of action here. If we've made it this far we damn well better be referencing a valid `targetObjectType` and `targetRelation`, and so it seems very unlikely we'd hit this case. I wonder if logging a warning is simply best here.

WDYT?

server/commands/list_objects.go
```diff
@@ -49,7 +50,39 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
...
+	if !ok {
+		return nil, serverErrors.RelationNotFound(targetRelation, targetObjectType, nil)
+	}
+
+	hasTypeInfo, ok := typesys.HasTypeInfo(targetObjectType, targetRelation)
+	if !ok {
+		// todo(jon-whit): handle error
+	}
+
+	resultsChan := make(chan string, 1)
```
jon-whit, line 81: Need a minor refactor here to avoid goroutine leaks. 

The `ConnectedObjects` API should create a top-level errgroup and set a concurrency limit on the group and then pass that errgroup down the call chain so that the total concurrency is limited instead of concurrency at each level of the recursive expansion.

Then this should be set to `resultsChan := make(chan string, maxConcurrency)` or whatever..

server/commands/list_objects.go
```diff
@@ -94,13 +154,49 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
...
+	if !ok {
+		return serverErrors.RelationNotFound(targetRelation, targetObjectType, nil)
+	}
+
+	hasTypeInfo, ok := typesys.HasTypeInfo(targetObjectType, targetRelation)
+	if !ok {
+		// todo(jon-whit): handle error
+	}
+
+	resultsChan := make(chan string, 1)
```
jon-whit, line 195: Same here as in https://github.com/openfga/openfga/pull/279/files#r993924033

pkg/typesystem/typesystem.go
elbuo8: I'm unclear why it's not using just the left hand side and have a single one.

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+
+	g := errgroup.Group{}
+	g.SetLimit(100) // set some concurrency limit
+
+	for {
+		t, err := iter.Next()
+		if err != nil {
+			if errors.Is(err, storage.ErrIteratorDone) {
+				break
+			}
```
elbuo8, line 204: Missing handling when the error isn't the iterator has been consumed.

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+			if foundCount != nil && atomic.LoadUint32(foundCount) >= req.limit {
+				break
+			}
+
+			if foundCount != nil {
+				atomic.AddUint32(foundCount, 1)
+			}
+
+			resultChan <- foundObject
+			foundObjectsMap.Store(foundObject, foundObject)
```
elbuo8, line 230: Do we have to store the actual object instead of something nil sized?

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+type reverseExpandRequest struct {
+	storeID          string
+	ingress          *graph.RelationshipIngress
+	sourceObjectRef  *openfgapb.RelationReference
+	targetUserRef    *openfgapb.ObjectRelation
+	contextualTuples []*openfgapb.TupleKey
+	limit            uint32
+}
+
+func (c *ConnectedObjectsCommand) reverseExpandTupleToUserset(
```
elbuo8, line 143: Could we have some tracing/instrumentation in this bit (can be in another PR but please create an issue)

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+			tuples = append(tuples, &openfgapb.Tuple{Key: t})
+		}
+	}
+	iter1 := storage.NewStaticTupleIterator(tuples)
+
+	iter2, err := c.Datastore.ReadStartingWithUser(ctx, store, storage.ReadStartingWithUserFilter{
+		ObjectType: ingress.GetType(),
+		Relation:   ingress.GetRelation(),
+		UserFilter: []*openfgapb.ObjectRelation{
+			req.targetUserRef,
```
elbuo8, line 290: Could we maybe be explicit here similar to above?

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+	}
+
+	iter := storage.NewCombinedIterator(iter1, iter2)
+
+	g := errgroup.Group{}
+	g.SetLimit(100) // set some concurrency limit
+
+	for {
+		t, err := iter.Next()
+		if err != nil {
```
elbuo8, line 305: Ditto above?

server/commands/list_objects.go
```diff
@@ -49,7 +51,39 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
...
+		return nil, serverErrors.TypeNotFound(targetObjectType)
+	}
+
+	relation, ok := typesys.GetRelation(targetObjectType, targetRelation)
+	if !ok {
+		return nil, serverErrors.RelationNotFound(targetRelation, targetObjectType, nil)
+	}
+
+	hasTypeInfo, ok := typesys.HasTypeInfo(targetObjectType, targetRelation)
+	if !ok {
```
elbuo8, line 78: I assume this is to find out the usage of the previous version? Perhaps add a comment for that context. Also, please use named instead of a single string as it's harder to search and do aggregations.

Also, I believe there is a utility `WithContext` that further enriches the log with bits such as the request ID and store ID.

server/commands/list_objects.go
```diff
@@ -69,7 +103,34 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
 			user:        req.User,
 			ctxTuples:   req.ContextualTuples,
 		}, resultsChan, errChan, resolvedChan)
-	}()
+	}
+
+	if !typesystem.RewriteContainsIntersection(relation) &&
```
elbuo8, line 108: Could you please add a comment on what is happening (more specifically, why it's happening).

elbuo8: Maybe tying in both the need for `typeinfo` and the current support for just `union` etc.

server/server_test.go
```diff
@@ -48,6 +49,16 @@ func TestOpenFGAServer(t *testing.T) {
 		ds := memory.New(telemetry.NewNoopTracer(), 10, 24)
 		test.RunAllTests(t, ds)
 	})
+
+	t.Run("TestMySQLDatastore", func(t *testing.T) {
```
elbuo8, line 53: 😬 I'd leave this for a quick PR and then rebase.

server/test/connected_objects_test.go
```diff
@@ -0,0 +1,671 @@
...
+			},
+			expectedObjects: []string{"group:auth0", "group:eng"},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			require := require.New(t)
+
+			ds := memory.New(telemetry.NewNoopTracer(), 10, 24)
```
elbuo8, line 625: I'm guessing this is WIP?

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+		Relation: req.Relation,
+	}
+
+	// build the graph of possible edges between object types in the graph based on the authz model's type info
+	g := graph.BuildConnectedObjectGraph(c.Typesystem)
+
+	// find the possible incoming edges (ingresses) between the target user reference and the source (object, relation) reference
+	ingresses, err := g.RelationshipIngresses(sourceObjRef, targetUserRef)
+	if err != nil {
+		return err
```
jon-whit, line 76: - [ ]: make sure we're surfacing errors from the `server/errors` package here and elsewhere.

server/commands/connected_objects.go
jon-whit: Good point. I'll store an empty struct instead.

server/commands/connected_objects.go
jon-whit: Definitely! I was going to submit a follow up PR with more tracing and instrumentation. I want to think through the exact bits I want to capture, but database queries and resolution depths etc.. are probably good starting points.

jon-whit: https://github.com/openfga/openfga/issues/280

server/server_test.go
jon-whit: Sure! I'll go ahead and do that instead.

server/test/connected_objects_test.go
jon-whit: Oh, good find. That's just residual. It should be wired up to a datastore input at the top-level of the function. Will do what we've done elsewhere here with

```
func ConnectedObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
    ...
    _ = ds // use ds
}
```

server/commands/connected_objects.go
jon-whit: Resolved.

server/commands/connected_objects.go
jon-whit: Now using `foundObjectsMap.Store(foundObject, struct{}{})`. The empty struct is 0 sized.

server/commands/connected_objects.go
jon-whit: I'm not sure I understand what you mean? Could you please clarify?

server/commands/list_objects.go
jon-whit: Yep, I'll do that 👍

server/test/connected_objects_test.go
jon-whit: Resolved.

server/server_test.go
jon-whit: https://github.com/openfga/openfga/pull/281 - will rebase once that's merged.

server/commands/list_objects.go
jon-whit: > I assume this is to find out the usage of the previous version?

Not necessarily, but it could also be used for that I suppose. We need to know if the model being used has type information because if it doesn't then we can't use the `ConnectedObjects` optimization. So this is mainly for that purpose.

> please use named instead of a single string as it's harder to search and do aggregations.

I could change to named fields with something like `{"error": "failed to lookup type info for relation", "relation": targetRelation, "objectType": targetObjectType}`, but having named fields for the relation and objectType there is odd. The only real valuable named param in this context would be the error field itself, and then the relation and objectType fields are just tag along and contextually specific to that error.

Not sure I understand the full reasoning here? Could you clarify?

Are you suggesting
```
msg := fmt.Sprintf("failed to lookup type info for relation '%s' on object type '%s'", targetRelation, targetObjectType)

q.Logger.WarnWithContext(ctx, msg, zap.Error("failed to lookup type info"))
```

> Also, I believe there is a utility WithContext that further enriches the log with bits such as the request ID and store ID.

I switched it to `Logger.WarnWithContext` 👍

server/commands/connected_objects.go
elbuo8: `{Object: req.targetUserRef.Object},`

server/commands/list_objects.go
elbuo8: Im suggesting:
`(ctx, "failed to lookup type info", zap.String("relation", targetRelation), zap.String("type", targetObjectType))

server/commands/connected_objects.go
jon-whit: Oh, gotcha! Keyed literals is what you are after. Thanks for the clarification. I'll make that change.

jon-whit: Oh, this is intentional. In the `expandDirectUserset` case you reverse read the whole `targetUserRef` (which includes `object#relation`). In the `expandTTU` case you only reverse read the tupleset relation on the `targetUserRef.Object`. 

So these cases are different purposefully.

server/commands/list_objects.go
jon-whit: Ok, I'm happy to do that, but it's quite weird IMO. Relation and type have no need to be searchable fields in this context, but that's ok.

I'd be searching logs on the string starting with "failed to lookup type info" and the relation and object type would be in the log message. I don't see a need to break those out into individual fields of the log message. If it was something like `requestID` or some worth searching/indexing a field for then I'd feel differently, but not in this case.

elbuo8: When performing an audit of the system I can immediately think of:
filter by: `"failed to lookup type info`, group by: `store id, type`, from the last week

That would give me all the offending stores and I can communicate them specifically which `type` needs migration/love per store. Relation would be a bit too much 😅

jon-whit: Mmm, the `groupby(storeId, type)` is pretty sexy 😄 I likey 👍 

So how about we go with, for example,

```
(ctx, "failed to lookup type info for relation 'viewer'", zap.String("storeId", store), zap.String("type", targetObjectType))
```

how does that sound?

pkg/typesystem/typesystem.go
jon-whit: Do you mean just return a single bool? I'm confused what you mean by 

> why it's not using just the left hand side and have a single one.

What are you referring to with left hand side and have a single one?

server/commands/connected_objects.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+// StreamedConnectedObjects yields all of the objects of the provided objectType that
+// the given user has a specific relation with. The results will be limited by the request
+// limit. If a 0 limit is provided then all objects of the provided objectType will be
+// returned.
+func (c *ConnectedObjectsCommand) StreamedConnectedObjects(
+	ctx context.Context,
+	req *ConnectedObjectsRequest,
+	resultChan chan<- string, // object string (e.g. document:1)
```
miparnisari, line 122:
```suggestion
	resultChan chan<- Object, // object string (e.g. document:1)
```

server/commands/connected_objects.go
```diff
@@ -0,0 +1,355 @@
...
+) error {
+
+	var foundCount *uint32
+	if req.Limit > 0 {
+		foundCount = new(uint32)
+	}
```
miparnisari, line 128: This looks weird because then you have to check `if foundCount != nil`

server/commands/connected_objects.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+
+		if foundObjectType == sourceObjectType {
+			if foundCount != nil && atomic.LoadUint32(foundCount) >= req.limit {
+				break
+			}
+
+			if foundCount != nil {
+				atomic.AddUint32(foundCount, 1)
+			}
```
miparnisari, line 229: Isn't this prone to the same race condition that we had found in ListObjects v0?

```suggestion
if atomic.AddUint32(foundCount, 1) > req.Limit {
	break
}
```

server/commands/connected_objects.go
```diff
@@ -0,0 +1,355 @@
...
+			// from the database in the first place
+
+			// if we've already evaluated/found the object, then continue
+			continue
+		}
+
+		if foundObjectType == sourceObjectType {
+			if foundCount != nil && atomic.AddUint32(foundCount, 1) > req.limit {
+				break
+			}
```
miparnisari, line 331: wait why didn't you do this in `reverseExpandTupleToUserset`? haha

server/commands/list_objects.go
```diff
@@ -69,7 +103,36 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
 			user:        req.User,
 			ctxTuples:   req.ContextualTuples,
 		}, resultsChan, errChan, resolvedChan)
-	}()
+	}
+
+	// ConnectedObjects currently only supports models that do not include intersection and exclusion,
+	// and the model must include type info for ConnectedObjects to work.
+	if !typesystem.RewriteContainsIntersection(relation) &&
+		!typesystem.RewriteContainsExclusion(relation) &&
```
miparnisari, line 111: What if I have a model like this

```
type document
    relations
         define viewer as self
         define restricted as self
         define editor as viewer but not restricted  
         define admin as editor
````

and then I call `ListObjects(user=maria, relation=admin, type: document)`

There is no exclusion in `admin` but it is defined as `as editor`, and `editor` does have exclusion. Will it work then?

File: server/commands/list_objects.go [hunk 21]
```diff
@@ -94,13 +157,49 @@ func (q *ListObjectsQuery) Execute(ctx context.Context, req *openfgapb.ListObjec
...
+
+	relation, ok := typesys.GetRelation(targetObjectType, targetRelation)
+	if !ok {
+		return serverErrors.RelationNotFound(targetRelation, targetObjectType, nil)
+	}
+
+	hasTypeInfo, ok := typesys.HasTypeInfo(targetObjectType, targetRelation)
+	if !ok {
+		q.Logger.WarnWithContext(ctx, fmt.Sprintf("failed to lookup type info for relation '%s' on object type '%s'", targetRelation, targetObjectType))
+	}
```
miparnisari, line 196: why is all this code duplicate? i'd put it inside `validateInput`

server/test/list_objects.go
```diff
@@ -51,24 +53,151 @@ func newListObjectsRequest(store, objectType, relation, user, modelID string, co
...
+			TypeDefinitions: gitHubTypeDefinitions.GetTypeDefinitions(),
+		}
+		err = ds.WriteAuthorizationModel(ctx, store, model)
+		require.NoError(t, err)
+
+		writes := []*openfgapb.TupleKey{tKAllAdminsRepo6, tkAnnaRepo1, tkAnnaRepo2, tkAnnaRepo3, tkAnnaRepo4, tkBobRepo2}
+		err = ds.Write(ctx, store, nil, writes)
+		require.NoError(t, err)
+
+		testCases := []listObjectsTestCase{
```
miparnisari, line 83: This is duplicate in the other test

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+		name             string
+		model            *openfgapb.AuthorizationModel
+		tuples           []*openfgapb.TupleKey
+		request          *commands.ConnectedObjectsRequest
+		resolveNodeLimit uint32
+		expectedObjects  []string
+		expectedError    error
+	}{
+		{
+			name: "Test-1",
```
miparnisari, line 31:
```suggestion
			name: "Success with tuples and contextual tuples",
```

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+				},
+			},
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("folder:folder1", "viewer", "user:jon"),
+				tuple.NewTupleKey("folder:folderY", "parent", "*"),
+			},
+			expectedObjects: []string{"folder:folder1", "folder:folderX", "folder:folderY"},
+		},
+		{
+			name: "Test-2",
```
miparnisari, line 90:
```suggestion
			name: "Success with no tuples but with contextual tuples",
```

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+				},
+			},
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("folder:folder1", "viewer", "user:jon"),
+				tuple.NewTupleKey("folder:folderX", "parent", "*"),
+			},
+			expectedObjects: []string{"folder:folder1"},
+		},
+		{
+			name: "Test-4",
```
miparnisari, line 206:
```suggestion
			name: "Success with direct relationships",
```

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+					},
+				},
+			},
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("document:doc1", "viewer", "user:jon"),
+			},
+			expectedObjects: []string{"document:doc1", "document:doc3"},
+		},
+		{
+			name: "Test-5",
```
miparnisari, line 245:
```suggestion
			name: "Success with direct relationships and usersets",
```

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("document:doc1", "viewer", "user:jon"),
+				tuple.NewTupleKey("document:doc2", "viewer", "user:bob"),
+				tuple.NewTupleKey("document:doc3", "viewer", "group:openfga#member"),
+				tuple.NewTupleKey("group:openfga", "member", "user:jon"),
+			},
+			expectedObjects: []string{"document:doc1", "document:doc3"},
+		},
+		{
+			name: "Test-6",
```
miparnisari, line 299:
```suggestion
			name: "Success with direct relationships and computed usersets",
```

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+			},
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("folder:folder1", "viewer", "user:jon"),
+				tuple.NewTupleKey("folder:folder2", "parent", "folder:folder1"),
+				tuple.NewTupleKey("folder:folder3", "parent", "folder:folder2"),
+			},
+			expectedError: serverErrors.AuthorizationModelResolutionTooComplex,
+		},
+		{
+			name: "Test-10",
```
miparnisari, line 580:
```suggestion
			name: "Success when req.User is a userset",
```

server/commands/connected_objects.go
jon-whit: Ah, I updated this in one spot but not here. It's supposed to be

```
if foundCount != nil && atomic.AddUint32(foundCount, 1) > req.limit { ... }
```

will fix this occurance.

server/commands/connected_objects.go
jon-whit: See my comment above. Just an oversight. That's why we have reviewers.

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+						},
+					},
+				},
+			},
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("folder:folder1", "viewer", "user:jon"),
+				tuple.NewTupleKey("folder:folder2", "parent", "folder:folder1"),
+				tuple.NewTupleKey("folder:folder3", "parent", "folder:folder2"),
+			},
+			expectedError: serverErrors.AuthorizationModelResolutionTooComplex,
```
miparnisari, line 577: I don't understand why this would fail. `folder1` is parent of `folder2` which is parent of `folder3`. `jon` can view folder1. Therefore, `jon` can also view folder2 and folder3?

server/commands/list_objects.go
jon-whit: That case should default to the old behavior (concurrent checks). Let me add a test for that to make sure that the `typesystem` package's `RewriteContainsIntersection` and `RewriteContainsExclusion` methods behave this way. If they don't then we'll need to fix that behavior.

The idea is that if intersection or exclusion is involved whatsoever in the evaluation, then we should fall back to prior behavior.

server/commands/list_objects.go
jon-whit: I'm working on refactoring this whole thing. We have duplication across the streamed and regular that doesn't need to exist either. The logic can be shared for many things. I'll let you know when I've updated.

server/test/connected_objects.go
jon-whit: Because of the `resolveNodeLimit`. The resolution depth is constrained to a depth of 2, but the nested hierarchy contains a depth of 3, and so it returns an error.

server/test/connected_objects.go
```diff
@@ -0,0 +1,668 @@
...
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("document:doc1", "owner", "user:jon"),
+				tuple.NewTupleKey("document:doc2", "owner", "user:bob"),
+				tuple.NewTupleKey("document:doc3", "owner", "group:openfga#member"),
+				tuple.NewTupleKey("group:openfga", "member", "user:jon"),
+			},
+			expectedObjects: []string{"document:doc1", "document:doc3"},
+		},
+		{
+			name: "Test-7",
```
miparnisari, line 358:
```suggestion
			name: "Success with many tuples",
```

? 🤷‍♀️

pkg/typesystem/typesystem.go
miparnisari: > Do you mean just return a single bool?

I think so. I'm also confused because the function name is a question and it returns two answers :P

server/commands/list_objects.go
jon-whit: Yeah, this is an issue. These methods only operate directly on the relation's rewrite itself. It doesn't follow other relations across types. Good catch! Will fix tomorrow and make sure tests cover those cases.

server/commands/connected_objects.go
jon-whit: This was less confusing to me than the alternative where `foundCount == 0`. What if foundCount is legitimately 0 instead of just the zero value?

The reason for the pointer here is that `foundCount` is unused if the limit is unspecified (0 value).

server/commands/connected_objects.go
jon-whit: This is resolved now.

server/commands/connected_objects.go
jon-whit: I originally had this signature, but because of the composition of ListObjects using a `string` it became more conversions to/from string to Object for little to no value. I'd like to refactor this like you suggest (and I originally had), but I think I'll refactor the whole ListObjects composition to do so (hell the API for that matter).

Our API should change to

```
message ListObjectsResponse {
    repeated Object objects = 1;
}
```
but right now we have

```
message ListObjectsResponse {
    repeated string object_ids = 1;
}
```

pkg/typesystem/typesystem.go
jon-whit: It returns `(bool, bool)` because the `objectType` or `relation` may be undefined for the typesystem constructed from the authorization model.

For example,

```
type user

type document
    relations
      define viewer: [user]
```

```
hasTypeInfo, ok := HasTypeInfo("undefined", "viewer")
fmt.Println(ok) // returns false

hasTypeInfo, ok := HasTypeInfo("document", "undefined")
fmt.Println(ok) // returns false

hasTypeInfo, ok := HasTypeInfo("document", "viewer")
fmt.Println(ok) // returns true
```

server/commands/list_objects.go
jon-whit: Alright, I refactored into a shared `handler` that is used for `ExecuteStreamed` and `Execute`. It's a lot nicer than it was before in v0 and what I had before my most recent commit. Let me know what you think of the newly refactor structure. Way less duplication!

server/commands/list_objects.go
```diff
@@ -31,15 +34,100 @@ type ListObjectsQuery struct {
...
+	relation, ok := typesys.GetRelation(targetObjectType, targetRelation)
+	if !ok {
+		return serverErrors.RelationNotFound(targetRelation, targetObjectType, nil)
+	}
+
+	hasTypeInfo, ok := typesys.HasTypeInfo(targetObjectType, targetRelation)
+	if !ok {
+		q.Logger.WarnWithContext(
+			ctx, fmt.Sprintf("failed to lookup type info for relation '%s'", targetRelation),
+			zap.String("storeID", req.GetStoreId()),
```
jon-whit, line 84: This and the `objectType` should be renamed to `store_id` and `object_type` to remain consistent with the conventions elsewhere.

server/test/list_objects.go
jon-whit: `model.Id` differs because the others have to be against v1.1 models. I'll think through how I can refactor to reduce duplication.

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+				contextualTuples: req.ContextualTuples,
+				limit:            req.Limit,
+			}, resultChan, foundObjectsMap, foundCount)
+			if err != nil {
+				return err
+			}
+
+		default:
+			return errors.Errorf("unsupported ingress type")
+		}
```
miparnisari, line 109:
```suggestion
	for _, ingress := range ingresses {
		request := &reverseExpandRequest{
			storeID:          storeID,
			ingress:          ingress,
			sourceObjectRef:  sourceObjRef,
			targetUserRef:    req.User,
			contextualTuples: req.ContextualTuples,
			limit:            req.Limit,
		}
		var err error
		switch ingress.Type {
		case graph.DirectIngress:
			err = c.reverseExpandDirect(ctx, request, resultChan, foundObjectsMap, foundCount)
		case graph.TupleToUsersetIngress:
			err = c.reverseExpandTupleToUserset(ctx, request, resultChan, foundObjectsMap, foundCount)

		default:
			return errors.Errorf("unsupported ingress type")
		}

		if err != nil {
			return err
		}
	}
```

server/commands/connected_objects.go
miparnisari: > foundCount is unused if the limit is unspecified

Can't we count regardless of whether the limit was set or not?

server/commands/connected_objects.go
```diff
@@ -0,0 +1,351 @@
...
+			foundObjectsMap.Store(foundObject, struct{}{})
+		}
+
+		g.Go(func() error {
+			return c.streamedConnectedObjects(ctx, &ConnectedObjectsRequest{
+				StoreID:          store,
+				ObjectType:       sourceObjectType,
+				Relation:         sourceObjectRel,
+				User:             &openfgapb.ObjectRelation{Object: foundObject, Relation: req.targetUserRef.Relation},
+				ContextualTuples: req.contextualTuples,
```
miparnisari, line 237: isn't `Limit:            req.limit,` missing here?

server/server.go
```diff
@@ -196,6 +212,19 @@ func (s *Server) StreamedListObjects(req *openfgapb.StreamedListObjectsRequest,
...
+	if err != nil {
+		return serverErrors.HandleError("", err)
+	}
+
+	typesys := typesystem.New(model)
+
+	connectObjCmd := &commands.ConnectedObjectsCommand{
+		Datastore:  s.datastore,
+		Typesystem: typesys,
+	}
```
miparnisari, line 226: why is `ResolveNodeLimit: s.config.ResolveNodeLimit,` not here?

server/commands/list_objects.go
```diff
@@ -31,15 +34,100 @@ type ListObjectsQuery struct {
...
+		userObj, userRel := tuple.SplitObjectRelation(req.GetUser())
+
+		handler = func() {
+			err = q.ConnectedObjects(ctx, &ConnectedObjectsRequest{
+				StoreID:          req.GetStoreId(),
+				ObjectType:       targetObjectType,
+				Relation:         targetRelation,
+				User:             &openfgapb.ObjectRelation{Object: userObj, Relation: userRel},
+				ContextualTuples: req.GetContextualTuples().GetTupleKeys(),
+				Limit:            q.ListObjectsMaxResults,
```
miparnisari, line 108: Minor: everything being passed here is related to the request, except for `q.ListObjectsMaxResults`. Maybe move `Limit` to the `ConnectedObjectsCommand` struct?

server/test/connected_objects.go
jon-whit: renamed

server/test/connected_objects.go
jon-whit: renamed

server/test/connected_objects.go
jon-whit: renamed

server/test/connected_objects.go
jon-whit: renamed

server/test/connected_objects.go
jon-whit: renamed

server/server.go
jon-whit: Good catch. Resolved!

server/commands/list_objects.go
jon-whit: ❤️ the recommendation. I've made these changes. Much better!

server/commands/connected_objects.go
jon-whit: Resolved by addressing the other recommendation.

server/test/connected_objects.go
jon-whit: Renamed the test for clarity on this one.

server/commands/connected_objects.go
jon-whit: Sure, but then what do we do in the check
```
if foundCount > X {...}
```
what should `X` be? I could set it to `math.MaxUint64`, but semantically there should be no limit if the limit is omitted. So this is why I chose the path I did. If limit is omitted (default 0 value) then we don't constrain the search to a maximum number of objects.

server/commands/list_objects.go
jon-whit: @miparnisari I have addressed this behavior now in the most recent changes. Would you mind taking a thorough second look and tell me what you think 👍 ?

server/commands/list_objects.go
jon-whit: Will save this for a later optimization outside of the scope of this PR, because this will require a lot of changes to the orchestration of this code.

server/server.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/openfga/openfga/pkg/typesystem"
+
```
craigpastro, line 14: Nit: no space

server/commands/connected_objects.go
```diff
@@ -0,0 +1,340 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"sync/atomic"
+
+	serverErrors "github.com/openfga/openfga/server/errors"
+
```
craigpastro, line 10: Nit: no space

craigpastro (commented):
A couple of small nitpicks.

I think this is going to be a case of smalls prs get a lot of comments and large prs just lgtm. It passes tests and I can't see anything obviously wrong, so lgtm!

server/test/list_objects.go
```diff
@@ -203,33 +338,149 @@ func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObje
...
+
 	ctx := context.Background()
-	data, err := os.ReadFile(gitHubTestDataFile)
-	if err != nil {
-		return nil, nil, "", err
+	store := ulid.Make().String()
+
+	model := &openfgapb.AuthorizationModel{
+		Id:            ulid.Make().String(),
+		SchemaVersion: typesystem.SchemaVersion1_0,
```
miparnisari, line 351: This is 1.1 🤔  why didn't the WriteAuthModel fail?

server/test/list_objects.go
```diff
@@ -203,33 +338,149 @@ func runListObjectsTests(t *testing.T, ctx context.Context, testCases []listObje
...
+		},
+	}
+	err := ds.WriteAuthorizationModel(ctx, store, model)
+	require.NoError(b, err)
+
+	n := 0
+	for i := 0; i < 200; i++ {
+		var tuples []*openfgapb.TupleKey
+
+		for j := 0; j < 100; j++ {
```
miparnisari, line 382: `< 100` could be `< ds.MaxTuplesInWriteOperation()`

But `i < 200` looks like an arbitrary number

assets/migrations/mysql/003_reverse_lookup_index.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+CREATE INDEX idx_reverse_lookup_user on tuple (store, object_type, relation, _user);
```
miparnisari, line 2: Maybe rename this file to `003_add_connected_objects_to_user_index.sql`

pkg/typesystem/typesystem.go
```diff
@@ -143,42 +150,231 @@ func (t *TypeSystem) GetRelations(objectType string) (map[string]*openfgapb.Rela
...
+func (t *TypeSystem) GetRelation(objectType, relation string) (*openfgapb.Relation, error) {
+	relations, err := t.GetRelations(objectType)
+	if err != nil {
+		return nil, err
 	}
 
 	r, ok := relations[relation]
 	if !ok {
-		return nil, false
+		return nil, &RelationUndefinedError{
```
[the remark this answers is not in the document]
jon-whit: It doesn't. I didn't want to change it too significantly though from what we were doing, just start moving us in the right direction. We shouldn't really introduce custom error types unless we have a strong reason to do so. I'd prefer we just wrap errors like `ErrRelationUndefined` in this case with more context. WDYT?

server/server.go
jon-whit: 👍 interesting my editor didn't adjust those imports like it usually does 🤷

server/commands/list_objects.go
```diff
@@ -31,15 +34,103 @@ type ListObjectsQuery struct {
...
+		q.Logger.WarnWithContext(
+			ctx, fmt.Sprintf("failed to lookup type info for relation '%s'", targetRelation),
+			zap.String("store_id", req.GetStoreId()),
+			zap.String("object_type", targetObjectType),
+		)
+	}
+
+	handler := func() {
+		q.performChecks(ctx, req, resultsChan, errChan, resolvedChan)
+	}
```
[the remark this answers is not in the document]
jon-whit: I went the way I did because it involves less nested code. I think it's ok to set the default case and then modify it based on other conditions.

server/test/list_objects.go
jon-whit: The datastore doesn't have validation logic like that, so that's why it passed. I'll change this to `typesystem.SchemaVersion1_1` though. Good catch

server/test/connected_objects.go
```diff
@@ -0,0 +1,670 @@
...
+							},
+						},
+					},
+				},
+			},
+			tuples:          []*openfgapb.TupleKey{},
+			expectedObjects: []string{"folder:folder1", "folder:folderX", "folder:folderY"},
+		},
+		{
+			name: "Restrict results based on limit",
```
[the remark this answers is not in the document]
jon-whit: Yeah, you aren't guaranteed to get the same result every time because the evaluation is done concurrently.

server/test/connected_objects.go
```diff
@@ -0,0 +1,670 @@
...
+							},
+						},
+					},
+				},
+			},
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("folder:folder1", "viewer", "user:jon"),
+				tuple.NewTupleKey("folder:folderX", "parent", "*"),
+			},
+			expectedObjects: []string{"folder:folder1"},
```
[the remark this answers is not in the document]
jon-whit: Technically this test case needs to be changed and some changes in ConnectedObjects as well, because with our recent changes the `folder:folderX#parent@*` tuple should lead to an error. Wildcards aren't allowed on tupleset relations.

assets/migrations/mysql/003_reverse_lookup_index.sql
jon-whit: The name of the file matches the index being added. Do you recommend we change the name of the index? I thought it made sense to have a reverse_lookup index since we're reverse looking up on the `user` field.

server/test/connected_objects.go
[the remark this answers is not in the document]
jon-whit: I've refactored the ConnectedObjects API a bit so that it returns the same error as the Check path if a `*` is encountered on a tupleset relation (e.g. `folder#parent`).

Let me know what you think 👍

assets/migrations/mysql/003_reverse_lookup_index.sql
miparnisari: Oh I kind of wanted to follow the pattern of the previous migrations that start with a verb: `initialize`, `add`, etc.

jon-whit: Ok, I'll rename to `003_add_reverse_lookup_index.sql`.

server/test/list_objects.go
miparnisari: > 100 is the max concurrency limit

The call to `ds.Write(tuples)` will fail if you pass in more tuples that the data storage can accept

jon-whit: Resolved

craigpastro (approved):
LGTM 🚀

craigpastro (dismissed):
LGTM

craigpastro (approved):
LGTM

PR #297, opened 2022-10-19
fix: ensure object type and relation exists in check

craigpastro:
Related:
1. https://discord.com/channels/759188666072825867/920359187949707305/1019251304251523112
2. https://github.com/openfga/openfga/pull/234

~I need to add a test for this.~ Test added

## Review Checklist


server/test/check.go
```diff
@@ -1167,6 +1167,45 @@ func TestCheckQuery(t *testing.T, datastore storage.OpenFGADatastore) {
...
+			name:             "edge case 1",
+			resolveNodeLimit: defaultResolveNodeLimit,
+			request: &openfgapb.CheckRequest{
+				TupleKey: tuple.NewTupleKey("document:1", "viewer", "anne"),
+			},
+			typeDefinitions: []*openfgapb.TypeDefinition{
+				{
+					Type: "folder",
+					Relations: map[string]*openfgapb.Userset{
+						"viewer": {
```
jon-whit, line 1180: Please use the shorthands for these

`typesystem.This()`
`typesystem.TupleToUserset("parent", "viewer")`

etc..

jon-whit:
@craigpastro are we sure this related to https://discord.com/channels/759188666072825867/920359187949707305/1019251304251523112? 

The reported issue seems to be a problem with the string validation of the `user` value, not behavior related to the evaluation of objects encountered in user values during Check evaluation. For example, the reported issue didn't include any user values that had something like `type:id`. All of his examples were something just directly of the form `id`. Could you clarify how this relates to that issue?

craigpastro:
In the context of TTU rewrites:

The first fix for that bug was to skip user_ids. This fix is to skip objects which don't have the corresponding computed userset relation. As of now, a client can write a user of the form "object:id" where "object" can be anything. Our current code will consider this an object, attach the computed userset "relation" to it, and recursively call check on it. This doesn't make sense unless the "object" type has the "relation" defined on it.

In terms of what we have discovered recently, this may need rethinking.

server/commands/check.go
```diff
@@ -563,17 +561,24 @@ func (query *CheckQuery) resolveTupleToUserset(
 			continue // TupleToUserset tuplesets should be of the form 'objectType:id' or 'objectType:id#relation' but are not guaranteed to be because it is neither a user or userset
 		}
 
-		usersetRel := node.TupleToUserset.GetComputedUserset().GetRelation() //reader
+		userObjType, _ := tupleUtils.SplitObject(userObj)
```
jon-whit, line 564: `tupleUtils.GetType(userObj)`

craigpastro: So many functions. Updated. Thanks!

server/test/check.go
```diff
@@ -1205,6 +1205,34 @@ func TestCheckQuery(t *testing.T, datastore storage.OpenFGADatastore) {
...
 			},
 		},
+		{
+			name:             "EdgeCase1",
+			resolveNodeLimit: defaultResolveNodeLimit,
+			request: &openfgapb.CheckRequest{
+				TupleKey: tuple.NewTupleKey("document:1", "viewer", "abigail"),
+			},
+			typeDefinitions: []*openfgapb.TypeDefinition{
+				{
```
jon-whit, line 1215: Add the `user` type definition so it's not ambiguous.

craigpastro: Done! Thanks.

jon-whit (approved):

Issue #314, opened 2022-10-26
[Bug] OpenFGA returning 500 on ListObjects


fschulte-eiwa:
Hello, we are testing OpenFGA for a project and had some issues with the list_objects (normal) endpoint. Not missing data but server error.
Our tested model is as follows:

```py
type organization
  relations
    define admin as self
    define editor as self or admin
    define member as self or viewer
    define viewer as self or editor
type type_1
  relations
    define owner as self
    define editor as self or owner
    define viewer as self or editor
type type_2
  relations
    define owner as self
    define editor as self or owner or editor from parent
    define viewer as self or editor or viewer from parent
    define parent as self
```

Server setup as the documentation shows for postgres engine:

```sh
docker network create openfga
docker run -d --name postgres --network=openfga -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password postgres:14
docker run --rm --network=openfga openfga/openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@postgres:5432/postgres?sslmode=disable'
docker run --name openfga --network=openfga -p 3000:3000 -p 8080:8080 -p 8081:8081 openfga/openfga run --datastore-engine postgres --datastore-uri 'postgres://postgres:password@postgres:5432/postgres?sslmode=disable' --listObjects-max-results 50000  --listObjects-deadline 360s
```

First we added a user as admin of an organization ORG1: `{"user": "user_1","relation": "admin","object": "organization:ORG1"}`

We then performed 600 iterations of adding 25 tuples: `{"user": "type_1: UUID_1","relation": "parent","object": "type_2:UUID_n"}`
and 1 tuple giving members of ORG1 "editor" relation to each type_1 element: `{"user": "organization:ORG1#member", "relation": "editor", "object": "type_1:UUID_1"}`.

This created 600 type_1 elements and 15000 type_2 elements where each type_1 element is parent of 25 type_2 elements

When we tried listing the objects of type_2 that some user has by using the list_objects endpoint with `{"type": type_2,"relation": "editor","user": "user_1"}` we get the 15000 ids but if we try listing with a "viewer" relation rather than "editor" `{"type": type_2,"relation": "viewer","user": "user_1"}` we get a 500 response.

In the server we get the error:

```sh
2022-10-26T15:50:17.700Z        INFO    db_stats        {"method": "Check", "reads": 50, "writes": 0}
[...]

PR #315, opened 2022-10-26
fix: throw error if config.http.upstreamTimeout is lower than config.listObjectsDeadline

miparnisari:
Throw an error if attempting to run the server and `config.http.upstreamTimeout` and `config.listObjectsDeadline`.

## References
https://github.com/openfga/openfga/issues/314

## Review Checklist

### DIFF
--- a/pkg/cmd/service/service.go
+++ b/pkg/cmd/service/service.go
@@ -179,7 +179,7 @@ func DefaultConfig() *Config {
 		MaxTypesPerAuthorizationModel: 100,
 		ChangelogHorizonOffset:        0,
 		ResolveNodeLimit:              25,
-		ListObjectsDeadline:           3 * time.Second, // there is a 3-second timeout elsewhere
+		ListObjectsDeadline:           3 * time.Second, // cannot be higher than HTTP.UpstreamTimeout
 		ListObjectsMaxResults:         1000,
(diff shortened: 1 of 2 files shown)


craigpastro (approved):
Nice. Does this fix this bug we discussed this morning?

This LGTM, but I am wondering if we simply have too many knobs to control. Perhaps we should be more opinionated with our config. For example, perhaps `ListObjectsDeadline` should always be set to 1 second less than `UpstreamTimeout`? 🤷

miparnisari:
> Nice. Does this fix this bug we discussed this morning?

I think so :) 

`./openfga run --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --listObjects-max-results 50000  --listObjects-deadline 360s --http-upstream-timeout 360s`

and I inserted ` time.Sleep(359 * time.Second)` inside function `internalCheck`:

image


> This LGTM, but I am wondering if we simply have too many knobs. Perhaps we should be more opinionated with our config. For example, perhaps `ListObjectsDeadline` should always be set to 1 second less than `UpstreamTimeout`? 🤷

Yeah I agree, I'm not sure why we have two configs in the first place.

jon-whit (approved):

PR #317, opened 2022-10-28
refactor: move all server run code to one place

craigpastro:
## Review Checklist


jon-whit:
Let's move the `cmd/openfga/util` package up one level so that it's under `cmd/util`.

cmd/openfga/run.go
```diff
@@ -0,0 +1,773 @@
...
+		go func() {
+			err = playground.ListenAndServe()
+			if err != http.ErrServerClosed {
+				logger.Fatal("failed to start the openfga playground server", zap.Error(err))
+			}
+			logger.Info("shutdown the openfga playground server")
+		}()
+	}
+
+	logger.Info(
```
jon-whit, line 625: Can we move this up before we actually start the OpenFGA server?

cmd/openfga/run.go
```diff
@@ -0,0 +1,773 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	if err := runServer(context.Background(), config); err != nil {
+		panic(err)
+	}
+}
+
+func runServer(ctx context.Context, config *Config) error {
```
jon-whit, line 326: Let's do a sanity check on the usage of `ctx` in this method body. I see a couple of spots where we don't propagate the provided context.

cmd/openfga/run.go
craigpastro: Yep. Done. Thanks!

cmd/openfga/run.go
craigpastro: There are two spots that don't use the passed in context: one is where we use it as a timeout for a `grpc.Dialcontext`, and the other is for a shutdown context.

pkg/logger/logger.go
```diff
@@ -120,6 +120,13 @@ func NewJSONLogger() (*ZapLogger, error) {
 	}, nil
 }
 
+func Must(logger *ZapLogger, err error) *ZapLogger {
```
jon-whit, line 123: Instead of introducing this to be composed as `logger.Must(logger.NewTextLogger())`, I'd recommend just moving this up one level.

`MustNewTextLogger() *ZapLogger`
`MustNewJSONLogger() *ZapLogger`

then just use those directly 👍

craigpastro: Done. Thanks!

jon-whit (approved):

PR #325, opened 2022-11-02
refactor: remove dbCounter

craigpastro:
## Description

Remove dbCounter for now. The plan is to add it back in the context and increment the counters at the storage layer.

## Review Checklist


internal/utils/utils.go
```diff
@@ -0,0 +1,42 @@
+package utils
+
+import (
+	"sync"
+)
+
+// ResolutionMetadata stores the number of times database relationship are resolved. It also returns number of times
+// the database is read from and written to.
+type ResolutionMetadata struct {
```
jon-whit, line 9: At some point in the (hopefully) near future we will be dispatching requests across peers in a cluster of OpenFGA servers. Consequently, we'll need this `ResolutionMetadata` as part of the API contract between dispatchers. We'll also need the `ResolutionDepth` field to be part of this metadata, because when requests are dispatched they will have to propagate the ResolutionDepth to the peer.

For the reasons above, can we move this into our `openfga/api` repository OR create a `proto` top-level directory in this repository and start the beginnings of our internal protobuf API contract which we will extend when we add dispatching support etc..

jon-whit: Sure, but what's the point of having ResolutionMetadata anymore any all if it isn't used? If we're not going to have it right now, let's get rid of it. Otherwise, let's put it somewhere it will have a good home for near future work.

jon-whit (approved):

PR #327, opened 2022-11-03
feat: add db counter

craigpastro:
## Description

TODO: add dbCounter to the streaming context

## Review Checklist


storage/wrapper.go
```diff
@@ -0,0 +1,168 @@
...
+	Reads  atomic.Int32
+	Writes atomic.Int32
+}
+
+// StorageWrapper is a wrapper around any OpenFGADatastore and is used to record metadata around how the underlying
+// implementation is being used.
+//
+// Currently, we are using it to record the number of reads and writes that are performed. Note that the values for the
+// reads and writes may not be accurate as we are only measuring calls that flow through the interface, and, such a call
+// may involve any number of database accesses under the hood.
```
miparnisari, line 27: > the values for the  reads and writes may not be accurate as we are only measuring calls that flow through the interface, and, such a call may involve any **number of database accesses under the hood**

but this is exactly what i'd expect to see, especially if we are using one of the iterators... each call to `fetchResults()` should add 1 to the `read` counter

storage/wrapper.go
```diff
@@ -0,0 +1,168 @@
...
+	"sync/atomic"
+	"time"
+
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+type ctxKey string
+
+const (
+	DBCounterCtxKey ctxKey = "db_counter"
```
miparnisari, line 14: This used to be `db_stats`, i wouldn't change it or we'll have to remember when querying old data

Also, do we need db stats at all in OpenFGA?

storage/wrapper.go
craigpastro: I think that is too much work for too little value. We would have to add it to each implementation. As it is now, it gives a general idea... which there may be some value in:

storage/wrapper.go
craigpastro: I think the msg key used to be `db_stats`. I think a whole log line dedicated to logging db_stats is a bit overkill and adds too much noise. In any case, I don't think we actually ever queried on it did we? We do have a dashboard widget with it, but, again, I don't think it adds much value.

storage/wrapper.go
craigpastro: FWIW it definitely did not have that accuracy before either. In fact, it is more accurate now than before.

PR #341, opened 2022-11-08
chore: cleanup tuple validations across packages

jon-whit:
## Description
This PR cleans up tuple validations across various packages by consolidating validation into a single place and by creating a reusable `TupleFilterFunc` to the storage interface that can be used to filter out tuples as they are encountered that are incompatible with the authorization model used in the query.

## Review Checklist


server/test/write.go
```diff
@@ -1398,40 +1407,10 @@ var writeCommandTests = []writeCommandTest{
 				tuple.NewTupleKey("document:budget", "parent", "folder:budgets#admin"),
 			}},
 		},
-		err: serverErrors.InvalidTuple("Userset 'folder:budgets#admin' is not allowed to have relation 'parent' with 'document:budget'",
+		err: serverErrors.InvalidTuple("unexpected userset relationship 'folder:budgets#admin' with tupleset relation 'document#parent'",
 			tuple.NewTupleKey("document:budget", "parent", "folder:budgets#admin"),
 		),
 	},
-	{
```
jon-whit, line 1405: This test is an invalid test since the `folder#parent` relation is a tupleset relation and is defined under a Union rewrite.

server/test/write.go
```diff
@@ -1239,36 +1271,13 @@ var writeCommandTests = []writeCommandTest{
...
+				tuple.NewTupleKey("document:budget", "reader", "group:*"),
 			}},
 		},
-		err: serverErrors.InvalidTuple("User '*' is not allowed to have relation reader with document:budget",
-			tuple.NewTupleKey("document:budget", "reader", "*"),
+		err: serverErrors.InvalidTuple("User 'group:*' is not allowed to have relation reader with document:budget",
+			tuple.NewTupleKey("document:budget", "reader", "group:*"),
 		),
 	},
-	{
```
jon-whit, line 1249: Removing this in favor of the added test in `server/test/write_authzmodel.go` that makes sure you can't write a 1.1 model without type information.

server/commands/expand.go
```diff
@@ -43,21 +41,44 @@ func (query *ExpandQuery) Execute(ctx context.Context, req *openfgapb.ExpandRequ
...
 
-	model, err := query.datastore.ReadAuthorizationModel(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
+	typesys := typesystem.New(model)
+
+	rel, err := typesys.GetRelation(tuple.GetType(object), relation)
 	if err != nil {
-		if err == storage.ErrNotFound {
-			return nil, serverErrors.AuthorizationModelNotFound(req.GetAuthorizationModelId())
+		if errors.Is(err, typesystem.ErrObjectTypeUndefined) {
+			// todo(jon-whit): handle this case. Should we ever be able to get here if the code above was correct?
```
jon-whit, line 68: TODO: handle this case

server/commands/expand.go
```diff
@@ -43,21 +41,44 @@ func (query *ExpandQuery) Execute(ctx context.Context, req *openfgapb.ExpandRequ
...
+		}
+
+		if errors.Is(err, typesystem.ErrRelationUndefined) {
+			// todo(jon-whit): handle this case. Should we ever be able to get here if the code above was correct?
```
jon-whit, line 72: TODO: handle this case

server/commands/expand.go
```diff
@@ -194,21 +226,6 @@ func (query *ExpandQuery) resolveTupleToUserset(
 		}
 	}
 
-	tuplesetRewrite := relation.GetRewrite().GetUserset()
-	if tuplesetRewrite != nil && reflect.TypeOf(tuplesetRewrite) != reflect.TypeOf(&openfgapb.Userset_This{}) {
```
jon-whit, line 198: This case is covered in the `validation/validation.go` (`validateTuplesetRestrictions`) method now. A tuple encountered with a tupleset relation that involves a rewrite according to the model will now be skipped.

pkg/tuple/validation.go
```diff
@@ -56,10 +56,3 @@ type IndirectWriteError struct {
 func (i *IndirectWriteError) Error() string {
 	return fmt.Sprintf("Cannot write tuple '%s'. Reason: %s", i.TupleKey, i.Reason)
 }
-
-func ValidateUser(tk *openfgapb.TupleKey) error {
```
jon-whit, line 60: Refactored and repurposed in `server/validation/validation.go`

pkg/typesystem/typesystem.go
```diff
@@ -105,6 +105,7 @@ func Difference(base *openfgapb.Userset, sub *openfgapb.Userset) *openfgapb.User
 }
 
 type TypeSystem struct {
+	model           *openfgapb.AuthorizationModel
```
jon-whit, line 108: todo: follow up and refactor the typsystem to work off of the model instead of storing `typeDefinitions`, `schemaVersion`, and `model`

pkg/typesystem/typesystem.go
```diff
@@ -748,37 +762,53 @@ func InvalidRelationTypeError(objectType, relation, relatedObjectType, relatedRe
...
 		}
 	}
+
 	return response
 }
 
-func (t *TypeSystem) getAllTupleToUsersetsDefinitions(relationDef *openfgapb.Userset, resp *[]*openfgapb.TupleToUserset) []*openfgapb.TupleToUserset {
+// IsTuplesetRelation returns a boolean indicating if the provided relation is defined under a
+// TupleToUserset rewrite as a tupleset relation (e.g. the right hand side of a `X from Y`).
+func (t *TypeSystem) IsTuplesetRelation(objectType, relation string) (bool, error) {
```
jon-whit, line 782: - [ ] @jon-whit add a test for this

server/commands/expand.go
```diff
@@ -43,21 +40,44 @@ func (query *ExpandQuery) Execute(ctx context.Context, req *openfgapb.ExpandRequ
...
-		return nil, err
+		if errors.Is(err, storage.ErrNotFound) {
+			return nil, serverErrors.AuthorizationModelNotFound(modelID)
+		}
+
+		return nil, serverErrors.HandleError("", err)
 	}
 
-	model, err := query.datastore.ReadAuthorizationModel(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
+	err = validation.ValidateObject(model, tk)
```
jon-whit, line 52: Unfortunately this validation and the one below cannot be delegated to `validation.ValidateTuple` because we don't valid the `user` field here since we're expanding an `object#relation`. So this is about as consolidated as it gets for Expand.

server/validation/validation.go
```diff
@@ -1,47 +1,313 @@
...
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
 
-// ValidateTuple checks whether a tuple is valid. If it is not, returns as error.
-// If it is, it grabs the type of the tuple's object, and the tuple's relation, and returns its corresponding userset.
-func ValidateTuple(ctx context.Context, backend storage.TypeDefinitionReadBackend, store, authorizationModelID string, tk *openfgapb.TupleKey) (*openfgapb.Userset, error) {
-	if err := tuple.ValidateUser(tk); err != nil {
-		return nil, err
+// ValidateTuple checks whether a tuple is welformed and valid according to the provided model.
+func ValidateTuple(model *openfgapb.AuthorizationModel, tk *openfgapb.TupleKey) error {
```
jon-whit, line 15: I like taking the `*openfgapb.AuthorizationModel` but then we subsequently construct a typesystem off of it all over the place. I'm beginning to lean toward taking a `*typesystem.TypeSystem` instead. 

@miparnisari @craigpastro do you have a preference/opinion?

server/validation/validation.go
```diff
@@ -1,47 +1,313 @@
...
+				Reason:   fmt.Sprintf("undefined type for user '%s'", tk.GetUser()),
+				TupleKey: tk,
+			}
+		}
+	}
+
+	relationInformation := relationsForObject[tk.Relation]
+
+	// case 1
+	if userRel == "" && userID != "*" {
```
jon-whit, line 163: This code needs some work now that we have typed and untyped wildcard behavior. This path seems to only account for the untyped wildcard behavior. 

- [ ] @jon-whit make sure the typed wildcard path is accounted for

server/validation/validation.go
craigpastro: I think taking a `TypeSystem` makes more sense.

server/validation/validation_test.go
```diff
@@ -0,0 +1,424 @@
...
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+func TestValidateTuple(t *testing.T) {
+
+	tests := []struct {
+		name        string
+		tuple       *openfgapb.TupleKey
+		model       *openfgapb.AuthorizationModel
+		expectError bool
```
jon-whit, line 18: - [ ] @jon-whit change this to be the expected error instead of a bool

server/validation/validation.go
jon-whit: Refactored to take `TypeSystem`. That is better 👍

server/validation/validation.go
```diff
@@ -1,76 +1,303 @@
 package validation
```
jon-whit, line 1: Thinking about moving this package. It doesn't really belong in `server/`. I think it makes more sense for this to belong to the `tuple` package. @craigpastro @miparnisari any recommendations?

craigpastro: Either tuple or in the validations package itself. I would probably prefer the latter.

server/validation/validation_test.go
```diff
@@ -0,0 +1,425 @@
...
+	}{
+		{
+			name:        "Malformed Object 1",
```
jon-whit, line 21: rename tests here to use underscore pattern

e.g. `malformed_object_1` which will make it easier to copy/paste and search the test.

server/commands/expand.go
```diff
@@ -43,21 +40,44 @@ func (query *ExpandQuery) Execute(ctx context.Context, req *openfgapb.ExpandRequ
...
+	objectType := tupleUtils.GetType(object)
+	rel, err := typesys.GetRelation(objectType, relation)
+	if err != nil {
+		if errors.Is(err, typesystem.ErrObjectTypeUndefined) {
+			return nil, serverErrors.TypeNotFound(objectType)
+		}
+
+		if errors.Is(err, typesystem.ErrRelationUndefined) {
+			return nil, serverErrors.RelationNotFound(relation, objectType, tk)
 		}
```
miparnisari, line 73: This isn't covered by tests, but we can remove it because `validation.ValidateRelation(typesys, tk)` already throws these errors

server/test/write.go
miparnisari: I don't understand - the tuple being written in this test is referencing the `parent` relation of the `document` type, which is defined as `self`

jon-whit: 🤔 I guess I'm confused what this is testing then. At face value it looks wrong and is confusing because `docuemnt#parent` is often referred to in a tupleset relation and so writing a userset value to it seems incorrect.

Now that I look at it after having a break though, it appears as though this is just testing that if a `parent` tupleset relation exists somewhere else in the model on some other type then the write is OK. Was that the intent @miparnisari ?

storage/storage_test.go
```diff
@@ -190,3 +190,36 @@ func ExampleNewUniqueObjectIterator() {
...
+			}
+
+			// handle the error in some way
+			panic(err)
+		}
+
+		filtered = append(filtered, fmt.Sprintf("%s#%s@%s", tuple.GetObject(), tuple.GetRelation(), tuple.GetUser()))
+	}
+
+	fmt.Println(filtered)
```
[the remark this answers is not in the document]
jon-whit: This is the standard for Go Examples in tests. Go will assert the output of the print statements matches the `// Output: [document:doc1#editor@user:elbuo]` statements.

This is how the Go playground and how examples in pkg.go.dev work.

storage/storage.go
```diff
@@ -223,6 +223,49 @@ func NewStaticObjectIterator(objects []*openfgapb.Object) ObjectIterator {
...
+	filter TupleKeyFilterFunc
+}
+
+var _ TupleKeyIterator = &filteredTupleKeyIterator{}
+
+// Next() returns the next most tuple in the underlying iterator that meets
+// the filter function this iterator was constructed with.
+func (f *filteredTupleKeyIterator) Next() (*openfgapb.TupleKey, error) {
+
+	for {
```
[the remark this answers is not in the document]
jon-whit: Good point. Will replace with non-looping variant.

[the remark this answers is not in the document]
jon-whit: How do you figure? You have to walk over each tuple and filter them out (potentially filtering all of the tuples in the iterator out). 

The semantics of this iterator is to find the first tuple that doesn't need to be filtered out and return it. That may involve iterating over the whole iterator, which implies a `do/while` style of loop which is best achieved by simply using a `for` loop in Go.

server/commands/expand.go
jon-whit: We have to get the `*openfgapb.Relation` from the typesystem by calling `typesys.GetRelation` because we use it for the rewrite rule.

```
rel, err := typesys.GetRelation(...)
if err != nil {
  // this error handling code
}

userset := rel.GetRewrite()

// use the 'userset' for Expand
```

So though this error handling code is redundant because presumably you know the type and relation exists (otherwise the validation code above would have thrown an error), it's still the appropriate error handling mechanism here. If we hit an error in this code path that we didn't hit above in our validation code path then that's a serious issue. 

I'm happy to change this to a panic instead, but otherwise we should leave as is with the appropriate error handling as it is now.

server/validation/validation.go
jon-whit: Ok I've moved this into `internal/validation`.

server/validation/validation.go
```diff
@@ -1,76 +1,310 @@
...
+	if relationsForObject == nil {
+		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 {
+			// if we get here, there's a bug in the validation of WriteAuthorizationModel API
+			msg := "invalid authorization model"
+			return fmt.Errorf(msg)
+		}
+
+		// authorization model is old/unspecified and does not have type information
+		return nil
 	}
```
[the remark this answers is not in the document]
jon-whit: True

Removed this extra code that is redundant.

server/validation/validation.go
```diff
@@ -1,76 +1,310 @@
...
+
+	// at this point we know the auth model has type information
+	if userType != "" {
+		if _, ok := typesys.GetTypeDefinition(userType); !ok {
+			return &tuple.InvalidTupleError{
+				Reason:   fmt.Sprintf("undefined type for user '%s'", tk.GetUser()),
+				TupleKey: tk,
+			}
+		}
 	}
```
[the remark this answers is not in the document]
jon-whit: True

Removed the redundant code.

server/commands/check.go
```diff
@@ -110,16 +108,40 @@ func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckReques
 }
 
 // getTypeDefinitionRelationUsersets validates a tuple and returns the userset corresponding to the "object" and "relation"
-func (query *CheckQuery) getTypeDefinitionRelationUsersets(ctx context.Context, rc *resolutionContext) (*openfgapb.Userset, error) {
-	ctx, span := query.tracer.Start(ctx, "getTypeDefinitionRelationUsersets")
-	defer span.End()
+func (query *CheckQuery) getTypeDefinitionRelationUsersets(tk *openfgapb.TupleKey, typesys *typesystem.TypeSystem) (*openfgapb.Userset, error) {
```
craigpastro, line 111: Does this need to be a method on `*CheckQuery`?

jon-whit:
The filtered tuple iterator will help fix https://github.com/openfga/openfga/issues/354

internal/validation/validation.go
```diff
@@ -0,0 +1,290 @@
...
+			Reason:   fmt.Sprintf("unexpected user '%s' with tupleset relation '%s#%s'", user, objectType, relation),
+			TupleKey: tk,
+		}
+	}
+
+	return nil
+}
+
+// validateTypeRestrictions makes sure the type restrictions are enforced.
+// 1. If the tuple is of the form (user=person:bob, relation=reader, object=doc:budget), then the type "doc", relation "reader" must allow type "person".
```
craigpastro, line 120: Can we write these tuples using the Zanzibar notation, i.e., `doc:budget#reader@person:bob`?

server/commands/write.go
```diff
@@ -71,32 +71,41 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
...
-		if err := c.validateTypesForTuple(authModel, tk); err != nil {
-			return err
+		// Validate that we are not trying to write to an indirect-only relationship
+		if !typesystem.RewriteContainsSelf(relation.GetRewrite()) {
+			return serverErrors.HandleTupleValidateError(&tupleUtils.IndirectWriteError{Reason: IndirectWriteErrorReason, TupleKey: tk})
 		}
 	}
 
 	for _, tk := range deletes {
 		// For delete, we only need to ensure it is well form but no need to validate whether relation exists
```
craigpastro, line 106: While you are here, do you mind fixing this comment, or just deleting it?

server/commands/write.go
```diff
@@ -63,6 +62,7 @@ func (c *WriteCommand) validateTuplesets(ctx context.Context, req *openfgapb.Wri
 	}
 
 	var authModel *openfgapb.AuthorizationModel
```
craigpastro, line 64: Do we need to have this defined here now that we are using the typesystem in what follows?

server/commands/write.go
jon-whit: Yep! Removed. Seemed redundant, I agree.

internal/validation/validation.go
jon-whit: Yep! Changed that.

server/commands/check.go
jon-whit: Nope! It's a static method for sure. Good call. Changed it.

internal/validation/validation.go
```diff
@@ -0,0 +1,290 @@
...
+
+	rewrite := rel.GetRewrite().GetUserset()
+
+	// tupleset relation involving a rewrite
+	if rewrite != nil && reflect.TypeOf(rewrite) != reflect.TypeOf(&openfgapb.Userset_This{}) {
+		return &tuple.InvalidTupleError{
+			Reason:   fmt.Sprintf("unexpected rewrite encountered with tupelset relation '%s#%s'", objectType, relation),
+			TupleKey: tk,
+		}
+	}
```
miparnisari, line 90: Hmmm does this code belong here? It is not validating the tuple, but the rewrite itself

jon-whit: I agree, ideally it wouldn't belong here. The reason it does is because we may still be encountering developers that were victims of the recent bug reports that may have models involving rewrites on tupleset relations. At least this consolidates the validation into one place instead of having to exist in Expand, Check, etc.. individually. 

It'd be nice to remove it in favor of these validations at the `WriteAuthorizationModel` API.

server/commands/write.go
jon-whit: Nope, good call. I've moved into the containing block as a `authModel, err := ...` still variable initializer.

internal/validation/validation.go
miparnisari: Ok so we can leave this code for a while and we'll remove it.. when? Maybe leave a TODO?

jon-whit: That's the million dollar question.. I'm not sure when is best to remove this code. I think that we should just bite the bullet and remove it as soon as we have verified (if we haven't already) that `WriteAuthorizationModel` RPC no longer allows models with rewrites on tupleset relations.

server/commands/check.go
```diff
@@ -110,16 +108,40 @@ func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckReques
...
 
 // getTypeDefinitionRelationUsersets validates a tuple and returns the userset corresponding to the "object" and "relation"
-func (query *CheckQuery) getTypeDefinitionRelationUsersets(ctx context.Context, rc *resolutionContext) (*openfgapb.Userset, error) {
-	ctx, span := query.tracer.Start(ctx, "getTypeDefinitionRelationUsersets")
-	defer span.End()
+func getTypeDefinitionRelationUsersets(tk *openfgapb.TupleKey, typesys *typesystem.TypeSystem) (*openfgapb.Userset, error) {
+
+	objectType := tupleUtils.GetType(tk.GetObject())
+
+	relation, err := typesys.GetRelation(objectType, tk.GetRelation())
```
miparnisari, line 115: Can we move these two lines below the `ValidateTuple` call? And then we can remove the error handling code here because that method would have thrown

jon-whit: We can, but we still need an error handling block. Whether we panic or handle the error based on the return contract is up to us, but we shouldn't just throw away the error. I'm wondering if we should log errors here instead 🤔 ..? But then any "invalid" tuples because of model differences would be logged and that would suck.. 🤮 

Not sure the best way to handle this in a way that feels elegant at all.

server/test/write.go
miparnisari: Yes, I think so. I remember while developing this code that, when retrieving all tupleset relations, I had forgotten to check that the type of the relation matched the type of the tuple.

storage/storage.go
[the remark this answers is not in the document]
miparnisari: Ah never mind, you are right 👍

server/test/write.go
jon-whit: Alright, cool. I'll add it back.

internal/validation/validation.go
miparnisari: > WriteAuthorizationModel RPC no longer allows models with rewrites on tupleset relations.

Yes, we have that: https://github.com/openfga/openfga/blob/57d9b3da62c4634a73c47d839c6be7ac0e0b6dde/pkg/typesystem/typesystem.go#L532-L537

internal/validation/validation.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		return err
+	}
+
+	rewrite := rel.GetRewrite().GetUserset()
+
+	// tupleset relation involving a rewrite
+	if rewrite != nil && reflect.TypeOf(rewrite) != reflect.TypeOf(&openfgapb.Userset_This{}) {
+		return &tuple.InvalidTupleError{
+			Reason:   fmt.Sprintf("unexpected rewrite encountered with tupelset relation '%s#%s'", objectType, relation),
```
miparnisari, line 87:
```suggestion
			Reason:   fmt.Sprintf("unexpected rewrite encountered with tupleset relation '%s#%s'", objectType, relation),
```

pkg/tuple/validation.go
```diff
@@ -18,6 +18,11 @@ func (i *InvalidTupleError) Error() string {
 	return fmt.Sprintf("Invalid tuple '%s'. Reason: %s", i.TupleKey, i.Reason)
 }
 
+func (i *InvalidTupleError) Is(target error) bool {
```
craigpastro, line 21: Are these `Is` functions used anywhere? Are they necessary? Is this just a convenience over using `errors.As`?

jon-whit: They are used in various places in the tests `require.ErrorIs(...)` uses them.

pkg/typesystem/typesystem.go
```diff
@@ -557,7 +571,7 @@ func isUsersetRewriteValid(
 
 func validateRelationTypeRestrictions(model *openfgapb.AuthorizationModel) error {
```
craigpastro, line 572: I forgot if we talked about this, but should this function just take a typesystem (instead of an auth model)?

pkg/tuple/validation.go
jon-whit: In general the standard Go error spec outlines the `As` and `Is` functions. These implement the `Is` spec for the custom go errors we use all over.

internal/validation/validation.go
```diff
@@ -0,0 +1,290 @@
...
+			}
+		}
+	}
+
+	return &tuple.InvalidTupleError{Reason: fmt.Sprintf("User '%s' is not allowed to have relation %s with %s", tk.User, tk.Relation, tk.Object), TupleKey: tk}
+}
+
+// NoopFilterFunc returns a filter function that does not filter out any tuple provided
+// to it.
+func NoopFilterFunc() storage.TupleKeyFilterFunc {
```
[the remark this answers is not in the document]
jon-whit: It's not. I just introduced it because of standard practice of having a Noop implementation of things. I'll remove it though and we can introduce something like it when/if we need it.

pkg/tuple/validation.go
craigpastro: Is it okay that we are just checking the type? We don't need to check that the tuple and the reason are equal?

internal/validation/validation.go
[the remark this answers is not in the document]
jon-whit: Removed 👍

internal/validation/validation.go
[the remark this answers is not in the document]
jon-whit: Fixed

pkg/tuple/validation.go
jon-whit: I chose to do it like it is for the scope of this PR just to check for type equality which allows me to `require.ErrorIs(..)` in the tests and make sure the underlying error is of the specific custom error types. If we wanted our custom error check for `errors.Is(..)` to include the strings of the error itself then we'd want to do that, yeah.

In the context of where you use `errors.Is` most, you are doing so to get typed error checking semantics on wrapped custom errors. So I think implementing the `Is` method for our custom errors by just checking the type is a great way to do that.

craigpastro: Cool cool.

craigpastro (approved):
LGTM

server/commands/check.go
miparnisari: Future PR!

server/commands/expand.go
miparnisari: Future PR!

miparnisari (approved):

server/validation/validation.go
jon-whit: Follow up.

Issue #235, opened 2022-09-14
bug: conflicting authorization models cause unexpected errors


jon-whit:
Consider the following two authorization models.

**model1**
```
type group
  relations
    define member as self

type document
  relations
    define viewer as self
```

**model2**
```
type team
  relations
    define member as self

type document
  relations
    define viewer as self
```

The following writes are allowed by OpenFGA.
```
write(document:doc1, viewer, group:eng#member, model1)
write(group:eng, member, jon, model1)

write(document:doc1, viewer, team:buzz#member, model2)
write(team:buzz, member, et, model2)
```
But, when evaluating `Check(document:doc1, viewer, someuser, model1)`, I get a `400 Bad Request` error with message
```
{
    "code": "type_not_found",
    "message": "Type 'team' not found"
}
```

I would expect a 200 OK with `{"allowed": false}`, because the Check is issued as of model1 and the 'team' type definition does not pertain whatsoever to model1.

PR #342, opened 2022-11-09
fix: #235

miparnisari:
Fix for https://github.com/openfga/openfga/issues/235

### DIFF
--- a/server/commands/check.go
+++ b/server/commands/check.go
@@ -83,6 +83,11 @@ func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckReques
 	if err != nil {
 		return nil, err
 	}
+	if userset == nil {
+		// the tuple in the Check request is invalid according to the model, so throw an error
+		_, actualErr := validation.ValidateTuple(ctx, query.datastore, rc.store, rc.modelID, rc.tk)
+		return nil, serverErrors.HandleTupleValidateError(actualErr)
+	}
 
 	if err := query.resolveNode(ctx, rc, userset, typesys); err != nil {
 		return nil, err
@@ -111,7 +116,8 @@ func (query *CheckQuery) getTypeDefinitionRelationUsersets(ctx context.Context, 
 
 	userset, err := validation.ValidateTuple(ctx, query.datastore, rc.store, rc.modelID, rc.tk)
 	if err != nil {
-		return nil, serverErrors.HandleTupleValidateError(err)
+		// the tuple in the request context is invalid according to the current model, so ignore it and swallow the error
+		return nil, nil
 	}
 	return userset, nil
 }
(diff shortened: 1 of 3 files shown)


server/test/check.go
```diff
@@ -1810,6 +1807,87 @@ func TestCheckQueryAuthorizationModelsVersioning(t *testing.T, datastore storage
...
+// for the one passed in the request
+func TestCheckQueryIgnoresTuplesCurrentlyInvalid(t *testing.T, datastore storage.OpenFGADatastore) {
+	ctx := context.Background()
+	tracer := telemetry.NewNoopTracer()
+	meter := telemetry.NewNoopMeter()
+	logger := logger.NewNoopLogger()
+	store := ulid.Make().String()
+
+	// write one model with some tuples
+	oldModel := &openfgapb.AuthorizationModel{
```
jon-whit, line 1820: Can I suggest refactoring this test so it is more simplified and part of the main `TestCheckQuery` test suite?

Something like
```
func TestCheckQuery(...) {
    ...
    {
            name: "Invalid direct tuples are skipped over",
            typeDefinitions: []*openfgapb.TypeDefinition{
                    {
                            Type: "document",
                            Relations: map[string]*openfgapb.Userset{
                                    "viewer": typesystem.This(),
                            },
                    },
            },
            tuples: []*openfgapb.TupleKey{
                    tuple.NewTupleKey("group:eng", "member", "jon"),
            },
            resolveNodeLimit: defaultResolveNodeLimit,
            request: &openfgapb.CheckRequest{
                    TupleKey: tuple.NewTupleKey("document:1", "viewer", "jon"),
            },
            response: &openfgapb.CheckResponse{
                    Allowed: false,
            },
    },
```

server/commands/check.go
```diff
@@ -83,6 +83,11 @@ func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckReques
 	if err != nil {
 		return nil, err
 	}
+	if userset == nil {
```
jon-whit, line 86: Why not just add
```
objectType, _ := tupleUtils.SplitObject(object)

_, err = typesys.GetRelation(objectType, relation)
if err != nil {
    continue
}
```
below [this line of code](https://github.com/openfga/openfga/pull/342/files#diff-9f6cc2751e9bcd87e5d066d5d601ded4d87e40818657f670bc0b6d84816c63c6R272) in the `resolveDirectUserset` function?

The suggestion above makes the behavior way more obvious that if you are reading userset values from the database and the userset you received from the database doesn't validate against the model then it is skipped with the `continue` line. The way this code is written is not obvious that we're skipping userset values with an undefined objectType or relation (according to the current model), and that's because the code isn't co-located with the immediate body of code where we are reading those kind of values from the database.

server/test/check.go
miparnisari: I was wondering how these tests were going to pass when I noticed that the test is not calling `commands.WriteTuples` but `storage.Write` 🤣

server/commands/check.go
miparnisari: Good idea, but as discussed offline, i can't remove this block of code, only the `if nsUS == nil { return nil }` block.

jon-whit (approved):

PR #343, opened 2022-11-11
feat: allow idempotent writes and deletes

### LINKED ISSUE #258: WriteRequest Write is not idempotent
It would be nice that sending the same write request twice would result in the same outcome. However for the second call, the SDK will return an error saying that 
```
cannot write a tuple which already exists: <tuple>
```
However, as a user I'd expect this call to succeed as the desired end state is that the tuple exists, so if it already exists before I make the call, this is essentially a no-op to end up in the desired end state, rather than a failure/error.

craigpastro:
## Description

Closes #258.

This PR makes changes to the Write API. Writes and deletes will become idempotent. This means:
- There will be no error if you try to insert the same tuple twice. The changelog will not contain the attempted second write.
- There will be no error if you try to delete a tuple which does not exist. The deletion attempt will not appear in the changelog.

## Review Checklist

### DIFF
--- a/server/test/write.go
+++ b/server/test/write.go
@@ -364,6 +364,28 @@ var writeCommandTests = []writeCommandTest{
 		err: serverErrors.DuplicateTupleInWrite(tk),
 	},
 	{
+		_name: "ExecuteDeleteTupleWhichDoesNotExistSucceeds",
+		// state
+		model: &openfgapb.AuthorizationModel{
+			Id:            ulid.Make().String(),
+			SchemaVersion: typesystem.SchemaVersion1_0,
+			TypeDefinitions: []*openfgapb.TypeDefinition{
+				{
+					Type: "repository",
+					Relations: map[string]*openfgapb.Userset{
+						"administrator": {},
+					},
+				},
+			},
+		},
+		// input
+		request: &openfgapb.WriteRequest{
+			Deletes: &openfgapb.TupleKeys{TupleKeys: []*openfgapb.TupleKey{tk}},
+		},
+		// output
+		err: nil,
+	},
+	{
 		_name: "ExecuteWithWriteTupleWithInvalidAuthorizationModelReturnsError",
 		// state
 		model: &openfgapb.AuthorizationModel{
--- a/server/test/write.go
+++ b/server/test/write.go
@@ -364,28 +364,6 @@ var writeCommandTests = []writeCommandTest{
 		err: serverErrors.DuplicateTupleInWrite(tk),
 	},
 	{
-		_name: "ExecuteDeleteTupleWhichDoesNotExistReturnsError",
-		// state
-		model: &openfgapb.AuthorizationModel{
-			Id:            ulid.Make().String(),
-			SchemaVersion: typesystem.SchemaVersion1_0,
-			TypeDefinitions: []*openfgapb.TypeDefinition{
-				{
-					Type: "repository",
-					Relations: map[string]*openfgapb.Userset{
-						"administrator": {},
-					},
-				},
-			},
-		},
-		// input
-		request: &openfgapb.WriteRequest{
-			Deletes: &openfgapb.TupleKeys{TupleKeys: []*openfgapb.TupleKey{tk}},
-		},
-		// output
-		err: serverErrors.WriteFailedDueToInvalidInput(storage.InvalidWriteInputError(tk, openfgapb.TupleOperation_TUPLE_OPERATION_DELETE)),
-	},
-	{
 		_name: "ExecuteWithWriteTupleWithInvalidAuthorizationModelReturnsError",
 		// state
 		model: &openfgapb.AuthorizationModel{
@@ -1434,7 +1412,7 @@ var writeCommandTests = []writeCommandTest{
 	},
 }
 
-func TestWriteCommand(t *testing.T, datastore storage.OpenFGADatastore) {
+func WriteCommandTest(t *testing.T, datastore storage.OpenFGADatastore) {
 	require := require.New(t)
 	ctx := context.Background()
 	tracer := telemetry.NewNoopTracer()
--- a/storage/memory/memory.go
+++ b/storage/memory/memory.go
@@ -270,10 +270,6 @@ func (s *MemoryBackend) Write(ctx context.Context, store string, deletes storage
 
 	now := timestamppb.Now()
 
-	if err := validateTuples(s.tuples[store], deletes, writes); err != nil {
-		return err
-	}
-
 	var tuples []*openfgapb.Tuple
 
 Delete:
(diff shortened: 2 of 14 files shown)


storage/test/tuples.go
```diff
@@ -151,11 +151,10 @@ func ReadChangesTest(t *testing.T, datastore storage.OpenFGADatastore) {
 }
 
 func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore) {
-
 	ctx := context.Background()
 
-	t.Run("deletes would succeed and write would fail, fails and introduces no changes", func(t *testing.T) {
-		store := testutils.CreateRandomString(10)
+	t.Run("delete with a duplicate write succeeds", func(t *testing.T) {
```
miparnisari, line 156:
```suggestion
	t.Run("two deletes and a duplicate write succeeds", func(t *testing.T) {
```

miparnisari:
I took the liberty of renaming this PR to `feat` instead of `refactor`

storage/test/tuples.go
craigpastro: Thank you!

storage/storage.go
```diff
@@ -253,11 +253,9 @@ type TupleBackend interface {
...
 	) ([]*openfgapb.Tuple, []byte, error)
 
-	// Write updates data in the tuple backend, performing all delete operations in
-	// `deletes` before adding new values in `writes`, returning the time of the transaction, or an error.
-	// It is expected that
-	// - there is at most 10 deletes/writes
-	// - no duplicate item in delete/write list
+	// Write updates data in the tuple backend, performing all delete operations in `deletes` before writing the values
+	// in `writes`, returning the time of the transaction, or an error. It is expected that there are at most
+	// MaxTuplesInWriteOperation deletes/writes.
```
miparnisari, line 258:
```suggestion
	// Write updates data in the tuple backend, performing all delete operations in `deletes` before writing the values
	// in `writes`, returning the time of the transaction, or an error. It is expected that there are at most
	// MaxTuplesInWriteOperation deletes/writes. 
	// It's an idempotent operation: writing an existing tuple will be a no-op, and deleting a non-existing tuple is also a no-op.
```

MarceloPeresToi:
Really looking forward to this feature! 
How long would you say it would take for having it released? (no pressure hahaha) 😄

server/test/write.go
```diff
@@ -363,28 +363,6 @@ var writeCommandTests = []writeCommandTest{
 		// output
 		err: serverErrors.DuplicateTupleInWrite(tk),
 	},
-	{
-		_name: "ExecuteDeleteTupleWhichDoesNotExistReturnsError",
```
miparnisari, line 367: do we need to delete the test? can we not assert that it doesn't return an error?

miparnisari (approved):

server/test/write.go
craigpastro: Good point. I'll do that. Thanks!

storage/storage.go
craigpastro: Updated. Thank you!

miparnisari (approved):

storage/memory/memory.go
```diff
@@ -270,11 +270,8 @@ func (s *MemoryBackend) Write(ctx context.Context, store string, deletes storage
 
 	now := timestamppb.Now()
 
-	if err := validateTuples(s.tuples[store], deletes, writes); err != nil {
-		return err
-	}
-
 	var tuples []*openfgapb.Tuple
+
 Delete:
```
jon-whit, line 275: Let's refactor to drop these control flow labels. They are confusing and this logic can easily be rewritten to avoid them.

storage/test/tuples.go
```diff
@@ -193,18 +190,22 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		require.ErrorIs(t, err, storage.ErrNotFound)
 	})
 
-	t.Run("inserting a tuple twice fails", func(t *testing.T) {
+	t.Run("inserting a tuple twice won't fail", func(t *testing.T) {
 		storeID := ulid.Make().String()
 		tk := &openfgapb.TupleKey{Object: "doc:readme", Relation: "owner", User: "10"}
```
jon-whit, line 195: `tuple.NewTupleKey(...)` please

storage/test/tuples.go
```diff
@@ -151,29 +151,26 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
...
 
-	t.Run("delete fails if the tuple does not exist", func(t *testing.T) {
+	t.Run("delete does not fail if the tuple does not exist", func(t *testing.T) {
 		storeID := ulid.Make().String()
-		tk := &openfgapb.TupleKey{Object: "doc:readme", Relation: "owner", User: "10"}
-		expectedError := storage.InvalidWriteInputError(tk, openfgapb.TupleOperation_TUPLE_OPERATION_DELETE)
 
-		err := datastore.Write(ctx, storeID, []*openfgapb.TupleKey{tk}, nil)
-		require.EqualError(t, err, expectedError.Error())
+		err := datastore.Write(ctx, storeID, []*openfgapb.TupleKey{{Object: "doc:readme", Relation: "owner", User: "10"}}, nil)
```
jon-whit, line 172: `tuple.NewTupleKey(...)` please

storage/storage.go
```diff
@@ -253,11 +253,10 @@ type TupleBackend interface {
 		opts PaginationOptions,
 	) ([]*openfgapb.Tuple, []byte, error)
 
-	// Write updates data in the tuple backend, performing all delete operations in
-	// `deletes` before adding new values in `writes`, returning the time of the transaction, or an error.
-	// It is expected that
-	// - there is at most 10 deletes/writes
-	// - no duplicate item in delete/write list
+	// Write updates data in the tuple backend, performing all delete operations in `deletes` before writing the values
+	// in `writes`, returning the time of the transaction, or an error. It is expected that there are at most
```
jon-whit, line 257: We don't return the time of the transaction in this method signature. Let's omit that bit.

storage/postgres/utils_test.go
```diff
@@ -7,17 +7,12 @@ import (
 
 	"github.com/openfga/openfga/storage"
 	"github.com/stretchr/testify/require"
-	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
 
 func TestHandlePostgresError(t *testing.T) {
 	t.Run("duplicate key value error with tuple key wraps ErrInvalidWriteInput", func(t *testing.T) {
```
jon-whit, line 13: This test name should change since we've removed InvalidWriteInput.

jon-whit (approved):
LGTM

Double checked the ReadChanges API and everything and the behavior of MySQL, Postgres, and Memory stores behave as expected 👍

craigpastro:
In light of https://github.com/openfga/openfga/issues/258#issuecomment-1320969688, I'll close this for now.

PR #347, opened 2022-11-15
chore: add functional check tests

craigpastro:
## Description

Add check functional tests. Remove redundant tests. Remove unused files.

## Review Checklist


cmd/run.go
```diff
@@ -320,12 +320,12 @@ func run(_ *cobra.Command, _ []string) {
...
 	}
 
-	if err := runServer(context.Background(), config); err != nil {
+	if err := RunServer(context.Background(), config); err != nil {
 		panic(err)
 	}
 }
 
-func runServer(ctx context.Context, config *Config) error {
+func RunServer(ctx context.Context, config *Config) error {
```
jon-whit, line 328: Does this need to be publicly exported? Is it intended to be used by the public Go community? Seems like this should stay private since it's primary usage isn't a reusable function but instead is just a helper to start the server in standalone mode (when running the server from the CLI `run` entrypoint).

tests/checktestmatrix/check_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+type checkTest struct {
+	Name       string
+	Model      string
+	Tuples     []tuple
+	Assertions []assertion
+}
+
+type tuple struct {
```
jon-whit, line 35: Instead of introducing another type called `tuple` can we please use the `*openfgapb.TupleKey`.

cmd/run.go
craigpastro: I want to use this from the test. What is the alternative?

tests/checktestmatrix/check_test.go
craigpastro: Oh yeah. Why did I do that? I think I was thinking about breaking up the tuples to be more fine-grained, but I guess that doesn't have to happen now and independently.

Fixed. Thanks!

cmd/openfga/main_test.go
```diff
@@ -561,158 +560,6 @@ func GRPCCheckTest(t *testing.T, tester OpenFGATester) {
 				errorCode: codes.InvalidArgument,
 			},
 		},
-		{
-			name: "union_with_direct_check_single_operand",
-			input: &openfgapb.CheckRequest{
-				StoreId: ulid.Make().String(),
-				// AuthorizationModelId is generated automatically during testData bootstrap
-				TupleKey: tuple.NewTupleKey("repo:expressjs/express", "admin", "github|bob@openfga.com"),
```
jon-whit, line 569: If we're going to delete the tests that have `Trace: true` then let's please add the trace resolution expectation to our `yaml` test suite 👍

tests/check/check_test.go
```diff
@@ -0,0 +1,133 @@
...
+	cancel()
+}
+
+func runTests(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
+	ctx := context.Background()
+
+	for _, test := range tests.Tests {
+		for _, stage := range test.Stages {
+			t.Run(test.Name, func(t *testing.T) {
+				resp, err := client.CreateStore(ctx, &pb.CreateStoreRequest{Name: "EnsureServiceIsUp"})
```
jon-whit, line 102: This is a duplicate of the retry function above on lines 83-87.

Let's either a) use the Health Check endpoint for the "service up" check above or b) remove this code since we have a store created from our "service up" check above.

tests/check/tests.yaml
```diff
@@ -0,0 +1,1424 @@
...
+          type user
+          
+          type document
+            relations
+              define viewer: [user:*] as self
+        tuples:
+          - object: document:public
+            relation: viewer
+            user: user:*
+        assertions:
```
jon-whit, line 1419: Mind adding one more assertion for the direct check on the `user:*` tuple as well?

```
- tuple:
     object: document:public
     relation: viewer
     user: user:*
  expectation: true
```

tests/check/tests.yaml
```diff
@@ -0,0 +1,1424 @@
...
+              define member: [user] as self
+    
+          type document
+            relations
+              define viewer: [group#member] as self
+        tuples:
+          - object: document:1
+            relation: viewer
+            user: group:x#member
+        assertions:
```
jon-whit, line 1400: Mind adding one more tuple and assertion to this to make sure we can Check on `user` types as well here (the fix I submitted recently).

```
tuples:
  - object: document:1
    relation: viewer
    user: group:x#member
  - object: group:x
    relation: member
    user: user:jon

assertions:
  - tuple:
      object: document:1
      relation: viewer
      user: group:x#member
    expectation: true
  - tuple:
      object: document:1
      relation: viewer
      user: user:jon
    expectation: true
```

tests/oldcheck/tests.yaml
```diff
@@ -0,0 +1,1341 @@
...
+          user: aardvark
+        expectation: true
+
+  - name: typed_wildcard_string_is_treated_as_a_literal
+    model: |
+      type document
+        relations
+          define viewer as self
+    tuples:
+      - object: document:public
```
jon-whit, line 1333: This isn't a public document, so this is misleading. `user:*` is the literal user_id in 1.0 models, so this isn't a public document.

tests/check/check_test.go
craigpastro: Don't we want to run each test in a new store?

But, yes, the name was leftover from a refactor, so I updated that to use `test.Name`.

tests/check/tests.yaml
craigpastro: Yep! Done. Thanks.

tests/check/tests.yaml
craigpastro: Yep! Done. Thanks.

tests/oldcheck/tests.yaml
craigpastro: Fixed. Thanks!

tests/oldcheck/tests.yaml
```diff
@@ -0,0 +1,1341 @@
...
+        relation: viewer
+        user: user:*
+    assertions:
```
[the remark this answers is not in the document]
craigpastro: Ah, right. Added. Thanks!

cmd/openfga/main_test.go
craigpastro: Added trace to the test suite. Added trace to a few of the tests. Thanks!

tests/check/check_test.go
jon-whit: Oh, right. Is it cool if we use the Health check endpoint for the readiness check instead then?

Like we're doing here https://github.com/openfga/openfga/blob/main/cmd/openfga/main_test.go#L162

For some reason this threw me off because of that.

tests/check/check_test.go
```diff
@@ -0,0 +1,138 @@
...
+			t.Run(test.Name, func(t *testing.T) {
+				resp, err := client.CreateStore(ctx, &pb.CreateStoreRequest{Name: test.Name})
+				require.NoError(t, err)
+
+				storeID := resp.GetId()
+
+				_, err = client.WriteAuthorizationModel(ctx, &pb.WriteAuthorizationModelRequest{
+					StoreId:         storeID,
+					TypeDefinitions: parser.MustParse(stage.Model),
+					SchemaVersion:   "1.1",
```
jon-whit, line 111: `typesystem.SchemaVersion_1_1`

craigpastro: Ah, good call. Thanks!

jon-whit (dismissed):

jon-whit (dismissed):

jon-whit (approved):

PR #348, opened 2022-11-15
chore: update api dependency with wildcards

craigpastro:
## Review Checklist

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -128,8 +128,10 @@ func (g *ConnectedObjectGraph) findIngressesWithRewrite(
 			res = append(res, &RelationshipIngress{
 				Type: DirectIngress,
 				Ingress: &openfgapb.RelationReference{
-					Type:     target.GetType(),
-					Relation: target.GetRelation(),
+					Type: target.GetType(),
+					RelationOrWildcard: &openfgapb.RelationReference_Relation{
+						Relation: target.GetRelation(),
+					},
 				},
 			})
 		}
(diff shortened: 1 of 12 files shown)


internal/graph/graph.go
```diff
@@ -128,8 +128,10 @@ func (g *ConnectedObjectGraph) findIngressesWithRewrite(
 			res = append(res, &RelationshipIngress{
 				Type: DirectIngress,
 				Ingress: &openfgapb.RelationReference{
-					Type:     target.GetType(),
-					Relation: target.GetRelation(),
+					Type: target.GetType(),
+					RelationOrWildcard: &openfgapb.RelationReference_Relation{
```
jon-whit, line 132: Can we add a helper for this called `DirectRelationReference` in the typesystem package and also add a `WildcardRelationReference`? This is to cut down on the inline struct instantiation.

```
package typesystem

func WildcardRelationReference(type string) *openfgapb.RelationReference {
    return &openfgapb.RelationReference{
        Type: type,
        RelationOrWildcard: &openfgapb.RelationReference_Wildcard{}
    }
}

func DirectRelationReference(type, relation string) *openfgapb.RelationReference {
    return &openfgapb.RelationReference {
        Type: type,
        RelationOrWildcard: &openfgapb.RelationReference_Relation{
            Relation: relation,
        },
    }
```

Then this line becomes `typesystem.DirectRelationReference(target.GetType(), target.GetRelation())`

craigpastro: I think it helps. Added. Thanks!

craigpastro: I wonder if we should move these "util" functions to another file within the `typesystem` package though?

jon-whit (approved):

PR #350, opened 2022-11-16
feat: add wildcard validation to write model

craigpastro:
## Review Checklist


pkg/typesystem/typesystem.go
```diff
@@ -556,18 +556,17 @@ func isUsersetRewriteValid(
 }
 
 func validateRelationTypeRestrictions(model *openfgapb.AuthorizationModel) error {
-	t := New(model)
-	allTupleToUsersetDefinitions := t.GetAllTupleToUsersetsDefinitions()
+	typeSystem := New(model)
```
jon-whit, line 559: Minor nit pick - we've been using `typesys` elsewhere. Mind either changing it here to `typesys` or `typeSystem` elsewhere?

pkg/typesystem/typesystem_test.go
```diff
@@ -960,6 +960,54 @@ func TestInvalidRelationTypeRestrictionsValidations(t *testing.T) {
...
+									DirectlyRelatedUserTypes: []*openfgapb.RelationReference{
+										DirectRelationReference("user", ""),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			err: &InvalidRelationError{ObjectType: "document", Relation: "parent"},
```
jon-whit, line 1009: Should we add more context to this error so that it is more clear what the violation was due to?

For example, it'd be nice to have an error message along the lines of
> 'folder:*' is an invalid type restriction for relation 'document#parent' because it is defined under a tupleset relation.

pkg/typesystem/typesystem.go
craigpastro: Done. Thanks!

pkg/typesystem/typesystem.go
```diff
@@ -556,18 +556,17 @@ func isUsersetRewriteValid(
...
 			return err
 		}
 
 		for name, relation := range relations {
 			relatedTypes := relation.GetTypeInfo().GetDirectlyRelatedUserTypes()
-			assignable := t.IsDirectlyAssignable(relation)
+			assignable := typeSystem.IsDirectlyAssignable(relation)
 
 			if assignable && len(relatedTypes) == 0 {
 				return AssignableRelationError(objectType, name)
```
jon-whit, line 572: We have very very specific custom errors here that I feel are overkill and add to code bloat for little to no value.

I recommend that we simplify in one of two ways

1) Create a custom error such as `ModelValidationError` and then the error specific message can point out the underlying issue that caused the error such as error messages for un assignable relations or type restrictions that are invalid etc..

2) just return an error with `fmt.Errorf` with an appropriate error message.

We don't consume any specific internal error and switch on or handle the errors differently based on the custom error type, so these custom error types here are just adding to code bloat.

pkg/typesystem/typesystem.go
```diff
@@ -581,24 +580,27 @@ func validateRelationTypeRestrictions(model *openfgapb.AuthorizationModel) error
...
-				if relatedRelation != "" {
-					if _, err := t.GetRelation(relatedObjectType, relatedRelation); err != nil {
-						return InvalidRelationTypeError(objectType, name, relatedObjectType, relatedRelation)
-					}
-
-					// you cannot specify a userset if the relation is being used in a `x from y` definition (in the `y` part)
-					for _, arrayOfTtus := range allTupleToUsersetDefinitions[objectType] {
+				if related.GetRelationOrWildcard() != nil {
+					// you cannot specify a userset or wildcard if the relation is being used in a `x from y` definition (in the `y` part)
+					for _, arrayOfTtus := range typeSystem.GetAllTupleToUsersetsDefinitions()[objectType] {
```
jon-whit, line 589: Can we simplify this code by adding a helper method that we can use here and elsewhere for 'tupleset' relations.

It'd be nice if our Typesystem has a method

`IsTuplesetRelation(objectType, relation string) (bool, error)`

Then wherever we're using `GetAllTupleToUsersetDefinitions` and indexing to a specific one we can just call `IsTuplesetRelation(...)`.

This code would simplify to

```
if related.GetRelationOrWildcard() != nil {
    isTupleset, err := IsTuplesetRelation(...)
    if isTupleset {
        return fmt.Errorf("invalid type type restriction for tupleset relation ...")
    }
}
```

pkg/typesystem/typesystem_test.go
craigpastro: Yeah. I am happy to have better errors. What I am not so happy about is trying to "match" them in tests. Should we refactor the error to be something like
```
type InvalidRelationError struct {
	Reason string
}
```
and fill in the reason on a case-by-case basis? Or have a general error wrapping an
```
var InvalidRelationError = errors.New("invalid relation")
```
which we can add context to, and also match against in tests?

pkg/typesystem/typesystem.go
craigpastro: Yeah, :100: agree. Let's discuss this tomorrow, then refactor the errors here, in a future PR if necessary.

pkg/typesystem/typesystem.go
craigpastro: Good idea. I think you added this function in your validation PR.

Done. Thanks!

pkg/typesystem/typesystem.go
jon-whit: Sounds good 👍

pkg/typesystem/typesystem_test.go
jon-whit: I think until we rework our errors it would be more consistent with the existing code to have a custom error type `InvalidRelationError` with the underlying message (your first example). Then just implement the `Is(target interface{})` method and we can match in tests using `require.ErrorIs(...)`

jon-whit (approved):

PR #351, opened 2022-11-17
chore: ensure no goroutine leaks at the end of tests

miparnisari:


cmd/openfga/main_test.go
```diff
@@ -201,7 +210,6 @@ func TestFunctionalGRPC(t *testing.T) {
 
 	t.Run("TestCreateStore", func(t *testing.T) { GRPCCreateStoreTest(t, tester) })
 	t.Run("TestGetStore", func(t *testing.T) { GRPCGetStoreTest(t, tester) })
-	t.Run("TestListStores", GRPCListStoresTest) // run an isolated tester from the others so bootstrapped stores don't collide
```
miparnisari, line 204: For some reason that I could not understand, if I leave this as is, I would get a leak reported about an HTTP or GRPC connection remaining open.

cmd/run.go
```diff
@@ -459,8 +459,9 @@ func RunServer(ctx context.Context, config *Config) error {
 		}()
 	}
 
+	cachedOpenFGADatastore := caching.NewCachedOpenFGADatastore(datastore, config.Datastore.MaxCacheSize)
```
craigpastro, line 462: Can you move this to where we initialize the datastore? Line 390 or so. Just so we can keep all the datastore initialization stuff close together.

server/server_test.go
```diff
@@ -45,12 +45,14 @@ func TestOpenFGAServer(t *testing.T) {
 		uri := testDatastore.GetConnectionURI()
 		ds, err := postgres.NewPostgresDatastore(uri)
 		require.NoError(t, err)
+		defer ds.Close(context.Background())
```
craigpastro, line 48: Slightly out of scope for this PR, but would you mind removing the context argument from the `Close` function. None of the implementations use it so I don't know what it is doing there anyway.

server/server_test.go
```diff
@@ -73,12 +76,23 @@ func BenchmarkOpenFGAServer(b *testing.B) {
...
 		test.RunAllBenchmarks(b, ds)
 	})
 
 	b.Run("BenchmarkMemoryDatastore", func(b *testing.B) {
 		ds := memory.New(telemetry.NewNoopTracer(), 10, 24)
+		defer ds.Close(context.Background())
+		test.RunAllBenchmarks(b, ds)
+	})
+
+	b.Run("BenchmarkMysqlDatastore", func(b *testing.B) {
```
craigpastro, line 89:
```suggestion
	b.Run("BenchmarkMySQLDatastore", func(b *testing.B) {
```

craigpastro (approved):
LGTM

Issue #254, opened 2022-09-27
telemetry: add an OpenTelemetry metric exporter to export telemetry information


jon-whit:
The OpenFGA server captures OpenTelemetry metrics but we currently do not export them. We should add a metric exporter so that developers can hook up an OpenTelemetry collector to collect the exported metrics.

craigpastro:
Done in #360.

PR #360, opened 2022-11-23
feat: implement otlp exporter

AlexandreBrg:
This PR adds the ability to export metrics through OpenTelemetry protocol.

## Description

This PR implements a way to use metrics given by OpenFGA. The default behavior is that no metrics are exposed `NoopMeter` is used.

**How OTLP is implemented ?**

Metrics export can be either done by gRPC or by HTTP. The endpoint can also be configured. There isn't a lot of parameters implemented as the SDK can contain a lot of breaking changes through time. Moreover, OpenTelemetry SDK can be configured through environment variables (see [SDK](https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlpmetric/otlpmetrichttp/config.go#L68)).

**How to test OTLP implementation ?**

```
openfga run --metrics-exporter "otlp" --metrics-endpoint "localhost:9000" --metrics-protocol "grpc"
```

(An Otel Collector can be installed easily with [`opentelemetry-collector-contrib/examples/demo`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/examples/demo))

## References

#254 

## Review Checklist
- [ ] (Pending) I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]


craigpastro:
Wow! Thanks for this great PR. The OpenFGA team is on vacation this week, but we’ll try to get around to reviewing this early next week. Thanks again for the contribution!

jon-whit:
@AlexandreBrg thanks for contributing 👍 We've had telemetry on the todo list for a hot-minute. 

I like the approach you've taken here, and we'd love to move forward with your contribution. However, we'd like to minimize the maintenance footprint of the OpenTelemetry integration in OpenFGA. For this reason, we'd prefer to support a single OpenTelemetry exporter and encourage the community to use the OTEL Collector if they need to collect and transform to an alternative format (e.g. prometheus, jaeger, etc..).

Would you mind repurposing this PR to expose only the `otlp` exporter ?

server/commands/check.go
```diff
@@ -49,12 +49,11 @@ func NewCheckQuery(datastore storage.OpenFGADatastore, t trace.Tracer, m metric.
 
 // Execute the query in `checkRequest`, returning the response or an error.
 func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckRequest) (*openfgapb.CheckResponse, error) {
-	statCheckResolutionDepth, _ := query.meter.AsyncInt64().Gauge(
+	statCheckResolutionDepth, _ := query.meter.SyncInt64().Counter(
```
miparnisari, line 52: What is the reasoning for this change? 


 See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter

> Asynchronous Counter is an [asynchronous Instrument](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-instrument-api) which reports [monotonically](https://wikipedia.org/wiki/Monotonic_function) increasing value(s) when the instrument is being observed. Example uses for Asynchronous Counter: [CPU time](https://wikipedia.org/wiki/CPU_time)

Resolution depth is not monotonically increasing, it varies per request.

cmd/run.go
```diff
@@ -332,11 +365,28 @@ func runServer(ctx context.Context, config *Config) error {
...
+	switch config.Metrics.MetricsExporter {
+	case "prometheus":
+		meter, err = telemetry.NewPrometheusMeter()
+		if err != nil {
+			return fmt.Errorf("failed to generate prometheus exporter: %w", err)
+		}
+	case "otlp":
+		meter, err = telemetry.NewOTLPMeter(logger, ctx, config.Metrics.OTLP.Protocol, config.Metrics.OTLP.Endpoint)
+		if err != nil {
+			return fmt.Errorf("failed to generate otlp exporter: %w", err)
```
miparnisari, line 381: Very minor, but i'm confused by the naming - the variable is called `meter` but the error says `exporter`

AlexandreBrg:
> @AlexandreBrg thanks for contributing 👍 We've had telemetry on the todo list for a hot-minute.
> 
> I like the approach you've taken here, and we'd love to move forward with your contribution. However, we'd like to minimize the maintenance footprint of the OpenTelemetry integration in OpenFGA. For this reason, we'd prefer to support a single OpenTelemetry exporter and encourage the community to use the OTEL Collector if they need to collect and transform to an alternative format (e.g. prometheus, jaeger, etc..).
> 
> Would you mind repurposing this PR to expose only the `otlp` exporter ?

Thank you! I'll update the PR accordingly with your request 😄

cmd/run.go
AlexandreBrg: Indeed, I updated the error message.

server/commands/check.go
AlexandreBrg: For an unknown reason, I couldn't manage this metric to be working when using asynchronous counter. I'll try to dig into it more in the coming days

AlexandreBrg: I tried to investigate why I did this: 
- Gauge has been changed to counter because it is not available in sync mode, but counter doesn't fit here as you mentioned; 
- With an async method, I not sure why, but the gauge doesn't register any metric in this context

Even though it would be working, would that make sense to keep a gauge here ? As its values would change too frequently to be accurate or to have an average depth of queries.

.config-schema.json
```diff
@@ -119,6 +119,33 @@
                 }
             }
         },
+        "metrics": {
+            "type": "object",
+            "properties": {
+                "type": {
```
jon-whit, line 125: Since we're committing to support `otlp`, but not other exporter formats, I'm not sure it makes sense to have multiple `type` of metrics. We should maybe just change this to



```
"metrics": {
    "type": "object",
    "properties": {
        "enabled" : { "type": "boolean", "default": false },
        "otlp": {...}
    }
}
```

AlexandreBrg: Indeed, I've also moved otlp properties to `metrics`, so now it can be configured with following keys:

```yaml
metrics:
   enabled: true
   endpoint: 127.0.0.1:4817
   protocol: http
```

pkg/telemetry/metrics.go
```diff
@@ -1,9 +1,69 @@
...
 	return metric.NewNoopMeter()
 }
+
+func NewOTLPMeter(logger logger.Logger, ctx context.Context, protocol, endpoint string) (metric.Meter, error) {
+	var exporter sdkMetrics.Exporter
+	var err error
+	switch protocol {
+	case "http":
+		var options []otlpmetrichttp.Option
+		if protocol == "http" {
```
craigpastro, line 28: This `if` is redundant I think

pkg/telemetry/metrics.go
```diff
@@ -1,9 +1,69 @@
...
+
+		if endpoint != "" {
+			options = append(options, otlpmetricgrpc.WithEndpoint(endpoint))
+		}
+		exporter, err = otlpmetricgrpc.New(ctx, options...)
+	default:
+		return NewNoopMeter(), fmt.Errorf("unknown open telemetry protocol %q", protocol)
+	}
+	if err != nil {
+		return NewNoopMeter(), err
```
craigpastro, line 48: Is it better to return nil here?
```suggestion
		return nil, err
```

cmd/run.go
```diff
@@ -15,6 +15,8 @@ import (
 	"syscall"
 	"time"
 
+	"go.opentelemetry.io/otel/metric"
+
```
craigpastro, line 19: Small nit: remove space here

craigpastro (commented):
Hi @AlexandreBrg, great PR! Thank you for this work. I left a couple of very small comments. I suppose the big issue is the change from gauge to counter. I don't have an opinion yet, but I'll form one and join the conversation there.

Thanks again!

cmd/run.go
```diff
@@ -222,6 +232,11 @@ func DefaultConfig() *Config {
 			Enabled: false,
 			Addr:    ":3001",
 		},
+		Metrics: MetricsConfig{
+			Enabled:  false,
+			Protocol: "grpc",
+			Endpoint: "localhost:4317",
```
jon-whit, line 238: To be consistent with the other defaults do you mind if we bind to `0.0.0.0:4317` instead of `localhost`?

server/commands/check.go
craigpastro: I had a bit of a read and am now convinced that Gauge is correct here. As you say, there is no sync method now, but maybe there will be one soon 🤞 https://github.com/open-telemetry/opentelemetry-specification/issues/2318

I think it probably doesn't work as we are not registering the instrument in a callback. The docs for `AsyncInt64()` say
> 	// To Observe data with instruments it must be registered in a callback.

I don't know, but do we have to do something like the following instead of just calling `Observe` like we are now? 
```
_ = query.meter.RegisterCallback([]instrument.Asynchronous{statCheckResolutionDepth}, func(ctx context.Context) {
	statCheckResolutionDepth.Observe(ctx, int64(rc.metadata.GetResolve()))
})
```

Maybe it makes more sense to just remove this instrument for now until there is a sync gauge?

cmd/run.go
```diff
@@ -332,11 +347,20 @@ func RunServer(ctx context.Context, config *Config) error {
...
-	meter := telemetry.NewNoopMeter()
 	tokenEncoder := encoder.NewBase64Encoder()
 
-	var datastore storage.OpenFGADatastore
 	var err error
+	var meter metric.Meter
+	if config.Metrics.Enabled {
+		meter, err = telemetry.NewOTLPMeter(logger, ctx, config.Metrics.Protocol, config.Metrics.Endpoint)
+		if err != nil {
+			return fmt.Errorf("failed to generate otlp meter: %w", err)
```
jon-whit, line 357: nit pick

"failed to initialize otlp metrics meter: %w"

cmd/run.go
```diff
@@ -332,11 +347,20 @@ func RunServer(ctx context.Context, config *Config) error {
 
 	logger := buildLogger(config.Log.Format)
 	tracer := telemetry.NewNoopTracer()
-	meter := telemetry.NewNoopMeter()
 	tokenEncoder := encoder.NewBase64Encoder()
 
-	var datastore storage.OpenFGADatastore
 	var err error
+	var meter metric.Meter
```
jon-whit, line 353: meter := metrics.NewNoopMeter()

then set it if `config.Metrics.Enabled`.

No need for the `else {..}` below.

cmd/run.go
```diff
@@ -769,4 +793,13 @@ func bindRunFlags(cmd *cobra.Command) {
 
 	cmd.Flags().Uint32("listObjects-max-results", defaultConfig.ListObjectsMaxResults, "the maximum results to return in ListObjects responses")
 	util.MustBindPFlag("listObjectsMaxResults", cmd.Flags().Lookup("listObjects-max-results"))
+
+	cmd.Flags().Bool("metrics", defaultConfig.Metrics.Enabled, "toggle OTel metrics exporter")
```
jon-whit, line 797: `metrics-enabled` 

to be consistent with the other flags (e.g `--playground-enabled`, `--http-enabled`, etc..)

cmd/run_test.go
```diff
@@ -730,6 +730,46 @@ func TestHTTPServerEnabled(t *testing.T) {
...
+	_, err := retryablehttp.Get(fmt.Sprintf("http://%s", cfg.Metrics.Endpoint))
+	require.Error(t, err)
+	require.ErrorContains(t, err, "Get \"http://0.0.0.0:9090\": dial tcp 0.0.0.0:9090: connect: connection refused")
+}
+
+func TestOTLPExporter(t *testing.T) {
+	cfg := MustDefaultConfigWithRandomPorts()
+	cfg.HTTP.Enabled = true
+	cfg.Metrics.Enabled = true
+	cfg.Metrics.Endpoint = "localhost:4317"
```
jon-whit, line 757: The default is this value. Please omit it here since it's the default.

cmd/run_test.go
```diff
@@ -730,6 +730,46 @@ func TestHTTPServerEnabled(t *testing.T) {
...
+	cfg.Metrics.Protocol = "grpc"
```
jon-whit, line 758: grpc is the default, so please omit this

cmd/run_test.go
```diff
@@ -730,6 +730,46 @@ func TestHTTPServerEnabled(t *testing.T) {
...
+	defer cancel()
+
+	go func() {
+		if err := runServer(ctx, cfg); err != nil {
+			log.Fatal(err)
+		}
+	}()
+
+	_, err := retryablehttp.Get(fmt.Sprintf("http://%s", cfg.HTTP.Addr))
+	require.NoError(t, err)
```
jon-whit, line 770: Please add an assertion on the response code

pkg/telemetry/metrics.go
AlexandreBrg: Indeed

cmd/run.go
AlexandreBrg: ✅

AlexandreBrg:
Hello, I applied all of your reviews, thank you for the time given on this PR :)

cmd/run_test.go
```diff
@@ -730,6 +731,45 @@ func TestHTTPServerEnabled(t *testing.T) {
 	ensureServiceUp(t, cfg.GRPC.Addr, cfg.HTTP.Addr, nil, true)
 }
 
+func TestNoopMetricExporter(t *testing.T) {
+	cfg := MustDefaultConfigWithRandomPorts()
+	cfg.HTTP.Enabled = false
+	cfg.Metrics.Enabled = false
```
jon-whit, line 737: This is the default so we can omit.

pkg/telemetry/metrics.go
```diff
@@ -1,9 +1,68 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
 )
 
+const providerName = "OpenFGA"
+
 func NewNoopMeter() metric.Meter {
 	return metric.NewNoopMeter()
 }
+
+func NewOTLPMeter(logger logger.Logger, ctx context.Context, protocol, endpoint string) (metric.Meter, error) {
```
jon-whit, line 23: Go standard is for ctx to come first.

pkg/telemetry/metrics.go
```diff
@@ -1,9 +1,68 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create resource: %w", err)
+	}
+	meterProvider := sdkMetrics.NewMeterProvider(sdkMetrics.WithReader(reader), sdkMetrics.WithResource(res))
+
+	defer global.SetMeterProvider(meterProvider)
+	meter := meterProvider.Meter(providerName)
+
+	logger.Info(fmt.Sprintf("OTLP metrics send through protocol '%s'", protocol))
```
jon-whit, line 66: I'd prefer this move up in the caller instead of in the constructor.

pkg/telemetry/metrics_test.go
```diff
@@ -0,0 +1,39 @@
...
+	_, filename, _, _ := runtime.Caller(0)
+	dir := path.Join(path.Dir(filename), "../../..")
+	err := os.Chdir(dir)
+	if err != nil {
+		panic(err)
+	}
+	os.Exit(m.Run())
+}
+
+func TestNewOTLPMeterWithHttp(t *testing.T) {
```
jon-whit, line 24: HTTP

pkg/telemetry/metrics_test.go
```diff
@@ -0,0 +1,39 @@
...
+	meter, err := NewOTLPMeter(logger.NewNoopLogger(), context.Background(), "http", "localhost:4318")
+	require.NoError(t, err)
+	require.NotNil(t, meter)
+}
+
+func TestNewOTLPMeterWithGrpc(t *testing.T) {
```
jon-whit, line 30: GRPC

pkg/telemetry/metrics.go
AlexandreBrg: Sure

craigpastro (approved):
Thank you for your contribution @AlexandreBrg 🙌  This is great!

I added host and runtime metrics to your PR just so we can get it in quickly.

Thanks again!

rhamzeh:
@AlexandreBrg, we're considering replacing our OpenTelemetry metrics for Prometheus metrics. @jon-whit expanded more on our reasoning here: https://github.com/openfga/openfga/pull/514#issuecomment-1419878023

We wanted to check in with you and get your opinion on this change, please feel free to add any comments on @jon-whit's [PR here](https://github.com/openfga/openfga/pull/514)

Issue #354, opened 2022-11-18
ListObjects returning IDs not from the chosen type when contextual tuples are involved


rhamzeh:
To reproduce:

```py
type user
type resource
  relations
    define parent as self
    define viewer as viewer from parent 
type folder
  relations
    define allowed as self
    define viewer as allowed
```

With the following tuples:
```yaml
- user: folder:folder-1
  relation: parent
  object: resource:resource-1
```

Note that this issue does not appear when the relation definition is `define viewer as self or allowed` or `define viewer as self`, and it would appear if the relation definition is `define viewer as self and allowed`

To try this out, the sample store: https://play.fga.dev/stores/create/?id=01GJ41921A4W4A3DCN9A5JQEJ8

```sh
curl  -X POST 'https://api.playground.fga.dev/stores/01GJ41921A4W4A3DCN9A5JQEJ8/list-objects' \
-H "Content-Type: application/json" \
-d '{"user": "user:emily", "relation":"viewer","type":"resource","contextual_tuples":{"tuple_keys":[{"user":"user:emily", "relation":"allowed", "object":"folder:folder-1"}]}}' | jq
```

Thanks to Suresh for raising this issue [on discord](https://discord.com/channels/759188666072825867/1042596768031973466)

<img width="641" alt="Screen Shot 2022-11-18 at 5 15 02 PM" src="https://user-images.githubusercontent.com/536147/202812120-ae2a7df3-931d-4c15-bb81-26db4536ae59.png">

miparnisari:
I think the root cause is this

jon-whit:
Fixed in release `v0.3.0`

PR #366, opened 2022-11-29
fix: ListObjects returning IDs not from the chosen type when contextu…

miparnisari:
Fix for https://github.com/openfga/openfga/issues/354

### DIFF
--- a/server/commands/list_objects.go
+++ b/server/commands/list_objects.go
@@ -238,11 +238,11 @@ func (q *ListObjectsQuery) performChecks(
 	errChan chan<- error,
 	resolvedChan chan<- struct{},
 ) {
-	iter1 := storage.NewTupleKeyObjectIterator(
-		req.GetContextualTuples().GetTupleKeys(),
+	iter1 := storage.NewObjectIteratorFromTupleKeyIterator(storage.NewFilteredTupleKeyIterator(
+		storage.NewStaticTupleKeyIterator(req.GetContextualTuples().GetTupleKeys()),
 		func(tk *openfgapb.TupleKey) bool {
 			return tuple.GetType(tk.GetObject()) == req.GetType()
-		})
+		}))
 
 	iter2, err := q.Datastore.ListObjectsByType(ctx, req.GetStoreId(), req.GetType())
 	if err != nil {
--- a/server/commands/list_objects.go
+++ b/server/commands/list_objects.go
@@ -238,25 +238,26 @@ func (q *ListObjectsQuery) performChecks(
 	errChan chan<- error,
 	resolvedChan chan<- struct{},
 ) {
-	g := new(errgroup.Group)
-	g.SetLimit(maximumConcurrentChecks)
-	var objectsFound = new(uint32)
-
-	iter1 := storage.NewTupleKeyObjectIterator(req.GetContextualTuples().GetTupleKeys())
+	iter1 := storage.NewTupleKeyObjectIterator(
+		req.GetContextualTuples().GetTupleKeys(),
+		func(tk *openfgapb.TupleKey) bool {
+			return tuple.GetType(tk.GetObject()) == req.GetType()
+		})
 
 	iter2, err := q.Datastore.ListObjectsByType(ctx, req.GetStoreId(), req.GetType())
 	if err != nil {
+		iter1.Stop()
 		errChan <- err
 		return
 	}
 
 	// pass contextual tuples iterator (iter1) first to exploit uniqueness optimization
 	iter := storage.NewUniqueObjectIterator(iter1, iter2)
-	if err != nil {
-		errChan <- err
-		return
-	}
 	defer iter.Stop()
+
+	g := new(errgroup.Group)
+	g.SetLimit(maximumConcurrentChecks)
+	var objectsFound = new(uint32)
 
 	// iterate over all object IDs in the store and check if the user has relation with each
 	for {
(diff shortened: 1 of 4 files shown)


server/commands/list_objects.go
```diff
@@ -238,26 +238,27 @@ func (q *ListObjectsQuery) performChecks(
...
-	iter1 := storage.NewTupleKeyObjectIterator(req.GetContextualTuples().GetTupleKeys())
+	iter1 := storage.NewTupleKeyObjectIterator(
+		req.GetContextualTuples().GetTupleKeys(),
+		func(tk *openfgapb.TupleKey) bool {
+			return tuple.GetType(tk.GetObject()) == req.GetType()
+		})
 
 	iter2, err := q.Datastore.ListObjectsByType(ctx, req.GetStoreId(), req.GetType())
 	if err != nil {
+		iter1.Stop()
```
craigpastro, line 249: Should this be `defer`ed after its definition?

craigpastro (approved):
Left one comment, but LGTM.

server/commands/list_objects.go
miparnisari: I don't think so - under normal conditions, the iterator will be stopped in line 256. This is just for the error case, where I need to clean up manually

craigpastro: Ah, I see. Thanks!

storage/storage.go
```diff
@@ -185,13 +185,16 @@ func NewTupleKeyIteratorFromTupleIterator(iter TupleIterator) TupleKeyIterator {
 }
 
 // NewTupleKeyObjectIterator returns an ObjectIterator that iterates over the objects
-// contained in the provided list of TupleKeys.
-func NewTupleKeyObjectIterator(tupleKeys []*openfgapb.TupleKey) ObjectIterator {
+// contained in the provided list of TupleKeys. You can optionally pass in a function that
+// will remove the objects that don't match the filter.
+func NewTupleKeyObjectIterator(tupleKeys []*openfgapb.TupleKey, filterFunc TupleKeyFilterFunc) ObjectIterator {
```
jon-whit, line 190: Please don't modify this signature. We have the [`FilteredTupleKeyIterator`](https://github.com/openfga/openfga/blob/main/storage/storage.go#L261) that serves this same purpose.

There's an example of how to do it here https://github.com/openfga/openfga/blob/main/storage/storage_test.go#L194

miparnisari: The problem is that `FilteredTupleKeyIterator` returns a `TupleKeyIterator`, but in line 256 I need a `ObjectIterator`. 

I could add:

```go
func NewObjectIteratorFromTupleKeyIterator(iter TupleKeyIterator) ObjectIterator {
	objects := make([]*openfgapb.Object, 0)
	for {
		tk, err := iter.Next()
		if err == ErrIteratorDone {
			break
		}
		objectType, objectID := tuple.SplitObject(tk.GetObject())
		objects = append(objects, &openfgapb.Object{Type: objectType, Id: objectID})
	}

	return NewStaticObjectIterator(objects)
}
```

and then in ListObjects
```go
iter1 := storage.NewObjectIteratorFromTupleKeyIterator(storage.NewFilteredTupleKeyIterator(
		storage.NewStaticTupleKeyIterator(req.GetContextualTuples().GetTupleKeys()),
		func(tk *openfgapb.TupleKey) bool {
			return tuple.GetType(tk.GetObject()) == req.GetType()
		}))
iter2 := q.Datastore.ListObjectsByType(ctx, req.GetStoreId(), req.GetType())

iter := storage.NewUniqueObjectIterator(iter1, iter2)
```

Do you prefer it this way? I don't have a strong opinion.

jon-whit: Yeah, I like that chaining approach better as it is more consistent with what we've been doing with the other iterator chaining.

miparnisari: Updated!

storage/storage.go
```diff
@@ -197,6 +198,24 @@ func NewTupleKeyObjectIterator(tupleKeys []*openfgapb.TupleKey) ObjectIterator {
 	return NewStaticObjectIterator(objects)
 }
 
+func NewObjectIteratorFromTupleKeyIterator(iter TupleKeyIterator) ObjectIterator {
```
jon-whit, line 201: Can we please implement this in a way that doesn't require the extra memory to statically store the list of objects?

Instead, we should implement the `ObjectIterator` interface with a struct that uses the `TupleKeyIterator` and yields the objects from that iterator dynamically. For example,

```
type tupleKeyObjectIterator struct {
    iter TupleKeyIterator
}

var _ ObjectIterator = &tupleKeyObjectIterator{} // make sure we implement the ObjectIterator interface

func (t *tupleKeyObjectIterator) Next() (*openfgapb.Object, error) {
    tk, err := t.iter.Next()
    if err != nil {
        return err
    }

    return tk.GetObject(), nil
}
```
then this function simply becomes
```
func NewObjectIteratorFromTupleKeyIterator(iter TupleKeyIterator) ObjectIterator {
    return &tupleKeyObjectIterator{iter}
}
```

jon-whit (changes requested):
https://github.com/openfga/openfga/pull/366/files#r1037342734

storage/storage.go
miparnisari: Good idea. Thanks!

jon-whit (approved):

PR #369, opened 2022-12-02
fix: close Postgres and MySQL rows

miparnisari:


craigpastro:
In each of these cases we iterate over all of the rows and according to the documentation
> // Close closes the Rows, preventing further enumeration. If Next is called
> // and returns false and there are no further result sets,
> // the Rows are closed automatically and it will suffice to check the
> // result of Err.

so explicitly calling `Close` is unnecessary, but I don't mind being explicit. If we do add these should we also add them to the MySQL implementation as well?

jon-whit (approved):

craigpastro (approved):
LGTM

jon-whit (approved):

PR #372, opened 2022-12-02
refactor: read with no tuple in the request returns all tuples

craigpastro:
## Description

Also, remove readByStore from the storage interface

~TODO: update the API version~

## Review Checklist


storage/memory/memory.go
```diff
@@ -219,22 +216,22 @@ func (s *MemoryBackend) ReadChanges(ctx context.Context, store, objectType strin
...
-		telemetry.TraceError(span, err)
-		return nil, err
-	}
 	var matches []*openfgapb.Tuple
-	for _, t := range s.tuples[store] {
-		if match(key, t.Key) {
-			matches = append(matches, t)
+	if tk.GetObject() == "" && tk.GetRelation() == "" && tk.GetUser() == "" {
+		matches = make([]*openfgapb.Tuple, len(s.tuples[store]))
+		copy(matches, s.tuples[store])
```
jon-whit, line 229: Instead of `copy` why not just return a static iterator over the `s.tuples[store]` that you already have?

server/server.go
```diff
@@ -174,31 +174,32 @@ func (s *Server) Read(ctx context.Context, req *openfgapb.ReadRequest) (*openfga
...
+		StoreId:           store,
+		TupleKey:          tk,
+		PageSize:          req.GetPageSize(),
+		ContinuationToken: req.GetContinuationToken(),
 	})
 }
 
+// ReadTuples returns all tuples for a given store.
+//
+// Deprecated: Please use Read with an empty tuple instead.
```
jon-whit, line 188: > Please use Read without a request body instead.

server/commands/read.go
```diff
@@ -70,35 +70,3 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*o
 		ContinuationToken: encodedContToken,
 	}, nil
 }
-
-func (q *ReadQuery) validateAndAuthenticateTupleset(ctx context.Context, store, authorizationModelID string, tupleKey *openfgapb.TupleKey) error {
```
jon-whit, line 74: Glory 👏  TY

server/commands/read.go
```diff
@@ -35,25 +34,26 @@ func NewReadQuery(datastore storage.OpenFGADatastore, tracer trace.Tracer, logge
...
-// Execute the ReadQuery, returning paginated `openfga.Tuple`(s) that match the tupleset
+// Execute the ReadQuery, returning paginated `openfga.Tuple`(s) that match the tuple. Return all tuples if the tuple is
+// nil or empty.
 func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*openfgapb.ReadResponse, error) {
 	store := req.GetStoreId()
-	modelID := req.GetAuthorizationModelId()
 	tk := req.GetTupleKey()
 
+	// Restrict our reads due to some compatibility issues in one of our storage implementations.
+	if !(tk.GetObject() == "" && tk.GetRelation() == "" && tk.GetUser() == "") {
```
jon-whit, line 44: The request body
```
rpc Read(ReadRequest) ...

message TupleKey {...}

message ReadRequest {
    ...
    TupleKey tuple_key = ...;
}
```
The semantics should be such that the `tuple_key` is omitted/nil if the request body is omitted. So if `tuple_key` is non-nil, then these extra validations on the non-empty tuplekey should apply.

Hence, I vote for simplification of this validation code to simply

```
if tk != nil {
    ... // old (regular) Read validation
}
```

server/test/read.go
```diff
@@ -673,11 +567,119 @@ func ReadQueryErrorTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+		Id:            ulid.Make().String(),
+		SchemaVersion: typesystem.SchemaVersion1_0,
+		TypeDefinitions: []*openfgapb.TypeDefinition{
+			{
+				Type: "repo",
+			},
+		},
+	}
+
+	err := datastore.WriteAuthorizationModel(ctx, store, model)
```
jon-whit, line 592: Why is this write necessary if the `Read` API no longer requires an AuthorizationModel to exist before serving the Read query? Shouldn't it suffice to just write the tuples and then call `Read` without a `tuple_key` field?

server/test/read.go
```diff
@@ -673,11 +567,119 @@ func ReadQueryErrorTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+	for _, tuple := range firstResponse.Tuples {
+		receivedTuples = append(receivedTuples, tuple.Key)
+	}
+
+	secondRequest := &openfgapb.ReadRequest{StoreId: store, ContinuationToken: firstResponse.ContinuationToken}
+	secondResponse, err := cmd.Execute(ctx, secondRequest)
+	require.NoError(t, err)
+
+	require.Len(t, secondResponse.Tuples, 2)
+	require.Empty(t, secondResponse.ContinuationToken)
```
jon-whit, line 638: Shouldn't there be a third page? There are 3 tuples and the page_size=1 so there should be 3 Reads.

server/commands/read.go
craigpastro: What if the user supplied an empty tuple
```
{
  "tuple_key": {}
}
```
then `tk` is not nil is it? It is a pointer to an empty thing `*pb.TupleKey{}`?

server/server.go
craigpastro: The request body may still include pagination options. So maybe
> Please use Read with a null tuple instead
??

server/test/read.go
craigpastro: Good point. I thought I needed it for write, but that is only at the server/command layer. Done. Thanks!

server/test/read.go
craigpastro: Two requests are done. One with page_size=1 and one with page_size=2. No continuation token because that seems to be different for every endpoint 🤷

server/commands/read.go
craigpastro: As discussed offline we would like the tuple key to be `nil` in-order to read all tuples

storage/memory/memory.go
craigpastro: Because of the way that the pagination code works below, this way seems like the most straightforward.

server/test/read.go
craigpastro: Whoops. Sorry. I misread it as well. Three tuples are written, the first request has page_size=1, the second request the continuation token and the default page size (of 50) so gets back the remaining two tuples, and no continuation token.

server/commands/read.go
```diff
@@ -35,25 +34,26 @@ func NewReadQuery(datastore storage.OpenFGADatastore, tracer trace.Tracer, logge
...
 
-// Execute the ReadQuery, returning paginated `openfga.Tuple`(s) that match the tupleset
+// Execute the ReadQuery, returning paginated `openfga.Tuple`(s) that match the tuple. Return all tuples if the tuple is
+// nil or empty.
 func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*openfgapb.ReadResponse, error) {
 	store := req.GetStoreId()
-	modelID := req.GetAuthorizationModelId()
 	tk := req.GetTupleKey()
 
+	// Restrict our reads due to some compatibility issues in one of our storage implementations.
```
miparnisari, line 43: I don't understand this comment

storage/storage.go
```diff
@@ -355,13 +355,6 @@ type TupleBackend interface {
 		filter ReadStartingWithUserFilter,
 	) (TupleIterator, error)
 
-	// ReadByStore reads the tuples associated with `store`.
```
miparnisari, line 358: Can you update the documentation in line 300?

craigpastro: Good call. Thank you! Updated.

server/test/read.go
jon-whit: Ah, gotcha! Thanks for the clarification. That makes sense.

storage/memory/memory.go
jon-whit: Sounds good 👍

jon-whit (approved):
Looks good to me. Seems to work as expected.

server/commands/read.go
miparnisari: ah so it's meant to deter people from supporting other kinds of filters in the query?

craigpastro: If we want to maintain wire compatibility with Auth0 FGA, then it is meant to cause us less pain at the moment 🙂

Issue #380, opened 2022-12-06
Postgres connections are created and closed for each request


OysteinVesth:
## Bug/issue

After upgrading from `v0.2.2` to `v0.2.3` we've noticed some weird behaver, an increased amount of calls against OpenFGA have become slow (over 800ms, our client times out after that).

The postgres logs in our AWS Aurora instance shows that OpenFGA is creating and closing connections _very_ often, looks like it connects and disconnects on every request.

We first saw this problem when we went from `v0.2.1` to `v0.2.5`, but after some testing it looks like the change happened in version `v0.2.3`, might be related to the pgx upgrade.

I suspect that the decrease in performance comes from creating the connection/authenticating each time. Is there an issue with connection pooling?

### Logs and connection URI

The postgres disconnection logs for `v0.2.2` looks like this:
```sh
LOG:  disconnection: session time: 1:13:33.121 user=postgres database=openfga host=10.1.1.252 port=43411
```

After deploying `v0.2.3` we get a bunch of these in the postgres logs, with connections closing in milliseconds (often one digit ms): 
```sh
LOG:  disconnection: session time: 0:00:00.110 user=postgres database=openfga host=10.1.1.22 port=28235
```

We've made no other changes than the upgrade, and when I revert back to `v0.2.2` the issue is gone.

We use a plain postgres URI without any params/options: 
```sh
postgres://XXX:XXX@XXX.rds.amazonaws.com:5432/openfga
```

Just let me know if you need any more details. We are currently running OpenFGA in production, hopefully so hoping this will be fixed - or that we can change this through config.

rhamzeh:
Thanks for raising this issue @OysteinVesth! We'll look into it and get back to you shortly

craigpastro:
Hi @OysteinVesth. Thanks for raising this issue 🙌 

This behaviour seems to have started after we changed from `pgx` to `database/sql`. `database/sql` by default only allows two idle connections, and a check seems to use more than two connections, which it opens and then `database/sql` closes. Updating `MaxIdleConns` seems to help.

Idle connections seem to consume around [10MB of memory](https://aws.amazon.com/blogs/database/resources-consumed-by-idle-postgresql-connections/) so I think conservatively raising the `MaxIdleConns` to 5 and seeing if that helps is a good start. I'd rather not make this a config option as we already have so many, but we can certainly consider it if necessary.

WDYT?

craigpastro:
[...]

PR #381, opened 2022-12-06
feat: increase configurability of Postgres and MySQL

craigpastro:
## Description

Allow the configuration of the following `database/sql` settings:
```
db.SetMaxOpenConns
db.SetMaxIdleConns
db.SetConnMaxIdleTime
db.SetConnMaxLifetime
```

If a configuration is supplied then use the default `database/sql` settings.

This address #380. I tested with our current version and with a previous version that used pgx. I turned on connection logging in Postgres and corroborated the problem in the issue. The culprit is that `database/sql` only allows two idle connections in the pool, whereas pgx does not have such a setting, so I suppose it allows as many idle connections as the pool can hold (not verified although I did log stats and saw that there were more than two idle connections in the pool). Anyway, during a check, OpenFGA seems to open more than two connections to Postgres, and once the check finishes, `database/sql` would close those connections. Thus, the churning on connections. If I increased the number of allowed idle connections I no longer saw that churning.

Since everyone who uses OpenFGA may want to have different settings we have decided to make the settings above configurable. If no configuration is set, then it will use the default `database/sql` settings.

## References

#380.

## Review Checklist


jon-whit:
Subsequent PR will wire these up t flags in the `openfga run` CLI entrypoint?

craigpastro:
@jon-whit I'll get to that in this PR

miparnisari:
> Resolves https://github.com/openfga/openfga/issues/380.

can you put more details as to how you tested that it solves the problem?

cmd/run.go
```diff
@@ -344,6 +356,12 @@ func runServer(ctx context.Context, config *Config) error {
 		opts := []mysql.MySQLOption{
 			mysql.WithLogger(logger),
 			mysql.WithTracer(tracer),
+			mysql.WithMaxTuplesPerWrite(config.MaxTuplesPerWrite),
```
craigpastro, line 359: We must always pass these options so does it make more sense to just add them in the config?

server/commands/write.go
```diff
@@ -131,8 +131,8 @@ func (c *WriteCommand) validateNoDuplicatesAndCorrectSize(deletes []*openfgapb.T
 		tuples[key] = struct{}{}
 	}
 
-	if len(tuples) > c.datastore.MaxTuplesInWriteOperation() {
-		return serverErrors.ExceededEntityLimit("write operations", c.datastore.MaxTuplesInWriteOperation())
+	if len(tuples) > c.datastore.MaxTuplesPerWriteOperation() {
```
craigpastro, line 134: Rename the config values to be the same as the flag values.

storage/mysql/mysql.go
```diff
@@ -24,41 +24,70 @@ import (
...
+	maxOpenConns    int
+	maxIdleConns    int
+	connMaxIdleTime time.Duration
+	connMaxLifetime time.Duration
 }
 
 type MySQLOption func(*MySQL)
 
-func WithMaxTuplesInWrite(maxTuples int) MySQLOption {
+func WithTracer(t trace.Tracer) MySQLOption {
```
craigpastro, line 46: Weird diff. Just moved this a bit so that MySQL and Postgres look the same and have these option functions in order.

craigpastro:
> can you put more details as to how you tested that it solves the problem?

@miparnisari Sure. More details added in the comment. And also, good point, I won't resolve the issue, but just reference it. Thanks!

.config-schema.json
```diff
@@ -83,6 +83,26 @@
                     "description": "The maximum number of cache keys that the storage cache can store before evicting old keys.",
                     "type": "integer",
                     "default": 100000
+                },
+                "maxOpenConns": {
+                    "description": "The maximum number of open connections to the datastore.",
+                    "type": "integer",
+                    "default": "database/sql default"
```
jon-whit, line 90: People using OpenFGA might not be coming from a Go background and have knowledge of `database/sql`. Let's just be explicit with what the default is based on the `database/sql` implementation.

cmd/run.go
craigpastro: Ah, right. That makes sense. Thanks for the explanation.

cmd/run.go
```diff
@@ -741,6 +765,15 @@ func bindRunFlags(cmd *cobra.Command) {
 	cmd.Flags().Int("datastore-max-cache-size", defaultConfig.Datastore.MaxCacheSize, "the maximum number of cache keys that the storage cache can store before evicting old keys")
 	util.MustBindPFlag("datastore.maxCacheSize", cmd.Flags().Lookup("datastore-max-cache-size"))
 
+	cmd.Flags().Int("datastore-max-open-conns", defaultConfig.Datastore.MaxOpenConns, "the maximum number of open connections to the datastore")
```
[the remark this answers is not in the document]
craigpastro: These currently mirror the config names in database/sql, and I think in English "max open connections" and "connection max idle time" sound natural. Not so much "connections max open". I would prefer to keep them as it, but my mind can be changed.

.config-schema.json
craigpastro: How writing something like:
```
"default": "2 - the database/sql default"
```
If we say `2` and the underlying implementation changes that may not be very good. For example, the docs for database/sql say:
```
// The default max idle connections is currently 2. This may change in
// a future release.
```

cmd/run.go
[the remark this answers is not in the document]
jon-whit: > These currently mirror the config names in database/sql, and I think in English "max open connections" and "connection max idle time" sound natural. Not so much "connections max open".

The community using the flags of our CLI don't care so much about `database/sql` package of Go, so though it maps nicely to the configs exported out of that package, I don't think that's a compelling reason to name them that way.

I do agree it sounds more natural to say `datastore-max-open-conns`, but the objective behind the names and structure of the CLI flags is to organize by category and sub-category etc.. Since these configs deal with datastore connection settings, I think it makes sense to follow the other practices established herein and follow the structure `datastore-conn-max-open` and `datastore-conn-max-idle` because we're configuring the subcategory of datastore specific settings specifically for connection related settings. `max-idle`,  `max-idle-time`, `max-open`, `max-lifetime` are all settings specific to `datastore-conn` subcategory of settings.

I suppose I'm not strongly held to the opinion above, but we would be following status quo established herein a little better if we organized by sub-category of settings.

cmd/run.go
```diff
@@ -78,6 +78,18 @@ type DatastoreConfig struct {
 	// old keys. The storage cache is used to cache query results for various static resources
 	// such as type definitions.
 	MaxCacheSize int
+
+	// MaxOpenConns is the maximum number of open connections to the database.
+	MaxOpenConns int
+
+	// MaxIdleConns is the maximum number of connections in the idle connection pool.
```
miparnisari, line 85:
```suggestion
	// MaxIdleConns is the maximum number of connections to the datastore in the idle connection pool.
```

cmd/run.go
```diff
@@ -78,6 +78,18 @@ type DatastoreConfig struct {
...
+	MaxIdleConns int
+
+	// ConnMaxIdleTime is the maximum amount of time a connection may be idle.
```
miparnisari, line 88:
```suggestion
	// ConnMaxIdleTime is the maximum amount of time a connection to the datastore may be idle.
```

cmd/run.go
```diff
@@ -78,6 +78,18 @@ type DatastoreConfig struct {
...
+	ConnMaxIdleTime time.Duration
+
+	// ConnMaxLifetime is the maximum amount of time a connection may be reused.
```
miparnisari, line 91:
```suggestion
	// ConnMaxLifetime is the maximum amount of time a connection to the datastore may be reused.
```

miparnisari: (same suggestion for the .config-schema.json file)

cmd/run.go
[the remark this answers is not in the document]
miparnisari:
```
datastore-max-open-conns
datastore-max-idle-conns
datastore-conn-max-idle-time
datastore-conn-max-lifetime
```

The first two deal with connections (plural), the second two deal with one connection. I think it's fine the way it is

jon-whit (dismissed):

cmd/run.go
craigpastro: Thank you! Fixed.

cmd/run.go
craigpastro: Thank you! Fixed.

cmd/run.go
craigpastro: Thank you! Fixed.

jon-whit (approved):

PR #382, opened 2022-12-07
chore: refactor iterators

miparnisari:
Suggestion: click this:

image

---

This is a follow-up of https://github.com/openfga/openfga/pull/368.

The `Next()` function of the iterators was modified to take in a `context` that, when cancelled or deadlined, will make the iterator's `Next()` function return `ErrIteratorDone`. 

Load tests were run against ListObjects to ensure there is no goroutine leaks.

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -73,50 +73,39 @@ var _ ObjectIterator = (*uniqueObjectIterator)(nil)
 // Next returns the next most unique object from the two underlying iterators.
 // If the context is cancelled or times out, it should return ErrIteratorDone
 func (u *uniqueObjectIterator) Next(ctx context.Context) (*openfgapb.Object, error) {
-ForLoop:
 	for {
-		select {
-		case <-ctx.Done():
-			return nil, ErrIteratorDone
-		default:
-			obj, err := u.iter1.Next(ctx)
-			if err != nil {
-				if err == ErrIteratorDone {
-					break ForLoop
-				}
-
-				return nil, err
+		obj, err := u.iter1.Next(ctx)
+		if err != nil {
+			if err == ErrIteratorDone {
+				break
 			}
 
-			// if the object has not already been seen, then store it and return it
-			_, ok := u.objects.Load(tuple.ObjectKey(obj))
-			if !ok {
-				u.objects.Store(tuple.ObjectKey(obj), struct{}{})
-				return obj, nil
-			}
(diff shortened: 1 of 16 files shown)


storage/storage.go
```diff
@@ -52,57 +52,71 @@ type TupleKeyIterator = Iterator[*openfgapb.TupleKey]
...
 }
 
 // NewUniqueObjectIterator returns an ObjectIterator that iterates over two ObjectIterators and yields only distinct
 // objects with the duplicates removed.
 //
 // iter1 should generally be provided by a constrained iterator (e.g. contextual tuples) and iter2 should be provided
 // by a storage iterator that already guarantees uniqueness.
-func NewUniqueObjectIterator(iter1, iter2 ObjectIterator) ObjectIterator {
+// If the context is cancelled, Next() will return ErrIteratorDone
+func NewUniqueObjectIterator(iter1, iter2 ObjectIterator, ctx context.Context) ObjectIterator {
```
jon-whit, line 64: Go standard is to pass `ctx` as the first parameter. To be more consistent with Go code in general would you mind moving it to the front?

storage/storage.go
```diff
@@ -69,40 +71,52 @@ func NewUniqueObjectIterator(iter1, iter2 ObjectIterator) ObjectIterator {
...
+// If the context is cancelled or times out, it should return ErrIteratorDone
+func (u *uniqueObjectIterator) Next(ctx context.Context) (*openfgapb.Object, error) {
+ForLoop:
 	for {
-		obj, err := u.iter1.Next()
-		if err != nil {
-			if err == ErrIteratorDone {
-				break
+		select {
+		case <-ctx.Done():
```
jon-whit, line 79: Why are these necessary? The underlying iterators respect the context cancellation and this simple wrapper ontop of them will inherit the cancellation automatically.

jon-whit: For example, `u.iter1.Next(ctx)` and `u.iter2.Next(ctx)` should both be cancelled when the parent `ctx` is cancelled, and so the `ctx.Done()` herein should be redundant

jon-whit: In general our iterators should probably be adapted to be more async to handle context cancellation.

https://pauldigian.com/advanced-go-make-synchronous-code-asynchronous-with-contextcontext-and-channels

jon-whit (approved):

PR #384, opened 2022-12-07
chore: update error message

craigpastro:
## Description

Update dangling refs to Go 19.3 as well.

## Review Checklist

### DIFF
--- a/server/commands/read.go
+++ b/server/commands/read.go
@@ -44,7 +44,7 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*o
 	if tk != nil {
 		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
 		if objectType == "" || (objectID == "" && tk.GetUser() == "") {
-			return nil, serverErrors.ValidationError("to read all tuples pass an empty tuple, otherwise object type is required and both object id and user cannot be empty")
+			return nil, serverErrors.ValidationError("to read all tuples pass a null tuple, otherwise object type is required and both object id and user cannot be empty")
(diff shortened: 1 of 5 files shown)


server/commands/read.go
```diff
@@ -44,7 +44,7 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*o
 	if tk != nil {
 		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
 		if objectType == "" || (objectID == "" && tk.GetUser() == "") {
-			return nil, serverErrors.ValidationError("to read all tuples pass an empty tuple, otherwise object type is required and both object id and user cannot be empty")
+			return nil, serverErrors.ValidationError("to read all tuples pass a null tuple, otherwise object type is required and both object id and user cannot be empty")
```
jon-whit, line 47: Is the "to read all tuples pass a null tuple, ..." even necessary here? I don't feel the error message needs to teach how to use the API, that should be reserved for the API doc. 

I'd recommend we keep it directly related to the error in this code path. Something like "the 'tuple_key' field was provided but the object 'type' field is required and both the object 'id' and 'user' cannot be empty" 🤷‍♂️

craigpastro, line 47:
```suggestion
			return nil, serverErrors.ValidationError("the 'tuple_key' field was provided but the object type field is required and both the object id and user cannot be empty")
```

craigpastro: Updated! Thanks!

jon-whit (approved):

jon-whit (approved):

PR #386, opened 2022-12-08
chore: add tests to Expand API to cover 1.1 authorization models

miparnisari:
I did this:

- The tests now take a whole input model as opposed to just `TypeDefinitions`, so that I can pass in the schema version.
- I fixed the existing 1.0 tests that had weird definitions, like 
```go
Relations: map[string]*openfgapb.Userset{
	"admin": {}
},
```
- For the happy path tests, I copied the 1.0 tests and added 1.1 tests, making sure that the model and the `tuples` written in the test are OK. I did not change the expected responses.
- For the unhappy path tests, I added 1.1 tests where it made sense (i.e. not when the problem is in the request itself)


server/test/expand.go
```diff
@@ -258,30 +365,35 @@ func TestExpandQuery(t *testing.T, datastore storage.OpenFGADatastore) {
...
+				Id:            ulid.Make().String(),
+				SchemaVersion: typesystem.SchemaVersion1_0,
+				TypeDefinitions: []*openfgapb.TypeDefinition{
+					{
+						Type: "repo",
+						Relations: map[string]*openfgapb.Userset{
+							"admin": {
+								Userset: &openfgapb.Userset_TupleToUserset{
+									TupleToUserset: &openfgapb.TupleToUserset{
+										Tupleset: &openfgapb.ObjectRelation{
```
jon-whit, line 379: Here and elsewhere, let's please use the `typesystem` helpers. 

`typesystem.TupleToUserset("manager", "repo_admin")`

and the others (where applicable) `typesystem.ComputedUserset(...)` and `typesystem.This()`

server/test/expand.go
```diff
@@ -86,22 +71,78 @@ func TestExpandQuery(t *testing.T, datastore storage.OpenFGADatastore) {
...
-							Userset: &openfgapb.Userset_ComputedUserset{
-								ComputedUserset: &openfgapb.ObjectRelation{
-									Relation: "admin",
+					Object:   "repo:openfga/foo",
+					Relation: "admin",
+					User:     "user:jon",
+				},
+			},
+			request: &openfgapb.ExpandRequest{
+				TupleKey: &openfgapb.TupleKey{
```
jon-whit, line 109: Here and elsewhere would you please change these to `tuple.NewTupleKey("object", "relation", "user")` (for example).

miparnisari: Normally i'd agree but Expand in particular does not need the `User` parameter 🤷‍♀️

jon-whit: So use `""`

jon-whit (approved):

PR #387, opened 2022-12-08
fix: resolve indirect relationships with type restrictions in Check

jon-whit:
## Description
This PR fixes Check support for 1.1 models by allowing Checks to be run on types that are indirectly related. For example

```
type user

type group
  relations
    define member: [user]

type document
  relations
    define viewer: [group#member]
```
| object     | relation | user             |
|------------|----------|------------------|
| group:eng  | member   | user:jon         |
| document:1 | viewer   | group:eng#member |

`Check(document:1#viewer@user:jon)` --> `{allowed: true}`. 

Prior to this PR strict validation would cause an error to be returned because the `user` type is not directly related to `document#viewer`.

## Review Checklist

### DIFF
--- a/server/commands/check.go
+++ b/server/commands/check.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"sync"
 
+	"github.com/openfga/openfga/internal/graph"
 	"github.com/openfga/openfga/internal/utils"
 	"github.com/openfga/openfga/internal/validation"
 	"github.com/openfga/openfga/pkg/logger"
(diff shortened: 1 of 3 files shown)


server/commands/check.go
```diff
@@ -75,13 +76,60 @@ func (query *CheckQuery) Execute(ctx context.Context, req *openfgapb.CheckReques
...
+	if err := validation.ValidateRelation(typesys, tk); err != nil {
+		return nil, serverErrors.HandleTupleValidateError(err)
+	}
+
+	if err := validation.ValidateUser(typesys, tk); err != nil {
 		return nil, serverErrors.HandleTupleValidateError(err)
 	}
 
+	if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 {
+		g := graph.BuildConnectedObjectGraph(typesys)
```
jon-whit, line 92: Actually, now that I'm thinking about this more.. I think this block of code is completely unnecessary. The graph evaluation will still skip invalid tuples, but we just want to make sure we don't return an error on indirect types.

So I should be able to omit this block of code and just have the object, relation, and user validation checks above and we should be G2G. I will remove this and verify that's alright.

craigpastro (approved):
LGTM

PR #391, opened 2022-12-10
fix: proper validation for contextual tuples in Check and ListObjects

miparnisari:


internal/contextualtuples/contextual_tuples.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// Read returns the list of contextual tuples that match 'object#relation'
+func (c *ContextualTuples) Read(tk *openfgapb.TupleKey) []*openfgapb.TupleKey {
+	return c.usersets[tupleUtils.ToObjectRelationString(tk.GetObject(), tk.GetRelation())]
+}
+
+// ReadUserTuple returns, if found, a contextual tuple that matches the input 'object#relation@user' and a bool indicating
+// whether such tuple was found.
+func (c *ContextualTuples) ReadUserTuple(tk *openfgapb.TupleKey) (*openfgapb.TupleKey, bool) {
```
adriantam, line 46: Wondering why we need to return the boolean.  Could the caller just check whether *openfgapb.TupleKey is nil?  I see that we just move this function from server/commands to here ... so, this should not block to merge.

adriantam (approved):

internal/contextualtuples/contextual_tuples.go
```diff
@@ -0,0 +1,66 @@
+package contextualtuples
```
jon-whit, line 1: I'm not sure these changes warrant a whole package specific to contextual tuples. Since contextual tuples are just an extra set of `[]*openfgapb.TupleKey` provided at the request, I'd prefer that we continue to use the `validation` package to validate the tuples (like we've been doing everywhere else).

May I suggest instead of these changes that we simplify a bit to something like:

```
package validation

func ValidateTuples(typesys *typesystem.TypeSystem, tuples ...*openfgapb.TupleKey) error {
    for _, tuple := range tuples {
        if err := ValidateTuple(typesys, tuple); err != nil {
            return err
        }
    }

    return nil
}
```

We already have [`ValidateTuple`](https://github.com/openfga/openfga/blob/main/internal/validation/validation.go#L15). This approach would just extend this existing function to provide a way to validate more than one tuple at a time.

internal/contextualtuples/contextual_tuples.go
miparnisari: Yeah i moved stuff around. We can change this later. But, i worry whether we will forget to check for `nil` 🤔

internal/contextualtuples/contextual_tuples.go
miparnisari: > And, regarding the read methods, those could be enforced by passing the ContextualTuples iterator into the filtered iterator so any invalid tuples in the ContextualTuples set are filtered out just like the database iterator.

I just moved stuff around in this PR, this feels like scope creep to me 😬 Can we do this on a follow-up PR after we release today?

miparnisari: > We're not only moving packages/code we're also fixing an issue (contextual tuple validation).

I moved stuff around because the existing code validating contextual tuples in Check needed to be used also in ListObjects. So i put them in one place that could be used by both. So this is to say, it's the smallest change I thought of that could make the fix work. If I put the `New` method inside the `validation` package, where would I put the `Read` methods? They don't do validations

jon-whit: We already have storage interfaces that deal with filtering reads, and so moving this to a separate package splits the concerns. A single responsibility separation of concerns is a better approach here. Let the storage interface (the filtered tuple iterator) deal with the filtering of the reads not another `contexualtuples` package.

The validation can be re-used across Checks and ListObjects if it's shared in the `validate` package, which is shared across those methods already as well.

server/test/list_objects.go
```diff
@@ -142,6 +142,21 @@ func ListObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 				expectedResult: []string{},
 				expectedError:  nil,
 			},
+			{
+				name: "returns error if duplicate contextual tuples",
```
jon-whit, line 146: Mind adding `_` here instead of whitespace. That way this test is more searchable when Go reports the test as having failed etc..

internal/contextualtuples/contextual_tuples.go
miparnisari: Agreed offline to send another PR with this change

jon-whit (approved):

PR #392, opened 2022-12-10
feat: add more check tests

craigpastro:
## Description

More check tests focusing on validation and tests with contextual tuples.

## Review Checklist


jon-whit:
I wonder if it would be better to just merge the check tests into one. If the `errorCode` is anything other than the zero value for a successful response, then we add the error assertion logic. Otherwise we assert the assertions are correct. 

There's a lot of boilerplate that's the same in here and the division across directories adds a bit of overhead. I think it's ok to split error responses vs successful response tests, but I think if makes sense to merge. WDYT?

tests/check/check_test.go
```diff
@@ -37,9 +37,10 @@ type stage struct {
 }
 
 type assertion struct {
-	Tuple       *pb.TupleKey
-	Expectation bool
-	Trace       string
+	Tuple            *pb.TupleKey
+	ContextualTuples []*pb.TupleKey `yaml:"contextualTuples"`
```
jon-whit, line 41: What's the default behavior here? I thought `ContextTuples` would, by default, marshal to camelCase `contextualTuples`.

jon-whit: If you do add this, please also add it to `oldcheck/check_test.go`. But yeah, I think the default is this behavior so you should be able to omit it 🤔 ?

tests/validation/check_validation_test.go
```diff
@@ -0,0 +1,132 @@
...
+	ctx := context.Background()
+
+	for _, test := range tests.Tests {
+		t.Run(test.Name, func(t *testing.T) {
+			resp, err := client.CreateStore(ctx, &pb.CreateStoreRequest{Name: test.Name})
+			require.NoError(t, err)
+
+			storeID := resp.GetId()
+
+			fmt.Println(parser.MustParse(test.Model))
```
jon-whit, line 102: Residual?

tests/validation/check_validation_test.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+type checkValidationTests struct {
+	Tests []checkValidationTest
+}
+
+type checkValidationTest struct {
+	Name    string
+	Model   string
+	Tuples  []*pb.TupleKey
```
jon-whit, line 32: We don't provide any tuples in the tests since these are all request validation related errors. We can omit, yeah?

tests/validation/check_validation_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+			fmt.Println(parser.MustParse(test.Model))
+			_, err = client.WriteAuthorizationModel(ctx, &pb.WriteAuthorizationModelRequest{
+				StoreId:         storeID,
+				SchemaVersion:   typesystem.SchemaVersion1_1,
+				TypeDefinitions: parser.MustParse(test.Model),
+			})
+			require.NoError(t, err)
+
+			for _, tuple := range test.Tuples {
```
jon-whit, line 110: Similar comment above. We don't have any tuples to write if we are testing request validation errors, yeah?

jon-whit (changes requested):
Left some comments. Thanks for adding more tests, this is awesome 👍

tests/check/check_test.go
craigpastro: That's what you would think right? But it seems it is necessary.

tests/validation/check_validation_test.go
craigpastro: Yep. Thanks! Updated.

tests/validation/check_validation_test.go
craigpastro: Good point. We don't yet so I'll remove it. Thanks!

tests/validation/check_validation_test.go
craigpastro: Yep! Thanks.

tests/check/check_test.go
```diff
@@ -110,24 +120,24 @@ func runTests(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
...
+			for _, stage := range test.Stages {
 				_, err = client.WriteAuthorizationModel(ctx, &pb.WriteAuthorizationModelRequest{
 					StoreId:         storeID,
 					SchemaVersion:   typesystem.SchemaVersion1_1,
 					TypeDefinitions: parser.MustParse(stage.Model),
 				})
 				require.NoError(t, err)
 
-				for _, tuple := range stage.Tuples {
+				if len(stage.Tuples) > 0 {
```
craigpastro, line 137: I think all of our tests do not have many tuples so this is fine.

tests/check/check_test.go
```diff
@@ -18,28 +18,38 @@ import (
...
+	Stages []*stage
 }
 
+// stage is a stage of a test. All stages will be run in a single store.
 type stage struct {
 	Model      string
 	Tuples     []*pb.TupleKey
-	Assertions []assertion
+	Assertions []*assertion
+	Validation *validation
```
jon-whit, line 39: Adding `validation` here feels a bit meta. I think these test cases would be more uniform if we just added a `code` field to an assertion (default value of 0 will be `OK`). Then we can add validation test assertions whose expected code is a validation error code. 

This would make each test case logic the same and wouldn't mean there are metadata differences between individual test cases.

jon-whit: For example,

```
  - name: validation_relation_not_in_model
    stages:
      - model: |
          type user

        assertions:
          - tuple:
              object: user:aardvark
              relation: viewer
              user: user:badger
            expectation: false
            code: 2009
```

jon-whit (approved):
Looks great! Thanks for the updates 👍

PR #410, opened 2022-12-15
test: refactor `graph` package tests to use DSL model strings

jon-whit:
## Description
Refactors the `graph` package tests to use DSL model strings instead of the verbose protobuf structs

## Review Checklist


internal/graph/graph_test.go
```diff
@@ -663,42 +315,17 @@ func TestConnectedObjectGraph_RelationshipIngresss(t *testing.T) {
...
-										typesystem.DirectRelationReference("organization", "viewer"),
-									},
-								},
-							},
-						},
-					},
-				},
-			},
+			model: `
+			type organization
```
jon-whit, line 319: @craigpastro this test was actually bad before because `document#parent` was restricted to `organization#viewer` (which is an invalid type restriction for a tupleset relation).

I modified the model to make more sense, but I noticed it fails now.

Given the model
```
type organization
  relations
    define viewer: [organization]
    define can_view as viewer

type document
  relations
    define parent: [organization]
    define view as can_view from parent
```

I would expect `RelationshipIngresses(document#view, organization)` to return `["organization#viewer" (direct)]`

So the assertion is correct, but right now we're getting the response `["organization#viewer" (direct), "document#view" (direct)]`. The second in the list is erroneous. 

Mind taking a look into this?

internal/graph/graph_test.go
```diff
@@ -661,77 +313,15 @@ func TestConnectedObjectGraph_RelationshipIngresss(t *testing.T) {
 				},
 			},
 		},
-		{
```
jon-whit, line 664: Removed this test because it's an invalid test. I have a follow up PR that introduces this test back in with a proper model, but our `RelationshipIngresses` needs some work since it has some bugs in the test that replaces this.

Will follow up with this immediately after this PR.

internal/graph/graph_test.go
jon-whit: Will open a follow up PR with more tests and will fix the issues in that PR.

craigpastro (approved):
LGTM

Issue #411, opened 2022-12-15
`v0.3.0`: Write validation is being performed on assertions


rhamzeh:
### Affected Version
`v0.3.0`

### Model Used
```py
model
  schema 1.1
type user
type doc
  relations
    define owner: [user]
    define can_share: owner
```

### Steps To Reproduce
```sh
curl -X PUT 'http://127.0.0.1:8080/stores/01GM9JTY114HWA8XRCDG46K3PY/assertions/01GM9R4RGBPQC66CGPYA499G93' \
  -H 'Content-Type: application/json' \
  -d '{"assertions":[{"tuple_key":{"user":"user:henry","relation":"can_share","object":"doc:roadmap"},"expectation":true}]}'
```

### Response Expected
204 No Content

### Response Received
400 Bad Request
```json
{
  "code": "validation_error",
  "message": "Invalid tuple 'doc:roadmap#can_share@user:henry'. Reason: type 'user' is not an allowed type restriction for 'doc#can_share'"
}

```

Reported by [kdubb on Discord](https://discord.com/channels/759188666072825867/1052745870786973726)

kdubb:
Just to be clear, I encountered this issue with `0.3.0`.

adriantam:
@kdubb , the issue is now fixed and should be available in the latest "main" branch.  It will be available in the next release.

PR #412, opened 2022-12-15
fix: write assertion should not care about type restriction

adriantam:
## Description
write_assertion should not care about type restriction as it is similar to check

## References
Close https://github.com/openfga/openfga/issues/411

## Review Checklist


jon-whit (changes requested):
We shouldn't change `ValidateTuple` as it is critical for the correctness of filtering tuples from the storage iterators. Instead, let's just add the appropriate validation in the `WriteAssertions` RPC directly.

I'll take this over 👍

jon-whit (approved):

PR #416, opened 2022-12-16
fix: send the Grpc-Metadata-Openfga-Authorization-Model-Id header on …

miparnisari:
…the StreamedListObjects API

I also got rid of this logging error:

```
2022-12-09T10:22:39.185-0500	ERROR	failed to set grpc header	
{"error": "rpc error: code = Internal desc = grpc: failed to fetch the stream from the context context.Background.WithValue(type trace.traceContextKeyType, val <not Stringer>).WithValue(type trace.traceContextKeyType, val <not Stringer>)", 
"header": "openfga-authorization-model-id"}

github.com/openfga/openfga/pkg/logger.(*ZapLogger).ErrorWithContext
```

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -82,15 +82,10 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgapb.ListObjectsRequ
 
 	modelID := req.GetAuthorizationModelId()
 
-	if modelID == "" {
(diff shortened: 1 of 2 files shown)


craigpastro (approved):
LGTM

PR #417, opened 2022-12-16
Check API: add unit test and new benchmark

miparnisari:
NOT APPLICABLE ANYMORE:

> If the model is
> ```
> define team
>     define member as self
> 
> define document
>     define reader as self
> ```
> and we have N tuples like `team:N#member@anne` and one `document:budget#reader@team:999#member`
> and one direct tuple like `document:budget#reader@anne`
> 
> in this case only, Check will be faster because we will short-circuit once we find the direct path and not check all the usersets.

---

Before PR

```
BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheckWithDirectResolution-8       399           3107096 ns/op           34173 B/op     713 allocs/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkCheckWithDirectResolution-8         28429             43405 ns/op       10089 B/op         139 allocs/op
BenchmarkOpenFGAServer/BenchmarkMySQLDatastore/BenchmarkCheckWithDirectResolution-8          152          10166852 ns/op       32566 B/op         673 allocs/op
```

With this PR

```
BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheckWithDirectResolution-2 	    1876	    632637 ns/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkCheckWithDirectResolution-2              	   31150	     38395 ns/op
BenchmarkOpenFGAServer/BenchmarkMySQLDatastore/BenchmarkCheckWithDirectResolution-2               	    1082	   1030872 ns/op
```

--- 

After latest merge of main 🤔 

Without PR

```
BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheckWithDirectResolution-8       468           2455873 ns/op           49123 B/op        948 allocs/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkCheckWithDirectResolution-8        20574             53524 ns/op           15093 B/op        195 allocs/op
BenchmarkOpenFGAServer/BenchmarkMySQLDatastore/BenchmarkCheckWithDirectResolution-8          230           5367676 ns/op           46241 B/op        895 allocs/op
```

With PR

```
BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheckWithDirectResolution-8       409           2660690 ns/op           49160 B/op        949 allocs/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkCheckWithDirectResolution-8         26252             45269 ns/op           15159 B/op        196 allocs/op
BenchmarkOpenFGAServer/BenchmarkMySQLDatastore/BenchmarkCheckWithDirectResolution-8          252           5394148 ns/op           46311 B/op        896 allocs/op
```


server/commands/check.go
```diff
@@ -214,16 +214,17 @@ func (q *CheckQuery) resolveDirectUserSet(
 	rc *resolutionContext,
 	typesys *typesystem.TypeSystem,
 ) error {
-	done := make(chan struct{})
-	defer close(done)
-
 	var wg sync.WaitGroup
-	c := make(chan *chanResolveResult)
+	c := make(chan *chanResolveResult, 2)
```
miparnisari, line 218: Since now we launch two goroutines, leave enough space in the channel for both to write in it. If we leave this as is, it is possible for the 1st goroutine to write to it and the 2nd goroutine to also attempt a write and becoming blocked.

server/commands/check.go
```diff
@@ -574,26 +585,29 @@ func (q *CheckQuery) resolveTupleToUserset(
...
-	for res := range c {
-		if res.found {
+		close(errorChannel)
+	}(errorChannel)
+
+	// if the user is found we swallow any errors that may have be thrown
+	// otherwise we return the last error we found (if any)
+	for potentialError := range errorChannel {
+		if rc.userFound() {
+			go utils.DrainChannel(errorChannel)
```
jon-whit, line 606: Why drain the channel instead of just returning? Is there a way to make this code more easily understandable? It's not obvious why we're calling `DrainChannel` on the same error channel that we're already iterating over, and it seems quite likely that the `DrainChannel` could hang indefinitely waiting on the read of the error channel.

miparnisari: It's the forgotten sender problem... When we return here, we stop consuming `errorChannel`, which means that any goroutine that tries to write to the channel will become blocked. This is because `errorChannel` is unbuffered.

> Is there a way to make this code more easily understandable?

We don't know :( If there is, feel free to send an improvement

server/server_test.go
```diff
@@ -97,6 +97,60 @@ func BenchmarkOpenFGAServer(b *testing.B) {
 	})
 }
 
+func TestCheckDoesNotThrowBecauseDirectTupleWasFound(t *testing.T) {
```
miparnisari, line 100: this test fails in `main`

craigpastro (approved):
LGTM. Very nice!

jon-whit:
How did you benchmark this? If we didn't have a sufficient number of benchmark samples, then the differences between the prior runs and the latest runs might make sense.

miparnisari:
@jon-whit 

> How did you benchmark this? 

`make bench` in my local machine

> If we didn't have a sufficient number of benchmark samples, then the differences between the prior runs and the latest runs might make sense.

True, but i'm still surprised that before the merge of main the improvement gain was a lot more notorious than after..

pkg/server/server_test.go
```diff
@@ -96,6 +97,132 @@ func BenchmarkOpenFGAServer(b *testing.B) {
 	})
 }
 
+func TestCheckDoesNotThrowBecauseDirectTupleWasFound(t *testing.T) {
+	t.SkipNow() //TODO remove
+	ctx := context.Background()
+	tracer := telemetry.NewNoopTracer()
```
jon-whit, line 103: Can you please pass these inline where applicable? For example:

```
s := &Server{
    tracer: telemetry.NewNoopTracer()
    ...
}
```

pkg/server/test/check.go
```diff
@@ -388,3 +389,74 @@ func BenchmarkCheckWithTrace(b *testing.B, datastore storage.OpenFGADatastore) {
...
+
+	// add direct access to the repo
+	err = datastore.Write(ctx, store, []*openfgapb.TupleKey{}, []*openfgapb.TupleKey{{
+		User:     "anne",
+		Relation: "owner",
+		Object:   "repo:openfga",
+	}})
+	require.NoError(b, err)
+
+	checkQuery := commands.NewCheckQuery(datastore, tracer, meter, logger, defaultResolveNodeLimit)
```
jon-whit, line 444: Instead of using the `command` can you please use the server instead?

```
s := &Server{...}

resp, err := s.Check(...)
```

This way this test is portable to whatever implementation details implement `Check`. As it stands this won't translate if we refactor to use a different Check implementation, because it's only testing the `CheckQuery` command.

jon-whit (dismissed):
Left a couple of minor comments.

jon-whit:
> @jon-whit
> 
> > How did you benchmark this?
> 
> `make bench` in my local machine
> 
> > If we didn't have a sufficient number of benchmark samples, then the differences between the prior runs and the latest runs might make sense.
> 
> True, but i'm still surprised that before the merge of main the improvement gain was a lot more notorious than after..

Just to make sure, it may be worth running multiple iterations of `go bench` so that you get a more accurate approximation. 

`go -bench=... -count 5`

pkg/server/test/check.go
miparnisari: I don't understand

> because it's only testing the CheckQuery command.

This is what we want, no?

`s.Check` delegates to `commands.NewCheckQuery`.

jon-whit: Nah, we want to check `s.Check`. If we only test the command then if `s.Check` implementation changes we won't necessarily be testing what we hoped to be testing. Instead of testing the command, test the RPC endpoint.

miparnisari: I tried to make that change and I couldn't - I get a circular reference error.

But in any case

> if s.Check implementation changes

https://github.com/openfga/openfga/blob/33a5ef166936e916f540d0381048b1143363cf0b/pkg/server/server.go#L215

`s.Check` is supposed to be, and is, a delegator to the command. This shouldn't change, IMO.

miparnisari: The other two benchmarks also use `CheckCommand`, so if we are going to make this change, we should send a different PR where we update them all, IMO

craigpastro: ☝️ I agree with Maria here. If we want let's refactor all of them in a future PR.

pkg/server/server_test.go
```diff
@@ -96,6 +97,124 @@ func BenchmarkOpenFGAServer(b *testing.B) {
...
+
+	start := time.Now()
+	checkResponse, err := s.Check(ctx, &openfgapb.CheckRequest{
+		StoreId:              storeID,
+		TupleKey:             tupleKey,
+		AuthorizationModelId: modelID,
+	})
+	end := time.Since(start)
+	// we expect the Check call to be short-circuited after ReadUsersetTuples runs
+	require.Truef(t, end < 200*time.Millisecond, fmt.Sprintf("end was %s", end))
```
craigpastro, line 213: Do you think this could be flaky? Once we have added resolution depth perhaps that will be a nicer thing to compare against?

craigpastro (approved):
LGTM. Left one comment, but it is not so relevant anyway as the test is skipped for now.

pkg/server/server_test.go
miparnisari: Yeah, it could be. And yeah, if resolution depth was in the response, we could check against that :(

PR #418, opened 2022-12-16
style: put underscores in all test names

miparnisari:


craigpastro (approved):
LGTM. Thanks for this!

PR #419, opened 2022-12-16
chore: add the experimentals config to the `config-schema`

jon-whit:
## Description
Add an `--experimentals` flag to enable experimental features and use this flag to allow developers to enable/disable the ListObjects optimizations.

## Review Checklist


.config-schema.json
```diff
@@ -32,6 +32,15 @@
             "minimum": 0,
             "default": 1000
         },
+        "enableExperimentals": {
+            "description": "a list of experimental features to enable",
+            "type": "array",
+            "items": {
+                "type": "string",
+                "enum": ["openfga.v1.list-objects-optimized"]
```
craigpastro, line 40: Do we need the `openfga.v1` prefix for the flag?

cmd/run.go
```diff
@@ -461,6 +465,13 @@ func RunServer(ctx context.Context, config *Config) error {
 		}()
 	}
 
+	logger.Info(fmt.Sprintf("🧪 experimental features enabled: %v", config.ExperimentalsEnabled))
+
+	var experimentals []server.ExperimentalFeatureFlag
+	for _, feature := range config.ExperimentalsEnabled {
```
jon-whit, line 471: I looked into implementing the pflag interface for the custom `ExperimentalFeatureFlag` type and for a `[]ExperimentalFeatureFlag` it looked more complicated than it was worth just for converting a slice of strings to a custom type. So I chose this route instead.

I may revisit that here shortly and find a better way, but this is simple and straightforward and works just fine to get this functionality out.

.config-schema.json
jon-whit: The reason I added it is to distinguish a feature specific to a version of the API (hence the server package version prefix). The thinking is that experimental features may be long lived and specific to one version of an API.

jon-whit: Since this flag is planned to be so short-lived, I've renamed this experimental flag to simply `list-objects-optimized`

cmd/run.go
```diff
@@ -711,6 +723,9 @@ func bindRunFlags(cmd *cobra.Command) {
 
 	defaultConfig := DefaultConfig()
 
+	cmd.Flags().StringSlice("enable-experimentals", defaultConfig.ExperimentalsEnabled, "a list of experimental features to enable")
```
miparnisari, line 726:
```suggestion
	cmd.Flags().StringSlice("enable-experimentals", defaultConfig.ExperimentalsEnabled, "a list of experimental features to enable")
```

`enable-experimentals` sounds like a boolean. `experimentals-enabled` ?

.config-schema.json
craigpastro: I still see the old flag

cmd/run.go
jon-whit: Renamed to `--experimentals`

craigpastro (approved):
LGTM

PR #433, opened 2022-12-21
chore: pull out common code from Postgres and MySQL

craigpastro:
## Description

The next step will be to refactor both implementations to use Squirrel underneath and extract that common code too hopefully.

## Review Checklist


cmd/run.go
```diff
@@ -397,31 +397,31 @@ func RunServer(ctx context.Context, config *Config) error {
...
-		opts := []postgres.PostgresOption{
-			postgres.WithLogger(logger),
-			postgres.WithTracer(tracer),
-			postgres.WithMaxTuplesPerWrite(config.MaxTuplesPerWrite),
-			postgres.WithMaxTypesPerAuthorizationModel(config.MaxTypesPerAuthorizationModel),
-			postgres.WithMaxOpenConns(config.Datastore.MaxOpenConns),
-			postgres.WithMaxIdleConns(config.Datastore.MaxIdleConns),
-			postgres.WithConnMaxIdleTime(config.Datastore.ConnMaxIdleTime),
-			postgres.WithConnMaxLifetime(config.Datastore.ConnMaxLifetime),
+		opts := []storage.DatastoreOption{
```
jon-whit, line 416: These options here are common above as well. Let's move these above and make one set of calls instead of duplicating this code.

storage/sql.go
```diff
@@ -0,0 +1,378 @@
...
+	"github.com/openfga/openfga/pkg/telemetry"
+	tupleUtils "github.com/openfga/openfga/pkg/tuple"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	DefaultMaxTuplesPerWrite             = 100
```
jon-whit, line 24: Should we move these into `storage/storage.go` since they are more generic to the storage interface than the `sql` specific signatures?

storage/sql.go
```diff
@@ -0,0 +1,378 @@
...
+func (o *SQLObjectIterator) Stop() {
+	_ = o.rows.Close()
+}
+
+func HandleSQLError(err error, args ...interface{}) error {
+	if errors.Is(err, sql.ErrNoRows) {
+		return ErrNotFound
+	} else if errors.Is(err, ErrIteratorDone) {
+		return ErrIteratorDone
+	} else if strings.Contains(err.Error(), "duplicate key value") { // Postgres
```
jon-whit, line 361: This case and the one below (for `MySQLError`) are specific to those datastores. I'm wondering if we even need this `HandleSQLError`? I think I'd prefer to see this error handling inline in the specific code paths where these conditions to have to take place.

Also it feels weird to me to blindly convert `sql.ErrNoRows` to `ErrNotFound`, because that implies something about the caller's expected behavior. If I were to blindly handle an error by calling `HandleSQLError(err)` and I didn't want to convert `sql.ErrNoRows` to `ErrNotFound`, then this code would be working against me. This is why I believe the inline error handling approach is more obvious.

WDYT?

storage/sql.go
```diff
@@ -0,0 +1,378 @@
+package storage
```
jon-whit, line 1: Should we put this under a `sql` package?

storage/postgres/postgres.go
```diff
@@ -208,40 +81,40 @@ func (p *Postgres) ReadPage(ctx context.Context, store string, tupleKey *openfga
 	}
 	defer iter.Stop()
 
-	return iter.toArray(ctx, opts)
+	return iter.ToArray(ctx, opts)
 }
 
-func (p *Postgres) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, opts storage.PaginationOptions) (*tupleIterator, error) {
-	ctx, span := p.tracer.Start(ctx, "postgres.read")
+func (p *Postgres) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, opts storage.PaginationOptions) (*storage.SQLTupleIterator, error) {
```
jon-whit, line 87: Can we keep this more generic and just return `storage.TupleIterator`?

storage/sql.go
```diff
@@ -0,0 +1,378 @@
...
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	DefaultMaxTuplesPerWrite             = 100
+	DefaultMaxTypesPerAuthorizationModel = 100
+)
+
+type Datastore struct {
```
jon-whit, line 28: It doesn't look like we benefit much from this abstraction and it adds a little structural indirection that I don't think is worth the abstraction. The shared code here is the underlying `*sql.DB`, `tracer`, `logger`, etc.. Everything but the underlying `*sql.DB` is shared database config, which we could simply provide a helper function for to bootstrap/initialize specific sql datastore implementations. My guiding rule behind this rationale is that this `Datastore` struct doesn't provide functionality in it's own right, and so it feels weird to introduce it just to share an abstraction over an existing abstraction (e.g. `*sql.DB`) and for config simplification which could easily be achieved with helper functions instead of a struct.

I like the changes that consolidate the iterators and the config options etc.. But I think we should move back towards specific datastore implementations here instead.

craigpastro: Doesn't it allow us to have one entry point to construct a new Postgres or MySQL datastore? If we split it won't we need separate `WithTracer`, etc. functions as well? Or is that what you mean and that it is okay?

storage/sql.go
craigpastro: Ah, yes. Thank you. Done.

storage/sql.go
craigpastro: Then it would conflict with `database/sql` everywhere. I don't mind it here, but what do you think?

storage/sql.go
craigpastro: Yes, this seems like a good idea. I'll refactor. Thanks!

storage/sql.go
jon-whit: Sorry, that wasn't clear. We would consolidate the storage options into `storage.go`but then leave the rest as it was. So essentially just de-duplicating the datastore options, but leaving the rest as is.

storage/sql.go
jon-whit: Most of the signatures contained herein are specific to `database/sql` (row iterators etc..). It's ok if `sql` conflicts with `database/sql`. I see that a lot in plenty of idiomatic Go projects. Most people end up aliasing their imports. However, if we want to avoid the collision maybe a better package name would just be `common`. So `storage/common` is a package with shared/common code across various datastores (of which the row iterators etc.. can reside).

storage/postgres/postgres.go
craigpastro: It doesn't seem so. `ReadPage` uses `.toArray` which is not defined on `storage.TupleIterator`.

storage/sql.go
craigpastro: This is huge and error-prone. Mind if I do it in a follow-up?

jon-whit: Sure

storage/postgres/postgres.go
jon-whit: Seems like a good candidate to add to `storage.TupleIterator` or as a helper method `ToArray(iter storage.TupleIterator)`. Introducing another structure just to attach the `.toArray` method to is a bit overkill.

jon-whit (approved):

PR #444, opened 2023-01-04
fix: add typed wildcard support to ListObjects

jon-whit:
## Description
Add support for typed wildcards in ListObjects.

This PR ensures that typed wildcards are resolved appropriately during reverse expansion.

## Review Checklist


server/commands/connected_objects.go
```diff
@@ -21,10 +21,58 @@ type ConnectedObjectsRequest struct {
 	StoreID          string
 	ObjectType       string
 	Relation         string
-	User             *openfgapb.ObjectRelation
+	User             isUser_Ref
 	ContextualTuples []*openfgapb.TupleKey
 }
 
+type isUser_Ref interface {
```
jon-whit, line 28: I introduced this interface to distinguish between different user types. This structure resembles what would be generated from the protobuf compiler if our protobuf API distinguished between objects, typed wildcards, and object#relation users.

I like it in theory, but I'm not sure how I feel about it in terms of the code changes that were required herein.

During review, if you have alternative suggestions that you believe would make this handling more elegant, I would love to hear that feedback!

jon-whit:
@craigpastro @miparnisari as you're reviewing this please consider how it could be improved for code clarity. I don't love the ConnectedObjects reverse expansion code as it is written today. In particular, ContextualTuples are a pain in the ass, and I feel like there is a bug hidden somewhere there.

If you have any suggestions on code restructure and composition let's make it happen! 🙇

craigpastro (approved):

adriantam (changes requested):
With the following model

```
model
  schema 1.1
type user
type org
  relations
    define dept: [group]
    define dept_member: member from dept
type group
  relations
    define member: [user]
type resource
  relations
    define reader: [user,user:*,group#member] or writer
    define writer: [user,user:*,group#member,org#dept_member]

```
and the following tuples
(resource:eng_manual#writer@org:eng#dept_member)
(org:eng#dept@group:fga)
(group:fga#member@user:anne)

check shows user:anne has reader relationship with resource:eng_manual
However, list does not return relationship with 
```
{
  "type": "resource",
  "relation": "reader",
  "user": "user:anne"
}
```

server/commands/connected_objects.go
```diff
@@ -21,10 +21,58 @@ type ConnectedObjectsRequest struct {
 	StoreID          string
 	ObjectType       string
 	Relation         string
-	User             *openfgapb.ObjectRelation
+	User             isUserRef
 	ContextualTuples []*openfgapb.TupleKey
 }
 
+type isUserRef interface {
```
adriantam, line 28: since the conncted_objects.go file is getting large and the isUserRef/UserRefObject etc. are used by both connected_objects.go and list_objects.go, will it make sense to move them to a separate file and keep this file to be strictly regarding ConnectedObjectsCommand?

server/commands/connected_objects.go
```diff
@@ -92,10 +161,16 @@ func (c *ConnectedObjectsCommand) streamedConnectedObjects(
...
-					User: &openfgapb.ObjectRelation{
-						Object:   req.User.Object,
-						Relation: innerLoopIngress.Ingress.GetRelation(),
+					User: &UserRefObjectRelation{
+						ObjectRelation: &openfgapb.ObjectRelation{
+							Object:   targetUserObj,
+							Relation: innerLoopIngress.Ingress.GetRelation(),
+						},
 					},
+					// User: &openfgapb.ObjectRelation{
```
adriantam, line 170: Do we intend to have this commented out code?

server/commands/connected_objects.go
```diff
@@ -167,22 +242,52 @@ func (c *ConnectedObjectsCommand) reverseExpandTupleToUserset(
 			continue
 		}
 
-		targetUserStr := req.targetUserRef.GetObject()
+		user := t.GetUser()
```
adriantam, line 245: line 245-258 is rather similar to line 398-411?  Will it make sense to move to a function called constructTargetUserStr (or something like that)?

server/commands/connected_objects.go
jon-whit: Nope, that was residual. Thanks for pointing it out 👍  I've removed it

server/commands/connected_objects.go
jon-whit: Honestly, I introduced the `isUserRef` interface to distinguish between different types of users, but I'm not sure I like the changes. I am seriously considering going back to the `ObjectRelation` approach (like the code we had before), because it makes the code more uniform.

I'm debating what direction I want to go...

server/commands/connected_objects.go
jon-whit: Realistically this should belong in our protobuf files. Such as

```
message User {
    oneof user_ref {
        Object object = 1;
        TypedWildcard typed_wildcard = 2;
        ObjectRelation object_relation = 3;
    }
}

message Object {
    string type = 1;
    string id = 2;
}

message TypedWildcard {
    string type = 1;
}

message ObjectRelation {
    Object object = 1;
    string relation = 2;
}
```

jon-whit: Because of the scope of impact of these changes, I'd like to do this promptly in a follow up PR. Changing it in the scope of this PR has some pretty broad impact.

server/commands/connected_objects.go
jon-whit: Similar comment above.. I want to simplify/refactor this. I think I've devised a good way to do so, but changing the structure here a bit has some broad impact to the scope of this PR. I'd like to tackle this work promptly in a follow up PR.

adriantam (approved):

PR #458, opened 2023-01-18
refactor: move server and storage to pkg

craigpastro:
## Review Checklist


cmd/run.go
```diff
@@ -28,18 +28,18 @@ import (
 	"github.com/openfga/openfga/internal/gateway"
 	"github.com/openfga/openfga/internal/middleware"
 	httpmiddleware "github.com/openfga/openfga/internal/middleware/http"
+	"github.com/openfga/openfga/internal/server"
```
jon-whit, line 31: Ah man.. I forgot about this too. We don't want to do this move. If a developer wants to embed an OpenFGA server inside of their Go application then they should be able to do so. If we move the `server` package into `internal` then we remove that ability.

craigpastro: How about pkg?

jon-whit (approved):

Issue #282, opened 2022-10-14
register grpc request context tags middleware to inject common context


jon-whit:
It'd be nice to use the [`go-grpc-middleware/tags`](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/tags#pkg-overview) package to register a middleware that injects things like the request `store_id`, `authorization_model_id`, etc.. into the request context. Then we can use the injected tags in the request context in our logging and telemetry  throughout the app. 

Something like

```
opts := []grpc_ctxtags.Option{
        // TagBasedRequestFieldExtractor extracts the request field based on the Go struct tags of the generated protobufs

	grpc_ctxtags.WithFieldExtractor(grpc_ctxtags.TagBasedRequestFieldExtractor("store_id")),
        grpc_ctxtags.WithFieldExtractor(grpc_ctxtags.TagBasedRequestFieldExtractor("authorization_model_id")),

        // ... others
}

_ = grpc.NewServer(
	grpc.StreamInterceptor(grpc_ctxtags.StreamServerInterceptor(opts...)),
	grpc.UnaryInterceptor(grpc_ctxtags.UnaryServerInterceptor(opts...)),
)
```
and then those ctx tags will be available to us in the request context. For example

```
package logger

type ZapLogger struct {...}

var _ Logger = &ZapLogger{} // verify we implement the interface

func (l *ZapLogger) InfoWithContext(ctx context.Context, msg string, fields ...zap.Field) {
    ctxfields := zaptags.TagsToFields(ctx)
    fields = append(fields, ctxfields...)
    l.underlying.Info(msg, fields)
}
```

See these integrations for additional niceties:
https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/tags/zap

craigpastro:
From the comment in #468 it seems we won't be able to use this, so I'll close it for now.

PR #468, opened 2023-01-23
chore: add extra fields to log middleware

craigpastro:
## Review Checklist


internal/middleware/logging.go
```diff
@@ -14,11 +18,24 @@ func NewLoggingInterceptor(logger logger.Logger) grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 		start := time.Now()
 
+		spanCtx := trace.SpanContextFromContext(ctx)
```
jon-whit, line 21: Any reason why we shouldn't just use https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/google.golang.org/grpc/otelgrpc?

internal/middleware/logging.go
```diff
@@ -14,11 +18,24 @@ func NewLoggingInterceptor(logger logger.Logger) grpc.UnaryServerInterceptor {
...
+
+		ctx = trace.ContextWithSpanContext(ctx, spanCtx)
+
 		resp, err := handler(ctx, req)
 
 		fields := []zap.Field{
 			zap.String("method", info.FullMethod),
 			zap.Duration("took", time.Since(start)),
+			zap.String("trace_id", traceID.String()),
+			zap.String("raw_request", fmt.Sprintf("{%s}", req)),
```
jon-whit, line 38: We should json serialize the request instead of passing it straight through. The String serialization format of the request will tend to be a pain to read.

internal/middleware/logging.go
craigpastro: Oh, good point. Thanks! Updated.

internal/middleware/logging.go
craigpastro: I tried both and it doesn't seem to make much difference. In the end both look like:
```
"raw_response": "{\"id\":\"01GQK227ACEV4KTGZJT62FV5NE\",\"name\":\"github\",\"created_at\":{\"seconds\":1674603208,\"nanos\":12914000},\"updated_at\":{\"seconds\":1674603208,\"nanos\":12914000}}"
```

In this case we can save milliseconds by not marshalling to JSON 🤷

jon-whit:
r.e. `trace_id` vs `request_id` these are two different things, we should have both. A trace can span multiple requests across services and or components within a single service. It's common to correlate traces with requests, and so I think we should have both.

jon-whit:
Can we include #282 as part of some of this work? It'd be nice to register the `grpc_ctxtags` middleware and add the `store_id`, `model_id`, and other fields (as appropriate) to the context and then derive a logger with that context auto-injected.

internal/middleware/logging.go
jon-whit: I don't see the updates. Maybe you haven't pushed them yet?

internal/middleware/logging.go
craigpastro: Fixed. Thanks!

internal/middleware/logging.go
```diff
@@ -2,25 +2,56 @@ package middleware
...
+		fields = append(fields, ctxzap.TagsToFields(ctx)...)
+
+		jsonReq, err := json.Marshal(req)
+		if err == nil {
+			fields = append(fields, zap.Any("raw_request", json.RawMessage(jsonReq)))
+		}
+
+		resp, err := handler(ctx, req)
+
+		fields = append(fields, zap.Duration("took", time.Since(start)))
```
jon-whit, line 53: Can we please rename this to `request_duration`? `took` is just too generic.

internal/middleware/logging.go
```diff
@@ -2,25 +2,56 @@ package middleware
...
 		start := time.Now()
 
-		resp, err := handler(ctx, req)
+		id, err := uuid.NewRandom()
+		if err != nil {
+			logger.Error("failed to generate uuid", zap.Error(err))
+		}
+		requestID := id.String()
+
+		err = grpc.SetHeader(ctx, metadata.Pairs("X-Request-Id", requestID))
```
jon-whit, line 29: Can we move this into it's own middleware `internal/middleware/requestid`?

Inject the autogenerated request id into the context and extract it in this middleware (if it's present) and add it to the logged fields.

This way we're separating concerns. Having the request id in the context is beneficial beyond just having it for logging, so this is why I believe it warrants separation of concerns.

cmd/run.go
```diff
@@ -461,14 +466,23 @@ func RunServer(ctx context.Context, config *Config) error {
 		return fmt.Errorf("failed to initialize authenticator: %w", err)
 	}
 
+	ctxtagsOpts := []grpc_ctxtags.Option{
+		grpc_ctxtags.WithFieldExtractor(grpc_ctxtags.TagBasedRequestFieldExtractor("store_id")),
```
jon-whit, line 470: After looking at this in more detail, I don't think this package is going to work for us. For starters, it only allows extraction of one field from what I can tell. Also, it doesn't work at all even just for one tag 😄 . It seems it depends on custom struct tags generated from a different protoc plugin, and from the examples in the repo I cannot tell which it is.

I may suggest we just create a `internal/middleware/storeid` and `internal/middleware/modelid` package that extracts the StoreID and AuthorizationModelID from the request manually and injects it into the context, and then this logging package can extract it from the context (if it exists) and add it to the logged fields.

For example,
```
// internal/middleware/storeid.go
package storeid

type ctxKey string

const (
    storeIdContextKey = "store-id"
)

func ContextWithStoreId(parent context.Context, storeID string) context.Context {
    ...
}

func StoreIdFromContext(ctx context.Context) (string, bool) {
    ...
}

type requestWithStoreID interface {
    GetStoreId()
}

func NewStoreIdUnaryInterceptor() grpc.UnaryServerInterceptor {
    return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler rpc.UnaryHandler) (interface{}, error) {
        r, ok := req.(requestWithStoreId)
        if ok {
            ctx = ContextWithStoreId(ctx, r.GetStoreId())
        }
}
}
```

jon-whit (approved):

jon-whit (changes requested):
Left some comments

internal/middleware/logging.go
craigpastro: Good suggestion! Thanks! Done in #478. It will need to be reviewed first.

internal/middleware/logging.go
craigpastro: Done!

cmd/run.go
craigpastro: That would be nice actually. Then we don't even need to pass it in the methods.

internal/middleware/logging.go
miparnisari: This will break https://github.com/auth0/dwh-dbt-snowflake/blob/master/models/marts/fga/fga_store_logs.sql#L19

craigpastro: Oh... good catch!

craigpastro: @jon-whit Shall we just leave it as took?

jon-whit: Let's inject `request_duration` and `took` (same values, but different names). We can refactor references to `took` so they reference `request_duration` instead, and then we can drop `took`.

craigpastro: Added back for now. Thanks!

internal/middleware/logging.go
```diff
@@ -2,37 +2,76 @@ package middleware
...
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
 
+const (
+	methodKey          = "method"
+	requestIDKey       = "request_id"
+	traceIDKey         = "trace_id"
+	requestDurationKey = "request_duration"
+	tookKey            = "took" // TODO: remove in the future
```
elbuo8, line 21: This can be dropped all together and we can adjust either the log pipeline or the ingestion pipeline.

pkg/logger/logger.go
```diff
@@ -117,6 +117,8 @@ func NewLogger(logFormat, logLevel string) (*ZapLogger, error) {
 
 	cfg := zap.NewProductionConfig()
 	cfg.Level = zap.NewAtomicLevelAt(level)
+	cfg.EncoderConfig.TimeKey = "timestamp"
+	cfg.EncoderConfig.CallerKey = "" // removes it
```
elbuo8, line 121: Removes what?

pkg/logger/logger.go
```diff
@@ -117,6 +117,8 @@ func NewLogger(logFormat, logLevel string) (*ZapLogger, error) {
 
 	cfg := zap.NewProductionConfig()
 	cfg.Level = zap.NewAtomicLevelAt(level)
+	cfg.EncoderConfig.TimeKey = "timestamp"
```
elbuo8, line 120: Given consideration to possible side effects of this internally?

pkg/logger/logger.go
```diff
@@ -153,11 +155,3 @@ func Error(err error) zap.Field {
 func String(k, v string) zap.Field {
 	return zap.String(k, v)
 }
-
-func Int(k string, v int) zap.Field {
```
elbuo8, line 157: Considerations for how this affects internally? Or how users whom may be using this package for their custom interceptors?

Filtering with >< operators is not the same with strings vs int.

pkg/telemetry/tracing.go
```diff
@@ -1,10 +1,22 @@
 package telemetry
 
 import (
+	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/codes"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
 )
 
+func MustNewTracerProvider() trace.TracerProvider {
```
elbuo8, line 10: Why are there tracing changes in a logger focused PR?

craigpastro: Yeah, good point. I will revert this stuff and add it to the tracer PR. Thanks!

pkg/logger/logger.go
craigpastro: Removes the "caller" key from the log.

pkg/logger/logger.go
craigpastro: We are pretty locked in with zap so we just zap.Int everywhere now.

pkg/logger/logger.go
craigpastro: Good point. I think this has to be reverted. Thanks!

craigpastro: Actually, it is okay.

internal/middleware/logging.go
```diff
@@ -2,37 +2,69 @@ package middleware
...
 			return nil, err
 		}
 
-		logger.Info("grpc_complete", fields...)
+		jsonResp, err := json.Marshal(resp)
+		if err == nil {
+			fields = append(fields, zap.Any(rawResponseKey, json.RawMessage(jsonResp)))
+		}
+
+		logger.Info(grpcCompleteKey, fields...)
```
jon-whit, line 67: Debug level. This is too noisy for Info level logs.

internal/middleware/logging.go
```diff
@@ -42,25 +74,33 @@ func NewStreamingLoggingInterceptor(logger logger.Logger) grpc.StreamServerInter
...
-			fields = append(fields, zap.String("public_error", err.Error()))
-			logger.Error("grpc_error", fields...)
+			fields = append(fields, zap.String(publicErrorKey, err.Error()))
+			logger.Error(grpcErrorKey, fields...)
 
 			return err
 		}
 
-		logger.Info("grpc_complete", fields...)
+		logger.Info(grpcCompleteKey, fields...)
```
jon-whit, line 103: Same comment above. Debug is more appropriate here.

internal/middleware/logging.go
```diff
@@ -42,25 +74,33 @@ func NewStreamingLoggingInterceptor(logger logger.Logger) grpc.StreamServerInter
...
 		}
 
+		spanCtx := trace.SpanContextFromContext(stream.Context())
+		if spanCtx.HasTraceID() {
+			fields = append(fields, zap.String(traceIDKey, spanCtx.TraceID().String()))
+		}
+
+		err := handler(srv, stream)
+
+		fields = append(fields, zap.Duration(reqDurationKey, time.Since(start)))
```
jon-whit, line 90: Screen Shot 2023-02-03 at 2 20 18 PM

I see cases where request duration is `0ms`. We should investigate.

----

Also, bringing this up for further conversation for the maintainer team..

We should be more methodical about the observability componentry in OpenFGA. We have traces, metrics, and logs. The way we capture `req_duration_ms` will lead to differences in reported values as compared to the traces and metrics. For example, the traces and metrics wrap the endpoints early on in the middleware chain and this logger middleware wraps the endpoints after those middleware chains.

I'm concerned that we're reporting values across 3 different components (traces, metrics, and logs) and we should be more strategic about how we report latency based metrics in OpenFGA. I would prefer to reserve this metric for custom request duration metrics and traces. Log based metrics on this field aren't a best practice IMO.

internal/middleware/logging.go
craigpastro: I'm torn. If we do that then we will only have error logs. Which may be what we want? 🤷 Let me create a poll just to get a sense of things.

internal/middleware/logging.go
craigpastro: I think the `0ms` request duration is because it actually took less than `0.5ms` from the pov of that middleware. E.g. create store in memory gives something like:
Screen Shot 2023-02-03 at 14 28 39

craigpastro: Perhaps the default of fractions of a second is better?

jon-whit: Yeah, if we lose resolution then I would go that route. It doesn't make sense to have a `0ms` value there.

jon-whit: Also just noting that this discrepancy is "exhibit a" of my point above 😄 

> The way we capture `req_duration_ms` will lead to differences in reported values as compared to the traces and metrics.

As we can see, we get `187us` precision in the trace and we see `0ms` in the log. That's misleading at best.

craigpastro: Agreed 💯

internal/middleware/logging.go
```diff
@@ -2,65 +2,95 @@ package middleware
...
 
 	"github.com/openfga/openfga/pkg/logger"
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
 
+const (
+	methodKey       = "method"
```
jon-whit, line 15: `grpc_method` just to be precise. That will also play nicely with the other standards in the `grpc-middleware` ecosystem.

craigpastro: Done. Thanks!

internal/middleware/logging.go
```diff
@@ -2,65 +2,103 @@ package middleware
...
 
+const (
+	grpcServiceKey  = "grpc_service"
+	grpcMethodKey   = "grpc_method"
+	requestIDKey    = "request_id"
+	traceIDKey      = "trace_id"
+	rawRequestKey   = "raw_request"
+	rawResponseKey  = "raw_response"
+	publicErrorKey  = "public_error"
+	grpcErrorKey    = "grpc_error"
```
jon-whit, line 23: I also recommend we introduce this as `grpc_code` and it's values will be things like `0`, `Code(2022)` (for our custom error codes 😢 ), `InvalidArgument`, etc... 


Can we also introduce `grpc_type` to capture the kind of grpc request.

```
grpc_type: {
    enum: ["unary", "client_stream", "server_stream", "bidi_stream"]
}
```

internal/middleware/logging.go
```diff
@@ -2,65 +2,103 @@ package middleware
...
 
 	"github.com/openfga/openfga/pkg/logger"
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 )
 
+const (
```
jon-whit, line 15: I like this approach with the `grpc_` prefix, because I can see the community asking us to add logged fields for the HTTP variations, which we'd have to implement in the grpc-gateway `runtime` middleware for the grpc-gateway HTTP --> gRPC transcoding process. In that case we'd also introduce `http_method`, `http_path`, etc..

So we'd have example values such as

`grpc_service: "openfga.v1.OpenFGAService"`
`grpc_method: "Check"`
`grpc_type: "unary"
`grpc_code: "OK"`, `grpc_code: "Code(2002)"`, etc..

`http_method: "POST"`
`http_path: "/stores/:id/check"`
`http_status: 200` etc..

pkg/logger/logger.go
elbuo8: Might as well strip the others and fully adopt zap.

craigpastro: Good point. Done. Thanks!

internal/middleware/logging.go
craigpastro: grpc_code and grpc_type introduced. Thanks!

internal/middleware/logging.go
```diff
@@ -2,66 +2,129 @@ package middleware
...
+	grpcServiceKey   = "grpc_service"
+	grpcMethodKey    = "grpc_method"
+	grpcTypeKey      = "grpc_type"
+	grpcCodeKey      = "grpc_code"
+	requestIDKey     = "request_id"
+	traceIDKey       = "trace_id"
+	rawRequestKey    = "raw_request"
+	rawResponseKey   = "raw_response"
+	internalErrorKey = "internal_error"
+	grpcCompleteKey  = "grpc_complete"
```
jon-whit, line 26: `grpc_request_complete` ?? `grpc_complete` is pretty vague. Does it mean the server has completed start up or does it mean request complete or does it mean some other grpc component has finalized it's initialization etc..?

jon-whit (approved):

jon-whit (approved):

PR #471, opened 2023-01-25
fix: ensure derived ctx is passed

craigpastro:
## Review Checklist


pkg/server/commands/list_objects.go
```diff
@@ -284,8 +284,8 @@ func (q *ListObjectsQuery) performChecks(ctx context.Context, req listObjectsReq
 	iter := storage.NewUniqueObjectIterator(iter1, iter2)
 	defer iter.Stop()
 
-	subg, subgctx := errgroup.WithContext(ctx)
-	subg.SetLimit(maximumConcurrentChecks)
+	g, ctx := errgroup.WithContext(ctx)
```
jon-whit, line 287: How does this fix it? It appears this is merely renaming the variable?

In the old case `subgctx` was derived from `ctx` and then subsequently used. Is that not the same behavior as derived a context from `ctx` and then reassigning?

craigpastro: `ctx` was used in `Next` (on line 292) which calls `fetchResults`. So renaming `subgctx` ensures that the correct context is propagated.

jon-whit: Gotcha 👍 I didn't see that bit in the diff.

jon-whit (approved):

PR #474, opened 2023-01-26
perf: thin context wrapper over datastore

craigpastro:
## Description

In a future PR I will add a timeout context to the datastore calls, but this requires removing the ctx from the datastore iterators.

## References

#380

## Review Checklist

### DIFF
--- a/pkg/storage/context.go
+++ b/pkg/storage/context.go
@@ -1,0 +1,73 @@
+package storage
+
+import (
+	"context"
+	"time"
+
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/trace"
+)
+
+var (
+	queryTimeout = 500 * time.Millisecond
+)
+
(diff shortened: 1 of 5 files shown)


pkg/storage/context.go
```diff
@@ -0,0 +1,67 @@
...
+
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// ContextTracerWrapper is a wrapper around a datastore that passes a new
+// context to the underlying datastore methods. It must be the first wrapper
+// around the datastore if traces are to work properly.
+type ContextTracerWrapper struct {
+	OpenFGADatastore
```
jon-whit, line 14: If we're going to wrap the whole datastore here, should we wrap every method of the datastore? If not, can we just wrap the `RelationshipTupleReader` interface instead?

My fear is that we'll blindly think that the whole datastore is wrapped with the alternative context, but to our surprise it won't be.

jon-whit (approved):
Rebase then this looks good 👍

PR #476, opened 2023-01-27
refactor: reimplement Check with more functional composition

### LINKED ISSUE #300: Sporadic authorization_model_resolution_too_complex error
Docker on Windows
OpenFGA v.0.2.2

I have a reproducible model, [dataset](https://github.com/openfga/openfga/files/9824865/complex_error_reproducible.zip), and check that gives me the following sporadic error;
```json
{
    "code": "authorization_model_resolution_too_complex",
    "message": "Authorization Model resolution required too many rewrite rules to be resolved. Check your authorization model for infinite recursion or too much nesting"
}
```  

I make the following Check call via POSTMAN
```json
{
    "tuple_key": {
      "object": "org:f708c725-b072-4678-87d1-7fc47c52ffd3",
      "relation": "member",
      "user": "a2226717-7702-480a-a4ce-4608d0dcaea5"
    },
    "contextual_tuples": {
      "tuple_keys": [
        {
          "object": "org:f708c725-b072-4678-87d1-7fc47c52ffd3",
          "relation": "user_in_context",
          "user": "a2226717-7702-480a-a4ce-4608d0dcaea5"
        }
      ]
    } 
  }
```
or Curl
```curl
curl --location --request POST 'http://localhost:3601/stores/01GFS6AQP6WTBCHNNQR7YSA9AW/check' \
--data-raw '{
    "tuple_key": {
      "object": "org:f708c725-b072-4678-87d1-7fc47c52ffd3",
      "relation": "member",
      "user": "a2226717-7702-480a-a4ce-4608d0dcaea5"
    },
    "contextual_tuples": {
      "tuple_keys": [
        {
          "object": "org:f708c725-b072-4678-87d1-7fc47c52ffd3",
          "relation": "user_in_context",
          "user": "a2226717-7702-480a-a4ce-4608d0dcaea5"
        }
      ]
    } 
  }'
```

My Authorization Model
```
type feature
  relations
    define access as subscriber_member from associated_plan
    define associated_plan as self
    define subscriber as subscriber from associated_plan
type feature_repo
  relations
    define feature as self
type integrity
  relations
    define member as self
type org
  relations
    define admin as self and member
    define member as self and member from user_repo and user_in_context
    define plan_provider as self
    define user_in_context as self
    define user_repo as self
type permission
  relations
    define access as access from associated_feature and member from associated_permission_group
    define associated_feature as self
    define associated_permission_group as self
    define org_access as subscriber from associated_feature
type permission_group
  relations
    define member as self or admin from owner
    define owner as self
type plan
  relations
    define subscriber as self
    define subscriber_member as member from subscriber
type plan_provider
  relations
    define plan as self and plan from plan_repo and plan_in_context
    define plan_in_context as self
    define plan_repo as self
type plan_repo
  relations
    define plan as self
type user_repo
  relations
    define integrity as self
    define member as self and member from integrity

```
[complex_error_reproducible.zip](https://github.com/openfga/openfga/files/9824865/complex_error_reproducible.zip)

### LINKED ISSUE #413: Performance improvement possible in resolveDirectUserSet
One of the short circuit exits is not kicking in for `resolveDirectUserSet `.

https://github.com/openfga/openfga/blob/aabbabf647bafadff76fce1abb8ed8e0f54562ad/server/commands/check.go#L220

In a sample trace we saw in our logs, the `ReadUserTuple` returned a count of 1, therefore the parent span `resolveNode` should have taken 7ms (not 50ms) waiting for the other lookup to complete.

jon-whit:
## Description
Refactor Check to make use of better functional composition and also enforce maximum concurrency limits. The refactor introduces an interface for request dispatching which will allow us to implement a Check dispatcher that delegates requests to peers in an OpenFGA cluster (future work).

## References
Fixes #300 
Fixes #413

## Review Checklist


internal/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,73 @@
...
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+type Dispatcher interface {
+	CheckDispatcher
+}
+
+// CheckDispatcher defines an interface that can be implemented to resolve dispatched Check
+// requests. Implementations of the interface can provide local or remote dispatching mechansims.
+type CheckDispatcher interface {
```
jon-whit, line 15: Eventually we'll implement this interface over an RPC medium and each OpenFGA server will have service discovery to discover other OpenFGA replicas so that each one can dispatch/distribute requests to one another. It doesn't really make sense to introduce this until we have caching to go along with it, because otherwise we'll just be adding a lot of network traffic and extra latency for little to no benefit.

internal/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,73 @@
...
+
+type DispatchCheckResponse struct {
+	Allowed bool
+}
+
+type ResolutionMetadata struct {
+	Depth uint32
+}
+
+func (d *DispatchCheckRequest) GetStoreId() string {
```
jon-whit, line 35: These mimic protobuf generated helpers on purpose, because the idea is that this would move to a protobuf RPC definition eventually (when we implement the above interface for peer dispatching).

tests/oldcheck/check_test.go
```diff
@@ -100,9 +100,6 @@ func runTest(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
 				})
 				require.NoError(t, err)
 				require.Equal(t, assertion.Expectation, resp.Allowed, assertion)
-				if assertion.Trace != "" {
```
jon-whit, line 103: Removed this since we're moving away from the `resolution` field and the refactor doesn't have support for the trace resolution string. I can add the trace resolution string easily, but since we're moving away from it I decided to omit that from the scope of this PR.

pkg/typesystem/typesystem.go
```diff
@@ -22,6 +27,17 @@ var (
 	ErrInvalidUsersetRewrite = errors.New("invalid userset rewrite definition")
 )
 
+// ContextWithTypesystem attaches the provided TypeSystem to the parent context.
+func ContextWithTypesystem(parent context.Context, typesys *TypeSystem) context.Context {
```
jon-whit, line 31: We may consider using this more elsewhere now that it exists. Instead of passing around the TypeSystem everywhere we can inject it into the context and extract it where needed.

pkg/storage/storage_test.go
```diff
@@ -46,182 +36,3 @@ func TestNewPaginationOptions(t *testing.T) {
 		}
 	}
 }
-
-func TestStaticTupleKeyIterator(t *testing.T) {
```
jon-whit, line 50: Moved the iterator specific code into `iterators.go` and `iterators_test.go`

pkg/storage/storage.go
```diff
@@ -36,285 +32,14 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
 	}
 }
 
-type Iterator[T any] interface {
```
jon-whit, line 39: Moved iterator stuff into `iterators.go`.

pkg/server/server.go
```diff
@@ -242,33 +245,68 @@ func (s *Server) Write(ctx context.Context, req *openfgapb.WriteRequest) (*openf
...
+	for _, ctxTuple := range req.GetContextualTuples().GetTupleKeys() {
+		if err := validation.ValidateTuple(typesys, ctxTuple); err != nil {
+			return nil, serverErrors.HandleTupleValidateError(err)
+		}
+	}
+
+	g := graph.NewConcurrentChecker(storage.NewContextualTupleDatastore(s.datastore), 100)
+
+	ctx = typesystem.ContextWithTypesystem(ctx, typesystem.New(model))
+	ctx = storage.ContextWithContextualTuples(ctx, req.ContextualTuples.GetTupleKeys())
```
jon-whit, line 289: Notice that ContextualTuples are now truly "contextual" 😄 . We pass contextual tuples in through the context and the storage layer is responsible for extracting any tuples in the context, filtering them based on the query filters, and including/merging the filtered tuples in the returned tuple iterator (which combines the tuples from the database query and the contextual tuples).

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+	req *dispatcher.DispatchCheckRequest,
+) (*dispatcher.DispatchCheckResponse, error) {
+
+	if req.GetResolutionMetadata().Depth == 0 {
+		return nil, ErrResolutionDepthExceeded
+	}
+
+	typesys, ok := typesystem.TypesystemFromContext(ctx)
+	if !ok {
+		panic("typesystem missing in context")
```
miparnisari, line 263: instead of panic can we build it?

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+	}
+
+	ctx = typesystem.ContextWithTypesystem(ctx, typesys)
```
miparnisari, line 266: is this line necessary? it's being run in line 288 of `server/server.go`

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+	case exclusionSetOperator:
+		if len(children) != 2 {
+			panic(fmt.Sprintf("expected two rewrite operands in exclusion operator, but got '%d'", len(children)))
+		}
+
+		handlers = append(handlers,
+			c.checkRewrite(ctx, req, children[0]),
+			c.checkRewrite(ctx, req, children[1]),
+		)
+	}
```
miparnisari, line 492:
```suggestion
	switch setOpType {
	case unionSetOperator, intersectionSetOperator, exclusionSetOperator:
		for _, child := range children {
			handlers = append(handlers, c.checkRewrite(ctx, req, child))
		}
	default:
		// unknown operator
	}
```

and check for the right size of `children` for the exclusion case within `checkRewrite`

internal/graph/check.go
jon-whit: The typesystem is expected to be injected into the `ctx` before the `DispatchCheck` call immediately below. So yeah, this is necessary here.

internal/graph/check.go
jon-whit: Not really, because it's expected that the caller fetched it and injected it into the context. We avoid doing extra work in the Check evaluation path in this packag and save the responsibility of the typesystem construction for the caller. 

This is to separate concerns and responsibilities.

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+// exclusion implements a CheckFuncReducer that requires a 'base' CheckHandlerFunc to resolve to an allowed
+// outcome and a 'sub' CheckHandlerFunc to resolve to a falsey outcome. The base and sub computations are
+// handled concurrently relative to one another.
+func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*openfgapb.CheckResponse, error) {
+
+	if len(handlers) != 2 {
+		panic(fmt.Sprintf("expected two rewrite operands for exclusion operator, but got '%d'", len(handlers)))
+	}
+
+	limiter := make(chan struct{}, concurrencyLimit)
```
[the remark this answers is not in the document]
jon-whit: I'm not sure I understand.. The limiter here does the same thing it does in `intersection` and `union`. It limits the total number of concurrent goroutines can be inflight for the evaluation of the `CheckHandlerFunc` passed in. For the `exclusion` case if `concurrencyLimit` is 1, then only the base or subtract handlers can be run at a time (the other has to wait for the other to finish).

For intersection and union, if the number of `CheckHandlerFunc` is greater than the `concurrencyLimit`, then we only evaluate up to `concurrencyLimit` handlers at a time, and the others have to wait until those are complete.

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+// evaluations in flight at any point.
+func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- checkOutcome, handlers ...CheckHandlerFunc) func() {
+	limiter := make(chan struct{}, concurrencyLimit)
+
+	var wg sync.WaitGroup
+
+	checker := func(fn CheckHandlerFunc) {
+		resp, err := fn(ctx)
+		resultChan <- checkOutcome{resp, err}
+		<-limiter
```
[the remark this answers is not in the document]
jon-whit: It's a channel scoped to this function that is used to limit goroutines. So `limiter` seemed appropriately named.

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
+package graph
```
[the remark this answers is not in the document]
jon-whit: The plan for the `graph` package is to evolve it with dispatch based implementations of ConnectedObjects, Check, and Expand.

The graph package is for implementations of queries for graph based evaluation of relationship tuples.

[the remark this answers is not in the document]
jon-whit: Since these queries share certain behavior and share common errors, it makes sense for them to be under a package called something like `graph`.

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+}
+
+// NewConcurrentChecker constructs a ConcurrentChecker that can be used to evaluate a Check
+// request locally and with a high degree of concurrency.
+func NewConcurrentChecker(
+	ds storage.RelationshipTupleReader,
+	concurrencyLimit uint32,
+) *ConcurrentChecker {
+	checker := &ConcurrentChecker{ds: ds, concurrencyLimit: concurrencyLimit}
+	checker.dispatcher = checker // todo: replace with a different CheckDispatcher once we support remote dispatching
```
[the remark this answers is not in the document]
jon-whit: What is weird about it? Just the variable names used?

go.mod
```diff
@@ -40,13 +40,22 @@ require (
...
 	go.uber.org/zap v1.24.0
-	golang.org/x/exp v0.0.0-20230108222341-4b8118a2686a
+	golang.org/x/exp v0.0.0-20230125214544-b3c2aaf6208d
 	golang.org/x/sync v0.1.0
 	google.golang.org/grpc v1.52.0
 	google.golang.org/protobuf v1.28.1
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
```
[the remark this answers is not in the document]
jon-whit: Ah, good catch. Fixed 👍

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+type CheckFuncReducer func(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*openfgapb.CheckResponse, error)
+
+// resolver concurrently resolves one or more CheckHandlerFunc and yields the results on the provided resultChan.
+// Callers of the 'resolver' function should be sure to invoke the callback returned from this function to ensure
+// every concurrent check is evaluated. The concurrencyLimit can be set to provide a maximum number of concurrent
+// evaluations in flight at any point.
+func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- checkOutcome, handlers ...CheckHandlerFunc) func() {
+	limiter := make(chan struct{}, concurrencyLimit)
+
+	var wg sync.WaitGroup
```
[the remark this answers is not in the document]
jon-whit: In this usage of the `sync.WaitGroup`, I think `conc` doesn't seem to add much value to warrant bringing in an external library over primitives in the language. I'd prefer to stick to the standard library in this case.

pkg/storage/storage.go
```diff
@@ -36,285 +32,14 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
...
 type TupleBackend interface {
-	// Read the set of tuples associated with `store` and `TupleKey`, which may be nil or partially filled. If nil,
-	// Read will return an iterator over all the `Tuple`s in the given store. If the `TupleKey` is partially filled,
-	// it will return an iterator over those `Tuple`s which match the `TupleKey`. Note that at least one of `Object`
-	// or `User` (or both), must be specified in this case.
-	//
-	// The caller must be careful to close the TupleIterator, either by consuming the entire iterator or by closing it.
-	Read(context.Context, string, *openfgapb.TupleKey) (TupleIterator, error)
+	RelationshipTupleReader
+	RelationshipTupleWriter
```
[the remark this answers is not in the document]
jon-whit: Sure. I can get these changes in upstream and then rebase so these don't show as part of this diff. I'll do that next week 👍

pkg/storage/iterators.go
```diff
@@ -0,0 +1,286 @@
+package storage
```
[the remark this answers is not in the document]
jon-whit: Sure, I'll do this upstream and rebase afterwards. I'll do that next week 👍

internal/graph/check.go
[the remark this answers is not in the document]
miparnisari: Hmm no, conceptually... i'm not sure how to explain it... you are creating an object and then assigning a property to it that is the object itself?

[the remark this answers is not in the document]
craigpastro: > you are creating an object and then assigning a property to it that is the object itself?

This also seems weird to me.

internal/validation/validation.go
```diff
@@ -174,11 +174,9 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgapb.Tupl
 // a generic filtering mechanism when reading tuples. It is particularly useful to filter
 // out tuples that aren't valid according to the provided model, which can help filter
 // tuples that were introduced due to another authorization model.
-func FilterInvalidTuples(model *openfgapb.AuthorizationModel) storage.TupleKeyFilterFunc {
+func FilterInvalidTuples(typesys *typesystem.TypeSystem) storage.TupleKeyFilterFunc {
```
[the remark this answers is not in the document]
craigpastro: Pulled out into a separate PR.

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+		wg.Wait()
+		close(limiter)
+	}
+}
+
+// union implements a CheckFuncReducer that requires any of the provided CheckHandlerFunc to resolve
+// to an allowed outcome. The first allowed outcome causes premature termination of the reducer.
+func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*openfgapb.CheckResponse, error) {
+
+	cctx, cancel := context.WithCancel(ctx)
```
craigpastro, line 103: Any reason to create a new variable for the context?

internal/graph/check.go
craigpastro: Wait, didn't you just get the typesystem from the context in line 261? Why are you putting it back in?

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+							TupleKey:             tuple.NewTupleKey(usersetObject, usersetRelation, tk.GetUser()),
+							ResolutionMetadata: &dispatcher.ResolutionMetadata{
+								Depth: req.GetResolutionMetadata().Depth - 1,
+							},
+						}))
+				}
+
+			}
+
+			if len(handlers) > 0 {
```
craigpastro, line 383: Small nitpick, but I think it would read a bit nicer if the `if` handled the "negative" case of `len(handlers) == 0`. Similarly elsewhere. But up to you.

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+			c.checkRewrite(ctx, req, children[1]),
+		)
+	}
+
+	return func(ctx context.Context) (*openfgapb.CheckResponse, error) {
+		return reducer(ctx, c.concurrencyLimit, handlers...)
+	}
+}
+
+func (c *LocalChecker) checkRewrite(
```
craigpastro, line 499: Should we add some tracing to this function? Maybe we need to merge https://github.com/openfga/openfga/pull/477 first?

internal/graph/check.go
```diff
@@ -0,0 +1,536 @@
...
+	case *openfgapb.Userset_TupleToUserset:
+		return c.checkTTU(ctx, req, rewrite)
+	case *openfgapb.Userset_Union:
+		return c.checkSetOperation(ctx, req, unionSetOperator, union, rw.Union.GetChild()...)
+	case *openfgapb.Userset_Intersection:
+		return c.checkSetOperation(ctx, req, intersectionSetOperator, intersection, rw.Intersection.GetChild()...)
+	case *openfgapb.Userset_Difference:
+		return c.checkSetOperation(ctx, req, exclusionSetOperator, exclusion, rw.Difference.GetBase(), rw.Difference.GetSubtract())
+	default:
+		panic("unexpected userset rewrite encountere")
```
craigpastro, line 534:
```suggestion
		panic("unexpected userset rewrite encountered")
```

internal/graph/graph.go
```diff
@@ -16,8 +16,9 @@ const (
 )
 
 var (
-	ErrTargetError    = errors.New("graph: target incorrectly specified")
-	ErrNotImplemented = errors.New("graph: intersection and exclusion are not yet implemented")
+	ErrResolutionDepthExceeded = fmt.Errorf("resolution depth exceeded")
```
craigpastro, line 19:
```suggestion
	ErrResolutionDepthExceeded = errors.New("resolution depth exceeded")
```

pkg/server/server.go
craigpastro: I'd rather simply remove the `contextualtuples` package in this PR, as we may forget if we don't. Let's just accept this as the way forward and fix it here if there are problems. WDYT?

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: You cannot initialize an object's field with the object that you are trying to initialize. How else would you recommend this be done?

It may help to understand the vision here...

Eventually the wrapped `dispatcher` will be the delegate that remote dispatching gets routed to. Since we don't yet have remote dispatching, the delegate simply becomes itself (the `LocalChecker`).

pkg/server/server.go
jon-whit: To minimize the scope of this PR, I'd like to save the removal of existing packages for a subsequent PR. But I agree, we should drop the `contextualtuples` package and remove the existing Check implementation once this is merged 👍

pkg/storage/storage.go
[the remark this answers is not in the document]
jon-whit: @miparnisari https://github.com/openfga/openfga/pull/493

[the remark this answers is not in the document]
jon-whit: Ok, rebased.

internal/graph/check.go
jon-whit: Ah yes, no need. This is superfluous. Will remove.

internal/graph/check.go
jon-whit: Nice, I like that it's more uniform 👍  Done.

internal/graph/check.go
jon-whit: 👍 done.

internal/graph/check.go
jon-whit: Added some tracing methods that should get wired up automatically when #477 is merged.

How does that look?

internal/graph/check.go
jon-whit: Not really. I renamed to override `ctx`.

internal/graph/check.go
craigpastro: Looks good! Thanks.

craigpastro (dismissed):
🚢 it

craigpastro (approved):
LGTM

internal/graph/check.go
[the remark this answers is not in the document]
miparnisari: Can we get rid of the concept of dispatching in this PR? It's not relevant now

internal/graph/check.go
miparnisari: Okay. Then, can we not panic and return an error? (There's 3 occurrences of this, can you replace all?)

internal/graph/check.go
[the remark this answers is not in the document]
miparnisari: > For the exclusion case if concurrencyLimit is 1, then only the base or subtract handlers can be run at a time (the other has to wait for the other to finish).

But why would I even set it to 1? Could that happen?

I don't think we need a limiter in the exclusion case because we know that we will always run 2 goroutines. If each branch of `base` and `sub` are more complex, those will be rate-limited separately.

On a separate note, if we do keep the limiter: do we need to close it in the `defer` function?

pkg/server/server.go
```diff
@@ -232,33 +235,68 @@ func (s *Server) Write(ctx context.Context, req *openfgapb.WriteRequest) (*openf
...
+	for _, ctxTuple := range req.GetContextualTuples().GetTupleKeys() {
+		if err := validation.ValidateTuple(typesys, ctxTuple); err != nil {
+			return nil, serverErrors.HandleTupleValidateError(err)
+		}
+	}
+
+	g := graph.NewLocalChecker(storage.NewContextualTupleDatastore(s.datastore), 100)
+
+	ctx = typesystem.ContextWithTypesystem(ctx, typesystem.New(model))
+	ctx = storage.ContextWithContextualTuples(ctx, req.ContextualTuples.GetTupleKeys())
```
miparnisari, line 279: is there a reason why this code is here and not within `graph/check.go`? Also: can you create a task in our backlog that removes the old code, if this refactor proves to be successful in prod?

pkg/server/server.go
```diff
@@ -232,33 +235,68 @@ func (s *Server) Write(ctx context.Context, req *openfgapb.WriteRequest) (*openf
...
+
+	for _, ctxTuple := range req.GetContextualTuples().GetTupleKeys() {
+		if err := validation.ValidateTuple(typesys, ctxTuple); err != nil {
+			return nil, serverErrors.HandleTupleValidateError(err)
+		}
+	}
+
+	g := graph.NewLocalChecker(storage.NewContextualTupleDatastore(s.datastore), 100)
+
+	ctx = typesystem.ContextWithTypesystem(ctx, typesystem.New(model))
```
miparnisari, line 278:
```suggestion
	ctx = typesystem.ContextWithTypesystem(ctx, typesys))
```

internal/graph/check.go
```diff
@@ -0,0 +1,558 @@
...
+	}, nil
+}
+
+// checkDirect composes two CheckHandlerFunc which evaluate direct relationships with the provided
+// 'object#relation'. The first handler looks up direct matches on the provided 'object#relation@user',
+// while the second handler looks up relationships between the target 'object#relation' and any usersets
+// related to it.
+func (c *LocalChecker) checkDirect(parentctx context.Context, req *dispatcher.DispatchCheckRequest) CheckHandlerFunc {
+
+	typesys, ok := typesystem.TypesystemFromContext(parentctx)
```
miparnisari, line 297: The parameter `parentctx` is used here and then never again, can we pass the `typesys` directly? same comment for `checkTTU`

jon-whit: I'm favoring propagating the typesystem via the context because I want us to start propagating the typesystem in the context. Other refactors will come along with this change. If we've resolved the authorization model somewhere up the call chain, then there should be a typesystem in the context. I don't want to have to pass down the TypeSystem reference everywhere in the call chain that needs it.

pkg/server/server.go
jon-whit: Yes, because this implementation of Check doesn't care about request scoped context like contextual tuples, and so this code shouldn't belong in `graph/check.go`.

The storage layer now respects handling ContextualTuples if they are in the context, and so the responsibility is on the caller (e.g. `server/server.go`) to inject the ContextualTuple context into the context.

This seperates responsibilities/concerns between the Check implementation and the storage layer. The storage layer returns the tuples matching both the query AND those contextual tuples matching the query.

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: I don't think removing it in this PR is warranted, because some work has been put into it already. Why would we remove it if it's been done well compositionally? One of the reasons for the refactor _was_ to set us up so that we can dispatch in the near future if we start going that direction architecturally. In the meantime, the dispatching is just a loopback mechanism that dispatches to itself (e.g. no remote dispatching).

This code sets us up to be able to dispatch requests to peers over a remote RPC (or similar) protocol and it adds little to no overhead compositionally in the scope of this PR. This would also be a pretty significant change to this PR, which is already a reasonably sized PR and has already had rounds of review.

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: The function accepts a `concurrencyLimit`. The function doesn't assume that variable is a certain range (e.g. > 1). So if you call this function with a configured concurrency limit of `1` (which may be possible if we allow concurrency configurations in the future), then this code works regardless of what the limit is. The limit could be 1 or it could be 1000, and this code still works.

internal/graph/check.go
jon-whit: Sure 👍

jon-whit: Done

internal/graph/check.go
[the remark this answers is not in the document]
craigpastro: I think what Jon says above is valid here. We are moving in this direction and have it on our roadmap, so I don't think it hurts to introduce it in this PR. In fact, I think it sets us up nicely because check was written around it.

internal/graph/check.go
craigpastro: I don't have a strong opinion on this either way. Is there any documentation of best practices for what to store in the context? From what I read so far it seems like it is a nice place to store request-specific data there. In this sense perhaps it does makes sense the store the storeID and modelID or typesystem in there. It would make our signatures simpler, but then we also lose some explicitness... I need to do some reading.

jon-whit: Guidance I've read on Go's official blog and doc is to use context for request scoped data. 

Here's my rationale for the way I did this.. Contextual tuples seemed like a good fit to propagate down to the storage layer to remove the responsibility of handling tuples in the request context differently than the other tuples coming from the storage layer. It's always been a struggle for me to look at our current code and see this special handling of contextual tuples. To me they are just an "extension" of the tuples returned from the database, and they should be treated that way. Our Check, Expand, and ListObjects code should operate on tuples/iterators returned from the storage layer and not be bothered with contextual specific tuples coming in the request context. I believe that responsibility should be saved for the storage layer which is providing tuple filtering and querying capabilities for us already.

internal/graph/check.go
```diff
@@ -18,6 +17,64 @@ import (
 
 var tracer = otel.Tracer("internal/graph/check")
 
+// CheckResolver represents an interface that can be implemented to provide recursive resolution
+// of a Check.
+type CheckResolver interface {
```
craigpastro, line 22: I'll just leave this here https://go.dev/doc/effective_go#interface-names, and say how much I would like a `Checker` interface!

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: Ok, I tried to meet us all in the middle somewhere. I removed the notion of dispatching but kept things behind some interfaces so that if we ever do add the notion of dispatching it's super easy to port to 👍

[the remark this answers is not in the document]
jon-whit: Resolving this conversation for now. If you have further comments on the new structure, please start a new discussion.

internal/graph/check.go
jon-whit: Talked offline. The reason for not naming this interface `Checker` is because then that will lead to obscurities with our existing `Check` interface for the gRPC server definition. We'll have

`Check(ctx context.Context, req *openfgapb.CheckRequest) (*openfgapb.CheckResponse, error)`
and
`Check(ctx context.Context, req *graph.CheckRequest) (*graph.CheckResponse, error)`

I purposefully chose `CheckResolver` isntead of simply `Checker` because it avoids these overlapping signatures. The only distinguishing piece above is the package prefix (`openfgapb` vs `graph`). With this interface name we avoid the signature collisions altogether.

`ResolveCheck(ctx context.Context, req *graph.ResolveCheckRequest) (*graph.ResolveCheckResponse, error)`.

miparnisari:
Can you remove these two lines?
- https://github.com/openfga/openfga/blob/cdcdd94787be794646a01ff67140a9fec27f274e/pkg/server/server_test.go#L101
- https://github.com/openfga/openfga/blob/cdcdd94787be794646a01ff67140a9fec27f274e/pkg/server/server_test.go#L155

pkg/server/server.go
```diff
@@ -212,32 +220,65 @@ func (s *Server) Write(ctx context.Context, req *openfgapb.WriteRequest) (*openf
...
-		return nil, err
+		if errors.Is(err, graph.ErrResolutionDepthExceeded) {
+			return nil, serverErrors.AuthorizationModelResolutionTooComplex
+		}
+
+		return nil, serverErrors.HandleError("", err)
+	}
+
+	res := &openfgapb.CheckResponse{
+		Allowed: resp.Allowed,
```
craigpastro, line 281: Should we add a span attribute here with `allowed: resp.Allowed`?

jon-whit: Sure!

craigpastro: 👍

craigpastro (approved):
LGTM

internal/graph/check.go
```diff
@@ -0,0 +1,612 @@
...
+func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*openfgapb.CheckResponse, error) {
+
+	ctx, cancel := context.WithCancel(ctx)
+	resultChan := make(chan checkOutcome, len(handlers))
+
+	drain := resolver(ctx, concurrencyLimit, resultChan, handlers...)
+
+	defer func() {
+		cancel()
+		drain()
```
miparnisari, line 168: As you found offline, this will wait until all functions finish, which is not what we want.

jon-whit: Ok, I've fixed the issue now. I've enabled to the two tests in `pkg/server/server_test.go` (`TestCheckDoesNotThrowBecauseDirectTupleWasFound` and `TestShortestPathToSolutionWins`) to demonstrate that we a) don't return an error if another resolution path returned before the error and b) short-circuit if a resolution path was found (in the union case) or if a resolution path was rejected (in the exclusion and intersection cases) prior to other check function handlers resolve.

jon-whit: The changes also have little/to no impact on the overall goroutine footprint as well. The following charts show a 2m load tests (250rps of random Checks involving direct, computed_userset, and ttu resolutions). The first chart is this refactored Check and the second chart shows the results before the refactor.

Screen Shot 2023-02-06 at 7 31 15 PM

Screen Shot 2023-02-06 at 7 37 53 PM

miparnisari (approved):

internal/graph/check.go
miparnisari: Can run a longer test of a few hours?

jon-whit: Sure!

jon-whit: I ran the k6-load-test script last night for 4 hours to check for memory leaks etc.. Everything looked really good. No signs of memory leaks or excessive processing.

The test run started just before 4:00.

Screen Shot 2023-02-07 at 7 38 34 AM
Screen Shot 2023-02-07 at 7 38 53 AM
Screen Shot 2023-02-07 at 7 39 03 AM

PR #477, opened 2023-01-27
feat: wire up tracing

craigpastro:
Wire up tracing and refactor tracing to be more idiomatic. (There is an issue or discussion about this somewhere, but I can't find it now. When I do I will update the description.)

## Review Checklist


pkg/telemetry/tracing.go
```diff
@@ -1,12 +1,51 @@
...
+	"google.golang.org/grpc"
 )
 
-func NewNoopTracer() trace.Tracer {
-	return trace.NewNoopTracerProvider().Tracer("noop")
+func MustNewTracerProvider(endpoint string, ratio float64) *sdktrace.TracerProvider {
+	res, err := resource.Merge(
+		resource.Default(),
+		resource.NewSchemaless(
+			semconv.ServiceNameKey.String("openfga"),
```
jon-whit, line 22: mind making this a `const` at the top of this file?

craigpastro: Yep. Done. Thanks!

pkg/telemetry/tracing.go
```diff
@@ -1,12 +1,51 @@
...
+			semconv.ServiceVersionKey.String(build.Version),
+		))
+	if err != nil {
+		panic(err)
+	}
+
+	exp, err := otlptracegrpc.New(context.Background(),
```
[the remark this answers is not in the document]
craigpastro: This makes sense. Hopefully we won't need to add more in the future because everyone will just use the grpc endpoint. For the moment, how about we just rename the flag to `trace-otlgrpc-endpoint` and if we do need to add more in the future we can without breaking anything?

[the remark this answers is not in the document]
jon-whit: I think that's a good plan 👍

cmd/run.go
```diff
@@ -745,6 +748,10 @@ func RunServer(ctx context.Context, config *Config) error {
 
 	datastore.Close()
 
+	_ = tp.ForceFlush(ctx)
+
```
jon-whit, line 752: Collapse newline here?

cmd/run.go
```diff
@@ -876,6 +883,18 @@ func bindRunFlags(cmd *cobra.Command) {
 	util.MustBindPFlag("log.level", cmd.Flags().Lookup("log-level"))
 	util.MustBindEnv("log.level", "LOG_LEVEL")
 
+	cmd.Flags().Bool("trace-enabled", defaultConfig.Trace.Enabled, "enable tracing")
+	util.MustBindPFlag("trace.enabled", cmd.Flags().Lookup("trace-enabled"))
+	util.MustBindEnv("trace.enabled", "OPENFGA_TRACE_ENABLED")
+
+	cmd.Flags().String("trace-grpc-endpoint", defaultConfig.Trace.Endpoint, "the endpoint of the trace collector")
```
jon-whit, line 890: `trace-otlpgrpc-endpoint` instead? That way it's more clear that it's wiring up the `otlpgrpc` exporter.

cmd/run.go
craigpastro: Done. Thanks!

cmd/run.go
craigpastro: Used `otlp-grpc` as that seems more common. Done. Thanks!

jon-whit: Perfect 👍

cmd/run.go
```diff
@@ -462,23 +474,15 @@ func RunServer(ctx context.Context, config *Config) error {
...
-		middleware.NewLoggingInterceptor(logger),
-	}
-
-	streamingServerInterceptors := []grpc.StreamServerInterceptor{
-		grpc_validator.StreamServerInterceptor(),
-		grpc_auth.StreamServerInterceptor(middleware.AuthFunc(authenticator)),
-		middleware.NewStreamingRequestIDInterceptor(logger),
-		middleware.NewStreamingLoggingInterceptor(logger),
-	}
+	unaryInterceptors, streamingInterceptors := middleware.NewInterceptors(&middleware.Config{
```
elbuo8, line 477: Perhaps this should have been in a different PR.

cmd/run.go
```diff
@@ -462,23 +474,15 @@ func RunServer(ctx context.Context, config *Config) error {
...
+		Logr:           logger,
```
elbuo8, line 478: Keep consistency in naming as this is found pretty much everywhere.

cmd/run/run.go
```diff
@@ -459,23 +471,15 @@ func RunServer(ctx context.Context, config *Config) error {
 		return fmt.Errorf("failed to initialize authenticator: %w", err)
 	}
 
-	unaryServerInterceptors := []grpc.UnaryServerInterceptor{
```
jon-whit, line 462: You're gonna hate me, but do you mind if we save this for a different PR? I'd like to separate the code reveiw for this change from the scope of work for the tracing stuff.

I have some comments on this stuff, but don't want it to hinder progress on the tracing stuff.

craigpastro: haha. No worries. Reverted... mostly.

cmd/run.go
craigpastro: Different one it is. Thanks!

cmd/run.go
craigpastro: Okay. Thanks!

cmd/run/run.go
```diff
@@ -459,23 +473,25 @@ func RunServer(ctx context.Context, config *Config) error {
...
+	unaryInterceptors := []grpc.UnaryServerInterceptor{
 		grpc_validator.UnaryServerInterceptor(),
-		grpc_auth.UnaryServerInterceptor(middleware.AuthFunc(authenticator)),
+		otelgrpc.UnaryServerInterceptor(otelgrpc.WithTracerProvider(tp)),
```
jon-whit, line 478: I'm wondering about the order of these middlewares..

Should we have a `log`, `request_id` and `trace` middleware run before the validation? Even if the request does not meet the protobuf validation requirements, it would be good to have the request id in the log context and the trace info. If validation was taking a long time the trace's would give us a good indication of that and if users are having validation problems we could see the requests and correlate with the request_id.

What's the right approach here? On one hand I don't super chatty logs with request validation problems, but maybe that just means those should fall under a debug level instead of error level.

otel-collector-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+processors:
+  batch:
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [jaeger]
+    metrics:
```
jon-whit, line 25: Let's omit the `metrics` bit since we're moving in the direction of prometheus instead of otel with otlp protocol.

craigpastro: Good point. Done! Thanks.

pkg/telemetry/tracing.go
```diff
@@ -1,12 +1,53 @@
...
+		resource.Default(),
+		resource.NewSchemaless(
+			semconv.ServiceNameKey.String(serviceName),
+			semconv.ServiceVersionKey.String(build.Version),
+		))
+	if err != nil {
+		panic(err)
+	}
+
+	exp, err := otlptracegrpc.New(context.Background(),
```
jon-whit, line 31: Let's add a context with a timeout here so we don't potentially hang indefinitely on the `grpc.WithBlock()`.

jon-whit (approved):

otel-collector-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+exporters:
+  jaeger:
+    endpoint: jaeger:14250
+    tls:
+      insecure: true
+  logging:
+    loglevel: debug
+  prometheus:
+    endpoint: "0.0.0.0:1234"
```
miparnisari, line 14: unused?

miparnisari (commented):
LGTM!

PR #490, opened 2023-01-30
update dependencies, change ListObjects gauge to updowncounter

miparnisari:
## Description
- Update dependencies
- Remove dependency `x/exp` because only one function within it was being used 
- Change ListObjects gauge to updowncounter
   > UpDownCounter is a [synchronous Instrument](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#synchronous-instrument-api) which supports increments and decrements. Note: if the value is [monotonically](https://wikipedia.org/wiki/Monotonic_function) increasing, use [Counter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#counter) instead.
Example uses for UpDownCounter: the number of active requests, **the number of items in a queue**


tools/go.mod
```diff
@@ -1,22 +1,23 @@
 module github.com/openfga/openfga/tooling
 
-go 1.18
+go 1.19
 
 require (
-	github.com/golang/mock v1.7.0-rc.1
-	github.com/golangci/golangci-lint v1.48.0
-	golang.org/x/vuln v0.0.0-20221103225512-4f561ca73b59
+	github.com/golang/mock v1.6.0
```
miparnisari, line 6: Same as `go.mod`

tools/go.mod
```diff
@@ -1,22 +1,23 @@
 module github.com/openfga/openfga/tooling
 
-go 1.18
+go 1.19
```
miparnisari, line 3: Same as `go.mod`

tools/go.mod
rhamzeh: Small question, why do we need golintci and vuln as part of the tools dependencies? Aren't those part of the CI process?

What is the benefit of having them here?

rhamzeh (approved):

tools/go.mod
craigpastro: My understanding is that it helps with local development. When you `make run` it will install this stuff and so everything will work seamlessly. Otherwise you may have to do a `go get ...` yourself before running. (In the case of mock and golangci anyway. I don't know why vuln is there. Oh wait, that is so it gets installed in ci.)

PR #494, opened 2023-01-31
refactor: wrap storage in server

craigpastro:
## Description

This will ensure that users of OpenFGA (as a library) won't forget this important step.

## Review Checklist

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -53,7 +53,6 @@ type Dependencies struct {
 }
 
 type Config struct {
-	DatastoreMaxCacheSize  int
 	ResolveNodeLimit       uint32
 	ChangelogHorizonOffset int
 	ListObjectsDeadline    time.Duration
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -62,11 +64,13 @@ type Config struct {
 // New creates a new Server which uses the supplied backends
 // for managing data.
 func New(dependencies *Dependencies, config *Config) *Server {
+	datastore := caching.NewCachedOpenFGADatastore(storage.NewContextWrapper(dependencies.Datastore), config.DatastoreMaxCacheSize)
```
jon-whit, line 67: If a client constructs a server with a specific datastore implementation, then we shouldn't override it by default. I can see this leading to very confusing behavior.

I recommend we provide alternate constructors for behavior that overrides user provided implementations. Maybe `NewWithCachedDatastore(...)`  is a good alternate constructor name?

craigpastro: They do have to supply the config value `DatastoreMaxCacheSize` as well. By default this is 0 which will have no effect. So maybe it is okay? Or perhaps a more specific name like `DatastoreAuthModelMaxCacheSize`. WDYT?

jon-whit: I think this is a leaky design detail. A datastore config shouldn't have to include cache settings, that's an implementation detail of the datastore implementation. If a datastore implementation implements caching for Authz models then that's it's choice, but it shouldn't be part of the config of the server.

I vote we remove this config from the server. It's not a server config anyways.

craigpastro: Yeah, this makes sense. I'll update.

elbuo8: Don’t think this contructor will scale as other options/permutations similar to it get added. Perhaps using with WithOptions pattern in the storage layer is more suitable.

If not, keep this PR scoped to the wrapper in particular.

jon-whit: > Don’t think this contructor will scale as other options/permutations similar to it get added. Perhaps using with WithOptions pattern in the storage layer is more suitable.

I agree. The options builder pattern would be quite nice in this scenario.

craigpastro: So many things can go wrong here, that I am beginning to think that wrapping the datastore needs to be left up to the user, even with the builder pattern.

Namely, the context datastore wrapper should be wrapped around the datastore implementation as it creates a new parent context and if you are depending on that for anything it would be bad. Thus, in our current implementation if we passed a cached wrapped datastore it would wrap the cached datastore... which is not bad actually, but still.

In the builder patter what happens if we called `WithDatastore(ds Datastore)` and also `WithCachedDatastore(ds Datastore)`, and also what if the datastore that was passed in was also wrapped with something else? Seems dangerous.

elbuo8: > If not, keep this PR scoped to the wrapper in particular.

craigpastro: I am not sure that is a good idea either though. Because that will mean we wrap whatever comes in, but the context one should be the first wrapper.

craigpastro:
We should think about this.

PR #504, opened 2023-02-03
fix: listObjects test expectations

miparnisari:
This PR only changes the test expectations for ListObjects where the model is using the 1.1 syntax.

- In the test model added the `user:*` restriction to the `repo#admin` relation
- In the test tuples I fixed one tuple which would be invalid and added a few tuples that would support a new test
- Added a new test
- Fixed the response of existing tests


craigpastro (approved):
LGTM. It would be nice, in the future, if we can move these tests to something similar to what we have for check in `tests`.

jon-whit (approved):

PR #513, opened 2023-02-06
feat(internal): make CombinedIterator accept more than two inputs

miparnisari:
I don't know why we didn't do this from the get-go :(

This PR also should bring a performance improvement because now, if the iterator is finished, we don't call `Next()`.

### DIFF
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -102,38 +102,41 @@ func (u *uniqueObjectIterator) Stop() {
 }
 
 type combinedIterator[T any] struct {
-	iter1, iter2 Iterator[T]
+	iters []Iterator[T]
 }
 
 func (c *combinedIterator[T]) Next() (T, error) {
-	val, err := c.iter1.Next()
-	if err != nil {
-		if !errors.Is(err, ErrIteratorDone) {
-			return val, err
+	for i := 0; i < len(c.iters); i++ {
+		if c.iters[i] != nil {
+			val, err := c.iters[i].Next()
+			if err != nil {
+				if !errors.Is(err, ErrIteratorDone) {
+					return val, err
+				} else {
+					c.iters[i] = nil // end of this iterator
+				}
+			} else {
+				return val, nil
+			}
 		}
-	} else {
-		return val, nil
 	}
 
-	val, err = c.iter2.Next()
-	if err != nil {
-		if !errors.Is(err, ErrIteratorDone) {
-			return val, err
-		}
-	}
-
-	return val, err
+	var val T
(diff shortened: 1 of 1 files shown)


pkg/storage/iterators.go
```diff
@@ -102,38 +102,41 @@ func (u *uniqueObjectIterator) Stop() {
...
+	for i := 0; i < len(c.iters); i++ {
+		if c.iters[i] != nil {
+			val, err := c.iters[i].Next()
+			if err != nil {
+				if !errors.Is(err, ErrIteratorDone) {
+					return val, err
+				} else {
+					c.iters[i] = nil // end of this iterator
+				}
+			} else {
```
jon-whit, line 118: You can remove excessive  indentation and make this code more easily readable with:
```

for i, iter := range c.iters {
    if iter == nil {
        continue
    }

    val, err := iter.Next()
    if err != nil {
        if !errors.Is(err, ErrIteratorDone) {
            return val, err
        }

       c.iters[i] = nil
       continue
    }

    return val, nil
}
```

pkg/storage/iterators.go
```diff
@@ -102,38 +102,41 @@ func (u *uniqueObjectIterator) Stop() {
...
+			c.iters[i].Stop()
+		}
+	}
 }
 
-// NewCombinedIterator takes two generic iterators of a given type T and combines them into a single iterator that yields
-// all of the values from both iterators. If the two iterators yield the same value then duplicates will be returned.
-func NewCombinedIterator[T any](iter1, iter2 Iterator[T]) Iterator[T] {
-	return &combinedIterator[T]{iter1, iter2}
+// NewCombinedIterator takes an array of generic iterators of a given type T and combines them into a single iterator that yields
```
craigpastro, line 136:
```suggestion
// NewCombinedIterator takes generic iterators of a given type T and combines them into a single iterator that yields
```

pkg/storage/iterators.go
```diff
@@ -102,38 +102,43 @@ func (u *uniqueObjectIterator) Stop() {
...
-	iter1, iter2 Iterator[T]
+	iters []Iterator[T]
 }
 
 func (c *combinedIterator[T]) Next() (T, error) {
-	val, err := c.iter1.Next()
-	if err != nil {
-		if !errors.Is(err, ErrIteratorDone) {
-			return val, err
+	for i := 0; i < len(c.iters); i++ {
```
jon-whit, line 109: Why not use `range`? It removes the chatty `c.iters[i]` stuff everywhere it's reference in this code.

craigpastro (approved):
LGTM

PR #514, opened 2023-02-06
feat: add Prometheus metrics and Grafana integration

jon-whit:
## Description
Adds Prometheus as the default metrics integration in OpenFGA along with a Grafana integration for local `docker-compose` based development. This makes it easy for developers using Docker to spin up OpenFGA and start getting metrics/observability out of the box.

## Review Checklist


jon-whit:
@AlexandreBrg - the OpenFGA maintainer team talked about supporting Prometheus for the standard metric provider in OpenFGA instead of OpenTelemetry. The primary reasons for this include:

* Prometheus is more standard in the CNCF ecosystem. Prometheus is a CNCF Graduated project, whereas OpenTelemetry is Incubating.

* The Prometheus Go SDK is stable with major releases, whereas the OpenTelemetry Metrics SDK is not marked as stable.

   https://opentelemetry.io/docs/instrumentation/go/#status-and-releases
   https://opentelemetry.io/docs/instrumentation/go/manual/#creating-metrics

* Prometheus is more widely adopted in many large CNCF projects as compared to the OpenTelemetry Metrics API (not to be confused with OpenTelemetry Tracing).

* Prometheus is part of a larger ecosystem of integrations in the open-source community to date.

* Prometheus is evolving with other large CNCF initiatives like the OpenMetrics format, which didn’t exist at the time OpenTelemetry started. Some of these initiatives (namely around metrics) are starting to supersede some of the original design ideas behind the OpenTelemetry initiative.

We don't want to discredit any of the work you previously did on the OpenTelemetry Metrics work. We appreciate the contribution, and we would love to see more collaboration with you in the future as well. 

You should be able to get all of the runtime metrics that were registered by default through your OpenTelemetry integration, but they will be exported by Prometheus instead. The `docker-compose.yaml` file setups up Prometheus and Grafana by default and so you should be able to `docker compose up` and then go to Grafana in the browser and there is a default "OpenFGA Dashboard" that you can use as a reference. We'd like your feedback on this approach? Thoughts/questions/concerns?

cmd/run/run.go
```diff
@@ -175,14 +175,11 @@ type ProfilerConfig struct {
...
-}
-
-// OpenTelemetryConfig defines configurations for OpenTelemetry telemetry settings.
-type OpenTelemetryConfig struct {
-	OpenTelemetryMetricsConfig `mapstructure:"metrics"`
+// MetricConfig defines configurations for serving custom metrics from OpenFGA.
+type MetricConfig struct {
+	Enabled             bool
+	Addr                string
+	EnableRPCHistograms bool
```
elbuo8, line 182: 🤔 why have this as configurable?

pkg/telemetry/metrics.go
```diff
@@ -1,80 +1 @@
 package telemetry
-
```
craigpastro, line 2: Do we need this file at all anymore?

cmd/run/run.go
jon-whit: Histograms with large cardinality can be quite costly, and so I felt it was a good flag to have.  The cardinality of the metrics collected for RPC request duration isn't excessive necessarily, but if you have no need for RPC histograms then this cuts down on the impact of the metrics reported by OpenFGA. 

We can omit it in favor of enabling RPC histograms if `metrics.enabled=true` (by default). If others feel that is appropriate I can make that change 👍

pkg/server/commands/list_objects.go
```diff
@@ -165,15 +160,6 @@ func (q *ListObjectsQuery) Execute(
 	req *openfgapb.ListObjectsRequest,
 ) (*openfgapb.ListObjectsResponse, error) {
 
-	listObjectsCounter, err := q.Meter.Int64UpDownCounter(
```
jon-whit, line 168: Removed this metric for now since a) we don't use it anywhere and b) don't particularly have a plan/strategy around why it is useful for us.

If we find that having this information is valuable to us, then we can introduce it into the `pkg/telemetry/metrics` package as a custom Prometheus metric and then build monitoring integrations around it at that time.

grafana/dashboards/openfga.json
```diff
@@ -0,0 +1,1161 @@
+{
```
craigpastro, line 1: Does it make sense to put all the observability stuff like this and `promethus` in an `observability` folder (or `telemetry` or something)? Also the `otel-collector-config.yaml` file.

pkg/server/commands/check.go
```diff
@@ -26,16 +25,15 @@ import (
...
 type CheckQuery struct {
 	logger           logger.Logger
-	meter            metric.Meter
 	datastore        storage.OpenFGADatastore
 	resolveNodeLimit uint32
 }
 
-// NewCheckQuery creates a CheckQuery with specified `tupleBackend` and `typeDefinitionReadBackend` to use for storage
-func NewCheckQuery(datastore storage.OpenFGADatastore, meter metric.Meter, logger logger.Logger, resolveNodeLimit uint32) *CheckQuery {
+// NewCheckQuery creates a CheckQuery that uses the datastore and loggers provided. Check resolution
```
craigpastro, line 32: Check resolution depth?

pkg/telemetry/metrics.go
jon-whit: Not right now, but I thought I'd leave it so that we can start adding custom metrics in this file. But we can get rid of it in favor of adding them in later 👍

grafana/dashboards/openfga.json
jon-whit: Probably 😄 .  Let's put it in a `telemetry` folder 👍

pkg/telemetry/metrics.go
jon-whit: Removed for now. We can introduce custom metrics in each package that defines them 👍

grafana/dashboards/openfga.json
jon-whit: Put the telemetry configuration manifests in the top level `telemetry` folder.

cmd/run/run.go
```diff
@@ -415,20 +412,7 @@ func RunServer(ctx context.Context, config *Config) error {
 		experimentals = append(experimentals, server.ExperimentalFeatureFlag(feature))
 	}
 
-	meter := metric.NewNoopMeter()
 	var err error
```
craigpastro, line 415: Can you move this to closer to where it is used?

craigpastro (dismissed):
Left one comment, but we can do that later too. LGTM

cmd/run/run.go
jon-whit: Yep, done.

craigpastro (dismissed):
LGTM

cmd/run/run.go
```diff
@@ -482,23 +465,41 @@ func RunServer(ctx context.Context, config *Config) error {
...
+		streamingInterceptors = append(streamingInterceptors, grpc_prometheus.StreamServerInterceptor)
+
+		if config.Metrics.EnableRPCHistograms {
+			grpc_prometheus.EnableHandlingTimeHistogram()
+		}
+	}
+
+	if config.Trace.Enabled {
+		unaryInterceptors = append(unaryInterceptors, otelgrpc.UnaryServerInterceptor(otelgrpc.WithTracerProvider(tp)))
+		streamingInterceptors = append(streamingInterceptors, otelgrpc.StreamServerInterceptor(otelgrpc.WithTracerProvider(tp)))
```
craigpastro, line 487: Can you drop the `otelgrpc.WithTracerProvider(tp)`? If it is not supplied it uses the global tracer provider, so this is redundant.

jon-whit: Yep, good call. Done.

craigpastro (approved):
LGTM

emsearcy:
Comment: OTEL metrics is now a stable API (even if the SDK is still being finalized). Tracing was in a similar situation last year (I was using it for some of our middleware that integrates with Auth0), and both the API and SDK are now stable. Tracing through OpenTelemetry Collector is a good way to get traces to something like Datadog, while also being able to route them to Jaeger for local development.

I know it's out of scope for this ticket, but having per-trace observability for OpenFGA (how long is being spent making Postgres queries, etc) would be good. Might be worth bringing back in OTEL, and then supporting OTLP metrics and traces to OTEL Collector sidecar (which supports Prometheus) might be *less* overhead.

https://opentelemetry.io/docs/reference/specification/status/

PR #516, opened 2023-02-07
chore: add utility method for object relations

miparnisari:


pkg/tuple/tuple.go
```diff
@@ -55,6 +55,14 @@ func BuildObject(objectType, objectID string) string {
 	return fmt.Sprintf("%s:%s", objectType, objectID)
 }
 
+// GetObjectRelationAsString returns a string like "object#relation". If there is no relation it returns "object"
+func GetObjectRelationAsString(objectRelation *openfgapb.ObjectRelation) string {
```
craigpastro, line 59: You are introducing a new function, but not using it anywhere?

jon-whit, line 59: Please add some tests for this 👍

jon-whit (changes requested):
Please add some tests.

pkg/tuple/tuple.go
miparnisari: It's used in Auth0 FGA :)

craigpastro: Thank you! Yes, we talked offline about it. My personal opinion is that in that case it shouldn't be here, but I will defer to others.

jon-whit: There are other places in our code within OpenFGA that should delegate to this function as well. Here is a list of them that I found from a quick search (there may be others):

* `pkg/server/commands/connected_objects.go` (line 400)
* `pkg/storage/memory/memory.go` (line 387)

Please refactor those to use this method as well. Then this debate is non-applicable 😄

miparnisari: Done!  Thanks - i don't know why I didn't check where I could use this here, that was dumb..

jon-whit (approved):

PR #519, opened 2023-02-07
chore: wire up ListObjects with the refactored Check implementation

jon-whit:
## Description
Wires up ListObjects (for the concurrent Check evaluation path) with the newly refactored Check implementation.

## Review Checklist

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/openfga/openfga/internal/contextualtuples"
+	"github.com/openfga/openfga/internal/graph"
 	"github.com/openfga/openfga/internal/validation"
 	"github.com/openfga/openfga/pkg/logger"
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
@@ -35,6 +36,7 @@ type ListObjectsQuery struct {
 	ListObjectsMaxResults uint32
 	ResolveNodeLimit      uint32
 	ConnectedObjects      func(ctx context.Context, req *ConnectedObjectsRequest, results chan<- string) error
+	CheckResolver         graph.CheckResolver
 }
 
 type listObjectsRequest interface {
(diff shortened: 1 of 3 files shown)


pkg/server/commands/list_objects.go
```diff
@@ -315,19 +317,22 @@ func (q *ListObjectsQuery) internalCheck(
...
 		TupleKey:             tuple.NewTupleKey(tuple.ObjectKey(obj), req.GetRelation(), req.GetUser()),
-		ContextualTuples:     req.GetContextualTuples(),
+		ContextualTuples:     req.GetContextualTuples().GetTupleKeys(),
+		ResolutionMetadata: &graph.ResolutionMetadata{
+			Depth: q.ResolveNodeLimit,
+		},
 	})
 	if err != nil {
 		// ignore the error. we don't want to abort everything if one of the checks failed.
 		q.Logger.ErrorWithContext(ctx, "check_error", zap.Error(err))
```
craigpastro, line 332: Do we want to log these as errors?

craigpastro (dismissed):
LGTM. Left an unrelated comment though.

pkg/server/commands/list_objects.go
miparnisari: why not? 🤔

pkg/server/test/list_objects.go
```diff
@@ -530,6 +533,7 @@ func BenchmarkListObjectsWithConcurrentChecks(b *testing.B, ds storage.OpenFGADa
 		Logger:           logger.NewNoopLogger(),
 		Meter:            telemetry.NewNoopMeter(),
 		ResolveNodeLimit: defaultResolveNodeLimit,
+		CheckResolver:    graph.NewLocalChecker(storage.NewContextualTupleDatastore(ds), 100),
```
miparnisari, line 536: can you create a constant for the 100? it's being set in a few places now

pkg/server/commands/list_objects.go
jon-whit: We should at least fix the error message. The error message should simply be the error 😄 . I shouldn't have to look at the `error` field of the logged message.

craigpastro: Now I am not sure. Writing aloud for a sec: if a check fails and it was `allowed: true` then ignoring it may mean we return a result which does not contain all the results. This could be bad if one wants to use this result to delete objects a particular user has access to or something...

pkg/server/test/list_objects.go
jon-whit: Done.

pkg/server/commands/list_objects.go
craigpastro: Now we have https://github.com/orgs/openfga/discussions/110, but obviously this is not the PR to solve this question.

craigpastro (approved):
LGTM

pkg/server/commands/list_objects.go
jon-whit: Will trace this work as part of that discussion 👍

PR #520, opened 2023-02-08
chore: add `grpc-health-probe` for Health checks

jon-whit:
## Description
Adds the `grpc-health-probe` binary to the OpenFGA container during the Docker build process.

Since OpenFGA is a gRPC server with an HTTP proxy on top of it, it makes sense to expose the Health checks via the grpc-health-probe. This is the defacto standard way of doing Health checks for gRPC, and it aligns with the [gRPC Health Check Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) as well.

https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/
https://github.com/grpc-ecosystem/grpc-health-probe/

## References
https://github.com/orgs/openfga/discussions/109

## Review Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,12 @@
 FROM golang:1.20-alpine AS builder
+ARG OS
+ARG ARCH
 
 WORKDIR /app
 
 COPY . .
 RUN go build -o ./openfga ./cmd/openfga
+RUN wget -q -O grpc_health_probe "https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.13/grpc_health_probe-${OS}-${ARCH}"
 
 FROM alpine as final
 EXPOSE 8081
@@ -11,5 +14,7 @@ EXPOSE 8080
 EXPOSE 3000
 COPY --from=builder /app/openfga /app/openfga
 COPY --from=builder /app/assets /app/assets
+COPY --from=builder /app/grpc_health_probe /bin/grpc_health_probe
+RUN chmod +x /bin/grpc_health_probe
 WORKDIR /app
 ENTRYPOINT ["./openfga"]
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -38,6 +38,10 @@ dockers:
     use: buildx
     build_flag_templates:
       - "--platform=linux/amd64"
+      - "--build-arg"
+      - "OS=linux"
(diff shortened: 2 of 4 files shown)


jon-whit:
Looks like I need to fix the `make function-test` target (which uses `docker build ..`). Docker builds need to happen against a particular OS/ARCH combination since the container embeds a `grpc-health-probe` binary inside of it.

That's why the functional tests are failing right now.

rhamzeh:
This does not seem to be working (based on the included `docker-compose.yaml`)

Screen Shot 2023-02-08 at 2 26 47 PM

rhamzeh:
Realized I needed to update the image first, but note that building on this branch seems broken:
Screen Shot 2023-02-08 at 3 25 08 PM

jon-whit:
> Realized I needed to update the image first, but note that building on this branch seems broken: Screen Shot 2023-02-08 at 3 25 08 PM

The issue is that the container build depends on os/arch specific combinations, and so a simple docker build doesn’t do it. You’ll need to set the build args for it to work e.g.
`docker build --build-arg OS=linux --build-arg ARCH=amd64 -t openfga/openfga:testing .`

I'm working on adding these changes to the Makefile 👍

Dockerfile
```diff
@@ -1,15 +1,20 @@
 FROM golang:1.20-alpine AS builder
+ARG OS
```
rhamzeh, line 2:
```suggestion
ARG OS=linux
```

Dockerfile
```diff
@@ -1,15 +1,20 @@
...
+ARG ARCH
```
rhamzeh, line 3:
```suggestion
ARG ARCH=amd64
```

rhamzeh (approved):
Tested and working as expected 👍🏽 

Any chance we can default the OS to "linux" and ARCH to "amd64"? I think we and folks are most likely to build for these for this combo, and we have the ability to override for those who need it

jon-whit:
> Tested and working as expected 👍🏽
> 
> Any chance we can default the OS to "linux" and ARCH to "amd64"? I think we and folks are most likely to build for these for this combo, and we have the ability to override for those who need it

@rhamzeh yep, I committed your suggestions. That's a good idea, thanks!

rhamzeh (approved):

PR #521, opened 2023-02-08
chore: add a span attr to indicate what ListObjects handler was used

jon-whit:
## Description
This PR adds a span attribute to the `ListObjects` span that indicates what handler was used (optimized or non-optimized variant).

## References
This work is essentially superseded by https://github.com/openfga/openfga/pull/522, because at that point it will be obvious which path ListObjects took.

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #522, opened 2023-02-08
chore: add tracing to internal ConnectedObjects API

jon-whit:
## Description
This adds tracing to the internal `ConnectedObjects` API, which is a reverse graph expansion algorithm that is used in the optimized version of ListObjects (the one that works with models involving type restrictions).

Here's an example screenshot of the traces produced by the changes in this PR:
Screen Shot 2023-02-09 at 4 13 09 PM

Each span is annotated with attributes that represent the relationship tuples that are expanded at each level in the reverse expansion graph evaluation.

## References
Related to https://github.com/openfga/openfga/pull/521

## Review Checklist

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -49,6 +49,19 @@ const (
 	ComputedUsersetIngress
 )
 
+func (r RelationshipIngressType) String() string {
+	switch r {
+	case DirectIngress:
+		return "direct"
+	case ComputedUsersetIngress:
+		return "computed_userset"
+	case TupleToUsersetIngress:
+		return "ttu"
+	default:
+		return "undefined"
+	}
+}
+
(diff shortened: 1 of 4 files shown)


miparnisari:
Can you add new parent spans also to the un-optimized ListObjects so that we don't see dangling spans?  E.g. the ones here, they belong inside a `Check` span:
image

jon-whit:
> Can you add new parent spans also to the un-optimized ListObjects so that we don't see dangling spans? E.g. the ones here, they belong inside a `Check` span: image

Can you do that in a separate PR? This PR is scoped to adding traces in the ConnectedObjects API, not in the ListObjects code paths.

I also recommend that we not add that span to ListObjects but instead that should be a top-level span inside of the Check handler (whichever one is used). That way whoever calls it will get that behavior.

craigpastro (approved):

PR #523, opened 2023-02-09
fix: add missing typesystem context in ListObjects (concurrent check)

jon-whit:
## Description
Fixes ListObjects in the event that concurrent Checks are evaluated (for 1.0 models or models involving intersection or exclusion).

## References
Related to #524 

## Review Checklist

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -90,6 +90,8 @@ func (q *ListObjectsQuery) handler(
 	}
 
 	handler := func() {
+		ctx = typesystem.ContextWithTypesystem(ctx, typesys)
+
 		err = q.performChecks(ctx, req, resultsChan)
 		if err != nil {
 			errChan <- err
--- a/cmd/openfga/main_test.go
+++ b/cmd/openfga/main_test.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
+	parser "github.com/craigpastro/openfga-dsl-parser/v2"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/container"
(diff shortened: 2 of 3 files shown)


pkg/server/commands/list_objects.go
```diff
@@ -90,6 +90,8 @@ func (q *ListObjectsQuery) handler(
 	}
 
 	handler := func() {
+		ctx = typesystem.ContextWithTypesystem(ctx, typesys)
```
jon-whit, line 93: This is exhibit A as to why we use panic in handlers where there are primitive expectations that should be met during runtime..

(referring back to this comment https://github.com/openfga/openfga/pull/476#discussion_r1089557291)

The problem here was that the caller chooses to just log an error and continue on, and so we actually end up returning an empty result list with a successful status code, but the server should have panic'd at runtime. In fact, our tests should have failed in our CI/CD pipeline because the server should have panic'd during tests, but it did not beceause we chose to return an error and the caller chose to handle the error by logging and continuing on as normal.

This demonstrates one of the good reasons to use panic in core/critical code paths that have expectations from calling code AND you want to detect failures to wire things up properly at runtime during the test phase of development.

jon-whit: Also, we should change https://github.com/openfga/openfga/blob/main/pkg/server/commands/list_objects.go#L330-L333 to return an error promptly. Ignoring an error and trying to complete the request despite the error is not a good call, because at best it leads to potential false reports. If you have access to 1 object and that Check fails, then you'll get an empty object list back, and that's not correct.

OpenFGA, as a system, should prioritize correctness over the attempt to fulfill the request at any cost. If an error happens during the lifetime of a request, and the system cannot correctly guarantee the results it returns as a result of it, then the server should return an error.

craigpastro (approved):
LGTM

PR #528, opened 2023-02-10
fix: race creating duplicates in ConnectedObjects reverseExpandDirect

jon-whit:
## Description
This fixes a race condition that leads to duplicates occasionally being returned. We had a similar issue we fixed in #475, but we missed this code path as well.

## Review Checklist


pkg/server/commands/connected_objects.go
```diff
@@ -489,15 +490,6 @@ func (c *ConnectedObjectsCommand) reverseExpandDirect(
...
 
-	// e.g. 'user:*'
-	if val, ok := req.targetUserRef.(*UserRefTypedWildcard); ok {
-		targetRelationRef.RelationOrWildcard = &openfgapb.RelationReference_Wildcard{}
-
-		userFilter = append(userFilter, &openfgapb.ObjectRelation{
-			Object: tuple.BuildObject(val.Type, "*"),
-		})
-	}
-
```
miparnisari, line 500: if the query is `ListObjects(user=user:*, relation=x, type=y)` but the model does not allow `user:*` for `x`, will we throw an error?

jon-whit: We shouldn't have hit this path in the first place, because there would not have been a valid ingress for the reverse expansion target if you tried to reverse expand a relationship that didn't have an allowed type restriction.

Also, even if we did, we should just filter it out with the filtered tuple iterator because it's not a valid type restriction. So in both cases we should be covered.

craigpastro (approved):

PR #538, opened 2023-02-13
chore: ensure tracing propagation at http layer

craigpastro:
## Description

Add a tracer interceptor around the http server and grpc client so that traces using the http endpoints can be propagated.

Screen Shot 2023-02-13 at 15 49 23

## Review Checklist


cmd/run/run.go
```diff
@@ -632,13 +634,15 @@ func RunServer(ctx context.Context, config *Config) error {
...
-			Handler: cors.New(cors.Options{
+			Handler: otelhttp.NewHandler(cors.New(cors.Options{
 				AllowedOrigins:   config.HTTP.CORSAllowedOrigins,
 				AllowCredentials: true,
 				AllowedHeaders:   config.HTTP.CORSAllowedHeaders,
 				AllowedMethods: []string{http.MethodGet, http.MethodPost,
 					http.MethodHead, http.MethodPatch, http.MethodDelete, http.MethodPut},
-			}).Handler(mux),
+			}).Handler(mux), "openfga_http_server", otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
+				return fmt.Sprintf("%s %s%s", r.Method, r.Host, r.URL.Path)
```
jon-whit, line 644: It'll be tricky to search these traces if the path involves parameters, and adding the `host` in the trace name also creates variability.

We should strive to keep the trace names uniform so that they are more searchable. For example,
`GET /stores/:storeId` in favor of `GET /stores/myid` and `POST /stores/:storeId/check` in favor of `POST api.us1.fga.dev/stores/:storeId/check`

craigpastro: hmm, we don't have individual routes to apply route tags, and it seems like not the best idea to do this globally with string manipulations. I wouldn't even mind skipping the span here and just using the one at the client, but I don't see how this is possible without resorting to custom code wrapping the handler.

kdubb: FYI, OpenTelemetry for HTTP I believe requires (heavily suggests?) using templatized path names for server span names. It talks a lot about reducing the cardinality of span names by, for example, showing a unique span name for every invocation of an endpoint.

Issue #540, opened 2023-02-14
Check internal error when Authorization Model is valid but no such model


adriantam:
When I call check with an authorization model ID that is valid (i.e., that fits the format) but there is no such model, check will return 500 Internal Error

For example:
POST http://localhost:8080/stores/{{store_id}}/check with 
```
{
    "tuple_key": {
        "object": "doc:X",
        "relation": "can_read",
        "user": "user:1"
    }, 
     "authorization_model_id": "01G5JAVJ41T49E9TT3SKVS7X1J"
}
```

Note that this issue only happens with check, but not expand.

PR #541, opened 2023-02-14
fix: error handling for check in auth model id not found

adriantam:
## Description
When auth model id is not found in check, it should return error auth model id not found instead of internal error.

## References
Close https://github.com/openfga/openfga/issues/540

## Review Checklist

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -246,7 +246,7 @@ func (s *Server) Check(ctx context.Context, req *openfgapb.CheckRequest) (*openf
 	model, err := s.datastore.ReadAuthorizationModel(ctx, storeID, modelID)
 	if err != nil {
 		if errors.Is(err, storage.ErrNotFound) {
-			return nil, serverErrors.AuthorizationModelNotFound(req.GetAuthorizationModelId())
+			return nil, serverErrors.AuthorizationModelNotFound(modelID)
 		}
 		return nil, err
 	}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -245,6 +245,9 @@ func (s *Server) Check(ctx context.Context, req *openfgapb.CheckRequest) (*openf
 
 	model, err := s.datastore.ReadAuthorizationModel(ctx, storeID, modelID)
 	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return nil, serverErrors.AuthorizationModelNotFound(req.GetAuthorizationModelId())
+		}
 		return nil, err
 	}
 
--- a/tests/check/check_test.go
+++ b/tests/check/check_test.go
@@ -152,13 +152,13 @@ func testBadAuthModelID(t *testing.T, engine string) {
 
 	storeID := resp.GetId()
 	model := `
-type user
+	type user
 
-type doc
-  relations
-    define viewer: [user] as self
-    define can_view as viewer
-`
+	type doc
+	  relations
+	    define viewer: [user] as self
+	    define can_view as viewer
+	`
 	_, err = client.WriteAuthorizationModel(ctx, &pb.WriteAuthorizationModelRequest{
 		StoreId:         storeID,
 		SchemaVersion:   typesystem.SchemaVersion1_1,
--- a/tests/check/check_test.go
+++ b/tests/check/check_test.go
@@ -41,20 +41,20 @@ type assertion struct {
 }
 
 func TestCheckMemory(t *testing.T) {
-	testCheck(t, "memory")
-	testBadAuthModelID(t, "memory")
+	testRunAll(t, "memory")
 }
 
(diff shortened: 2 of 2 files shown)


tests/check/check_test.go
```diff
@@ -126,3 +132,64 @@ func runTests(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
 		})
 	}
 }
+
+func testBadAuthModelID(t *testing.T, engine string) {
```
adriantam, line 136: I had considered adding this test as part of testCheck.  However, those tests are more focused on relationship between model/tuples/check tuples while for this test, there are no relationships between tuple/model/check tuples. Thus, it may make more sense to separate this out.

adriantam: Feel free to unresolve this if you feel we should move this as part of testCheck

pkg/server/server.go
```diff
@@ -245,6 +245,9 @@ func (s *Server) Check(ctx context.Context, req *openfgapb.CheckRequest) (*openf
 
 	model, err := s.datastore.ReadAuthorizationModel(ctx, storeID, modelID)
 	if err != nil {
+		if errors.Is(err, storage.ErrNotFound) {
+			return nil, serverErrors.AuthorizationModelNotFound(req.GetAuthorizationModelId())
```
jon-whit, line 249:
```suggestion
			return nil, serverErrors.AuthorizationModelNotFound(modelID)
```

tests/check/check_test.go
```diff
@@ -126,3 +132,64 @@ func runTests(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
...
+	ctx := context.Background()
+	resp, err := client.CreateStore(ctx, &pb.CreateStoreRequest{Name: "bad auth id"})
+	require.NoError(t, err)
+
+	storeID := resp.GetId()
+	_, err = client.WriteAuthorizationModel(ctx, &pb.WriteAuthorizationModelRequest{
+		StoreId:       storeID,
+		SchemaVersion: typesystem.SchemaVersion1_1,
+		TypeDefinitions: []*pb.TypeDefinition{
+			{
```
jon-whit, line 158: Please use the DSL parser and a DSL string here instead of an inline TypeDefinition struct.

e.g.

```
type user

type doc
  relations
    define viewer: [user] as self
    define can_view as viewer
```
and `parser.MustParse(modelStr)`

tests/check/check_test.go
```diff
@@ -126,3 +132,64 @@ func runTests(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
...
+		TupleKey:             tuple.NewTupleKey("doc:x", "viewer", "user:y"),
+		AuthorizationModelId: "01GS89AJC3R3PFQ9BNY5ZF6Q97",
+	})
+
+	require.Error(t, err)
+	e, ok := status.FromError(err)
+	require.True(t, ok)
+	require.Equal(t, int(pb.ErrorCode_authorization_model_not_found), int(e.Code()))
+
+	cancel()
```
jon-whit, line 194: Redundant. We have `defer cancel()` above.

tests/check/check_test.go
```diff
@@ -41,14 +42,19 @@ type assertion struct {
 
 func TestCheckMemory(t *testing.T) {
 	testCheck(t, "memory")
```
jon-whit, line 44: Let's compose these in a single `RunAll(t, engine string)` and then call that instead. This way we share all of the tests across invocations.

```
    RunAll(t, "memory")
}

func TestCheckPostgres(t *testing.T) {
    RunAll(t, "postgres")
}

...
```

jon-whit (changes requested):
Left some suggestions for changes 👍

pkg/server/server.go
adriantam: done

tests/check/check_test.go
adriantam: done

tests/check/check_test.go
adriantam: removed as suggested

tests/check/check_test.go
adriantam: updated

tests/check/check_test.go
```diff
@@ -126,3 +132,45 @@ func runTests(t *testing.T, client pb.OpenFGAServiceClient, tests checkTests) {
...
+	resp, err := client.CreateStore(ctx, &pb.CreateStoreRequest{Name: "bad auth id"})
+	require.NoError(t, err)
+
+	storeID := resp.GetId()
+	model := `type user
+
+type doc
+  relations
+    define viewer: [user] as self
+    define can_view as viewer`
```
jon-whit, line 159: Newline after the string declaration please.

jon-whit (approved):

PR #545, opened 2023-02-14
feat: improve the speed of Check by using type restrictions

miparnisari:
This change brings a performance improvement for 1.1 models:

- when doing `Check(user:jon, viewer, document:budget)` (direct check), if relation `viewer` does not allow type `user`, then we return `false` and **we save one database read**
- when doing `Check(group:fga:#member, viewer, document:budget)` (direct check), if relation `viewer` does not allow type `group#member`, then we return `false`  and **we save one database read**
- when retrieving all userset tuples for `document:budget#viewer`, if relation `viewer` only allows type `group#member`, then we return only tuples like `document:budget#viewer@group:....#member`  and we **decrease the number of tuples returned from the database**
- when retrieving all userset tuples for `document:budget#viewer`, if relation `viewer` only allows type `user:*`, then we return only tuples like `document:budget#viewer@user:*` and we **decrease the number of tuples returned from the database**

This PR includes the benchmarks in https://github.com/openfga/openfga/pull/554. We can merge #554 first and then this one will be shorter

----

image


internal/graph/check.go
```diff
@@ -372,28 +372,40 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+		}
+
 		fn1 := func(ctx context.Context) (*openfgapb.CheckResponse, error) {
-			t, err := c.ds.ReadUserTuple(ctx, storeID, tk)
-			if err != nil {
-				if errors.Is(err, storage.ErrNotFound) {
-					return &openfgapb.CheckResponse{Allowed: false}, nil
-				}
+			if c.shouldCheckDirectTuple(typesys, allowedTypesForUser, tk.GetUser()) {
+				t, err := c.ds.ReadUserTuple(ctx, storeID, tk)
```
jon-whit, line 383: Instead of nesting this inside the handler, avoid the CheckFuncHandler altogether if there is no need to check it. For example,

```
var handlers []CheckHandlerFunc

var condition bool

if condition {
    handler := func(ctx context.Context) (*openfgapb.CheckResponse, error) {
        ... // leave code as is
    }
   
    handlers = append(handlers, handler)
}

... 

return union(ctx, c.concurrencyLimit, handlers...)
```

internal/graph/check.go
```diff
@@ -372,6 +372,14 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		storeID := req.GetStoreID()
 		tk := req.GetTupleKey()
 
+		var allowedTypesForUser []*openfgapb.RelationReference
+		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 {
+			// allowedTypesForUser could be "user" or "user:*" or "group#member"
+			allowedTypesForUser, _ = typesys.GetDirectlyRelatedUserTypes(tuple.GetType(tk.GetObject()), tk.Relation)
+		}
+
+		var outerHandlers []CheckHandlerFunc
```
jon-whit, line 381: Maybe `checkfuncs`

internal/graph/check.go
```diff
@@ -459,8 +474,28 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+		outerHandlers = append(outerHandlers, fn2)
+
+		return union(ctx, c.concurrencyLimit, outerHandlers...)
+	}
+}
+
+// allowedTypesForUser could be "team" or "team:*" or "group#member"
+// shouldCheckDirectTuple returns true if allowedTypes contains "team" and the user is like type "team:fga"
+// or if allowedTypes contains "group#member" and the user is like type "group:fga#member"
+func (c *LocalChecker) shouldCheckDirectTuple(typesys *typesystem.TypeSystem, allowedTypes []*openfgapb.RelationReference, user string) bool {
```
jon-whit, line 486: There's already a function for this in the `typesystem` package. We shouldn't add this here.

You should be able to use `typesys.IsDirectlyRelated(...)`

jon-whit (changes requested):
Left a couple of comments.

internal/graph/check.go
```diff
@@ -372,6 +372,12 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		storeID := req.GetStoreID()
 		tk := req.GetTupleKey()
 
+		var allowedTypesForUser []*openfgapb.RelationReference
```
jon-whit, line 375: This code isn't necessary anymore

jon-whit, line 375: Please put this code in the CheckFuncHandler that it pertains to.

```
fn2 := func(ctx context.Context) (*openfgapb.CheckResponse, error) {
    // put code here
}
```

internal/graph/check.go
```diff
@@ -372,6 +372,12 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 {
+			// allowedTypesForUser could be "user" or "user:*" or "group#member"
+			allowedTypesForUser, _ = typesys.GetDirectlyRelatedUserTypes(tuple.GetType(tk.GetObject()), tk.Relation)
```
jon-whit, line 378: Please create a closure around these variables since we use them in multiple places in both `fn1` and `fn2`.

```
objectType := tuple.GetType(tk.GetObject())
relation := tk.GetRelation()

fn1 := func(ctx conetxt.Context) ... {
    ..
    typesystem.DirectRelationReference(objectType, relation),
    ...
}

fn2 := func(ctx context.Context) ... {
    // use 'objectType' and 'relation'
}
```

pkg/storage/storage.go
```diff
@@ -119,6 +127,12 @@ type ReadStartingWithUserFilter struct {
 	UserFilter []*openfgapb.ObjectRelation
 }
 
+type ReadUsersetTuplesFilter struct {
```
jon-whit, line 130: I noticed where we use this we have to split it everywhere. Can we please break it apart so it's more explicit?

```
type ReadUsersetTuplesFilter struct {
    ObjectType string
    ObjectID string
    ...
}
```

pkg/storage/storage.go
```diff
@@ -119,6 +127,12 @@ type ReadStartingWithUserFilter struct {
...
+	ObjectID            string                         // required
+	Relation            string                         // required
+	AllowedTypesForUser []*openfgapb.RelationReference // optional
```
jon-whit, line 133: minor nitpick, mind if we call this `AllowedUserTypeRestrictions`?

jon-whit (dismissed):

pkg/storage/storage.go
miparnisari: > where we use this we have to split it everywhere

only in `mysql.go` and `postgres.go`. Every other place (Check, the tests) don't have to. I'm gonna keep this as is

craigpastro: Is ObjectID an id or an object though? If it is an object I would prefer to name it as such.

miparnisari: currently, `ObjectID` is e.g. `doc:budget`. Do you prefer just `Object`?

craigpastro: If we have a `TupleKey` and we call `GetObject` on it, then it will return a `doc:budget` thing, so, in my case anyway, I would prefer `Object`.

pkg/storage/storage.go
```diff
@@ -214,13 +228,10 @@ var _ RelationshipTupleReader = (*ctxRelationshipTupleReader)(nil)
...
 // that don't match the object and relation provided in the filterKey.
-func filterTuples(tuples []*openfgapb.TupleKey, filterKey *openfgapb.TupleKey) []*openfgapb.Tuple {
-	targetObject := filterKey.GetObject()
-	targetRelation := filterKey.GetRelation()
-
+func filterTuples(tuples []*openfgapb.TupleKey, targetObjectID, targetRelation string) []*openfgapb.Tuple {
 	var filtered []*openfgapb.Tuple
 	for _, tk := range tuples {
-		if tk.GetObject() == targetObject && tk.GetRelation() == targetRelation {
+		if tk.GetObject() == targetObjectID && tk.GetRelation() == targetRelation {
```
craigpastro, line 234: Keep it `targetObject`?
```suggestion
		if tk.GetObject() == targetObject && tk.GetRelation() == targetRelation {
```

craigpastro (approved):
LGTM

PR #555, opened 2023-02-17
feat: modelID interceptor

craigpastro:
## Description

Model id interceptor. I am a little bit undecided if this is good or bad. It moves the "resolution" of the model id to the middleware layer. Is this where it belongs?

## Review Checklist


miparnisari:
> I am a little bit undecided if this is good or bad. It moves the "resolution" of the model id to the middleware layer. Is this where it belongs?

This would mean that all APIs that need to resolve the auth model ID don't have to do that anymore, which sounds like less code overall 🤔

miparnisari:
I'm getting two headers in the response 

```shell
[21/02/23 11:44:07] ~/GitHub/openfga (model-id-interceptor) $ curl --verbose --request POST 'http://localhost:8080/stores/01GR7A9GJPEHPSR95JBE1BV56V/check' \
--data-raw '{
    "tuple_key": {
        "user": "maria",
        "object": "doc:1",
        "relation": "writer"
    }
}'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /stores/01GR7A9GJPEHPSR95JBE1BV56V/check HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.79.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 109
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< Authorization-Model-Id: 01GSTQ6AY6XT0Z38SB1N4R1R4K <-----------------------
< Openfga-Authorization-Model-Id: 01GSTQ6AY6XT0Z38SB1N4R1R4K <-----------------------
< Vary: Origin
< X-Request-Id: db0ef803-a1b1-4ed7-8fbd-f089186b31fd
< Date: Tue, 21 Feb 2023 19:44:10 GMT
< Content-Length: 152
< 
{"code":"validation_error","message":"the 'user' field must be an object (e.g. document:1) or an 'object#relation' or a typed wildcard (e.g. group:*)"}
* Connection #0 to host localhost left intact
[21/02/23 11:44:10] ~/GitHub/openfga (model-id-interceptor) $ 

```

internal/middleware/middleware.go
```diff
@@ -40,6 +45,27 @@ func (s *wrappedServerStream) RecvMsg(m interface{}) error {
...
 		s.wrappedContext = context.WithValue(s.Context(), storeIDCtxKey, storeID)
 		fields = append(fields, zap.String(storeIDKey, storeID))
+
+		if r, ok := m.(hasGetAuthorizationModelID); ok {
+			modelID := r.GetAuthorizationModelId()
+
+			if modelID == "" {
+				if s.datastore == nil {
+					return errors.New("failed to retrieve model id")
+				}
```
miparnisari, line 55: i'm not fond of this check... I think we can change the constructor to accept a datastore, and if `nil` is passed we use a NoOp Datastore?

internal/middleware/modelid_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	t.Run("empty_modelID_in_request_but_not_in_ds", func(t *testing.T) {
+		interceptor := NewModelIDInterceptor(ds)
+
+		handler := func(ctx context.Context, req interface{}) (interface{}, error) {
+			return nil, nil
+		}
+
+		_, err := interceptor(context.Background(), &openfgapb.CheckRequest{StoreId: ulid.Make().String()}, nil, handler)
+		require.Error(t, err)
```
miparnisari, line 52: can you add a line like `require.Contains(err, "the expected error here")` so that the tests are easier to follow? i'm not sure what error is expected in this case, and `empty_modelID_in_request_but_not_in_ds` isn't helping me

internal/middleware/modelid.go
```diff
@@ -0,0 +1,74 @@
...
+func NewModelIDInterceptor(ds storage.AuthorizationModelReadBackend) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if r, ok := req.(hasGetAuthorizationModelID); ok {
+			modelID := r.GetAuthorizationModelId()
+
+			if modelID == "" {
+				storeID := r.GetStoreId()
+				if storeID == "" {
+					return nil, errors.New("no store id")
+				}
```
miparnisari, line 41: this code is not covered by tests. Can it ever happen? We have non-empty validation at the API layer, e.g. https://github.com/openfga/api/blob/main/openfga/v1/openfga_service.proto#L803

cmd/run/run.go
```diff
@@ -637,6 +639,7 @@ func RunServer(ctx context.Context, config *Config) error {
 				return status.Convert(encodedErr)
 			}),
 			runtime.WithHealthzEndpoint(healthv1pb.NewHealthClient(conn)),
+			runtime.WithOutgoingHeaderMatcher(func(s string) (string, bool) { return s, true }),
```
craigpastro, line 642: This ensures we don't add `Grpc-Metadata-` in front of our headers.

craigpastro:
The fact that this doesn't really work in the streaming endpoints means we don't have a consistent way to extract the model id from the context. If we really wanted the same to work for the streaming endpoints I think it would require a quite ugly refactor. Furthermore it doesn't seem worth it to do a database lookup to add the model id in a span one level up. For those reasons let's close this for now.

Issue #557, opened 2023-02-18
[v0.3.6] ListObjects endpoint not working correctly after new version.


xXAvoraXx:
Hello, I've described the problem here https://github.com/orgs/openfga/discussions/113#discussioncomment-5017936 before, I'm moving this issue to a topic.

In v0.3.4 before the update, the ListObjects endpoint was returning a list as expected, but with the new update (v0.3.6) the ListObjects endpoint is returning empty. Below are all scenarios and requirements.

The scenario of the diagram above is:

plan-> feature -> permission
plan -> feature -> page
plan -> feature -> page -> component

businesses subscribe to a plan_group type.
`plan_group:businessId/planId`

Each resource_access_groub has a business parent.
`resource_access_group:businessId`

Each resource_access_group is associated with a permission_access.
`permission_access:businessId/permissionName`
`resource_access_group:businessId`

Each permission_access type is associated with a permission type.
`permission_access:businessId/permissionName`

The access of the permission_access type is restricted according to some scenarios with;

- `business_member`: The person who wants to access the permission type in the context context must be a business member.

- `can_member_access`: Must be a member of the business and have access to the feature type of the business.

- `can_owner_access`: It must be the owner of the business and have access to the feature type of the business.

- `business_role_member`: Must be both a business member and have a role.

- `can_access_individual_or_role`: It must be able to access the role or directly.

- `can_access_member_and_individual_or_role`: Must be both a member of the business and have access to the feature type of the business, and must be able to access or directly access the Role.

- `can_access`: Must be both a business member and able to access the business's feature type and be able to access or directly access the Role, or Both should be a business owner and have access to the business's feature type.

The purpose is;

We offer plans to our businesses that are associated with a feature. Properties are related to permissions. Permissions keep our endpoints, for example;
`permission:UserService.GetUser`

Every business has its own user pool. Businesses can reduce and increase the authorizations of their users. I'm using the permission_access type for this. But businesses' access to permissions for plans must be managed by system administrators. The point that provides this is permission and feature.

Businesses can assign or reduce permissions to their users based on the plans they have. Businesses can both authorize individually and authorize by role.

A user can be a user in more than one business.

The business owner can access directly if the user has the owner relationship with the business type.

75 permissions are registered in permission_repo.

If the business has the highest plan: 75 permission_accesses are registered for the business.

Every registered permission_access type is registered in resource_access_group. 75 more relationships are registered.

[...]

PR #565, opened 2023-02-21
fix: propagate contextual tuples to ListObjects with concurrent checks

jon-whit:
## Description
This fixes an issue that leads to ListObjects reporting different results than Check if the unoptimized version of ListObjects is delegated to.

## References
https://github.com/openfga/openfga/issues/557

## Review Checklist


pkg/server/test/list_objects.go
```diff
@@ -51,6 +52,63 @@ func newListObjectsRequest(store, objectType, relation, user, modelID string, co
 func ListObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 	ctx := context.Background()
 
+	t.Run("list_objects_with_concurrent_checks_and_contextual_tuples", func(t *testing.T) {
```
jon-whit, line 55: Added this here as a means to ensure we don't fail this issue again in the future. I'm working on a separate PR to add more functional tests for ListObjects like we've done with Check 👍 

In the future this test will move into that other test suite once it is done.

pkg/server/test/list_objects.go
```diff
@@ -141,21 +201,6 @@ func ListObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 				expectedResult: []string{},
 				expectedError:  nil,
 			},
-			{
-				name: "returns_error_if_duplicate_contextual_tuples",
```
jon-whit, line 145: Removed this test and functionality associated with it. If the developer wants to provide a duplicate contextual tuple, then so be it. I don't see a reason why we should return an error if they provide a duplicate.

If they want to provide duplicates, then that should be fine so long as they still meet the maximum contextual tuple request limits.

pkg/server/test/list_objects.go
```diff
@@ -51,6 +52,63 @@ func newListObjectsRequest(store, objectType, relation, user, modelID string, co
...
+			Id:              ulid.Make().String(),
+			SchemaVersion:   typesystem.SchemaVersion1_1,
+			TypeDefinitions: typedefs,
+		}
+		err := ds.WriteAuthorizationModel(ctx, store, model)
+		require.NoError(t, err)
+
+		typesys := typesystem.New(model)
+
+		listObjectsQuery := &commands.ListObjectsQuery{
```
jon-whit, line 78: Note to self - we have this in a lot of places. We should consolidate to

```
func NewListObjectsQuery(...) {
    // set defaults if none are overriden (e.g. `NoopLogger`, `defaultListObjectsDeadline`, etc..)
}
```

this will cut down on boilerplate in a lot of places.

pkg/server/test/list_objects.go
```diff
@@ -51,6 +52,63 @@ func newListObjectsRequest(store, objectType, relation, user, modelID string, co
...
+			User:                 "user:jon",
+			ContextualTuples: &openfgapb.ContextualTupleKeys{
+				TupleKeys: []*openfgapb.TupleKey{
+					tuple.NewTupleKey("doc:1", "reader", "user:jon"),
+					tuple.NewTupleKey("doc:2", "reader", "user:jon"),
+				},
+			},
+		}
+
+		ctx = typesystem.ContextWithTypesystem(ctx, typesys)
```
adriantam, line 101: a bit sad that we have to do this in the test instead of calling server's code to ensure this is setup properly :(

pkg/server/test/list_objects.go
```diff
@@ -51,6 +52,63 @@ func newListObjectsRequest(store, objectType, relation, user, modelID string, co
...
+		ctx = storage.ContextWithContextualTuples(ctx, request.ContextualTuples.GetTupleKeys())
+
+		resp, err := listObjectsQuery.Execute(ctx, request)
+		require.NoError(t, err)
+
+		objects := resp.GetObjects()
+		require.Len(t, objects, 2)
```
adriantam, line 108: curious why do we need this instead of relying on the elements match below

pkg/server/test/list_objects.go
jon-whit: Yep! That's why we need the functional tests for ListObjects. Our test entrypoint should be `server.ListObjects` instead of `commands.ListObjectsQuery`.

pkg/server/test/list_objects.go
jon-whit: Good point - this is redundant. Removed 👍

adriantam (changes requested):
Please add test to validate behaviour for duplicate contextual tuple for list object + check

adriantam (approved):

pkg/server/server.go
```diff
@@ -153,11 +161,21 @@ func (s *Server) StreamedListObjects(req *openfgapb.StreamedListObjectsRequest,
...
 		return serverErrors.HandleError("", err)
 	}
 
 	typesys := typesystem.New(model)
 
+	for _, ctxTuple := range req.GetContextualTuples().GetTupleKeys() {
+		if err := validation.ValidateTuple(typesys, ctxTuple); err != nil {
+			return serverErrors.HandleTupleValidateError(err)
+		}
+	}
```
miparnisari, line 174: why isn't this also inside the non-streaming version of ListObjects?

miparnisari: also, this code is already inside `evaluate`... i'm confused

pkg/server/commands/list_objects.go
```diff
@@ -58,21 +57,18 @@ func (q *ListObjectsQuery) handler(
...
 
-	model, err := q.Datastore.ReadAuthorizationModel(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
-	if err != nil {
-		if errors.Is(err, storage.ErrNotFound) {
-			return serverErrors.AuthorizationModelNotFound(req.GetAuthorizationModelId())
-		}
-		return err
+	typesys, ok := typesystem.TypesystemFromContext(ctx)
+	if !ok {
+		panic("typesystem missing in context")
```
miparnisari, line 62: this should not be a panic since we can recover from this error by actually building the type system

PR #569, opened 2023-02-22
refactor: middleware to use grpc-middleware v2

craigpastro:
## Description

Refactor middleware to use grpc-middleware v2. For streaming rpcs I don't think we can count on adding information to the context and extracting it later. This is because in `PostMsgReceive` there is no ctx to add things to that will propate through. So let's just remove that part from the store id interceptor.

## Review Checklist


internal/middleware/authn/authn.go
```diff
@@ -1,4 +1,4 @@
-package middleware
+package authn
```
jon-whit, line 1: Moved this back into `internal` as we don't necessarily need to expose it as part of the public API yet. If we find we need it externally or get interest from the community in some way, then we can make it public. But until then let's keep things private.

jon-whit:
Here's a little video demonstrating this work. You can see that both unary and server-streaming endpoints inject the `store_id` and `authorization_model_id` fields into the logged fields. e.g. `WriteAuthorizationModel, `Check, `ListObjects`, and `StreamedListObjects` in this demo show the behavior. `CreateStore` doesn't return any of these fields because these fields don't exist as part of that API definition, and the `authorization_model_id` field isn't logged in the `WriteAuthorizationModel` RPC because it creates the model and does not have a request field for that metadata.

https://user-images.githubusercontent.com/2899204/222365935-842b166f-3ac5-41d4-9ce6-2683089bd50d.mov

jon-whit (approved):

pkg/server/server.go
```diff
@@ -464,33 +467,40 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
...
 func (s *Server) resolveAuthorizationModelID(ctx context.Context, store, modelID string) (string, error) {
 	ctx, span := tracer.Start(ctx, "resolveAuthorizationModelID")
 	defer span.End()
 
+	defer func() {
+		span.SetAttributes(attribute.KeyValue{Key: authorizationModelIDKey, Value: attribute.StringValue(modelID)})
+		grpc_ctxtags.Extract(ctx).Set(authorizationModelIDKey, modelID)
+		_ = grpc.SetHeader(ctx, metadata.Pairs(AuthorizationModelIDHeader, modelID))
+	}()
+
```
miparnisari, line 487: this is changing behavior - now, the auth model ID header is being set even if this function errors. Is there a test for this?

also, can we not introduce this new `defer func` pattern until we have agreed with the team that we want it? at least for me, i'm not used to it, so it takes me time to reason through it

pkg/middleware/storeid/storeid_test.go
```diff
@@ -55,21 +55,21 @@ func (s *mockServerStream) RecvMsg(m interface{}) error {
...
-func TestStreamingStoreIDInterceptor(t *testing.T) {
+func TestStreamingInterceptor(t *testing.T) {
 	t.Run("streaming_interceptor_with_no_GetStoreId_in_request", func(t *testing.T) {
 		handler := func(srv interface{}, stream grpc.ServerStream) error {
 			err := stream.RecvMsg(nil)
 			require.NoError(t, err)
 
 			storeID, ok := StoreIDFromContext(stream.Context())
-			require.False(t, ok)
+			require.True(t, ok)
```
miparnisari, line 65: this is confusing to me, if `ok` is true then why is `storeId` empty?

internal/middleware/authn/authn.go
miparnisari: please don't move around files in the same PR that you are changing them as it makes it hard to see what actually changed

Issue #571, opened 2023-02-23
runtime error: invalid memory address or nil pointer dereference


wemod123:
Version: 0.3.7

When call the `../list-objects` API, get below errors in console
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xd54520]
goroutine 161 [running]:
github.com/openfga/openfga/pkg/typesystem.RewriteContainsSelf.func1(0x30?)
/home/runner/work/openfga/openfga/pkg/typesystem/typesystem.go:909
github.com/openfga/openfga/pkg/typesystem.WalkUsersetRewrite(0x0, 0x11127a8)
/home/runner/work/openfga/openfga/pkg/typesystem/typesystem.go:1101 +0x2d
github.com/openfga/openfga/pkg/typesystem.RewriteContainsSelf(0xc00071ba60?)
/home/runner/work/openfga/openfga/pkg/typesystem/typesystem.go:908 +0x25
github.com/openfga/openfga/pkg/typesystem.(*TypeSystem).IsDirectlyAssignable(...)
/home/runner/work/openfga/openfga/pkg/typesystem/typesystem.go:901
github.com/openfga/openfga/internal/graph.(*ConnectedObjectGraph).findIngressesWithRewrite(0xc00071ba60, 0xc00071a530, 0xc00071bd38, 0xc0006ce140?, 0xf?)
/home/runner/work/openfga/openfga/internal/graph/graph.go:202 +0xdad
github.com/openfga/openfga/internal/graph.(*ConnectedObjectGraph).findIngressesWithRewrite(0xc00071ba60, 0xc00071a530, 0xc00071bd38, 0xc0006170f8?, 0x5?)
/home/runner/work/openfga/openfga/internal/graph/graph.go:233 +0xbaa
...
...
...
/home/runner/work/openfga/openfga/internal/graph/graph.go:129 +0x1cf
github.com/openfga/openfga/internal/graph.(*ConnectedObjectGraph).RelationshipIngresses(...)
/home/runner/work/openfga/openfga/internal/graph/graph.go:113
github.com/openfga/openfga/pkg/server/commands.(*ConnectedObjectsCommand).streamedConnectedObjects(0xc00061cc00, {0x1245370, 0xc000795d80}, 0xc000516780, 0xc0000a4240, 0xc000792450, 0xc00061736c)
[...]

PR #572, opened 2023-02-23
fix: list objects null pointer if computed userset does not exist

craigpastro:
## Description
fix: list objects null pointer if computed userset does not exist

## References
https://github.com/openfga/openfga/issues/571

## Review Checklist


internal/graph/graph.go
```diff
@@ -196,6 +196,9 @@ func (g *ConnectedObjectGraph) findIngressesWithRewrite(
 		for _, typeRestriction := range tuplesetTypeRestrictions {
 
 			r, err := g.typesystem.GetRelation(typeRestriction.GetType(), computedUserset)
+			if err != nil {
+				continue // skip computed relations on tupleset relationships if they are undefined
```
jon-whit, line 200: We should probably be more specific and check specifically on `typesystem.ErrRelationUndefined`

```
if errors.Is(err, typesystem.ErrRelationUndefined) {
    continue
}

return nil, err
```

jon-whit: Otherwise we might blindly skip something that should have led to an error otherwise.

craigpastro: Good point! Updated. Thanks!

assets/tests/listobjects_1_0_tests.yaml
```diff
@@ -0,0 +1 @@
+tests:
```
miparnisari, line 1: TODO 😆

jon-whit (approved):

PR #573, opened 2023-02-24
chore: add more tests to ListObjects

miparnisari:
Given a ListObjects query for user `u` and relation `r` and type `o` and response with objects `o1` and `o2`: `Check(u, r, 01)` and `Check(u, r, o2)` should return true


miparnisari:
Made as part of https://github.com/openfga/openfga/pull/599

Issue #576, opened 2023-02-27
One of the response objects in ListObjects returns Check->false


miparnisari:
With the following model

```python
model
  schema 1.1
type user
type folder
  relations
    define owner: [user]
    define viewer: [user,user:*] or owner
type document
  relations
    define can_read: viewer from parent
    define parent: [document,folder]
    define viewer: [user,user:*]
```

The following list objects request  

```json
{
    "type": "document",
    "relation": "can_read",
    "user": "user:anne",
    "contextual_tuples": {
        "tuple_keys": [
            {
                "object": "folder:a",
                "relation": "owner",
                "user": "user:anne"
            },
            {
                "object": "document:c",
                "relation": "parent",
                "user": "folder:a"
            },
            {
                "object": "document:d",
                "relation": "parent",
                "user": "document:c"
            }
        ]
    }
}
```

will return `{document:c, document:d}`. 

But `document:d` should not be returned as `Check(document:d#can_read@user:anne)` returns `false`.  

_Originally posted by @adriantam in https://github.com/openfga/openfga/issues/571#issuecomment-1442288119_

PR #577, opened 2023-02-27
fix: ListObjects should not return objects where Check->false

miparnisari:
Fixes https://github.com/openfga/openfga/issues/576


internal/graph/graph.go
```diff
@@ -117,13 +117,13 @@ func (g *ConnectedObjectGraph) findIngresses(target *openfgapb.RelationReference
 	key := tuple.ToObjectRelationString(target.GetType(), target.GetRelation())
 	if _, ok := visited[key]; ok {
 		// We've already visited the target so no need to do so again.
-		return nil, nil
+		return make([]*RelationshipIngress, 0), nil
```
miparnisari, line 120: in the graph unit tests we do [`require.ElementsMatch(t, test.expected, ingresses)`](https://github.com/openfga/openfga/blob/main/internal/graph/graph_test.go#L854). When those tests break the output of the diff was gigantic so i temporarily switched to `cmp.Diff`, which was very nitpicky about the fact that a nil response and a zero array response are not the same.

internal/graph/graph.go
```diff
@@ -117,13 +117,13 @@ func (g *ConnectedObjectGraph) findIngresses(target *openfgapb.RelationReference
...
 	}
 	visited[key] = struct{}{}
 
 	relation, err := g.typesystem.GetRelation(target.GetType(), target.GetRelation())
 	if err != nil {
-		return nil, err
+		return make([]*RelationshipIngress, 0), err
```
jon-whit, line 126: Why not leave this as `nil`. The semantics were such that the return should be `nil, err` if `err` is non-nil.

internal/graph/graph.go
jon-whit: Let's pass compare options to `cmp.Diff` then so that we can keep this contract the same. The semantics of returning a nil slice if a non-nil error is returned is idiomatic Go.

internal/graph/graph_test.go
```diff
@@ -545,8 +545,9 @@ func TestConnectedObjectGraph_RelationshipIngresses(t *testing.T) {
 			  relations
 			    define viewer: [group#member] as self
 			`,
-			target: typesystem.DirectRelationReference("document", "viewer"),
-			source: typesystem.DirectRelationReference("user", ""),
+			target:   typesystem.DirectRelationReference("document", "viewer"),
+			source:   typesystem.DirectRelationReference("user", ""),
+			expected: []*RelationshipIngress{},
```
jon-whit, line 550: Please add another test for this case:

```
		{
			name: "unrelated_source_and_target_relationship_involving_ttu",
			model: `
			type user

			type folder
			  relations
			    define viewer: [user] as self

			type document
  			  relations
    		    define can_read as viewer from parent
    		    define parent: [document,folder] as self
    		    define viewer: [user] as self
			`,
			target:   typesystem.DirectRelationReference("document", "can_read"),
			source:   typesystem.DirectRelationReference("document", ""),
			expected: []*RelationshipIngress{},
		},
```

pkg/server/commands/connected_objects.go
```diff
@@ -155,7 +155,18 @@ func (c *ConnectedObjectsCommand) streamedConnectedObjects(
 	g := graph.BuildConnectedObjectGraph(c.Typesystem)
 
 	// find the possible incoming edges (ingresses) between the target user reference and the source (object, relation) reference
+	_, span2 := tracer.Start(ctx, "RelationshipIngresses", trace.WithAttributes(
```
jon-whit, line 158: Another PR please? Let's follow up with this after the bug fix so that the changes herein are more apparently related to the bug fix.

miparnisari: the spans that I added in this PR i used exclusively to debug this. But in the end they didn't prove very useful. I removed this. Let me know if I should remove the `tupleEval` span as well

internal/graph/graph.go
```diff
@@ -179,32 +201,33 @@ func (g *ConnectedObjectGraph) findIngressesWithRewrite(
...
 
 		for _, typeRestriction := range tuplesetTypeRestrictions {
-
+			//TODO we ignore typeRestriction.GetRelationOrWildcard()
 			r, err := g.typesystem.GetRelation(typeRestriction.GetType(), computedUserset)
 
 			var directlyAssignable bool
-			if typeRestriction.GetType() == source.GetType() {
+			matchesSourceRelation := typeRestriction.GetType() == source.GetType() && computedUserset == source.GetRelation()
+			if matchesSourceRelation {
```
miparnisari, line 224: this is the fix

internal/graph/graph.go
```diff
@@ -19,6 +21,29 @@ var (
 	ErrResolutionDepthExceeded = errors.New("resolution depth exceeded")
 	ErrTargetError             = errors.New("graph: target incorrectly specified")
 	ErrNotImplemented          = errors.New("graph: intersection and exclusion are not yet implemented")
+
+	RelationshipIngressTransformer = cmp.Transformer("Sort", func(in []*RelationshipIngress) []*RelationshipIngress {
```
adriantam, line 25: Is this only used in the graph_test.go?  If so would it make sense to define it there?

pkg/server/commands/connected_objects.go
jon-whit: Yes, please do. The level of annotations on the traces feels very chatty.

pkg/tuple/tuple.go
```diff
@@ -63,6 +63,19 @@ func GetObjectRelationAsString(objectRelation *openfgapb.ObjectRelation) string
 	return objectRelation.GetObject()
 }
 
+func GetRelationReferenceAsString(rr *openfgapb.RelationReference) string {
+	if rr == nil {
+		return ""
+	}
+	if _, ok := rr.RelationOrWildcard.(*openfgapb.RelationReference_Relation); ok {
```
jon-whit, line 70: Nested if/else is redundant. More elegantly written as

```
if _, ok := ...; ok {
    return ...
}

if _, ok := ...; ok {
    return ...
}

panic("unexpected relation reference")
```

Please add a test for this in `pkg/tuple/tuple_test.go`.

internal/graph/graph.go
```diff
@@ -179,22 +201,21 @@ func (g *ConnectedObjectGraph) findIngressesWithRewrite(
 				},
 			}, nil
 		}
-
+		// else, we recurse on document#writer
```
jon-whit, line 204: @adriantam do these comments help you read this code? To me they add noise and don't really help, but if they help you then that's majority and we should keep em'

pkg/server/commands/connected_objects.go
```diff
@@ -345,6 +345,10 @@ func (c *ConnectedObjectsCommand) reverseExpandTupleToUserset(
 		}
 
 		tk := t.GetKey()
+		_, span := tracer.Start(ctx, "tupleEval", trace.WithAttributes(
```
jon-whit, line 348: Let's remove these debug level traces.

pkg/server/commands/connected_objects.go
miparnisari: Agreed... if there's many tuples eval'ed this would be super noisy 🤦‍♀️

internal/graph/graph.go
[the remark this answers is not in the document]
miparnisari: I can't move it there since it would create a cycle in the `imports`.

miparnisari: I don't have a preference. I mimicked what we did for `TupleKeyCmpTransformer` which is defined in `testutils` (but i can't put it next to it or i'll create a cycle of imports)

pkg/tuple/tuple.go
```diff
@@ -63,6 +63,21 @@ func GetObjectRelationAsString(objectRelation *openfgapb.ObjectRelation) string
 	return objectRelation.GetObject()
 }
 
+// GetRelationReferenceAsString returns team#member, or team:*, or an empty string if the input is nil.
+func GetRelationReferenceAsString(rr *openfgapb.RelationReference) string {
```
jon-whit, line 67: We have methods in the `typesystem` package to deal with `RelationReferences`. The RelationReference structure was introduced as a result of adding type restrictions and is technically part of the typesystem ecosystem. For that reason, can we place this in the `typesystem` package instead?

Then the tests below can change to:

```
func TestGetRelationReferenceAsString(t *testing.T) {
	require.Equal(t, "", GetRelationReferenceAsString(nil))
	require.Equal(t, "team#member", GetRelationReferenceAsString(DirectRelationReference("team", "member")))
	require.Equal(t, "team:*", GetRelationReferenceAsString(WildcardRelationReference("team")))
}
```

We should reserve the `tuple` package for operations/helpers with relationship tuples structures, not typesystem related stuff.

jon-whit (approved):

PR #582, opened 2023-03-01
chore: enhance integration tests for list objects by asserting on the streaming endpoint

miparnisari:


tests/listobjects/listobjects.go
```diff
@@ -115,23 +118,59 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
+					}, []grpc.CallOption{}...)
+					go func() {
+						for {
+							resp, err := clientStream.Recv()
+							if err != nil {
+								break
+							}
+							streamedObjectIds = append(streamedObjectIds, resp.Object)
 						}
+						done <- struct{}{}
```
adriantam, line 147: Will a more "go" approach be passing the err and streamedObjectIds to the channel?

tests/listobjects/listobjects.go
```diff
@@ -103,7 +104,9 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
 					require.NoError(t, err)
 				}
 
+				// act
```
adriantam, line 107: what is act?

miparnisari: https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/ :D

tests/listobjects/listobjects.go
miparnisari: That would be unnecessary, `done <- struct{}{}` is used to signal that `streamedObjectIds` has received all the objects  and that we can proceed with the assertions after line 149

I will add code that checks that `err is EOF`. If it's not we have a problem in the Streaming code.

tests/listobjects/listobjects.go
```diff
@@ -115,23 +119,62 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
 						e, ok := status.FromError(err)
 						require.True(t, ok)
 						require.Equal(t, assertion.ErrorCode, int(e.Code()))
 					}
+
+					if assertion.ErrorCode != 0 {
+						continue
+					}
+					// assert 3: each object in the response of ListObjects should return check -> true
+					for _, object := range resp.Objects {
```
jon-whit, line 169: Instead of `if assertion.ErrorCode != 0 { continue }` above, do you mind putting this for loop in the if statement above?

```
if assertion.ErrorCode == 0 {
    ...
    
    for _, object := range resp.Objects {
        checkResp, err := client.Check(...)
        require.NoError(t, err)
        require.True(t, ...)
    }
}
```

adriantam (dismissed):

adriantam (approved):

Issue #581, opened 2023-03-01
parallelize tests


miparnisari:
https://github.com/openfga/openfga/blob/fcfad88703d51f3e573511f4b92ea6a31d61febb/tests/check/check.go#L80

These tests should be parallelized. Also these:

https://github.com/openfga/openfga/blob/fcfad88703d51f3e573511f4b92ea6a31d61febb/tests/check/check_test.go#L25-L33

adriantam:
We probably also want to do this for ListObject tests too

PR #584, opened 2023-03-01
chore: parallelize test

adriantam:
## Description
Speed up test time by marking check/listobjects/writemodel test as parallel

## References
Close https://github.com/openfga/openfga/issues/581

## Review Checklist


adriantam:
For now, I did not convert the higher level TestCheckMemory/TestCheckPostgres/TestCheckMySQL as parallel because they require combining these tests as a single test.  There are occasions where it is useful to run the check tests against a specific database type.

tests/listobjects/listobjects.go
```diff
@@ -70,70 +70,73 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
 	require.NoError(t, err)
 
 	ctx := context.Background()
+	t.Run("runListObjectsTests"+schemaVersion, func(t *testing.T) {
```
jon-whit, line 73: This adds a lot of redundant nesting. I think it'd be more preferrable to make changes in `tests/check/check_test.go`

Namely,

```
    RunAll(t, "memory")
}

func RunAll(t *testing.T, engine string) {
    t.Run("Check", ...)
   t.Run("BadAuthModelID", ...)
}

func checkTest(t *testing.T, engine string) {
    ...
    t.Run("SchemaVersion1_1", ...)
    t.Run("SchemaVersion1_0", ...)
}
```

We should strive to be consistent with the composition in `pkg/server/test` and `pkg/server/server_test.go` as well where we can.

tests/check/check.go
```diff
@@ -72,12 +72,14 @@ func runTests(t *testing.T, schemaVersion string, client CheckTestClientInterfac
 	ctx := context.Background()
 
 	for _, test := range testCases.Tests {
-		resp, err := client.CreateStore(ctx, &openfgapb.CreateStoreRequest{Name: test.Name})
-		require.NoError(t, err)
+		test := test
+		t.Run(test.Name, func(t *testing.T) {
+			t.Parallel()
```
miparnisari, line 77: There's something odd going on. This doesn't seem to be parallelizing them.  Or, something else (in the setup?) is taking a very long time

image

jon-whit: That may just be the startup process for the MySQL container

miparnisari: I think it's that, but it's odd because I see the container pop-up in the UI very quickly. Maybe the tear down is also slow?

miparnisari: In any case, I see an improvement 

main

thisPR

tests/check/check_test.go
```diff
@@ -53,9 +54,12 @@ func testCheck(t *testing.T, engine string) {
 	)
 	require.NoError(t, err)
 	defer conn.Close()
+	t.Run("testCheck", func(t *testing.T) {
```
miparnisari, line 57: i'm curious why we need this here?

adriantam: need to ensure all the child test are done before tearing down connection

miparnisari: https://pkg.go.dev/testing#T.Run

> Run runs f as a subtest of t called name. It runs f in a separate goroutine and **blocks until f returns or calls t.Parallel to become a parallel test.** Run reports whether f succeeded (or at least did not fail before calling t.Parallel). Run may be called simultaneously from multiple goroutines, but all such calls must return before the outer test function for t returns.

I'm not sure if this will be enough 🤔

adriantam: We didn't see premature tear down of connection so far.

miparnisari: Yesterday when I attempted something similar (not your change but very similar), I saw that error.

https://go.dev/play/p/MsVMZ5mrqPs

miparnisari: Ok, yeah, this works! https://go.dev/blog/subtests

```go
func TestTeardownParallel(t *testing.T) {
    // <setup code>
    // This Run will not return until its parallel subtests complete.
    t.Run("group", func(t *testing.T) {
        t.Run("Test1", parallelTest1)
        t.Run("Test2", parallelTest2)
        t.Run("Test3", parallelTest3)
    })
    // <tear-down code>
}
```

miparnisari (approved):

tests/check/check_test.go
```diff
@@ -53,9 +54,12 @@ func testCheck(t *testing.T, engine string) {
...
+
+		RunSchema1_1CheckTests(t, pb.NewOpenFGAServiceClient(conn))
```
jon-whit, line 59: These should be parallelized as well. Right now they are running sequentially.

This is why I suggested we refactor a bit and move these up a level (see https://github.com/openfga/openfga/pull/584#discussion_r1122058504).

I think what we want, for consistency and the most parallelism, is something like the following:

```
    RunAllCheckTests(t, "memory")
}

func RunAllCheckTests(t *testing.T, engine string) {
    t.Run("Check", func(t *testing.T) { t.Parallel(); checkTest(t, engine) }
    t.Run("BadAuthModelID", func(t *testing.T) { t.Parallel(); badAuthModelIDTest(t, engine) }
}

func checkTest(t *testing.T, engine string) {
    ... // boilerplate setup

    t.Run("SchemaVersion1_0", func(t *testing.T) {
    
    })

   t.Run("SchemaVersion1_1", func(t *testing.T) {
   
   })
}
```

This will make our tests more parallel and will also make them more consistent with other tests in this codebase. For example:
https://github.com/openfga/openfga/blob/main/pkg/server/test/server.go

The pattern we've been trying to follow lately (though we need to refactor some) is to name test signatures of the form `func somethingTest(t *testing.T, ... args)` and reserve function names like `TestSomething` for actual go test signatures (e.g. `TestSomething(t *testing.T)`).

jon-whit:
> For now, I did not convert the higher level TestCheckMemory/TestCheckPostgres/TestCheckMySQL as parallel because they require combining these tests as a single test. There are occasions where it is useful to run the check tests against a specific database type.

I'm not sure I understand what you mean, could you elaborate? Why would have they have to be combined as a single test?

adriantam:
> > For now, I did not convert the higher level TestCheckMemory/TestCheckPostgres/TestCheckMySQL as parallel because they require combining these tests as a single test. There are occasions where it is useful to run the check tests against a specific database type.
> 
> I'm not sure I understand what you mean, could you elaborate? Why would have they have to be combined as a single test?

I am saying we shouldn't combine into one as noted in the original issue.

tests/listobjects/listobjects_test.go
```diff
@@ -24,10 +24,6 @@ func TestCheckMySQL(t *testing.T) {
 }
 
 func testRunAll(t *testing.T, engine string) {
-	testListObjects(t, engine)
-}
-
-func testListObjects(t *testing.T, engine string) {
 	cfg := run.MustDefaultConfigWithRandomPorts()
 	cfg.Log.Level = "none"
 	cfg.Datastore.Engine = engine
```
miparnisari, line 29: One of the server configs is `ListObjectsMaxResults`. In an upcoming PR, I will add testing for this parameter, so I will need each test to create a new server. Can you make the change in this PR so I don't have a lot of merge conflicts?

tests/check/check.go
```diff
@@ -72,12 +72,14 @@ func runTests(t *testing.T, schemaVersion string, client CheckTestClientInterfac
 	ctx := context.Background()
 
 	for _, test := range testCases.Tests {
-		resp, err := client.CreateStore(ctx, &openfgapb.CreateStoreRequest{Name: test.Name})
-		require.NoError(t, err)
+		test := test
```
jon-whit, line 75: newline after this please, just to add some visual spacing.

tests/check/check_test.go
```diff
@@ -54,25 +49,30 @@ func testCheck(t *testing.T, engine string) {
 	require.NoError(t, err)
 	defer conn.Close()
 
-	RunSchema1_1CheckTests(t, pb.NewOpenFGAServiceClient(conn))
-	RunSchema1_0CheckTests(t, pb.NewOpenFGAServiceClient(conn))
+	t.Run("RunAll", func(t *testing.T) {
+		t.Run("testCheck", func(t *testing.T) {
```
jon-whit, line 53: I'd omit the `test` prefix. It just stutters. Lets name as `Check` and `BadAuthModelID`. So we'll get, for example:

`TestCheckMemory/RunAll/Check`
`TestCheckMemory/RunAll/BadAuthModelID`

tests/check/check_test.go
```diff
@@ -54,25 +49,30 @@ func testCheck(t *testing.T, engine string) {
...
 
-func testBadAuthModelID(t *testing.T, engine string) {
-
-	cfg := run.MustDefaultConfigWithRandomPorts()
-	cfg.Log.Level = "none"
-	cfg.Datastore.Engine = engine
-
-	cancel := tests.StartServer(t, cfg)
-	defer cancel()
+func testCheck(t *testing.T, conn *grpc.ClientConn) {
```
jon-whit, line 64: Please refactor this a bit so that the composition will allow us to import a single test entrypoint in Sandcastle. For example, like we talked about,

```
package check
...

func RunAllTests(t *testing.T, client CheckTestClientInterface) {
    t.Run("RunAllTests", func(t *testing.T) {
        t.Run("Check", ...)
        t.Run("BadAuthModelID", ...)
    })
}

func testCheck(t *testing.T, client CheckTestClientInterface) {
    t.Run("Schema1_1", func(t *testing.T) {
        t.Parallel()
        schema1_1CheckTests(t, client)
    }
    ...
```

This way we only have to import `RunAllTests` to run all of the Check tests, instead of having to be careful to run each and every one of them.

tests/writemodel/writemodel.go
```diff
@@ -354,13 +354,16 @@ type ClientInterface interface {
...
 func RunTests(t *testing.T, client ClientInterface) {
 	ctx := context.Background()
-	resp, err := client.CreateStore(ctx, &pb.CreateStoreRequest{Name: "write_model_test"})
-	require.NoError(t, err)
-
-	storeID := resp.GetId()
 
 	for name, test := range testCases {
+		name := name
+		test := test
```
jon-whit, line 360: Newline please

tests/writemodel/writemodel.go
```diff
@@ -354,13 +354,16 @@ type ClientInterface interface {
 // for the server.
 func RunTests(t *testing.T, client ClientInterface) {
```
jon-whit, line 355: For consistency, do you mind structuring this the same as our other test suites? There's no need for this to be publicly exported.

In Check, for example, we have

```
// check.go
func runTests(t *testing.T, ..., client ClientInterface) {
    ...
}    
```
and then we use that to compose the other top-level tests.

tests/listobjects/listobjects_test.go
```diff
@@ -41,7 +37,26 @@ func testListObjects(t *testing.T, engine string) {
...
-	RunSchema1_1ListObjectsTests(t, pb.NewOpenFGAServiceClient(conn))
-	RunSchema1_0ListObjectsTests(t, pb.NewOpenFGAServiceClient(conn))
+func testListObjects(t *testing.T, client ListObjectsClientInterface) {
+	t.Run("Schema1_1", func(t *testing.T) {
+		t.Parallel()
+		RunSchema1_1ListObjectsTests(t, client)
+	})
+	t.Run("Schema1_0", func(t *testing.T) {
+		t.Parallel()
+		RunSchema1_0ListObjectsTests(t, client)
```
jon-whit, line 60: Please unexport the 1.0 Check and ListObjects tests as we do not want anyone depending on them, especially given we are on the deprecation cycle for them. Technically we don't really need the 1.1 `RunSchema1_1...` tests to be exported as well, because we want to encourage usage of `RunAllTests` externally anyways.

adriantam: make runschema to be private

PR #595, opened 2023-03-01
chore: update linter version

miparnisari:


jon-whit (approved):

Makefile
```diff
@@ -13,7 +13,7 @@ install-tools: download ## Install developer tooling
 	@cd tools && go list -f '{{range .Imports}}{{.}} {{end}}' tools.go | CGO_ENABLED=0 xargs go install -mod=readonly
 
 .PHONY: lint
-lint:  ## Lint Go source files
+lint: install-tools ## Lint Go source files
```
miparnisari, line 16: i want make to use the version installed in `tools`, not the global `golangci-lint` that i have in my machine

PR #596, opened 2023-03-01
fix: read userset tuples with a filter like group#member

miparnisari:
This is a follow-up fix for https://github.com/openfga/openfga/pull/545


pkg/storage/mysql/mysql.go
```diff
@@ -310,7 +310,7 @@ func (m *MySQL) ReadUsersetTuples(ctx context.Context, store string, filter stor
 		orConditions := sq.Or{}
 		for _, userset := range filter.AllowedUserTypeRestrictions {
 			if _, ok := userset.RelationOrWildcard.(*openfgapb.RelationReference_Relation); ok {
-				orConditions = append(orConditions, sq.Like{"_user": userset.Type + "%#" + userset.GetRelation()})
+				orConditions = append(orConditions, sq.Like{"_user": userset.Type + ":%#" + userset.GetRelation()})
```
jon-whit, line 313: Can we verify that we have proper indexes to cover this query pattern?

pkg/storage/test/tuples.go
```diff
@@ -317,6 +318,98 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
...
+			},
+		})
+		require.NoError(t, err)
+
+		iter := storage.NewTupleKeyIteratorFromTupleIterator(gotTuples)
+		defer iter.Stop()
+
+		var gotTupleKeys []*openfgapb.TupleKey
+		for {
+			tk, err := iter.Next()
```
jon-whit, line 348: I suggest simplifying this to

```
tk, err := iter.Next()
require.NoError(t, err)
require.Equal(t, tk, tuple.NewTupleKey("document:1", "viewer", "group:eng#member")

_, err := iter.Next()
```
This is because you expect the iterator to only yield one value and then the second time the iterator should yield `storage.ErrIteratorDone`.

pkg/storage/test/tuples.go
```diff
@@ -317,6 +318,98 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		_, err = gotTuples.Next()
 		require.ErrorIs(t, err, storage.ErrIteratorDone)
 	})
+
+	t.Run("reading_userset_tuples_with_filter_made_of_direct_relation_reference", func(t *testing.T) {
```
jon-whit, line 322: Would you mind just extending this test slightly so that it returns filtered iterator with more than one type restriction?

e.g. add one more tuple

```
document:1#viewer@team:tigers#member
```
and then
```
ReadUsersetTuples(ctx, ..., storage.ReadUsersetTuplesFilter {
    Object: "document:1",
    Relation: "viewer",
    AllowedUserTypeRestrictions: []*openfgapb.RelationReference{
        typesystem.DirectRelationReference("group", "member"),
        typesystem.DirectRelationReference("team", "member"),
    },
})
```
This will give us confidence that our SQL `IN` or union works as we expect.

pkg/storage/test/tuples.go
```diff
@@ -317,6 +318,98 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
...
+
+		// Then the iterator should run out
+		_, err = gotTuples.Next()
+		require.ErrorIs(t, err, storage.ErrIteratorDone)
+
+		require.Len(t, gotTupleKeys, 1)
+		require.Equal(t, gotTupleKeys[0], tuple.NewTupleKey("document:1", "viewer", "group:eng#member"))
+	})
+
+	t.Run("reading_userset_tuples_with_filter_made_of_wildcard_relation_reference", func(t *testing.T) {
```
jon-whit, line 368: Would you also just mix/match with wildcard + `type#relation` just to make sure our SQL union works across variations of typed wildcards and userset relations.

e.g. just change to 

```
ReadUsersetTuples(ctx, ..., storage.ReadUsersetTuplesFilter{
    Object: "document:1",
    Relation: "viewer",
    AllowedUserTypeRestrictions: []*openfgapb.RelationReference{
        typesystem.DirectRelationReference("group", "member"),
        typesystem.WildcardRelationReference("user"),
    },
})
```
and then call `Next()` 2x and assert no error on each call and on the 3rd call to `Next()` assert that you get the `storage.ErrIteratorDone`.

This will give us confidence the union across typed wildcard and userset relation type restrictions works as we expect.

jon-whit (changes requested):
Left some recommendations.

pkg/storage/mysql/mysql.go
miparnisari: image

adriantam (approved):

Issue #583, opened 2023-03-01
Consolidate check and list object unit test cases


adriantam:
Currently we had separate models used in testing check and list object.  Ideally, we use the same model/tuples between the tests.

PR #601, opened 2023-03-02
chore: consolidate integration tests between Check and ListObjects APIs

adriantam:
## Description
Consolidate between check and list object unit tests

## References
Close https://github.com/openfga/openfga/issues/583

## Review Checklist


tests/listobjects/listobjects.go
```diff
@@ -104,25 +105,36 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
 						Relation:         assertion.Request.Relation,
 						User:             assertion.Request.User,
-						ContextualTuples: assertion.Request.ContextualTuples,
+						ContextualTuples: contextualTuples,
 					})
 
 					if assertion.ErrorCode == 0 {
 						require.NoError(t, err)
-						require.Subset(t, assertion.Expectation, resp.Objects)
+						require.ElementsMatch(t, assertion.Expectation, resp.Objects,
```
adriantam, line 125: note that the original UT infrastructure has bugs in that it cannot detect expected element in list that is not in the actual list.

tests/listobjects/listobjects.go
```diff
@@ -104,25 +105,36 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
 					require.NoError(t, err)
 				}
 
-				for _, assertion := range stage.Assertions {
+				for _, assertion := range stage.ListAssertions {
+					var contextualTuples *pb.ContextualTupleKeys
```
adriantam, line 109: this is needed so that the YAML contextual tuple structure looks identical

tests/listobjects/listobjects.go
miparnisari: This was there because if `ListObjectsMaxResults = 5` and `len(expected) = 6` and `len(actual) = 5`, we could not do a strict equality check

adriantam: But why would expected be 6 if LIstObjectsMaxResults is 5?  Should we create a unique test so that the expected = 5?

adriantam: Otherwise, there is a risk that the UT does not truly detect if list object misses some of the expected tuples

miparnisari: > But why would expected be 6 if LIstObjectsMaxResults is 5?

Because there are 6 possible results that can be returned, and we cannot predict which of those will be returned since neither algorithm is deterministic

adriantam: I would argue that these are more exception rather than the rule ... we can use the subset specifically for these tests.  For normal test, we definitely should use exact match.

tests/listobjects/listobjects.go
```diff
@@ -108,19 +109,31 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
+							TupleKeys: assertion.ContextualTuples,
+						}
+					}
 					resp, err := client.ListObjects(ctx, &pb.ListObjectsRequest{
 						StoreId:          storeID,
 						Type:             assertion.Request.Type,
 						Relation:         assertion.Request.Relation,
 						User:             assertion.Request.User,
-						ContextualTuples: assertion.Request.ContextualTuples,
+						ContextualTuples: contextualTuples,
```
miparnisari, line 126:
```suggestion
						ContextualTuples: &pb.ContextualTupleKeys{
							TupleKeys: assertion.ContextualTuples,
						},
```

tests/listobjects/listobjects.go
```diff
@@ -108,19 +109,31 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
 
-				for _, assertion := range stage.Assertions {
+				for _, assertion := range stage.ListAssertions {
 					// assert 1: on regular list objects endpoint
+
+					var contextualTuples *pb.ContextualTupleKeys
+					if len(assertion.ContextualTuples) > 0 {
+						contextualTuples = &pb.ContextualTupleKeys{
+							TupleKeys: assertion.ContextualTuples,
+						}
```
miparnisari, line 119: this isn't needed, it's OK to pass an empty array i think

tests/listobjects/listobjects.go
```diff
@@ -108,19 +109,31 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
 						User:             assertion.Request.User,
-						ContextualTuples: assertion.Request.ContextualTuples,
+						ContextualTuples: contextualTuples,
 					})
 
 					if assertion.ErrorCode == 0 {
 						require.NoError(t, err)
-						require.ElementsMatch(t, assertion.Expectation, resp.Objects)
+						require.ElementsMatch(t, assertion.Expectation, resp.Objects,
+							fmt.Sprintf("ListObject(Type:%s,Relation:%s,User:%s,ContextualTuples:%v)\n",
```
miparnisari, line 132: what is the `fmt.Sprintf("ListObject ...` doing? if it's a debug message, i'm pretty sure the standard output is enough

adriantam: Unfortunately it does not.
```
                    listobjects.go:130:
                        	Error Trace:	/Users/adriantam/src/github.com/openfga/openfga/tests/listobjects/listobjects.go:130
                        	Error:      	elements differ

                        	            	extra elements in list A:
                        	            	([]interface {}) (len=1) {
                        	            	 (string) (len=10) "document:1"
                        	            	}


                        	            	extra elements in list B:
                        	            	([]interface {}) (len=1) {
                        	            	 (string) (len=11) "document:1a"
                        	            	}


                        	            	listA:
                        	            	([]string) (len=1) {
                        	            	 (string) (len=10) "document:1"
                        	            	}


                        	            	listB:
                        	            	([]string) (len=1) {
                        	            	 (string) (len=11) "document:1a"
                        	            	}
                        	Test:       	TestListObjectsMySQL/RunAll/ListObjects/Schema1_1/this_with_contextual_tuples
```

I cannot tell what is the user/relation/object + any contextual tuples that cause the error.  It becomes problematic with more than 1 list object assertion in the test

tests/listobjects/listobjects.go
adriantam: changed

tests/listobjects/listobjects.go
miparnisari: Before that error trace you should see the test name, once you have that it's easy to find the test data

adriantam: Yes, I see the name.  However, how can we tell which list object assertion fail if there are more than 1 assertion?

miparnisari: Ah... in that case i think it's prettier to add `t.Logf("ListObject request: %s. Contextual tuples: %s", assertion.Request, assertion.ContextualTuples)` before the `require`. It will only print if you add `-v` to `go test`, or in case there was a failure.

assets/tests/consolidate_1_1_tests.yaml
```diff
@@ -666,12 +915,17 @@ tests:
...
           - tuple:
               object: document:1
               relation: viewer
               user: user:aardvark
             expectation: true
+          - tuple:
+              object: document:1
+              relation: viewer
+              user: user:badger
+            expectation: true
```
miparnisari, line 928: This assertion and this one

```
- tuple:
              object: document:3
              relation: viewer
              user: user:cheetah
            expectation: true
```

are pretty much the same in the current setup of tuples. 

Can you replace line 904 with `relation: writer`?

assets/tests/consolidate_1_1_tests.yaml
```diff
@@ -27,6 +27,18 @@ tests:
               relation: viewer
               user: user:badger
             expectation: false
+        listAssertions:
```
miparnisari, line 30:
```suggestion
        listObjectsAssertions:
```

assets/tests/consolidate_1_1_tests.yaml
```diff
@@ -2097,3 +2909,84 @@ tests:
                 relation: parent
                 user: user:*
             errorCode: 2027
```
miparnisari, line 2911:
```suggestion
            errorCode: 2027
        listAssertions:
          - request:
              type: document
              relation: viewer
              user: user:aardvark
            contextualTuples:
              - object: document:1
                relation: parent
                user: user:*
            errorCode: 2027
```

assets/tests/consolidate_1_1_tests.yaml
```diff
@@ -2065,7 +2877,7 @@ tests:
           type document
             relations
               define viewer: [user] as self
-        assertions:
+        checkAssertions:
```
miparnisari, line 2880:
```
        listAssertions:
          - request:
              type: document
              relation: viewer
              user: user:aardvark
            contextualTuples:
              - object: document:1
                relation: viewer
                user: user:*
            errorCode: 2027
```

assets/tests/consolidate_1_1_tests.yaml
```diff
@@ -2087,7 +2899,7 @@ tests:
             relations
               define parent: [folder] as self
               define viewer as viewer from parent
-        assertions:
+        checkAssertions:
```
miparnisari, line 2902: i see there's a few test cases where we haven't added `listAssertions`. Can you add assertions for the test cases involving contextual tuples? Since this code is not well covered with tests:

image

adriantam: Thanks for the note - I will add them in subsequent PR (this one is large enough as is :) )

assets/tests/consolidate_1_1_tests.yaml
```diff
@@ -1633,6 +2297,20 @@ tests:
                 relation: writer
                 user: user:badger
             expectation: true
+        listAssertions:
```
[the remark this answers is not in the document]
adriantam: I will also add the corresponding check test

tests/listobjects/listobjects.go
```diff
@@ -133,11 +142,13 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
-						Relation:         assertion.Request.Relation,
-						User:             assertion.Request.User,
-						ContextualTuples: assertion.Request.ContextualTuples,
+						StoreId:  storeID,
+						Type:     assertion.Request.Type,
+						Relation: assertion.Request.Relation,
+						User:     assertion.Request.User,
+						ContextualTuples: &pb.ContextualTupleKeys{
+							TupleKeys: assertion.ContextualTuples,
+						},
```
miparnisari, line 151:
```suggestion
						ContextualTuples: contextualTuples,
```

tests/listobjects/listobjects.go
```diff
@@ -108,19 +109,27 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
...
 				}
 
-				for _, assertion := range stage.Assertions {
+				for _, assertion := range stage.ListObjectAssertions {
 					// assert 1: on regular list objects endpoint
+					t.Logf("ListObject request: %s. Contextual tuples: %s", assertion.Request, assertion.ContextualTuples)
+
+					contextualTuples := &pb.ContextualTupleKeys{
+						TupleKeys: assertion.ContextualTuples,
+					}
```
miparnisari, line 118: is this used?

tests/listobjects/listobjects.go
```diff
@@ -108,19 +109,27 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
 					require.NoError(t, err)
 				}
 
-				for _, assertion := range stage.Assertions {
+				for _, assertion := range stage.ListObjectAssertions {
 					// assert 1: on regular list objects endpoint
+					t.Logf("ListObject request: %s. Contextual tuples: %s", assertion.Request, assertion.ContextualTuples)
+
+					contextualTuples := &pb.ContextualTupleKeys{
```
jon-whit, line 116: Please inline this below

```
ListObjects(ctx, &pb.ListObjectsRequest{
    ...
    ContextualTuples: &pb.ContextualTupleKeys{
        TupleKeys: assertion.ContextualTuples,
    },
})
```

tests/listobjects/listobjects.go
```diff
@@ -176,7 +187,7 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
 							checkResp, err := client.Check(ctx, &pb.CheckRequest{
 								StoreId:          storeID,
 								TupleKey:         tuple.NewTupleKey(object, assertion.Request.Relation, assertion.Request.User),
-								ContextualTuples: assertion.Request.ContextualTuples,
+								ContextualTuples: contextualTuples,
 							})
 							require.NoError(t, err)
 							require.True(t, checkResp.Allowed, fmt.Sprintf("Expected Check(%s#%s@%s) to be true, got false", object, assertion.Request.Relation, assertion.Request.User))
```
jon-whit, line 193: Adding this extra log message below isn't necessary, because the default output makes it clear that the expectation was true and not false. We're purposefully not using this field for that reason elsewhere, so to be consistent let's drop the verbose message here as well.

tests/listobjects/listobjects.go
```diff
@@ -108,19 +109,27 @@ func runTests(t *testing.T, schemaVersion string, client ListObjectsClientInterf
 					require.NoError(t, err)
 				}
 
-				for _, assertion := range stage.Assertions {
+				for _, assertion := range stage.ListObjectAssertions {
 					// assert 1: on regular list objects endpoint
+					t.Logf("ListObject request: %s. Contextual tuples: %s", assertion.Request, assertion.ContextualTuples)
```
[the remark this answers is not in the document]
adriantam: See https://github.com/openfga/openfga/pull/601#discussion_r1125309481 - we need this.  Otherwise, it becomes very difficult to tell if there are more than 1 list obj assertion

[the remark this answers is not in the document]
miparnisari: It only appears if you run `go test -v` or if the test fails

tests/listobjects/listobjects.go
adriantam: removed contextualTuples so we need to stick with &pb.ContextualTupleKeys

tests/listobjects/listobjects.go
adriantam: removed as per above

tests/listobjects/listobjects.go
adriantam: we no longer have contextualTuples variable

tests/check/check.go
```diff
@@ -59,9 +59,9 @@ func runTests(t *testing.T, schemaVersion string, client CheckTestClientInterfac
 	var b []byte
 	var err error
 	if schemaVersion == typesystem.SchemaVersion1_1 {
-		b, err = assets.EmbedTests.ReadFile("tests/check_1_1_tests.yaml")
+		b, err = assets.EmbedTests.ReadFile("tests/consolidate_1_1_tests.yaml")
```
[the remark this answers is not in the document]
adriantam: done

tests/listobjects/listobjects.go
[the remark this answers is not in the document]
jon-whit: If you need to provide more context in the logged failure message, then that's the good use for the `msg` field in the `require` assertion. I think that's a better approach than helper logging in the tests. For example, in this case it'd be more advantageous to do:

```
extraErrorMsg := "Expectation failed. ListObjects request (%s), ContextualTuples (%s)"

...
require.Equal(t, ...., extraErrorMsg)
```

This way you literally only see the message if an expectation was  not met, and in that case you'll have clear context where it came from.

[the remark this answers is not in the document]
adriantam: swap to use extraErrorMsg as suggested above

jon-whit (approved):

miparnisari (approved):

PR #570, opened 2023-02-22
feat: tuning of OTEL parameters

adriantam:
## Description
Tuning of OTEL parameters for retry/memory etc.

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #605, opened 2023-03-04
chore: update otel-collector-config.yaml

miparnisari:
Follow-up of https://github.com/openfga/openfga/pull/570

Related docs: https://github.com/openfga/openfga.dev/pull/379

Changes:
- I reviewed https://opentelemetry.io/docs/collector/configuration/ and modified according to the docs.
- OTEL now exports Prometheus metrics on `http://otel-collector:2113/metrics`:

    otel collector

- The Prometheus container now listens to `otel-collector:2113/metrics` instead of `openfga:2112/metrics`:
     prometheus server

The big picture:

big picture


.config-schema.json
```diff
@@ -186,29 +186,6 @@
...
-                            "description": "Protocol used to send metrics to the collector",
-                            "type": "string",
-                            "enum": ["grpc", "http"],
-                            "default": "grpc",
-                            "x-env-variable": "OPENFGA_OTEL_METRICS_PROTOCOL"
-                        }
-                    }
-                }
-            }
-        },
```
miparnisari, line 211: don't see them mentioned in https://github.com/openfga/openfga/blob/58591cd8dc6de6de599397216163485f159fae6e/cmd/run/flags.go#

telemetry/otel-collector-config.yaml
```diff
@@ -18,27 +27,31 @@ exporters:
       num_consumers: 10
       queue_size: 10000
       # this should work out to about 8M batches/second for 10s buffer
+  # Data sources: traces, metrics, logs
   logging:
     loglevel: debug
-  prometheus:
-    endpoint: "0.0.0.0:1234"
```
miparnisari, line 24: this isn't doing anything

telemetry/otel-collector-config.yaml
```diff
@@ -1,9 +1,26 @@
...
     protocols:
+      http:
+        endpoint: "0.0.0.0:4318"
       grpc:
+        endpoint: "0.0.0.0:4317"
+  # Data sources: metrics
+  prometheus:
+    config:
+      scrape_configs:
+        - job_name: otel-collector
```
jon-whit, line 16: Please use the same `job_name` structure as we are in FGA. This was named intentionally based on the recommendation of OTEL.

The job name should be a fully qualified name composed of a namespace and service name descriptor (e..g `fga/openfga`). And would you mind adding the `service` label here?

https://opentelemetry.io/docs/reference/specification/compatibility/prometheus_and_openmetrics/#resource-attributes-1 for more info.

telemetry/otel-collector-config.yaml
```diff
@@ -1,9 +1,29 @@
...
+          scrape_interval: 1s
+          static_configs:
+            - targets: [ openfga:2112 ]
+              labels:
+                service: openfga
 
+# How you send data to one or more backends. Configuring an exporter does not enable it.
 exporters:
+  # Data sources: traces
+  # TODO will be deprecated July 2023 https://github.com/open-telemetry/opentelemetry-specification/pull/2858
```
jon-whit, line 26: Please add this as an issue instead of a comment. It'll be forgotten if we leave it here as a comment.

.config-schema.json
jon-whit: That is correct, we dropped the OTEL metrics integration altogether, so this can be removed. It's residual 👍

telemetry/otel-collector-config.yaml
```diff
@@ -1,9 +1,29 @@
+# https://opentelemetry.io/docs/collector/configuration/
+
+# How data gets into the Collector
 receivers:
+  # Data sources: traces, metrics, logs
   otlp:
     protocols:
+      http:
```
jon-whit, line 8: Please omit this as it is unused. We export traces in `otlp` format over the `grpc` protocol. We do not currently have any additional configuration that we can provide to switch the protocol from `grpc` to `http`.

telemetry/prometheus/prometheus.yml
```diff
@@ -7,4 +7,4 @@ scrape_configs:
     metrics_path: /metrics
     static_configs:
       - targets:
-          - 'openfga:2112'
\ No newline at end of file
+          - 'otel-collector:1234'
```
jon-whit, line 10: Can we keep this more consistent with the default metrics port? Maybe `2113` instead of `2112`?

telemetry/otel-collector-config.yaml
miparnisari: https://github.com/openfga/openfga/issues/609

telemetry/otel-collector-config.yaml
miparnisari: Ah! You are right..

telemetry/prometheus/prometheus.yml
miparnisari: Sounds good

telemetry/otel-collector-config.yaml
```diff
@@ -18,27 +35,39 @@ exporters:
...
       queue_size: 10000
       # this should work out to about 8M batches/second for 10s buffer
+  # Data sources: traces, metrics, logs
   logging:
     loglevel: debug
+  # Data sources: metrics
   prometheus:
-    endpoint: "0.0.0.0:1234"
+    endpoint: otel-collector:2113
+    namespace: default
```
jon-whit, line 44: Please omit this, because it prefixes every metric with `default_`, and this breaks our Grafana dashboard metrics.

miparnisari: Ah! I mis-read the docs for this exporter, I thought the `namespace` was mandatory, but it's not.

Removed and confirmed Grafana still works:

image

jon-whit (approved):

PR #614, opened 2023-03-07
fix: inject typesystem into context in benchmarks

jon-whit:
## Description
This fixes the ListObjects benchmarks by injecting the needed `typesystem` into the context before calling `Execute`.

## Review Checklist

### DIFF
--- a/pkg/server/test/list_objects.go
+++ b/pkg/server/test/list_objects.go
@@ -539,6 +539,8 @@ func BenchmarkListObjectsWithReverseExpand(b *testing.B, ds storage.OpenFGADatas
 
 	var r *openfgapb.ListObjectsResponse
 
+	ctx = typesystem.ContextWithTypesystem(ctx, typesystem.New(model))
+
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		r, _ = listObjectsQuery.Execute(ctx, &openfgapb.ListObjectsRequest{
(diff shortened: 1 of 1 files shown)


miparnisari:
how did this break? this is passing on the latest `main`

adriantam:
> how did this break? this is passing on the latest `main`: https://github.com/openfga/openfga/actions/runs/4359615923/jobs/7621593177

Main also broke - https://github.com/openfga/openfga/actions/runs/4359615923/jobs/7621593177#step:5:42
Unfortunately it does not mark as fail on CICD

adriantam (approved):

miparnisari:
> Main also broke - https://github.com/openfga/openfga/actions/runs/4359615923/jobs/7621593177#step:5:42
Unfortunately it does not mark as fail on CICD

@adriantam can you mark it as fail please?

jon-whit:
> 🤦‍♀️ @adriantam can you mark it as fail please?
@adriantam @miparnisari please don't mark the bench workflow as required, because it takes 10+ minutes. This will kill our PR CI/CD responsiveness. That workflow was meant to be a reviewable artifact but not a blocker for PR merges.

jon-whit:
> I don't understand what the 10 minutes have to do with it. The step will run no matter what, i'm saying we should fail the workflow if the return code is non-zero. Otherwise how are we going to detect problems?

I think we're not communicating clearly here.. I don't want to wait 10 minutes every PR build to finish waiting on the benchmark stage. If we can fail the benchmark workflow but not require it to pass, then that's fine. But let's not make it a required build stage to pass.

PR #617, opened 2023-03-08
chore: mark benchmark step as failure if status code is not 0

miparnisari:
Follow-up of https://github.com/openfga/openfga/pull/614

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -30,10 +30,6 @@ jobs:
         run: |
           set -o pipefail
           make bench | tee bench_output.txt
-          if [ $? -ne 0 ]; then
-            echo "Benchmark failed"
-            exit 1
-          fi
 
       # Download previous benchmark result from cache (if exists)
       - name: Download previous benchmark data
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -27,7 +27,12 @@ jobs:
 
       # Run benchmark with `go test -bench` and stores the output to a file
       - name: Run benchmark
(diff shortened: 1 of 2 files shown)


adriantam (approved):

.github/workflows/main.yaml
```diff
@@ -27,7 +27,12 @@ jobs:
 
       # Run benchmark with `go test -bench` and stores the output to a file
       - name: Run benchmark
-        run: make bench | tee bench_output.txt
+        run: |
+          make bench | tee bench_output.txt
+          if [ $? -ne 0 ]; then
```
jon-whit, line 32: This doesn't appear any different than what we had previously. The last command we run in the pipe is `tee bench_output.txt`, and so the `$?` will evaluate the return code of the `tee` command. For example, try these in a bash shell

```bash
$ false | true
$ echo $? // returns 0

$ true | false
$ echo $? // returns 1
```
I think what we want is either
* Split the commands and let the individual command failures dictate the return code
* Use `set -o pipefail` before running this command, which will return a non-zero code if any of the commands in the pipe return a non-zero code (not just the last one).

miparnisari: Excellent point, i didn't really test this command when I submitted the PR 🤦‍♀️ 

I tried it locally: introduced a unit test bug and ran

```
set -o pipefail
	make unit-test | tee debug.txt
	if [ $? -ne 0 ]; then
		echo "unit test failed"
	fi
```

this prints `"unit test failed"` 👍

jon-whit (approved):

Issue #619, opened 2023-03-09
Simplify setup code for `commands.ListObjectsQuery.Execute`


miparnisari:
The setup code to run a ListObjects call that uses the reverse expansion implementation is more complicated than it should be, IMO.

https://github.com/openfga/openfga/blob/c9d53e9b49ef69d6a7fa373aadbe934db264038f/pkg/server/server.go#L118-L143

`q` always needs an instance of `connectObjCmd`. So rather than passing it as a field I think it should be built within `q.Execute` itself. Otherwise, if we forget to set the field, we will be inadvertently running the concurrent checks implementation, which is never intended. E.g. i discovered by pure chance that i forgot to setup `connectObjCmd` in here: https://github.com/openfga/openfga/blob/2be676faa85c3061cef651e6b2fc348cb14feebf/pkg/server/test/list_objects.go#L162-L174

It didn't fail because of this check

https://github.com/openfga/openfga/blob/a383e03ad6aa85ff1d939e743ea5baa762e0f253/pkg/server/commands/list_objects.go#L100

which is... weird, in my opinion. The `q.Execute` code should decide which implementation to run based on the schema version, not on whether that object has been set.

PR #620, opened 2023-03-09
fix: list objects test setup

miparnisari:
- Fix test setup (https://github.com/openfga/openfga/issues/619)
- Remove unused code path
- Add coverage for this code block https://github.com/openfga/openfga/blob/3a032d1346a6065f5a1cf8c9970fc3450977d91a/pkg/server/commands/connected_objects.go#L542-L544

### DIFF
--- a/pkg/server/commands/connected_objects.go
+++ b/pkg/server/commands/connected_objects.go
@@ -528,7 +528,7 @@ func (c *ConnectedObjectsCommand) reverseExpandDirect(
 		tk := t.GetKey()
 
 		foundObject := tk.GetObject()
-		foundObjectType, foundObjectID := tuple.SplitObject(foundObject)
+		foundObjectType, _ := tuple.SplitObject(foundObject)
 
 		if _, ok := foundObjectsMap.LoadOrStore(foundObject, struct{}{}); ok {
 			// todo(jon-whit): we could optimize this by avoiding reading this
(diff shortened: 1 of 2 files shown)


pkg/server/commands/connected_objects.go
```diff
@@ -554,10 +554,6 @@ func (c *ConnectedObjectsCommand) reverseExpandDirect(
 			},
 		}
 
-		if tuple.IsTypedWildcard(foundObject) {
-			targetUserRef = &UserRefTypedWildcard{Type: foundObjectType}
-		}
```
miparnisari, line 559: this is never hit by our tests. Plus i think it can never happen, since you can't have `user:*` as the `object` field of a tuple 🤔

pkg/server/commands/connected_objects.go
```diff
@@ -554,10 +554,6 @@ func (c *ConnectedObjectsCommand) reverseExpandDirect(
...
-
 		if tk.GetRelation() != "" {
```
miparnisari, line 557: will this line ever be false?? 🤔

pkg/server/commands/connected_objects.go
miparnisari: do you mean we should do something like this?

```go
if tuple.IsTypedWildcard(foundObject) {
			panic("did not expect type:* on the object field of a tuple")
		}
```

hmmmmmm

> should these be a panic/assert (0)?

Worry that later on when we relax the model restriction and we forgot to update this

what do you mean by relax model restrictions?

adriantam: I meant that if/when we allow wildcard in TTU later on for model schema 1.1.

miparnisari: So for example you say what if in the future this test case  doesn't throw an error?

https://github.com/openfga/openfga/blob/eb8e348ced49a4bb514400ed720b7acabef52e9a/internal/validation/validation_test.go#L426-L468

if/when we change this, i think the code i deleted still won't be run ever

adriantam: I see.

jon-whit (approved):

PR #622, opened 2023-03-09
chore: wireup model schema_version usage metrics

jon-whit:
## Description
Introduce a new metric `model_schema_version_counter` which tracks a count of the various `schema_version` values that we encounter when resolving an authorization model, and wire it up by making use of the `TypesystemResolverFunc` at the top-level in the server.

## Review Checklist

### DIFF
--- a/pkg/server/commands/write.go
+++ b/pkg/server/commands/write.go
@@ -22,13 +22,19 @@ const (
 type WriteCommand struct {
 	logger    logger.Logger
 	datastore storage.OpenFGADatastore
+	typesys   *typesystem.TypeSystem
 }
 
 // NewWriteCommand creates a WriteCommand with specified storage.TupleBackend to use for storage.
-func NewWriteCommand(datastore storage.OpenFGADatastore, logger logger.Logger) *WriteCommand {
+func NewWriteCommand(
+	datastore storage.OpenFGADatastore,
+	logger logger.Logger,
+	typesys *typesystem.TypeSystem,
+) *WriteCommand {
 	return &WriteCommand{
-		logger:    logger,
 		datastore: datastore,
+		logger:    logger,
+		typesys:   typesys,
 	}
 }
 
@@ -50,8 +56,6 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 	ctx, span := tracer.Start(ctx, "validateWriteRequest")
 	defer span.End()
 
-	store := req.GetStoreId()
-	modelID := req.GetAuthorizationModelId()
 	deletes := req.GetDeletes().GetTupleKeys()
 	writes := req.GetWrites().GetTupleKeys()
 
@@ -61,22 +65,15 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 
 	if len(writes) > 0 {
 
-		authModel, err := c.datastore.ReadAuthorizationModel(ctx, store, modelID)
-		if err != nil {
-			return err
-		}
-
-		typesys := typesystem.New(authModel)
-
 		for _, tk := range writes {
-			err := validation.ValidateTuple(typesys, tk)
+			err := validation.ValidateTuple(c.typesys, tk)
 			if err != nil {
 				return serverErrors.ValidationError(err)
 			}
 
 			objectType, _ := tupleUtils.SplitObject(tk.GetObject())
 
-			relation, err := typesys.GetRelation(objectType, tk.GetRelation())
+			relation, err := c.typesys.GetRelation(objectType, tk.GetRelation())
 			if err != nil {
 				if errors.Is(err, typesystem.ErrObjectTypeUndefined) {
 					return serverErrors.TypeNotFound(objectType)
--- a/pkg/server/commands/write_test.go
+++ b/pkg/server/commands/write_test.go
@@ -11,6 +11,7 @@ import (
 	mockstorage "github.com/openfga/openfga/pkg/storage/mocks"
 	"github.com/openfga/openfga/pkg/testutils"
 	"github.com/openfga/openfga/pkg/tuple"
+	"github.com/openfga/openfga/pkg/typesystem"
 	"github.com/stretchr/testify/require"
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 )
@@ -41,7 +42,7 @@ func TestValidateNoDuplicatesAndCorrectSize(t *testing.T) {
 		}
 	}
 
-	cmd := NewWriteCommand(mockDatastore, logger)
+	cmd := NewWriteCommand(mockDatastore, logger, nil)
 
 	tests := []test{
 		{
@@ -142,11 +143,7 @@ func TestValidateWriteRequest(t *testing.T) {
 			maxTuplesInWriteOp := 10
 			mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)
 			mockDatastore.EXPECT().MaxTuplesPerWrite().AnyTimes().Return(maxTuplesInWriteOp)
-			cmd := NewWriteCommand(mockDatastore, logger)
-
-			if len(test.writes) > 0 {
-				mockDatastore.EXPECT().ReadAuthorizationModel(gomock.Any(), gomock.Any(), gomock.Any()).Return(&openfgapb.AuthorizationModel{}, nil)
-			}
+			cmd := NewWriteCommand(mockDatastore, logger, typesystem.New(&openfgapb.AuthorizationModel{}))
 
 			ctx := context.Background()
 			req := &openfgapb.WriteRequest{
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	grpc_ctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
-	"github.com/oklog/ulid/v2"
 	"github.com/openfga/openfga/internal/gateway"
 	"github.com/openfga/openfga/internal/graph"
 	"github.com/openfga/openfga/internal/validation"
@@ -49,7 +48,8 @@ type Server struct {
 	transport gateway.Transport
 	config    *Config
 
-	checkResolver graph.CheckResolver
+	checkResolver      graph.CheckResolver
+	typesystemResolver typesystem.TypesystemResolverFunc
 }
 
 type Dependencies struct {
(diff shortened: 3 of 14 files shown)


pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,99 @@
...
+		model, err := reader.ReadAuthorizationModel(ctx, storeID, modelID)
+		if err != nil {
+			if errors.Is(err, storage.ErrNotFound) {
+				return nil, ErrModelNotFound
+			}
+
+			return nil, fmt.Errorf("failed to ReadAuthorizationModel: %w", err)
+		}
+
+		modelSchemaVersionCounter.WithLabelValues(model.GetSchemaVersion()).Add(1)
```
jon-whit, line 95: Here's where we increment the counter based on the `schema_version` label. So we get counters like:

`model_schema_version_counter{"schema_version": "1.0"}`
`model_schema_version_counter{"schema_version": "1.1"}`

Note that we won't increment the counter if we return a memoized value (see `MemoizedTypesystemResolverFunc` above). We only increment the counter if we have to resolve a model (e.g. by fetching the latest one OR resolving a specific one for the first time).

adriantam: We may want to increment the counter every time it is being resolved regardless whether  MemoizedTypesystemResolverFunc is used or not.  Otherwise, we may miss any customers that had specified auth model id but continue to use model with schema 1.0

jon-whit: Yep, good point. Fixed in the latest commits.

pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+// TypesystemResolverFunc is a function that implementations can implement to provide lookup and
+// resolution of a Typesystem.
+type TypesystemResolverFunc func(ctx context.Context, storeID, modelID string) (*TypeSystem, error)
+
+// MemoizedTypesystemResolverFunc memoizes the provided TypesystemResolverFunc. It returns
+// the value from the first invocation of the underlying resolver function.
+//
+// The memoized resolver function is safe for concurrent use.
+func MemoizedTypesystemResolverFunc(fn TypesystemResolverFunc) TypesystemResolverFunc {
+	var cache sync.Map
```
adriantam, line 34: This may cause memory to grow overtime as it is not bounded nor has any timeout.  We may want to reconsider that will cause problem in a long running system

pkg/server/commands/write.go
```diff
@@ -47,11 +53,9 @@ func (c *WriteCommand) Execute(ctx context.Context, req *openfgapb.WriteRequest)
 }
 
 func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.WriteRequest) error {
-	ctx, span := tracer.Start(ctx, "validateWriteRequest")
```
miparnisari, line 50: nitpick: this will create a span that is not the child of the calling function

pkg/server/commands/write_test.go
```diff
@@ -41,7 +42,7 @@ func TestValidateNoDuplicatesAndCorrectSize(t *testing.T) {
 		}
 	}
 
-	cmd := NewWriteCommand(mockDatastore, logger)
+	cmd := NewWriteCommand(mockDatastore, logger, nil)
```
miparnisari, line 45: if for some reason `cmd.validateNoDuplicatesAndCorrectSize` starts using `typesystem` this whole set of tests will panic. I think it makes sense to change the signature of `validateNoDuplicatesAndCorrectSize` to

```golang
func validateNoDuplicatesAndCorrectSize(deletes []*openfgapb.TupleKey, writes []*openfgapb.TupleKey, maxTuplesPerwrite int) error {
```

pkg/server/server.go
```diff
@@ -136,7 +130,7 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgapb.ListObjectsRequ
 	return q.Execute(ctx, &openfgapb.ListObjectsRequest{
 		StoreId:              storeID,
 		ContextualTuples:     req.GetContextualTuples(),
-		AuthorizationModelId: modelID,
+		AuthorizationModelId: typesys.GetAuthorizationModelID(), // resolved model id
```
miparnisari, line 133: the `q` object now has a reference to the `typesys` so why do we need to pass a field of `typesys` to `Execute()`?

pkg/server/server.go
```diff
@@ -122,6 +115,7 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgapb.ListObjectsRequ
 		ListObjectsMaxResults: s.config.ListObjectsMaxResults,
 		ResolveNodeLimit:      s.config.ResolveNodeLimit,
 		CheckResolver:         s.checkResolver,
+		Typesystem:            typesys,
```
miparnisari, line 118: this is breaking the pattern we currently have where the `command` object has fields that are not request-specific. `typesys` is request specific

pkg/server/server_test.go
```diff
@@ -149,6 +149,9 @@ func TestCheckDoesNotThrowBecauseDirectTupleWasFound(t *testing.T) {
 			ResolveNodeLimit: 25,
 		},
 		checkResolver: graph.NewLocalChecker(storage.NewContextWrapper(mockDatastore), 100),
+		typesystemResolver: typesystem.MemoizedTypesystemResolverFunc(
+			typesystem.NewTypesystemResolver(mockDatastore),
+		),
 	}
```
miparnisari, line 155: please change this so that it uses `server.New( ... )` and we don't have to manually specify these fields

pkg/server/server_test.go
```diff
@@ -255,11 +261,14 @@ func TestResolveAuthorizationModel(t *testing.T) {
 			datastore: mockDatastore,
 			transport: transport,
 			logger:    logger,
+			typesystemResolver: typesystem.MemoizedTypesystemResolverFunc(
+				typesystem.NewTypesystemResolver(mockDatastore),
+			),
 		}
```
miparnisari, line 267: same here

pkg/server/server_test.go
```diff
@@ -220,6 +223,9 @@ func TestShortestPathToSolutionWins(t *testing.T) {
 			ResolveNodeLimit: 25,
 		},
 		checkResolver: graph.NewLocalChecker(storage.NewContextWrapper(mockDatastore), 100),
+		typesystemResolver: typesystem.MemoizedTypesystemResolverFunc(
+			typesystem.NewTypesystemResolver(mockDatastore),
+		),
 	}
```
miparnisari, line 229: same here

pkg/typesystem/resolver.go
miparnisari, line 34: why do we need another cache of models? we already have one here https://github.com/openfga/openfga/blob/038c5c1f8c2dca1df38a32e04143fb162d476d83/pkg/storage/caching/caching.go#L22-L27

we can do the `modelSchemaVersionCounter++` in there

miparnisari (changes requested):

miparnisari:
This PR is too big, and it's doing two things: introducing the `typeResolver` object, and introducing the new metric. Can we split into two?

pkg/server/server.go
jon-whit: We don't, but since `Execute` accepts the `*openfgapb.ListObjectsRequest` this is the accurate way of passing the resolved authorization model id.

I didn't feel it was pertinent to refactor all of this in the scope of this PR.

pkg/typesystem/resolver.go
jon-whit: The cached store caches the models, but it doesn't cache the resolution of the model and the typesystem constructed from it.

Instead of having to call `ReadAuthorizationModel` and `FindLatestAuthorizationModelID` everywhere we need it, we simply memoize the resolution of the typesystem once for each `storeID` and `modelID` we see. The memoization bypasses the cached storage layer altogether in favor of returning a prior constructed typesystem.

pkg/server/server.go
jon-whit: I'm not sure I understand. The `ListObjectsQuery` itself is request-bound. The query is constructed per-request, and it serves the inbound `ListObjects` RPC that is being invoked. Consequently, the Typesystem is very applicable to the incoming request. If we only constructed a single `ListObjectsQuery` up-front and then called `Execute` per invocation, then I could understand the point, but that's not the case here.

I was under the impression this was the direction you preferred our code-base trended towards, so this comment is surprising to me? We've been passing Typesystem through context, and you have been arguing in favor of passing Typesystem explicitly to the `commands`.

pkg/server/commands/write_test.go
jon-whit: That's a good point. That method doesn't need to be an instance method (instance of `WriteCommand`).

jon-whit: Refactored 👍

pkg/server/server_test.go
jon-whit: See https://github.com/openfga/openfga/pull/623

pkg/server/server_test.go
jon-whit: See https://github.com/openfga/openfga/pull/623

Will rebase after that is merged.

pkg/server/server.go
miparnisari: Sorry i didn't explain myself super well. If you look at the codebase today, at our commands, e.g.

https://github.com/openfga/openfga/blob/5185d4e041e8b8ad4692a34e912a8bc891e465b1/pkg/server/commands/read_authzmodel.go#L14-L17

or 

https://github.com/openfga/openfga/blob/5185d4e041e8b8ad4692a34e912a8bc891e465b1/pkg/server/commands/read_changes.go#L15-L20

the fields in the `Query` struct (commands) are global objects that do not vary per request. Today, if we want to, we could build one of these objects at the server constructor and re-use it per request.

But in this PR, you are constructing `typesys` _from the request parameters_ and adding it to this struct. This would make us not be able to do what I mentioned above

pkg/typesystem/resolver.go
miparnisari: Can't we change the existing cache to cache typesys instead of models?

miparnisari:
>  another PR introducing the metric in the TypesystemResolverFunc (which would be like a 1 line change).

you can make this change right now in the existing model cache and it would likely get approved immediately. With this PR, you are asking us to approve a big change before we can get that metric in

pkg/typesystem/resolver.go
jon-whit: If we change our storage layer then yes. But I see that separate concerns.

PR #624, opened 2023-03-10
chore: refactor typesystem

miparnisari:
This should reduce memory usage.

### DIFF
--- a/pkg/typesystem/typesystem.go
+++ b/pkg/typesystem/typesystem.go
@@ -126,39 +126,37 @@ func Difference(base *openfgapb.Userset, sub *openfgapb.Userset) *openfgapb.User
 }
 
 type TypeSystem struct {
-	model           *openfgapb.AuthorizationModel
+	modelID         string
 	schemaVersion   string
 	typeDefinitions map[string]*openfgapb.TypeDefinition
-}
 
-type ValidatedTypeSystem struct {
-	*TypeSystem
+	// used only for validation
+	duplicateTypes []string
 }
 
 // New creates a *TypeSystem from an *openfgapb.AuthorizationModel.
 func New(model *openfgapb.AuthorizationModel) *TypeSystem {
+	duplicateTypes := make([]string, 0)
 	tds := map[string]*openfgapb.TypeDefinition{}
 	for _, td := range model.GetTypeDefinitions() {
+		if _, duplicateKey := tds[td.GetType()]; duplicateKey {
+			duplicateTypes = append(duplicateTypes, td.GetType())
+		}
 		tds[td.GetType()] = td
 	}
 
 	return &TypeSystem{
-		model:           model,
+		modelID:         model.GetId(),
+		duplicateTypes:  duplicateTypes,
 		schemaVersion:   model.GetSchemaVersion(),
 		typeDefinitions: tds,
 	}
 }
 
-// GetAuthorizationModel returns the underlying AuthorizationModel this TypeSystem was
-// constructed from.
-func (t *TypeSystem) GetAuthorizationModel() *openfgapb.AuthorizationModel {
-	return t.model
-}
-
 // GetAuthorizationModelID returns the id for the authorization model this
 // TypeSystem was constructed for.
 func (t *TypeSystem) GetAuthorizationModelID() string {
-	return t.model.GetId()
+	return t.modelID
 }
 
 func (t *TypeSystem) GetSchemaVersion() string {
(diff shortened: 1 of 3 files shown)


pkg/typesystem/typesystem.go
```diff
@@ -126,39 +126,37 @@ func Difference(base *openfgapb.Userset, sub *openfgapb.Userset) *openfgapb.User
...
+	// used only for validation
+	duplicateTypes []string
 }
 
 // New creates a *TypeSystem from an *openfgapb.AuthorizationModel.
 func New(model *openfgapb.AuthorizationModel) *TypeSystem {
+	duplicateTypes := make([]string, 0)
 	tds := map[string]*openfgapb.TypeDefinition{}
 	for _, td := range model.GetTypeDefinitions() {
+		if _, duplicateKey := tds[td.GetType()]; duplicateKey {
```
adriantam, line 142: Instead of carrying the duplicateTypes around and waiting for validation to fail, will it make more sense to just return error?

miparnisari: that would change the signature of `New()` and i don't want to do that yet. But yeah this file deserves another refactor. The only reason we need this is for `Validate()` and this is only called in `WriteAuthorizationModel`

adriantam: Ok.

adriantam (dismissed):

pkg/typesystem/typesystem.go
```diff
@@ -126,39 +126,37 @@ func Difference(base *openfgapb.Userset, sub *openfgapb.Userset) *openfgapb.User
 }
 
 type TypeSystem struct {
-	model           *openfgapb.AuthorizationModel
+	modelID         string
 	schemaVersion   string
 	typeDefinitions map[string]*openfgapb.TypeDefinition
-}
 
-type ValidatedTypeSystem struct {
```
jon-whit, line 134: The reason for this was so that you can have the Go type system help you pass around a struct that you know has already been validated. Without it it's not clear if a `TypeSystem` struct has already been validated or not.

For example,

```
func blah(ts *ValidatedTypeSystem) { // ts has already been validated }
```
vs
```
func blah(ts *TypeSystem) { // has ts been validated?? }
```

miparnisari: this struct is never used though

miparnisari: > any time we've called typesystem.New(...).Validate()

this is the only usage

https://github.com/openfga/openfga/blob/89554e8b6a0fb5eca0101bdd9fed3c7d944ebed9/pkg/server/commands/write_authzmodel.go#L49

and it's ignoring the result 😆

jon-whit: Well that's not good, lol. If you don't pass model validation then `WriteAuthorizationModel` RPC should fail.

miparnisari: Nono, it's ignoring the `validatedTypeSystem` struct, but it's using the `err`

jon-whit: Oh, I see. Yeah, so ideally we propagate the `ValidatedTypeSystem` anywhere we assumed we are working with a pre-validated typesystem. That was the goal, but it seems when we added that we didn't properly wire it up??

miparnisari: you mean that all the other APIs would go from accepting a `typesystem` to accepting a `ValidTypeSystem`? but what does that mean, it was valid at the time it was written for sure, but what if the validation rules changed after that?

jon-whit: What if we had the constructor do the validation so that we know any `*TypeSystem` returned has already been validated. Basically, I want to just ensure that

```go
typesys, err := typesystem.New(model) // should always return a validated typesystem
```
and we could have a helper if needs be like we do elsewhere
```go
typesys := typesystem.MustNew(model) // panics if the typesystem failed validation during construction.
```
then we could just get rid of the `ValidatedTypeSystem` in favor of that.

miparnisari: We can send a future PR with that, but I need to think about your idea a bit more. Off the top of my head

> What if we had the constructor do the validation

- Check for example should not do validations, to keep it fast.
- In general i think `New`s should be kept as thin as possible as not throw any errors, otherwise it becomes `NewAndValidate`.

pkg/typesystem/typesystem.go
miparnisari: Introduced a new constructor method that does validations

jon-whit (approved):

Issue #615, opened 2023-03-08
Spans in Check are not properly nested


miparnisari:
With this model:

```python
model
  schema 1.1
type user
type team
  relations
    define member: [user,team#member]
type repo
  relations
    define admin: [user,team#member] or repo_admin from owner
    define maintainer: [user,team#member] or admin
    define owner: [organization]
    define reader: [user,team#member] or triager or repo_reader from owner
    define triager: [user,team#member] or writer
    define writer: [user,team#member] or maintainer or repo_writer from owner
type organization
  relations
    define member: [user] or owner
    define owner: [user]
    define repo_admin: [user,organization#member]
    define repo_reader: [user,organization#member]
    define repo_writer: [user,organization#member]

```

and these tuples

```json
{
    "writes": {
        "tuple_keys": [
            {
                "user": "user:erik",
                "relation": "member",
                "object": "organization:openfga"
            },{
                "user": "organization:openfga#member",
                "relation": "repo_admin",
                "object": "organization:openfga"
            },{
                "user": "team:openfga/core#member",
                "relation": "admin",
                "object": "repo:openfga/openfga"
            },{
                "user": "organization:openfga",
                "relation": "owner",
                "object": "repo:openfga/openfga"
            }
        ]
    }
}
```

this call to Check

```json
{
    "tuple_key": {
                "user": "user:erik",
                "relation": "reader",
                "object": "repo:openfga/openfga"
            }
}
```

prints the following trace

<img width="1922" alt="image" src="https://user-images.githubusercontent.com/5374887/223630036-dde87cb4-7847-40d1-817d-787436888d76.png">

but the `ResolveCheck` span should be at least as long as all of its children spans.

My hypothesis is that when we spin up goroutines we are losing the parent span. We would need to do 

```go
ctx, span := tracer.Start(ctx, "childSpan", trace.WithLinks(trace.Link{
      SpanContext: parentSpanCtx,
}))
```

PR #627, opened 2023-03-11
fix: tracing in Check API

miparnisari:
Fixes https://github.com/openfga/openfga/issues/615

Changes:

- The inner-most `span` that found the Check resolution will now have the attribute `allowed:true`. This helps visually identify which path led to the resolution (similar to the UI in the playground)
- I renamed a few spans because they made more sense visually
- I added a few attributes to some spans to help debugging

overall

compare to playground

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -13,6 +13,7 @@ import (
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 var tracer = otel.Tracer("internal/graph/check")
@@ -326,6 +327,8 @@ func (c *LocalChecker) ResolveCheck(
 	ctx, span := tracer.Start(ctx, "ResolveCheck")
 	defer span.End()
 
+	span.SetAttributes(attribute.String("tuple_key", req.GetTupleKey().String()))
+
 	if req.GetResolutionMetadata().Depth == 0 {
 		return nil, ErrResolutionDepthExceeded
 	}
@@ -375,6 +378,9 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		relation := tk.GetRelation()
 
 		fn1 := func(ctx context.Context) (*openfgapb.CheckResponse, error) {
+			ctx, span := tracer.Start(ctx, "checkDirectUserTuple", trace.WithAttributes(attribute.String("tuple_key", tk.String())))
+			defer span.End()
+
 			t, err := c.ds.ReadUserTuple(ctx, storeID, tk)
 			if err != nil {
 				if errors.Is(err, storage.ErrNotFound) {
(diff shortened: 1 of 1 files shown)


internal/graph/check.go
```diff
@@ -626,9 +634,6 @@ func (c *LocalChecker) checkRewrite(
 	case *openfgapb.Userset_This:
 		return c.checkDirect(ctx, req)
 	case *openfgapb.Userset_ComputedUserset:
-		ctx, span := tracer.Start(ctx, "checkComputedUserset")
-		defer span.End()
```
miparnisari, line 630: this span is not needed. `c.dispatch` will create the right span when it calls `ResolveCheck`

internal/graph/check.go
```diff
@@ -323,7 +323,7 @@ func (c *LocalChecker) ResolveCheck(
 	ctx context.Context,
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
-	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	ctx, span := tracer.Start(ctx, "checkRewrite")
 	defer span.End()
```
miparnisari, line 327: I realize the span name is not consistent but the function name, but this is the right name from a customer perspective. `ResolveCheck` calls `checkRewrite`. I think we need to do some name changes after this PR. `ResolveCheck` is called both by `ListObjects` and `Check` AND within this file, which is confusing

jon-whit: Yep, that's purposeful. `ResolveCheck` recurses on itself (or dispatches/delegates to itself) because the composition is meant to be able to delegate the resolution to any implementation of a `CheckResolver`, which could be a remote `CheckResolver` (dispatch sub-problems to a peer or to a cached local resolver etc..).

internal/graph/check.go
```diff
@@ -344,6 +344,8 @@ func (c *LocalChecker) ResolveCheck(
 		return nil, fmt.Errorf("relation '%s' undefined for object type '%s'", relation, objectType)
 	}
 
+	span.SetAttributes(attribute.String("rewrite", tuple.ToObjectRelationString(object, relation)))
```
jon-whit, line 347: This attribute is misleading to me, because it's not a rewritten value. The only scenario where we evaluate a rewritten value is with the `computed_userset`. Maybe `relationship` or `object_relation` is a more aptly named attribute here.

miparnisari: Okay then `checkRewrite` should be called `checkRelationship`?

internal/graph/check.go
```diff
@@ -375,6 +377,7 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		relation := tk.GetRelation()
 
 		fn1 := func(ctx context.Context) (*openfgapb.CheckResponse, error) {
+			span.SetAttributes(attribute.String("user_tuple", tk.String()))
```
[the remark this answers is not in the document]
miparnisari: This attribute now belongs to the span named `checkDirectUserTuple`. Sso I think `user_tuple_ is a good attribute name, let me know if you think otherwise

jon-whit (approved):

PR #628, opened 2023-03-11
build: use chainguard images in Dockerfile

craigpastro:
## Description

👋 Build using the chainguard go image (which uses glibc rather than musl) and then run the resulting executable in the distroless chainguard image. This should be more secure than a generic alpine image.

I remember a long time ago wondering why it wouldn't work in a distroless image because we had no dependency on CGO, but you must explicitly have `CGO_ENABLED=0` in the build step for it to work. 

## Review Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,16 @@
-FROM golang:1.20-alpine AS builder
+FROM cgr.dev/chainguard/go:1.20 AS builder
 
 WORKDIR /app
 
 COPY . .
-RUN go build -o ./openfga ./cmd/openfga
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,16 +1,16 @@
-FROM golang:1.20-alpine AS builder
+FROM cgr.dev/chainguard/go:1.20 AS builder
```
jon-whit, line 1: We will also want to use the `cgr.dev/chainguard/static` image in the `Dockerfile.goreleaser` file. That's the docker image that actually gets released.

jon-whit (approved):

PR #641, opened 2023-03-16
chore: add span attributes after computing relationship ingresses

miparnisari:
This should help with debugging:

image

### DIFF
--- a/pkg/server/commands/connected_objects.go
+++ b/pkg/server/commands/connected_objects.go
@@ -155,6 +155,9 @@ func (c *ConnectedObjectsCommand) streamedConnectedObjects(
 	g := graph.BuildConnectedObjectGraph(c.Typesystem)
 
 	// find the possible incoming edges (ingresses) between the target user reference and the source (object, relation) reference
+	span.SetAttributes(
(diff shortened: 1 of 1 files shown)


pkg/server/commands/connected_objects.go
```diff
@@ -163,7 +166,8 @@ func (c *ConnectedObjectsCommand) streamedConnectedObjects(
 	subg, subgctx := errgroup.WithContext(ctx)
 	subg.SetLimit(maximumConcurrentChecks)
 
-	for _, ingress := range ingresses {
+	for i, ingress := range ingresses {
+		span.SetAttributes(attribute.String(fmt.Sprintf("_ingress %d", i), fmt.Sprintf("ingress %s, type %s, tupleset %s", ingress.Ingress.String(), ingress.Type.String(), ingress.TuplesetRelation.String())))
```
jon-whit, line 170: Let's add a `String()` method on the `RelationshipIngress` type itself and then just do

```go
span.SetAttributes(attribute.String(fmt.Sprintf("_ingress %d", i), ingress.String()))
```

jon-whit (approved):

Issue #621, opened 2023-03-09
Improve test by removing hardcoded port


miparnisari:
https://github.com/openfga/openfga/blob/8e3988306fa6da4ce8314e9615a55b35e59f95a9/cmd/run/run_test.go#L539-L540

It should choose a random port that is available

Issue #639, opened 2023-03-16
HTTP upstream connection to gRPC uses `grpc.addr` incorrectly


kdubb:
The HTTP server immediately make an upstream connection to the gRPC server with the `grpc.addr`; most of the time this will be the default value of `0.0.0.0:<port>`.

Strictly speaking, this isn't a valid address to use for connections. It's only valid as a source address; although in practice it is basically equivalent to `localhost` or `127.0.0.1`. This should be translated to `127.0.0.1` before attempting the connection.

PR #649, opened 2023-03-17
fix: do not hard-code ports in test

adriantam:
## Description

Avoid hard coding ports in test

## References

Close https://github.com/openfga/openfga/issues/621

## Review Checklist

### DIFF
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -293,8 +293,8 @@ func DefaultConfig() *Config {
 	}
 }
 
-// TCPRandomPort returns a random TCP Port along with a function that releases the port.
-// It is the responsibility of the caller to release the port. Otherwise, the port will be orphaned.
+// TCPRandomPort tries to find a random TCP Port. If it can't find one, it panics. Else, it returns the port and a function that releases the port.
+// It is the responsibility of the caller to call the release function.
 func TCPRandomPort() (int, func()) {
 	l, err := net.Listen("tcp", "")
 	if err != nil {
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -316,8 +316,8 @@ func MustDefaultConfigWithRandomPorts() *Config {
 	defer l.Close()
 	grpcPort := l.Addr().(*net.TCPAddr).Port
 
-	config.GRPC.Addr = fmt.Sprintf("0.0.0.0:%d", grpcPort)
-	config.HTTP.Addr = fmt.Sprintf("0.0.0.0:%d", httpPort)
+	config.GRPC.Addr = fmt.Sprintf("localhost:%d", grpcPort)
+	config.HTTP.Addr = fmt.Sprintf("localhost:%d", httpPort)
 
 	return config
 }
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -293,6 +293,18 @@ func DefaultConfig() *Config {
 	}
 }
 
+// TCPRandomPort returns a random TCP Port along with a function that releases the port.
+// It is the responsibility of the caller to release the port. Otherwise, the port will be orphaned.
+func TCPRandomPort() (int, func()) {
+	l, err := net.Listen("tcp", "")
+	if err != nil {
+		panic(err)
+	}
+	return l.Addr().(*net.TCPAddr).Port, func() {
+		l.Close()
+	}
+}
+
 // MustDefaultConfigWithRandomPorts returns the DefaultConfig, but with random ports for the grpc and http addresses.
 // This function may panic if somehow a random port cannot be chosen.
 func MustDefaultConfigWithRandomPorts() *Config {
@@ -302,22 +314,14 @@ func MustDefaultConfigWithRandomPorts() *Config {
 	config.Playground.Enabled = false
 	config.Metrics.Enabled = false
 
-	l, err := net.Listen("tcp", "")
-	if err != nil {
-		panic(err)
-	}
-	defer l.Close()
-	httpPort := l.Addr().(*net.TCPAddr).Port
+	httpPort, httpPortReleaser := TCPRandomPort()
 
-	l, err = net.Listen("tcp", "")
-	if err != nil {
-		panic(err)
-	}
-	defer l.Close()
-	grpcPort := l.Addr().(*net.TCPAddr).Port
+	grpcPort, grpcPortReleaser := TCPRandomPort()
+	httpPortReleaser()
+	grpcPortReleaser()
 
-	config.GRPC.Addr = fmt.Sprintf("localhost:%d", grpcPort)
-	config.HTTP.Addr = fmt.Sprintf("localhost:%d", httpPort)
+	config.GRPC.Addr = fmt.Sprintf("0.0.0.0:%d", grpcPort)
+	config.HTTP.Addr = fmt.Sprintf("0.0.0.0:%d", httpPort)
 
 	return config
 }
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -315,10 +315,11 @@ func MustDefaultConfigWithRandomPorts() *Config {
 	config.Metrics.Enabled = false
 
 	httpPort, httpPortReleaser := TCPRandomPort()
-
-	grpcPort, grpcPortReleaser := TCPRandomPort()
 	defer func() {
 		httpPortReleaser()
+	}()
+	grpcPort, grpcPortReleaser := TCPRandomPort()
+	defer func() {
 		grpcPortReleaser()
 	}()
 
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -317,8 +317,10 @@ func MustDefaultConfigWithRandomPorts() *Config {
 	httpPort, httpPortReleaser := TCPRandomPort()
 
 	grpcPort, grpcPortReleaser := TCPRandomPort()
-	httpPortReleaser()
-	grpcPortReleaser()
+	defer func() {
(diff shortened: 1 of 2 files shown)


cmd/run/run.go
```diff
@@ -316,8 +316,8 @@ func MustDefaultConfigWithRandomPorts() *Config {
 	defer l.Close()
 	grpcPort := l.Addr().(*net.TCPAddr).Port
 
-	config.GRPC.Addr = fmt.Sprintf("0.0.0.0:%d", grpcPort)
-	config.HTTP.Addr = fmt.Sprintf("0.0.0.0:%d", httpPort)
+	config.GRPC.Addr = fmt.Sprintf("localhost:%d", grpcPort)
```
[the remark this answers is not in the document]
adriantam: TLS requires localhost (not 0.0.0.0 ) because it doesn’t contain any IP SANs .  Thus, we will either have to specifically change TLS case to use localhost or hard code those cases.

cmd/run/run_test.go
```diff
@@ -563,10 +564,20 @@ func TestHTTPServerWithCORS(t *testing.T) {
 	}
 }
 
+// helper function to obtain default config and extra tcp port
+func obtainDefaultConfigWithExtraPort(t *testing.T) (*Config, int) {
```
jon-whit, line 568: Mind just adding a testutil to the `pkg/testutils` package that can be used to obtain a random port and then using it directly instead of a very specific helper method for this? For example,

```go
// package pkg/testutils/testutils.go

func RandomTCPPort() (int, error) {
    	l, err := net.Listen("tcp", "")
         if err != nil {
             return -1, err
         }
	defer l.Close()
	
        return l.Addr().(*net.TCPAddr).Port, nil
}
```

```go
// package cmd/run/run_test.go

func TestBuildServerWithOIDCAuthentication(t *testing.T) {
    cfg := MustDefaultConfigWithRandomPorts()

    oidcPort, err := testutils.RandomTCPPort()
    require.NoError(t, err)

    localOIDCServerURL := fmt.Sprintf("http://localhost:%d", oidcPort)
```

adriantam: The example above will not work because we have released the port when we exit MustDefaultConfigWithRandomPorts() and testutils.RandomTCPPort(). Thus, it is entirely possible that we use the same port for oidcPort and one of either HTTP/GRPC port.

cmd/run/run.go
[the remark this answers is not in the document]
adriantam: We may also want to consider https://github.com/openfga/openfga/issues/639 and update the default to localhost as well (or address this in a separate PR)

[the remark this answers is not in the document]
jon-whit: We're talking about unrelated things here. In the scope of this PR we're not changing 0.0.0.0 to `localhost` to address TLS specific issues. If we want to do that, as part of work to address #639 for example, then let's save that for a separate PR. The purpose of this work is to avoid hard-coded ports. So let's keep it focused to that.

cmd/run/run_test.go
jon-whit: `defer l.Close()` releases the port. The code you have currently is functionally the same as the code I proposed above.

Your concern is valid, but it's also already present in https://github.com/openfga/openfga/blob/main/cmd/run/run.go#L305-L320

It's possible we re-use a returned port in other code, but in practice we haven't experienced it (AFAIK).

adriantam: The code that is proposed will make it worse.  Previously, https://github.com/openfga/openfga/blob/main/cmd/run/run.go#L305-L320 will obtain two random ports before releasing.  Therefore, these two ports are guarantee not to collide and you will need to use have another caller trying to obtain the ports immediately for collision.  However, if we separate out via RandomTCPPort, you will release either the extra random port OR release both ports for default config immediately before obtaining a new port.  Thus, the chance of collision is much higher.

adriantam: Discussed offline, we can have RandomTCPPort but we need to return a function that can be called to release the port.  Otherwise, the ports may be re-used

cmd/run/run.go
```diff
@@ -302,19 +314,11 @@ func MustDefaultConfigWithRandomPorts() *Config {
...
+	httpPort, httpPortReleaser := TCPRandomPort()
 
-	l, err = net.Listen("tcp", "")
-	if err != nil {
-		panic(err)
-	}
-	defer l.Close()
-	grpcPort := l.Addr().(*net.TCPAddr).Port
+	grpcPort, grpcPortReleaser := TCPRandomPort()
+	httpPortReleaser()
```
jon-whit, line 320: Should we defer these so that we know there isn't a race on the port being acquired before the return of this function?

```go
defer func() {
    httpPortReleaser()
    grpcPortReleaser()
}()
```

adriantam: updated

jon-whit (approved):

cmd/run/run.go
```diff
@@ -293,6 +293,18 @@ func DefaultConfig() *Config {
 	}
 }
 
+// TCPRandomPort returns a random TCP Port along with a function that releases the port.
+// It is the responsibility of the caller to release the port. Otherwise, the port will be orphaned.
```
miparnisari, line 297:
```suggestion
// TCPRandomPort tries to find a random TCP Port. If it can't find one, it panics. Else, it returns the port and a function that releases the port.
// It is the responsibility of the caller to call the release function.
```

cmd/run/run_test.go
```diff
@@ -573,6 +575,9 @@ func TestBuildServerWithOIDCAuthentication(t *testing.T) {
 		Issuer:   localOIDCServerURL,
 	}
 
+	// release the oidc server port after obtaining cfg ports
+	oidcServerPortReleaser()
```
miparnisari, line 579: i don't understand why this is here. at this point in the code, the test hasn't run. should we run this code as a last step?

cmd/run/run.go
```diff
@@ -302,19 +314,13 @@ func MustDefaultConfigWithRandomPorts() *Config {
...
-	httpPort := l.Addr().(*net.TCPAddr).Port
+	httpPort, httpPortReleaser := TCPRandomPort()
 
-	l, err = net.Listen("tcp", "")
-	if err != nil {
-		panic(err)
-	}
-	defer l.Close()
-	grpcPort := l.Addr().(*net.TCPAddr).Port
+	grpcPort, grpcPortReleaser := TCPRandomPort()
```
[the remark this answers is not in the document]
miparnisari: Okay but then you have a leaked open port, no?

cmd/run/run_test.go
adriantam: We need to release the port before we call RunServer.  Otherwise, we try to use a port that is in use (i.e., net.Listen in line 299 of cmd/run/run.go and cause RunServer to fail.

cmd/run/run.go
```diff
@@ -302,19 +314,14 @@ func MustDefaultConfigWithRandomPorts() *Config {
...
-	l, err = net.Listen("tcp", "")
-	if err != nil {
-		panic(err)
-	}
-	defer l.Close()
-	grpcPort := l.Addr().(*net.TCPAddr).Port
+	httpPort, httpPortReleaser := TCPRandomPort()
+	defer func() {
+		httpPortReleaser()
+	}()
```
miparnisari, line 320:
```suggestion
	defer httpPortReleaser()
```

cmd/run/run.go
```diff
@@ -302,19 +314,14 @@ func MustDefaultConfigWithRandomPorts() *Config {
...
+	grpcPort, grpcPortReleaser := TCPRandomPort()
+	defer func() {
+		grpcPortReleaser()
+	}()
```
miparnisari, line 324:
```suggestion
	defer grpcPortReleaser()
```

cmd/run/run_test.go
miparnisari: Something seems off to me.. the OpenFGA server and the mock OIDC server should run in different ports, no?

cmd/run/run.go
```diff
@@ -293,6 +293,18 @@ func DefaultConfig() *Config {
 	}
 }
 
+// TCPRandomPort tries to find a random TCP Port. If it can't find one, it panics. Else, it returns the port and a function that releases the port.
+// It is the responsibility of the caller to call the release function.
+func TCPRandomPort() (int, func()) {
```
miparnisari, line 298: Sorry for being all over the place 😆 i was trying to understand how it works, and now I do

I think this function should call `Listen` to pick the port and then `Close` it immediately. So that when `http.ListenAndServe` runs, the port is available

cmd/run/run_test.go
adriantam: The issue is that server will fail due to OIDC port already in use.

cmd/run/run.go
miparnisari: The part i don't understand is this: `net.Listen("tcp", "")` already starts listening on the port. So how come `http.ListenAndServer(port)`doesn't give a "address already in use"?

miparnisari: Oh wait this is why we call the `defer release()` function...

jon-whit (dismissed):

cmd/run/run_test.go
miparnisari, line 579:
```suggestion
	oidcServerPortReleaser()
```

miparnisari (approved):

miparnisari (approved):

PR #618, opened 2023-03-08
WIP: refactor: swap typesystem in the context for explicit propagation

jon-whit:
## References
Closes #616 

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #650, opened 2023-03-18
refactor: #619

### LINKED ISSUE #619: Simplify setup code for `commands.ListObjectsQuery.Execute`

miparnisari:
Closes https://github.com/openfga/openfga/issues/619

Reviewers: please click on this

image


pkg/server/commands/connected_objects.go
```diff
@@ -20,6 +20,7 @@ import (
 
 type ConnectedObjectsRequest struct {
 	StoreID          string
+	Typesystem       *typesystem.TypeSystem
```
jon-whit, line 23: I'd like to talk about this as a maintainer team. It's odd for the request to take the Typesystem, because the request is the client provided request body, and the Typesystem is not part of those semantics. The Typesystem is resolved as a result of the client provided request metadata (e.g. the `authorization_model_id` field). The lifecycle of a command/query struct begins at the beginning of a request and ends (and is garbage collected) after the response is served.

In our usages of the commands and queries pattern, for example `ListObjectsQuery` and `WriteCommand`, the command/query itself is scoped to the request. In every usage we do something like:

```go
query := NewExpandQuery(...)

resp, err := query.Execute(ctx, &openfgapb.ExpandRequest{...})
if err != nil {...}

return resp, nil
```

In other words, we always construct an `ExpandQuery` object 1:1 for every request that we received. It is request bound. Consequently, instead of a wiring and propagating the Typesystem in every `...Request` object (like we've had to do here), I think it'd make more sense to keep the Typesystem bound to the query or command struct itself.

jon-whit:
This is directly related to https://github.com/openfga/openfga/pull/618.

pkg/server/commands/connected_objects.go
miparnisari: > It's odd for the request to take the Typesystem, because the request is the client provided request body, and the Typesystem is not part of those semantics. The Typesystem is resolved as a result of the client provided request metadata (e.g. the authorization_model_id field).

Typesystem is not part of the request body, but it's derived from the request body. Poteito, potato...

> the command/query itself is scoped to the request. (...) we always construct an ExpandQuery object 1:1 for every request that we received. It is request bound. 

True. But the `query`/`command` don't receive request-specific objects. E.g.

https://github.com/openfga/openfga/blob/288f58e682be6e04d373f1db15444ace2b305527/pkg/server/commands/expand.go#L18-L22

or 

https://github.com/openfga/openfga/blob/3a1142eecdba268d92331df6644ea3d08584917e/pkg/server/commands/read_changes.go#L15-L20

> instead of a wiring and propagating the Typesystem in every ...Request object (like we've had to do here), I think it'd make more sense to keep the Typesystem bound to the query or command struct itself.

Why? Typesystem is request-scoped

jon-whit: > Typesystem is not part of the request body, but it's derived from the request body. Poteito, potato...

I think this is the crux of the argument though. If we construct a `ReadChangesQuery`, `ListObjectsQuery`, `ExpandQuery`, `WriteCommand`, or something similar, then the only purpose these structs serve is to encapsulate a set of methods that share bindings to reference objects or constants that are required/dependend upon (e.g. the logger or datastore). When you say
>  But the query/command don't receive request-specific objects. E.g.

This is kind of my point though. The Typesystem isn't a request-specific object, it's an object constructed from a request-specific field (e.g. the `authorization_model_id` field).

----

In our case the commands and queries only have private helpers and a public `Execute` function, and the `Execute` function has developed a pattern of taking the API level request body (`*openfgapb.ExpandRequest`, `*openfgapb.WriteRequest`, `*openfgapb.ReadChangesRequest`, etc..). In a sense, you could argue that the commands and query structs themselves are wasteful, because they are request bound (we construct and destroy one for every request). We could get away with a closure closing the datastore and logger just as elegantly.

The way I look at it is this - the command/query struct itself provides a means to mock or inject dependencies into the `Execute` function. The `Execute` functions _needs a_ storage interface, logger, encoder, etc.. The `Execute` function also _needs a_ Typesystem to do it's work. It doesn't care how the Typesystem is constructed/created upstream, but it does need it to do its work. For this reason, instead of constructing the Typesystem in every command/query that needs a Typesystem, why not just provide that as another injected depednecy of the struct itself? That way upstream code in `pkg/server/server.go` can provide a common utility method for resolving a `TypeSystem` reference from a request-level field and then inject it into the command/query for usage/evaluation?

Anyhow, I suppose whatever direction we go we just take the time to make the codebase consistent. It'd be nice to get some feedback from @adriantam on this as well.

adriantam: IMO, either way is reasonable.  I understand why @miparnisari wants to approach this way though - it is "cleaner" in a sense that we only construct the Typesystem when we need it rather than carrying it all over the place. One aspect I think that ties to a more per request approach is that it has tight coupling with the model ID - which is part of the request itself.

miparnisari: > The Typesystem isn't a request-specific object, it's an object constructed from a request-specific field

In practice, what is the difference?

> instead of constructing the Typesystem in every command/query that needs a Typesystem, why not just provide that as another injected dependency of the struct itself? 


> In our case the commands and queries only have private helpers and a public Execute function, and the Execute function has developed a pattern of taking the API level request body

My point in this PR is to make all commands / queries consistent with the current convention we started out with back in 2021, which was having global dependencies in the `struct`, request-level specifics in the `execute`.

> In a sense, you could argue that the commands and query structs themselves are wasteful, because they are request bound (we construct and destroy one for every request). We could get away with a closure closing the datastore and logger just as elegantly.

That's a good debate that we can have, but if we agree to change it, we need to do this in a separate PR and it needs to change all commands and queries. I think: either they all follow this pattern or none of them do. Having both patterns is confusing. I recently read a [book](https://milkov.tech/assets/psd.pdf) that has a paragraph on this very topic: 

> Don’t change existing conventions. Resist the urge to “improve” on existing conventions. Having a “better idea” is not a sufficient excuse to introduce inconsistencies. Your new idea may indeed be better, but the value of
consistency over inconsistency is almost always greater than the value of one approach over another. Before introducing inconsistent behavior, ask yourself two questions. First, do you have significant new information justifying your approach that wasn’t available when the old convention was established? Second, is the new approach so much better that it is worth taking the time to update all of the old uses? If your organization agrees that the answers to both questions are “yes,” then go ahead and make the upgrade; when you are done, there 
no sign of the old convention. However, you still run the risk that other developers will not know about the new convention, so they may reintroduce the old approach in the future. Overall, reconsidering established conventions is rarely a good use of developer time.

adriantam (approved):

pkg/server/commands/connected_objects.go
rhamzeh: > In a sense, you could argue that the commands and query structs themselves are wasteful, because they are request bound (we construct and destroy one for every request). We could get away with a closure closing the datastore and logger just as elegantly.

I feel that's a good idea to work towards. But this PR does not nullify that. If I'm understanding @jon-whit correctly, we want to make the commands not request-bound as an optimization. I feel this PR is the first step of doing that by moving the typesystem out of the command.

Now, there may be other changes we'd like to do in the future, but I don't see this as a blocker, but more as the road that.

rhamzeh (approved):

PR #651, opened 2023-03-21
Feature[credentials]: Support explicit datastore username & password configuration points

kdubb:
This adds explicit support for separately providing a username and/or password to the datastore as separate explicit configuration values.

All the expected sources are supported: config.yaml, flags and environment variables. Both postgrsql & mysql are implemented.

## Description

When using secret managers like Vault or even Kubernetes secrets only allowing the specifying a full connection URI only becomes problematic.Usually _only_ the credentials are stored in these services and they may need to be rotated in an automated fashion.

Allowing the username and/or password to be provided separately eases integration of OpenFGA with these credential managements systems. For example, passing the base URI to a Helm chart value and then providing the username and password via K8s secrets mapped as environment variables.

## Review Checklist

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -113,6 +113,16 @@
                     ],
                     "x-env-variable": "OPENFGA_DATASTORE_URI"
                 },
+                "username": {
+                    "description": "The username for the connection to the datastore. If provided it will overwrite any username provided in the connection string.",
+                    "type": "string",
+                    "x-env-variable": "OPENFGA_DATASTORE_USERNAME"
+                },
+                "password": {
+                    "description": "The password for the connection to the datastore. If provided it will overwrite any password provided in the connection string.",
+                    "type": "string",
+                    "x-env-variable": "OPENFGA_DATASTORE_PASSWORD"
+                },
                 "maxCacheSize": {
                     "description": "The maximum number of cache keys that the storage cache can store before evicting old keys.",
                     "type": "integer",
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -119,7 +119,7 @@
                     "x-env-variable": "OPENFGA_DATASTORE_USERNAME"
                 },
                 "password": {
-                    "description": "The connection password to use to connect to the datastore (overwrites any password provided in the connection uri).",
+                    "description": "The connection password to connect to the datastore (overwrites any password provided in the connection uri).",
                     "type": "string",
                     "x-env-variable": "OPENFGA_DATASTORE_PASSWORD"
                 },
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -114,7 +114,7 @@
                     "x-env-variable": "OPENFGA_DATASTORE_URI"
                 },
                 "username": {
-                    "description": "The connection username to use to connect to the datastore (overwrites any username provided in the connection uri).",
+                    "description": "The connection username to connect to the datastore (overwrites any username provided in the connection uri).",
                     "type": "string",
                     "x-env-variable": "OPENFGA_DATASTORE_USERNAME"
                 },
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -114,12 +114,12 @@
                     "x-env-variable": "OPENFGA_DATASTORE_URI"
                 },
                 "username": {
-                    "description": "The username for the connection to the datastore. If provided it will overwrite any username provided in the connection string.",
+                    "description": "The connection username to use to connect to the datastore (overwrites any username provided in the connection uri).",
                     "type": "string",
                     "x-env-variable": "OPENFGA_DATASTORE_USERNAME"
                 },
                 "password": {
-                    "description": "The password for the connection to the datastore. If provided it will overwrite any password provided in the connection string.",
+                    "description": "The connection password to use to connect to the datastore (overwrites any password provided in the connection uri).",
                     "type": "string",
                     "x-env-variable": "OPENFGA_DATASTORE_PASSWORD"
                 },
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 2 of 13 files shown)


jon-whit:
> @jon-whit I took a shot at this to jumpstart its implementation, if it's going to being implemented at some point.

Hey @kdubb , if we're going to separate out the individual fields of the datastore/database connection uri, then we should separate it in totality. It is odd to have just the `username` and `password` fields be deconstructed from the whole uri. My counter proposal here would be to introduce two ways of configuring the connection uri.

1. A single connection uri (like we have today) - `postgres://username:password@hostname:port/db-name?opts...`
2. Individual configurations per field. For example:

    `OPENFGA_DATASTORE_USERNAME` (required)
    `OPENFGA_DATASTORE_PASSWORD` (required)
    `OPENFGA_DATASTORE_HOSTNAME` (required)
    `OPENFGA_DATASTORE_PORT` (required)
    `OPENFGA_DATASTORE_DB_NAME` (required)
    `OPENFGA_DATASTORE_OPTS` (optional)

     We'd concatenate these individual fields to construct the whole uri that we use to connect to the database. If the operator of OpenFGA provides the `--datastore-connection-uri` flag (option 1), then that flag will take precedence over providing the individual fields. If the developer specifies any `--datastore-engine` other than `memory`, and does not provide the `--datastore-connection-uri`, then each individual required field (referring to option 2 above) must be validated as non-empty.

kdubb:
That's fine but I wouldn't say it's required. Credentials are "secrets" that should be allowed separately. For example, JDBC's Datasource uses a connection string but requires the implementation to allow separately setting the username and password even when using it.

The URI connection string is nice, easy to use, and convenient. It's just the credentials that need to be opened up to alternate configuration sources due to the nature of them.

jon-whit:
> That's fine but I wouldn't say it's required. Credentials are "secrets" that should be allowed separately. For example, JDBC's Datasource uses a connection string but requires the implementation to allow separately setting the username and password even when using it.
> 
> The URI connection string is nice, easy to use, and convenient. It's just the credentials that need to be opened up to alternate configuration sources due to the nature of them.

My concern is that the behavior adds more nuanced variability to the configuration structure and semantics. If you can provide a connection uri but with seperated `username` and `password` fields, then now the single `--datastore-connection-uri` has nuances to it. It becomes a uri string with an optional `username` and `password` field etc.. `postgres://[username]:[password]@hostname:port/db-name[?opts]`. That just introduces more variations of deconstructing and validating the uri, and more semantics to understand of how to configure the database connection settings in OpenFGA, which impacts developer experience a bit. 

For example, what if the username field is provided in the connection uri but the password field isn't or vice-versa etc..? Does `postgres://myuser:@myhost:5432/mydb` with `OPENFGA_DATASTRE_PASSWORD=secret` imply that the `myuser:secret` credentials are to be used or does it imply that the `myuser:` credentials are to be used, or is the `myuser:` invalid because the `OPENFGA_DATASTORE_PASSWORD` env variable was provided, etc.. These are the kinds of variations that cause ambiguities that concern me.

It may make sense to require that `username:password` both be included if you are providing them in the connection uri directly. Otherwise we will construct them from `OPENFGA_DATASTORE_USERNAME` and `OPENFGA_DATASTORE_PASSWORD`. That would help minimize the ambiguities that I've mentioned above. What do you think?

.config-schema.json
```diff
@@ -113,6 +113,16 @@
                     ],
                     "x-env-variable": "OPENFGA_DATASTORE_URI"
                 },
+                "username": {
+                    "description": "The username for the connection to the datastore. If provided it will overwrite any username provided in the connection string.",
```
jon-whit, line 117: Please use the same `description` here as the one you provided in the `flags.go` file.

.config-schema.json
```diff
@@ -113,6 +113,16 @@
...
+                    "type": "string",
+                    "x-env-variable": "OPENFGA_DATASTORE_USERNAME"
+                },
+                "password": {
+                    "description": "The password for the connection to the datastore. If provided it will overwrite any password provided in the connection string.",
```
jon-whit, line 122: Please use the same `description` here as the one you provided in the `flags.go` file.

jon-whit:
@kdubb do you mind if I take this PR over and make some minor changes? We want to merge the changes in, but I wanted to make a few minor changes (nothing major functionallity, just nit picks here and there).

kdubb:
Sure! Glad I could start the process!

kdubb:
@jon-whit Any update on this? I can help out more if needed.

jon-whit:
@kdubb alright, it seems I've fixed the issue by treating config values more idiomatically. Flags, env variables, and config values should treat default/zero values as defaults if unprovided.

Now the configs override correctly.

kdubb:
Wahoo!

.config-schema.json
```diff
@@ -113,6 +113,16 @@
                     ],
                     "x-env-variable": "OPENFGA_DATASTORE_URI"
                 },
+                "username": {
+                    "description": "The connection username to use to connect to the datastore (overwrites any username provided in the connection uri).",
```
adriantam, line 117:
```suggestion
                    "description": "The connection username to connect to the datastore (overwrites any username provided in the connection uri).",
```

.config-schema.json
```diff
@@ -113,6 +113,16 @@
...
+                    "type": "string",
+                    "x-env-variable": "OPENFGA_DATASTORE_USERNAME"
+                },
+                "password": {
+                    "description": "The connection password to use to connect to the datastore (overwrites any password provided in the connection uri).",
```
adriantam, line 122:
```suggestion
                    "description": "The connection password to connect to the datastore (overwrites any password provided in the connection uri).",
```

pkg/storage/postgres/postgres.go
```diff
@@ -37,6 +38,38 @@ type Postgres struct {
...
+		}
+
+		// Username always set to some value
+		username := ""
+		if cfg.Username != "" {
+			username = cfg.Username
+		} else if parsed.User != nil {
+			username = parsed.User.Username()
+		}
+
```
adriantam, line 56: Do we need to do anything if username is still ""?

adriantam (dismissed):

pkg/storage/postgres/postgres.go
jon-whit: Username will still be "" in the following cases, for example:

`postgres://:password@localhost:5432/postgres`
`postgres://localhost:5432/postgres`

```shell
export OPENFGA_DATASTORE_USERNAME=""

./openfga run --datastore-engine mysql --datastore-uri 'root:secret@tcp(localhost:3306)/openfga?parseTime=true'
```

In these cases we constrct the credentials correctly, and we attempt to connect with a username that is an empty string, which may or may not work depending on the user credentials. In any case though, the behavior seems to override correctly and accurately based on what is provided.

adriantam: Thanks for the explanation.  In this case, the comment `// Username always set to some value` is a bit misleading

jon-whit: Gotcha. Those comments are a bit unecessary. The code is pretty self-explanatory, so I just removed them.

adriantam (approved):

PR #659, opened 2023-03-27
fix: broken unit test

miparnisari:
Fixes broken build: https://github.com/openfga/openfga/actions/runs/4536794174/jobs/7993912100?pr=658#step:7:24

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/openfga/openfga/pkg/tuple"
 	"github.com/openfga/openfga/pkg/typesystem"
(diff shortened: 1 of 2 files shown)


internal/graph/graph_test.go
```diff
@@ -39,19 +40,42 @@ var (
...
+			expected: "ingress type:\"document\" relation:\"viewer\", type direct",
+			ingress: RelationshipIngress{
+				Type:    DirectIngress,
+				Ingress: typesystem.DirectRelationReference("document", "viewer"),
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := strings.ReplaceAll(tc.ingress.String(), " ", "")
+			require.Equal(t, strings.ReplaceAll(tc.expected, " ", ""), actual)
```
adriantam, line 76: Feels a bit weird to to the trimming it the UT instead of in graph.go?  Is the failure pointing to inconsistency in the output message? Should we use strings.Trim(r.Ingress.String()) in graph.go?  (Though I am not sure if the same approach will fix for r.TuplesetRelation.String() as the whitespace is in the middle :( ).  We may need to do `strings.Join(strings.Fields(r.TuplesetRelation.String()), " ")`

miparnisari: So the problem is in 

```go
func (x *RelationReference) String() string {
	return protoimpl.X.MessageStringOf(x)
}
```

For some reason that I don't know, this sometimes outputs one whitespace, and sometimes two. 

In the usage of this method, i don't really care which one it does, because it's used only in the UI of Jaeger.

miparnisari: On second thought, i think what you said is a good idea. Updated the PR!

internal/graph/graph.go
```diff
@@ -80,10 +81,13 @@ type RelationshipIngress struct {
 }
 
 func (r RelationshipIngress) String() string {
+	val := ""
```
adriantam, line 84: another way to do it is
```
val := "ingress " + r.Ingress.String() + ", type " + r.Type.String()
if r.TuplesetRelation != nil {
    val += ", tupleset " + r.TuplesetRelation.String()
}

adriantam: Either way should be fine

adriantam (approved):

Issue #616, opened 2023-03-08
Remove `typesystem.ContextWithTypesystem`


miparnisari:
There's been two instances at least where we have forgotten to write `ctx = typesystem.ContextWithTypesystem(ctx, typesys)` and this caused an issue:

- https://github.com/openfga/openfga/pull/523
- https://github.com/openfga/openfga/pull/614

We should get rid of this. If a function needs the typesystem, it should either build it internally or accept it as function parameter.

jon-whit:
I'm not opposed to refactoring this in favor of explicit parameters in various places, but please don't remove it. I just ask that care is taken to not adjust the function signature of the `ResolveCheck`. We can adjust elsewhere, but leave `ResolveCheck` function signature unmodified please.

https://github.com/openfga/openfga/blob/main/internal/graph/check.go#L322

This implementation specifically makes use of the context scoped TypeSystem on purpose because this implementation is  based upon an RPC interface for remote dispatching. `ResolveCheck` gets executed recursively, and so it is purposeful to not read the model and construct the typesystem each invocation. The assumption of the `ResolveCheck` function was that the higher order function `server.Check` resolves the model and constructs the typesystem from it, and the `ResolveCheck` would be implemented as an RPC interface on each replica of OpenFGA and they'd delegate to one other. Consequently, you wouldn't be passing the `TypeSystem` around between function invocation within `ResolveCheck`. This orchestration is very purposeful.

My recommendation , and maybe a good alternative here, would be to memoize the typesystem construction in `ResolveCheck` and so if the request gets dispatched (locally or remotely) we don't reconstruct the typesystem and we just return the reference to it that was constructed from a prior invocation. Essentially replacing context passed parameters with a closure that is memoized.

PR #636, opened 2023-03-16
feat(tests): automatically test contextual tuples

adriantam:
## Description
Automatically add the contextual tuples counter in test.  This transforms the storage tuples into contextual tuples and run the same tests with purely contextual tuples

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #665, opened 2023-03-30
refactor: use combinedTupleReader in ListObjects and Check

miparnisari:
This is similar to https://github.com/openfga/openfga/issues/616, but it doesn't close it because it's for contextual tuples, not for the typesystem.

Instead of setting the contextual tuples as part of the `ctx` object (implicit), we pass those around explicitly via:

```go
combinedDatastore := storage.NewCombinedTupleReader(q.Datastore, req.GetContextualTuples().GetTupleKeys())
localChecker := graph.NewLocalChecker(combinedDatastore, maximumConcurrentChecks)
```


pkg/server/commands/list_objects.go
```diff
@@ -33,7 +33,6 @@ type ListObjectsQuery struct {
 	ListObjectsMaxResults uint32
 	ResolveNodeLimit      uint32
 	ConnectedObjects      func(ctx context.Context, req *ConnectedObjectsRequest, results chan<- string) error
-	CheckResolver         graph.CheckResolver
```
miparnisari, line 36: this is only needed if the implementation is v0, so i only set it when it's needed

pkg/server/test/list_objects.go
```diff
@@ -200,11 +194,9 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 				ListObjectsMaxResults: test.maxResults,
 				ConnectedObjects:      connectedObjCmd.StreamedConnectedObjects,
 				ResolveNodeLimit:      defaultResolveNodeLimit,
-				CheckResolver:         checker,
 			}
 			typesys := typesystem.New(model)
 			ctx = typesystem.ContextWithTypesystem(ctx, typesys)
-			ctx = storage.ContextWithContextualTuples(ctx, test.contextualTuples.GetTupleKeys())
```
miparnisari, line 207: this is my ultimate goal, get rid of this line. I should not have to remember to set the `ctx` when writing a test

pkg/server/commands/list_objects.go
```diff
@@ -301,13 +292,15 @@ func (q *ListObjectsQuery) performChecks(ctx context.Context, req listObjectsReq
 
 func (q *ListObjectsQuery) internalCheck(
 	ctx context.Context,
+	combinedDatastore storage.RelationshipTupleReader,
 	obj *openfgapb.Object,
 	req listObjectsRequest,
 	objectsFound *uint32,
 	resultsChan chan<- string,
 ) error {
+	checkResolver := graph.NewLocalChecker(combinedDatastore, maximumConcurrentChecks)
```
jon-whit, line 301: This should only be constructed once, not per `internalCheck` invocation. Constructing it each invocation is wasteful.

pkg/storage/storage.go
```diff
@@ -241,129 +230,129 @@ func filterTuples(tuples []*openfgapb.TupleKey, targetObject, targetRelation str
...
+
+	iter2, err := c.wrapped.ReadStartingWithUser(ctx, store, filter)
 	if err != nil {
 		return nil, err
 	}
 
 	return NewCombinedIterator(iter1, iter2), nil
 }
+
+func (c *combinedTupleReader) ListObjectsByType(ctx context.Context, store string, objectType string) (ObjectIterator, error) {
```
adriantam, line 342: mind adding description for exportable function (i.e., to satisfy godoc)?

adriantam (commented):
looks like we don't have any test covering several branches from ReadStartingWithUser

miparnisari:
> looks like we don't have any test covering several branches from ReadStartingWithUser

Right, our code was not using this method. I changed this PR so that now it's used. But I will wait until https://github.com/openfga/openfga/pull/636 is merged because I don't want to introduce regressions.

pkg/storage/storage.go
miparnisari: The documentation for all these methods is in the interface declaration: https://github.com/openfga/openfga/blob/c4a8d8f5ad36c02d13c0470cc68e3b5c204bac8d/pkg/storage/storage.go#L103

pkg/server/commands/connected_objects.go
```diff
@@ -503,17 +424,16 @@ func (c *ConnectedObjectsCommand) reverseExpandDirect(
 		userFilter = append(userFilter, val.ObjectRelation)
 	}
 
-	iter2, err := c.Datastore.ReadStartingWithUser(ctx, store, storage.ReadStartingWithUserFilter{
+	combinedTupleReader := storage.NewCombinedTupleReader(c.Datastore, req.contextualTuples)
```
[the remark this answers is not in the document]
jon-whit: > Which is two APIs: ListObjects, and Check

And `Expand`. The number of APIs is besides the point. The point is that we're trying to increase readability and explicitness, preferably without having to repeat ourselves in multiple critical code paths.

> I wouldn't consider one line of code (combinedDatastore := storage.NewCombinedTupleReader(q.Datastore, req.GetContextualTuples().GetTupleKeys())) in 4 places to be "all over the place". 3 of those places are for ListObjects, i could try in a future PR to have that in one place only.

`ctx = storage.ContextualTuplesWithContext(ctx, req.GetContextualTuples())` is one line of code as well.

We just as easily could have changed the code to inject the context at this point in the code and the storage layer could have constructed the combined iterator and abstracted that away. It would have had the same impact to LOC changed. Again, this is coming from a place of personal preference. Let's bring objectivity to the evaluation of these changes.

We should reduce repetition, increase explicitness, and avoid having to construct derivative datastores for context specific code, because failing to do so leads you equally as confused as a developer as failing to inject the contextual tuples into the Go context (like the code is today).

> I don't like option 2 because it shouldn't be the data layer's responsibility to even know that contextual tuples exist (IMO).

It's counter productive to have to remember to construct a derivative datastore with the contextual tuples in every code path that needs to iterate over contextual tuples + persisted tuples. Your code is already delegating the combined iterator responsibility to the datastore layer, but it requires the caller to set it up. It's the datastore that iterates over the contextual tuples and persistent tuples, and so that's already the datastore's responsibility. The only difference between this code and the old code is that you construct a derivative datastore instead of injecting values into the context. And my argument is that's only minorly different than passing it into the context (albeit it requires calling `NewCombinedTupleReader(..)` rather than passing it in the context).

This is why option 2 above, in my opinion, is a better approach. If we want readability and explicitness, then let's make it explicitly part of the storage interface and propagate it appropriately instead of requiring every caller to construct a derivative datastore to merge them.

> Then who's responsibility is it? I can't think of a more appropriate owner. The layer that talks to the DB shouldn't even know that contextual tuples exist

At the end of the day we don't want the iterator code in ListObjects, Check, etc.. to have to change to handle contextual tuples. We want to receive an iterator and handle the tuples in the iterator uniformly, regardless of if they were persistent tuples or contextual tuples. Your code achieves this, and the existing code (as it exists today) achieves this. This means that the datastore layer is responsible for merging the contextual tuples and the persistent tuples.

So the question is... What is the most idiomatic and DRY way of passing in the additional contextual tuples to the datastore layer so that it can merge those contextual tuples and return a single combined iterator? If we're trying to increase readability and explicitness, then passing the contextual tuples in explicitly is the best way to do that instead of passing them through context or constructing a derivative datastore implementation.

**Option 1**
```go
var contextualTuples *openfgapb.ContextualTuples

iter, err := datastore.Read(ctx, ..., contextualTuples)
_ = iter // use iterator (which combines all tuples)
```
is more explicit, less error prone, and more idiomatic than the two following examples:

**Option 2**
```go
var contextualTuples *openfgapb.ContextualTuples

combinedDatastore := storage.NewCombinedTupleReader(s.datastore, contextualTuples)

iter, err := combinedDatastore.Read(ctx, ...)
_ = iter // user iterator (which combines all tuples)
```

**Option 3**
```go
var contextualTuples *openfgapb.ContextualTuples
ctx = storage.ContextWithContextualTuples(ctx, contextualTuples)

iter, err := datastore.Read(ctx, ...)
_ = iter // user iterator (which combines all tuples)
```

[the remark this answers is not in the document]
rhamzeh: Reading this now - I will take some time to draft a better response once I've reviewed the PR, currently I'm basing my opinion on just this thread.

Looking at @jon-whit's 3 suggestions [here](https://github.com/openfga/openfga/pull/665#discussion_r1160276570):
- Option 1: I don't really like that the DB layer has to handle the contextual tuples, that seems to me to be very weird. It should be the application logic that is responsible for merging the data sources.

Also, for the sake of argument, imagine we want to allow `Read` to accept contextual tuples and return them. It would be much harder if the logic is in the datastore, because that means you have to update all the implementations for something that is not the datastore's responsibility.

- Option 2:
The naming is throwing me off. Is this a new DataStore when we say (`combinedDatastore`)? If yes, this seems weird. Is it implementing the DataStore interface in it's entirety? If yes, I'm not sure that is the best idea for just handling the contextual tuples.

If it is meant to be just a combined iterator, that appends the datastore iterator to the contextual tuples, this seems fine by me. Maybe we can call it something else? But other than the naming, this seems fine.

If this wraps around the datastore methods to then extend the DB iterator with contextual tuples, maybe we can call it `datastoreWrapper` or `extendedDatastore`? Though I guess `combinedDatastore` is OK if it means combining the contextual tuples with the datastore. Can we document that though?

Naming is hard, so I'm OK with this option regardless of name, provided it's not a whole new datastore that's being implemented.

- Option 3:
I'm not really sure ctx is the place for that. I may be off coming from node, but having to defer to ctx for a critical piece of the application interfaces to be passed around could cause it to have a bit less visibility (only in this case we need to look at ctx for data needed for application logic).

As in: imagine you are a new dev taking a look at the code, you might miss that ctx has more context here, and why, being explicitly passed around (even though more verbose) brings a bit more clarity. Note: I mention a new dev, but that would similarly apply to any of us 6 months down the line.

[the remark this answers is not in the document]
adriantam: My 2 cents - ideally, the datastore layer (i.e., mysql / postgresql / etc.) should only know how to read/write to their respective database. Any higher logic should be handled in a more abstract layer/higher level layer.  In this way, we don't need to repeat the handling for contextual tuple in multiple database.  Maybe a 3 layer approach (i.e., Command or API layer that deals with logic of check/expand/list -> Storage layer that deals with layer of contextual tuple vs. "real" tuple generically -> Datastore layer that deals with writing/reading to specific database technology such as Postgresql / MySQL / etc.) may be a good way to abstract different level of responsibilities.

Also, another thing that I would like to mention is that we should avoid stuffing too many things in the context (i.e., `ctx = storage.ContextWithContextualTuples(ctx, req.GetContextualTuples())`). Some member of the golang community - i.e., (How To Use Contexts in Go)[https://www.digitalocean.com/community/tutorials/how-to-use-contexts-in-go#using-data-within-a-context], (How to correctly use context.Context in Go 1.7)[https://medium.com/@cep21/how-to-correctly-use-context-context-in-go-1-7-8f2c0fafdf39] have suggested that "any data required for a function to run should be passed as parameters".  It will make the code more readable if these parameters are explicitly specified.

adriantam (approved):

PR #666, opened 2023-03-30
refactor: test assertions

miparnisari:
A few changes:

- use `require` whenever possible, because it's a lot less code to write for us
- all test names have no spaces in them (easier for when a test fails to locate it in the codebase)
- simplify assertions like 

```go
if actual != expected {
	t.Errorf("%s: Expected message to be %s, actual %s", tc.expected, tc.actual)
}
```

to 
```go
t.Run(tc.name, func(t *testing.T) {
	require.Equal(t, tc.expected, tc.actual)
})
```
- standardize all our uses of `cmp.diff`. The [correct usage](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) is

```go
got, want := Something()

if diff := cmp.Diff(want, got); diff != "" {
	t.Errorf("mismatch (-want +got):\n%s", diff)
}
```


pkg/logger/logger_test.go
```diff
@@ -47,26 +48,18 @@ func TestMessage(t *testing.T) {
...
-		if actualMessage.Level != tc.expectedLevel {
-			t.Errorf("%s: Expected message level to be %d, actual %d", tc.name, tc.expectedLevel, actualMessage.Level)
+		require.Equal(t, 1, logs.Len())
 
-		}
+		actualMessage := logs.All()[0]
+		require.Equal(t, testMessage, actualMessage.Message)
 
+		expectedZapFields := map[string]interface{}{}
+		require.True(t, reflect.DeepEqual(actualMessage.ContextMap(), expectedZapFields))
```
adriantam, line 57: We should be able to just use require.Equal.  Looking at the implementation (https://github.com/stretchr/testify/blob/c5fc9d6b6b21ea89be8480c0dc35e2977ab988f6/require/require.go#L158, https://github.com/stretchr/testify/blob/c5fc9d6b6b21ea89be8480c0dc35e2977ab988f6/assert/assertions.go#L58), it looks like under the hood it is doing deep equal.

pkg/logger/logger_test.go
```diff
@@ -88,37 +81,31 @@ func TestMessageWithEmptyId(t *testing.T) {
...
+				t.Errorf("%s: Unknown name", tc.name)
+			}
+			require.Equal(t, 1, logs.Len())
 
-		}
+			actualMessage := logs.All()[0]
+			require.Equal(t, testMessage, actualMessage.Message)
 
+			expectedZapFields := map[string]interface{}{}
+			require.True(t, reflect.DeepEqual(actualMessage.ContextMap(), expectedZapFields))
```
adriantam, line 106: similar to above, it appears that require.Equal should be sufficient.

pkg/logger/logger_test.go
```diff
@@ -131,13 +118,10 @@ func TestWithFields(t *testing.T) {
...
-	}
+	require.Equal(t, testMessage, actualMessage.Message)
+
 	expectedZapFields := map[string]interface{}{
 		"TestOption": "Message",
 	}
-	if !reflect.DeepEqual(actualMessage.ContextMap(), expectedZapFields) {
-		t.Errorf("Expected zap fields %v actual %v", expectedZapFields, actualMessage.ContextMap())
-	}
+	require.True(t, reflect.DeepEqual(actualMessage.ContextMap(), expectedZapFields))
```
adriantam, line 126: same as above

jon-whit (approved):

Issue #676, opened 2023-04-03
Postgresql check - investigate whether checkTTU required order by ulid


adriantam:
Based on performance test, we see that the checkTTU call (https://github.com/openfga/openfga/blob/2c5dfe2081803a47d63644d63f061fced6c3c09f/internal/graph/check.go#L553) will lead to the use of order by ulid (https://github.com/openfga/openfga/blob/2c5dfe2081803a47d63644d63f061fced6c3c09f/pkg/storage/postgres/postgres.go#L137) for Postgresql .  This probably is not necessary and slow down the common path of check.

PR #677, opened 2023-04-03
feat(performance): postgresql/mysql Read interface to be unordered

adriantam:
## Description

Previously, Read storage interface implementation for postgresql and mysql is ordered by ulid. We now change to unordered to improve performance.

## References
Close https://github.com/openfga/openfga/issues/676

## Review Checklist

### DIFF
--- a/cmd/util/util.go
+++ b/cmd/util/util.go
@@ -1,4 +1,4 @@
-// Package util provides sqlcommon utilities for spf13/cobra CLI utilities
+// Package util provides common utilities for spf13/cobra CLI utilities
 // that can be used for various commands within this project.
 package util
 
--- a/cmd/util/util.go
+++ b/cmd/util/util.go
@@ -1,4 +1,4 @@
-// Package util provides common utilities for spf13/cobra CLI utilities
+// Package util provides sqlcommon utilities for spf13/cobra CLI utilities
 // that can be used for various commands within this project.
 package util
 
--- a/pkg/server/errors/errors.go
+++ b/pkg/server/errors/errors.go
@@ -133,7 +133,7 @@ func HandleError(public string, err error) error {
 	return NewInternalError(public, err)
 }
 
-// HandleTupleValidateError provide common routines for handling tuples validation error
+// HandleTupleValidateError provide sqlcommon routines for handling tuples validation error
 func HandleTupleValidateError(err error) error {
 	switch t := err.(type) {
 	case *tuple.InvalidTupleError:
--- a/pkg/server/errors/errors.go
+++ b/pkg/server/errors/errors.go
@@ -133,7 +133,7 @@ func HandleError(public string, err error) error {
 	return NewInternalError(public, err)
 }
 
-// HandleTupleValidateError provide sqlcommon routines for handling tuples validation error
+// HandleTupleValidateError provide common routines for handling tuples validation error
 func HandleTupleValidateError(err error) error {
 	switch t := err.(type) {
 	case *tuple.InvalidTupleError:
--- a/pkg/storage/mysql/mysql.go
+++ b/pkg/storage/mysql/mysql.go
@@ -11,10 +11,9 @@ import (
 	sq "github.com/Masterminds/squirrel"
 	"github.com/cenkalti/backoff/v4"
 	_ "github.com/go-sql-driver/mysql"
-	"github.com/oklog/ulid/v2"
 	"github.com/openfga/openfga/pkg/logger"
 	"github.com/openfga/openfga/pkg/storage"
-	"github.com/openfga/openfga/pkg/storage/common"
+	"github.com/openfga/openfga/pkg/storage/sqlcommon"
 	tupleUtils "github.com/openfga/openfga/pkg/tuple"
 	"github.com/openfga/openfga/pkg/typesystem"
 	"github.com/pkg/errors"
@@ -37,7 +36,7 @@ type MySQL struct {
 
 var _ storage.OpenFGADatastore = (*MySQL)(nil)
 
-func New(uri string, cfg *common.Config) (*MySQL, error) {
+func New(uri string, cfg *sqlcommon.Config) (*MySQL, error) {
 	db, err := sql.Open("mysql", uri)
 	if err != nil {
 		return nil, fmt.Errorf("failed to initialize mysql connection: %w", err)
@@ -103,10 +102,10 @@ func (m *MySQL) ListObjectsByType(ctx context.Context, store string, objectType 
 		}).
 		QueryContext(ctx)
 	if err != nil {
-		return nil, common.HandleSQLError(err)
+		return nil, sqlcommon.HandleSQLError(err)
 	}
 
-	return common.NewSQLObjectIterator(rows), nil
+	return sqlcommon.NewSQLObjectIterator(rows), nil
 }
 
 func (m *MySQL) Read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey) (storage.TupleIterator, error) {
@@ -133,7 +132,7 @@ func (m *MySQL) ReadPage(ctx context.Context, store string, tupleKey *openfgapb.
 	return iter.ToArray(opts)
 }
 
-func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, opts storage.PaginationOptions) (*common.SQLTupleIterator, error) {
+func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, opts storage.PaginationOptions) (*sqlcommon.SQLTupleIterator, error) {
 	ctx, span := tracer.Start(ctx, "mysql.read")
 	defer span.End()
 
@@ -158,7 +157,7 @@ func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgapb.Tupl
 		sb = sb.Where(sq.Eq{"_user": tupleKey.GetUser()})
 	}
 	if opts.From != "" {
-		token, err := common.UnmarshallContToken(opts.From)
+		token, err := sqlcommon.UnmarshallContToken(opts.From)
 		if err != nil {
 			return nil, err
 		}
@@ -170,10 +169,10 @@ func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgapb.Tupl
 
 	rows, err := sb.QueryContext(ctx)
 	if err != nil {
-		return nil, common.HandleSQLError(err)
+		return nil, sqlcommon.HandleSQLError(err)
 	}
 
-	return common.NewSQLTupleIterator(rows), nil
+	return sqlcommon.NewSQLTupleIterator(rows), nil
 }
 
 func (m *MySQL) Write(ctx context.Context, store string, deletes storage.Deletes, writes storage.Writes) error {
@@ -185,82 +184,8 @@ func (m *MySQL) Write(ctx context.Context, store string, deletes storage.Deletes
 	}
 
 	now := time.Now().UTC()
-	txn, err := m.db.BeginTx(ctx, nil)
-	if err != nil {
-		return common.HandleSQLError(err)
-	}
-	defer func() {
-		_ = txn.Rollback()
-	}()
 
-	changelogBuilder := m.stbl.
-		Insert("changelog").
-		Columns("store", "object_type", "object_id", "relation", "_user", "operation", "ulid", "inserted_at")
-
-	deleteBuilder := m.stbl.Delete("tuple")
-
-	for _, tk := range deletes {
-		id := ulid.MustNew(ulid.Timestamp(now), ulid.DefaultEntropy()).String()
-		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
-
-		res, err := deleteBuilder.
-			Where(sq.Eq{
-				"store":       store,
-				"object_type": objectType,
-				"object_id":   objectID,
-				"relation":    tk.GetRelation(),
-				"_user":       tk.GetUser(),
-				"user_type":   tupleUtils.GetUserTypeFromUser(tk.GetUser()),
-			}).
-			RunWith(txn). // Part of a txn
-			ExecContext(ctx)
-		if err != nil {
-			return common.HandleSQLError(err)
-		}
-
-		rowsAffected, err := res.RowsAffected()
-		if err != nil {
-			return common.HandleSQLError(err)
-		}
-
-		if rowsAffected != 1 {
-			return storage.InvalidWriteInputError(tk, openfgapb.TupleOperation_TUPLE_OPERATION_DELETE)
-		}
-
-		changelogBuilder = changelogBuilder.Values(store, objectType, objectID, tk.GetRelation(), tk.GetUser(), openfgapb.TupleOperation_TUPLE_OPERATION_DELETE, id, sq.Expr("NOW()"))
-	}
-
-	insertBuilder := m.stbl.
-		Insert("tuple").
-		Columns("store", "object_type", "object_id", "relation", "_user", "user_type", "ulid", "inserted_at")
-
-	for _, tk := range writes {
-		id := ulid.MustNew(ulid.Timestamp(now), ulid.DefaultEntropy()).String()
-		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
-
-		_, err = insertBuilder.
-			Values(store, objectType, objectID, tk.GetRelation(), tk.GetUser(), tupleUtils.GetUserTypeFromUser(tk.GetUser()), id, sq.Expr("NOW()")).
-			RunWith(txn). // Part of a txn
-			ExecContext(ctx)
-		if err != nil {
-			return common.HandleSQLError(err, tk)
-		}
-
-		changelogBuilder = changelogBuilder.Values(store, objectType, objectID, tk.GetRelation(), tk.GetUser(), openfgapb.TupleOperation_TUPLE_OPERATION_WRITE, id, sq.Expr("NOW()"))
-	}
-
-	if len(writes) > 0 || len(deletes) > 0 {
-		_, err := changelogBuilder.RunWith(txn).ExecContext(ctx) // Part of a txn
-		if err != nil {
-			return common.HandleSQLError(err)
-		}
-	}
-
-	if err := txn.Commit(); err != nil {
-		return common.HandleSQLError(err)
-	}
-
-	return nil
+	return sqlcommon.Write(ctx, sqlcommon.NewDBInfo(m.db, m.stbl, sq.Expr("NOW()")), store, deletes, writes, now)
 }
 
 func (m *MySQL) ReadUserTuple(ctx context.Context, store string, tupleKey *openfgapb.TupleKey) (*openfgapb.Tuple, error) {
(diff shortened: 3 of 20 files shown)


pkg/storage/postgres/postgres.go
```diff
@@ -130,15 +130,23 @@ func (p *Postgres) ReadPage(ctx context.Context, store string, tupleKey *openfga
 	return iter.ToArray(opts)
 }
 
-func (p *Postgres) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, opts storage.PaginationOptions) (*common.SQLTupleIterator, error) {
+func (p *Postgres) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, requiresOrder bool, opts storage.PaginationOptions) (*common.SQLTupleIterator, error) {
```
rhamzeh, line 133: Instead of adding a new param, maybe we can have it in the pagination options?

Or (in case it's always the same), deduce it from whether there are pagination params?

pkg/storage/postgres/postgres.go
```diff
@@ -130,15 +130,23 @@ func (p *Postgres) ReadPage(ctx context.Context, store string, tupleKey *openfga
...
 	ctx, span := tracer.Start(ctx, "postgres.read")
 	defer span.End()
 
-	sb := p.stbl.
-		Select("store", "object_type", "object_id", "relation", "_user", "ulid", "inserted_at").
-		From("tuple").
-		Where(sq.Eq{"store": store}).
-		OrderBy("ulid")
+	var sb sq.SelectBuilder
+	if requiresOrder {
```
rhamzeh, line 138: Can't this be:

```go
sb = p.stbl.
	Select("store", "object_type", "object_id", "relation", "_user", "ulid", "inserted_at").
	From("tuple").
	Where(sq.Eq{"store": store})
if requiresOrder {
  sb = sb.OrderBy("ulid")
}
```

pkg/storage/mysql/mysql.go
```diff
@@ -129,7 +129,7 @@ func (m *MySQL) ReadPage(ctx context.Context, store string, tupleKey *openfgapb.
 	return iter.ToArray(opts)
 }
 
-func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, opts storage.PaginationOptions) (*common.SQLTupleIterator, error) {
+func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgapb.TupleKey, requiresOrder bool, opts storage.PaginationOptions) (*common.SQLTupleIterator, error) {
```
rhamzeh, line 132: Why isn't MySQL the new field?

pkg/storage/postgres/postgres.go
adriantam: we can do that

pkg/storage/mysql/mysql.go
adriantam: what do you mean?

adriantam: fixed

jon-whit (changes requested):
Could you please add some tests for this new behavior in the storage conformance test suite `pkg/storage/test`?

adriantam:
> Could you please add some tests for this new behavior in the storage conformance test suite `pkg/storage/test`?

Added

pkg/storage/postgres/postgres_test.go
```diff
@@ -46,3 +48,94 @@ func TestReadAuthorizationModelPostgresSpecificCases(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, typesystem.SchemaVersion1_0, model.SchemaVersion)
 }
+
+// TestReadEnsureNoOrder asserts that the read response is not ordered by ulid
+func TestReadEnsureNoOrder(t *testing.T) {
```
jon-whit, line 53: Instead of duplicating these across MySQL and Postgres, please add this to the storage conformance tests `pkg/storage/test/tuples.go`

adriantam: The setup call is a bit different - $1 vs. ?

adriantam: As discuss offline, we will use common facility that requires manipulation of time.

pkg/storage/storage.go
```diff
@@ -187,13 +190,29 @@ type ChangelogBackend interface {
 	ReadChanges(ctx context.Context, store, objectType string, paginationOptions PaginationOptions, horizonOffset time.Duration) ([]*openfgapb.TupleChange, []byte, error)
 }
 
+// CurrentTime has functionality of Now that allows mocking of current time
+type CurrentTime interface {
```
jon-whit, line 194: nit pick - `TimeProvider` I think is a better interface name here.

See https://go.dev/doc/effective_go#interface-names

pkg/storage/storage.go
```diff
@@ -187,13 +190,29 @@ type ChangelogBackend interface {
...
+
 type OpenFGADatastore interface {
 	TupleBackend
 	AuthorizationModelBackend
 	StoresBackend
 	AssertionsBackend
 	ChangelogBackend
 
+	// UpdateCurrentTime allows overriding time.Now functionality
+	UpdateCurrentTime(currentTime TimeProvider)
```
jon-whit, line 214:
```suggestion
        // SetTimeProvider allows overriding time.Now functionality to set timestmaps in writes.
	SetTimeProvider(provider TimeProvider)
```

pkg/storage/postgres/postgres.go
```diff
@@ -81,6 +82,7 @@ func New(uri string, cfg *common.Config) (*Postgres, error) {
 		logger:                 cfg.Logger,
 		maxTuplesPerWriteField: cfg.MaxTuplesPerWriteField,
 		maxTypesPerModelField:  cfg.MaxTypesPerModelField,
+		currentTime:            storage.DefaultTimeProvider{},
```
jon-whit, line 85:
```suggestion
		timeProvider:            storage.DefaultTimeProvider{},
```

pkg/storage/postgres/postgres.go
```diff
@@ -33,6 +33,7 @@ type Postgres struct {
 	logger                 logger.Logger
 	maxTuplesPerWriteField int
 	maxTypesPerModelField  int
+	currentTime            storage.TimeProvider
```
jon-whit, line 36:
```suggestion
	timeProvider            storage.TimeProvider
```

pkg/storage/mysql/mysql.go
```diff
@@ -33,6 +33,7 @@ type MySQL struct {
 	logger                 logger.Logger
 	maxTuplesPerWriteField int
 	maxTypesPerModelField  int
+	currentTime            storage.TimeProvider
```
jon-whit, line 36:
```suggestion
	timeProvider            storage.TimeProvider
```

pkg/storage/memory/memory.go
```diff
@@ -106,6 +107,7 @@ func New(maxTuplesPerWrite int, maxTypesPerAuthorizationModel int) *MemoryBacken
 		authorizationModels:           make(map[string]map[string]*AuthorizationModelEntry),
 		stores:                        make(map[string]*openfgapb.Store, 0),
 		assertions:                    make(map[string][]*openfgapb.Assertion, 0),
+		currentTime:                   storage.DefaultTimeProvider{},
```
jon-whit, line 110:
```suggestion
		timeProvider:                   storage.DefaultTimeProvider{},
```

pkg/storage/memory/memory.go
```diff
@@ -69,6 +69,7 @@ type MemoryBackend struct {
 	maxTuplesPerWrite             int
 	maxTypesPerAuthorizationModel int
 	mu                            sync.Mutex
+	currentTime                   storage.TimeProvider
```
jon-whit, line 72:
```suggestion
	timeProvider                   storage.TimeProvider
```

pkg/storage/test/tuples.go
```diff
@@ -474,6 +476,98 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		_, err = iter.Next()
 		require.ErrorIs(t, err, storage.ErrIteratorDone)
 	})
+
+}
+
+// TupleReadOrder tests the ordering of read and read page.
+// They are for specific datastore implementation and may not be appropriate
+// for all storage implementation (i.e,. memory)
+
```
jon-whit, line 485: Remove extra newline here.

pkg/storage/test/tuples.go
```diff
@@ -474,6 +476,98 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
...
+func TupleReadOrder(t *testing.T, datastore storage.OpenFGADatastore) {
```
jon-whit, line 486:
```suggestion
func TupleReadOrderTest(t *testing.T, datastore storage.OpenFGADatastore) {
```

pkg/storage/storage.go
```diff
@@ -187,13 +190,29 @@ type ChangelogBackend interface {
 	ReadChanges(ctx context.Context, store, objectType string, paginationOptions PaginationOptions, horizonOffset time.Duration) ([]*openfgapb.TupleChange, []byte, error)
 }
 
+// TimeProvider has functionality of Now that allows mocking of current time
+type TimeProvider interface {
+	Now() time.Time
+}
+
+// DefaultTimeProvider allows returning current time as Now()
+type DefaultTimeProvider struct{}
```
jon-whit, line 199: I suggest we do the following:

```go
package storage

// imports...

var DefaultTimeProvider = &defaultTimeProvider{}

... // other code

type defaultTimeProvider struct {}

var _ TimeProvider = (*defaultTimeProvider)(nil) // ensures we always implement the TimeProvider interface

func (d *defaultTimeProvider) Now() time.Time {
    return time.Now()
}
```

then usage looks like:

```go
var ds OpenFGADatastore

ds.SetTimeProvider(DefaultTimeProvider) // more idiomatic than ds.SetTimeProvider(DefaultTimeProvider{})
```

pkg/storage/postgres/postgres.go
```diff
@@ -90,6 +92,11 @@ func (p *Postgres) Close() {
 	p.db.Close()
 }
 
+// UpdateCurrentTime allows overwriting current time
+func (p *Postgres) UpdateCurrentTime(timeNow storage.TimeProvider) {
+	p.currentTime = timeNow
```
jon-whit, line 97: this poses a race condition

goroutineA is calling `Write` which calls `p.timeProvider.Now()`
goroutineB is calling `SetTimeProvider` which assignes `p.timeProvider = provider`

The race occurs on the read and assignment.

To protect ourselves from this we should take care to either synchronize concurrent access or use `sync.Once`.

e.g.

```
var once sync.Once

func (p *Postgres) SetTimeProvider(provider storage.TimeProvider) {
    once.Do(func() {
        p.timeProvider = provider
    })
}
```

This approach seems like it'd be fine because it avoids the race and still allows you to have full control over the implementation/mock behavior of the `TimeProvider` at the caller. You'd be able to reset/fast-forward/rewind time with the mocked TimeProvider.

adriantam: I am a bit hesitant in adding synchronization (i.e., mutex/lock) to protect a method that should only be invoked in unit test.  Rather, we should make it explicit that the method SetTimeProvider must not be used outside .  I will add in check to ensure SetTimeProvider is a no-op in non-test code.

adriantam: As discussed offline, we will revert back to use the separate test for mysql and postgresql.  However, we will abstract out the Write functionality into common.

pkg/storage/storage.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/test/tuples.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/test/tuples.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/memory/memory.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/memory/memory.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/mysql/mysql.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/postgres/postgres.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/storage.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/postgres/postgres.go
adriantam: Obsolete as we don't have the time provider anymore

pkg/storage/mysql/mysql_test.go
```diff
@@ -1,21 +1,113 @@
-package mysql_test
+package mysql
```
rhamzeh, line 1: Any reason for changing this?

pkg/storage/storage.go
```diff
@@ -16,8 +16,9 @@ const (
 )
 
 type PaginationOptions struct {
-	PageSize int
-	From     string
+	PageSize      int
+	From          string
+	OrderRequired bool
```
rhamzeh, line 21: Didn't we discuss removing OrderRequired and use the presence of Pagination options instead?

If there is any pagination then it necessitates ordering

rhamzeh (dismissed):

pkg/storage/storage.go
adriantam: Having no pagination options does not necessary mean that we should not have order (i.e., page 0 and default page size).

pkg/storage/mysql/mysql_test.go
adriantam: allow access to internal structure of mysql if needed (also be consistent with postgres_test.go as well).

pkg/storage/storage.go
rhamzeh: But in that case there still are paginationOptions (with the options being 0 and default size)

paginationOptions itself is not nil

cmd/util/util.go
```diff
@@ -1,4 +1,4 @@
-// Package util provides common utilities for spf13/cobra CLI utilities
+// Package util provides sqlcommon utilities for spf13/cobra CLI utilities
```
jon-whit, line 1: This doesn't need to change. It's referring to the literal word "common" not the package `common`.

pkg/server/errors/errors.go
```diff
@@ -133,7 +133,7 @@ func HandleError(public string, err error) error {
 	return NewInternalError(public, err)
 }
 
-// HandleTupleValidateError provide common routines for handling tuples validation error
+// HandleTupleValidateError provide sqlcommon routines for handling tuples validation error
```
jon-whit, line 136: This doesn't need to change. It's referring to the literal word "common" not the package `common`.

pkg/storage/mysql/mysql_test.go
```diff
@@ -1,21 +1,113 @@
...
+// TestReadEnsureNoOrder asserts that the read response is not ordered by ulid
+func TestReadEnsureNoOrder(t *testing.T) {
+	testDatastore := storagefixtures.RunDatastoreTestContainer(t, "mysql")
+
+	uri := testDatastore.GetConnectionURI()
+	ds, err := New(uri, sqlcommon.NewConfig())
+	require.NoError(t, err)
+
+	ctx := context.Background()
+	defer ds.Close()
```
jon-whit, line 37: Please move the defer directly below the construction.

```go
ds, err : New(...)
defer ds.Close()

require.NoError(t, err)
```

cmd/util/util.go
adriantam: fixed

pkg/server/errors/errors.go
adriantam: fixed

pkg/storage/mysql/mysql_test.go
```diff
@@ -1,21 +1,113 @@
...
+
+	store := "store"
+	objectType := "doc"
+	objectID1 := "object_id_1"
+	relation := "relation"
+	user1 := "user:user_1"
+	firstTuple := tuple.NewTupleKey(objectType+":"+objectID1, relation, user1)
```
jon-whit, line 44: Here, and elsewhere, this code can be simplified quite a bit to simply

```go
store := "store"
relation := "relation"
firstTuple := tuple.NewTupleKey("doc:1", relation, "user:1")
secondTuple := tuple.NewTupleKey("doc:2", relation, "user:2")

... // inner code

// we expect that objectID1 will return first because it is inserted first
curTuple, err := iter.Next()
require.NoError(t, err)
require.Equal(t, firstTuple, curTuple.Key)

curTuple, err = iter.Next()
require.NoError(t, err)
require.Equal(t, secondTuple, curTuple.Key)
```

Reconstructing the TupleKey objects and string concatenating the objects is not necessary here.

pkg/storage/mysql/mysql_test.go
```diff
@@ -1,21 +1,113 @@
...
```
jon-whit, line 90: Same comment here as in https://github.com/openfga/openfga/pull/677/files#r1164446781

pkg/storage/mysql/mysql_test.go
adriantam: done

adriantam: Actually, lint suggests we need to defer close after checking for error

pkg/storage/mysql/mysql_test.go
adriantam: done

pkg/storage/mysql/mysql_test.go
jon-whit: That's fine if the linter is warning.

```go
ds, err := New(...)
require.NoError(t, err)
defer ds.Close()

ctx := context.Background()
```

keeping the defer as close to the instantiation as possible

pkg/storage/mysql/mysql_test.go
```diff
@@ -1,21 +1,112 @@
...
+
+	objectID2 := "object_id_2"
+	user2 := "user:user_2"
+	secondTuple := tuple.NewTupleKey(objectType+":"+objectID2, relation, user2)
+
+	err = sqlcommon.Write(ctx, sqlcommon.NewDBInfo(ds.db, ds.stbl, sq.Expr("NOW()")), store, []*openfgav1.TupleKey{},
```
jon-whit, line 49: minor nitpick, not critical

Would you mind formatting these lines so they wrap. They are challenging to read.

```go
err = sqlcommon.Write(
  ctx,
  sqlcommon.NewDBInfo(...),
  store,
  []*openfgav1.TupleKey{},
  []*oepnfgav1.TupleKey{firstTuple},
  time.Now(),
)
```

It'd be nice if we could enforce a golint rule for this. I'll look into that.

pkg/storage/mysql/mysql_test.go
jon-whit: You changed the last bits, but you didn't simplify at the top.

```go
store := "store"
firstTuple := tuple.NewTupleKey("doc:1", "relation", "user:1")
secondTuple := tuple.NewTupleKey("doc:2", "relation", "user:2")
```

pkg/storage/mysql/mysql_test.go
adriantam: done

rhamzeh (dismissed):

jon-whit (approved):

PR #694, opened 2023-04-17
test: add perf testing workflow

miparnisari:
New secrets uploaded in https://github.com/openfga/openfga/settings/secrets/actions:

- ${{ secrets.AWS_IAM_ROLE }}
- ${{ secrets.AWS_ACCOUNT }}
- ${{ secrets.AWS_REGION }}
- ${{ secrets.OPENFGA_REGISTRY }}
- ${{ secrets.OPENFGA_ECR_URI }}
- ${{ secrets.OPENFGA_EKS_CLUSTER }}

Tested in private fork (will delete private fork after this PR is merged)


.github/workflows/load-test.yaml
```diff
@@ -0,0 +1,103 @@
...
+            helm repo update
+            helm install openfgabench-${{ steps.vars.outputs.sha_short }} openfga/openfga-benchmark \
+              --set replicaCount=1 \
+              --set image.repository=${{ secrets.OPENFGA_ECR_URI }} \
+              --set image.tag=openfgabench-${{ steps.vars.outputs.sha_short }} \
+              --set datastore.engine=postgres \
+              --set datastore.uri="postgres://postgres:password@openfga-postgresql.default.svc.cluster.local:5432/postgres?sslmode=disable" \
+              --set postgres.enabled=true \
+              --set postgresql.auth.postgresPassword=password \
+              --set postgresql.auth.database=postgres
```
adriantam, line 97: 1. Do we want to specify the M/N/K param?  Otherwise, by default, M, N and K will be 1.
2. Do we want to specify the CPU and memory for the postgres?
3. Which use case we want to run for?  I think by default it is use case 1.

.github/workflows/load-test.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+      - name: helm destroy
+        uses: koslib/helm-eks-action@master
+        with:
+          command: |
+            helm uninstall openfgabench-${{ steps.vars.outputs.sha_short }}
```
adriantam, line 103: Do we need to clean up kubectl persistent storage?

jon-whit: Yes, we should also do this here

```shell
kubectl delete pvc -l app.kubernetes.io/instance=openfgabench-${{ steps.vars.outputs.sha_short }}
```

That will delete the PVC (volume and associated data) for the Postgres instance that was spun up.

.github/workflows/load-test.yaml
jon-whit: We need the CPU and memory parameters for Postgres to get consistent benchmark runs. For our test runs we've been running with

```
  --set openfga.postgresql.primary.resources.requests.memory=2Gi \
  --set openfga.postgresql.primary.resources.requests.cpu=1.0
```

We also need to set the CPU and memory for the OpenFGA deployment itself. We've been running with:

```
  --set openfga.resources.requests.memory=1Gi \
  --set openfga.resources.requests.cpu=500m
```

miparnisari: > 1. Do we want to specify the M/N/K param? Otherwise, by default, M, N and K will be 1.
> 3. Which use case we want to run for? I think by default it is use case 1.

Added input parameters for all 👍

jon-whit (dismissed):
Looks good to me from what I can tell. Let's see how it goes!

.github/workflows/load-test.yaml
```diff
@@ -0,0 +1,141 @@
...
+  deploy:
+    runs-on: ubuntu-latest
+    environment: AWSLoadTest
+    # needed for aws-actions/configure-aws-credentials. See https://github.com/aws-actions/configure-aws-credentials/issues/271
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
adriantam, line 51: can you use the SHA instead of v3?

adriantam (approved):

.github/workflows/load-test.yaml
```diff
@@ -0,0 +1,141 @@
...
+      K:
+        required: true
+        description: large K implies a single object has a relationship with multiple users
+        type: number
+        default: 1
+      testCase:
+        required: true
+        description: the test case to run (see https://github.com/openfga/benchmarks)
+        type: choice
+        options:
```
adriantam, line 30: Do we want to omit options so that if we don't need to update this file if there are more options later on?

miparnisari: For now i'd rather keep as-is so that i don't have to figure out what test cases we have. Later on if we see that this list becomes unmanagable/frequently updated, we can change this.

.github/workflows/load-test.yaml
```diff
@@ -0,0 +1,139 @@
...
+        run: echo "sha_short=$(git rev-parse --short=8 ${{ github.event.inputs.ref }})" >> $GITHUB_OUTPUT
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4b4e9c3e2d4531116a6f8ba8e71fc6e2cb6e6c8c # v2.5.0
+
+      - name: Cache Docker layers
+        uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.event.inputs.ref }}
```
rhamzeh, line 66: This would only cache for a single commit - maybe we should cache across commits with a deterministic key?

Something like: `${{ runner.os }}-buildx-1` that we can always change if we need to evict cache

jon-whit (approved):

.github/workflows/load-test.yaml
miparnisari: Removed this step

rhamzeh (approved):

PR #695, opened 2023-04-18
test: add tests for rollbacks of database scripts

miparnisari:
This is to help detect problems like this: https://github.com/openfga/openfga/pull/691

Tested manually:

```shell
[17/04/23 7:48:07] ~/GitHub/openfga (test-migrations) $  make start-postgres 
docker run -d --name postgres -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password postgres:14
c6c088272b07293dbc12f4611dedbfbc8307267ebda45fc36a5e4059a05fa3d6
> use 'postgres://postgres:password@localhost:5432/postgres' to connect to postgres
[17/04/23 7:48:11] ~/GitHub/openfga (test-migrations) $ ./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --version 4
2023/04/17 19:48:15 current version 0
2023/04/17 19:48:15 migrating to 4
2023/04/17 19:48:15 migration done
[17/04/23 7:48:15] ~/GitHub/openfga (test-migrations) $ ./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --version 4
2023/04/17 19:48:21 current version 3
2023/04/17 19:48:21 migrating to 4
2023/04/17 19:48:21 migration done
[17/04/23 7:48:21] ~/GitHub/openfga (test-migrations) $ ./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --version 3
2023/04/17 19:48:24 current version 3
2023/04/17 19:48:24 migrating to 3
2023/04/17 19:48:24 nothing to do
[17/04/23 7:48:24] ~/GitHub/openfga (test-migrations) $ ./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --version 2
2023/04/17 19:48:26 current version 3
2023/04/17 19:48:26 migrating to 2
2023/04/17 19:48:26 migration done
[17/04/23 7:48:26] ~/GitHub/openfga (test-migrations) $ ./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --version 1
2023/04/17 19:48:28 current version 2
2023/04/17 19:48:28 migrating to 1
2023/04/17 19:48:28 migration done
[17/04/23 7:48:28] ~/GitHub/openfga (test-migrations) $ ./openfga migrate --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --version 0
2023/04/17 19:48:29 current version 1
2023/04/17 19:48:29 running all migrations
2023/04/17 19:48:29 migration done
[17/04/23 7:48:29] ~/GitHub/openfga (test-migrations) $ 
```

### DIFF
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -116,13 +116,20 @@ func runMigration(_ *cobra.Command, _ []string) error {
 
 	goose.SetBaseFS(assets.EmbedMigrations)
 
-	if targetVersion > 0 {
-		currentVersion, err := goose.GetDBVersion(db)
-		if err != nil {
+	currentVersion, err := goose.GetDBVersion(db)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Printf("current version %d", currentVersion)
+
+	if targetVersion == 0 {
+		log.Println("running all migrations")
+		if err := goose.Up(db, migrationsPath); err != nil {
 			log.Fatal(err)
 		}
-		log.Printf("current version %d, migrating to %d", currentVersion, targetVersion)
-
+	} else {
+		log.Printf("migrating to %d", targetVersion)
 		targetInt64Version := int64(targetVersion)
 		if targetInt64Version < currentVersion {
 			if err := goose.DownTo(db, migrationsPath, targetInt64Version); err != nil {
@@ -136,15 +143,6 @@ func runMigration(_ *cobra.Command, _ []string) error {
 			log.Println("nothing to do")
 			return nil
 		}
-
-		log.Println("migration done")
-
-		return nil
-	}
-
-	// no target version specified, will apply all migrations available
-	if err := goose.Up(db, migrationsPath); err != nil {
-		log.Fatal(err)
 	}
 
 	log.Println("migration done")
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -61,7 +61,7 @@ func NewMigrateCommand() *cobra.Command {
 func runMigration(_ *cobra.Command, _ []string) error {
 	engine := viper.GetString(datastoreEngineFlag)
 	uri := viper.GetString(datastoreURIFlag)
-	version := viper.GetUint(versionFlag)
+	targetVersion := viper.GetUint(versionFlag)
 	timeout := viper.GetDuration(timeoutFlag)
 
 	goose.SetLogger(goose.NopLogger())
@@ -116,26 +116,33 @@ func runMigration(_ *cobra.Command, _ []string) error {
 
 	goose.SetBaseFS(assets.EmbedMigrations)
 
-	if version > 0 {
+	if targetVersion > 0 {
 		currentVersion, err := goose.GetDBVersion(db)
 		if err != nil {
 			log.Fatal(err)
 		}
+		log.Printf("current version %d, migrating to %d", currentVersion, targetVersion)
 
-		int64Version := int64(version)
-		if int64Version < currentVersion {
-			if err := goose.DownTo(db, migrationsPath, int64Version); err != nil {
+		targetInt64Version := int64(targetVersion)
+		if targetInt64Version < currentVersion {
+			if err := goose.DownTo(db, migrationsPath, targetInt64Version); err != nil {
 				log.Fatal(err)
 			}
+		} else if targetInt64Version > currentVersion {
+			if err := goose.UpTo(db, migrationsPath, targetInt64Version); err != nil {
+				log.Fatal(err)
+			}
+		} else {
+			log.Println("nothing to do")
+			return nil
 		}
 
-		if err := goose.UpTo(db, migrationsPath, int64Version); err != nil {
-			log.Fatal(err)
-		}
+		log.Println("migration done")
 
 		return nil
 	}
 
+	// no target version specified, will apply all migrations available
 	if err := goose.Up(db, migrationsPath); err != nil {
 		log.Fatal(err)
 	}
--- a/pkg/storage/mysql/mysql_test.go
+++ b/pkg/storage/mysql/mysql_test.go
@@ -2,10 +2,12 @@ package mysql
 
 import (
 	"context"
+	"strconv"
 	"testing"
 	"time"
 
 	sq "github.com/Masterminds/squirrel"
+	"github.com/openfga/openfga/cmd"
 	"github.com/openfga/openfga/pkg/storage"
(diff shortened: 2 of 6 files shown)


cmd/migrate.go
```diff
@@ -116,26 +116,33 @@ func runMigration(_ *cobra.Command, _ []string) error {
 
 	goose.SetBaseFS(assets.EmbedMigrations)
 
-	if version > 0 {
+	if targetVersion > 0 {
```
rhamzeh, line 119: Can we add the shortest case first? And make the logs more consistent

E.g. always print the current version and always print what migrations are about to be run and when they are done, e.g.

```go
currentVersion, err := goose.GetDBVersion(db)
if err != nil {
	log.Fatal(err)
}
if targetVersion == 0 {
  log.Println("current version %d, running all migrations")
...
} else {
  log.Printf("current version %d, migrating to %d", currentVersion, targetVersion)
  targetInt64Version := int64(targetVersion)
  if targetInt64Version == currentVersion {
    log.Println("nothing to do")
  } else ...
  ...
  ...
}
```
log.Println("migration done")

pkg/storage/mysql/mysql_test.go
```diff
@@ -25,6 +27,30 @@ func TestMySQLDatastore(t *testing.T) {
...
+	// starts the container and runs migration up to the latest migration version available
+	testDatastore := storagefixtures.RunDatastoreTestContainer(t, engine)
+
+	uri := testDatastore.GetConnectionURI()
+	ds, err := New(uri, sqlcommon.NewConfig())
+	require.NoError(t, err)
+	defer ds.Close()
+
+	// going from version 3 to 4 when migration #4 doesn't exist is a no-op
+	version := testDatastore.GetDatabaseVersion() + 1
```
adriantam, line 41: Do we have control on the database version in the test?  (Reason I am asking is that how do we know that the current version is 4 and maximum version is 4 - which will result in a no-op).

Should we run downgrade until 0 (which you already do) and then run upgrade to maximum to ensure both upgrade and downgrade are tested?

cmd/migrate.go
```diff
@@ -116,28 +116,33 @@ func runMigration(_ *cobra.Command, _ []string) error {
...
 			log.Fatal(err)
 		}
-
-		int64Version := int64(version)
-		if int64Version < currentVersion {
-			if err := goose.DownTo(db, migrationsPath, int64Version); err != nil {
+	} else {
+		log.Printf("migrating to %d", targetVersion)
+		targetInt64Version := int64(targetVersion)
+		if targetInt64Version < currentVersion {
```
adriantam, line 134: Will the most likely case be 
i) no-op
ii) upgrade
iii) downgrade
?

In this case, does it make sense to order the if statement according to this order to optimize path?

pkg/testfixtures/storage/storage.go
```diff
@@ -10,6 +10,9 @@ type DatastoreTestContainer interface {
 	// GetConnectionURI returns a connection string to the datastore instance running inside
 	// the container.
 	GetConnectionURI() string
+
+	// GetDatabaseVersion returns the last migration applied (e.g. 3) when the container was created
+	GetDatabaseVersion() int64
```
jon-whit, line 15:
```suggestion
	GetDatabaseSchemaVersion() int64
```

pkg/storage/postgres/postgres_test.go
```diff
@@ -26,6 +28,30 @@ func TestPostgresDatastore(t *testing.T) {
...
+	// starts the container and runs migration up to the latest migration version available
+	testDatastore := storagefixtures.RunDatastoreTestContainer(t, engine)
+
+	uri := testDatastore.GetConnectionURI()
+	ds, err := New(uri, sqlcommon.NewConfig())
+	require.NoError(t, err)
+	defer ds.Close()
+
+	// going from version 3 to 4 when migration #4 doesn't exist is a no-op
+	version := testDatastore.GetDatabaseVersion() + 1
```
jon-whit, line 42: This behavior and comment is a bit misleading. If you try to migrate to revision 4 and it is an undefined revision, then that should return an error. There should be two cases worth testing

1. Omitting the revision defaults to latest.

   `openfga migrate --datastore-engine=postgres --datastore-uri="..."` (defaults to latest)

2. Specifying the revision explicitly should require a valid revision number.

    `openfga migrate --datastore-engine=postgres --datastore-uri="..." 10` (error, revision 10 is undefined)

    `openfga migrate --datastore-engine=postgres --datastore-uri="..." 3` (success)

pkg/storage/mysql/mysql_test.go
miparnisari: By default, when you create a test container, it runs all `up` migrations up to the latest one available.

So in the test i'm doing:
1. try to update to one version more -> no-op
2. run `down` until version 0 is reached

pkg/storage/postgres/postgres_test.go
miparnisari: > If you try to migrate to revision 4 and it is an undefined revision, then that should return an error.

If you're on version 3 and try to migrate to version 4, Goose doesn't do anything. It feels weird to introduce an error when there is none, I see it as a deletion where it's already deleted, in REST that results in a no-op as well

cmd/migrate.go
miparnisari: I don't understand, what would we be optimizing?

adriantam: What I meant is that should we do something like this?

```
   if targetInt64Version == currentVersion {
     // no-op
   } else if targetInt64Version < currentVersion {
     //upgrade
   } else {
     //downgrade
   }
```
as downgrade seldomly happen (save a comparison or two).

miparnisari: Migrations are run once in a blue moon.. i don't think that would classify as an optimization

adriantam (approved):

pkg/storage/postgres/postgres_test.go
miparnisari: @jon-whit i'm going to merge this as-is but if you think we should introduce an error when users try to migrate from version X to version X+1 when the highest one available is X, please open an issue. Thanks!

PR #701, opened 2023-04-20
chore: pin MySQL container image to v8

miparnisari:
- chore: pin MySQL container image to v8
- normalize tabs/spaces in migration script

### DIFF
--- a/pkg/testfixtures/storage/mysql.go
+++ b/pkg/testfixtures/storage/mysql.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	mySQLImage = "mysql:latest"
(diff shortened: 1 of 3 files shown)


pkg/testfixtures/storage/mysql.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	mySQLImage = "mysql:latest"
+	mySQLImage = "mysql:8"
```
miparnisari, line 24: same as https://github.com/openfga/openfga/blob/11797960cc3e3040785d5f19f02c1d786d9a3f97/Makefile#L45-L47

jon-whit (approved):

Issue #703, opened 2023-04-20
Internal server error of `context deadline exceeded` for list object with `but not` relationship under load


adriantam:
While running performance test with the following model

```
model
  schema 1.1
type user
type folder
  relations
    define viewer: [user]
type group
  relations
    define member: [user]
type document
  relations
    define allowed: [user]
    define editor: [user] or owner
    define editor_and_allowed: editor and allowed
    define editor_not_restricted: editor but not restricted
    define owner: [user]
    define parent: [folder]
    define restricted: [user]
    define viewer: [user,group#member] or editor or viewer from parent
    define viewer_and_allowed: viewer and allowed
    define viewer_not_restricted: viewer but not restricted
```

List object query of 
```
{
   authorization_model_id: "...",
   type: "document",
   relation: "viewer_not_restricted",
   user: "user:C-0-0"
}
```
occasionally results in 500 Internal Error.  The benchmark tool shows the OpenFGA response as 

```
[...]

PR #704, opened 2023-04-20
fix: list objects to respect configurable ListObjectsDeadline

adriantam:
## Description
List objects should return results when ListObjectsDeadline is hit rather than waiting for results to be completed. Otherwise, 500 internal error will be returned if the request takes a long time (i.e., maybe due to many items to be looked up).

## References
- Close https://github.com/openfga/openfga/issues/703

## Tests
- Added unit test
- Manual test by adding a 10 seconds timeout in the [reverseExpandObject](https://github.com/openfga/openfga/blob/f33657319d7885600a19c112509868a6ee1c5b3e/pkg/server/commands/connected_objects.go#L452).  Observe than 500 Internal Error is returned before the change.  200 OK with empty list is returned after the change.

## Review Checklist

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -197,9 +197,8 @@ func (q *ListObjectsQuery) Execute(
 
 		case <-timeoutCtx.Done():
 			q.Logger.WarnWithContext(
-				ctx, fmt.Sprintf("list objects timeout with relation '%s'", req.Relation),
-				zap.String("store_id", req.GetStoreId()),
-				zap.String("object_type", req.Type),
+				ctx, "list objects timeout with list object configuration timeout",
+				zap.String("timeout duration", q.ListObjectsDeadline.String()),
 			)
 			return &openfgapb.ListObjectsResponse{
 				Objects: objects,
@@ -250,9 +249,8 @@ func (q *ListObjectsQuery) ExecuteStreamed(
 
 		case <-timeoutCtx.Done():
 			q.Logger.WarnWithContext(
-				ctx, fmt.Sprintf("streamed list objects timeout with relation '%s'", req.Relation),
-				zap.String("store_id", req.GetStoreId()),
-				zap.String("object_type", req.Type),
+				ctx, "list objects timeout with list object configuration timeout",
+				zap.String("timeout duration", q.ListObjectsDeadline.String()),
 			)
 			return nil
 
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -194,6 +194,17 @@ func (q *ListObjectsQuery) Execute(
 
 	for {
 		select {
+
+		case <-timeoutCtx.Done():
+			q.Logger.WarnWithContext(
+				ctx, fmt.Sprintf("list objects timeout with relation '%s'", req.Relation),
+				zap.String("store_id", req.GetStoreId()),
+				zap.String("object_type", req.Type),
+			)
+			return &openfgapb.ListObjectsResponse{
+				Objects: objects,
+			}, nil
+
 		case objectID, ok := <-resultsChan:
 			if !ok {
 				return &openfgapb.ListObjectsResponse{
@@ -236,6 +247,15 @@ func (q *ListObjectsQuery) ExecuteStreamed(
 
 	for {
 		select {
+
+		case <-timeoutCtx.Done():
+			q.Logger.WarnWithContext(
+				ctx, fmt.Sprintf("streamed list objects timeout with relation '%s'", req.Relation),
+				zap.String("store_id", req.GetStoreId()),
+				zap.String("object_type", req.Type),
+			)
+			return nil
+
 		case object, ok := <-resultsChan:
 			if !ok {
 				// Channel closed! No more results.
--- a/pkg/server/test/list_objects.go
+++ b/pkg/server/test/list_objects.go
@@ -29,6 +29,119 @@ func (x *mockStreamServer) Send(m *openfgapb.StreamedListObjectsResponse) error 
 	return nil
 }
 
+// mockSlowDataStorage is a proxy to the actual ds except the Reads are slow time by the sleepTime
+// This allows simulating list objection condition that times out
+type mockSlowDataStorage struct {
+	sleepTime time.Duration
+	ds        storage.OpenFGADatastore
+}
+
+func NewMockSlowDataStorage(ds storage.OpenFGADatastore, sleepTime time.Duration) storage.OpenFGADatastore {
+	return &mockSlowDataStorage{
+		sleepTime: sleepTime,
+		ds:        ds,
+	}
+}
+
+func (m *mockSlowDataStorage) Close() {}
+
+func (m *mockSlowDataStorage) ListObjectsByType(ctx context.Context, store string, objectType string) (storage.ObjectIterator, error) {
+	return m.ds.ListObjectsByType(ctx, store, objectType)
+}
+
+func (m *mockSlowDataStorage) Read(ctx context.Context, store string, key *openfgapb.TupleKey) (storage.TupleIterator, error) {
+	time.Sleep(m.sleepTime)
+	return m.ds.Read(ctx, store, key)
+}
+
+func (m *mockSlowDataStorage) ReadPage(ctx context.Context, store string, key *openfgapb.TupleKey, paginationOptions storage.PaginationOptions) ([]*openfgapb.Tuple, []byte, error) {
+	time.Sleep(m.sleepTime)
+	return m.ds.ReadPage(ctx, store, key, paginationOptions)
+}
+
+func (m *mockSlowDataStorage) ReadChanges(ctx context.Context, store, objectType string, paginationOptions storage.PaginationOptions, horizonOffset time.Duration) ([]*openfgapb.TupleChange, []byte, error) {
+	return m.ds.ReadChanges(ctx, store, objectType, paginationOptions, horizonOffset)
+}
+
+func (m *mockSlowDataStorage) Write(ctx context.Context, store string, deletes storage.Deletes, writes storage.Writes) error {
+	return m.ds.Write(ctx, store, deletes, writes)
+}
+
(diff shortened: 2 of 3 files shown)


pkg/server/test/list_objects.go
```diff
@@ -29,6 +29,119 @@ func (x *mockStreamServer) Send(m *openfgapb.StreamedListObjectsResponse) error
 	return nil
 }
 
+// mockSlowDataStorage is a proxy to the actual ds except the Reads are slow time by the sleepTime
+// This allows simulating list objection condition that times out
+type mockSlowDataStorage struct {
```
jon-whit, line 34: Please put mocked storage interfaces in the `pkg/storage/mocks` package.

pkg/server/test/list_objects.go
```diff
@@ -158,6 +273,33 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 			minimumResultsExpected: 1,
 			allResults:             []string{"team:1"},
 		},
+		{
+			// should not return any useful data as it times out
+			// however, there should not be any error
+			name:   "connected objects timeout",
```
jon-whit, line 279: Use underscores here please `connected_objects_timeout`

pkg/server/test/list_objects.go
adriantam: Done

pkg/server/test/list_objects.go
adriantam: Done

pkg/server/test/list_objects.go
```diff
@@ -158,6 +161,33 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
...
+			},
+			user:       "user:alice",
+			objectType: "repo",
+			relation:   "admin",
+			contextualTuples: &openfgapb.ContextualTupleKeys{
+				TupleKeys: []*openfgapb.TupleKey{tuple.NewTupleKey("repo:3", "admin", "user:alice")},
+			},
+			maxResults:             2,
+			minimumResultsExpected: 0,
+			allResults:             []string{},
```
miparnisari, line 187: `allResults` should be the list of results that you'd expect if you ran `Check` for all objects, without caring about the ListObjects API settings. So in this test `repo:1 , repo:2, repo:3` are all the valid results.

The expectations that we have for this test is that:
-  `err == nil`
- `0 <= len(results)`


If so, `minimumResultsExpected: 0` is okay, but `maxResults` can be set to a million and it should still pass. Perhaps we should have a default of 100000 and just not set it in this test

pkg/storage/mocks/slow_storage.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/openfga/openfga/pkg/storage"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+// SlowDataStorage is a proxy to the actual ds except the Reads are slow time by the sleepTime
+// This allows simulating list objection condition that times out
+type SlowDataStorage struct {
+	sleepTime time.Duration
+	ds        storage.OpenFGADatastore
+}
```
miparnisari, line 16:
```suggestion
type slowDataStorage struct {
	readTuplesDelay time.Duration
	ds        storage.OpenFGADatastore
}
```

pkg/storage/mocks/slow_storage.go
```diff
@@ -0,0 +1,124 @@
...
+
+// NewMockSlowDataStorage returns SlowDataStorage which is a proxy to the actual ds except the Reads are slow time by the sleepTime
+// This allows simulating list objection condition that times out
```
miparnisari, line 19:
```suggestion
// NewMockSlowDataStorage returns a wrapper of a datastore that adds artificial delays into the reads of tuples
```

pkg/server/test/list_objects.go
```diff
@@ -158,6 +161,33 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 			minimumResultsExpected: 1,
 			allResults:             []string{"team:1"},
 		},
+		{
+			// should not return any useful data as it times out
+			// however, there should not be any error
+			name:   "list_objects_timeout",
```
miparnisari, line 167:
```suggestion
			name:   "respects_max_results_when_deadline_timeout_and_returns_no_error_and_no_results",
```

miparnisari (changes requested):

pkg/server/test/list_objects.go
adriantam: No, we specifically expect there will be no objects returned as all reads are timed out

pkg/server/test/list_objects.go
```diff
@@ -158,6 +161,33 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
...
+			tuples: []*openfgapb.TupleKey{
+				tuple.NewTupleKey("repo:1", "admin", "user:alice"),
+				tuple.NewTupleKey("repo:2", "admin", "user:alice"),
+			},
+			user:       "user:alice",
+			objectType: "repo",
+			relation:   "admin",
+			contextualTuples: &openfgapb.ContextualTupleKeys{
+				TupleKeys: []*openfgapb.TupleKey{tuple.NewTupleKey("repo:3", "admin", "user:alice")},
+			},
```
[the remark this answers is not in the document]
adriantam: I will remove the contextual tuples.

pkg/server/commands/list_objects.go
```diff
@@ -194,6 +194,17 @@ func (q *ListObjectsQuery) Execute(
 
 	for {
 		select {
+
+		case <-timeoutCtx.Done():
+			q.Logger.WarnWithContext(
+				ctx, fmt.Sprintf("list objects timeout with relation '%s'", req.Relation),
```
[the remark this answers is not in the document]
adriantam: The reason that I didn't choose to include the actual timeout field is that it is not changing for the duration of the openfga process.  Thus, it does not convey any new info per log.

pkg/server/commands/list_objects.go
```diff
@@ -194,6 +194,17 @@ func (q *ListObjectsQuery) Execute(
...
+				zap.String("store_id", req.GetStoreId()),
+				zap.String("object_type", req.Type),
```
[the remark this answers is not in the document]
adriantam: This is similar reason why we log store id in warn message in line 102.

pkg/server/test/list_objects.go
```diff
@@ -41,6 +42,8 @@ type listObjectsTestCase struct {
 	allResults             []string //all the results. the server may return less
 	maxResults             uint32
 	minimumResultsExpected uint32
+	listObjectsDeadline    time.Duration // 1 minute if not set
+	dsSlowTime             time.Duration // if set, purposely use a slow storage to slow down read and simulate timeout
 }
```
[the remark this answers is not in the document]
adriantam: Agree on this

pkg/server/test/list_objects.go
```diff
@@ -41,6 +42,8 @@ type listObjectsTestCase struct {
 	allResults             []string //all the results. the server may return less
 	maxResults             uint32
 	minimumResultsExpected uint32
+	listObjectsDeadline    time.Duration // 1 minute if not set
+	dsSlowTime             time.Duration // if set, purposely use a slow storage to slow down read and simulate timeout
```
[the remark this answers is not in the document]
adriantam: I am a bit hesitant in doing this as we could later add test where the dsSlowTime is slightly smaller than the list object deadline and one of the item will be returned.

pkg/server/commands/list_objects.go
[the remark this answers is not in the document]
adriantam: However, I can add this in to make things more explicit.

pkg/server/test/list_objects.go
adriantam: I will add comments to make it explicit that we expect empty array and test consider failure if there are any.

pkg/storage/mocks/slow_storage.go
```diff
@@ -0,0 +1,123 @@
...
+	"context"
+	"time"
+
+	"github.com/openfga/openfga/pkg/storage"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+// SlowDataStorage is a proxy to the actual ds except the Reads are slow time by the readTuplesDelay
+// This allows simulating list objection condition that times out
+type SlowDataStorage struct {
```
miparnisari, line 13:
```suggestion
// SlowDataStorage is a proxy to the actual ds except the Reads are slow time by the readTuplesDelay
// This allows simulating list objection condition that times out
type slowDataStorage struct {
```

no need to export this struct

pkg/storage/mocks/slow_storage.go
```diff
@@ -0,0 +1,123 @@
...
+		readTuplesDelay: readTuplesDelay,
+		ds:              ds,
+	}
+}
+
+func (m *SlowDataStorage) Close() {}
+
+func (m *SlowDataStorage) ListObjectsByType(ctx context.Context, store string, objectType string) (storage.ObjectIterator, error) {
+	return m.ds.ListObjectsByType(ctx, store, objectType)
+}
```
miparnisari, line 30: this one is a read so it should also have `time.Sleep`

pkg/server/test/list_objects.go
```diff
@@ -158,6 +161,31 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 			minimumResultsExpected: 1,
 			allResults:             []string{"team:1"},
 		},
+		{
+			// should not return any useful data as it times out
+			// however, there should not be any error
```
miparnisari, line 166: the comment is not adding any new information, i would remove it

pkg/server/test/list_objects.go
miparnisari: > we specifically expect there will be no objects returned as all reads are timed out.

This is checked via `minimumResultsExpected: 0`. You should still set `allResults = [repo:1 , repo:2, repo:3]` because this assertion will still hold true:

```
require.Subset(t, test.allResults, res.Objects)
```

miparnisari (approved):

PR #702, opened 2023-04-20
feat: add three more columns to tuple and changelog tables, part 1 of 3

miparnisari:
This PR is the first step towards
- deprecating the `_user` column from the `tuple` and `changelog` tables
- deprecating the `user_type` column from the `tuple` table
- replacing the reverse index with a new one
- replacing the partial indexes in postgres with new ones

The final goal is to make `ReadUsersetTuples` faster by not having a SQL LIKE statement.

This PR:
- adds a new migration script that backfills existing tuples and makes sure that future tuples have the new columns
- changes the server's Writes to write to the new columns
- changes the server's health check so that it also ensures the latest DB migration has been run

Sample migration of store with 4 million tuples in Postgres:

- 5 minutes to run `UPDATE tuple` command
- 1 minute to run `UPDATE changelog` command

<details>
  <summary>Sample migration</summary>

  ```shell
[24/04/23 5:28:06] ~/GitHub/openfga (feat-split-user-column) $ make test-migration-postgres 
{ \
    URI=postgres://postgres:password@localhost:5432/postgres; \
    ENGINE=postgres; \
    TOTALTUPLES=4000000; \
    docker rm -f ${ENGINE} && \
        make start-${ENGINE} && \
        ./openfga migrate --datastore-engine ${ENGINE} --datastore-uri ${URI} --version 3 && \
        go run ./scripts/loaddata.go ${ENGINE} ${URI} ${TOTALTUPLES} && \
        ./openfga migrate --datastore-engine ${ENGINE} --datastore-uri ${URI} --version 4 && \
        ./openfga migrate --datastore-engine ${ENGINE} --datastore-uri ${URI} --version 3; \
        } > migration-postgres.log 2>&1
[24/04/23 5:37:54] ~/GitHub/openfga (feat-split-user-column) $ 
```

```
postgres
docker run -d --name postgres -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password postgres:14
8b087ef53b877b8e7c0bd384b93989b359b8f5d02fd28ede229b5e932fe2d329
> use 'postgres://postgres:password@localhost:5432/postgres' to connect to postgres
2023/04/24 17:28:31 current version 0
2023/04/24 17:28:31 migrating to 3
2023/04/24 17:28:31 -- +goose Up
2023/04/24 17:28:31 [90mStateMachine: set 0 => 1[00m
2023/04/24 17:28:31 CREATE TABLE tuple (
[...]

PR #706, opened 2023-04-21
[DO NOT MERGE] feat: add three more columns to tuple and changelog tables, part 2

miparnisari:
This is part 2 of the migration. 

Pre-requisite: https://github.com/openfga/openfga/pull/702


rhamzeh:
To have the tests run here, add in the relevant workflow, under branches alongside `main`: `feat-split-user-column-part2`

assets/migrations/postgres/005_drop_indices.sql
```diff
@@ -0,0 +1,40 @@
+-- +goose NO TRANSACTION
+-- This migration updates the primary key of the tuple table, drops old indexes
+
+-- +goose Up
+
+ALTER TABLE tuple
```
jon-whit, line 6: To do this with nearly 0 locking and reduce interruptions to existing queries you'll need to first build a unique index concurrently and then replace the primary key using that index.

For example:

```sql
CREATE UNIQUE INDEX CONCURRENTLY idx_tuple_pkey
   ON tuple (store, object_type, object_id, relation, user_object_type, user_object_id, user_relation);

ALTER TABLE tuple
   ADD CONSTRAINT tuple_pkey PRIMARY KEY USING INDEX idx_tuple_pkey;
```

The `ALTER TABLE .. ADD CONSTRAINT .. PRIMARY KEY` will lock the whole table, but the constraint swaps over to the index immediately and the indexing continues concurrently. So the effect to table/row level locks is much more negligible than the alternative.

assets/migrations/postgres/005_drop_indices.sql
```diff
@@ -0,0 +1,40 @@
...
+    ALTER COLUMN user_relation SET NOT NULL;
+
+-- +goose Down
+
+CREATE INDEX idx_tuple_partial_user ON tuple (store, object_type, object_id, relation, _user) WHERE user_type = 'user';
+CREATE INDEX idx_tuple_partial_userset ON tuple (store, object_type, object_id, relation, _user) WHERE user_type = 'userset';
+
+ALTER TABLE tuple
+    DROP CONSTRAINT tuple_pkey,
+    ADD PRIMARY KEY (store, object_type, object_id, relation, _user),
```
jon-whit, line 32: Similar to https://github.com/openfga/openfga/pull/706/files#r1204728556, we'll need to swap the PRIMARY KEY back using a index to avoid excessive locking.

assets/migrations/mysql/005_drop_indices.sql
```diff
@@ -0,0 +1,28 @@
+--  +goose Up
+--   This migration updates the primary key of the tuple table
+
+ ALTER TABLE tuple
```
jon-whit, line 4: It doesn't appear that this is a locking operation. I was able to continue reading/writing concurrently while this was doing it's work.

assets/migrations/postgres/005_drop_indices.sql
```diff
@@ -0,0 +1,46 @@
...
+
+ALTER TABLE tuple
+    ALTER COLUMN user_object_type SET NOT NULL,
+    ALTER COLUMN user_object_id SET NOT NULL,
+    ALTER COLUMN user_relation SET NOT NULL,
+    DROP CONSTRAINT tuple_pkey,
+    ADD CONSTRAINT tuple_pkey PRIMARY KEY USING INDEX idx_tuple_pkey;
+
+-- +goose StatementBegin
+DROP INDEX CONCURRENTLY idx_tuple_partial_user;
```
jon-whit, line 17: If we drop these indexes and we don't change our queries to start using the new columns that have been backfilled in the prior migration, then we'll suffer performance penalities. I test this out, and our latency went up because these indexes no longer serve the `ReadUserTuple` and `ReadUsersetTuples` queries, because they use the old column(s) in their query.

jon-whit:
```sql
Index Scan using idx_reverse_lookup_user on tuple  (cost=0.55..8.58 rows=1 width=121) (actual time=0.026..0.027 rows=1 loops=1)
   Index Cond: ((store = '01H1A9YPAYYH7H54PPPJMVN4J1'::text) AND (object_type = 'document'::text) AND (relation = 'viewer'::text) AND (_user = 'user:0'::text))
   Filter: ((object_id = 'budget'::text) AND (user_type = 'user'::text))
 Planning Time: 0.083 ms
 Execution Time: 0.040 ms
```

versus

```
 Index Scan using idx_tuple_partial_user on tuple  (cost=0.55..8.58 rows=1 width=121) (actual time=0.043..0.044 rows=1 loops=1)
   Index Cond: ((store = '01H1A9YPAYYH7H54PPPJMVN4J1'::text) AND (object_type = 'document'::text) AND (object_id = 'budget'::text) AND (relation = 'viewer'::text) AND (_user = 'user:0'::text))
 Planning Time: 0.292 ms
 Execution Time: 0.054 ms
```

curfew-marathon:
**We're closing this PR and will circle back when the work is prioritized.  Do not delete the branch**

PR #708, opened 2023-04-21
chore: every PR runs the PR workflow no matter what branch they target

miparnisari:
I need this so that I can run the tests on https://github.com/openfga/openfga/pull/706 and https://github.com/openfga/openfga/pull/707, which target a branch that is not `main`

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -1,8 +1,6 @@
 name: Pull Request
 on:
   pull_request:
-    branches:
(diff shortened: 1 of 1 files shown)


rhamzeh:
You do not need this. You can make the change in those PRs to add the relevant branch to the branches alongside main.

In that way it will run for those PRs, and you can remove that before merging, without forcing it to run on every branch.

rhamzeh (changes requested):
.

Issue #690, opened 2023-04-12
Writing Authorization Models to a non-existent store returns 201


aaguiarz:
When you call POST to `/stores/{store_id}/authorization-models` with a non-existent store_id you get back a 201 and a new model_id.  

We should return an error.

Himanshu372:
Addressed this issue in the PR - https://github.com/openfga/openfga/pull/710

hello-caleb:
Due to the complexity of this ongoing issue, recommend that the `good first issue` tag be removed.

PR #710, opened 2023-04-21
Adding validation for nonexistent storeID write

Himanshu372:
A write request for nonexistent storeID shouldn't return 201, but should return an error

## Description

Changes 
1. Added a check on storeID being non empty when validating WriteRequest 
2. Added a custom error for nonexistent storeID error, but resued error code `ErrorCode_invalid_write_input`
3. Added test  

## References

Addresses issue - https://github.com/openfga/openfga/issues/690 

## Review Checklist

### DIFF
--- a/pkg/server/commands/write.go
+++ b/pkg/server/commands/write.go
@@ -51,11 +51,16 @@ func (c *WriteCommand) Execute(ctx context.Context, req *openfgapb.WriteRequest)
 func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.WriteRequest) error {
 	ctx, span := tracer.Start(ctx, "validateWriteRequest")
 	defer span.End()
-
 	store := req.GetStoreId()
+
 	// Check if store is empty, can't write to empty store
-	if store == "" {
-		return serverErrors.NonExistentStoreID
+	if _, err := c.datastore.GetStore(ctx, store); err != nil {
+		if err != nil {
+			if errors.Is(err, storage.ErrNotFound) {
+				return serverErrors.StoreIDNotFound
+			}
+			return serverErrors.HandleError("", err)
+		}
 	}
 	modelID := req.GetAuthorizationModelId()
 	deletes := req.GetDeletes().GetTupleKeys()
--- a/pkg/server/commands/write.go
+++ b/pkg/server/commands/write.go
@@ -53,6 +53,10 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 	defer span.End()
 
 	store := req.GetStoreId()
+	// Check if store is empty, can't write to empty store
+	if store == "" {
+		return serverErrors.NonExistentStoreID
+	}
 	modelID := req.GetAuthorizationModelId()
 	deletes := req.GetDeletes().GetTupleKeys()
 	writes := req.GetWrites().GetTupleKeys()
--- a/pkg/server/commands/write.go
+++ b/pkg/server/commands/write.go
@@ -53,7 +53,7 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 	defer span.End()
 	store := req.GetStoreId()
 
-	// Check if store is empty, can't write to empty store
+	// Check if store is invalid/empty, can't write to invalid/empty store
 	if _, err := c.datastore.GetStore(ctx, store); err != nil {
 		if err != nil {
 			if errors.Is(err, storage.ErrNotFound) {
--- a/pkg/server/commands/read.go
+++ b/pkg/server/commands/read.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/openfga/openfga/pkg/encoder"
@@ -37,6 +38,16 @@ func NewReadQuery(datastore storage.OpenFGADatastore, logger logger.Logger, enco
 func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*openfgapb.ReadResponse, error) {
 	store := req.GetStoreId()
 	tk := req.GetTupleKey()
+
+	// Check if store is empty, can't write to empty store
+	if _, err := q.datastore.GetStore(ctx, store); err != nil {
+		if err != nil {
+			if errors.Is(err, storage.ErrNotFound) {
+				return nil, serverErrors.StoreIDNotFound
+			}
+			return nil, serverErrors.HandleError("", err)
+		}
+	}
 
 	// Restrict our reads due to some compatibility issues in one of our storage implementations.
 	if tk != nil {
--- a/pkg/server/commands/read.go
+++ b/pkg/server/commands/read.go
@@ -39,7 +39,7 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgapb.ReadRequest) (*o
 	store := req.GetStoreId()
 	tk := req.GetTupleKey()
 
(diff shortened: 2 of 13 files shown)


Himanshu372:
@adriantam Plz check the PR when you have time. Thanks

pkg/server/commands/write.go
```diff
@@ -53,6 +53,10 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 	defer span.End()
 
 	store := req.GetStoreId()
+	// Check if store is empty, can't write to empty store
+	if store == "" {
```
adriantam, line 57: Instead of checking whether the store is null - as we have already validate it as part of [protobuf API](https://github.com/openfga/api/blob/157c45499d9ba7218b98601a0608281ddc82ee1b/openfga/v1/openfga_service.proto#L809), we should call [GetStore](https://github.com/openfga/openfga/blob/c108f3c5142d3aa8a10909b9ee253d4a4151ea92/pkg/storage/storage.go#L175) to see whether store is returned or the store is non-existing.

adriantam (changes requested):
@Himanshu372 : thank you for your contribution.  In addition to do the validation on write, can we also do it for the [other APIs](https://openfga.dev/api/service#/) where {store_id} is part of the path parameter?  If it is too much or outside of your comfort zone, please feel free to let us know and we can do the other pieces :)

pkg/server/commands/write.go
jon-whit: We'll also want to ensure that we do the GetStore read in the same transaction as the other statements. Otherwise we'll have a race on the store's visibility relative to other concurrent transactions. As part of this work, let's consider adding a `WrapTransaction` or similarly named signature to the storage interface. For example:


```go

type TxnFuncHandler func(ctx context.Context) error

type OpenFGADatstore interface {
    ...
    WrapTransaction(ctx context.Context, txnFunc TxnFuncHandler) error
}
```

then, in calling code, we can check for the store's existence within the same txn

```go

func (cmd *WriteCommand) Write(ctx context.Context, ...) (...) {
    ... // other code
    
    err := cmd.datastore.WrapTransaction(ctx, func(ctx contextContext) error {
        _, err := cmd.datastore.GetStore(ctx, req.GetStoreId())
        if err != nil {
                // handle non-existence case
            }

            // handle general error
        }

    ... // handle other Write database code.
    })
}
```

How we compose this is up for debate, but the importance is that we avoid races on the store's visibility relative to the transactions that have committed.

miparnisari: > Otherwise we'll have a race on the store's visibility relative to other concurrent transactions

@jon-whit I don't follow, can you give an example of what problems may happen if we don't do a transaction?

jon-whit: @miparnisari Sure. Consider this scenario:

Two clients are concurrently interacting with the OpenFGA API. Client 1 calls the `Write` API against `store1`. Meanwhile, client 2 calls the `DeleteStore` API against `store1`. Consider the following code in the `Write` API:

```go
func (cmd *WriteCommand) Execute(ctx context.Context, req *openfgapb.WriteRequest) (*openfgapb.WriteResponse, error) {

    store, err := cmd.datastore.GetStore(ctx, req.GetStoreId()) // step 1
    if err != nil {
            return nil, status.Errorf(codes.InvalidArgument, "the provided store_id does not exist")
        }
    
        return nil, err
    }

    // other code for validation etc..                         // step 2

   err = cmd.datastore.Write(....)                             // step 3
   if err != nil {
       // handle error
   }
}
```

If client 2 concurrently deletes the store between step 1 and step 2, then the Write at step 3 is erroneous. The set of rows satisfying the prior query condition (GetStore) may have changed due to another recently-committed transaction. This is known as a phantom read (from the perspective of the API client). For data consistency and API consistency reasons, API behavior should behave in a serializable manner in regards to the client. If a client is concurrently deleting a store while another client is concurrently writing to it, then whichever transaction commits first should win. Otherwise you get concurrency anomalies. 

See https://www.postgresql.org/docs/current/transaction-iso.html for more information, but we should be using an isolation level of at-least "repeatable-read" (the default in postgres), but that only applies to a transaction itself. If you call `datastore.GetStore` followed by `datastore.Write`, then those two queries aren't executed in the same transaction, and so a race on the previously read data can happen between step 1 and step 2. It's pretty common to see code written like this in projects, and it's always bothersome because it's not correct and/or transactionally accurate.

Himanshu372: I kind of agree on above @jon-whit . Do we have examples of similar `Transaction` in repo, otherwise I can add some module which can be reused. 
Cc @miparnisari

jon-whit: @Himanshu372, we don't at the moment. The current storage interface (and the way it is consumed) doesn't demand transaction support as part of the storage interface. This would be one of the first use cases of having such a signature added to the storage interface. Like I said above, we should just consider adding a `WrapTransaction` or something similar, and then we can implement that signature to hook into a txn across multiple datastore calls.

miparnisari: > If client 2 concurrently deletes the store between step 1 and step 2, then the Write at step 3 is erroneous. 

I feel like we are adding a big change to our storage interface to support one scenario that happens once (the deletion)

jon-whit: > I feel like we are proposing a big change to our storage interface to support just one scenario that happens once (the store deletion)

We wouldn't write code with data race conditions on shared memory (e.g. `go test -race ...`), so why would we compromise on design accuracy and consistency on data at the storage layer?

Himanshu372: @jon-whit @miparnisari  When doing store check for every API which has `store_id` as param, I do see a lot of diff. Adding `Transaction` logic here would be inflate the diff more and PR would be difficult of review. 
I can open up another issue to take that work forward, is it OK?

Himanshu372: @adriantam as you suggested I have used `GetStore` logic to validate store, plz let me know if it looks good to you

rhamzeh:
Hey @Himanshu372 - we've been discussing this internally, and we have a few thoughts.

We are thinking of approaching this in a different way: By introducing a foreign key where StoreID is used.

This includes:
- [ ] Creating a migration script that would create stores if they do not already exist for the distinct store_ids in the authorization models table, with a the store name being an empty string. An N+1 would be if we set the `created_at` and `updated_at` to the earliest relevant entry in the authorization model table.
- [ ] Create a migration script that would make all the store_ids across all the tables foreign keys pointing to the Stores table
- [ ] Catch the DB errors to expose our own errors
- [ ] Tests to ensure that the writes are not succeeding (these can be the same tests you wrote)

Our main reasoning for this is that doing that extra DB call can adversely affect latency, and it is something we can get for free (or as close to it) by using the DB. Later on, we can extend it to add more logic, if needed. But it will get us most of the way there.

Apologies for the long discussion about this topic, and for the sudden change in direction from where you were going with this PR. We're just trying to make sure not to introduce anything that would affect latency without a lot of consideration, and adding a foreign key constraint seems the simplest way for now.

What do you think about this? Would you be willing to do that change in this PR instead of the work you did?

Himanshu372:
@rhamzeh Hmm, so for latency, we are saying, we'll introduce FK constraints on tables themselves. I can take this up, but it depends on the priority/timeline of this task. 

I'll take a bit of time, as I may need to look around a bit. Let me know your thoughts. Thanks

curfew-marathon:
**Thank you @Himanshu372 for the contribution.  At this time we're closing this PR and will circle back when the work is prioritized.  We would like to collaborate on this as a wider initiative in the future.**

PR #711, opened 2023-04-21
fix: database calls for ListObjects v0 use a new context object to avoid db connection churning

miparnisari:
Fixes: 

- database calls for ListObjects v0 use a new context object to avoid database connection churning
- flaky test

### DIFF
--- a/pkg/storage/context.go
+++ b/pkg/storage/context.go
@@ -8,8 +8,11 @@ import (
 )
 
 // ContextTracerWrapper is a wrapper around a datastore that passes a new
-// context to the underlying datastore methods. It must be the first wrapper
-// around the datastore if traces are to work properly.
+// context to the underlying datastore methods.
+// This is so that if the context gets cancelled (e.g by the client), the underlying database connection isn't closed.
+// So, we let outstanding queries run their course even if the context gets cancelled to avoid database connection churning.
+//
+// ContextTracerWrapper must be the first wrapper around the datastore if traces are to work properly.
 type ContextTracerWrapper struct {
 	OpenFGADatastore
 }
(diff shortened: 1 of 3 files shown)


pkg/storage/context.go
```diff
@@ -8,8 +8,11 @@ import (
 )
 
 // ContextTracerWrapper is a wrapper around a datastore that passes a new
-// context to the underlying datastore methods. It must be the first wrapper
-// around the datastore if traces are to work properly.
+// context to the underlying datastore methods.
+// This is so that if the context gets cancelled (e.g by the client), the underlying database connection isn't closed.
+// So, we let outstanding queries run their course even if the context gets cancelled, so that we don't cause database connection churning.
```
adriantam, line 13:
```suggestion
// So, we let outstanding queries run their course even if the context gets cancelled to avoid database connection churning.
```

pkg/storage/storage.go
```diff
@@ -54,6 +54,7 @@ type RelationshipTupleReader interface {
 
 	// ReadPage is similar to Read, but with PaginationOptions. Instead of returning a TupleIterator, ReadPage
 	// returns a page of tuples and a possibly non-empty continuation token.
+	// There is NO guarantee on the order returned on the iterator.
```
adriantam, line 57: I don't get this - ReadPage does not have any iterator

adriantam (changes requested):

pkg/storage/storage.go
```diff
@@ -54,6 +54,7 @@ type RelationshipTupleReader interface {
 
 	// ReadPage is similar to Read, but with PaginationOptions. Instead of returning a TupleIterator, ReadPage
 	// returns a page of tuples and a possibly non-empty continuation token.
+	// There is NO guarantee on the order returned on the list.
```
adriantam, line 57: I thought there is guarantee order on the tuples for ReadPage?  Both Postgres and MySQL quries with order by ulid.  If there is no order on the tuples, then continuation_token will not work.

miparnisari: Ah you're right, my bad!

adriantam (approved):

Issue #724, opened 2023-05-03
Write internal Error when authorization model ID valid but not found


adriantam:
Write results in 500 internal error if request's auth model id is valid but not found.

PR #725, opened 2023-05-03
fix: in write API, if auth model ID not found, throw 400 instead of 500 error

adriantam:
## Description
Should respond with proper error message when auth model id is not found for write.

## References
Close https://github.com/openfga/openfga/issues/724

## Testing

1. Create store
2. Write auth model
3. Write tuples but specify auth model id that is a valid ulid but no such auth model id
4. Ensure 400 Bad Request is returned (not 500)

## Review Checklist

### DIFF
--- a/pkg/server/commands/write.go
+++ b/pkg/server/commands/write.go
@@ -65,6 +65,9 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 
 		authModel, err := c.datastore.ReadAuthorizationModel(ctx, store, modelID)
 		if err != nil {
+			if errors.Is(err, storage.ErrNotFound) {
+				return serverErrors.AuthorizationModelNotFound(modelID)
+			}
 			return err
 		}
 
--- a/pkg/server/test/write.go
+++ b/pkg/server/test/write.go
@@ -634,6 +634,47 @@ var writeCommandTests = []writeCommandTest{
 		),
 	},
 	{
+		_name: "ExecuteReturnsErrorIfAuthModelNotFound",
+		// state
+		model: &openfgapb.AuthorizationModel{
+			Id:            ulid.Make().String(),
+			SchemaVersion: typesystem.SchemaVersion1_0,
+			TypeDefinitions: []*openfgapb.TypeDefinition{
+				{
+					Type: "repo",
+					Relations: map[string]*openfgapb.Userset{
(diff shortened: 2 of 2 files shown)


pkg/server/commands/write.go
```diff
@@ -65,6 +65,9 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 
 		authModel, err := c.datastore.ReadAuthorizationModel(ctx, store, modelID)
 		if err != nil {
+			if errors.Is(err, storage.ErrNotFound) {
```
jon-whit, line 68: Though this addition is a good one, the overall logic contained herein seems a bit faulty.

It appears as though we attempt to read (and therefore validate the existence of) the authorization model `if len(writes) > 0`. But what if you don't specify any writes but you do specify some deletes? Should we not also verify the authorization model exists in that case?

It seems this code should change to:

```go
authModel, err := c.datastore.ReadAuthorizationModel(ctx, store, modelID)
if err != nil {
        return serverErrors.AuthorizationModelNotFound(modelID)
    }
}

if len(writes) > 0 {
    // do existing logic
}

for _, tk := range deletes {
    // do existing delete validation
}
```

jon-whit: Please add a test for this case as well ☝️

jon-whit (changes requested):
Left a comment to be addressed.

pkg/server/commands/write.go
adriantam: For writes, we need to do read authorization model for type validation purpose.  For delete, we do not care about auth model as we do not need to do type validation.

pkg/server/test/write.go
```diff
@@ -633,6 +633,47 @@ var writeCommandTests = []writeCommandTest{
 			},
 		),
 	},
+	{
+		_name: "ExecuteReturnsErrorIfAuthModelNotFound",
+		// state
+		model: &openfgapb.AuthorizationModel{
+			Id:            ulid.Make().String(),
+			SchemaVersion: typesystem.SchemaVersion1_0,
```
jon-whit, line 641: We're doing away with 1.0 support, so we should prioritize tests that default to 1.1. Can we switch this specific test case to be for 1.1 please?

adriantam: Done

miparnisari (approved):

jon-whit (approved):

PR #726, opened 2023-05-04
feat: less CPU consumption by typesystem class

miparnisari:
`typesystem.GetRelations` is used a lot, so instead of computing the relations on each invocation of this function, we compute it once in `typesystem.New` and then store it in memory for later reuse.

Results:

image


rhamzeh (approved):

pkg/typesystem/typesystem.go
```diff
@@ -126,23 +126,45 @@ func Difference(base *openfgapb.Userset, sub *openfgapb.Userset) *openfgapb.User
 }
 
 type TypeSystem struct {
+	// [objectType] => typeDefinition
 	typeDefinitions map[string]*openfgapb.TypeDefinition
-	modelID         string
-	schemaVersion   string
+	// [objectType] => [relationName] => relation
+	relations     map[string]map[string]*openfgapb.Relation
```
jon-whit, line 132: What I'd really hope to see in the near term is the `openfgapb.TypeDefinition` to change altogether. The protobuf definition for a TypeDefinition should be:

```
message TypeDefintion {
    string type = 1;
    map<string, Relation> relations = 2;
}
```

The current definition involving `map<string, Userset>` requires this extra construction AND it also necessitates the `metadata` as well.

jon-whit: Left an issue in the `openfga/api` repository to address this.

https://github.com/openfga/api/issues/77

pkg/typesystem/typesystem.go
```diff
@@ -126,23 +126,45 @@ func Difference(base *openfgapb.Userset, sub *openfgapb.Userset) *openfgapb.User
...
+		for relation, rewrite := range td.GetRelations() {
+			r := &openfgapb.Relation{
+				Name:     relation,
+				Rewrite:  rewrite,
+				TypeInfo: &openfgapb.RelationTypeInfo{},
+			}
+
+			if metadata, ok := td.GetMetadata().GetRelations()[relation]; ok {
+				r.TypeInfo.DirectlyRelatedUserTypes = metadata.GetDirectlyRelatedUserTypes()
+			}
```
adriantam, line 156: Will it ever by !ok?  Should we throw panic for these cases?

miparnisari: if it's a 1.0 model, line 155 should be ignored

PR #732, opened 2023-05-10
fix: ensure Check evaluates deterministically on the eval boundary case

jon-whit:
## Description
This PR ensures Check evaluations are deterministic when two concurrent evaluation paths are on the boundary of evaluating a path that would encounter the AuthorizationModelTooComplexError case and another concurrent path that would yield an `{allowed: true}` outcome.

## Review Checklist


internal/graph/check.go
```diff
@@ -197,6 +199,10 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 		}
 	}
 
+	if err != nil {
+		return &openfgapb.CheckResponse{Allowed: false}, err
+	}
+
 	return &openfgapb.CheckResponse{Allowed: false}, nil
```
adriantam, line 206:
```suggestion
	return &openfgapb.CheckResponse{Allowed: false}, err
```
The check for err != nil is redundant.

internal/graph/check.go
```diff
@@ -197,6 +199,10 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 		}
```
adriantam, line 199: Want to verify a behaviour change.  Let's say that there is an error (i.e., err != nil) and then we receive ctx.Done(), we will not return the err.  Instead, we will return ctx.Err().  Is this expected or we want to return err if ctx.Err() is nil.

pkg/testfixtures/storage/storage.go
```diff
@@ -52,3 +59,26 @@ func RunDatastoreTestContainer(t testing.TB, engine string) DatastoreTestContain
...
+
+	switch engine {
+	case "memory":
+		ds = memory.New(10, 24)
+	case "postgres":
+		ds, err = postgres.New(uri, sqlcommon.NewConfig())
+	case "mysql":
+		ds, err = mysql.New(uri, sqlcommon.NewConfig())
+	default:
+		t.Fatalf("'%s' is not a supported datastore engine", engine)
```
adriantam, line 79: Should we have BootstrapDatastore as part of datastore interface (or something like testdatastore interface)?  In this way, custom datastore does not need to modify this method.  Instead, we have compile time check rather than test/run-time check.

adriantam (changes requested):

internal/graph/check.go
jon-whit: ✅ done

pkg/testfixtures/storage/storage.go
jon-whit: It's a utility method only used in tests. I don't think it belongs on the interface, as it is not needed for any storage functionality other than to help bootstrap a datastore implementation in tests.

internal/graph/check.go
jon-whit: If we receive the context cancellation (and thus receive on the `ctx.Done()` channel), and if this happens first, then we'll return `ctx.Err()`. This behavior is expected, yes.

If we encounter a case where `err != nil`, and the context is not yet cancelled, then we'll surface that error after we've determined that there is no `allowed=true` outcome. If the context is cancelled while waiting for the `allowed=true` outcome, then the statement above ☝️ applies, because the context cancellation wins. 

In both cases it seems the behavior makes sense.

pkg/testfixtures/storage/storage.go
adriantam: If someone have another storage adaptor (i.e., dynamo for example), what do they need to do?  Do they update this file or override the functionality?  My concern here is that the textfixture/storage/storage.go is generic and should not have any knowledge of whether there is mysql/postgresql/memory etc.

jon-whit: It's not generic though. It has to have knowledge about the supported datastore engines in order to provide a `DatastoreTestContainer` to callers. See https://github.com/openfga/openfga/blob/main/pkg/testfixtures/storage/storage.go#L42

In fact, we also must provide implementations of the `DatastoreTestContainer` for each implementing storage engine.

https://github.com/openfga/openfga/tree/main/pkg/testfixtures/storage

This package is just meant to help with construction of a test container for testing purposes. In any case, I moved this function out and into the place directly where it is used.

internal/graph/check_test.go
```diff
@@ -0,0 +1,72 @@
...
+	parser "github.com/craigpastro/openfga-dsl-parser/v2"
+	"github.com/oklog/ulid/v2"
+	"github.com/openfga/openfga/pkg/storage/memory"
+	"github.com/openfga/openfga/pkg/tuple"
+	"github.com/openfga/openfga/pkg/typesystem"
+	"github.com/stretchr/testify/require"
+	openfgav1 "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+func TestResolveCheckDeterministic(t *testing.T) {
```
adriantam, line 16: Do we need a test for union (the test right now is intersection)

jon-whit: Done. Added a condition on `define viewer: [group#member] as self or editor` and added some tuples that generate a model resolution too complex path on `document:1#editor`, and yet the test still passes with the expected outcome like it should. So this proves union is also deterministic in regards to an `{allowed: true}` outcome superseding a resolution error.

adriantam (approved):

adriantam (approved):

PR #734, opened 2023-05-11
chore: prune `tools/` dependencies

miparnisari:
This PR should help reduce the number of security advisories that we get for the dependencies declared in `tools/go.mod`, which are development dependencies.

- `govulncheck` is used when submitting a PR and run by Github Actions, there is no need for developers to run this locally, therefore i am removing it.
- `golangci-lint` in Github Actions runs with the latest version and locally it used to run with a pinned version. In this PR i'm changing the `make lint` command so that it always installs the latest version prior to running, and i remove it from `tools/go.mod`.


adriantam (changes requested):
Please update in .github/workflows/pull_request.yaml for downloading govulncheck

adriantam:
Please also update description why we want to make this PR

jon-whit (dismissed):

adriantam (approved):

PR #756, opened 2023-05-12
refactor: rewrite Typesystem without context propagation

jon-whit:
## Description
Refactors the way the Typesystem is resolved and passed around. Instead of passing through context we explicitly propagate it via a TypesystemResolverFunc that resolves it in a memoized way.

## References
https://github.com/openfga/openfga/issues/616

## Review Checklist


pkg/typesystem/resolver_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	parser "github.com/craigpastro/openfga-dsl-parser/v2"
+	"github.com/golang/mock/gomock"
+	"github.com/oklog/ulid/v2"
+	mockstorage "github.com/openfga/openfga/pkg/storage/mocks"
+	"github.com/stretchr/testify/require"
+	openfgav1 "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+func TestMemoizedTypesystemResolverFunc(t *testing.T) {
```
jon-whit, line 15: todo(@jon-whit) - You should also add one more test for the case where the authorization model identifier is `""`.

pkg/server/commands/connected_objects.go
```diff
@@ -98,11 +98,14 @@ type ConnectedObjectsCommand struct {
 
 	// Limit limits the results yielded by the ConnectedObjects API.
 	Limit uint32
+
+	TypesystemResolver typesystem.TypesystemResolverFunc
 }
 
 func (c *ConnectedObjectsCommand) streamedConnectedObjects(
```
jon-whit, line 105: Function signature is starting to get pretty bloated. Will consider some ways to make this a bit more elegant, but this works at least for the objective at hand.

pkg/server/queries.go
jon-whit, line 1: I added the `queries.go` file to separate out the Relationship Query RPC handlers.

https://openfga.dev/api/service#/Relationship%20Queries

jon-whit:
Looks like something broke in the tests. I'll investigate next week 👍

pkg/server/commands/connected_objects.go
miparnisari: Why can't `typesys` be kept as part of `ConnectedObjectsRequest` and `reverseExpandRequest`?

pkg/server/commands/list_objects.go
```diff
@@ -48,6 +50,7 @@ type listObjectsRequest interface {
 func (q *ListObjectsQuery) evaluate(
 	ctx context.Context,
 	req listObjectsRequest,
+	typesys *typesystem.TypeSystem,
```
miparnisari, line 53: Similar comment as above. `typesys` is a concrete instance of a model ID. It can be part of `listObjectsRequest`

miparnisari (changes requested):

pkg/server/queries.go
miparnisari: Please don't move code around in this PR

pkg/server/commands/connected_objects.go
jon-whit: Our pattern is clear - `<something>Request` structs are for user provided inputs. The typesys is an structure constructed internally during the request lifecycle. The suggestion of embedding this in the Request input is non-standard to our codebase. You could argue that every field of the `ConnectedObjectsCommand` could/should be part of the `ConnectedObjectsRequest` as well for the same reasons you mention, but yet we don't do that.

A `ConnectedObjectsCommand` represents a specific instance of a struct that is constructed to handle a single request to an endpoint.

If we're going to do this, then we need a larger refactor for consistency. This pattern aligns with the commands/queries structs and the patterns we are using with them as they are designed today.

pkg/server/commands/list_objects.go
jon-whit: See https://github.com/openfga/openfga/pull/756#discussion_r1199491826

internal/graph/check.go
```diff
@@ -371,9 +378,9 @@ func (c *LocalChecker) ResolveCheck(
 func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
 
 	return func(ctx context.Context) (*openfgapb.CheckResponse, error) {
-		typesys, ok := typesystem.TypesystemFromContext(parentctx) // note: use of 'parentctx' not 'ctx' - this is important
-		if !ok {
-			return nil, fmt.Errorf("typesystem missing in context")
+		typesys, err := c.typesysResolver(ctx, req.GetStoreID(), req.GetAuthorizationModelID())
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve typesystem: %v", err)
```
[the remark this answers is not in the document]
jon-whit: It's part of the `typesystem` package though, and that's an internal bit. The server (and by extension commands/queries) layer should be responsible for user facing error messages, since those are the code bits that handle the public API itself.

pkg/server/queries.go
```diff
@@ -0,0 +1,198 @@
...
+		attribute.KeyValue{Key: "user", Value: attribute.StringValue(tk.GetUser())},
+	))
+	defer span.End()
+
+	storeID := req.GetStoreId()
+
+	modelID, err := s.resolveAuthorizationModelID(ctx, storeID, req.GetAuthorizationModelId())
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
jon-whit: Yeah, that's residual. Good catch. That should go away.

pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,74 @@
...
+// the resolved model. The type-system resolution is memoized so if another lookup of the same model occurs,
+// then the earlier TypeSystem that was constructed will be used.
+//
+// The memoized resolver function is safe for concurrent use.
+func MemoizedTypesystemResolverFunc(reader storage.AuthorizationModelReadBackend) TypesystemResolverFunc {
+
+	cache := ccache.New(ccache.Configure[*TypeSystem]())
+
+	return func(ctx context.Context, storeID, modelID string) (*TypeSystem, error) {
+		tracer.Start(ctx, "MemoizedTypesystemResolverFunc")
```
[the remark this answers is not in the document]
jon-whit: It should be. Just missing some code.

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: It's the same reason why the `internal/graph` package doesn't return user facing errors (e.g. errors from the `pkg/server/errors` package).

pkg/server/queries.go
jon-whit: Ok, I'd be happy to move it back. But the reason I did it is that it separates the code a bit more, which makes it a little more manageable, and that helped me manage the changes.

jon-whit: Ok, moved back into their original spot for now. We can move those into another file in a subsequent PR.

pkg/server/commands/connected_objects.go
miparnisari: > Our pattern is clear - <something>Request structs are for user provided inputs.

This is fine for the requests defined in `api`. But `ResolveCheckRequest`, `ConnectedObjectsRequest` and `reverseExpandRequest` are not defined there, we have more flexibility for these, I think.

> The suggestion of embedding this in the Request input is non-standard to our codebase.

With this specific change you're undoing the change here https://github.com/openfga/openfga/pull/650 and adding more work to yourself. Wouldn't the code work as is today?

jon-whit: > This is fine for the requests defined in api. But ResolveCheckRequest, ConnectedObjectsRequest and reverseExpandRequest are not defined there, we have more flexibility for these, I think.

We are targeting a public API contract for the ConnectedObjects API. Even if we weren't, then it still is a good practice to keep things consistent. Whether an internal RPC or an external RPC it's good to keep consistent practices/semantics between public facing RPC definitions and internal facing ones. 

> With this specific change you're undoing the change here https://github.com/openfga/openfga/pull/650 and adding more work to yourself. Wouldn't the code work as is today?

#650 introduced the inconsistency. All command/query structs are constructed per request, and all other command/query structs define their parameters per struct, not per method. The change, albeit super small and hardly more work for myself, establishes more consistency within the scope of this work.

rhamzeh: (sidenote: if the scope of this discussion is not blocking this PR, let's resolve any other issues, merge and continue the discussion after)

Re: ConnectedObjects
Yes, we do plan to make it public. We need it in order to build a local search index without the user doing all the work.
However:
1) Not sure if `ConnectedObjects` will be the name, that was the name used internally in the API, but it may not be the best name for external usage
2) We need to discuss the interface
As in whatever gets exposed may or may not be this specific method, it will probably be a wrapper on top of it as we debate the API interface. So I wouldn't pre-optimize for that, although it is good to keep in mind that this will end up being exposed in some version or another.
For now, please do not add it to the proto files unless you hide it similar to how we did the UpdateStore, otherwise this will end up being reflected in all SDKs and docs as soon as it is merged.

My thinking is that this should be similar to all the internal methods that CheckRequest and similar wrap, it probably should not have the same inputs (middleware, etc..) that the API entry points do.
Ideally, the external interface would be consistent with the interfaces for the external methods, and the internal one would be consistent with the internal interfaces.

jon-whit:
Closing until we can get around to this at a later time.

PR #757, opened 2023-05-12
docs: if OPENFGA_LIST_OBJECTS_MAX_RESULTS=0, no limit

miparnisari:
## Description
Fixes for ListObjects:

- Undo a change made in https://github.com/openfga/openfga/pull/733 - if `ListObjectsMaxResults` is 0, we make a results channel of size 1, not 0.
- Fix goroutine leak that can happen if the main goroutine exits and the child goroutines attempt to write an error, previously they became blocked in [this](https://github.com/openfga/openfga/blob/2f3989def137634ac5009f93e94e91455853f6d3/pkg/server/commands/list_objects.go#L95) line or [this](https://github.com/openfga/openfga/blob/2f3989def137634ac5009f93e94e91455853f6d3/pkg/server/commands/list_objects.go#L159) line. (This is verifiable through https://github.com/uber-go/goleak)
- Document that if `ListObjectsMaxResults` is 0, there is no limit in the number of results returned

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #762, opened 2023-05-16
fix: regression in #757

miparnisari:
This fixes a regression introduced in https://github.com/openfga/openfga/pull/757. The failure is seen here https://github.com/openfga/openfga/actions/runs/4994450765/jobs/8945024934#step:5:38. Note that this failure happens once in a while, not all the time :)

The following benchmarks (not included in the PR) are minimum reproducible examples of the overall behavior before that PR, after that PR, and with this PR fixing it.

```go
// passes
func BenchmarkBeforePR757(b *testing.B) {
	for i := 0; i < b.N; i++ {
		errChan := make(chan error) //unbuffered
		resultsChan := make(chan error, 1000)
		go func() {
			errChan <- fmt.Errorf("error!")
			close(resultsChan)
		}()

		for {
			select {
			case _, ok := <-resultsChan:
				if !ok {
					panic("not expected")
				}
			case _, ok := <-errChan:
				if ok {
					return
				}
			}
		}
	}
}

// panics after some time
func BenchmarkAfterPR757(b *testing.B) {
	for i := 0; i < b.N; i++ {
		errChan := make(chan error, 1) //buffered, this is new
		resultsChan := make(chan error, 1000)
		go func() {
			errChan <- fmt.Errorf("error!")
			close(errChan) // this is new!
			close(resultsChan) 
		}()

		for {
			select {
                         // If multiple cases are ready, Go picks one randomly
			case _, ok := <-resultsChan:
				if !ok {
					panic("not expected")
				}
			case _, ok := <-errChan:
				if ok {
					return
				}
			}
		}
	}
}

// passes
func BenchmarkAfterPR762(b *testing.B) {
	for i := 0; i < b.N; i++ {
		resultsChan := make(chan error, 1000)
		go func() {
			errChan <- fmt.Errorf("error!")
			close(errChan)
		}()

		for {
			select {
			case _, ok := <-resultsChan:
				if !ok {
					panic("not expected")
				}
			case _, ok := <-errChan:
				if ok {
					return
				}
			}
		}
	}
}
```

💡 As a follow-up, instead of having `errChan` and `resultsChan`, we may consider having just one channel that can store both errors and results.

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -212,15 +212,15 @@ func (q *ListObjectsQuery) Execute(
 				Objects: objects,
 			}, nil
 
-		case objectID, ok := <-resultsChan:
-			if !ok { //channel was closed and no new values are available
+		case objectID, channelOpen := <-resultsChan:
+			if !channelOpen {
 				return &openfgapb.ListObjectsResponse{
 					Objects: objects,
 				}, nil
 			}
 			objects = append(objects, objectID)
-		case genericError, ok := <-errChan:
-			if ok { //a value is available
+		case genericError, valueAvailable := <-errChan:
+			if valueAvailable {
 				return nil, serverErrors.NewInternalError("", genericError)
 			}
 		}
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -94,9 +94,9 @@ func (q *ListObjectsQuery) evaluate(
 		if err != nil {
 			errChan <- err
 			close(errChan)
+		} else {
+			close(resultsChan)
 		}
-
-		close(resultsChan)
 	}
 
 	hasTypeInfo, err := typesys.HasTypeInfo(targetObjectType, targetRelation)
@@ -159,9 +159,9 @@ func (q *ListObjectsQuery) evaluate(
 			if err != nil {
 				errChan <- err
 				close(errChan)
+			} else {
+				close(resultsChan)
 			}
-
-			close(resultsChan)
 		}
 	}
 
@@ -213,14 +213,14 @@ func (q *ListObjectsQuery) Execute(
 			}, nil
 
 		case objectID, ok := <-resultsChan:
-			if !ok {
+			if !ok { //channel was closed and no new values are available
 				return &openfgapb.ListObjectsResponse{
 					Objects: objects,
 				}, nil
 			}
 			objects = append(objects, objectID)
 		case genericError, ok := <-errChan:
-			if ok {
+			if ok { //a value is available
 				return nil, serverErrors.NewInternalError("", genericError)
 			}
 		}
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -94,9 +94,9 @@ func (q *ListObjectsQuery) evaluate(
 		if err != nil {
 			errChan <- err
 			close(errChan)
-		} else {
-			close(resultsChan)
+			return
 		}
+		close(resultsChan)
 	}
 
 	hasTypeInfo, err := typesys.HasTypeInfo(targetObjectType, targetRelation)
@@ -159,9 +159,9 @@ func (q *ListObjectsQuery) evaluate(
 			if err != nil {
 				errChan <- err
 				close(errChan)
-			} else {
-				close(resultsChan)
+				return
 			}
+			close(resultsChan)
 		}
 	}
 
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 	"time"
 
+	parser "github.com/craigpastro/openfga-dsl-parser/v2"
 	"github.com/golang/mock/gomock"
 	"github.com/oklog/ulid/v2"
 	"github.com/openfga/openfga/internal/gateway"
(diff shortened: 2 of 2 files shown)


rhamzeh (approved):

rhamzeh:
Thanks for the steps! Can we add that as a test case so that we don't stumble upon this again?

pkg/server/commands/list_objects.go
```diff
@@ -213,14 +213,14 @@ func (q *ListObjectsQuery) Execute(
 			}, nil
 
 		case objectID, ok := <-resultsChan:
-			if !ok {
+			if !ok { //channel was closed and no new values are available
```
jon-whit, line 216: This comment and the one below seem redundant. This is the behavior of the language. It should go without documentation.

pkg/server/commands/list_objects.go
```diff
@@ -94,9 +94,9 @@ func (q *ListObjectsQuery) evaluate(
 		if err != nil {
 			errChan <- err
 			close(errChan)
+		} else {
+			close(resultsChan)
```
jon-whit, line 98: The `else` is redundant.

Instead

```go
if err != nil {
    errChan <- err
    return
}

close(resultsChan)
```

`StreamedConnectedObjects` is synchronous, but produces asynchronously on the `resultsChan`. So if we see an error, then just yield it and return.

pkg/server/commands/list_objects.go
miparnisari: I understand that it's part of the language but it's not super obvious to me. I don't think it hurts to have it

jon-whit: How about a different variable name then?

`objectID, closed := <-resultsChan`

pkg/server/commands/list_objects.go
miparnisari: In your snippet you don't have the line `close(errChan)`. Is this expected? And if yes, how is it better than the way I wrote it? To be clear, from my testing, both approaches seem to work just fine.

jon-whit: Sorry, I omitted the `close(errChan)` bit. That was supposed to be there. I've updated the snippet.

> And if yes, how is it better than the way I wrote it?

Less nesting == more readability. It's just a style thing mostly. In our code we see more

```go
if err != nil {
    // do something
    return
}

// do something else
```

more than we see

```go
if err != nil {
    // do something
} else {
   // do something else
}
```

And yeah, they are functionally equivalent. So it's more a stylistic pattern matching thing.

pkg/server/server_test.go
```diff
@@ -347,6 +347,68 @@ func (m *mockStreamServer) Send(*openfgapb.StreamedListObjectsResponse) error {
...
+	ctx := context.Background()
+	logger := logger.NewNoopLogger()
+	transport := gateway.NewNoopTransport()
+	store := ulid.Make().String()
+	modelID := ulid.Make().String()
+
+	mockController := gomock.NewController(b)
+	defer mockController.Finish()
+
+	data, err := os.ReadFile("testdata/github.json")
```
jon-whit, line 361: There's not a strong reason to use the github model here given that we're only testing `repo#owner` which is just a direct relationship with an `org` type.

I vote we drop this simply in favor of

```go
typedefs := parser.MustParse(`
    type org

    type repo
        relations
            define owner: [org] as self
    `)
    
...

// use typedefs below
```

pkg/server/server_test.go
```diff
@@ -347,6 +347,68 @@ func (m *mockStreamServer) Send(*openfgapb.StreamedListObjectsResponse) error {
...
+	require.NoError(b, err)
+
+	var gitHubTypeDefinitions openfgapb.WriteAuthorizationModelRequest
+	err = protojson.Unmarshal(data, &gitHubTypeDefinitions)
+	require.NoError(b, err)
+
+	mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)
+
+	mockDatastore.EXPECT().ReadAuthorizationModel(gomock.Any(), store, modelID).AnyTimes().Return(&openfgapb.AuthorizationModel{
+		SchemaVersion:   typesystem.SchemaVersion1_0,
```
jon-whit, line 371: Let's make this a 1.1 model please, because we're doing away with 1.0.

pkg/server/server_test.go
```diff
@@ -347,6 +347,68 @@ func (m *mockStreamServer) Send(*openfgapb.StreamedListObjectsResponse) error {
...
+	})
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, err := s.ListObjects(ctx, &openfgapb.ListObjectsRequest{
+			StoreId:              store,
+			AuthorizationModelId: modelID,
+			Type:                 "repo",
+			Relation:             "owner",
+			User:                 "bob",
```
jon-whit, line 394:
```suggestion
			User:                 "user:bob",
```

pkg/server/server_test.go
```diff
@@ -347,6 +347,68 @@ func (m *mockStreamServer) Send(*openfgapb.StreamedListObjectsResponse) error {
...
+		})
+
+		require.ErrorIs(b, err, serverErrors.NewInternalError("", errors.New("error reading from storage")))
+
+		err = s.StreamedListObjects(&openfgapb.StreamedListObjectsRequest{
+			StoreId:              store,
+			AuthorizationModelId: modelID,
+			Type:                 "repo",
+			Relation:             "owner",
+			User:                 "bob",
```
jon-whit, line 404:
```suggestion
			User:                 "user:bob",
```

pkg/server/server_test.go
jon-whit: The motivation behind this comment is because of a similar change taking place here: https://github.com/openfga/openfga/pull/763/files#diff-fe7a244e18e92d8f347af17363209172e1d79224f086b0585a1bbeef3b626c4eL360

We're using the Github model in some cases where we honestly don't need to be. Simpler is better, and in this case we can drop the dependency on the `.json` file for a simple direct relationship. Makes the test a tad cleaner and easier to manage.

pkg/server/server_test.go
miparnisari: ok, but it will have to have exclusion to force the v0 path :)

jon-whit (approved):
LGTM

PR #763, opened 2023-05-16
chore: drop 1.0 model support officially

jon-whit:
## Description
Officially drops support for 1.0 models. This removes the code that allows developers to override the 1.0 model support behavior in OpenFGA. After we merge this developers will no longer be able to `Write` or evaluate (e.g. `Check`, `ListObjects`, `Expand`) 1.0 models.

## Review Checklist


assets/tests/consolidated_1_0_tests.yaml
jon-whit, line 1: Drop 1.0 related tests.

pkg/server/commands/list_objects.go
```diff
@@ -94,6 +94,7 @@ func (q *ListObjectsQuery) evaluate(
 		if err != nil {
 			errChan <- err
 			close(errChan)
+			return
```
jon-whit, line 97: Related to https://github.com/openfga/openfga/pull/762/files

pkg/server/commands/write_test.go
```diff
@@ -142,10 +143,12 @@ func TestValidateWriteRequest(t *testing.T) {
 			maxTuplesInWriteOp := 10
 			mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)
 			mockDatastore.EXPECT().MaxTuplesPerWrite().AnyTimes().Return(maxTuplesInWriteOp)
-			cmd := NewWriteCommand(mockDatastore, logger, true)
+			cmd := NewWriteCommand(mockDatastore, logger)
 
 			if len(test.writes) > 0 {
-				mockDatastore.EXPECT().ReadAuthorizationModel(gomock.Any(), gomock.Any(), gomock.Any()).Return(&openfgapb.AuthorizationModel{}, nil)
+				mockDatastore.EXPECT().ReadAuthorizationModel(gomock.Any(), gomock.Any(), gomock.Any()).Return(&openfgapb.AuthorizationModel{
+					SchemaVersion: typesystem.SchemaVersion1_1,
```
jon-whit, line 150: Given the new behavior, this change was particularly relevant because we return a `ValidationError` otherwise.

pkg/typesystem/typesystem_test.go
```diff
@@ -495,40 +461,18 @@ func TestInvalidRewriteValidations(t *testing.T) {
...
+				type folder
+				  relations
+				    define parent: [folder] as self
+					define viewer as viewer from parent
+				`),
 			},
 			err: ErrCycle,
 		},
-		{
-			name: "Succeeds_If_Auth_Model_1.0_Has_A_Cycle_But_There_Is_More_Than_One_Type",
```
jon-whit, line 514: This test and the removed above are no longer relevant/possible under a 1.1 model paradigm.

rhamzeh (approved):
Glad to see at least one TODO done :)

pkg/server/test/read.go
```diff
@@ -82,54 +86,32 @@ func ReadQuerySuccessTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
-					},
-				},
+				SchemaVersion: typesystem.SchemaVersion1_1,
+				TypeDefinitions: parser.MustParse(`
+				type user
+
+				type repo
+				  relations
+				    define admin: [user] as self
+					define owner: [user] as self
```
miparnisari, line 96: nit: indent

pkg/server/test/write.go
```diff
@@ -75,278 +78,183 @@ var writeCommandTests = []writeCommandTest{
...
-
 		// output
-		err: serverErrors.WriteToIndirectRelationError("Attempting to write directly to an indirect only relationship", &openfgapb.TupleKey{
-			Object:   "repository:openfga/openfga",
-			Relation: "viewer",
-			User:     "github|alice@openfga.com",
-		}),
+		err: serverErrors.ValidationError(
+			&tuple.InvalidTupleError{
+				Cause:    fmt.Errorf("type 'user' is not an allowed type restriction for 'repo#viewer'"),
```
miparnisari, line 103: I'm looking at https://app.codecov.io/gh/openfga/openfga/pull/763/indirect-changes and we aren't hitting this code path:  https://github.com/openfga/openfga/blob/0f1a5d06073efe0fbd33823cda56a6202864f816/pkg/server/commands/write.go#L102-L104 anymore.

I feel like `"Attempting to write directly to an indirect only relationship"` is clearer than `"type 'user' is not an allowed type restriction for 'repo#viewer'"` . `repo#viewer` doesn't support writing, doesn't matter which type you try.

Any chance of bringing back the previous error result?

pkg/server/commands/write_assertions.go
```diff
@@ -46,8 +43,8 @@ func (w *WriteAssertionsCommand) Execute(ctx context.Context, req *openfgapb.Wri
 
 	typesys := typesystem.New(model)
 
-	if ProhibitModel1_0(typesys.GetSchemaVersion(), w.allowSchema10) {
-		return nil, serverErrors.ValidationError(ErrObsoleteAuthorizationModel)
+	if typesys.GetSchemaVersion() != typesystem.SchemaVersion1_1 {
+		return nil, serverErrors.ValidationError(typesystem.ErrInvalidSchemaVersion)
```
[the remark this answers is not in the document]
jon-whit: It should change. Good catch.

pkg/typesystem/typesystem.go
```diff
@@ -715,7 +724,7 @@ func (t *TypeSystem) isUsersetRewriteValid(objectType, relation string, rewrite
 				}
 			}
 
-			return fmt.Errorf("%s does not appear as a relation in any of the directly related user types %v", computedUserset, userTypes)
+			return errors.Join(ErrRelationUndefined, fmt.Errorf("%s does not appear as a relation in any of the directly related user types %v", computedUserset, userTypes))
 		} else {
 			// for 1.0 models, relation `computedUserset` has to be defined _somewhere_ in the model
 			for typeName := range t.relations {
```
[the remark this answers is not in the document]
jon-whit: Yep, that's coming in the cleanup PR following this one.

pkg/server/server_test.go
```diff
@@ -423,18 +422,18 @@ func TestListObjects_Unoptimized_UnhappyPaths(t *testing.T) {
...
-		SchemaVersion:   typesystem.SchemaVersion1_0,
-		TypeDefinitions: gitHubTypeDefinitions.GetTypeDefinitions(),
+		SchemaVersion: typesystem.SchemaVersion1_1,
+		TypeDefinitions: parser.MustParse(`
+		type user
+
+		type repo
+		  relations
+		    define allowed: [user] as self
+		    define viewer: [user] as self and allowed
```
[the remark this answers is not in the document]
jon-whit: Done 👍

pkg/server/test/write.go
jon-whit: I go both ways, but I see where you're coming from.

To me the type restrictions are the set of allowed types that the relation can be related to, and a set of allowed types that is greater than 1 implies the relationship is, in some way, related to itself (e.g. `as self` shows up somewhere in the rewrite rule) and therefore is assignable.

If there are no type restrictions specified (e.g. `directly_related_user_types: []`), then that implies there are no types allowed to be written with that relation and therefore is unassignable. We can call this an indirect relationship or a purely rewritten relationship (I like to call it an unasignable relation), but I think for the sake of uniformity and to reduce the variation of errors it makes sense to leave it as is and remove this code bit.

Maybe a change of the error message itself would help?

Go from
> type 'user' is not an allowed type restriction for 'repo#viewer'

to
> object's of type 'user' are not an allowed relationship with the relation 'repo#viewer'

pkg/server/commands/write.go
```diff
@@ -71,12 +69,12 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgapb.
 			return err
 		}
 
-		typesys := typesystem.New(authModel)
-
-		if ProhibitModel1_0(typesys.GetSchemaVersion(), c.allowSchema10) {
-			return serverErrors.ValidationError(ErrObsoleteAuthorizationModel)
+		if !typesystem.IsSchemaVersionSupported(authModel.GetSchemaVersion()) {
```
adriantam, line 72: One part of me is thinking why don't we put this check inside typesystem.New?

miparnisari: We can revisit this in a follow-up PR, but `typesystem.New` doesn't return an error. It assumes that the model is valid.

pkg/server/test/write.go
miparnisari: Your suggestion still has the issue that me as a consumer, if i get the error `object's of type 'user' are not an allowed relationship with the relation 'repo#viewer'` then i will tend to think "what if I try writing an object of type "group" or "team"?... but i'll get a similar error. 

`"Attempting to write directly to an indirect only relationship"` implies that _no_ type is allowed

pkg/server/commands/write.go
jon-whit: `typesystem.NewAndValidate` takes the model, validates it, and returns either an error (if invalid) or the typesystem constructed from the model (if no error).

So yeah, I think having this as is right now makes the most sense. Those were by design.

pkg/server/test/write.go
jon-whit: `group` or `team` are not valid type restrictions just like `user` is not. If it doesn't show up in the type restriction list (or if the list is empty), then are no valid relationships that are valid targets to be written.

miparnisari: let's resurrect in a future PR if we want to

miparnisari (approved):

PR #770, opened 2023-05-19
fix: forbid 1.0 models in streaming list objects

miparnisari:
## Description
Forbid 1.0 models in streaming list objects

## References
This is a follow-up of https://github.com/openfga/openfga/pull/763


rhamzeh (approved):
Do we already have a test for the non-streamed ListObjects? If not, can we add one?

miparnisari:
> Do we already have a test for the non-streamed ListObjects? If not, can we add one?

Yep: https://github.com/openfga/openfga/blob/a0d9381dc10934889222619e9b30c0745eab24db/pkg/server/server_test.go#L609-L621

rhamzeh (approved):

adriantam (approved):

Issue #801, opened 2023-06-02
Models with self-referencing type restrictions and non-empty entrypoints should be valid


jon-whit:
I get a validation error from the `WriteAuthorizationModel` endpoint when trying to write the model:

```
type user

type document
  relations
    define editor: [user]
    define viewer: [document#viewer] or editor
```
Consider the following tuples:
```
document:1#viewer@document:2#viewer
document:2#editor@user:jon
```
`Check(document:1#viewer@user:jon)` should return `{allowed: true}`, because `user:jon` is an editor of `document:2`.

The model above should be valid becuase there is _at least one_ entry point that can lead to a possible evaluation path.

jon-whit:
https://github.com/openfga/openfga/blob/main/pkg/typesystem/typesystem.go#L784-L786 is the offending code.

The code block between L784-L786 needs to recurse in some form to see if there is at least 1 or more entrypoints/ingresses that could lead to an evaluation path.

PR #802, opened 2023-06-02
fix: validate models with non-zero entrypoints properly

jon-whit:
## Description
This fixes a bug with model validation that wasn't allowing models with non-zero entrypoints to be written to OpenFGA.

## References
Fixes the issue reported in #801.

## Review Checklist

### DIFF
--- a/pkg/server/test/write_authzmodel.go
+++ b/pkg/server/test/write_authzmodel.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"testing"
 
+	parser "github.com/craigpastro/openfga-dsl-parser/v2"
 	"github.com/oklog/ulid/v2"
 	"github.com/openfga/openfga/pkg/logger"
 	"github.com/openfga/openfga/pkg/server/commands"
@@ -133,6 +134,21 @@ func WriteAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastor
 						},
 					},
 				},
+			},
+		},
+		{
+			name: "self_referencing_type_restriction_with_entrypoint",
+			request: &openfgapb.WriteAuthorizationModelRequest{
+				StoreId: storeID,
+				TypeDefinitions: parser.MustParse(`
+				type user
+
+				type document
+				  relations
+				    define editor: [user] as self
+				    define viewer: [document#viewer] as self or editor
+				`),
+				SchemaVersion: typesystem.SchemaVersion1_1,
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


pkg/server/test/write_authzmodel.go
```diff
@@ -135,6 +136,21 @@ func WriteAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastor
 				},
 			},
 		},
+		{
+			name: "self_referencing_type_restriction_with_entrypoint",
```
miparnisari, line 140: why do we need this test here? isn't the test in `typesystem` enough? 🤔

jon-whit: If the behavior of the WriteAuthorizationModel implementation changes relative to the typesystem (e.g. if we remove the typesystem usage), then WriteAuthorizationModel should still behave the same.

pkg/server/test/write_authzmodel.go
```diff
@@ -135,6 +136,21 @@ func WriteAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastor
...
+			request: &openfgapb.WriteAuthorizationModelRequest{
+				StoreId: storeID,
+				TypeDefinitions: parser.MustParse(`
```
rhamzeh, line 143: Can we add a test for a case like this (it must fail validation):
```py
model
  schema 1.1
type user
type document
  relations
    define editor: [user] as self
    define viewer: [document#viewer] and editor
```

pkg/typesystem/typesystem.go
```diff
@@ -781,7 +781,9 @@ func (t *TypeSystem) validateRelationTypeRestrictions() error {
 			if assignable && len(relatedTypes) == 1 {
 				relatedObjectType := relatedTypes[0].GetType()
 				relatedRelation := relatedTypes[0].GetRelation()
-				if objectType == relatedObjectType && name == relatedRelation {
+
+				if objectType == relatedObjectType && name == relatedRelation &&
+					reflect.TypeOf(relation.GetRewrite().Userset) == reflect.TypeOf(&openfgapb.Userset_This{}) {
```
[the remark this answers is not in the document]
jon-whit: We do not, because the relation is loaded dynamically at runtime. We lookup the relation and based on its rewrite's type (which is determined at runtime) we do some validation. So this behavior is at runtime, not statically at compile time.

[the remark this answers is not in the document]
adriantam: As discussed for now, we can proceed like this for now.

pkg/server/test/write_authzmodel.go
jon-whit: This test case introduces some significant complexities that will need to be addressed. I'm going to create an issue to circle back to this test case and makes sure it gets addressed.

jon-whit: Good test! It exposed a little bug with the logic. I've fixed it and added this test as well as a similar one for a model involving exclusion (e.g `but not`).

rhamzeh (approved):

adriantam (approved):

Issue #681, opened 2023-04-06
"datastore.uri" and "datastore.engine" should be required fields when doing `run` and `migrate`


miparnisari:
See https://github.com/openfga/openfga/pull/655

Himanshu372:
Have a PR in draft for the above - https://github.com/openfga/openfga/pull/712

jon-whit:
https://github.com/openfga/openfga/pull/810 supersedes #712

PR #712, opened 2023-04-22
Adding code to verify datastore engine according to URI

Himanshu372:
Adding code to verify datastore engine is in line with URI provided in configurations

### Added 
1. Added a `DatastoreEngine` type which takes up const permissible engine type values
2. Added helper methods for the type `DatastoreEngine`
3. Added `verifyDatastoreEngine` func to handle verification of engine in line with URI 
4. Added tests 

## References

Addresses issue - https://github.com/openfga/openfga/issues/681

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #810, opened 2023-06-11
Validate datastore engine

Himanshu372:
### Added 
1. Added a `DatastoreEngine` type which takes up const permissible engine type values
2. Added helper methods for the type `DatastoreEngine`
3. Added checks to validate engine AFTER it is marshalled into config type
4. Added tests 

## References
Addresses comment - https://github.com/openfga/openfga/pull/712#issuecomment-1574015631
Closes https://github.com/openfga/openfga/issues/681

## Review Checklist


Himanshu372:
Hi @jon-whit, based on our last discussion, raised this PR to verify Datastore engine. Please have a look. Thanks

cmd/shared.go
jon-whit, line 1: Mind renaming to `cmd/cmd.go`?

cmd/shared.go
```diff
@@ -0,0 +1,40 @@
...
+
+// String converts DatastoreEngine e to string
+func (e DatastoreEngine) String() string {
+	return string(e)
+}
+
+// IsValid checks if DatastoreEngine e is valid
+func (e DatastoreEngine) IsValid() bool {
+	var isValid bool
+	for _, engineType := range []DatastoreEngine{Memory, Postgres, MySQL} {
```
jon-whit, line 25: This code can change to simply:

```go
for _, engine := range []DatastoreEngine{Memory, Postgres, MySQL} {
    if engine.String() == e.String() {
        return true
    }
}

return false
```

The variable assignment and then return is redundant.

cmd/shared.go
```diff
@@ -0,0 +1,40 @@
...
+		if engineType.String() == e.String() {
+			isValid = true
+		}
+	}
+	return isValid
+}
+
+// NewDatastoreEngine inits a new valid DatastoreEngine type
+func NewDatastoreEngine(engine string) (*DatastoreEngine, error) {
```
jon-whit, line 34: Why return a pointer instead of a value type here?

cmd/run/run_test.go
```diff
@@ -283,6 +284,23 @@ func TestVerifyConfig(t *testing.T) {
 		err := VerifyConfig(cfg)
 		require.Error(t, err)
 	})
+
+	t.Run("correct datastore engine specified", func(t *testing.T) {
+		cfg := DefaultConfig()
+		cfg.Datastore.URI = "postgresql://localhost:5432"
```
jon-whit, line 290: I didn't think `postgresql` was a valid prefix? The enumerations in `cmd/shared.go` below define `postgres`.

This test should return an error, and we should change this test to assert `postgres://localhost:5432` does not.

jon-whit (changes requested):
Left some comments for review.

cmd/run/run_test.go
Himanshu372: I think, I accidently set URI in this test. I just wanted to test for correct DatastoreEngine. After changes, I am getting expected result. 
Although one thing which I noticed is, on setting wrong postgres URI, `RunServer` stalls and waits for connection. I think we should fail early. Let me know your thoughts

Himanshu372:
> Left some comments for review.

@jon-whit Addressed comments, plz check.

Himanshu372:
> > Left some comments for review.
> 
> @jon-whit Addressed comments, plz check.

Hey @jon-whit plz have a look and let me know if the changes are OK

cmd/cmd.go
```diff
@@ -0,0 +1,39 @@
...
+		}
+	}
+	return false
+}
+
+// NewDatastoreEngine inits a new valid DatastoreEngine type
+func NewDatastoreEngine(engine string) (DatastoreEngine, error) {
+	dsEngine := DatastoreEngine(engine)
+	if isValid := dsEngine.IsValid(); !isValid {
+		return "", fmt.Errorf("invalid datastore engine '(%s)'", engine)
```
miparnisari, line 36: We have similar code here https://github.com/openfga/openfga/blob/0f3e337dec1503fa3fc6b045f2efbb27a4defa19/cmd/run/run.go#L360 and here https://github.com/openfga/openfga/blob/8dc2ed62413e7cccc7da57da057f1f1fc17dcb8a/cmd/validatemodels/validate_models.go#L70

miparnisari:
Hi @Himanshu372, i apologize for taking so long to review this. I took the liberty of updating this PR with respect to main. I also added some more validations to the datastore configs.

curfew-marathon:
Thank you for opening the PR.  At this time we are closing the PR as it has gone stale and needs to be brought up to date.  In the future feel free to re-open the PR.  Thanks again.

PR #817, opened 2023-06-15
feat: support ./openfga validate-models

miparnisari:
## Description
- refactor: when different commands use the same flags (e.g. `--datastore-engine`), ensure the flag binding is correctly applied _only_ to the command being run. See a demo of this issue: https://gist.github.com/gesquive/c36501fbd84ecab00d3ff2c9502020f9
   -  Functionality for `./openfga migrate`, `./openfga run` and `./openfga version` should **NOT** be affected by this PR.
   
- feat: add `./openfga validate-models`. It fetches all stores and every model in those stores, and prints validation results in JSON format.

## Testing

```shell
# test 1: reads from config
[15/06/23 4:49:36] ~/GitHub/openfga (feat-validate-cmd) $ cat $HOME/.openfga/config.yaml
datastore:
  engine: invalid
[15/06/23 4:49:40] ~/GitHub/openfga (feat-validate-cmd) $ ./openfga validate-models     
Error: invalid datastore engine type: invalid
Usage:
  openfga validate-models [flags]

Flags:
      --datastore-engine string   the datastore engine
      --datastore-uri string      the connection uri to the datastore
  -h, --help                      help for validate-models

# test 2: validates and prints json
[15/06/23 4:49:44] ~/GitHub/openfga (feat-validate-cmd) $ cat $HOME/.openfga/config.yaml
datastore:
  engine: postgres
  uri: postgres://postgres:password@localhost:5432/postgres%     

[21/06/23 12:35:41] ~/GitHub/openfga (feat-validate-cmd) $ ./openfga validate-models
[
     {
         "store_id": "01H3FN0X01WDQ6KNDVR4RHKK99",
         "model_id": "01H3FN17M1XKY3AAMBBJW0AJD8",
         "is_latest_model": true,
         "error": ""
     }
 ]

```


jon-whit:
Do we want to have a `openfga validate-model [model-id]` that has a flag `openfga validation-model --all` so that we have more general use for the CLI command?

`openfga validate-models` is very specific to fetching all models and validating each and every one of them. What if I want to just validate a specific one and/or optionally validate all of them with the `--all` argument?

cmd/validate.go
```diff
@@ -0,0 +1,81 @@
...
+	flags := cmd.Flags()
+	flags.String(datastoreEngineFlag, "", "the datastore engine")
+	flags.String(datastoreURIFlag, "", "the connection uri to the datastore")
+
+	// NOTE: if you add a new flag here, add the binding in PreRunE
+
+	return cmd
+}
+
+func runValidate(_ *cobra.Command, _ []string) error {
```
jon-whit, line 41: Can we please emit a warning message at the very beginning of this command that notes that this command is in beta and may be removed in the future?

cmd/validate.go
jon-whit, line 1: Mind renmaing this to `validate_model.go`?

miparnisari:
> Do we want to have a openfga validate-model [model-id] that has a flag openfga validation-model --all so that we have more general use for the CLI command?

We discussed offline with the team and we decided to keep it as is, i.e. validate all existing models.

cmd/run/run.go
```diff
@@ -60,16 +61,256 @@ import (
...
 	cmd := &cobra.Command{
 		Use:   "run",
 		Short: "Run the OpenFGA server",
 		Long:  "Run the OpenFGA server.",
 		Run:   run,
 		Args:  cobra.NoArgs,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			flags := cmd.Flags()
+
+			util.MustBindPFlag("experimentals", flags.Lookup("experimentals"))
```
jon-whit, line 79: Why did we move these in here instead of in `run/flags.go`? I think these are more aptly placed in their own file, because they litter this command setup code.

miparnisari: `flags.go` has code that is supposed to be only for the `run` command, but it executes irrespective of what command is being used. This introduces inconsistent behavior if we decide to have other commands with the same flags.

See https://github.com/spf13/viper/issues/233#issuecomment-479336184

> Bind your viper flags in `PreRun() or Run()` -- **NOT** `init()`.

(In our case we don't have `init()`, but the binding still runs in every case)

cmd/validate_models.go
```diff
@@ -0,0 +1,81 @@
...
+	engine := viper.GetString(datastoreEngineFlag)
+	uri := viper.GetString(datastoreURIFlag)
+
+	ctx := context.Background()
+
+	var (
+		db  storage.OpenFGADatastore
+		err error
+	)
+	switch engine {
```
jon-whit, line 51: Please keep the error messages surfaced from this switch consistent with the other code:
https://github.com/openfga/openfga/blob/main/cmd/run/run.go#L441-L460

cmd/util/util.go
```diff
@@ -33,3 +46,125 @@ func Index[E comparable](s []E, v E) int {
...
+	confdir := PrepareTempConfigDir(t)
+	confFile, err := os.Create(filepath.Join(confdir, "config.yaml"))
+	require.Nil(t, err)
+	_, err = confFile.WriteString(config)
+	require.Nil(t, err)
+	require.Nil(t, confFile.Close())
+}
+
+type ValidationResult struct {
+	StoreID       string
```
jon-whit, line 94: For consistency with our API, please add struct tags so that these JSON marshal consistently with our API protobufs.

`store_id`
`authorization_model_id`

We also tend to use snake case instead of camelCase, so `is_latest_model` would be preferred over `IsLatestModel`.

cmd/validate_models.go
```diff
@@ -0,0 +1,81 @@
...
+	case "":
+		return fmt.Errorf("missing datastore engine type")
+	case "memory":
+		fallthrough
+	default:
+		return fmt.Errorf("invalid datastore engine type: %s", engine)
+	}
+
+	if err != nil {
+		return fmt.Errorf("could not connect to database")
```
jon-whit, line 65: Wrap the error here so we have the underlying error context. And please keep it consistent with our other error messages.
https://github.com/openfga/openfga/blob/main/cmd/migrate.go#L93

```go
return fmt.Errorf(failed to open a connection to the datastore: %v", err)
```

cmd/run/run.go
jon-whit: It can remain as a `PreRun` func, but the organization of the code in its own file creates better code structure and organization.

```go
// cmd/run/flags.go

func BindRunFlagsFunc(cmd *cobra.Command, args []string) {
    ...
    util.MustBindPFlag(...) // pflag bindings specific to `openfga run`.
}
```
and then in, for example, `cmd/run/run.go`
```go
cmd := &cobra.Command{
    ...
    PreRun: cmd.BindCmdFlagsFunc,
    ...
```

miparnisari: If I apply your suggestion, the declaration of flags will be in `NewRunCommand()` but the binding will be in `BindRunFlagsFunc` which is in a different file. So if someone tries to add a new flag, the change will be spread across  two files instead of one. I don't se the benefit of that 🤔

jon-whit: What's wrong with that? It's within the same package. Go source files organize code in a logical way within the same package. The alternative and anti-practice would be to put everything within a single Go file.

cmd/validate_models.go
miparnisari: I'll send another PR that standardizes across the three commands

cmd/run/run.go
miparnisari: > What's wrong with that? 

There's nothing wrong, i'm asking what is the benefit. That file will only be changed when we add a new flag.

jon-whit: The benefit to me as a maintainer is that it's easier to know to go to `flags.go` to change flag registration than it is to go to the PreRun function body of the `run.go` file.

cmd/validate_models.go
```diff
@@ -0,0 +1,161 @@
...
+	flags := cmd.Flags()
+	flags.String(datastoreEngineFlag, "", "the datastore engine")
+	flags.String(datastoreURIFlag, "", "the connection uri to the datastore")
+
+	// NOTE: if you add a new flag here, add the binding in PreRunE
+
+	return cmd
+}
+
+type ValidationResult struct {
```
jon-whit, line 42: Was making this publicly exported done on purpose? What's the use case for making a public contract on this? Can't we make this an unexported type that is only used in the scope of this CLI entrypoint?

cmd/run/run.go
miparnisari: Updated all commands so that each have their own `flags.go` file

cmd/validate_models.go
miparnisari: Not on purpose 😅

pkg/storage/mysql/mysql_test.go
```diff
@@ -27,30 +25,6 @@ func TestMySQLDatastore(t *testing.T) {
 	test.RunAllTests(t, ds)
 }
 
-func TestMigrate(t *testing.T) {
```
miparnisari, line 30: FYI i moved these tests so that there is no cycle in the import loop

adriantam: Where do we move these tests to?  I cannot find it in this review

miparnisari: I forgot to commit that file 🤣  added! The test is `TestMigrateCommandRollbacks`

cmd/run/flags.go
```diff
@@ -7,191 +7,145 @@ import (
 
 // bindRunFlags binds the cobra cmd flags to the equivalent config value being managed
 // by viper. This bridges the config between cobra flags and viper flags.
-func bindRunFlags(command *cobra.Command) {
-	defaultConfig := DefaultConfig()
+func bindRunFlags(command *cobra.Command, _ []string) {
 	flags := command.Flags()
```
jon-whit, line 11: I'm wondering why we duplicate the flags construction and why the flags don't live alongside the bindings here? Can't we pass in the flags at the caller layer and then return the PreRun func? Or is there an issue with that?

```go
func bindRunFlagFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
    return func(cmd *cobra.Command, args []string {
        util.MustBindPFlag(..., flags.Lookup(...))
    }
}
```
and then it's usage 
```go
// cmd/run/run.go
flags := command.Flags()

cmd := &cobra.Command{
    ...
    PreRun: bindRunFlagsFunc(flags)
}
```

miparnisari: `flags := command.Flags()` is referencing the created `cobra.Command`, i need to build that object first

jon-whit: So then
```go
cmd := &cobra.Command{
    ...
}

flags := cmd.Flags()

cmd.PreRun = bindRunFlagsFunc(flags)
```

jon-whit (approved):

PR #820, opened 2023-06-21
feat(perf): cache inflight requests for FindLastestAuthorizationModelId

elbuo8:
## Review Checklist


elbuo8:
No, we didn’t have it before. 

The reason this works is due to this line:
https://github.com/openfga/openfga/blob/5e8814be8e74b9e25693d73b273b20b699b42c55/pkg/storage/caching/caching.go#L18

So the struct inherits all the methods of the OpenFGAStorage it receives and any that gets declared simply becomes an override. Basically how you do inheritance in Go.

miparnisari:
> The cache is only during the duration of the first inflight request.

What does this mean? `cachedOpenFGADatastore` is shared across all requests

elbuo8:
Only concurrent calls to this method are cached together. And it will last for the duration of the first call. Not the whole lifecycle of the request. So if the FindLastestModelID takes 5ms, that's how long the cache will last even tho the request might take a total of 35ms.

Picture this scenario:
R1 -> FindLastestModelID -> Calls Datastore and creates cache
R2 -> FindLastestModelID -> Will share the response with R1 
R3 -> ... some delay -> FindLastestModelID -> Calls Datastore and creates new cache
R4 -> ... some delay -> FindLastestModelID -> Will share the response with R2

pkg/storage/caching/caching.go
```diff
@@ -16,6 +17,7 @@ var _ storage.OpenFGADatastore = (*cachedOpenFGADatastore)(nil)
 
 type cachedOpenFGADatastore struct {
 	storage.OpenFGADatastore
+	G     singleflight.Group
```
jon-whit, line 20:
```suggestion
	     lookupGroup singleflight.Group
```

jon-whit: No need to have the field as a public viewable member.

pkg/storage/caching/caching.go
```diff
@@ -46,6 +48,16 @@ func (c *cachedOpenFGADatastore) ReadAuthorizationModel(ctx context.Context, sto
 	return model, nil
 }
 
+func (c *cachedOpenFGADatastore) FindLatestAuthorizationModelID(ctx context.Context, storeID string) (string, error) {
+	v, err, _ := c.G.Do(fmt.Sprintf("FindLatestAuthorizationModelID:%s", storeID), func() (interface{}, error) {
+		return c.OpenFGADatastore.FindLatestAuthorizationModelID(ctx, storeID)
```
jon-whit, line 53: Can we please add a test for this to make sure that the mocked call to this method is not called more than once for 2 concurrent lookups?

pkg/storage/caching/caching.go
```diff
@@ -16,6 +17,7 @@ var _ storage.OpenFGADatastore = (*cachedOpenFGADatastore)(nil)
 
 type cachedOpenFGADatastore struct {
```
jon-whit, line 18: Do we not need to initialize the singleflight group in the constructor?

jon-whit (changes requested):
Left some comments for review.

pkg/storage/caching/caching.go
```diff
@@ -46,6 +48,16 @@ func (c *cachedOpenFGADatastore) ReadAuthorizationModel(ctx context.Context, sto
 	return model, nil
 }
 
+func (c *cachedOpenFGADatastore) FindLatestAuthorizationModelID(ctx context.Context, storeID string) (string, error) {
+	v, err, _ := c.G.Do(fmt.Sprintf("FindLatestAuthorizationModelID:%s", storeID), func() (interface{}, error) {
```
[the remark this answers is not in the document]
elbuo8: If we were to use the `lookupGroup` for other cache purposes, we wouldn't need to think of collisions or adding another group. Happy to shorten the name tho.

pkg/storage/caching/caching.go
elbuo8: No need. But if you prefer explicitness it can be done.

miparnisari (approved):

jon-whit (approved):

pkg/storage/caching/caching.go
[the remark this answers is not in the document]
jon-whit: I think I'd prefer it be shortened, because it's uniquely bound to this instance of the `lookupGroup`. Only if we were to use it within the cachedOpenFGADatastore in another method would it be an issue. However, it's not a huge deal. Mostly a nit pick 😄

PR #822, opened 2023-06-22
refactor: tracer

miparnisari:
## Description
~~Add support for flags `OPENFGA_TRACE_ENABLE_TAIL_LATENCY_EXPORTER` and `OPENFGA_TRACE_TAIL_LATENCY` which together ensure that only traces that are >= a given latency (default 1 second) are exported to the OTEL collector.~~

Refactor tracer:
- New constructor which allows customizable tracing attributes
- Add integration test


pkg/telemetry/tracing.go
```diff
@@ -40,6 +41,13 @@ func MustNewTracerProvider(endpoint string, serviceName string, ratio float64) *
 		panic(fmt.Sprintf("failed to establish a connection with the otlp exporter: %v", err))
 	}
 
+	if enableTailLatencySpanExporter {
+		if ratio != 1 {
```
jon-whit, line 45: Why require this? If you want to downsample AND only export high latency spans, then I don't see a problem with that.

I'd avoid being over pedantic here on this check.

pkg/telemetry/tracing.go
```diff
@@ -17,7 +17,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-func MustNewTracerProvider(endpoint string, serviceName string, ratio float64) *sdktrace.TracerProvider {
+func MustNewTracerProvider(endpoint string, serviceName string, ratio float64, enableTailLatencySpanExporter bool, tailLatencyInMs int) *sdktrace.TracerProvider {
```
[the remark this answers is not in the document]
miparnisari: Will do

pkg/telemetry/tracing.go
```diff
@@ -5,43 +5,77 @@ import (
...
+	}
+}
+
+func WithAttributes(attrs ...attribute.KeyValue) TracerOption {
+	return func(d *CustomTracer) {
+		d.attributes = attrs
+	}
+}
+
+type CustomTracer struct {
```
jon-whit, line 39: Does this need to be exported? If we're going to have a construct return the `TracerProvider`, then I'd recommend we unexport this, because the intent is to build a contract on the OTEL interfaces, not the CustomTracer directly.

internal/authn/mocks/mock_tracing_server.go
```diff
@@ -0,0 +1,48 @@
+package mocks
```
jon-whit, line 1: Since the mocked interface contained herein is not related to `authn`, can we please put these in an alternative package, maybe simply `internal/mocks` or `internal/telemetry/mocks`?

cmd/run/run_test.go
```diff
@@ -378,6 +379,43 @@ func TestBuildServiceWithPresharedKeyAuthentication(t *testing.T) {
 	}
 }
 
+func TestBuildServiceWithTailSpanExporterEnabled(t *testing.T) {
```
jon-whit, line 382: I don't think we need this test anymore, right?

internal/authn/mocks/mock_tracing_server.go
jon-whit: Also, do we need this mocked interface if we remove the test above?

Related to https://github.com/openfga/openfga/pull/822/files#r1244374750

pkg/telemetry/tracing.go
miparnisari: No need to be exported

internal/authn/mocks/mock_tracing_server.go
miparnisari: Yes the mock is still useful because the test is still useful :D

cmd/run/run_test.go
miparnisari: This test is still useful, but the name is wrong

pkg/telemetry/mocks/mock_tracing_server.go
```diff
@@ -0,0 +1,48 @@
+package mocks
```
jon-whit, line 1: Let's not put mocks in the public `pkg` package. That opens them up to a public contract with other Go devs importing the OpenFGA library.

pkg/telemetry/mocks/mock_tracing_server.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (s *MockTracingServer) Export(context.Context, *otlpcollector.ExportTraceServiceRequest) (*otlpcollector.ExportTraceServiceResponse, error) {
+	s.serviceMu.Lock()
+	defer s.serviceMu.Unlock()
+	s.exportCount++
+	return &otlpcollector.ExportTraceServiceResponse{}, nil
+}
+
+func NewMockTracingServer(port int) (*MockTracingServer, error) {
```
jon-whit, line 27:
```suggestion
func NewMockTracingServer(port int) (otlpcollector.TraceServiceServer, error) {
   	mockServer := &mockTracingServer{exportCount: 0}

	go func() {
		lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
		if err != nil {
			log.Fatalf("failed to listen: %v", err)
		}
		server := grpc.NewServer()
		otlpcollector.RegisterTraceServiceServer(server, mockServer)
		if err := server.Serve(lis); err != nil {
			log.Fatalf("failed to serve: %v", err)
		}
	}()
	
	return mockServer, nil
}
```

pkg/telemetry/mocks/mock_tracing_server.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"log"
+	"net"
+	"sync"
+
+	otlpcollector "go.opentelemetry.io/proto/otlp/collector/trace/v1"
+	"google.golang.org/grpc"
+)
+
+type MockTracingServer struct {
```
jon-whit, line 14:
```suggestion
type mockTracingServer struct {
```

jon-whit: You know what, why not just generate a mock for this using mockgen and assert that the `Export` function is called exactly once? I think that'd be more preferable than writing custom mock code and maintaining the mocked code.

I recommend we add a `go:generate mockgen ...` directive at the top of the test file that requires this mocked interface, and export the mocked interface to the `internal/mocks` package.

```
go:generate mockgen -destination <path>/mock_trace_service_server.go -package mocks  go.opentelemetry.io/proto/otlp/collector/trace/v1 TraceServiceServer
```

Then we can do something like:
```go
defer mockController.Finish()

mockTraceServiceServer := mockstorage.NewMockTraceServiceServer(mockController)

mockTraceServiceServer.EXPECT().Export(...).Times(1)
```

miparnisari: This is a good idea, but the mock has to listen on a specific port or else the thing fails: https://github.com/openfga/openfga/blob/df6f05a3d2ac1d8472079057d4b73c3e337f59ce/pkg/telemetry/tracing.go#L68-L75

I tried having the mock listen on a port with

```go
func TestBuildServiceWithTracingEnabled(t *testing.T) {
	// create mock OTLP server
	defer mockController.Finish()
	mockTraceServiceServer := mocks2.NewMockTraceServiceServer(mockController)
	mockTraceServiceServer.EXPECT().Export(gomock.Any(), gomock.Any()).Times(1)

	// attach it to a port
	otlpServerPort, otlpServerPortReleaser := TCPRandomPort()
	localOTLPServerURL := fmt.Sprintf("localhost:%d", otlpServerPort)
	otlpServerPortReleaser()

	go func() {
		lis, err := net.Listen("tcp", fmt.Sprintf(":%d", otlpServerPort))
		if err != nil {
			log.Fatalf("failed to listen: %v", err)
		}
		server := grpc.NewServer()
		otlpcollector.RegisterTraceServiceServer(server, mockTraceServiceServer) // <-------- fails here
		if err := server.Serve(lis); err != nil {
			log.Fatalf("failed to serve: %v", err)
		}
	}()

	// create OpenFGA server with tracing enabled
	cfg := MustDefaultConfigWithRandomPorts()
	cfg.Trace.Enabled = true
	cfg.Trace.SampleRatio = 1
	cfg.Trace.OTLP.Endpoint = localOTLPServerURL
```

However, the code doesn't build. 

```
cmd/run/run_test.go:405:52: cannot use mockTraceServiceServer (variable of type *"github.com/openfga/openfga/internal/mocks".MockTraceServiceServer) as "go.opentelemetry.io/proto/otlp/collector/trace/v1".TraceServiceServer value in argument to otlpcollector.RegisterTraceServiceServer: 

*"github.com/openfga/openfga/internal/mocks".MockTraceServiceServer does not implement "go.opentelemetry.io/proto/otlp/collector/trace/v1".TraceServiceServer (missing method mustEmbedUnimplementedTraceServiceServer)
```

i'm not sure why it complains because the mock does have an implementation of `mustEmbedUnimplementedTraceServiceServer`

pkg/telemetry/mocks/mock_tracing_server.go
miparnisari: Moved all our existing mocks to `internal/mocks`

pkg/telemetry/mocks/mock_tracing_server.go
miparnisari: I found this article https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a that says that it's better to return structs.

pkg/telemetry/mocks/mock_tracing_server.go
jon-whit: Taking a look 👀

pkg/telemetry/mocks/mock_tracing_server.go
jon-whit: The guidance of "Accept interfaces, return structs" is good general advice, but not always applicable. In this case, the only thing we're really testing for is that the function `Export` of the mocked interface (`TraceServiceServer`) is called exactly once. So creating a dependency on a custom struct when the actual unit under tests is a mocked interface seems counterproductive here. This issue goes away though if we can find a good approach for https://github.com/openfga/openfga/pull/822#discussion_r1244456438

Let me get back with ya.

internal/mocks/mock_tracing_server.go
```diff
@@ -0,0 +1,50 @@
...
+	defer s.serviceMu.Unlock()
+	s.exportCount++
+	return &otlpcollector.ExportTraceServiceResponse{}, nil
+}
+
+func NewMockTracingServer(port int) (*mockTracingServer, error) {
+	mockServer := &mockTracingServer{exportCount: 0}
+
+	go func() {
+		lis, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
```
jon-whit, line 33: I'm still looking into a way to mock this with `mockgen` and not having to have this, but in the meantime the https://pkg.go.dev/google.golang.org/grpc/test/bufconn package is a pretty standard package to use to test grpc services. You can avoid network and instead use a buffered loop connection. We should change this to the `bufconn` package, because then you won't get those firewall pop-ups 😄

cmd/run/run_test.go
```diff
@@ -378,6 +379,43 @@ func TestBuildServiceWithPresharedKeyAuthentication(t *testing.T) {
 	}
 }
 
+func TestBuildServiceWithTracingEnabled(t *testing.T) {
+	// create mock OTLP server
+	otlpServerPort, otlpServerPortReleaser := TCPRandomPort()
+	localOTLPServerURL := fmt.Sprintf("localhost:%d", otlpServerPort)
+	otlpServerPortReleaser()
+	otlpServer, err := mocks.NewMockTracingServer(otlpServerPort)
```
jon-whit, line 387: You need a way to close the mock server, otherwise we leak in our tests.

There should be a `defer otlpServer.Close()` that closes the underlying grpc server.

pkg/telemetry/mocks/mock_tracing_server.go
miparnisari: https://github.com/uber/mock/issues/23

pkg/telemetry/mocks/mock_tracing_server.go
miparnisari: So... I added a method `GetExportCount` in the mock that is used in the test to assert that  the `Export` method got called. If I change the return type, I lose access to this method.

Unless I figure out how to use `mockgen`, I don't see a way of solving this elegantly :/

internal/mocks/mock_tracing_server.go
miparnisari: I gave it a try and the test fails here https://github.com/openfga/openfga/blob/df6f05a3d2ac1d8472079057d4b73c3e337f59ce/pkg/telemetry/tracing.go#L68C1-L75

with `panic: failed to establish a connection with the otlp exporter: context deadline exceeded`

miparnisari:
Until I figure out a better approach, @jon-whit do you think this is worth merging as is?

internal/mocks/mock_storage.go
jon-whit, line 1: One last QQ

Do we use the storage mocks in Auth0 FGA? If so, we may need to leave that mock public? Or we can generate a mock of it in Auth0 FGA directly instead. Either way, we should look real quick.

miparnisari: Good point. It's not used in Auth0 FGA.

jon-whit (approved):

PR #831, opened 2023-06-28
fix: memoize typesystem resolution and model validation

jon-whit:
## Description
The changes herein introduce a new `TypesystemResolver` and a specific implementation of a typesystem resolver that memoizes (caches) typesystem resolution and the model validations applied to them.

**Screenshot 1a** (first API call without providing the model id explicitly)
Screen Shot 2023-06-28 at 3 30 31 PM

You can see that the `MemoizedTypesystemResolverFunc` was called and it looked up the latest model from the database and then called `NewAndValidate` on it.

**Screenshot 1b** (second API call without providing the model id explicitly)
Screen Shot 2023-06-28 at 3 30 38 PM

You can see that we called the `MemoizedTypesystemResolverFunc` again and we looked up the latest model from the database, but since it was unchanged we didn't have to go back through model validation (e.g. there isn't a `NewAndValidate` step).

**Screenshot 2a** (first API call if model id is provided explicitly)
Screen Shot 2023-06-28 at 3 37 11 PM

You see `MemoizedTypesystemResolverFunc` is called and that resolves the model with a database call and then validates the typesystem and returns it.

**Screenshot 2b** (second API call with model id provided explicitly)
Screen Shot 2023-06-28 at 3 37 17 PM

You see `MemoizedTypesystemResolverFunc` is called, but it immediately returns without doing any extra work, because the model has already been resolved and validated in the prior call (screenshot 2a above).

## Review Checklist


pkg/server/server.go
```diff
@@ -197,15 +177,18 @@ func (s *Server) Write(ctx context.Context, req *openfgapb.WriteRequest) (*openf
 
 	storeID := req.GetStoreId()
 
-	modelID, err := s.resolveAuthorizationModelID(ctx, storeID, req.GetAuthorizationModelId())
+	// todo(jon-whit): read/write API endpoints (e.g. non-query endpoints) don't necessarily need
```
jon-whit, line 180: We probably could avoid introducing the overhead of the model validation step in the read/write endpoints, but assuming the models are used across endpoints (which I assume they are), then it should be irrelevant anyways.. So I'm inclined to leave this as is and consistent with the rest, but I'm open to implementing another resolver that omits typesystem construction and validation and just resolves the model itself.

Thoughts? Concerns?

elbuo8: I agree, leave as is.

pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,79 @@
...
+// resolution of a Typesystem.
+type TypesystemResolverFunc func(ctx context.Context, storeID, modelID string) (*TypeSystem, error)
+
+// MemoizedTypesystemResolver returns a TypesystemResolverFunc that either fetches the provided authorization
+// model (if provided) or looks up the latest authorizaiton model, and then it constructs a TypeSystem from
+// the resolved model. The type-system resolution is memoized so if another lookup of the same model occurs,
+// then the earlier TypeSystem that was constructed will be used.
+//
+// The memoized resolver function is safe for concurrent use.
+func MemoizedTypesystemResolverFunc(reader storage.AuthorizationModelReadBackend) TypesystemResolverFunc {
```
elbuo8, line 28: nitpit: weird naming convention 🤔

jon-whit: How so? What would be your suggested name?

elbuo8: You normally use `ds` or `datastore`, kinda got used to that 😆

elbuo8 (dismissed):

pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		key := fmt.Sprintf("%s/%s", storeID, modelID)
+
+		item := cache.Get(key)
+		if item != nil {
+			return item.Value(), nil
+		}
+
+		model, err := reader.ReadAuthorizationModel(ctx, storeID, modelID)
```
elbuo8, line 61: For another PR: this method is a good candidate for singleflight. Specifically in the scenario where the cache is empty (deployment, scale up event).

pkg/typesystem/resolver.go
jon-whit: Oh, I see. Happy to change that. The change was because in most other places we are referring to `OpenFGADatastore` (e.g. the whole interface), whereas here we are just referring to the read interface for Authz models.

pkg/typesystem/resolver.go
jon-whit: For sure! I'll capture as an issue and save it for another PR. Thanks 👍

pkg/typesystem/resolver.go
miparnisari: Also please fix the doc: `MemoizedTypesystemResolver` to `MemoizedTypesystemResolverFunc` :)

The reason i bring it up is because Goland is underlining this for me and it's veeeery annoying
image

pkg/server/server.go
```diff
@@ -472,40 +432,34 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
...
 
-		return "", serverErrors.HandleError("", err)
+		return nil, serverErrors.HandleError("", err)
 	}
 
-	return modelID, nil
+	resolvedModelID := typesys.GetAuthorizationModelID()
+
+	span.SetAttributes(attribute.KeyValue{Key: authorizationModelIDKey, Value: attribute.StringValue(resolvedModelID)})
+	grpc_ctxtags.Extract(ctx).Set(authorizationModelIDKey, resolvedModelID)
```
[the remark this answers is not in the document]
jon-whit: If you set it in the ctx as grpc_ctxtags, then it's available anywhere in the call-chain, and we use it in the middleware to log the model id as a field in the log message. ctxtags get injected into log messages.

pkg/typesystem/resolver.go
jon-whit: Done and done 👍

pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,79 @@
...
+
+// MemoizedTypesystemResolverFunc returns a TypesystemResolverFunc that either fetches the provided authorization
+// model (if provided) or looks up the latest authorizaiton model, and then it constructs a TypeSystem from
+// the resolved model. The type-system resolution is memoized so if another lookup of the same model occurs,
+// then the earlier TypeSystem that was constructed will be used.
+//
+// The memoized resolver function is safe for concurrent use.
+func MemoizedTypesystemResolverFunc(datastore storage.AuthorizationModelReadBackend) TypesystemResolverFunc {
+
+	cache := ccache.New(ccache.Configure[*TypeSystem]())
```
miparnisari, line 30: I won't block this PR, but i don't feel super happy about having this cache here and one for `AuthorizationModels` [here](https://github.com/openfga/openfga/blob/5c7c5c88dd07c85666275621503f867428a59693/pkg/storage/caching/caching.go#L21).

jon-whit: I'd be happy to remove the code in https://github.com/openfga/openfga/blob/main/pkg/storage/caching/caching.go and move it over into here.

@elbuo8 already suggested we also integrate `singleflight into this work, and so I can just do that now and then delete that whole `storage/caching` package.

How does that sound to you?

pkg/typesystem/resolver.go
```diff
@@ -0,0 +1,79 @@
...
+				return nil, ErrModelNotFound
+			}
+
+			return nil, fmt.Errorf("failed to ReadAuthorizationModel: %w", err)
+		}
+
+		typesys, err := NewAndValidate(ctx, model)
+		if err != nil {
+			return nil, fmt.Errorf("%w: %v", ErrInvalidModel, err)
+		}
```
miparnisari, line 73: If someone continuously passes an invalid auth model, we are not storing it anywhere in the cache and we'll try to fetch it from the DB every time 🤔

pkg/typesystem/resolver.go
miparnisari: Yeah that'd be nice :D

pkg/server/server.go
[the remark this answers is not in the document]
miparnisari: Ah it's used here 

https://github.com/openfga/openfga/blob/5c7c5c88dd07c85666275621503f867428a59693/pkg/middleware/logging/logging.go#L51

pkg/typesystem/resolver.go
miparnisari, line 30: also, we need to call `cache.Stop()` when the server terminates - this is important when running the tests

jon-whit: Ok, done 👍 

Tell me what you think.

jon-whit: You have to call `.Stop()` on an in-memory cache. That's funky as hell and not cool.. What's the issue in the tests, is it the goroutine leak because of the background cache worker?

pkg/typesystem/resolver.go
jon-whit: Yeah, I purposefully chose not to cache invalid models because of the footprint, and if the model is invalid then the request can't be served successfully anyways, and so the SLO latency expectations around those cases aren't as tight tolerance as the alternative.

If you are referencing an invalid auth model and you experience a little more response latency, but you reduce memory footprint because you don't cache the invalid models, then that is a fine trade-off IMO.

pkg/typesystem/resolver.go
jon-whit: https://github.com/karlseguin/ccache/issues/59 seems that if you call stop you get races potentially as well..

jon-whit: Maybe we should look for an alternative cache package? The cache design is pretty odd to begin with TBH.

https://github.com/allegro/bigcache is widely used for high concurrency scenarios, which we have in this case.

miparnisari: > is it the goroutine leak because of the background cache worker? Is that just because of our usage of uber-go/goleak?

i haven't ran a test with it recently but i expect to see some leaks, yeah. Not a big deal though because it would only affect tests

miparnisari: I think we introduce a regression by removing `caching.go`. Because for example, `server.Expand` calls `s.resolveTypesystem()` which fetches the auth model, and then `NewExpandQuery.Execute()` calls `datastore.ReadAuthorizationModel` which will fetch the model again, no? 

If we remove `caching.go` we'd need to pass `typesys` in all `Execute()` calls, which is a reasonable path forward but it will blow up this PR with changes

jon-whit: Yeah, I guess that's why I need to revive and finish https://github.com/openfga/openfga/pull/756 after this PR is merged.

I'll revert the deletion of `pkg/storage/caching` and then we can drop it later.

jon-whit: @miparnisari ok I've reverted the commit that deletes the `pkg/storage/caching` code, but I've integrated the `singleflight` work into the memorized typesystem resolver. That should do it.

jon-whit: We'd have to change composition a good amount so we can introduce a `.Close` or `.Stop` method on a struct that implements some `TypesystemResolver` interface or order to manage this. It feels like a really dirty pattern just so we can call close/stop on the underlying cache. An in-memory cache really shouldn't have a Stop...

I'd vote we table it and look at swapping away from our existing cache lib, it's odd.

miparnisari (approved):

PR #832, opened 2023-06-29
chore: drop optimized-list-objects experimental flag and related code

jon-whit:
## Description
These changes drop the experimental flag `optimized-list-objects` and makes the optimized ListObjects implementation the defacto implementation. This is in preparation for the upcoming `v1.2.0` release that will include this new optimized ListObjects algorithm.

Additional changes were made in this PR to drop other code related to the old (unoptimized ListObjects) implementation. Namely, changes were made to the storage interface for methods no longer needed.

## Review Checklist


pkg/server/server.go
```diff
@@ -35,8 +34,6 @@ const (
 	authorizationModelIDKey    = "authorization_model_id"
 
 	checkConcurrencyLimit = 100
-
-	optimizedListObjects ExperimentalFeatureFlag = "optimized-list-objects"
```
miparnisari, line 39: Please remove flag from schema.json

pkg/storage/storage.go
```diff
@@ -100,16 +100,6 @@ type RelationshipTupleReader interface {
...
-
-	// ListObjectsByType returns all the objects of a specific type.
-	// You can assume that the type has already been validated.
-	// The result can't have duplicate elements.
-	// There is NO guarantee on the order returned on the iterator.
-	ListObjectsByType(
-		ctx context.Context,
-		store string,
-		objectType string,
-	) (ObjectIterator, error)
```
miparnisari, line 112: can we remove all this code in another PR so that if we ever want to revert it, we only have to undo that PR and not pieces of this one?

jon-whit: If you revert back to code before this it will include this code and the code involving concurrent Checks (e.g. `performChecks` and `internalCheck` in the `pkg/server/commands/list_objects.go` file), so I'm not sure I understand? Could you elaborate on the scenario where/why it would be pertinent to separate them?

pkg/server/server.go
jon-whit: Done.

pkg/storage/storage.go
jon-whit: For example, if we release this as `v1.2.0` and update Auth0 FGA to use that version of OpenFGA, then when we made that `go.mod` update in Auth0 FGA we'll have to drop this storage signature. If we want to revert then we just revert to the prior version that was before the `v1.2.0` bump of OpenFGA and we'll get the storage signature implementation(s) back along with the old ListObjects implementation.

miparnisari: > Could you elaborate on the scenario where/why it would be pertinent to separate them?

ListObjectsByType has code that could potentially be needed in the future. So i'm saying that in this PR we merely stop using it in ListObjects, and then in another PR we remove the SQL code itself, the storage definition, the caching around it, etc.

miparnisari (approved):
Won't block this PR with my comment :)

PR #834, opened 2023-06-29
fix: add in distinguish message for loop

adriantam:
## Description
Allow better identification in message when running validation command

Running validation program on a potential loop model will result in

```
validate              |          "error": "the definition of relation 'r1' in object type 'doc' is invalid : potential loop found for relation 'doc#r1'"
```

## Review Checklist

### DIFF
--- a/cmd/validatemodels/validate_models.go
+++ b/cmd/validatemodels/validate_models.go
@@ -135,13 +135,7 @@ func ValidateAllAuthorizationModels(ctx context.Context, db storage.OpenFGADatas
 					}
 
 					if err != nil {
-						if invalidRelation, ok := err.(*typesystem.InvalidRelationError); ok {
-							// This will give a more detailed error
-							validationResult.Error = invalidRelation.Error() + " : " + invalidRelation.Cause.Error()
-						} else {
-							validationResult.Error = err.Error()
-
-						}
+						validationResult.Error = err.Error()
 					}
 					validationResults = append(validationResults, validationResult)
 				}
--- a/cmd/validatemodels/validate_models.go
+++ b/cmd/validatemodels/validate_models.go
@@ -135,7 +135,13 @@ func ValidateAllAuthorizationModels(ctx context.Context, db storage.OpenFGADatas
 					}
 
 					if err != nil {
-						validationResult.Error = err.Error()
+						if invalidRelation, ok := err.(*typesystem.InvalidRelationError); ok {
+							// This will give a more detailed error
+							validationResult.Error = invalidRelation.Error() + " : " + invalidRelation.Cause.Error()
+						} else {
+							validationResult.Error = err.Error()
+
+						}
 					}
 					validationResults = append(validationResults, validationResult)
 				}
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -20,7 +20,7 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
-    image: openfga/openfga:v1.1.1
+    image: openfga/openfga:latest
     container_name: migrate
     environment:
       - OPENFGA_DATASTORE_ENGINE=postgres
@@ -29,28 +29,13 @@ services:
     networks:
       - default
 
-  validate:
-    depends_on:
-      postgres:
-        condition: service_healthy
-    image: openfga/openfga:latest
-    container_name: validate
-    environment:
-      - OPENFGA_DATASTORE_ENGINE=postgres
-      - OPENFGA_DATASTORE_URI=postgres://postgres:password@postgres:5432/postgres?sslmode=disable
-    command: validate-models
-    networks:
-      - default
-
   openfga:
     depends_on:
-      validate:
-        condition: service_completed_successfully
       migrate:
         condition: service_completed_successfully
       otel-collector:
         condition: service_started
-    image: openfga/openfga:v1.1.1
+    image: openfga/openfga:latest
     container_name: openfga
     command: run
     environment:
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -20,7 +20,7 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
-    image: openfga/openfga:latest
+    image: openfga/openfga:v1.1.1
     container_name: migrate
     environment:
       - OPENFGA_DATASTORE_ENGINE=postgres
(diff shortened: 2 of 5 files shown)


docker-compose.yaml
```diff
@@ -19,7 +19,7 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
-    image: openfga/openfga:latest
+    image: openfga/openfga:v1.1.1
```
miparnisari, line 22: 🤔 why? latest already includes 1.1.1. The digests are the same: https://hub.docker.com/r/openfga/openfga/tags

docker-compose.yaml
```diff
@@ -28,13 +28,28 @@ services:
     networks:
       - default
 
+  validate:
+    depends_on:
+      postgres:
+        condition: service_healthy
```
miparnisari, line 34:
```suggestion
  validate-models:
    depends_on:
      migrate:
        condition: service_healthy
```

Otherwise this container exits with `2023-06-29 17:14:57 Error: error reading stores: sql error: ERROR: relation "store" does not exist (SQLSTATE 42P01)`

docker-compose.yaml
```diff
@@ -28,13 +28,28 @@ services:
...
+    image: openfga/openfga:latest
+    container_name: validate
```
miparnisari, line 36:
```suggestion
    container_name: validate-models
```

docker-compose.yaml
```diff
@@ -28,13 +28,28 @@ services:
...
   openfga:
     depends_on:
+      validate:
+        condition: service_completed_successfully
       migrate:
         condition: service_completed_successfully
       otel-collector:
         condition: service_started
-    image: openfga/openfga:latest
+    image: openfga/openfga:v1.1.1
```
miparnisari, line 52: same as before, this isn't needed

docker-compose.yaml
```diff
@@ -28,13 +28,28 @@ services:
...
+      - OPENFGA_DATASTORE_ENGINE=postgres
+      - OPENFGA_DATASTORE_URI=postgres://postgres:password@postgres:5432/postgres?sslmode=disable
+    command: validate-models
+    networks:
+      - default
+
   openfga:
     depends_on:
+      validate:
+        condition: service_completed_successfully
```
miparnisari, line 47: what is the goal here, that if there are invalid models the service shouldn't start? i don't think we should do that. what we _could_ do that if the latest model has errors, then the service won't start.

in any case you have to change the `validate-models` command so that if there is at least one error, it throws a non-zero exit code to indicate error, because right now it doesn't work like that

image

pkg/typesystem/typesystem.go
```diff
@@ -34,6 +34,7 @@ var (
 	ErrReservedKeywords      = errors.New("self and this are reserved keywords")
 	ErrCycle                 = errors.New("an authorization model cannot contain a cycle")
 	ErrNoEntrypoints         = errors.New("no entrypoints defined")
+	ErrNoEntryPointsLoop     = errors.New("no entrypoints defined with potential loop")
```
miparnisari, line 37: this error message isn't clear to me, also how is it different from `ErrCycle`'s message? in other words, do we need `ErrNoEntryPointsLoop`?

cmd/validatemodels/validate_models.go
```diff
@@ -135,7 +135,13 @@ func ValidateAllAuthorizationModels(ctx context.Context, db storage.OpenFGADatas
 					}
 
 					if err != nil {
-						validationResult.Error = err.Error()
+						if invalidRelation, ok := err.(*typesystem.InvalidRelationError); ok {
+							// This will give a more detailed error
+							validationResult.Error = invalidRelation.Error() + " : " + invalidRelation.Cause.Error()
```
miparnisari, line 140: i think it's simpler to remove this change and to change 

```go
func (e *InvalidRelationError) Error() string {
	return fmt.Sprintf("the definition of relation '%s' in object type '%s' is invalid", e.Relation, e.ObjectType)
}
```

to 


```go
func (e *InvalidRelationError) Error() string {
	return fmt.Sprintf("the definition of relation '%s' in object type '%s' is invalid: %s", e.Relation, e.ObjectType, e.Cause)
}
```

i  personally want the API to return the details, too

pkg/typesystem/typesystem.go
```diff
@@ -862,12 +865,19 @@ func (t *TypeSystem) validateRelation(typeName, relationName string, relationMap
...
 	}
 
 	if !hasEntrypoints {
+		if loop {
+			return &InvalidRelationError{
+				ObjectType: typeName,
+				Relation:   relationName,
+				Cause:      fmt.Errorf("relation '%s#%s': %w", typeName, relationName, ErrNoEntryPointsLoop),
+			}
+		}
```
miparnisari, line 880:
```suggestion
cause := ErrNoEntrypoints
if loop {
       cause = ErrNoEntryPointsLoop
}
return &InvalidRelationError{
	ObjectType: typeName,
	Relation:   relationName,
	Cause:      cause,
}
```

miparnisari (changes requested):

docker-compose.yaml
adriantam: I didn't intend to check in the docker-compose.yaml file.  Will revert that file.

pkg/typesystem/typesystem.go
adriantam: We need to as the error returned from hasEntrypoints usually indicates some lookup error rather than no entry point.  We probably want to reformat the whole thing..but not now.

pkg/typesystem/typesystem.go
adriantam: updated

pkg/typesystem/typesystem.go
```diff
@@ -35,8 +35,17 @@ var (
 	ErrReservedKeywords      = errors.New("self and this are reserved keywords")
 	ErrCycle                 = errors.New("an authorization model cannot contain a cycle")
 	ErrNoEntrypoints         = errors.New("no entrypoints defined")
+	ErrNoEntryPointsLoop     = errors.New("potential loop in relations definitions")
```
miparnisari, line 38:
```suggestion
	ErrNoEntryPointsLoop     = errors.New("potential loop in relation definitions")
```

miparnisari:
```shell
curl --location 'http://localhost:8080/stores/01H44VBCR64GKJTVFWG8K59V0C/authorization-models' \
--data '{"type_definitions":[{"type":"user","relations":{}},{"type":"repo","relations":{"admin":{"this":{}},"action1":{"intersection":{"child":[{"computedUserset":{"object":"","relation":"admin"}},{"computedUserset":{"object":"","relation":"action2"}},{"computedUserset":{"object":"","relation":"action3"}}]}},"action2":{"intersection":{"child":[{"computedUserset":{"object":"","relation":"admin"}},{"computedUserset":{"object":"","relation":"action3"}},{"computedUserset":{"object":"","relation":"action1"}}]}},"action3":{"intersection":{"child":[{"computedUserset":{"object":"","relation":"admin"}},{"computedUserset":{"object":"","relation":"action1"}},{"computedUserset":{"object":"","relation":"action2"}}]}}},"metadata":{"relations":{"admin":{"directly_related_user_types":[{"type":"user"}]},"action1":{"directly_related_user_types":[]},"action2":{"directly_related_user_types":[]},"action3":{"directly_related_user_types":[]}}}}],"schema_version":"1.1"}'
```

prints

```
{
    "code": "invalid_authorization_model",
    "message": "the definition of relation 'action1' in object type 'repo' is invalid: relation 'repo#action1': potential loop in relations definitions"
}
```

Notice the repetition of the type and relation in the error message :)

miparnisari (approved):

PR #837, opened 2023-06-30
test: fix flakiness from singleflight use

miparnisari:
## Description
On rare occasions, test `TestSingleFlightMemoizedTypesystemResolverFunc` fails. I think the cause is that if the 1st goroutine’s call to fetch from the DB finishes quickly, before the 2nd goroutine needs it, the 2nd goroutine will also do the call.

A repro of the failure, with added print statements:
image

Test `TestSingleFlightFindLatestAuthorizationModelID` never failed on my machine but I think if we happened to add more code to `cachingBackend.FindLatestAuthorizationModelID`, it could start failing too.

## References
Failure https://github.com/openfga/openfga/actions/runs/5426454015/jobs/9868515707#step:5:55


adriantam:
@jon-whit can you take a look at this as you are most familiar with the test TestSingleFlightMemoizedTypesystemResolverFunc

pkg/typesystem/resolver_test.go
```diff
@@ -99,17 +101,17 @@ func TestSingleFlightMemoizedTypesystemResolverFunc(t *testing.T) {
 			Return(&openfgav1.AuthorizationModel{
 				Id:            modelID,
 				SchemaVersion: SchemaVersion1_1,
-			}, nil).Times(1),
+			}, nil).MinTimes(1).MaxTimes(numGoroutines),
```
jon-whit, line 104: Good catch. Tricky little race.

jon-whit (approved):

jon-whit:
Needs a rebase but looks good otherwise.

Issue #828, opened 2023-06-28
If listObjects is called with an overly complex model, it throws 500 instead of 400


miparnisari:
With this model:

```
model
  schema 1.1
type user
type resource
  relations
    define a1: a2
    define a2: a3
    define a3: a4
    define a4: a5
    define a5: a6
    define a6: a7
    define a7: a8
    define a8: a9
    define a9: a10
    define a10: a11
    define a11: a12
    define a12: a13
    define a13: a14
    define a14: a15
    define a15: a16
    define a16: a17
    define a17: a18
    define a18: a19
    define a19: a20
    define a20: a21
    define a21: a22
    define a22: a23
    define a23: a24
    define a24: a25
    define a25: a26
    define a26: a27
    define a27: a28
    define a28: a29
    define a29: a30
    define a30: a31
    define a31: a32
    define a32: a33
    define a33: a34
    define a34: a35
    define a35: a36
    define a36: a37
    define a37: a38
    define a38: a39
    define a39: a40
    define a40: a41
    define a41: a42
    define a42: a43
    define a43: a44
    define a44: a45
    define a45: a46
    define a46: a47
    define a47: a48
    define a48: a49
    define a49: a50
    define a50: a51
    define a51: a52
    define a52: a53
    define a53: a54
    define a54: a55
    define a55: a56
    define a56: a57
    define a57: a58
    define a58: a59
    define a59: a60
    define a60: a61
    define a61: a62
    define a62: a63
    define a63: a64
    define a64: a65
    define a65: a66
    define a66: a67
    define a67: a68
    define a68: a69
    define a69: a70
    define a70: a71
    define a71: a72
    define a72: a73
    define a73: a74
    define a74: a75
    define a75: a76
    define a76: a77
    define a77: a78
    define a78: a79
    define a79: a80
    define a80: a81
    define a81: a82
    define a82: a83
    define a83: a84
    define a84: a85
    define a85: a86
    define a86: a87
    define a87: a88
    define a88: a89
    define a89: a90
    define a90: a91
    define a91: a92
    define a92: a93
    define a93: a94
    define a94: a95
    define a95: a96
    define a96: a97
    define a97: a98
    define a98: a99
    define a99: a100
    define a100: a101
    define a101: a102
    define a102: [user]
```

this Check

```
curl --location 'http://localhost:8080/stores/01H3ZKV2RMXWD908XFTV29FJC3/check' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
[...]

PR #853, opened 2023-07-06
fix: return correct error for list object

adriantam:
## Description
Previously all errors for list objects are classified as internal error. However, we should return the correct error message if it is known.

## References
Close https://github.com/openfga/openfga/issues/828

## Testing
Follow instruction as listed in https://github.com/openfga/openfga/issues/828 and generate error in list object.

Verify that HTTP code 400 is returned with the following message:

```
{
    "code": "authorization_model_resolution_too_complex",
    "message": "Authorization Model resolution required too many rewrite rules to be resolved. Check your authorization model for infinite recursion or too much nesting"
}
```

## Review Checklist


pkg/server/commands/list_objects.go
```diff
@@ -259,6 +261,10 @@ func (q *ListObjectsQuery) Execute(
 
 		case result, channelOpen := <-resultsChan:
 			if result.Err != nil {
+				code := status.Code(result.Err)
```
jon-whit, line 264: We should be using our errors package here instead of the grpc codes directly.

`serverErrors.HandleError("", err)`

adriantam: Updated to use serverErrors.HandleError is possible but treat AuthorizationModelResolutionTooComplex separately

jon-whit (approved):

jon-whit:
@adriantam looks like we have a test flake.

PR #830, opened 2023-06-28
chore(tests): re-add complex list objects test, chunking write tuples into 40

rhamzeh:
## References

- closes #825 

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #855, opened 2023-07-06
chore: add a concurrency limited CheckResolver implementation

jon-whit:
## Description
Implement a `CheckResolver` implementation that limits the total amount of concurrent evaluations that can be in flight at any given time.

Here's a sample screenshot if the concurrency limit is set to 1 (e.g. only 1 path of evaluation in flight at a time).
Screen Shot 2023-07-07 at 10 39 42 AM
As you can see, there are 3 paths of evaluation contained under the 'union' operator, but we evaluate them serially because the overall concurrency limit is set to 1.

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -89,9 +89,30 @@ type checkOutcome struct {
 	err  error
 }
 
+// NewCheckResolver creates a CheckResolver that resolves Checks locally with a concurrency
+// limit applied to the top-level resolution (as opposed to per-branch).
+func NewCheckResolver(
+	ds storage.RelationshipTupleReader,
+	concurrencyLimit uint32,
+) CheckResolver {
+
+	localChecker := &LocalChecker{
+		ds:               ds,
+		concurrencyLimit: concurrencyLimit,
+	}
+
+	localChecker.resolver = MustNewLimitedLocalChecker(
+		WithCheckResolver(localChecker),
+		WithConcurrencyLimit(concurrencyLimit),
+	)
+
+	return localChecker
+}
+
 // LocalChecker implements Check in a highly concurrent and localized manner. The
 // Check resolution is limited per branch of evaluation by the concurrencyLimit.
 type LocalChecker struct {
+	resolver         CheckResolver
 	ds               storage.RelationshipTupleReader
 	concurrencyLimit uint32
 }
@@ -103,6 +124,8 @@ func NewLocalChecker(
 	concurrencyLimit uint32,
 ) *LocalChecker {
 	checker := &LocalChecker{ds: ds, concurrencyLimit: concurrencyLimit}
+	checker.resolver = checker
+
 	return checker
 }
 
@@ -316,7 +339,7 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 // was constructed with.
 func (c *LocalChecker) dispatch(ctx context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
 	return func(ctx context.Context) (*openfgapb.CheckResponse, error) {
-		resp, err := c.ResolveCheck(ctx, req)
+		resp, err := c.resolver.ResolveCheck(ctx, req)
 		if err != nil {
 			return nil, err
 		}
--- a/internal/graph/limited_checker.go
+++ b/internal/graph/limited_checker.go
@@ -1,0 +1,60 @@
+package graph
+
+import (
+	"context"
+	"fmt"
+)
+
+type LimitedLocalChecker struct {
+	resolver CheckResolver
+	limiter  chan struct{}
+}
+
+type LimitedLocalCheckerOpt func(l *LimitedLocalChecker)
+
+func WithCheckResolver(resolver CheckResolver) LimitedLocalCheckerOpt {
+	return func(l *LimitedLocalChecker) {
+		l.resolver = resolver
+	}
+}
+
+func WithConcurrencyLimit(limit uint32) LimitedLocalCheckerOpt {
+	return func(l *LimitedLocalChecker) {
+		l.limiter = make(chan struct{}, limit)
+	}
+}
+
+func MustNewLimitedLocalChecker(opts ...LimitedLocalCheckerOpt) *LimitedLocalChecker {
+	checker, err := NewLimitedLocalChecker(opts...)
+	if err != nil {
+		panic(err)
+	}
+
+	return checker
+}
+
+func NewLimitedLocalChecker(opts ...LimitedLocalCheckerOpt) (*LimitedLocalChecker, error) {
+
+	checker := &LimitedLocalChecker{
+		limiter: make(chan struct{}, 50),
+	}
+
+	for _, opt := range opts {
+		opt(checker)
+	}
+
+	if checker.resolver == nil {
+		return nil, fmt.Errorf("the provided CheckResolver cannot be nil")
+	}
+
+	return checker, nil
+}
+
+func (l *LimitedLocalChecker) ResolveCheck(ctx context.Context, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
+	l.limiter <- struct{}{}
+	defer func() {
+		<-l.limiter
+	}()
+
+	return l.resolver.ResolveCheck(ctx, req)
+}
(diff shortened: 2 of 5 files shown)


internal/graph/limited_checker.go
```diff
@@ -0,0 +1,60 @@
...
+		panic(err)
+	}
+
+	return checker
+}
+
+func NewLimitedLocalChecker(opts ...LimitedLocalCheckerOpt) (*LimitedLocalChecker, error) {
+
+	checker := &LimitedLocalChecker{
+		limiter: make(chan struct{}, 50),
```
jon-whit, line 39: Should make this a const `defaultConcurrencyLimit` or something.

jon-whit:
There is some potential danger with this PR in that it could create a deadlock if a single `ResolveCheck` depends on more than the concurrency limit, and so I'm not sure if this approach is generally viable. For that we'd have no other real option than to look into dispatching.

More investigation and testing on this specific scenario needs to be performed.

internal/graph/limited_checker.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	return checker, nil
+}
+
+func (l *LimitedLocalChecker) ResolveCheck(ctx context.Context, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
+	l.limiter <- struct{}{}
+	defer func() {
+		<-l.limiter
+	}()
```
miparnisari, line 57: This will rate-limit `LimitedLocalChecker.ResolveCheck`, but not `LocalChecker.ResolveCheck`, which is the one that is called recursively

I did a quick load test lasting 1 minute with 100 concurrent Checks, this is the number of goroutines:

image

jon-whit: @miparnisari try it now. I hadn't correctly wired up the CheckResolver so that it would loop back. It should now be limited properly.

internal/graph/limited_checker.go
jon-whit: The lack of channel closure _should_ be fine, because the other branches are cancelled and so the Go garbage collector should be able to reclaim the unused channel. You don't always have to explicitly close a channel in Go. The Go garbage collector can use escape analysis to determine the channel is no longer in use.

> It is only necessary to close a channel if the receiver is looking for a close. Closing the channel is a control signal on the channel indicating that no more data follows.

Closure is a signaling mechanism mostly. Otherwise Go GC can use escape analysis and collect it automatically. If, however, you have goroutines lying around that still reference the channel, then that is considered a memory leak (both because of the dangling goroutines and the channel that is still being referenced by said goroutines).

internal/graph/check.go
```diff
@@ -103,6 +124,8 @@ func NewLocalChecker(
 	concurrencyLimit uint32,
 ) *LocalChecker {
 	checker := &LocalChecker{ds: ds, concurrencyLimit: concurrencyLimit}
+	checker.resolver = checker
```
adriantam, line 127: can you add a few lines explaining this?  I have a hard time wrapping my head around the notion that the checker's resolver is itself.

internal/graph/limited_checker.go
adriantam: Agree on making this value configurable as depending on the database configuration, this value may be different.

internal/graph/check.go
miparnisari: +1. This exact comment came up in the original Check rewrite: https://github.com/openfga/openfga/pull/476#discussion_r1089530447

> can you add a few lines explaining this?

I personally don't think this is enough.. this may be a solution that works and is clever but it's not readable.

jon-whit: > +1. This exact comment came up in the original Check rewrite: https://github.com/openfga/openfga/pull/476#discussion_r1089530447

Let's not forget context. The original intent behind the comment was motivated by the fact that the term "dispatching" was a conflicting topic. At the time we only need a local concurrent CheckResolver, and so the extra wrapped `resolver` was a bit much to swallow at the time. The original code was already written to take into account what this PR introduces. Fact of the matter is, dispatch composition has many uses, and this is one of them.

I don't understand what's so terribly wrong with it? Could you elaborate and be more specific? It may be a bit inception(y), but there isn't anything inherently wrong with it. 

The primary driving factor behind this is that we're delegating to an implementation of an interface, and the interface you happen to be delegating to is yourself if you are only constructing a `LocalChecker`. But since the inner resolver can be dispatched through any implementation of a CheckResolver, then it allows for clever compositional routing through an alternative implementation of a CheckResolver, which for the scope of this PR is a concurrency limited resolver but in the future it could be a remote dispatcher etc..

jon-whit: This is a different way to look at it in simpler terms that may help to understand the _why_ behind the pattern here..

```go
type SomeInterface interface {
  SomeFunc()
}

type SomeStruct struct {
}

func (s *SomeStruct) SomeFunc() {...}

func NewSomeStruct() *SomeStruct {
    return &SomeStruct{}
}

s := NewSomeStruct()
s.SomeFunc()
```
is functionally equivalent to

```go
type SomeInterface interface {
  SomeFunc()
}

type SomeStruct struct {
    delegate SomeInterface
}

func (s *SomeStruct) SomeFunc() {s.delegate.SomeFunc()}

func NewSomeStruct() *SomeStruct {
    s := &SomeStruct{}
    s.delegate = s
    return s
}

s := NewSomeStruct()
s.SomeFunc()
```
The difference is that the first cannot benefit from functional composition chaining whereas the second can. If you have more than one implementation of `SomeInterface`, then you can chain/compose the delegate with those, and that allows for decorating and chaining the interface implementations.

jon-whit: @miparnisari mind taking a stab at what you would do differently?

From a diff standpoint, I think the top goal for me is just to avoid too many changes to the LocalChecker. But if we can do it differently that achieves the same outcome then let's do it.

adriantam: I am not arguing whether it is correct or not correct (likely correct).  However, the fact that it feel `inception(y)` to both @miparnisari and myself suggests we can explain in the code as comments what it does (or why we have resolver = checker) in that it is a dispatching pattern.  This will help others reading the code wrapping their head around it.

jon-whit: @adriantam I have no problem adding a comment. @miparnisari seemed to suggest we change it, and that was the motivation of my responses above.

I'll let ya'll take a stab at it and fix it up as you see fit. This was just to serve as an example.

miparnisari:
I made a test with the following settings:
- The model + tuples from this PR https://github.com/openfga/openfga/pull/830 and this check: `folder:3#can_read@user:harriette`.
- Set `UpstreamTimeout` set to 0s (no timeout)
- Added 1 second artificial delay in `postgres.ReadUserTuple` and `postgres.ReadUsersetTuples`
- Set `scrape_interval: 1s` in Prometheus.yml
- Set checkConcurrencyLimit` set to `1`

The Check took 41 seconds and this is the result for `goroutines`:

image

internal/graph/check.go
miparnisari: I'll post a link here to another version... when/if I actually do get a working version 😆

jon-whit:
Here's an example of a problem that this PR introduces:

Consider the following relationship tuples:
```
document:1#viewer@group:1#member
document:1#viewer@group:2#member

group:1#member@group:1a#member
group:1#member@group:1b#member
group:2#member@group:2a#member
group:2#member@group:2b#member

group:2b#member@user:jon
```

And we call `Check(document:1#viewer@user:jon)`, which we expect to return `{allowed: true}`:

```
ResolveCheck(document:1#viewer@user:jon) // consumed limit is now at 1
    document:1#viewer@group:1#member
    document:1#viewer@group:2#member

    Dispatch(
        ResolveCheck(group:1#member@user:jon) // consumed limit is now at 0
            group:1#member@group:1a#member
            group:1#member@group:1b#member
            
            Dispatch(
                ResolveCheck(group:1#member@group:1a#member) // no more room in the resolution queue
             )
             
             Dispatch(
                ResolveCheck(group:1#member@group:1b#member) // no more room in the resolution queue
             )
    )

    
    Dispatch(
        ResolveCheck(group:2#member@user:jon) // no more room in the resolution queue
            group:2#member@group:2a#member
            group:2#member@group:2b#member
            
            ...
    )
```

We'll experience a deadlock because as we attempt to dispatch subproblems of the graph resolution, we won't have room in the resolution queue to resolve the subproblems.


Closing this PR as a result of the issues it would introduce ☝️

PR #860, opened 2023-07-11
fix: in ListObjects API, Check calls have now a bounded number of concurrent database reads

miparnisari:
## Description

This PR implements a wrapper for the [`RelationshipTupleReader` interface](https://github.com/openfga/openfga/blob/main/pkg/storage/storage.go#L47) that has a channel limiter so that only so many database queries can be in flight per request.

It also measures how much time is spent waiting. For example: for a complex ListObjects call:

image

image

In this PR we use it only for ListObjects API calls, so that one complex ListObjects call won't hog all database connections.

In the future we can introduce this idea in Check API, too.

## References
This is a follow-up of https://github.com/openfga/openfga/pull/855

## Testing

Ran a load test for 10 minutes on both this PR and main: `ab -r -t 600 -c 100 -p body.txt -T application/json http://localhost:8080/stores/01H4SEESD718G51KWP3GHBKNE7/list-objects`

image

### DIFF
--- a/pkg/storage/storagewrappers/boundedconcurrency.go
+++ b/pkg/storage/storagewrappers/boundedconcurrency.go
@@ -8,12 +8,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
-	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 )
-
-var tracer = otel.Tracer("pkg/storage/boundedconcurrency")
 
 var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
 
--- a/pkg/storage/storagewrappers/boundedconcurrency.go
+++ b/pkg/storage/storagewrappers/boundedconcurrency.go
@@ -69,10 +69,6 @@ func (b *boundedConcurrencyTupleReader) ReadUsersetTuples(ctx context.Context, s
 	return b.RelationshipTupleReader.ReadUsersetTuples(ctx, store, filter)
 }
 
-func (b *boundedConcurrencyTupleReader) Close() {
-	close(b.limiter)
-}
-
 func (b *boundedConcurrencyTupleReader) waitForLimiter(ctx context.Context) {
 	start := time.Now()
 
--- a/pkg/storage/storagewrappers/boundedconcurrency.go
+++ b/pkg/storage/storagewrappers/boundedconcurrency.go
@@ -12,14 +12,14 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-const TimeWaitingSpanAttribute = "time_waiting"
+const timeWaitingSpanAttribute = "time_waiting"
 
 var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
 
 var (
-	timeWaitingCounter = promauto.NewHistogram(prometheus.HistogramOpts{
-		Name:    "time_waiting_for_bounded_query",
-		Help:    "Time (in ms) spent waiting for Read, ReadUserTuple and ReadUsersetTuples calls to the datastore",
+	timeWaitingHistogram = promauto.NewHistogram(prometheus.HistogramOpts{
+		Name:    "datastore_bounded_read_delay_ms",
+		Help:    "Time spent waiting for Read, ReadUserTuple and ReadUsersetTuples calls to the datastore",
 		Buckets: []float64{10, 25, 50, 100, 1000, 5000}, // milliseconds. Upper bound is config.UpstreamTimeout
 	})
 )
@@ -30,12 +30,12 @@ type boundedConcurrencyTupleReader struct {
 }
 
 // NewBoundedConcurrencyTupleReader returns a wrapper over a datastore that makes sure that there are, at most,
-// N concurrent calls to Read, ReadUserTuple and ReadUsersetTuples.
+// "concurrency" concurrent calls to Read, ReadUserTuple and ReadUsersetTuples.
 // Consumers can then rest assured that one client will not hoard all the database connections available.
-func NewBoundedConcurrencyTupleReader(wrapped storage.RelationshipTupleReader, N uint32) *boundedConcurrencyTupleReader {
+func NewBoundedConcurrencyTupleReader(wrapped storage.RelationshipTupleReader, concurrency uint32) *boundedConcurrencyTupleReader {
 	return &boundedConcurrencyTupleReader{
 		RelationshipTupleReader: wrapped,
-		limiter:                 make(chan struct{}, N),
+		limiter:                 make(chan struct{}, concurrency),
 	}
 }
 
@@ -80,7 +80,7 @@ func (b *boundedConcurrencyTupleReader) waitForLimiter(ctx context.Context) {
 
 	end := time.Now()
 	timeWaiting := end.Sub(start).Milliseconds()
-	timeWaitingCounter.Observe(float64(timeWaiting))
+	timeWaitingHistogram.Observe(float64(timeWaiting))
 	span := trace.SpanFromContext(ctx)
-	span.SetAttributes(attribute.Int64(TimeWaitingSpanAttribute, timeWaiting))
+	span.SetAttributes(attribute.Int64(timeWaitingSpanAttribute, timeWaiting))
 }
--- a/pkg/storage/storagewrappers/boundedconcurrency.go
+++ b/pkg/storage/storagewrappers/boundedconcurrency.go
@@ -8,7 +8,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
+
+var tracer = otel.Tracer("pkg/storage/boundedconcurrency")
 
 var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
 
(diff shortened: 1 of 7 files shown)


pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,97 @@
...
+	"go.opentelemetry.io/otel/trace"
+)
+
+var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
+
+var (
+	timeWaitingCounter = promauto.NewHistogram(prometheus.HistogramOpts{
+		Name:    "time_waiting_for_read_queries",
+		Help:    "Time (in ms) spent waiting for Read, ReadUserTuple and ReadUsersetTuples calls to the datastore",
+		Buckets: []float64{1, 10, 25, 50, 100, 1000, 5000}, // milliseconds
```
elbuo8, line 21:
```suggestion
		Buckets: []float64{10, 25, 50, 100, 1000}, // milliseconds
```

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,97 @@
...
+	start := time.Now()
+
+	b.limiter <- struct{}{}
+
+	end := time.Now()
+
+	timeWaiting := end.Sub(start).Milliseconds()
+	timeWaitingCounter.Observe(float64(timeWaiting))
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(attribute.Int64("time_waiting", timeWaiting))
```
elbuo8, line 50: `time_wating` can be const.

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,97 @@
...
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
+
+var (
+	timeWaitingCounter = promauto.NewHistogram(prometheus.HistogramOpts{
+		Name:    "time_waiting_for_read_queries",
```
elbuo8, line 19:
```suggestion
		Name:    "time_waiting_for_bounded_query",
```

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,97 @@
...
+	timeWaitingCounter.Observe(float64(timeWaiting))
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(attribute.Int64("time_waiting", timeWaiting))
+
+	defer func() {
+		<-b.limiter
+	}()
+
+	return b.RelationshipTupleReader.ReadUsersetTuples(ctx, store, filter)
+}
```
elbuo8, line 93: Isn't there a idiomatic Go way to handle these as there is only one line diff across these functions?

If there isn't its fine.

miparnisari: If the storage implementations had the same function signature I could pass that as a parameter to `waitForLimiter`. But they don't so.. I changed to the next best thing that I know of :P

pkg/storage/storagewrappers/boundedconcurrency.go
miparnisari: a request can take up to 5000 ms (config.UpstreamTimeout), so i'm inclined to leave that bucket

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,86 @@
...
+
+type boundedConcurrencyTupleReader struct {
+	storage.RelationshipTupleReader
+	limiter chan struct{}
+}
+
+// NewBoundedConcurrencyTupleReader returns a wrapper over a datastore that makes sure that there are, at most,
+// N concurrent calls to Read, ReadUserTuple and ReadUsersetTuples.
+// Consumers can then rest assured that one client will not hoard all the database connections available.
+func NewBoundedConcurrencyTupleReader(wrapped storage.RelationshipTupleReader, N uint32) *boundedConcurrencyTupleReader {
```
jon-whit, line 35:
```suggestion
func NewBoundedConcurrencyTupleReader(wrapped storage.RelationshipTupleReader, concurrencyLimit uint32) *boundedConcurrencyTupleReader {
```

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,86 @@
...
+
+	defer func() {
+		<-b.limiter
+	}()
+
+	return b.RelationshipTupleReader.ReadUsersetTuples(ctx, store, filter)
+}
+
+func (b *boundedConcurrencyTupleReader) Close() {
+	close(b.limiter)
```
jon-whit, line 73: You need to call `Close` on the underlying (e.g. wrapped) `RelationshipTupleReader`.

```go
b.RelationshipTupleReader.Close()
close(b.limiter)
```

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,86 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const TimeWaitingSpanAttribute = "time_waiting"
+
+var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
+
+var (
+	timeWaitingCounter = promauto.NewHistogram(prometheus.HistogramOpts{
```
jon-whit, line 20: This variable name should be a `___Histogram` not `___Counter`

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,86 @@
...
+		Name:    "time_waiting_for_bounded_query",
```
jon-whit, line 21: It's standard practice to name the metric with a suffix indicating the units. 

Also, minor nit, it would be nice to rename this to `datastore_bounded_concurrency_lag_ms` or `datastore_bounded_concurrency_delay_ms` or `datastore_read_limiter_delay_ms` or something similar. `time_waiting_for_bounded_query` isn't clear to me.

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const TimeWaitingSpanAttribute = "time_waiting"
```
jon-whit, line 15: This doesn't need to be publicly exported.

pkg/storage/storagewrappers/boundedconcurrency_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/openfga/openfga/pkg/storage/memory"
+	"github.com/openfga/openfga/pkg/storage/mocks"
+	"github.com/openfga/openfga/pkg/tuple"
+	"github.com/stretchr/testify/require"
+	openfgapb "go.buf.build/openfga/go/openfga/api/openfga/v1"
+)
+
+func TestBoundedConcurrencyWrapper(t *testing.T) {
+	store := ulid.Make().String()
+	t.Logf("create a slow backend that takes 1 second per read call")
```
jon-whit, line 20: This doesn't need to be logged, just leave a comment. It's noisy in the test output otherwise.

jon-whit (changes requested):
Left a few review items

pkg/storage/storagewrappers/boundedconcurrency.go
miparnisari: There is no method `Close()` on `RelationshipTupleReader`. Also, in this PR, we instantiate a new `boundedConcurrencyTupleReader` on every ListObjects request. I wouldn't want to close the underlying datastore on every request

pkg/storage/storagewrappers/boundedconcurrency.go
miparnisari: 😅 oops haha

pkg/storage/storagewrappers/boundedconcurrency.go
miparnisari: I wrote `N` because I wanted the godoc to be clear that N was the parameter. There is no way of having nice godocs that reference parameters :(

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,82 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+const timeWaitingSpanAttribute = "time_waiting"
+
+var _ storage.RelationshipTupleReader = (*boundedConcurrencyTupleReader)(nil)
+
+var (
+	timeWaitingHistogram = promauto.NewHistogram(prometheus.HistogramOpts{
```
jon-whit, line 20: Can we name the variable after the metric name please?

`boundedReadDelayMsHistogram` or `boundedReadDelayHistogram`

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -0,0 +1,82 @@
...
+		Name:    "datastore_bounded_read_delay_ms",
+		Help:    "Time spent waiting for Read, ReadUserTuple and ReadUsersetTuples calls to the datastore",
+		Buckets: []float64{10, 25, 50, 100, 1000, 5000}, // milliseconds. Upper bound is config.UpstreamTimeout
```
jon-whit, line 23: Please configure this as a native histogram so we have better resolution than these predefined buckets, and also we'll want to tighten up the low range of these buckets since hopefully, on avg, the amount of waiting that we'll have will be in the single-digit milliseconds. So I recommend at least having a `1ms`, `3ms`, and `5ms` in the range.

See:
https://promcon.io/2022-munich/talks/native-histograms-in-prometheus/
https://github.com/auth0/sandcastle/blob/main/internal/storage/dynamodb/utils/retryableclient.go#L37-L39

jon-whit (approved):

jon-whit:
Looks like unit tests are failing, but the changes look good otherwise.

miparnisari:
> Looks like unit tests are failing, but the changes look good otherwise.

It's govulncheck! I needed to wait until [this](https://github.com/actions/go-versions/blob/main/versions-manifest.json) file listing all Go versions was updated to include 1.20.6 which has the fix to the vulnerability reported

PR #862, opened 2023-07-12
ci: use govulncheck github action

miparnisari:


jon-whit (approved):

jon-whit:
Looks like the vuln check is failing with this new version, but looks good otherwise.

PR #833, opened 2023-06-29
refactor: change Server constructor to use options builder pattern

jon-whit:
## Description
The options builder pattern is used elsewhere in our codebase and it provides a clean and elegant way to compose the server structure. The changes herein provide a refactor to the Server structure to use the options builder pattern, and it simplifies test setup in some of our test code.

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #864, opened 2023-07-12
feat: add three server settings to control resources used by Check API

miparnisari:
## Description

(It'd be nice if this was merged first: https://github.com/openfga/openfga/pull/833)

Add two new server configuration settings:
- MAX_CONCURRENT_READS_FOR_CHECK controls, ultimately, database connections used
- MAX_CONCURRENT_READS_FOR_LIST_OBJECTS controls, ultimately, database connections used
- RESOLVE_NODE_BREADTH_LIMIT (not too happy with this name but i want it to mimic RESOLVE_NODE_LIMIT since they are kind of similar) controls, ultimately, goroutines spun up

## References
Follow-up PR of https://github.com/openfga/openfga/pull/860

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -12,6 +12,12 @@
             "default": 100,
             "x-env-variable": "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL"
         },
+        "maxConcurrentReads": {
+            "description": "The maximum allowed number of concurrent reads during queries.",
+            "type": "integer",
+            "default": 30,
+            "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS"
+        },
         "changelogHorizonOffset": {
             "description": "The offset (in minutes) from the current time. Changes that occur after this offset will not be included in the response of ReadChanges.",
             "type": "integer",
@@ -19,10 +25,16 @@
             "x-env-variable": "OPENFGA_CHANGELOG_HORIZON_OFFSET"
         },
         "resolveNodeLimit": {
-            "description": "Defines how deeply nested an authorization model can be.",
+            "description": "Defines how deeply nested an authorization model can be before a Check request errors out.",
             "type": "integer",
             "default": 25,
             "x-env-variable": "OPENFGA_RESOLVE_NODE_LIMIT"
+        },
+        "resolveNodeBreadthLimit": {
+            "description": "Defines how many nodes on a given level can be evaluated concurrently in a Check resolution tree.",
+            "type": "integer",
+            "default": 100,
+            "x-env-variable": "OPENFGA_RESOLVE_NODE_BREADTH_LIMIT"
         },
         "listObjectsDeadline": {
             "description": "The timeout deadline for serving ListObjects requests",
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -12,11 +12,17 @@
             "default": 100,
             "x-env-variable": "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL"
         },
-        "maxConcurrentReads": {
-            "description": "The maximum allowed number of concurrent reads during queries.",
+        "maxConcurrentReadsForCheck": {
+            "description": "The maximum allowed number of concurrent reads during Check queries.",
             "type": "integer",
             "default": 30,
-            "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS"
+            "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_CHECK"
+        },
+        "maxConcurrentReadsForListObjects": {
+            "description": "The maximum allowed number of concurrent reads during ListObjects queries.",
+            "type": "integer",
+            "default": 30,
+            "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_LIST_OBJECTS"
         },
         "changelogHorizonOffset": {
             "description": "The offset (in minutes) from the current time. Changes that occur after this offset will not be included in the response of ReadChanges.",
@@ -25,7 +31,7 @@
             "x-env-variable": "OPENFGA_CHANGELOG_HORIZON_OFFSET"
         },
         "resolveNodeLimit": {
-            "description": "Defines how deeply nested an authorization model can be before a Check request errors out.",
+            "description": "Defines how deeply nested an authorization model can be before a query errors out.",
             "type": "integer",
             "default": 25,
             "x-env-variable": "OPENFGA_RESOLVE_NODE_LIMIT"
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -13,13 +13,13 @@
             "x-env-variable": "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL"
         },
         "maxConcurrentReadsForCheck": {
-            "description": "The maximum allowed number of concurrent reads during Check queries.",
+            "description": "The maximum allowed number of concurrent reads in a single Check query.",
             "type": "integer",
             "default": 30,
             "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_CHECK"
         },
         "maxConcurrentReadsForListObjects": {
-            "description": "The maximum allowed number of concurrent reads during ListObjects queries.",
+            "description": "The maximum allowed number of concurrent reads in a single ListObjects query.",
             "type": "integer",
             "default": 30,
             "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_LIST_OBJECTS"
@@ -31,7 +31,7 @@
             "x-env-variable": "OPENFGA_CHANGELOG_HORIZON_OFFSET"
         },
         "resolveNodeLimit": {
-            "description": "Defines how deeply nested an authorization model can be before a query errors out.",
+            "description": "Maximum resolution depth to attempt before throwing an error (defines how deeply nested an authorization model can be before a query errors out).",
             "type": "integer",
             "default": 25,
             "x-env-variable": "OPENFGA_RESOLVE_NODE_LIMIT"
--- a/cmd/run/flags.go
+++ b/cmd/run/flags.go
@@ -137,11 +137,17 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
 		util.MustBindPFlag("maxTypesPerAuthorizationModel", flags.Lookup("max-types-per-authorization-model"))
 		util.MustBindEnv("maxTypesPerAuthorizationModel", "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL", "OPENFGA_MAXTYPESPERAUTHORIZATIONMODEL")
 
+		util.MustBindPFlag("maxConcurrentReads", flags.Lookup("max-concurrent-reads"))
+		util.MustBindEnv("maxConcurrentReads", "OPENFGA_MAX_CONCURRENT_READS", "OPENFGA_MAXCONCURRENTREADS")
+
 		util.MustBindPFlag("changelogHorizonOffset", flags.Lookup("changelog-horizon-offset"))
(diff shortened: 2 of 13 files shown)


cmd/run/flags.go
```diff
@@ -137,12 +137,18 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
 		util.MustBindPFlag("maxTypesPerAuthorizationModel", flags.Lookup("max-types-per-authorization-model"))
 		util.MustBindEnv("maxTypesPerAuthorizationModel", "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL", "OPENFGA_MAXTYPESPERAUTHORIZATIONMODEL")
 
+		util.MustBindPFlag("maxConcurrentReads", flags.Lookup("max-concurrent-reads"))
```
jon-whit, line 140: This belongs under the family of the datastore flags. Please add the datastore prefix.

`--datastore-max-concurrent-reads`.

cmd/run/run.go
```diff
@@ -448,6 +460,10 @@ func ReadConfig() (*Config, error) {
 }
 
 func VerifyConfig(cfg *Config) error {
+	if int(cfg.MaxConcurrentReads) > cfg.Datastore.MaxOpenConns {
```
jon-whit, line 463: Why so strict here? If max_concurrent_reads is higher than max_open_conns there isn't anything inherently wrong with that. It should be a `log.Warn` at most, because it _could_ mean that RPCs will get throttled a bit waiting for connections, but it's not an error.

.config-schema.json
```diff
@@ -12,18 +12,30 @@
...
+        },
         "changelogHorizonOffset": {
             "description": "The offset (in minutes) from the current time. Changes that occur after this offset will not be included in the response of ReadChanges.",
             "type": "integer",
             "default": 0,
             "x-env-variable": "OPENFGA_CHANGELOG_HORIZON_OFFSET"
         },
         "resolveNodeLimit": {
-            "description": "Defines how deeply nested an authorization model can be.",
+            "description": "Defines how deeply nested an authorization model can be before a Check request errors out.",
```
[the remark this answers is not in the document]
miparnisari: 🤔 Expand too? Where? I can't find where we throw an error in that case

cmd/run/flags.go
miparnisari: Hmmm. The `datastore` flag controls config of the SQL driver, but this flag is used in application logic. Unless we change the code so that commands have access to database configuration (which they don't have access to right now), I still need this flag to be part of this struct 

```
type server.Config struct {
	ResolveNodeLimit        uint32
	ResolveNodeBreadthLimit uint32
	MaxConcurrentReads      uint32
	ChangelogHorizonOffset  int
	ListObjectsDeadline     time.Duration
	ListObjectsMaxResults   uint32
	Experimentals           []ExperimentalFeatureFlag
}
```

I could add an interface method here:

```

type OpenFGADatastore interface {
	TupleBackend
	AuthorizationModelBackend
	StoresBackend
	AssertionsBackend
	ChangelogBackend

	IsReady(ctx context.Context) (bool, error)

	Close()

        MaxConcurrentReadsAllowed()
}
```

 WDYT?

cmd/run/run.go
miparnisari: My thinking here was that, say they set `maxOpenConns = 1` and `maxConcurrentReads = 100`. `maxOpenConns` will still dominate, no? And no more than 1 request at a time will go through. So my goal is to not give a false idea of what the performance will be like

cmd/run/flags.go
jon-whit: It controls `storage` layer parameters though (e.g. the `limiter` in `pkg/storage/storagewrappers`).

If we're going to leave as is, I vote we at least make it specific to the endpoint.

`--check-max-concurrent-reads`
`--list-objects-max-concurrent-reads`
`--expand-max-concurrent-read`

because those name make it more clear that it's controlling graph resolution behavior.

cmd/run/run.go
jon-whit: Yeah, but that doesn't need to be an error. Just `log.Warn`. An error implies that the system will not work if you have the values the way they are.

.config-schema.json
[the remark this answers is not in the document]
jon-whit: Sorry, not in Expand. Just in Check and ListObjects. But, in general, our resolution depth limitations apply to any endpoint that involves graph resolution, and so that's the intent behind the comment. It's not just specific to Check.

pkg/server/commands/list_objects.go
```diff
@@ -159,14 +160,15 @@ func (q *ListObjectsQuery) evaluate(
...
 		}()
 
-		limitedTupleReader := storagewrappers.NewBoundedConcurrencyTupleReader(q.Datastore, q.CheckConcurrencyLimit)
+		limitedTupleReader := storagewrappers.NewBoundedConcurrencyTupleReader(q.Datastore, q.MaxConcurrentReads)
 
 		checkResolver := graph.NewLocalChecker(
 			storage.NewCombinedTupleReader(limitedTupleReader, req.GetContextualTuples().GetTupleKeys()),
-			q.CheckConcurrencyLimit,
+			q.ResolveNodeBreadthLimit,
+			q.MaxConcurrentReads,
```
miparnisari, line 168: This code isn't pretty: callers of `NewLocalChecker` shouldn't have to remember to create these storage wrappers.

In a follow-up PR I would do

```go
checkResolver := graph.NewLocalChecker(q.ds,
			q.ResolveNodeBreadthLimit,
			q.MaxConcurrentReads)
```

and then, when calling `checkResolver.Execute`, I do this

```go
c.ds = storage.NewCombinedTupleReader(storagewrappers.NewBoundedConcurrencyTupleReader(c.ds, c.MaxConcurrentReads), req.GetContextualTuples().GetTupleKeys()),
```

jon-whit: Why is the `ds` bound to the `req` in your example? Decorating the incoming request with a deeply internal ref to a database client has code smells. Instead, if you need to decorate the datastore per request (e.g. for contextual tuples), then that's a perfect use for context. I know that's a controversial topic, but it is a good usage for context.

```go
ctx = WithDatastoreContext(storage.NewCombinedTupleReader(ds))

ds := DatastoreFromContext(ctx)
```

Otherwise I'd say the next best would be explicit propagation via parameter passing.

pkg/server/test/server.go
```diff
@@ -7,7 +7,9 @@ import (
 )
 
 const (
-	DefaultResolveNodeLimit = 25
+	DefaultResolveNodeLimit        = 25
+	DefaultResolveNodeBreadthLimit = 100
+	DefaultMaxConcurrentReads      = 30
```
jon-whit, line 12: Sanity check - is this an aggresively small value considering the resolve node breadth limit is more than 3x this value.

If you assume each resolution at a given level requires at least 1 database query, then setting the default max concurrent reads to a number much smaller than the breadth limit will significantly hinder concurrent evaluation.

I would recommend that we either
a) increase the max concurrent reads limit to be greater than the breadth limit
b) decrease the breadth limit to a number that is more reasonable for the aggressive read limit (with an understanding that this will noticeably impact performance).

jon-whit:
Do we have any benchmarks comparing before/after with the defaults? I would expect with these defaults that the performance on average will get worse.

cmd/run/run.go
```diff
@@ -337,13 +352,16 @@ type Config struct {
...
-		MaxTuplesPerWrite:             100,
-		MaxTypesPerAuthorizationModel: 100,
-		ChangelogHorizonOffset:        0,
-		ResolveNodeLimit:              25,
-		Experimentals:                 []string{},
-		ListObjectsDeadline:           3 * time.Second, // there is a 3-second timeout elsewhere
-		ListObjectsMaxResults:         1000,
+		MaxTuplesPerWrite:                100,
+		MaxTypesPerAuthorizationModel:    100,
+		MaxConcurrentReadsForCheck:       30, // same as Datastore.MaxOpenConns
```
jon-whit, line 357: Should we have this value reflect the `--datastore-max-open-conns` if this config is not provided?

e.g. `openfga run --datastore-max-open-conns=100` then that implies `--max-concurrent-reads-for-check=100`

That way we're not hindering/impacting latency with a more restrictive read limit for developers using OpenFGA if they don't have database issues.

pkg/server/commands/list_objects.go
miparnisari: > Why is the ds bound to the req in your example

because i don't want to overwrite `c.ds` on every request, since every request has different contextual tuples

> be explicit propagation via parameter passing.

okay yeah i prefer this :)

pkg/server/test/server.go
miparnisari: Do note that this list of defaults is only for tests. But we have the same numbers in `DefaultConfig()` that applies in production.

I'll go with (b). `MaxOpenConns` has been defaulted to 30 for a long time. `ResolveNodeBreadthLimit` is being introduced here, so I might as well set it to 30 now.

miparnisari: Actually, hold on, `ResolveNodeBreadthLimit` didn't exist before because we were using a hardcoded value of 100 :P

miparnisari:
> Do we have any benchmarks comparing before/after?

There shouldn't be any changes, I basically replaced hardcoded values with server settings 🤔

cmd/run/run.go
rhamzeh: @jon-whit it would be safer to limit them, as you do not 1 intensive check to be blocking all other checks happening at the same time. People who do not have this issue can increase it, but I think this is better as a default.

rhamzeh:
> Latency should be the only impacted metric, no?

RPS should theoretically also increase because the system is not blocked on serving the same check request so it should accept more requests which may be lighter and get processed faster.

.config-schema.json
```diff
@@ -12,18 +12,36 @@
...
             "x-env-variable": "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL"
         },
+        "maxConcurrentReadsForCheck": {
+            "description": "The maximum allowed number of concurrent reads during Check queries.",
+            "type": "integer",
+            "default": 30,
+            "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_CHECK"
+        },
+        "maxConcurrentReadsForListObjects": {
+            "description": "The maximum allowed number of concurrent reads during ListObjects queries.",
```
rhamzeh, line 22:
```suggestion
            "description": "The maximum allowed number of concurrent reads in a single ListObject query.",
```

.config-schema.json
```diff
@@ -12,18 +12,36 @@
             "default": 100,
             "x-env-variable": "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL"
         },
+        "maxConcurrentReadsForCheck": {
+            "description": "The maximum allowed number of concurrent reads during Check queries.",
```
rhamzeh, line 16:
```suggestion
            "description": "The maximum allowed number of concurrent reads during a single Check query.",
```

.config-schema.json
```diff
@@ -12,18 +12,36 @@
...
+        },
         "changelogHorizonOffset": {
             "description": "The offset (in minutes) from the current time. Changes that occur after this offset will not be included in the response of ReadChanges.",
             "type": "integer",
             "default": 0,
             "x-env-variable": "OPENFGA_CHANGELOG_HORIZON_OFFSET"
         },
         "resolveNodeLimit": {
-            "description": "Defines how deeply nested an authorization model can be.",
+            "description": "Defines how deeply nested an authorization model can be before a query errors out.",
```
rhamzeh, line 34:
```suggestion
            "description": "Maximum resolution depth to attempt before throwing an error (defines how deeply nested an authorization model can be before a query errors out).",
```

rhamzeh (approved):
A few suggested changes. But 👍🏽 from me, and thanks!

rhamzeh (approved):

PR #867, opened 2023-07-13
fix: graphs in Grafana dashboard

miparnisari:
## Description
before

image

after 

image

### DIFF
--- a/telemetry/grafana/dashboards/openfga.json
+++ b/telemetry/grafana/dashboards/openfga.json
@@ -103,14 +103,7 @@
             "__systemRef": "hideSeriesFrom",
             "matcher": {
               "id": "byNames",
-              "options": {
-                "mode": "exclude",
-                "names": [
-                  "openfga.v1.OpenFGAService/CreateStore"
-                ],
-                "prefix": "All except:",
-                "readOnly": true
-              }
+              "options": "openfga.v1.OpenFGAService/CreateStore"
             },
             "properties": [
               {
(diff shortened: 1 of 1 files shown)


jon-whit:
When you select a specific timeseries, say for `openfga.v1.OpenFGAService/CreateStore`, then Grafana adds an override to hide all timeseries except the one selected.

When I exported the dashboard originally, then I must have had the `openfga.v1.OpenFGAService/CreateStore` timeseries selected in the p50 and p90 panels. See:

(p50 request metrics) https://github.com/openfga/openfga/blob/main/telemetry/grafana/dashboards/openfga.json#L220-L246
(p90 request metrics) https://github.com/openfga/openfga/blob/main/telemetry/grafana/dashboards/openfga.json#L341-L367

Simply change the `overrides` to `overrides: []`.

telemetry/grafana/dashboards/openfga.json
```diff
@@ -103,14 +103,7 @@
             "__systemRef": "hideSeriesFrom",
```
jon-whit, line 103: Instead of hiding the `CreateStore` series from the panel, just omit the overrides altogether so that all timeseries show by default.

`overrides: []`

This is what is being done in the p99 request latency panel.

Good catch though! That's a bit of an annoyance.

miparnisari: Ah i thought we wanted to keep this filter (for whatever reason). Removed!

adriantam (approved):

PR #872, opened 2023-07-15
refactor: add constructors for checker, list objects, connected objects

miparnisari:
## Description
Update:
- Check
- ListObjects
- ConnectedObjects

## References
Follow-up of https://github.com/openfga/openfga/pull/833

PR #873, opened 2023-07-17
style: naming of connected objects query

miparnisari:
## Description
Follow-up of https://github.com/openfga/openfga/pull/872.

- `ConnectedObjectsCommand` -> `ConnectedObjectsQuery`  (command modifies data, query doesn't)
- `ConnectedObjectsQuery.StreamedConnectedObjects()` -> `ConnectedObjectsQuery.Execute()`  (same pattern as other queries/commands)
- `ListObjectsQuery`'s and `ConnectedObjectsQuery`'s struct fields are now private (same pattern as other queries/commands)
- `ConnectedObjectsQuery.Limit` -> `ConnectedObjectsQuery.MaxResults` (same as ListObjectsQuery)

No functional changes

### DIFF
--- a/pkg/server/commands/connectedobjects/connected_objects.go
+++ b/pkg/server/commands/connectedobjects/connected_objects.go
@@ -21,7 +21,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-var tracer = otel.Tracer("openfga/pkg/server/commands/connectedobjects")
+var tracer = otel.Tracer("openfga/pkg/server/commands/connected_objects")
 
 const (
 	// same values as run.DefaultConfig() (TODO break the import cycle, remove these hardcoded values and import those constants here)
--- a/pkg/server/commands/connectedobjects/connected_objects.go
+++ b/pkg/server/commands/connectedobjects/connected_objects.go
@@ -165,7 +165,7 @@ func (c *ConnectedObjectsQuery) execute(
 	foundObjectsMap *sync.Map,
 	foundCount *uint32,
 ) error {
-	ctx, span := tracer.Start(ctx, "execute", trace.WithAttributes(
+	ctx, span := tracer.Start(ctx, "connectedObjects.execute", trace.WithAttributes(
 		attribute.String("object_type", req.ObjectType),
 		attribute.String("relation", req.Relation),
 		attribute.String("user", req.User.String()),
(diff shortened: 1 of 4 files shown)


pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -14,20 +15,30 @@ import (
...
 	StoreID          string
 	ObjectType       string
 	Relation         string
-	User             isUserRef
+	User             IsUserRef
 	ContextualTuples []*openfgapb.TupleKey
 }
 
-type isUserRef interface {
+type IsUserRef interface {
```
jon-whit, line 41: Would you please add an issue to represent this more natively in our protobufs in `openfga/api`? The only reason it is here is because we don't have a good way to distinguish between different user/subject types in a strongly typed/structured way and so this was a first attempt to do so.

Ideally though we'd have a way, in our protobuf API definition, to distinguish between `user:jon`, `group:eng#member`, and `user:*` user/subject types.

miparnisari: https://github.com/openfga/api/issues/87

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -149,14 +158,14 @@ type ConnectedObjectsResult struct {
...
-func (c *ConnectedObjectsCommand) streamedConnectedObjects(
+func (c *ConnectedObjectsQuery) execute(
 	ctx context.Context,
 	req *ConnectedObjectsRequest,
 	resultChan chan<- *ConnectedObjectsResult,
 	foundObjectsMap *sync.Map,
 	foundCount *uint32,
 ) error {
-	ctx, span := tracer.Start(ctx, "streamedConnectedObjects", trace.WithAttributes(
+	ctx, span := tracer.Start(ctx, "execute", trace.WithAttributes(
```
jon-whit, line 168: Unfortunate we lose the name. `execute` is so generic as a span name it's almost not helpful. I realize it's scoped/namespaced to the package `openfga/pkg/server/commands/connected_objects`, but that requires looking at 2 fields to see what we're dealing with..

miparnisari: Renamed to `connectedObjects.execute`

jon-whit (approved):

Issue #874, opened 2023-07-18
Add a metric: database reads for Check requests


miparnisari:
The goal is to see how "expensive" a Check call is.

PR #875, opened 2023-07-18
feat: count db reads in check

miparnisari:
## Description
This PR changes the Check API. It adds a header in the response with the number of database reads that led to the solution (note: this number does NOT include the database reads that were done **after** the response was sent)

## References
Closes https://github.com/openfga/openfga/issues/874

## Testing

With the complex model & tuples from this PR https://github.com/openfga/openfga/pull/830:

```
[18/07/23 1:00:25] ~ $ curl --location 'http://localhost:8080/stores/01H4SEESD718G51KWP3GHBKNE7/check' \
--data '{
    "tuple_key": {
        "object": "folder:5",
        "relation": "can_read",
        "user": "user:harriette"
    }
}' -v
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> POST /stores/01H4SEESD718G51KWP3GHBKNE7/check?trace=true HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.88.1
> Content-Type: application/json
> Accept: application/json
> Content-Length: 123
>
< HTTP/1.1 200 OK
< Db-Reads: 33
< Openfga-Authorization-Model-Id: 01H5N99SQ8VKXEQR746DD28QS5
< Openfga-Store-Id: 01H4SEESD718G51KWP3GHBKNE7
< Vary: Origin
< X-Request-Id: ab14b8e3-f1de-4f10-a874-d615108d4442
< Date: Tue, 18 Jul 2023 20:00:37 GMT
< Content-Length: 33
<
* Connection #0 to host localhost left intact
{"allowed":true, "resolution":""}%
[18/07/23 1:00:37] ~ $
```

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -91,8 +91,9 @@ const (
 )
 
 type checkOutcome struct {
-	resp *openfgapb.CheckResponse
-	err  error
+	resp    *openfgapb.CheckResponse
+	dbReads uint32
+	err     error
 }
 
 // LocalChecker implements Check in a highly concurrent and localized manner. The
@@ -136,12 +137,12 @@ func NewLocalChecker(ds storage.RelationshipTupleReader, opts ...LocalCheckerOpt
 
 // CheckHandlerFunc defines a function that evaluates a CheckResponse or returns an error
 // otherwise.
-type CheckHandlerFunc func(ctx context.Context) (*openfgapb.CheckResponse, error)
+type CheckHandlerFunc func(ctx context.Context) (*openfgapb.CheckResponse, uint32, error)
 
 // CheckFuncReducer defines a function that combines or reduces one or more CheckHandlerFunc into
 // a single CheckResponse with a maximum limit on the number of concurrent evaluations that can be
 // in flight at any given time.
-type CheckFuncReducer func(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*openfgapb.CheckResponse, error)
+type CheckFuncReducer func(ctx context.Context, concurrencyLimit uint32, dbReads uint32, handlers ...CheckHandlerFunc) (*openfgapb.CheckResponse, uint32, error)
 
 // resolver concurrently resolves one or more CheckHandlerFunc and yields the results on the provided resultChan.
 // Callers of the 'resolver' function should be sure to invoke the callback returned from this function to ensure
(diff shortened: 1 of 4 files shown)


internal/graph/check.go
```diff
@@ -91,8 +91,9 @@ const (
 )
 
 type checkOutcome struct {
-	resp *openfgapb.CheckResponse
```
jon-whit, line 94: Instead of adding an additional return value to the Check signatures, you should add this as additional metadata in the `ResolutionMetadata` structure. The approach taken here won't work when dispatching is introduced later, and the purpose behind the current design is to allow the implementation to be flexible in that regard.

If you add `DatabaseReadCounter` to the `ResolutionMetadata` structure and increment it anytime you do a read, then if we dispatch locally it will work the same as if we dispatch remotely.

pkg/server/server.go
```diff
@@ -349,7 +349,8 @@ func (s *Server) Check(ctx context.Context, req *openfgapb.CheckRequest) (*openf
 	res := &openfgapb.CheckResponse{
 		Allowed: resp.Allowed,
 	}
-
+	s.transport.SetHeader(ctx, "db-reads", strconv.Itoa(int(dbReads)))
```
jon-whit, line 352: This should be put behind a configuration IMO.

If `--debug-trace` or something is enabled then these sub-specs of the API are enabled.

internal/graph/check.go
miparnisari: If I do this, I'll have to  change

```go
	Allowed bool
}
```

to 

```go
	Allowed            bool
	ResolutionMetadata *ResolutionMetadata
}
```

which i'm fine with. But, it will be confusing that the resolution metadata is also in the request object...


```go
	StoreID              string
	AuthorizationModelID string
	TupleKey             *openfgapb.TupleKey
	ContextualTuples     []*openfgapb.TupleKey
	ResolutionMetadata   *ResolutionMetadata
}
```

miparnisari: I feel like this change is a good candidate for a pair programming session

pkg/server/server.go
rhamzeh: This should honor the `trace` in the query params, rather than the `debug-trace` flag

Also, can we change the header to:
- `X-OpenFga-Query-ReadCount`
- or `OpenFga-Query-ReadCount`

adriantam: Query-Read seems a bit redundant.  Thus, I would recommend choosing one or the other.
Also, we should make the name a constant as it may be used later.

PR #880, opened 2023-07-21
feat: count how many datastore calls one Check does

miparnisari:
## Description

This PR changes the Check API implementation. It adds an attribute `datastore_call_count` to the Check trace and the response log with the number of calls to the storage layer that were made.

image

There are two ways of arriving to this number:

- If the solution is "allowed=true", one path was found. `datastore_call_count` will be all the reads made to arrive to the leaf node of that path (including paths that were eventually discarded)
- If the solution is "allowed=false", no paths were found. `datastore_call_count` will be the sum of all the reads in all the paths that had to be evaluated

Caveats ⚠️ :
- this number is NOT the number of actual DB calls made as this is DB-specific.
- this number does NOT include the calls that were done **after** the response was sent
- every contextual tuple that leads to the solution will add +1 to the counter. This isn't a trivial problem to solve so i'm deferring it to later

## References
Closes https://github.com/openfga/openfga/issues/874
Another approach to https://github.com/openfga/openfga/pull/875

## Testing
~~TO DO: make the tests more specific than this and use mocks~~

Unit tests can not be deterministic: I can't guarantee that one Check will be *exactly* X database reads; this will depend on the scheduling of goroutines, which i have no control of. So, i made them like this: `min <= dbReads(test) <= max`


internal/graph/check.go
```diff
@@ -253,8 +263,9 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 				err = result.err
 				continue
 			}
+			dbReads += result.resp.ResolutionMetadata.DatabaseReads
 
-			if !result.resp.GetAllowed() {
```
rhamzeh, line 257: What's the difference?

internal/graph/check.go
```diff
@@ -314,44 +337,76 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 		wg.Done()
 	}()
 
+	var dbReads uint32
 	for i := 0; i < len(handlers); i++ {
 		select {
 		case baseResult := <-baseChan:
 			if baseResult.err != nil {
-				return &openfgapb.CheckResponse{Allowed: false}, baseResult.err
+				return &ResolveCheckResponse{
```
rhamzeh, line 345: If there is an error why do we need to return this? And in case we do, can we make this a var first and return it in the case of err or !allowed?

internal/graph/check.go
miparnisari: Previously, `result.resp` was of type `*openfgav1.CheckResponse` which has that method defined in the generated code

internal/graph/check.go
```diff
@@ -253,8 +264,9 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 				err = result.err
 				continue
 			}
+			dbReads += result.resp.ResolutionMetadata.DatabaseReads
 
-			if !result.resp.GetAllowed() {
+			if !result.resp.Allowed {
```
adriantam, line 269: For this case, do we need to set the result.resp.ResolutionMetadata.DatabaseReads to be dbReads?

Otherwise, the result will only contains the dbRead from the last goroutine that returns the result.

internal/graph/check.go
```diff
@@ -314,44 +336,64 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
...
 
+	var dbReads uint32
 	for i := 0; i < len(handlers); i++ {
 		select {
 		case baseResult := <-baseChan:
 			if baseResult.err != nil {
-				return &openfgapb.CheckResponse{Allowed: false}, baseResult.err
+				return &ResolveCheckResponse{
+					Allowed:            false,
+					ResolutionMetadata: &ResolutionMetadata{},
```
adriantam, line 346: Do we have inconsistent behaviour?  In the union and intersection case, we will still return the dbRead that has been accumulated while we return 0 here.

internal/graph/check.go
```diff
@@ -314,44 +336,64 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
...
+					},
+				}, nil
 			}
+
 		case subResult := <-subChan:
 			if subResult.err != nil {
-				return &openfgapb.CheckResponse{Allowed: false}, subResult.err
+				return &ResolveCheckResponse{
+					Allowed:            false,
+					ResolutionMetadata: &ResolutionMetadata{},
```
adriantam, line 365: Similar to above, do we have inconsistent behaviour as we don't return dbRead that has been accumulated

internal/graph/check.go
```diff
@@ -634,13 +704,17 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
...
+						DatabaseReads: dbReads + req.GetResolutionMetadata().DatabaseReads,
 					},
 				}))
 		}
 
 		if len(handlers) == 0 {
-			return &openfgapb.CheckResponse{Allowed: false}, nil
+			return &ResolveCheckResponse{
+				Allowed:            false,
+				ResolutionMetadata: &ResolutionMetadata{},
```
adriantam, line 716: why don't we return dbReads?

internal/graph/check.go
```diff
@@ -218,21 +220,29 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 				err = result.err
 				continue
 			}
+			dbReads += result.resp.ResolutionMetadata.DatabaseReads
 
-			if result.resp.GetAllowed() {
+			if result.resp.Allowed {
```
jon-whit, line 225: Let's add getters to follow suite with the RPC stub design and the other RPC Reqest/Response patterns and usage of getters.

internal/graph/check.go
```diff
@@ -263,16 +275,26 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 	}
 
 	if err != nil {
-		return &openfgapb.CheckResponse{Allowed: false}, err
+		return &ResolveCheckResponse{
+			Allowed: false,
+			ResolutionMetadata: &ResolutionMetadata{
```
jon-whit, line 280: We're dropping the resolution depth here. If we're going to to respond with ResolutionMetadata we should carry/propagate that info along.

internal/graph/check.go
```diff
@@ -412,27 +452,35 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
 
-		fn1 := func(ctx context.Context) (*openfgapb.CheckResponse, error) {
+		fn1 := func(ctx context.Context) (*ResolveCheckResponse, error) {
 			ctx, span := tracer.Start(ctx, "checkDirectUserTuple", trace.WithAttributes(attribute.String("tuple_key", tk.String())))
 			defer span.End()
 
+			response := &ResolveCheckResponse{
+				Allowed: false,
+				ResolutionMetadata: &ResolutionMetadata{
+					DatabaseReads: req.ResolutionMetadata.DatabaseReads + 1,
```
jon-whit, line 462: getters please.

internal/graph/check.go
```diff
@@ -479,21 +534,23 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 
 			var handlers []CheckHandlerFunc
 			for {
+				response.ResolutionMetadata.DatabaseReads++
```
jon-whit, line 537: I would expect the counter to increase on the query execution, not the iterator. What was the rationale there?

I think we're conflating "dispatched" subproblems with database reads.

internal/graph/check.go
```diff
@@ -591,7 +651,10 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 			tuple.NewTupleKey(object, tuplesetRelation, ""),
 		)
 		if err != nil {
-			return &openfgapb.CheckResponse{Allowed: false}, err
+			return &ResolveCheckResponse{
+				Allowed:            false,
+				ResolutionMetadata: &ResolutionMetadata{},
```
jon-whit, line 656: `var emptyResolutionMetdata = &ResolutionMetadata{}` and then use `emptyResolutionMetadata` here and elsewhere.

pkg/server/server.go
```diff
@@ -335,9 +335,11 @@ func (s *Server) Check(ctx context.Context, req *openfgapb.CheckRequest) (*openf
 		TupleKey:             req.GetTupleKey(),
 		ContextualTuples:     req.ContextualTuples.GetTupleKeys(),
 		ResolutionMetadata: &graph.ResolutionMetadata{
-			Depth: s.resolveNodeLimit,
+			Depth:         s.resolveNodeLimit,
+			DatabaseReads: 0,
 		},
 	})
+	span.SetAttributes(attribute.Int64("db_reads", int64(resp.ResolutionMetadata.DatabaseReads)))
```
jon-whit, line 342: I'd like us to establish some more conventions around naming metadata things for query metadata, because we should in general have these things for each of our query endpoints (e.g. ListObjects, StreamedListObjects, Check etc..).

Can we go with something more like `<method_prefix>_database_query_count`

`check_database_query_count`
`list_objects_database_query_count`
etc..

internal/graph/check.go
miparnisari: This is an error case, so if you have `define diff as a but not b` and  the evaluation of `a` fails, there is no point in tracking the `dbReads` 🤔

internal/graph/check.go
miparnisari: What value should it have?

internal/graph/check.go
adriantam: Since we have no control over whether base/sub arrives first.  It is entirely be possible that we already made a db request for the sub.

In this case, it will analogous to union/intersection where one of the branch fail.  The logic for these cases are that we will return the number of db read

internal/graph/check.go
miparnisari: Huh? I'm counting database reads. A call to `iterator.Next()` is one database read

pkg/server/server.go
miparnisari: Why do we need to add the prefix? The span name already is `Check`.

internal/graph/check.go
miparnisari: I just realized that `iterator.Next()` may not be one database read in Auth0 FGA 😢

jon-whit: Yeah, that's my point, sorry if that wasn't clear. `iterator.Next()` is not always representative of a database query. In this case it's consuming another batch of rows on an existing query cursor. In DynamoDB it would be every time we fetch a new batch of results, etc..

So we just need to be carefully how we evaluate it.

pkg/server/server.go
jon-whit: We don't need the method, I guess I'm just asking for a better name beyond `db_reads`.

Can we rename to `database_query_count`

internal/graph/check.go
miparnisari: Talked with @elbuo8, agreed to count only the calls to Read, ReadUserTuple and ReadUsersetTuple

internal/graph/check.go
miparnisari: Updated 👍

script.sh
jon-whit, line 1: Can we delete this file? I dont' see it used anywhere.

miparnisari: Woops 😅

internal/graph/check.go
jon-whit: That's a good question. We haven't discussed what the expectation should be.. What do you and @adriantam and @rhamzeh think?

My thoughts - If we're counting database queries, we should be counting how many queries it took to find an outcome (regardless of if it was a permit or deny decision). It's also important for us to know how many database queries are involved in searching for a path and never finding one, because these kinds of queries are often more intensive on the app as well (it's more costly to search the whole graph space than to short-circuit and return early).

internal/graph/check.go
jon-whit: Right, we shouldn't be counting the `iter.Next()`. That's what I said above ☝️

internal/graph/check.go
miparnisari: Ok so, for the case of `allowed = true` where one path was found, `Depth` (height in the tree) should be the height of that solution in particular.

For the case of `allowed = false`, we probably examined a lot of paths each with their own depth. I guess it should be either the maximum of all those paths, or zero 🤷 

But in any case, let's do this in another PR.

elbuo8 (approved):

internal/graph/check_test.go
```diff
@@ -119,3 +121,200 @@ func TestCheckWithOneConcurrentGoroutineCausesNoDeadlock(t *testing.T) {
...
+	typedefs := parser.MustParse(`
+	type user
+	type org
+      relations
+		define member: [user] as self
+	type document
+	  relations
+		define a: [user] as self
+		define b: [user] as self
+		define union as a or b
```
adriantam, line 149: can we have 

```
define union_rewrite: union
```

and then for union we have

```
define union_or_ttu as union or ttu or rewrite_union
```

There should be no changes to the values as they are rewrite.

pkg/server/server.go
```diff
@@ -359,9 +359,13 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
 		ContextualTuples:     req.ContextualTuples.GetTupleKeys(),
 		ResolutionMetadata: &graph.ResolutionMetadata{
-			Depth: s.resolveNodeLimit,
+			Depth:              s.resolveNodeLimit,
+			DatastoreCallCount: 0,
 		},
 	})
+
+	grpc_ctxtags.Extract(ctx).Set("datastore_call_count", int64(resp.ResolutionMetadata.DatastoreCallCount))
+	span.SetAttributes(attribute.Int64("datastore_call_count", int64(resp.ResolutionMetadata.DatastoreCallCount)))
```
jon-whit, line 368: Can we please rename to `datastore_query_count` to be consistent with other vernacular we've established in metric names etc..?

internal/graph/graph.go
```diff
@@ -43,6 +43,11 @@ func ResolutionDepthFromContext(ctx context.Context) (uint32, bool) {
 
 type ResolutionMetadata struct {
 	Depth uint32
```
jon-whit, line 45: Newline

internal/graph/graph.go
```diff
@@ -43,6 +43,11 @@ func ResolutionDepthFromContext(ctx context.Context) (uint32, bool) {
...
+	// Number of calls to RelationshipTupleReader.ReadUserTuple + ReadUsersetTuples + Read.
+	// Thinking of a Check as a tree of evaluations:
+	// If the solution is "allowed=true", one path was found. This is the value in the leaf node of that path
+	// If the solution is "allowed=false", no paths were found. This is the sum of all the reads in all the paths that had to be evaluated
+	DatastoreCallCount uint32
```
jon-whit, line 50: Following along with this comment: https://github.com/openfga/openfga/pull/880/files#r1282079519

Can we rename to `DatastoreQueryCount`?

internal/graph/check_test.go
```diff
@@ -119,3 +121,201 @@ func TestCheckWithOneConcurrentGoroutineCausesNoDeadlock(t *testing.T) {
 	require.NoError(t, err)
 	require.True(t, resp.Allowed)
 }
+
+func TestCheckDbReads(t *testing.T) {
```
jon-whit, line 125: Can we name after the field we're testing?

`TestCheckDatastoreQueryCount`

Follows along with https://github.com/openfga/openfga/pull/880/files#r1282079519

internal/graph/check_test.go
```diff
@@ -119,3 +121,201 @@ func TestCheckWithOneConcurrentGoroutineCausesNoDeadlock(t *testing.T) {
...
+			name:       "intersection_of_ttus",
+			check:      tuple.NewTupleKey("document:x", "intersection_of_ttus", "user:maria"),
+			allowed:    true,
+			minDBReads: 4,
+			maxDBReads: 5,
+		},
+	}
+
+	// run the test many times to exercise all the possible DBReads
+	for i := 1; i < 1000; i++ {
```
jon-whit, line 293: 1000 iterations is a lot. It'll lead to slower tests.

5 seems reasonable to me?

internal/graph/check.go
```diff
@@ -221,6 +239,7 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 				err = result.err
 				continue
 			}
+			dbReads += result.resp.GetResolutionMetadata().DatastoreCallCount
 
 			if result.resp.GetAllowed() {
 				return result.resp, nil
```
jon-whit, line 245: Shouldn't the response returned here reference the value of `dbReads`?

Let's say you have a union of 2 check handlers here, the first handler takes 2 database queries and doesn't find an answer and the second handler takes 2 database queries and finds an answer. The result should be 4 total database queries at the time we respond, but this will only return 2.

Seems this should be:
```go
if result.resp.GetAllowed() {
    return &ResolveCheckResponse{
        Allowed: true, 
	    DatastoreCallCount: dbReads,
        },
    }
```

I may be overlooking something though? Could you clarify.

internal/graph/check_test.go
```diff
@@ -119,3 +121,201 @@ func TestCheckWithOneConcurrentGoroutineCausesNoDeadlock(t *testing.T) {
...
+		tuple.NewTupleKey("document:x", "a", "user:maria"),
+		tuple.NewTupleKey("document:x", "b", "user:maria"),
+		tuple.NewTupleKey("document:x", "parent", "org:fga"),
+		tuple.NewTupleKey("org:fga", "member", "user:maria"),
+	})
+	require.NoError(t, err)
+
+	typedefs := parser.MustParse(`
+	type user
+	type org
```
jon-whit, line 142: Newlines between type defs please. Makes it easier to quickly read.

internal/graph/check.go
```diff
@@ -257,6 +282,7 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 				continue
 			}
 
+			dbReads += result.resp.GetResolutionMetadata().DatastoreCallCount
 			if !result.resp.GetAllowed() {
 				return result.resp, nil
```
jon-whit, line 287: Same comment as in https://github.com/openfga/openfga/pull/880/files#r1282110051

pkg/server/server.go
```diff
@@ -359,9 +359,13 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		TupleKey:             req.GetTupleKey(),
 		ContextualTuples:     req.ContextualTuples.GetTupleKeys(),
 		ResolutionMetadata: &graph.ResolutionMetadata{
-			Depth: s.resolveNodeLimit,
+			Depth:              s.resolveNodeLimit,
+			DatastoreCallCount: 0,
```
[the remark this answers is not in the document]
miparnisari: I prefer to leave it here for clarity

internal/graph/check_test.go
miparnisari: The entire test takes 6 seconds to run. I'd rather spend this time and have the test give me confidence

pkg/server/server.go
```diff
@@ -359,9 +359,13 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
 		},
 	})
+
+	grpc_ctxtags.Extract(ctx).Set("datastore_call_count", int64(resp.ResolutionMetadata.DatastoreCallCount))
```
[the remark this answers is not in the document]
miparnisari: I'm not sure how to get that test passing. A unit test doesn't work because it doesn't setup the `grpc_ctxtags` interceptors. An integration test doesn't seem to work either because I lose access to the resulting `ctx` 🤔  i probably need to setup a test like [this](https://github.com/openfga/openfga/blob/0d97ad1996fc3dd3a00fb48b78cf0aa254f606e8/pkg/middleware/requestid/requestid_test.go)

internal/graph/check.go
miparnisari: Honestly I went back and forth with how to calculate the final number. I think at some point I experimented with this idea but I didn't get the tests to pass, most likely because I didn't figure out til later that TTU calculation is different than the rest. (I had the problem that I was over-counting).

I made the changes now, let me know if you agree with the results of the unit tests 👍

internal/graph/check_test.go
jon-whit: Why does 1000 iterations give you any more confidence than 5? It's still random concurrent behavior. The probability doesn't change with an increasing number of iterations. Concurrency oddness is just as likely to happen on iteration 1 as it is iteration 1000.

internal/graph/check_test.go
```diff
@@ -119,3 +121,203 @@ func TestCheckWithOneConcurrentGoroutineCausesNoDeadlock(t *testing.T) {
...
+			allowed:          true,
+			minDBReads:       1, // checkDirectUserTuple needs to run
+			maxDBReads:       2, // maybe checkDirectUsersetTuples also runs
+		},
+		{
+			name:       "union",
+			check:      tuple.NewTupleKey("document:x", "union", "user:maria"),
+			allowed:    true,
+			minDBReads: 1, // checkDirectUserTuple needs to run
+			maxDBReads: 2, // maybe checkDirectUsersetTuples also runs
```
jon-whit, line 205: `maxDBReads` would be up to 4.

`document#a` can have up to 2 queries (one for checkDirectUser and one for checkDirectUserset
`document#b` can have up to 2 queries (same reason as above)

jon-whit: maxDBReads here should always be the same as the worst case (e.g. you don't find a path in the evaluation tree). So this should match Line 211 below in the `union_no_access` test case.

internal/graph/check_test.go
```diff
@@ -119,3 +121,203 @@ func TestCheckWithOneConcurrentGoroutineCausesNoDeadlock(t *testing.T) {
...
+			minDBReads: 3, // one read to find org:fga + (one direct check + userset check) to see if user:jon is a member of org:fga
+			maxDBReads: 3,
+		},
+		// more complex scenarios
+		{
+			name:       "union_and_ttu",
+			check:      tuple.NewTupleKey("document:x", "union_and_ttu", "user:maria"),
+			allowed:    true,
+			minDBReads: 3, // union (1 read) + ttu (2 reads)
+			maxDBReads: 5, // union (2 reads) + ttu (3 reads)
```
jon-whit, line 262: This should be 7, worst case.

union decomposes into `a or b`
* 2 reads worst case for `document#a`
* 2 reads worst case for `document#b`

`document#ttu` has worst case of 3 reads (1 for tupleset lookup + 2 for direct and userset lookups)

4 reads worst case for `document#union` + 3 reads worst case for `document#ttu` is a total of 7 reads.

jon-whit: 🤔 wait, I'm probably wrong because it's intersection. Let me second guess myself..

jon-whit: Ok, correction..

This should be `max(max(union), max(ttu))` 

`max(union) == 4`
`max(ttu) == 3`

So this should be 4, not 5.

jon-whit: Ok, I stand corrected.

Shouldn't this be `max(max(union), max(ttu))`?

If the union evaluates first it could take up to `4`, but if the ttu evaluates first it could take up to `3`. The worst case of the two is due to the union case.

jon-whit: Oh, nvm. I see where you got this from.

`a or b` will resolve in  a max of 2 because `user:maria` has direct access, so one of the paths is redundant. Then we have to evaluate the TTU (because of the intersection) which will produce a up to a maximum of 3 additional reads. So 2 + 3 = 5.

Gotcha 👍

internal/graph/check_test.go
jon-whit: I see the distinction here. It wasn't clear to me.

The reason this is 2 is because with `a or b`  and `user:maria` having either `document#a` or `document#b` access, then one of them will resolve before the other, and so the max is 2.

In the cases below where `user:maria` doesn't have access this goes up to 4 because you have to evaluate `a` and `b` only to find she doesn't have access for either one.

internal/graph/check_test.go
jon-whit: Doesn't need to block, but it would be nice to lower it 👍

internal/graph/check.go
jon-whit: Looks good to me now! Thanks for the updates.

jon-whit (approved):

internal/graph/check_test.go
miparnisari: My thinking was that 5 iterations isn't enough to produce all the possible values of the number of db reads

internal/graph/check.go
```diff
@@ -41,7 +41,24 @@ type ResolveCheckRequest struct {
 }
 
 type ResolveCheckResponse struct {
-	Allowed bool
+	Allowed            bool
+	ResolutionMetadata *ResolutionMetadata
+}
+
+func (r *ResolveCheckResponse) GetAllowed() bool {
```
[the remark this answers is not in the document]
miparnisari: https://github.com/openfga/api/issues/89

PR #885, opened 2023-07-25
feat(tracing): configurable security

matoous:
## Description

Right now, OTEL tracing uses gRPC with the insecure option. This PR makes the insecure option optional. Users are free to opt in for secure traces export using `OPENFGA_TRACE_OTLP_TLS_ENABLED`.

Related PR on `openfga.dev`: https://github.com/openfga/openfga.dev/pull/485

## References

Also see: https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlpmetric/internal/oconf/options.go#L134 which is called from here: https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/otlptracegrpc/client.go#L65 when setting up new otlptracegrpc exporter/client. The `Insecure` configuration is handled internally based on env variables.

## Review Checklist

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -321,6 +321,12 @@
                             "type": "string",
                             "default": "0.0.0.0:4317",
                             "x-env-variable": "OPENFGA_TRACE_OTLP_ENDPOINT"
+                        },
+                        "insecure": {
+                            "description": "Whether to use insecure connectio for the trace collector",
+                            "type": "boolean",
+                            "default": false,
+                            "x-env-variable": "OPENFGA_TRACE_OTLP_INSECURE"
                         }
                     }
                 },
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -322,11 +322,16 @@
                             "default": "0.0.0.0:4317",
                             "x-env-variable": "OPENFGA_TRACE_OTLP_ENDPOINT"
                         },
-                        "insecure": {
-                            "description": "Whether to use insecure connectio for the trace collector",
-                            "type": "boolean",
-                            "default": false,
-                            "x-env-variable": "OPENFGA_TRACE_OTLP_INSECURE"
+                        "tls": {
+                            "type": "object",
+                            "properties": {
+                                "enabled": {
+                                    "description": "Whether to use TLS connection for the trace collector",
+                                    "type": "boolean",
+                                    "default": false,
+                                    "x-env-variable": "OPENFGA_TRACE_OTLP_TLS_ENABLED"
+                                }
+                            }
                         }
                     }
                 },
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -152,6 +152,8 @@ func NewRunCommand() *cobra.Command {
 
 	flags.String("trace-otlp-endpoint", defaultConfig.Trace.OTLP.Endpoint, "the endpoint of the trace collector")
 
+	flags.Bool("trace-otlp-insecure", defaultConfig.Trace.OTLP.Insecure, "use insecure connection for trace collector")
+
 	flags.Float64("trace-sample-ratio", defaultConfig.Trace.SampleRatio, "the fraction of traces to sample. 1 means all, 0 means none.")
 
 	flags.String("trace-service-name", defaultConfig.Trace.ServiceName, "the service name included in sampled traces.")
@@ -290,6 +292,7 @@ type TraceConfig struct {
 
 type OTLPTraceConfig struct {
 	Endpoint string
+	Insecure bool
 }
 
 // PlaygroundConfig defines OpenFGA server configurations for the Playground specific settings.
@@ -414,6 +417,7 @@ func DefaultConfig() *Config {
 			Enabled: false,
 			OTLP: OTLPTraceConfig{
 				Endpoint: "0.0.0.0:4317",
+				Insecure: true,
 			},
 			SampleRatio: 0.2,
 			ServiceName: "openfga",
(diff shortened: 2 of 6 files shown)


pkg/telemetry/tracing.go
```diff
@@ -19,9 +19,10 @@ import (
 
 type TracerOption func(d *customTracer)
 
-func WithOTLPEndpoint(endpoint string) TracerOption {
+func WithOTLPEndpoint(endpoint string, insecure bool) TracerOption {
```
jon-whit, line 22: Please don't overload this Trace option. Make a unique one for this individual option.

e.g.

```go
func WithOTLPInsecure() TracerOption {
    return func(d *customTracer) {
        d.insecure = true
    }
}
```

cmd/run/run.go
```diff
@@ -150,6 +150,8 @@ func NewRunCommand() *cobra.Command {
 
 	flags.String("trace-otlp-endpoint", defaultConfig.Trace.OTLP.Endpoint, "the endpoint of the trace collector")
 
+	flags.Bool("trace-otlp-insecure", defaultConfig.Trace.OTLP.Insecure, "use insecure connection for trace collector")
```
jon-whit, line 153: Please add this to the `config.schema.json` file as well 👍

jon-whit:
@matoous why is this a breaking change? Could you elaborate? It appears you've just introduced a new flag that can be used to control the behavior, but its default behavior is to act as it did before. That shouldn't be breaking.

jon-whit (changes requested):
Left a couple of change requests. Let me know when you've addressed them.

pkg/telemetry/tracing.go
matoous: Done, thanks!

cmd/run/run.go
matoous: I am sorry, I can't find that file. Is it supposed to be in this repository?

matoous:
@jon-whit it's a breaking change because now the `insecure` is the default. With these changes the OTLP export by default uses TLS and disabling it is explicit via this flag/param/OTLP env variable. Meaning whoever updates and relies on insecure connection for exporting the traces will have to configure the connection to be insecure. We could make it the other way around and make `insecure` the default but that seems less _secure_ and also doesn't follow the conventions for OTLP.

miparnisari:
> With these changes the OTLP export by default uses TLS and disabling it is explicit via this flag/param/OTLP env variable

That's not what i understood from reading this change - you set `Insecure: true` in `DefaultConfig()`.  With this PR:

```
$ ./openfga run --help
Run the OpenFGA server.

Usage:
  openfga run [flags]

Flags:


--trace-otlp-insecure                            use insecure connection for trace collector (default true)
```

So if you want to do it securely, you need to explicitly pass `--insecure=false`, no?

cmd/run/run.go
matoous: Haha, thanks! My editor ignored the file as a dotfile 🤦

jon-whit:
My recommendation here would be to just follow suite with the rest of our flags and allow them to provide a `--trace-otlp-tls-enabled` flag which implies `insecure=false`.

If that flag is provided then you don’t pass in Insecure creds.

.config-schema.json
```diff
@@ -310,6 +310,12 @@
                             "type": "string",
                             "default": "0.0.0.0:4317",
                             "x-env-variable": "OPENFGA_TRACE_OTLP_ENDPOINT"
+                        },
+                        "insecure": {
+                            "description": "Whether to use insecure connectio for the trace collector",
+                            "type": "boolean",
+                            "default": false,
+                            "x-env-variable": "OPENFGA_TRACE_OTLP_INSECURE"
```
miparnisari, line 318: Please update `cmd/run/flags.go` to include this new flag, otherwise if you change `image: openfga/openfga:latest` to `build: .` in the `docker-compose.yaml` file and then run `docker-compose up`, it panics on startup

matoous: Done.

matoous:
@miparnisari @jon-whit sorry for delay, updated the PR based on feedback, would you please give it another look?

cmd/run/run_test.go
```diff
@@ -1060,6 +1061,14 @@ func TestDefaultConfig(t *testing.T) {
...
 	require.Equal(t, val.String(), cfg.Trace.ServiceName)
 
+	val = res.Get("properties.trace.properties.OTLP.properties.endpoint.default")
+	require.True(t, val.Exists())
+	require.Equal(t, val.String(), cfg.Trace.OTLP.Endpoint)
+
+	val = res.Get("properties.trace.properties.OTLP.properties.tls.properties.enabled.default")
+	require.True(t, val.Exists())
+	require.Equal(t, val.Bool(), cfg.Trace.OTLP.TLS.Enabled)
+
```
miparnisari, line 1071:
```suggestion
	val = res.Get("properties.trace.properties.otlp.properties.endpoint.default")
	require.True(t, val.Exists())
	require.Equal(t, val.String(), cfg.Trace.OTLP.Endpoint)

	val = res.Get("properties.trace.properties.otlp.properties.tls.properties.enabled.default")
	require.True(t, val.Exists())
	require.Equal(t, val.Bool(), cfg.Trace.OTLP.TLS.Enabled)

```

miparnisari: (basically lowercase the `.otlp` 😅 )

miparnisari:
Can you update the PR description? This isn't a breaking change anymore

cmd/run/run_test.go
matoous: Done.

matoous:
@miparnisari done, thanks!

miparnisari (approved):

Issue #886, opened 2023-07-26
Honor MAX_CONCURRENT_READS_FOR_CHECK


miparnisari:
At this moment, MAX_CONCURRENT_READS_FOR_LIST_OBJECTS is honored, but MAX_CONCURRENT_READS_FOR_CHECK is not.

PR #887, opened 2023-07-26
feat: honor MAX_CONCURRENT_READS_FOR_CHECK

miparnisari:
## Description
Close https://github.com/openfga/openfga/issues/886. Note that there is a change in ListObjects code but the behavior of ListObjects should not be affected.


jon-whit (changes requested):
@miparnisari will you please make the default unbounded or a very large bound (e.g. math.MaxUint64)? This is to keep the existing behavior as it is today. I don't want to introduce a regression for those operators of OpenFGA that do not need to restrict the read concurrency.

pkg/server/server.go
```diff
@@ -124,12 +131,26 @@ func WithListObjectsMaxResults(limit uint32) OpenFGAServiceV1Option {
 	}
 }
 
+// WithMaxConcurrentReadsForListObjects sets a limit on the number of datastore reads that can be in flight for a given Check call.
+// This number should be set depending on the RPS expected for Check and ListObjects APIs, the number of OpenFGA replicas running,
```
jon-whit, line 135: This is for ListObjects, not also Check.

pkg/server/server.go
```diff
@@ -124,12 +131,26 @@ func WithListObjectsMaxResults(limit uint32) OpenFGAServiceV1Option {
...
+// - 1 OpenFGA replica and expected traffic of 1 RPS => set it to 100.
+// - 2 OpenFGA replicas and expected traffic of 1 RPS => set it to 50.
 func WithMaxConcurrentReadsForListObjects(max uint32) OpenFGAServiceV1Option {
 	return func(s *Server) {
 		s.maxConcurrentReadsForListObjects = max
 	}
 }
 
+// WithMaxConcurrentReadsForCheck sets a limit on the number of datastore reads that can be in flight for a given Check call.
+// This number should be set depending on the RPS expected for Check and ListObjects APIs, the number of OpenFGA replicas running,
```
jon-whit, line 148: This is just for Check, not ListObjects.

pkg/server/server.go
```diff
@@ -124,12 +131,26 @@ func WithListObjectsMaxResults(limit uint32) OpenFGAServiceV1Option {
...
+// and the number of connections the datastore allows.
+// E.g. if Datastore.MaxOpenConns = 100 and assuming that each Check call takes 1 second and no traffic to Check API:
+// - 1 OpenFGA replica and expected traffic of 100 RPS => set it to 1.
```
jon-whit, line 151: Same comment as earlier https://github.com/openfga/openfga/pull/887/files#r1276471391

pkg/server/server.go
```diff
@@ -94,12 +94,19 @@ func WithTransport(t gateway.Transport) OpenFGAServiceV1Option {
 	}
 }
 
+// WithResolveNodeLimit sets a limit on the number of recursive calls that one Check or ListObjects call will do.
```
jon-whit, line 97: "will allow" instead of "will do"

internal/graph/check.go
```diff
@@ -355,6 +358,8 @@ func (c *LocalChecker) dispatch(ctx context.Context, req *ResolveCheckRequest) C
 	}
 }
 
+// ResolveCheck resolves a node out of a tree of evaluations. If the depth of the tree has gotten too large,
```
jon-whit, line 361: nit - I'd like us to stop using "node" where possible. It's not at all clear what it refers to within the Zanzibar domain. Can we start referring to these as graph "subproblems" instead?

> ResolveCheck resolves a single graph subproblem out of the larger Check resolution.

miparnisari: > It's not at all clear what it refers to within the Zanzibar domain

I was using graph nomenclature (node, edge, etc) since this in the graph package

jon-whit: We don't really even agree what a "node" represents though. We have no visual representation for it at this time. That's why I mention it. That's why I think "graph subproblem" is more appropriate as it avoids the term node which doesn't have a concrete representation right now.

How do others feel about this common nomenclature? I'd like us to start converging on some terms so it's easier to communicate these things both internally and externally.

@adriantam @rhamzeh 

Maybe we can just avoid graph nomenclature altogether though. Instead say something like

> ResolveCheck resolve a single evaluation path out of a tree of evaluation paths. ....

pkg/server/server.go
miparnisari: Right, but it's important to account for traffic for both when setting these values

pkg/server/server.go
```diff
@@ -124,12 +131,26 @@ func WithListObjectsMaxResults(limit uint32) OpenFGAServiceV1Option {
 	}
 }
 
+// WithMaxConcurrentReadsForListObjects sets a limit on the number of datastore reads that can be in flight for a given Check call.
+// This number should be set depending on the RPS expected for Check and ListObjects APIs, the number of OpenFGA replicas running,
+// and the number of connections the datastore allows.
+// E.g. if Datastore.MaxOpenConns = 100 and assuming that each ListObjects call takes 1 second and no traffic to Check API:
+// - 1 OpenFGA replica and expected traffic of 100 RPS => set it to 1.
```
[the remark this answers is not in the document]
miparnisari: I'm torn on this one... because i'd like people to have concrete advice on how to tune this number

internal/graph/check.go
adriantam: I don't know if PR is the best avenue to converge on terms. Let's take this discussion outside of this PR

pkg/server/server.go
```diff
@@ -94,12 +94,20 @@ func WithTransport(t gateway.Transport) OpenFGAServiceV1Option {
...
 	return func(s *Server) {
 		s.resolveNodeLimit = limit
 	}
 }
 
+// WithResolveNodeBreadthLimit sets a limit on the number of goroutines that can be created
+// when evaluating a subtree of a Check or ListObjects call.
+// Thinking of a Check request as a tree of evaluations, this option controls,
+// on a given level of the tree, the maximum number of nodes that can be evaluated concurrently (the breadth).
+// If your authorization models are very complex, you should set this option to be a low number (e.g. 1000)
```
adriantam, line 110: > If your authorization models are very complex, you should set this option to be a low number (e.g. 1000)

Should this number related to the server/database capability? The way it is phrased suggest there is a need to tweak this value as you have more/less authorization models. In some case, the server operator has no control on the model complexity.

pkg/server/server.go
miparnisari: Right, but it's important to account for traffic for both when setting these values

pkg/server/server.go
miparnisari: I'm not exactly sure what is the relation between goroutines that you can create and CPU or memory of the OpenFGA server, that's why i left it ambiguous for now.

internal/graph/check.go
miparnisari: If anything i think the doc I added is better than having nothing 😆

.config-schema.json
```diff
@@ -15,7 +15,7 @@
         "maxConcurrentReadsForCheck": {
             "description": "The maximum allowed number of concurrent reads in a single Check query.",
             "type": "integer",
-            "default": 30,
+            "default": 4294967295,
```
jon-whit, line 18: Can we add comment that this is MaxUint32?

internal/graph/check.go
jon-whit: Fair point. My motivation was just trying to get us to have common vernacular, not because I care so much about `internal` godoc strings.

pkg/server/server.go
adriantam: In this case, probably just remove the last line `If your authorization models are very complex, you should set this option to be a low number (e.g. 1000)` as it may cause more confusion than intended

jon-whit: It's not just models, it's the data (e.g. the tuples). This doc needs to account for that as well.

> If your authorization models have a large number of relationship rewrites for a single relation or if you have a large degree of breadth in your relationship tuples, then you should set this option to bound the concurrency appropriately.

adriantam (dismissed):

pkg/server/server.go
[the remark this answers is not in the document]
jon-whit: It's a non-blocker, but please fix the numbering. Right now it's 1, 1, 2 

Did you mean 1, 2, 3? Or maybe we just drop the numbers, since it's not really ordered and it's just a bulleted list.

pkg/server/server.go
jon-whit: Change to just bulleted list instead of numbered list.

.config-schema.json
miparnisari: A comment in a JSON file? No :/

elbuo8: Put it in the description line.

pkg/server/server.go
miparnisari: updated

.config-schema.json
miparnisari: I feared that this description got printed to customers when using `openfga run --help`, but no, that is taken from somewhere else :) added

jon-whit: > A comment in a JSON file? No :/
I meant in the description e.g. `(defaults to max uint32)` 👍

pkg/server/server.go
miparnisari: It's not a numbered list, it's "one openfga replica" or "two openfga replicas"... i'll use words instead of numbers

jon-whit: Oh, lol that confused the shit out of me..

jon-whit (approved):

PR #891, opened 2023-07-27
feat: add CachedCheckResolver for caching Check subproblems

### LINKED ISSUE #889: Cache Query Subproblems
It's not uncommon for a query (e.g. Check, ListObjects, etc..) to have multiple paths of evaluation that evaluate similar subproblems. For example, consider the following

## Model
```
type user

type group
  relations
    define member: [user]

type folder
  relations
    define viewer: [group#member]

type document
  relations
    define allowed: [group#member]
    define viewer: viewer from parent and allowed
```

## Tuples
```
- document:1#parent@folder:1
- document:1#allowed@group:eng#member
- folder:1#viewer@group:eng#member
- group:eng#member@user:jon
```

`Check(document:1#viewer@user:jon)` will involve 2 identical lookups for `group:eng#member@user:jon` (one in the `folder:1#viewer` evaluation path and one in the `document:1#allowed` evaluation path).

Instead of evaluating `group:eng#member@user:jon` two times, we should cache subproblems that we've evaluated and attempt to use prior evaluations so as to avoid recomputation of subproblems we've already visited.

---
Similarly, two different query requests could result in similar subproblem evaluations, which would also benefit from prior cached subproblems. For example, following along with the model above, we may have the following tuples:

```
- document:1#allowed@group:eng#member
- document:1#parent@folder:X
- document:2#allowed@group:eng#member
- document:2#parent@folder:Y
- folder:X#viewer@group:eng#member
- folder:Y#viewer@group:eng#member
```
`Check(document:1#viewer@user:jon)` and `Check(document:2#viewer@user:jon)` both require evaluation of `group:eng#member@user:jon`. If you've already evaluated that subproblem in the recent past, then avoiding re-evaluation of that subproblem will reduce database contention and will improve performance (albeit at the expense of a configurable read staleness policy).

----

## Recommended Changes
Take these changes how you will, but they are just to provide some suggestions for a starting point of this work.

* Add a `CachedCheckResolver` implementation of the `graph.CheckResolver` interface - the implementation should delegate to another `CheckResolver` delegate if the resolution has not been cached for the provided CheckRequest. Otherwise it should return the cached result.
* Introduce a new experimental flag `--experimentals=query-subproblem-cache` to allow this behavior to be experimentally enabled
* Introduce a new config flag `--query-subproblem-cache-ttl` that bounds how long a query subproblem can be cached for
* Introduce a new config flag `--query-subproblem-cache-limit` that bounds the size of how large the subproblem cache can be
* Introduce some kind of metric to track the cache hit ratio for cached subproblems so that we can measure how impactful this change is. e.g. `dispatched_check_count` + `dispatched_check_cached_count`. The ratio `dispatched_check_cached_count / dispatched_check_count` is the cache hit ratio.

  Increment the `dispatched_check_count` on every `ResolveCheck` and only increment the `dispatched_check_cached_count` when the result is served from a cached subproblem.

jon-whit:
## Description
This introduces an implementation of the `CheckResolver` interface and wires up dispatch delegation in a way that caches Check subproblems between invocation of `ResolveCheck` calls.

## References
Closes #889 

## Review Checklist
- [x] I have added documentation for new/changed functionality in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) -- [chore: add check-query-cache as experimental flag](https://github.com/openfga/openfga.dev/pull/488)


jon-whit:
- [ ] todo(@jon-whit ) - add tests
- [ ] todo(@jon-whit) - wire up the ListObjects server RPC handler to use the CachedCheckResolver as well.
- [ ] todo(@jon-whit) - wire up the server flags to configure the Check query cache TTL and Check cache size limit.

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,128 @@
...
+// checkRequestCacheKey converts the ResolveCheckRequest into a canonical cache key that can be
+// used for Check resolution cache key lookups.
+//
+// We use the 'gob' package to produce a determinstic byte ordering for the contextual tuples provided
+// in the request body. The same tuple provided with the same contextual tuples should produce the same
+// cache key.
+func checkRequestCacheKey(req *ResolveCheckRequest) (string, error) {
+
+	var b bytes.Buffer
+	if err := gob.NewEncoder(&b).Encode(req.GetTupleKey()); err != nil {
```
jon-whit, line 114: Verify that this will produce a stable encoding even if the underlying protobuf message evolves.

We may need to generate a canonical representation of a TupleKey explicitly by serializing the `(object, relation, user)` bit and concatenating that with some deterministic and canonical representation of the slice of ContextualTuples (maybe we sort the slice and then .


See:
https://gist.github.com/kchristidis/39c8b310fd9da43d515c4394c3cd9510
https://github.com/golang/protobuf/issues/1121


Also,

> Proto serialization is not stable, thus you can't rely on marshalling and hashing the output to produce the same hash for the same message.

From https://developers.google.com/protocol-buffers/docs/reference/go/faq#hash

> How do I use a protocol buffer message as a hash key?

> You need canonical serialization, where the marshaled output of a protocol buffer message is guaranteed to be stable over time. Unfortunately, no specification for canonical serialization exists at this time. You'll need to write your own or find a way to avoid needing one.

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,128 @@
...
+// WithCacheTTL sets the TTL (as a duration) for any single Check cache key value.
+func WithCacheTTL(ttl time.Duration) CachedCheckResolverOpt {
+	return func(ccr *CachedCheckResolver) {
+		ccr.cacheTTL = ttl
+	}
+}
+
+func WithExistingCache(cache *ccache.Cache[*ResolveCheckResponse]) CachedCheckResolverOpt {
+	return func(ccr *CachedCheckResolver) {
+		ccr.cache = cache
```
miparnisari, line 53: Here, you should call `ccr.cache.Stop` before re-assigning

internal/graph/cached_resolver.go
miparnisari: Interesting..

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *CachedCheckResolver) ResolveCheck(
+	ctx context.Context,
+	req *ResolveCheckRequest,
+) (*ResolveCheckResponse, error) {
+
+	dispatchedCheckCounter.Inc()
+
+	cacheKey, err := checkRequestCacheKey(req)
+	if err != nil {
+		return nil, fmt.Errorf("cache key computation failed with error: %w", err)
```
miparnisari, line 87: if we can't compute the cache key we should avoid the cache altogether, no?

pkg/server/server.go
```diff
@@ -329,7 +356,15 @@ func (s *Server) Check(ctx context.Context, req *openfgapb.CheckRequest) (*openf
...
-	resp, err := checkResolver.ResolveCheck(ctx, &graph.ResolveCheckRequest{
+	cachedCheckResolver := graph.NewCachedCheckResolver(
+		checkResolver,
+		graph.WithExistingCache(s.checkCache),
+		graph.WithCacheTTL(s.checkQueryCacheTTL),
+	)
+
+	checkResolver.SetDelegate(cachedCheckResolver)
+
+	resp, err := cachedCheckResolver.ResolveCheck(ctx, &graph.ResolveCheckRequest{
```
miparnisari, line 367: This should be behind an experimental flag

adriantam: Instead of experimental flag, I would argue that this should be behind configuration (i.e., allow user to choose whether to enable or not)

internal/graph/cached_resolver.go
adriantam: In theory we can avoid the cache.  However, in reality, I don't see any scenario where the cache key fails to be computed - there is something seriously weird and it is better to return error at this point.

pkg/server/server.go
jon-whit: It should be both enabled with an experimental flag AND configurable. The reason we put it behind the experimental flag is to express that it is not yet ready for production adoption. When we move it out of experimental we are communicating that we're confident in using it in production scenarios.

internal/graph/mock_resolver.go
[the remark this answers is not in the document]
adriantam: Done

cmd/run/run.go
```diff
@@ -178,6 +178,8 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Uint32("listObjects-max-results", defaultConfig.ListObjectsMaxResults, "the maximum results to return in non-streaming ListObjects API responses. If 0, all results can be returned")
 
+	//ResolveCheckCacheConfig
```
adriantam, line 181: Missing TTL etc.

cmd/run/flags.go
```diff
@@ -157,5 +157,9 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
 
 		util.MustBindPFlag("listObjectsMaxResults", flags.Lookup("listObjects-max-results"))
 		util.MustBindEnv("listObjectsMaxResults", "OPENFGA_LIST_OBJECTS_MAX_RESULTS", "OPENFGA_LISTOBJECTSMAXRESULTS")
+
+		util.MustBindPFlag("resolveCheckCacheConfig.enabled", flags.Lookup("resolve-check-cache-config-enabled"))
+		util.MustBindEnv("resolveCheckCacheConfig.enabled", "OPENFGA_RESOLVE_CHECK_CACHE_CONFIG_ENABLED")
+
```
adriantam, line 163: missing bind flag for esolveCheckCacheConfig

pkg/server/server.go
adriantam: Added experimental flag

internal/graph/cached_resolver.go
miparnisari: Whatever way we decide, as long as we call `stop()` on the default cache if the user passed an existing cache, i'm happy :)

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"github.com/stretchr/testify/require"
+)
+
+func TestResolveCheck(t *testing.T) {
```
miparnisari, line 13:
```suggestion
func TestResolveCheckFromCache(t *testing.T) {
```

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// if the tuple is different, it should result in fetching from cache
+	tests := []struct {
+		_name   string
+		req     *ResolveCheckRequest
+		prepare func(mock *MockCheckResolver, request *ResolveCheckRequest)
+	}{
+		{
+			// same signature means data will be taken from cache
+			_name: "same_signature",
```
miparnisari, line 35:
```suggestion
			_name: "same_request_returns_results_from_cache",
```

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+					User:     "user:XYZ",
+				},
+			},
+			prepare: func(mock *MockCheckResolver, request *ResolveCheckRequest) {
+				// there should be no call
+			},
+		},
+		{
+			// different store means data is not from cache
+			_name: "different_store",
```
miparnisari, line 51:
```suggestion
			_name: "request_for_different_store_does_not_return_results_from_cache",
```

same suggestion for the remaining tests

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+				TupleKey: &openfgav1.TupleKey{
+					Object:   "document:abc",
+					Relation: "reader",
+					User:     "user:AAA",
+				},
+			},
+			prepare: func(mock *MockCheckResolver, request *ResolveCheckRequest) {
+				mock.EXPECT().ResolveCheck(ctx, request).Times(1).Return(result, nil)
+			},
+		},
```
miparnisari, line 128: no test for the contextual tuples?

internal/graph/cached_resolver.go
miparnisari: See https://protobuf.dev/getting-started/cpptutorial/#parsing-serialization

image

We should (in another PR that is a pre-requisite of this one) wrap the `openfgav1.CheckRequest` object into a `openfgav1.ExtendedCheckRequest` that defines the `Equals` method.

I've been wanting to do something similar with `openfgav1.AuthorizationModel`. Since `Typesystem` is basically an extended AuthorizationModel with validation and parsing methods, I've been wanting to stop referencing `openfgav1.AuthorizationModel` in our codebase and replace it with `Typesystem`

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,129 @@
...
+// in the request body. The same tuple provided with the same contextual tuples should produce the same
+// cache key.
+func checkRequestCacheKey(req *ResolveCheckRequest) (string, error) {
+
+	var b bytes.Buffer
+	if err := gob.NewEncoder(&b).Encode(req.GetTupleKey()); err != nil {
+		return "", err
+	}
+
+	contextualTuplesCacheKey := b.String()
```
miparnisari, line 119: This variable name is incorrect - it is not the cache key to the contextual tuples

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+			defer ctrl.Finish()
+
+			initialMockResolver := NewMockCheckResolver(ctrl)
+			initialMockResolver.EXPECT().ResolveCheck(ctx, req).Times(1).Return(result, nil)
+
+			// expect first call to result in actual resolve call
+			dut := NewCachedCheckResolver(initialMockResolver)
+			actualResult, err := dut.ResolveCheck(ctx, req)
+			require.Equal(t, result, actualResult)
+			require.Equal(t, nil, err)
```
miparnisari, line 145:
```suggestion
			require.Nil(t, err)
```

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+				StoreID:              "12",
+				AuthorizationModelID: "33",
+				TupleKey: &openfgav1.TupleKey{
+					Object:   "document:abc",
+					Relation: "reader",
+					User:     "user:XYZ",
+				},
+			},
+			prepare: func(mock *MockCheckResolver, request *ResolveCheckRequest) {
+				// there should be no call
```
miparnisari, line 46:
```suggestion
				mock.EXPECT().ResolveCheck(ctx, request).Times(0).Return(result, nil)
```

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,192 @@
...
+			User:     "user:XYZ",
+		},
+	}
+	result := &ResolveCheckResponse{Allowed: true}
+
+	// if the tuple is different, it should result in fetching from cache
+	tests := []struct {
+		_name   string
+		req     *ResolveCheckRequest
+		prepare func(mock *MockCheckResolver, request *ResolveCheckRequest)
```
miparnisari, line 31:
```suggestion
		setTestExpectations func(mock *MockCheckResolver, request *ResolveCheckRequest)
```

internal/graph/graph.go
```diff
@@ -23,6 +25,18 @@ var (
...
+var (
+	dispatchedCheckCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "dispatched_check_count",
+		Help: "The total number of dispatched ResolveCheck calls made for a single Check resolution.",
+	})
+
+	dispatchedCheckCachedCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "dispatched_check_cached_count",
+		Help: "The total number of dispatched ResolveCheck calls that were for a single Check resolution.",
+	})
```
miparnisari, line 37:
```suggestion
	checkCacheTotalCounter = promauto.NewCounter(prometheus.CounterOpts{
		Name: "check_cache_total_count",
		Help: "The total number of calls to ResolveCheck",
	})

	checkCacheHitCounter = promauto.NewCounter(prometheus.CounterOpts{
		Name: "check_cache_hit_count",
		Help: "The total number of cache hits for ResolveCheck",
	})
```

pkg/server/test/list_objects.go
```diff
@@ -194,9 +196,15 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
...
 
+			checkCache := ccache.New(
+				ccache.Configure[*graph.ResolveCheckResponse]().MaxSize(100),
+			)
+
 			opts := []commands.ListObjectsQueryOption{
 				commands.WithListObjectsMaxResults(test.maxResults),
 				commands.WithListObjectsDeadline(test.listObjectsDeadline),
+				commands.WithCheckCache(checkCache),
+				commands.WithCheckQueryCacheTTL(1 * time.Millisecond),
```
miparnisari, line 207: why do we need this change? the existing tests should still pass without it, since the caching is optional. Also note how the other options set here depend on each `test` :)

i think we are we missing new tests

pkg/server/server.go
```diff
@@ -260,13 +308,24 @@ func (s *Server) StreamedListObjects(req *openfgav1.StreamedListObjectsRequest,
...
 	}
 
+	var checkCache *ccache.Cache[*graph.ResolveCheckResponse]
+
+	if slices.Contains(s.experimentals, ExperimentalCacheResolveCheck) {
+		if s.checkQueryCacheEnabled {
+			checkCache = s.checkCache
+			// otherwise when we pass nil to the list object query, the list query will not use cache
+		}
+	}
```
miparnisari, line 318:
```suggestion
	if slices.Contains(s.experimentals, ExperimentalCacheResolveCheck) && s.checkQueryCacheEnabled {
		checkCache = s.checkCache
	}
```

pkg/server/server.go
```diff
@@ -189,12 +219,20 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+		checkQueryCacheTTL:     defaultCheckQueryCacheTTL,
 	}
 
 	for _, opt := range opts {
 		opt(s)
 	}
 
+	s.checkCache = ccache.New(
+		ccache.Configure[*graph.ResolveCheckResponse]().MaxSize(int64(s.checkQueryCacheLimit)),
+	)
```
miparnisari, line 234: Someone that opts out of this new feature will not want the memory usage increase

miparnisari (changes requested):

pkg/server/server.go
adriantam: I purposely split them as the experimental flag will be removed

internal/graph/cached_resolver.go
adriantam: > Or some other option that doesn't surface this error to users but we still know about it..

This is not possible as OpenFGA team is not the operator.

cmd/run/flags.go
```diff
@@ -157,5 +157,14 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
 
 		util.MustBindPFlag("listObjectsMaxResults", flags.Lookup("listObjects-max-results"))
 		util.MustBindEnv("listObjectsMaxResults", "OPENFGA_LIST_OBJECTS_MAX_RESULTS", "OPENFGA_LISTOBJECTSMAXRESULTS")
+
+		util.MustBindPFlag("resolveCheckCacheConfig.enabled", flags.Lookup("resolve-check-cache-config-enabled"))
+		util.MustBindEnv("resolveCheckCacheConfig.enabled", "OPENFGA_RESOLVE_CHECK_CACHE_CONFIG_ENABLED")
```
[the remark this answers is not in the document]
adriantam: For enable flag, we can have two settings (though I am a bit hesitant as the number of flags and knobs are getting rather large)

pkg/server/test/list_objects.go
adriantam: We don't - but I want to exercise having list object calls with cache.

internal/graph/cached_resolver.go
adriantam: We should do that - however, I think it is orthogonal to the caching work here.  Let's not bloat the scope of this PR.

internal/graph/cached_resolver.go
adriantam: actually, there is a bug above - we didn't really encode the contextual tuple key.  We need to encode both the tuple key and contextual tuple key.  Otherwise, the contextual tuple will not be taken into consideration

internal/graph/cached_resolver_test.go
adriantam: Added

pkg/server/test/list_objects.go
adriantam: Added new test

.config-schema.json
```diff
@@ -61,7 +61,7 @@
             "type": "array",
             "items": {
                 "type": "string",
-                "enum": []
+                "enum": ["cache-resolve-check"]
```
jon-whit, line 64: `check-query-cache`

.config-schema.json
```diff
@@ -349,6 +349,30 @@
                     "x-env-variable": "OPENFGA_METRICS_ENABLE_RPC_HISTOGRAMS"
                 }
             }
+        },
+        "resolveCheckCache": {
```
jon-whit, line 353: I know we've been back and forth, but `ResolveCheck` is a deeply internal interface. Please keep this more generic.

I suggest `--check-query-cache-enabled`, `--check-query-cache-limit`, `--check-query-cache-ttl`.

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,215 @@
...
+			Object:   "document:abc",
+			Relation: "reader",
+			User:     "user:XYZ",
+		},
+	}
+	result := &ResolveCheckResponse{Allowed: true}
+
+	// if the tuple is different, it should result in fetching from cache
+	tests := []struct {
+		_name               string
```
jon-whit, line 29: Please drop the `_` prefix here. `name` suffices.

internal/graph/cached_resolver.go
adriantam: Updated just to swallow + log

internal/graph/cached_resolver.go
adriantam: Implemented the pattern as @jon-whit discussed above

internal/graph/cached_resolver.go
jon-whit: I actually don't agree with swallow + log in this case. If we cannot compute the cache key, then that indicates something much more severely wrong (maybe tuplekey validation is broken in our code etc..). If we fail to compute a cache key, then there is an invariant that is not being met and we should seriously fix it. Failure to fix the issue can continue to lead to bypassing the cache, which will degrade performance unnecessarily and void the benefit of the cache in the first place. In such cases, it truly is an internal error worth noting and causing server errors for. We want to strongly encourage/enforce an invariant that requires this never fail. For that reason, I want to raise awareness to an error such as this very quickly, and logging an error isn't going to drive us to fix the issue faster that the alternatives. I would vote that we return an internal error and log the underlying error. Second vote would be panic, though I feel that's a bit drastic in this case within the request lifecycle.

jon-whit: If we were communicating with some remote cache server (e.g. Redis) and there was some transient network issue that caused an error, then I would agree that we should log the error and evaluate the request. But this cache key computation is on a fixed value that should have strong invariants met, and so it should be guaranteed. If it is not, then that is not a transient thing like networking issues to a Redis server, that's a severe bug.

pkg/server/server.go
```diff
@@ -26,15 +27,17 @@ import (
...
 )
 
 type ExperimentalFeatureFlag string
 
 const (
-	AuthorizationModelIDHeader = "openfga-authorization-model-id"
-	authorizationModelIDKey    = "authorization_model_id"
+	AuthorizationModelIDHeader  = "openfga-authorization-model-id"
+	authorizationModelIDKey     = "authorization_model_id"
+	ExperimentalCheckQueryCache = "check-query-cache"
```
elbuo8, line 40: Why the `Experimental` prefix?

internal/graph/cached_resolver.go
adriantam: Tracked via https://github.com/openfga/openfga/issues/925

cmd/run/run.go
```diff
@@ -410,6 +424,11 @@ func DefaultConfig() *Config {
 			Addr:                "0.0.0.0:2112",
 			EnableRPCHistograms: false,
 		},
+		CheckQueryCache: CheckQueryCache{
+			Enabled: true,
```
jon-whit, line 428: The default should be `false` IMO. Developers should have to opt-in to eventually consistent semantics in OpenFGA. The default right now is read-committed data, which implies strongly consistent semantics. If a developer want to trade off stale reads for lower latency, then that's their choice, but we shouldn't change the default behavior.

pkg/server/server.go
jon-whit: This should be a typed enum const so that it can be treated as an enum.

```go
CheckQueryCache ExperimentalFeatureFlag = "check-query-cache"
```

pkg/server/server.go
```diff
@@ -44,6 +47,9 @@ const (
 	defaultListObjectsMaxResults            = 1000
 	defaultMaxConcurrentReadsForCheck       = math.MaxUint32
 	defaultMaxConcurrentReadsForListObjects = math.MaxUint32
+	defaultCheckQueryCacheLimit             = 10000
+	defaultCheckQueryCacheTTL               = 10 * time.Second
+	defaultCheckQueryCacheEnable            = true
```
jon-whit, line 52: Default to false for the reasons mentioned earlier.

https://github.com/openfga/openfga/pull/891/files#r1284870129

pkg/server/server.go
```diff
@@ -167,6 +178,24 @@ func WithExperimentals(experimentals ...ExperimentalFeatureFlag) OpenFGAServiceV
 	}
 }
 
+func WithCheckQueryCacheEnabled(enabled bool) OpenFGAServiceV1Option {
```
jon-whit, line 181: Can we add godoc to these, like we did elsewhere? This is part of our public API (if used as a lib), so it feels pertinent.

.config-schema.json
```diff
@@ -349,6 +349,30 @@
...
                 }
             }
+        },
+        "checkQueryCache": {
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "description": "when executing Check and ListObjects requests, enables caching. This will turn Check and ListObjects responses into eventually consistent responses",
+                    "type": "bool",
+                    "default": "true",
```
jon-whit, line 359: Default should be `false`. The query cache should be an opt-in, not defaulted to true.

.config-schema.json
```diff
@@ -349,6 +349,30 @@
...
+                    "x-env-variable": "OPENFGA_CHECK_QUERY_CACHE_ENABLED"
+                },
+                "limit": {
+                    "description": "if caching of Check and ListObjects calls is enabled, this is the size limit of the cache",
```
jon-whit, line 363: Can we add that this is the "size limit (in bytes)" just to be clear it's unit is bytes.

pkg/server/test/list_objects.go
```diff
@@ -69,6 +72,7 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 			maxResults:             2,
 			minimumResultsExpected: 2,
 			allResults:             []string{"repo:1", "repo:2", "repo:3"},
+			useCheckCache:          false,
```
jon-whit, line 75: Can we omit in favor of brevity where the value is `false`?

internal/graph/mock_check_resolver.go
jon-whit, line 1: Please move this to the internal mocks package instead of here.

adriantam: If we move this to internal mocks, we will have cyclical dependency as we have UT that relies on the mock package.  But the mock package (i.e., mock_check_resolver.go line 38) relies on graph package.

pkg/server/test/list_objects.go
adriantam: Having more explicit value makes things more clear though

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	})
+
+	checkCacheHitCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "check_cache_hit_count",
+		Help: "The total number of cache hits for ResolveCheck.",
+	})
+)
+
+// CachedResolveCheckResponse is very similar to ResolveCheckResponse except we
+// do not store the ResolutionData
```
miparnisari, line 38: why?

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+type CachedResolveCheckResponse struct {
+	Allowed bool
+}
+
+func (c *CachedResolveCheckResponse) convertToResolveCheckResponse() *ResolveCheckResponse {
+	return &ResolveCheckResponse{
+		Allowed: c.Allowed,
+		ResolutionMetadata: &ResolutionMetadata{
+			Depth:               defaultResolveNodeLimit,
+			DatastoreQueryCount: 0,
```
miparnisari, line 48: i'm scared that setting `DatastoreQueryCount` to zero will break how we are counting the reads. Shouldn't this be set to `req.ResolutionMetadata.DatastoreQueryCount`?

There is a test for the `DatastoreQueryCount` called `TestCheckDatastoreQueryCount`. Can you create one test that is similar, but which uses this cached resolver? Specifically, I want to make sure that, if one Check call involves several levels in a tree, and the last level is cached, we don't return zero. We should return all the reads we had to do prior to the cached one.

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// WithExistingCache sets the cache to the provided cache
+func WithExistingCache(cache *ccache.Cache[*CachedResolveCheckResponse]) CachedCheckResolverOpt {
+	return func(ccr *CachedCheckResolver) {
+		ccr.cache = cache
+	}
+}
+
+// WithLogger sets the logger for the cached check resolver
```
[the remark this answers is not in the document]
adriantam: Ideally, we have comments to explain what each public method does for godoc.

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	return func(ccr *CachedCheckResolver) {
+		ccr.logger = logger
+	}
+}
+
+// NewCachedCheckResolver constructs a CheckResolver that delegates Check resolution to the provided delegate,
+// but before delegating the query to the delegate a cache-key lookup is made to see if the Check sub-problem
+// has already recently been computed. If the Check sub-problem is in the cache, then the response is returned
+// immediately and no re-computation is necessary.
+// NOTE: the ResolveCheck's resolution data will be set as the default values as we actually did no database lookup
```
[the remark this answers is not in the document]
adriantam: I disagree - convertToResolveCheckResponse is an internal method. Users of CachedCheckResolver should understand what it does without going into internal method

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+		logger:       logger.NewNoopLogger(),
+	}
+
+	for _, opt := range opts {
+		opt(checker)
+	}
+
+	if checker.cache == nil {
+		// this means there were no cache supplied
+		checker.cache = ccache.New(
```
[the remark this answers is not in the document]
adriantam: added

internal/graph/cached_resolver.go
adriantam: Change to return error. If it is transient error, let's have the caller retry.

pkg/server/test/list_objects.go
adriantam: Track via https://github.com/openfga/openfga/issues/933

pkg/server/test/list_objects.go
```diff
@@ -203,6 +236,15 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 				opts = append(opts, commands.WithListObjectsDeadline(test.listObjectsDeadline))
 			}
 
+			if test.useCheckCache {
+				checkCache := ccache.New(
```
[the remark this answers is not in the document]
adriantam: done

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,191 @@
...
+	c.cache.Set(cacheKey, newCachedResolveCheckResponse(resp), c.cacheTTL)
+	return resp, nil
+}
+
+// checkRequestCacheKey converts the ResolveCheckRequest into a canonical cache key that can be
+// used for Check resolution cache key lookups.
+//
+// We use the 'gob' package to produce a deterministic byte ordering for the contextual tuples provided
+// in the request body. The same tuple provided with the same contextual tuples should produce the same
+// cache key.
```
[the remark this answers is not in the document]
adriantam: documented

.config-schema.json
```diff
@@ -349,6 +349,30 @@
...
+                    "default": "false",
+                    "x-env-variable": "OPENFGA_CHECK_QUERY_CACHE_ENABLED"
+                },
+                "limit": {
+                    "description": "if caching of Check and ListObjects calls is enabled, this is the size limit (in items) of the cache",
+                    "type": "integer",
+                    "default": "10000",
+                    "x-env-variable": "OPENFGA_CHECK_QUERY_CACHE_LIMIT"
+                },
+                "TTL": {
```
jon-whit, line 368: Shouldn't this be lowercase?

`checkQueryCache.ttl`

pkg/server/server.go
```diff
@@ -354,6 +401,17 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
 
+	if s.checkCache != nil {
+		cachedCheckResolver := graph.NewCachedCheckResolver(
+			checkResolver,
+			graph.WithExistingCache(s.checkCache),
+			graph.WithCacheTTL(s.checkQueryCacheTTL),
+		)
+
+		checkResolver.SetDelegate(cachedCheckResolver)
+		defer cachedCheckResolver.Close()
```
jon-whit, line 412: The cache is shared across requests. You shouldn't close the cache here. The only place where we should be stopping the cache is at server shutdown.

pkg/server/server.go
```diff
@@ -354,6 +401,17 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+	}
+
 	resp, err := checkResolver.ResolveCheck(ctx, &graph.ResolveCheckRequest{
```
jon-whit, line 415: We're not using the `cachedCheckResolver` here. We should be. Right now nothing is being cached.

jon-whit: I suggest the following

```go
var resolver graph.CheckResolver

checkResolver := graph.NewLocalCheck(...)
resolver = checkResolver

if s.checkCache != nil {
  cachedCheckResolver := graph.NewCachedCheckResolver(...)

  checkResolver.SetDelegate(cachedCheckResolver)
  resolver = cachedCheckResolver
}

resp, err := resolver.ResolveCheck(...)
```

internal/graph/cached_resolver.go
```diff
@@ -0,0 +1,202 @@
...
+	if err := gob.NewEncoder(&c).Encode(req.GetContextualTuples()); err != nil {
+		return "", err
+	}
+
+	contextualTuplesCacheKey := c.String()
+
+	key := fmt.Sprintf("%s/%s/%s/%s",
+		req.GetStoreID(),
+		req.GetAuthorizationModelID(),
+		tuplesCacheKey,
```
miparnisari, line 197:
```suggestion
		tupleCacheKey,
```

pkg/server/server.go
miparnisari: +1

pkg/server/server.go
miparnisari: I don't understand how that code is different to adrian's code

pkg/server/server.go
adriantam: There is no stopping of the cache here as it is provided with checkCache.

pkg/server/server.go
adriantam: Added test + added in changes as suggested by @jon-whit .  The code is different in that the initial call to ResolveCheck should be through cached check resolver.

pkg/server/server.go
adriantam: See https://github.com/openfga/openfga/blob/c1b773a15e65f2c70ceddc339963326ac53aed21/internal/graph/cached_resolver.go#L139

cmd/run/run.go
```diff
@@ -179,6 +179,12 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Uint32("listObjects-max-results", defaultConfig.ListObjectsMaxResults, "the maximum results to return in non-streaming ListObjects API responses. If 0, all results can be returned")
 
+	flags.Bool("check-query-cache-enabled", defaultConfig.CheckQueryCache.Enabled, "when executing Check and ListObjects requests, enables caching. This will turn Check and ListObjects responses into eventually consistent responses")
```
jon-whit, line 182: Can we please add tests for these new config options like we've done here?

https://github.com/openfga/openfga/blob/main/cmd/run/run_test.go#L1079-L1120

We should test to make sure that the config file, env variables, and flags all behave as expected by asserting that `ReadConfig` parses the provided configs correctly (including the merging of values behavior).

pkg/server/server.go
jon-whit: Then why are we calling `defer cachedCheckResolver.Close()` here? It's misleading as a developer reading the code.

pkg/server/commands/list_objects.go
```diff
@@ -231,6 +254,16 @@ func (q *ListObjectsQuery) evaluate(
 			graph.WithMaxConcurrentReads(q.maxConcurrentReads),
 		)
 
+		if q.checkCache != nil {
+			cachedCheckResolver := graph.NewCachedCheckResolver(
```
jon-whit, line 258: We're not using the `cachedCheckResolver` below when we call `Check`, so we're not actually using the cache. You should follow the same pattern you followed in Check here.

https://github.com/openfga/openfga/pull/891/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R413-R429

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,514 @@
...
+			},
+			setTestExpectations: func(mock *MockCheckResolver, request *ResolveCheckRequest) {
+				mock.EXPECT().ResolveCheck(ctx, request).Times(0).Return(result, nil)
+			},
+		},
+		{
+			// Ideally we will have the same order. However, having different order
+			// will not be catastrophic - just result in a cache miss and potentially
+			// duplicate entry
+			name: "different_order_contextual_tuples_does_not_return_results_from_cache",
```
jon-whit, line 206: I suggested we take a checksum of a the contextual tuples to produce a cache key that is stable independent of order.

If we're going to save that work for a subsequent PR, would you please add an issue to track that work and cross reference it here?

Same contextual tuples but different order should result in a cache hit.

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+			_, _ = dut.ResolveCheck(ctx, initialReq)
+
+			newCtrl := gomock.NewController(t)
+			defer newCtrl.Finish()
+
+			newResolver := NewMockCheckResolver(newCtrl)
+			test.setTestExpectations(newResolver, test.subsequentReq)
+
+			dut2 := NewCachedCheckResolver(newResolver, WithExistingCache(dut.cache))
```
jon-whit, line 358: Should we be testing this with the same pattern we use in `pkg/server/server.go` and in `pkg/server/commands/list_objects.go`?

Namely,

```go
localChecker := NewLocalCheckResolver(...)
cachedChecker := NewCachedCheckResolver(localChecker, ...)
localChecker.SetDelegate(cachedChecker)

_, _ = cachedChecker.ResolveCheck(ctx, test.initialReq)
...
result, err := cachedChecker.ResolveCheck(ctx, test.subsequentReq)
require.Nil(t, err)
require.Equal(...)
```

adriantam: @jon-whit : I am not sure what you are asking here - the purpose of this test is to assert the number of times ResolveCheck is called in order to validate whether data is served from cache vs. from DS.  As such, we need to mock the check resolver.

pkg/server/commands/list_objects.go
```diff
@@ -98,6 +105,20 @@ func WithLogger(l logger.Logger) ListObjectsQueryOption {
 	}
 }
 
+// WithCheckQueryCacheTTL sets the check cache query TTL
+func WithCheckQueryCacheTTL(ttl time.Duration) ListObjectsQueryOption {
```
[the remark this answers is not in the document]
jon-whit: Sounds good. Can we add an issue to follow up and consolidate the boilerplate construction of CheckResolvers?

internal/graph/cached_resolver_test.go
jon-whit: Sorry I wasn’t clear. What I meant was to try mimic the pattern we use more closely, which implies not creating a second mock CheckResolver. It’s not necessary.

Instead, you can
```go
mockCheckResolver := NewMockCheckResolver(ctrl)
initialReq := req
if test.initialReq != nil {
	initialReq = test.initialReq
}
test.setInitialResult(mockCheckResolver, initialReq)

// expect first call to result in actual resolve call
dut := NewCachedCheckResolver(mockCheckResolver,
	WithLogger(logger.NewNoopLogger()),
	WithMaxCacheSize(10))
defer dut.Close()

_, _ = dut.ResolveCheck(ctx, initialReq)

test.setTestExpectations(mockCheckResolver, test.subsequentReq)

dut2 := NewCachedCheckResolver(dut, WithExistingCache(dut.cache))
defer dut2.Close()

actualResult, err := dut2.ResolveCheck(ctx, test.subsequentReq)
require.Equal(t, result.Allowed, actualResult.Allowed)
require.Nil(t, err)
```

e.g.
Use `dut` as the delegate for the `dut2` CheckResolver. That way we’re composing/chaining the resolvers in a way that exercises the delegate behavior.

pkg/server/commands/list_objects.go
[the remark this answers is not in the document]
adriantam: https://github.com/openfga/openfga/issues/945

internal/graph/cached_resolver_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+			// expect first call to result in actual resolve call
+			dut := NewCachedCheckResolver(mockResolver,
+				WithLogger(logger.NewNoopLogger()),
+				WithMaxCacheSize(10))
+			defer dut.Close()
+
+			_, _ = dut.ResolveCheck(ctx, initialReq)
+
+			newCtrl := gomock.NewController(t)
```
jon-whit, line 352: Not needed or used. The `ctrl` above is used for the `mockResolver`.

pkg/server/server.go
jon-whit: If we're going to leave this code here like this, without an alternative compositional pattern, then we should at minimum make sure we have a unit test in `cached_resolver_test.go` that asserts that the `CachedCheckResolver.Close()` method doesn't call 'Stop' on the underlying cache. If a developer were to accidentally change that code without realize the impact, then that could be bad and an unintended side-effect.

internal/graph/cached_resolver_test.go
adriantam: removed

pkg/server/server.go
```diff
@@ -363,11 +410,26 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+	localCheckResolver := graph.NewLocalChecker(
 		storagewrappers.NewCombinedTupleReader(s.datastore, req.ContextualTuples.GetTupleKeys()),
 		graph.WithResolveNodeBreadthLimit(s.resolveNodeBreadthLimit),
 		graph.WithMaxConcurrentReads(s.maxConcurrentReadsForCheck),
 	)
+	checkResolver = localCheckResolver
+
+	if s.checkCache != nil {
+		cachedCheckResolver := graph.NewCachedCheckResolver(
+			checkResolver,
```
jon-whit, line 423: Tricky tricky tricky! This should be `localCheckResolver`. `checkResolver` stores a pointer reference to an implementation of the interface. When you re-assign the `checkResolver` below on line https://github.com/openfga/openfga/pull/891/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R431 you effecitvely set the delegate of the CachedCheckResolver to itself. You can verify this by checking that the pointer reference is the same between `checkResolver` and `cachedCheckResolver`.

* CachedCheckerResolver should delegate to LocalCheckResolver
* LocalCheckResolver should delegate to CachedCheckResolver


Right now we have:
* CachedCheckResolver delegates to CachedCheckResolver
* LocalCheckResolver delegates to CachedCheckResolver

cmd/run/run.go
```diff
@@ -720,6 +739,9 @@ func (s *ServerContext) Run(ctx context.Context, config *Config) error {
 		server.WithListObjectsMaxResults(config.ListObjectsMaxResults),
```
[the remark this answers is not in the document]
adriantam: Instead of log.Warn, I think we just log that check query cache is enabled (s.Logger.Info)  Warn implies something is wrong.  Turning on query cache is an explicit configuration.

pkg/server/server.go
adriantam: Hmm... we have UT that shows check works.  However, I can change it to localCheckResolver to keep things clean.

jon-whit:
LGTM. Left one comment above about adding a test to `cached_resolver_test.go` for the `Close` behavior, but not necessarily a blocker.

pkg/server/server.go
adriantam: We already have it in https://github.com/openfga/openfga/blob/d9c04f2fd935bb693dbb5fa971dbc3b4050019e9/internal/graph/cached_resolver_test.go#L466 - here we close firstCachedResolver but have secondCachedResolver running with checkCache

pkg/server/server.go
```diff
@@ -363,11 +414,26 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+	checkResolver = localCheckResolver
+
+	if s.checkCache != nil {
+		cachedCheckResolver := graph.NewCachedCheckResolver(
+			localCheckResolver,
+			graph.WithExistingCache(s.checkCache),
+			graph.WithCacheTTL(s.checkQueryCacheTTL),
+		)
+
+		localCheckResolver.SetDelegate(cachedCheckResolver)
```
miparnisari, line 432: Echoing https://auth0.slack.com/archives/C2ZCK92Q6/p1692227188656589

This is like a brain teaser and we shouldn't have those. Is there any way we can make this simpler please?

miparnisari (changes requested):

miparnisari (approved):

pkg/server/server.go
miparnisari: @jpadilla what do you think? since you are a newcomer to the code. Is this easy to follow?

adriantam: Hopefully after this clean up is easier to read and understand

pkg/server/server.go
```diff
@@ -363,11 +414,25 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+			storagewrappers.NewCombinedTupleReader(s.datastore, req.ContextualTuples.GetTupleKeys()),
+			graph.WithResolveNodeBreadthLimit(s.resolveNodeBreadthLimit),
+			graph.WithMaxConcurrentReads(s.maxConcurrentReadsForCheck),
+			graph.WithCachedResolver(
+				graph.WithExistingCache(s.checkCache),
+				graph.WithCacheTTL(s.checkQueryCacheTTL),
+			),
+		)
+	}
+	defer checkResolver.Close()
```
miparnisari, line 435: This is too verbose and it's duplicate in two places.

I have a suggestion that builds upon your PR: https://gist.github.com/miparnisari/d833b97ba94e8d57ae50e6e8404ca4d6

The gist: we create all the necessary Check options on server startup, and then we use those options in Check and ListObjects

internal/graph/check.go
```diff
@@ -136,22 +131,41 @@ func WithMaxConcurrentReads(limit uint32) LocalCheckerOption {
...
 // NewLocalChecker constructs a LocalChecker that can be used to evaluate a Check
 // request locally.
-func NewLocalChecker(ds storage.RelationshipTupleReader, opts ...LocalCheckerOption) *LocalChecker {
+func NewLocalChecker(ds storage.RelationshipTupleReader, opts ...LocalCheckerOption) CheckResolver {
 	checker := &LocalChecker{
 		ds:                 ds,
 		concurrencyLimit:   defaultResolveNodeBreadthLimit,
 		maxConcurrentReads: defaultMaxConcurrentReadsForCheck,
 	}
+	checker.delegate = checker // by default, a LocalChecker delegates/dispatchs subproblems to itself (e.g. local dispatch) unless otherwise configured.
```
miparnisari, line 158: https://github.com/openfga/openfga/pull/476#discussion_r1089530447

I'm sorry I keep bringing this up but I still hold the opinion that this line is odd. LocalChecker has a `delegate` which can hold a CachedLocalChecker. Meanwhile, CachedLocalChecker has also a `delegate` which holds a LocalChecker. It's like the [Spiderman meme](https://knowyourmeme.com/memes/spider-man-pointing-at-spider-man)..

I won't block the PR though, I'll think if there is a better way of doing this and if there is i'll send a PR.

pkg/server/server.go
adriantam: I have taken the idea and update the. code. I could not take the code as is as it has a bug in setting incorrect resolve node breadth limit for list object (using maxConcurrentReadsForCheck instead of maxConcurrentReadsForListObjects).

internal/graph/check.go
adriantam: Let's have a follow up PR if there is a better way later.

miparnisari (approved):

Issue #915, opened 2023-08-02
Inconsistent max concurrent reads for check value between running as standalone vs. library


adriantam:
In the case OpenFGA ran as a standalone, the default max concurrent read for check is [4294967295](https://github.com/openfga/openfga/blob/d88474a3871cb50f40476dd8b70ec9f9ea6a870a/.config-schema.json#L18) .

However, if OpenFGA is imported as a library, the default max concurrent read for check is [100](https://github.com/openfga/openfga/blob/d88474a3871cb50f40476dd8b70ec9f9ea6a870a/pkg/server/server.go#L44).

The default configured value should be identical.

PR #916, opened 2023-08-02
fix: concurrency read settings should be unbounded by default

jon-whit:
## Description
The defaults for concurrency read limits should be unbounded to match preexisting behavior. Constraining the concurrency limits should be an "opt in" setting.

## References
Close https://github.com/openfga/openfga/issues/915

## Review Checklist

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -13,15 +13,15 @@
             "x-env-variable": "OPENFGA_MAX_TYPES_PER_AUTHORIZATION_MODEL"
         },
         "maxConcurrentReadsForCheck": {
-            "description": "The maximum allowed number of concurrent reads in a single Check query. The default is MaxUint32",
+            "description": "The maximum allowed number of concurrent reads in a single Check query (default is MaxUint32).",
             "type": "integer",
             "default": 4294967295,
             "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_CHECK"
         },
         "maxConcurrentReadsForListObjects": {
-            "description": "The maximum allowed number of concurrent reads in a single ListObjects query.",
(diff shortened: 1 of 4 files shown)


.config-schema.json
```diff
@@ -13,15 +13,15 @@
...
             "type": "integer",
             "default": 4294967295,
             "x-env-variable": "OPENFGA_MAX_CONCURRENT_READS_FOR_CHECK"
         },
         "maxConcurrentReadsForListObjects": {
-            "description": "The maximum allowed number of concurrent reads in a single ListObjects query.",
+            "description": "The maximum allowed number of concurrent reads in a single ListObjects query (default is MaxUint32).",
             "type": "integer",
-            "default": 30,
+            "default": 4294967295,
```
adriantam, line 24: This is a breaking behaviour - do we have consensus of tweaking this value

jon-whit: So was the unintended breaking change to the concurrency limits in Read.

It's new behavior. We got the values wrong.

adriantam: In this case, please
- ensure it is well documented in the changelog 
- update title of PR to reflect that we are changing the default behaviour of listobject

adriantam: Do we also want to update the sample docker-compose file and https://openfga.dev/docs/getting-started/running-in-production to recommend operator setting a reasonable value

miparnisari (approved):

.config-schema.json
jon-whit: I will update the production best practices documentation, but I think leaving the docker-compose file as is would be fine. If you and @miparnisari feel strongly otherwise, I'm happy to update the docker-compose file as well.

Consensus?

adriantam (approved):

Issue #920, opened 2023-08-03
List-Object endpoint returns inconsistent results for relations using usersets


rouven-s:
### Setup

**Build version:** 1.3.0
**Auth Model:**
```
model
  schema 1.1

type user

type test_type
  relations
    define relation1: [user]
    define relation2: [user]
    define relation3: [user]
    define relation4: [user]
    define relation5: [user]
    define relation6: [user]
    define relation7: [user]
    define relation8: [user]

type list_type
  relations
    define list_relation: [test_type#relation1,test_type#relation2,test_type#relation3,test_type#relation4,test_type#relation5,test_type#relation6,test_type#relation7,test_type#relation8]
```

**Tuples in Store:**
```json
[
	{
			"user": "user:test_user",
			"relation": "relation1",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation2",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation3",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation4",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation5",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation6",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation7",
			"object": "test_type:test_type1"
        },
	{
			"user": "user:test_user",
			"relation": "relation8",
			"object": "test_type:test_type1"
        },
	{
			"user": "test_type:test_type1#relation1",
			"relation": "list_relation",
			"object": "list_type:list_type1"
        }
]
```

### Issue
When I use the list-object endpoint with the following body, the endoint sometimes returns one object (as expected) and sometimes an empty list:
```json
{
    "user": "user:test_user",
    "relation": "list_relation",
    "type": "list_type"
}
```

I know that there is the limitiation of deadline (3 seconds by default) and maximum tuple result (1000 results by default). In this case neither of them are breached, the request only takes about 5 milliseconds locally and should only return 1 object. Is there any other limitation that I am not aware of or is this a bug?

adriantam:
We are able to reproduce issue with a smaller set

Model:

```
model
  schema 1.1
type user
type test_type
  relations
    define relation1: [user]
    define relation2: [user]
   
type list_type
  relations
[...]

Issue #863, opened 2023-07-12
Minor optimization in ListObjects


miparnisari:
i’m reading ListObjects code and i see this:

https://github.com/openfga/openfga/blob/d2360092635abe5ae7876b31faf5f2be2e141391/pkg/server/commands/list_objects.go#L190-L207

This is good, but shouldn’t we _also_ have the following before we do the Check call? 

```go
if atomic.LoadUint32(objectsFound) >= maxResults { 
     return
}
```

Right now, we’ll do the Check and, if we’ve already hit the maxResults, we’ll throw the result away :grimacing:

jingchu000:
Haha, you have already told me how to modify it, haven't you? i work it

Issue #800, opened 2023-06-02
update/explain documentation and comments in `graph.go`


rhamzeh:
In [graph.go](https://github.com/openfga/openfga/blob/2c0a0f5525902edb20caef67a37767a1b146f90b/internal/graph/graph.go#L230), the examples given on lines 230 and 253 in the if and else seem to be the same.

Issue #892, opened 2023-07-27
Flaky ListObjects & Check integration tests


miparnisari:
Use this script to try and reproduce:

```
#!/bin/bash

for ((i=1; i<=1000; i++)); do
    echo "Running test iteration $i"
    go test -race -count=1 ./tests/listobjects/...
done
```

I got this

```
[27/07/23 4:13:30] ~/GitHub/openfga (main) $ sh script.sh     
Running test iteration 1
ok      github.com/openfga/openfga/tests/listobjects    16.199s
Running test iteration 2
ok      github.com/openfga/openfga/tests/listobjects    15.652s
Running test iteration 3
ok      github.com/openfga/openfga/tests/listobjects    18.389s
Running test iteration 4
ok      github.com/openfga/openfga/tests/listobjects    16.651s
Running test iteration 5
ok      github.com/openfga/openfga/tests/listobjects    15.535s
Running test iteration 6
ok      github.com/openfga/openfga/tests/listobjects    15.088s
Running test iteration 7
ok      github.com/openfga/openfga/tests/listobjects    14.015s
Running test iteration 8
ok      github.com/openfga/openfga/tests/listobjects    20.835s
Running test iteration 9
ok      github.com/openfga/openfga/tests/listobjects    15.230s
Running test iteration 10
ok      github.com/openfga/openfga/tests/listobjects    13.769s
Running test iteration 11
ok      github.com/openfga/openfga/tests/listobjects    17.323s
Running test iteration 12
ok      github.com/openfga/openfga/tests/listobjects    20.563s
Running test iteration 13
ok      github.com/openfga/openfga/tests/listobjects    15.386s
Running test iteration 14
ok      github.com/openfga/openfga/tests/listobjects    14.283s
Running test iteration 15
ok      github.com/openfga/openfga/tests/listobjects    14.160s
Running test iteration 16
ok      github.com/openfga/openfga/tests/listobjects    16.574s
Running test iteration 17
ok      github.com/openfga/openfga/tests/listobjects    15.841s
Running test iteration 18
ok      github.com/openfga/openfga/tests/listobjects    14.224s
Running test iteration 19
ok      github.com/openfga/openfga/tests/listobjects    14.627s
Running test iteration 20
ok      github.com/openfga/openfga/tests/listobjects    14.562s
Running test iteration 21
ok      github.com/openfga/openfga/tests/listobjects    14.522s
Running test iteration 22
ok      github.com/openfga/openfga/tests/listobjects    17.121s
[...]

PR #921, opened 2023-08-03
fix: ListObjects gives non-deterministic results

adriantam:
## Description
- Fix ListObjects: sometimes it returns 1 result, sometimes it returns 0 results.

Things broken and fixed:
- Some unit tests.
- RelationshipIngresses handling for TTU.
- Keep 3 different maps:
   - ConnectedObjects keeps `visitedMap` to keep track of the ingresses visited for each userset, to avoid revisiting the same path more than once
   - ConnectedObjects keeps `candidateMap` to keep track of good objects AND potential candidate objects 
   - ListObjects keeps `foundObjectsMap`. This ONLY stores good objects.
- ConnectedObjects was not adding objects found while expanding a computed userset.
- ConnectedObjects was accepting a `maxResults`. It should not. We want it to give us as many candidate objects as possibles, and let ListObjects decide which of those are good objects. Otherwise we may miss some objects in the response.

NOTE:
- ~~I had to comment out the infamous `three_prong_relation_loop` test (sorry @adriantam ). I **strongly** recommend that we send a follow-up PR for that. I think that we need to keep track of the ingresses we visited and exit when we’ve entered a cycle.~~
- We may still see test failures. This is because of a [bug in our test parallelization](https://github.com/openfga/openfga/issues/892). Rest assured that when i removed [this](https://github.com/openfga/openfga/blob/ff41e318068dc0fadcb26186a91202bb729eab59/tests/listobjects/listobjects.go#L121) `t.Parallel()` locally and let the tests run forever (or: a reasonably long amount of time), they always pass.

This is how tracing looks like now:

image

## References
Closes https://github.com/openfga/openfga/issues/920
Closes https://github.com/openfga/openfga/issues/800
Closes https://github.com/openfga/openfga/issues/863


adriantam:
> I had to comment out the infamous three_prong_relation_loop test (sorry @adriantam ). I strongly recommend that we send a follow-up PR for that. I think that we need to keep track of the ingresses we visited and exit when we’ve entered a cycle.

@miparnisari - please ensure we have an issue tracking that as it indicates there are cyclical dependency where list object algorithm does not handle correctly.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -553,15 +488,39 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
...
+			candidateChan <- &ConnectedObjectsResult{
+				Object:       candidateObject,
+				ResultStatus: RequiresFurtherEvalStatus,
+			}
+		} else {
+			candidateChan <- &ConnectedObjectsResult{
+				Object:       candidateObject,
+				ResultStatus: NoFurtherEvalStatus,
+			}
+		}
```
adriantam, line 524:
```suggestion
resultStatus := NoFurtherEvalStatus
if ingress != nil && ingress.Condition == graph.RequiresFurtherEvalCondition {
	resultStatus = RequiresFurtherEvalStatus
}
candidateChan <- &ConnectedObjectsResult{
	Object:       candidateObject,
	ResultStatus: resultStatus,
}
```

pkg/server/commands/list_objects.go
```diff
@@ -288,6 +292,14 @@ func (q *ListObjectsQuery) evaluate(
 	return nil
 }
 
+func sendObject(object string, foundObjectsMap *sync.Map, foundCount *uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
+	if _, loaded := foundObjectsMap.LoadOrStore(object, struct{}{}); !loaded {
+		if foundCount != nil && atomic.AddUint32(foundCount, 1) <= maxResults {
```
adriantam, line 297: Want to confirm that it is our intention where if maxResults == 0, we will not be sending the result to the resultsChan.  This is because if maxResults==0, foundCount will be nil.

pkg/server/test/connected_objects.go
```diff
@@ -1155,6 +1162,41 @@ func ConnectedObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 				},
 			},
 		},
+		{
+			name: "new_test",
```
adriantam, line 1166: new_test is not a good description

pkg/server/test/connected_objects.go
```diff
@@ -1224,7 +1266,7 @@ func ConnectedObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 					return test.expectedResult[i].Object < test.expectedResult[j].Object
 				})
 
-				require.Equal(test.expectedResult, results)
+				require.Subset(test.expectedResult, results)
```
adriantam, line 1269: Curious why are we changing this to a subset?

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -158,20 +168,24 @@ type ConnectedObjectsResult struct {
 	ResultStatus ConditionalResultStatus
 }
 
-func (c *ConnectedObjectsQuery) execute(
+// Execute yields all the objects of the provided objectType that the given user has, possibly, a specific relation with
+// and sends those objects to resultChan. It guarantees no duplicate objects sent.
```
jon-whit, line 172: If we're guaranteeing no duplicate objects sent here, why are we doing the same in `commands/list_objects.go`?

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -3268,100 +3268,100 @@ tests:
...
-              - document:a
-              - document:b
-          - request:
-              user: user:anne
-              type: module
-              relation: viewer
-            expectation:
-              - module:a
-              - module:b
+#  - name: three_prong_relation_loop
```
jon-whit, line 3271: This test doesn't pass if I uncomment it.

internal/graph/graph.go
```diff
@@ -332,37 +319,6 @@ func (g *ConnectedObjectGraph) findIngressesWithTargetRewrite(
 					TuplesetRelation: typesystem.DirectRelationReference(target.GetType(), tupleset),
 					Condition:        condition,
 				})
-			} else {
```
jon-whit, line 335: Could we get an explanation for this change with a reference to a test that demonstrates the reasoning for it?

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -107,7 +108,11 @@ type ConnectedObjectsQuery struct {
 	typesystem              *typesystem.TypeSystem
 	resolveNodeLimit        uint32
 	resolveNodeBreadthLimit uint32
-	maxResults              uint32
+
+	// candidateObjectsMap map allows us to not return the same object twice
```
jon-whit, line 112: Slight change of verbiage here so it's more clear to the next developer

> candidateObjectsMap allows us to avoid expanding the same relationship more than once.

The bit "to not return the same object twice" implies avoiding sending duplicates, but it's not that we're avoiding sending duplicates, it's that we're avoiding expanding the same relationship more than once.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -225,9 +243,12 @@ func (c *ConnectedObjectsQuery) execute(
 	subg, subgctx := errgroup.WithContext(ctx)
 	subg.SetLimit(int(c.resolveNodeBreadthLimit))
 
-	for i, ingress := range ingresses {
-		span.SetAttributes(attribute.String(fmt.Sprintf("_ingress %d", i), ingress.String()))
+	for _, ingress := range ingresses {
 		innerLoopIngress := ingress
+		if req.Ingress != nil && req.Ingress.Condition == graph.RequiresFurtherEvalCondition {
+			// propagate the condition to upcoming reverse expansions
+			innerLoopIngress.Condition = graph.RequiresFurtherEvalCondition
```
jon-whit, line 250: The ingresses status should come directly from the ingress itself. I'm confused why we are mutating ingress conditions in the expansion paths?

Could you elaborate?

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -4938,4 +4938,38 @@ tests:
               object: resource:abc
               relation: a1
               user: user:maria
-            errorCode: 2002
\ No newline at end of file
+            errorCode: 2002
+  - name: issue_920
```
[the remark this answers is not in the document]
miparnisari: There is nothing special about this model, is as simple as you can get. I'm not sure what name to give it. `race_condition`?

pkg/server/commands/list_objects.go
miparnisari: We will still send through the channel in that case. What will be different is the `ResultStatus` will be `NoFurtherEval`

pkg/server/test/connected_objects.go
miparnisari: To make the test with `limit` pass. I'll send a follow up PR to change this back

pkg/server/commands/list_objects.go
miparnisari: If `maxresults == 0` we send all the objects. I'll send a followup PR to add tests for this

pkg/server/commands/connectedobjects/connected_objects.go
jon-whit: Talked about this offline. We discussed moving this to propagating a boolean or something indicating a conditional ingress was evaluated along the evaluation path.

internal/graph/graph.go
miparnisari: It is adding an edge that should not be added

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: No, `candidateObjectsMap` **is** avoiding sending duplicates

`visitedObjectsMap` is the one we use to avoiding expanding the same relationship more than once

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: good point, i think ListObjects doesn't need to have a map anymore 🤔

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -158,20 +168,24 @@ type ConnectedObjectsResult struct {
 	ResultStatus ConditionalResultStatus
 }
 
-func (c *ConnectedObjectsQuery) execute(
+// Execute yields all the objects of the provided objectType that the given user has, possibly, a specific relation with
+// and sends those objects to resultChan. It guarantees no duplicate objects sent.
+func (c *ConnectedObjectsQuery) Execute(
```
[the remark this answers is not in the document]
miparnisari: > This is a public facing API

It's not :p It's in `pkg` but it's not exposed

pkg/server/test/connected_objects.go
```diff
@@ -1200,7 +1266,7 @@ func ConnectedObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 				done <- struct{}{}
 			}()
 
-			timeoutCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			timeoutCtx, cancel := context.WithTimeout(context.Background(), 10*time.Hour)
```
miparnisari, line 1269: This was changed in a previous commit and i'm not sure why. @jon-whit @adriantam

jon-whit: Oh, that was probably me and I committed it accidentally. I change that when debugging locally so my request doesn't time out while I debug.

pkg/server/commands/connectedobjects/connected_objects.go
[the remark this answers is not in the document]
jon-whit: `pkg/server/connectedobjects` is a public package. You can import it and thus it is exposed. 

A public consumer has no need to provide an Ingress to `Execute`, that is the responsibility internally of the call to `Execute`. It's a minor change.

I know you're tired and this has been hard work, so save it for next week if it's more bearable.

internal/graph/graph.go
```diff
@@ -115,6 +115,7 @@ type RelationshipIngress struct {
 }
 
 func (r RelationshipIngress) String() string {
+	// TODO log the condition
```
jon-whit, line 118: I don't know that the ingress itself should ever have a logger associated with it. If you want to log the condition, that should be the callers responsibility, and the caller would presumably have a logger at that point..

Can we remove this comment?

internal/graph/graph_test.go
```diff
@@ -225,6 +225,32 @@ func TestPrunedRelationshipIngresses(t *testing.T) {
...
+			target: typesystem.DirectRelationReference("repo", "admin"),
+			source: typesystem.DirectRelationReference("organization", "member"),
+			expected: []*RelationshipIngress{
+				{
+					Type:      DirectIngress,
+					Ingress:   typesystem.DirectRelationReference("organization", "repo_admin"),
+					Condition: NoFurtherEvalCondition,
+				},
+			},
+		},
```
jon-whit, line 253: Can we add two more tests here to cover a similar but alternative case?

```go
{
	name: "ttu_through_direct_rewrite_1",
	model: `
	type folder
		relations
		define viewer: [folder] as self

	type document
		relations
		define parent: [folder] as self
	`,
	target: typesystem.DirectRelationReference("document", "viewer"),
	source: typesystem.DirectRelationReference("folder", "viewer"),
	expected: []*RelationshipIngress{
		{
			Type:             TupleToUsersetIngress,
			Ingress:          typesystem.DirectRelationReference("document", "viewer"),
			TuplesetRelation: typesystem.DirectRelationReference("document", "parent"),
			Condition:        NoFurtherEvalCondition,
		},
	},
},
{
	name: "ttu_through_direct_rewrite_2",
	model: `
	type folder
		relations
		define viewer: [folder] as self

	type document
		relations
		define parent: [folder] as self
	`,
	target: typesystem.DirectRelationReference("document", "viewer"),
	source: typesystem.DirectRelationReference("folder", ""),
	expected: []*RelationshipIngress{
		{
			Type:      DirectIngress,
			Ingress:   typesystem.DirectRelationReference("folder", "viewer"),
			Condition: NoFurtherEvalCondition,
		},
	},
},

File: internal/graph/graph.go (hunk 7 above)
Comment (jon-whit) [reply internal/graph/graph.go:335]: I verified this behavior and it's solid. I [left a comment](https://github.com/openfga/openfga/pull/921/files#r1298896832) to add a few tests just to cover our bases on the original reason why we added this code (albeit we added it in an erroneous way back then).

If we can add those tests then we're good here 👍

File: pkg/server/commands/connectedobjects/connected_objects.go [hunk 15]
```diff
@@ -553,15 +497,37 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
...
 
 	return subg.Wait()
 }
+
+func (c *ConnectedObjectsQuery) sendCandidate(ctx context.Context, ingress *graph.RelationshipIngress, candidateObject string, candidateChan chan<- *ConnectedObjectsResult) {
+	if _, ok := c.candidateObjectsMap.LoadOrStore(candidateObject, struct{}{}); !ok {
+		if c.candidatesFound != nil && atomic.AddUint32(c.candidatesFound, 1) > c.maxCandidates {
+			return
+		}
+		_, span := tracer.Start(ctx, "sendCandidate", trace.WithAttributes(
```
jon-whit, line 519: Should we start this at the beginning of the function call. It's possible with enough contention on the `LoadOrStore` that we could see higher trace latencies for this function, but with where this is placed we won't ever see contention on the synchronized map.

internal/graph/graph.go
miparnisari: oh i meant `print` the condition, not log

pkg/server/commands/list_objects.go
adriantam: But if `maxresults==0`, `found==nil`.  The condition will be false and we will not execute line 298.

assets/tests/consolidated_1_1_tests.yaml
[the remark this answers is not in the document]
adriantam: Maybe `race_condition_same_user_same_object_different_relation`?

pkg/server/commands/list_objects.go
adriantam: If possible, let's add the test before completing this PR?

miparnisari: > The condition will be false and we will not execute line 298.

Correct, that is the line where we return early because we've hit the limit :)

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: I put this trace here so that when I was debugging via looking at Jaeger, I could quickly see which objects were sent through the channel. I didn't care about performance. If I move the trace outside the first `if`, then `sendCandidate` is not an accurate name for the trace...

pkg/server/test/connected_objects.go
miparnisari: Reverted in this PR

pkg/server/commands/connectedobjects/connected_objects.go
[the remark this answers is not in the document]
miparnisari: fixed

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: added a TODO

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -324,17 +332,9 @@ func (c *ConnectedObjectsQuery) reverseExpandTupleToUserset(
...
-
 	var userFilter []*openfgav1.ObjectRelation
 
-	// e.g. 'user:bob'
-	if val, ok := req.sourceUserRef.(*UserRefObject); ok {
-		userFilter = append(userFilter, &openfgav1.ObjectRelation{
-			Object: tuple.BuildObject(val.Object.Type, val.Object.Id),
-		})
-	}
-
```
miparnisari, line 337: This is dead code that can never be hit.

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: See the original PR description with a screenshot of how tracing looks like

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -4938,4 +4938,43 @@ tests:
...
+        checkAssertions:
+          - tuple:
+              user: user:test_user
+              relation: list_relation
+              object: list_type:list_type1
+            expectation: true
+          - tuple:
+             user: user:test_user
+             relation: list_relation
+             object: list_type:list_typ2
```
adriantam, line 4972:
```suggestion
             object: list_type:list_type2
```

pkg/server/commands/list_objects.go
adriantam: but there should be no limit when maxresults==0 though

internal/graph/graph.go
jon-whit: Why would we print the condition in the `String()` method of the ingress? Shouldn't that be the callers responsibility to print the `ingress.String()`?

Can't we just remove this comment? What follow up todo is there to be done here?

pkg/server/commands/list_objects.go
```diff
@@ -201,14 +201,20 @@ func (q *ListObjectsQuery) evaluate(
...
 		connectedObjectsQuery := connectedobjects.NewConnectedObjectsQuery(q.datastore, typesys,
 			connectedobjects.WithResolveNodeLimit(q.resolveNodeLimit),
 			connectedobjects.WithResolveNodeBreadthLimit(q.resolveNodeBreadthLimit),
 			connectedobjects.WithMaxResults(maxResults),
 		)
 
+		// foundObjects map allows us to not return the same object twice
+		var foundObjectsMap sync.Map
+		var foundCount *uint32
+		if maxResults > 0 {
```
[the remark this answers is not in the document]
jon-whit: If `maxResults` == 0 then that's supposed to indicate "no limit". Maybe it would be better to change this to a `*uint32` so that if it is nil then we treat it as no limit 🤷 ?

internal/graph/graph.go
miparnisari: > Why would we print the condition in the String() method of the ingress?

Because the `condition` is a field of the `RelationshipIngress`, so why does it NOT print it is my question 😅

pkg/server/commands/connectedobjects/connected_objects.go
jon-whit: Mind creating the issue and cross referencing with this PR to track?

pkg/server/commands/connectedobjects/connected_objects.go
jon-whit: Please create the issue and cross referencing with this PR.

pkg/server/commands/list_objects.go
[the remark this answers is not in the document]
miparnisari: > If maxResults == 0 then that's supposed to indicate "no limit". 

Correct

internal/graph/graph.go
jon-whit: Oh, gosh. We're not communicating well here. We're not talking about "printing" anything. We're talking about adding the `condition` field to the returned string. Returned vs printed is the miscomm here 😄

pkg/server/commands/list_objects.go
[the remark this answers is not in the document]
adriantam: In this case if maxResults==0 indicates no limit, then https://github.com/openfga/openfga/pull/921/files#r1300329789 doesn't make sense for me.

internal/graph/graph.go
miparnisari: Yes :D hahaha

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: I'll change in this PR

pkg/server/commands/connectedobjects/connected_objects.go
jon-whit: @miparnisari I'm wondering if we should just `panic` in these cases. We don't ever expect to hit the code paths, but if we do it's worth panic'ing to bring attention to the invariant that was not met.

Instead of swallowing the issue and continuing on, can we panic and then just codecov ignore the blocks of code? That way we're not hurting our codecov but we're also catching the invariant problem if the code path gets hit someone because of a problem introduced somewhere.

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: > This pattern/usage of a span is odd and feels very much non-standard. Instead, can we consider starting the span at the top level so that it is still representative of any latency incurred in this function call but also add an attribute indicating if a result was yielded

ok

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: > can we panic and then just codecov ignore the blocks of code?

No, I don't like the idea of introducing a panic. It will never hit and it'll make the code just a tiny bit longer to read. 

If we could somehow introduce a test for this, it will suffice. But we can't, because we can never write a model where the invariant is broken 😞

pkg/server/commands/list_objects.go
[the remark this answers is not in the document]
miparnisari: I can't follow that link, but does it make more sense now?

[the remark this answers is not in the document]
adriantam: Yes, it makes more sense now.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -559,9 +492,32 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
...
+
+		resultStatus := NoFurtherEvalStatus
+		if ingress != nil && ingress.Condition == graph.RequiresFurtherEvalCondition {
+			resultStatus = RequiresFurtherEvalStatus
+		}
+		candidateChan <- &ConnectedObjectsResult{
+			Object:       candidateObject,
+			ResultStatus: resultStatus,
+		}
+		span.SetAttributes(attribute.Bool("sent", true))
```
adriantam, line 521: should we have attribute on resultStatus?

pkg/server/commands/list_objects.go
```diff
@@ -287,6 +288,13 @@ func (q *ListObjectsQuery) evaluate(
 	return nil
 }
 
+func sendObject(object string, foundCount *uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
+	if foundCount != nil && atomic.AddUint32(foundCount, 1) > maxResults {
+		return
```
adriantam, line 293: Curious that none of our unit test/integration test hits this condition.

pkg/server/test/connected_objects.go
```diff
@@ -1216,15 +1235,165 @@ func ConnectedObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
...
+
+			select {
+			case <-timeoutCtx.Done():
+				require.FailNow(t, "timed out waiting for response")
+			case <-done:
+			}
 
-				require.Equal(test.expectedResult, results)
+			require.GreaterOrEqual(t, len(results), int(test.minResultsExpected))
+			if test.maxResults > 0 {
```
adriantam, line 1392: should the condition be 
```suggestion
			if test.maxResults == test.minResultsExpected {
```

In this way, we only check for subset if we know that the results may be cut short due to max results limitation.

pkg/server/commands/list_objects.go
miparnisari: Yeah... i'm missing a test 🤔

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: Yes, we can

pkg/server/commands/connectedobjects/connected_objects.go
jon-whit: > No, I don't like the idea of introducing a panic.

Why? 


> If we could somehow introduce a test for this, it will suffice. But we can't (I think), because we can never write a model where the invariant is broken 😞

This is exactly why `panic` is warranted IMO. If you cannot add a test for it because it is an invariant that is always expected to be met, then you should be screaming an error if/when the invariant is met.

miparnisari: Ok, added the panic

pkg/server/test/connected_objects.go
miparnisari: This comment doesn't apply anymore. ConnectedObjects now doesn't accept a `maxResults`

pkg/server/commands/list_objects.go
miparnisari: Ok should be covered now.

pkg/server/test/list_objects.go
```diff
@@ -126,6 +99,31 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 			allResults:             []string{"org:1", "org:2", "org:3"},
 			useCheckCache:          false,
 		},
+		{
+			name:   "max_results_with_model_that_uses_exclusion_and_one_object_is_a_false_candidate",
```
miparnisari, line 103: This test fails on `main` 😅 
image

In `main` we are passing `maxResults: 2` to ConnectedObjects, which gives us 2 objects (org:2 and org:1). But `org:1` has to be analyzed via Check, which returns false, so in reality we should have kept on looking for candidates.

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: https://github.com/openfga/openfga/issues/957

jon-whit (approved):

jon-whit:
LGTM. Please rebase though.

Issue #918, opened 2023-08-02
Add integration test that asserts request logging


miparnisari:
We need to have an integration test that checks that after one Check request (for example, but applies to other endpoints as well), a log gets emitted with certain fields: `raw_request`, `raw_response`, `store_id`, `model_id`, etc.

~~One way of doing this: create a server with an `WithLogger` option that passes in a https://pkg.go.dev/go.uber.org/zap/zaptest/observer, and then make assertions on those logs.~~ This won't work because we need the interceptors to be running, and those run only after you actually run the server :(

miparnisari:
Once this is fixed https://github.com/openfga/openfga/issues/935

Add another test please :)

PR #926, opened 2023-08-05
test: assert emitted log during Check request

jpadilla:
## Description

Add an integration test that asserts that an expected request log is emitted with the expected shape for one Check request.

## References

https://github.com/openfga/openfga/issues/918

## Review Checklist

### DIFF
--- a/pkg/middleware/http/handler.go
+++ b/pkg/middleware/http/handler.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"log"
 	"net/http"
 	"net/textproto"
 	"strconv"
@@ -69,9 +68,6 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
 func CustomHTTPErrorHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, err *errors.EncodedError) {
 	// convert as error object
 	pb := err.ActualError
-
-	log.Println(err)
-	log.Println(pb)
 
 	w.Header().Del("Trailer")
 	w.Header().Del("Transfer-Encoding")
--- a/pkg/middleware/http/handler.go
+++ b/pkg/middleware/http/handler.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"net/http"
 	"net/textproto"
 	"strconv"
@@ -68,6 +69,9 @@ func handleForwardResponseTrailer(w http.ResponseWriter, md runtime.ServerMetada
 func CustomHTTPErrorHandler(ctx context.Context, w http.ResponseWriter, r *http.Request, err *errors.EncodedError) {
 	// convert as error object
 	pb := err.ActualError
+
+	log.Println(err)
+	log.Println(pb)
 
 	w.Header().Del("Trailer")
 	w.Header().Del("Transfer-Encoding")
--- a/tests/check/check_test.go
+++ b/tests/check/check_test.go
@@ -95,64 +95,70 @@ func TestCheckLogs(t *testing.T) {
 	})
 	require.NoError(t, err)
 
-	resp, err := client.Check(context.Background(), &openfgav1.CheckRequest{
-		StoreId:  storeID,
-		TupleKey: tuple.NewTupleKey("document:1", "viewer", "user:anne"),
-	})
-	require.NoError(t, err)
-	require.True(t, resp.Allowed)
-
-	filteredLogs := logs.Filter(func(e observer.LoggedEntry) bool {
-		if e.Message == "grpc_req_complete" {
-			for _, ctxField := range e.Context {
-				if ctxField.Equals(zap.String("grpc_method", "Check")) {
-					return true
-				}
-			}
-		}
-
-		return false
-	})
-
-	expectedLogs := filteredLogs.All()
-	require.Len(t, expectedLogs, 1)
-
-	fields := expectedLogs[0].Context
-
-	for _, field := range fields {
-		switch field.Key {
-		case "grpc_service":
-			require.Equal(t, field.String, "openfga.v1.OpenFGAService")
-		case "grpc_method":
-			require.Equal(t, field.String, "Check")
-		case "grpc_type":
-			require.Equal(t, field.String, "unary")
-		case "grpc_code":
-			require.Equal(t, field.Integer, int64(0))
-		case "raw_request":
-			raw, err := json.Marshal(&field.Interface)
-			require.NoError(t, err)
-			expected := fmt.Sprintf(`{"store_id":"%s","tuple_key":{"object":"document:1","relation":"viewer","user":"user:anne"},"contextual_tuples":null,"authorization_model_id":"","trace":false}`, storeID)
-			require.JSONEq(t, string(raw), expected)
-		case "raw_response":
-			raw, err := json.Marshal(&field.Interface)
-			require.NoError(t, err)
-			expected := `{"allowed":true,"resolution":""}`
-			require.JSONEq(t, string(raw), expected)
-		case "authorization_model_id":
-			require.Equal(t, field.String, authorizationModelID)
-		case "store_id":
-			require.Equal(t, field.String, storeID)
-		case "datastore_query_count":
-			require.GreaterOrEqual(t, field.Integer, int64(1))
-			require.LessOrEqual(t, field.Integer, int64(2))
-		case "peer.address":
-		case "request_id":
-			require.NotEmpty(t, field.String)
-		default:
-			require.Fail(t, "unexpected field: %s", field.Key)
-		}
+	type test struct {
+		_name           string
+		req             *openfgav1.CheckRequest
+		expectedContext map[string]interface{}
 	}
+
+	tests := []test{
+		{
+			_name: "check_success",
+			req: &openfgav1.CheckRequest{
+				AuthorizationModelId: authorizationModelID,
+				StoreId:              storeID,
+				TupleKey:             tuple.NewTupleKey("document:1", "viewer", "user:anne"),
+			},
+			expectedContext: map[string]interface{}{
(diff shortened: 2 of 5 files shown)


tests/check/check_test.go
```diff
@@ -33,12 +37,133 @@ func TestCheckMySQL(t *testing.T) {
...
+		},
+	})
+	require.NoError(t, err)
+
+	resp, err := client.Check(context.Background(), &openfgav1.CheckRequest{
+		StoreId:  storeID,
+		TupleKey: tuple.NewTupleKey("document:1", "viewer", "user:anne"),
+	})
+	require.NoError(t, err)
+	require.True(t, resp.Allowed)
```
miparnisari, line 103: This is the "happy path" Check. If we want to add tests for the logs in error cases, we have to change this test quite a bit.

I suggest that you change the test to use a table-driven approach, similar to how we do our unit tests in other parts of the repo ([example](https://github.com/openfga/openfga/blob/ff41e318068dc0fadcb26186a91202bb729eab59/pkg/server/commands/write_test.go#L19-L90))

Here you could do something like this:

```go
         type test struct {
		_name           string
		req             *openfgav1.CheckRequest
		expectedContext map[string]string
	}
	tests := []test{
		{
			_name: "check_success",
			req: &openfgav1.CheckRequest{
				AuthorizationModelId: authorizationModelID,
				StoreId:              storeID,
				TupleKey:             tuple.NewTupleKey("document:1", "viewer", "user:anne"),
			},
			expectedContext: map[string]string{
				"grpc_service": "openfga.v1.OpenFGAService",
				"grpc_method":  "Check",
				"raw_response": `{"allowed":true,"resolution":""}`,
			},
		},
		{
			_name: "check_fail",
			req: &openfgav1.CheckRequest{
				AuthorizationModelId: authorizationModelID,
				StoreId:              "123",
				TupleKey:             tuple.NewTupleKey("document:1", "viewer", "user:anne"),
			},
			expectedContext: map[string]string{
				"grpc_service": "openfga.v1.OpenFGAService",
				"grpc_method":  "Check",
				"raw_response": `{"code":"validation_error","message":"invalid CheckRequest.StoreId: value does not match regex pattern \"^[ABCDEFGHJKMNPQRSTVWXYZ0-9]{26}$\""}`,
			},
		},
	}

	for _, test := range tests {
		t.Run(test._name, func(t *testing.T) {
			_, err = client.Check(context.Background(), test.req)

			// read the last log emitted
			fields := logs.All()[len(logs.All())-1].ContextMap()

			// all the assertions go here
			require.JSONEq(t, test.expectedContext["raw_response"], string(fields["raw_response"].(json.RawMessage)))
		})
	}
```

(That last test is failing because we are not logging anything in case of an error 😱 )

tests/check/check_test.go
```diff
@@ -33,12 +37,133 @@ func TestCheckMySQL(t *testing.T) {
...
+			require.NotEmpty(t, field.String)
+		default:
+			require.Fail(t, "unexpected field: %s", field.Key)
+		}
+	}
+}
+
 func testRunAll(t *testing.T, engine string) {
 	cfg := run.MustDefaultConfigWithRandomPorts()
 	cfg.Log.Level = "none"
```
miparnisari, line 160: it's confusing that here we set the log level to `none` but then we pass a concrete `logger`.

tests/check/check_test.go
```diff
@@ -33,12 +37,133 @@ func TestCheckMySQL(t *testing.T) {
...
 	cfg.Datastore.Engine = engine
 
-	cancel := tests.StartServer(t, cfg)
+	logger := logger.NewNoopLogger()
+	serverCtx := &run.ServerContext{Logger: logger}
+
+	cancel := tests.StartServerWithContext(t, cfg, serverCtx)
```
miparnisari, line 166: Why does this have to change? All these other tests can have the logger enabled, that shouldn't affect the test that you wrote

miparnisari (changes requested):

tests/check/check_test.go
jpadilla: removed changes, they're not required.

tests/check/check_test.go
```diff
@@ -33,12 +37,133 @@ func TestCheckMySQL(t *testing.T) {
...
+
+	expectedLogs := filteredLogs.All()
+	require.Len(t, expectedLogs, 1)
+
+	fields := expectedLogs[0].Context
+
+	for _, field := range fields {
+		switch field.Key {
+		case "grpc_service":
+			require.Equal(t, field.String, "openfga.v1.OpenFGAService")
```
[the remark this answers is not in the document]
jpadilla: done

tests/check/check_test.go
```diff
@@ -33,12 +37,133 @@ func TestCheckMySQL(t *testing.T) {
...
+		return false
+	})
+
+	expectedLogs := filteredLogs.All()
+	require.Len(t, expectedLogs, 1)
+
+	fields := expectedLogs[0].Context
+
+	for _, field := range fields {
+		switch field.Key {
```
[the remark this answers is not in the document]
jpadilla: ah yep 👍

tests/check/check_test.go
```diff
@@ -33,6 +37,130 @@ func TestCheckMySQL(t *testing.T) {
...
+			require.Equal(t, test.expectedContext["grpc_service"], fields["grpc_service"])
+			require.Equal(t, test.expectedContext["grpc_method"], fields["grpc_method"])
+			require.Equal(t, test.expectedContext["grpc_type"], fields["grpc_type"])
+			require.Equal(t, test.expectedContext["grpc_code"], fields["grpc_code"])
+			require.JSONEq(t, test.expectedContext["raw_request"].(string), string(fields["raw_request"].(json.RawMessage)))
+			require.JSONEq(t, test.expectedContext["raw_response"].(string), string(fields["raw_response"].(json.RawMessage)))
+			require.Equal(t, test.expectedContext["authorization_model_id"], fields["authorization_model_id"])
+			require.Equal(t, test.expectedContext["store_id"], fields["store_id"])
+			require.GreaterOrEqual(t, fields["datastore_query_count"], int64(1))
+			require.LessOrEqual(t, fields["datastore_query_count"], int64(2))
```
miparnisari, line 155: We don't need to be so specific here. We have separate tests that check the value for this number. So just do `require.NotEmpty(t, fields["datastore_query_count"])`

pkg/middleware/http/handler.go
```diff
@@ -69,6 +70,9 @@ func CustomHTTPErrorHandler(ctx context.Context, w http.ResponseWriter, r *http.
 	// convert as error object
 	pb := err.ActualError
 
+	log.Println(err)
+	log.Println(pb)
```
miparnisari, line 74: 🤔

tests/check/check_test.go
```diff
@@ -33,6 +37,130 @@ func TestCheckMySQL(t *testing.T) {
...
+			require.NotEmpty(t, fields["peer.address"])
+			require.NotEmpty(t, fields["request_id"])
```
miparnisari, line 157: If you enable tracing, there is one more assertion you can do: `require.NotEmpty(t, fields["trace_id"])`

tests/check/check_test.go
```diff
@@ -33,6 +37,130 @@ func TestCheckMySQL(t *testing.T) {
 	testRunAll(t, "mysql")
 }
 
+func TestCheckLogs(t *testing.T) {
+	cfg := run.MustDefaultConfigWithRandomPorts()
+	cfg.Log.Level = "none"
```
miparnisari, line 42: Since you are passing a custom logger to `serverContext, `cfg.Log` will not be read within `StartServerWithContext`, right? So, this line is unnecessary..

tests/check/check_test.go
```diff
@@ -33,6 +37,130 @@ func TestCheckMySQL(t *testing.T) {
...
+							return true
+						}
+					}
+				}
+
+				return false
+			})
+
+			expectedLogs := filteredLogs.All()
+			require.Len(t, expectedLogs, 1)
```
miparnisari, line 142: this assertion will break when we add more tests 😅

tests/check/check_test.go
```diff
@@ -33,6 +37,130 @@ func TestCheckMySQL(t *testing.T) {
...
+	cfg.Datastore.Engine = "memory"
+
+	observerLogger, logs := observer.New(zap.DebugLevel)
+	serverCtx := &run.ServerContext{
+		Logger: &logger.ZapLogger{
+			Logger: zap.New(observerLogger),
+		},
+	}
+
+	cancel := tests.StartServerWithContext(t, cfg, serverCtx)
```
miparnisari, line 52: I would add a comment here that the reason why we have to start up a full fledged server is because otherwise the logging interceptor (or any other interceptor) will not be hooked up.

tests/check/check_test.go
miparnisari: @jpadilla Can you create an issue that we are not logging anything in case of 4xxs?

tests/check/check_test.go
jpadilla: fixed by clearing observed logs after reach run.

miparnisari (approved):

tests/check/check_test.go
```diff
@@ -33,6 +38,144 @@ func TestCheckMySQL(t *testing.T) {
...
+	// create mock OTLP server
+	otlpServerPort, otlpServerPortReleaser := run.TCPRandomPort()
+	localOTLPServerURL := fmt.Sprintf("localhost:%d", otlpServerPort)
+	otlpServerPortReleaser()
+	_, serverStopFunc, err := mocks.NewMockTracingServer(otlpServerPort)
+	defer serverStopFunc()
+	require.NoError(t, err)
+
+	cfg := run.MustDefaultConfigWithRandomPorts()
+	cfg.Trace.Enabled = true
```
miparnisari, line 51: @jpadilla i forget... why are we starting a tracing server for this test?

Issue #852, opened 2023-07-05
Improve `Check` speed by pruning userset lookups that cannot result in success


jbrooks2-godaddy:
The introduction of the `1.1` type system in model definitions paved the way for `Check` performance improvements. Some were implemented here: https://github.com/openfga/openfga/pull/545

I believe that we can build upon the changes introduced in that PR and use the type information to further prune and short-circuit userset lookups that cannot result in access being granted.

Consider this toy authorization model:
```
model
  schema 1.1
type user
type oauth_client
type document
  relations
    define owner: [user]
    define can_write: [oauth_client] or owner
```

And a `Check` request:
```
POST /check:
{
  "tuple_key": {
    "user": "oauth_client:123",
    "relation": "can_write",
    "object": "document:abc"
  }
}
```

The `Check` system ends up doing two things to resolve this in [`CheckDirect`](https://github.com/openfga/openfga/blob/main/internal/graph/check.go#L371-L511):
1. Checking the direct relationship; querying the database for the full `< document:abc, can_write, oauth_client:123 >`
2. Checking the userset relationship; querying the database for the list of tuples satisfying `object=document:abc, relation=owner`

https://github.com/openfga/openfga/pull/545 made some changes that improved both of these parts:
- (1) is completely avoided if the type definition makes the relationship impossible
- (2) passes a type filter to the storage system; instructing it to return fewer tuples.

I think that for some models, (2) can be short-circuited in a way similar to (1). In this example authorization model, an `oauth_client` cannot be an `owner` of a `document`. If the type system can be used to understand this restriction, then we can avoid a db call completely for (2).

jon-whit:
Yep, definitely. The relationship ingresses can be used to prune the search space. It's something we're aware of and have talked about adding optimizations for. We've also been working on the storage layer to reduce/filter the related types at the query layer. That will help as well.

jon-whit:
Just to elaborate on my last comment for any future readers.. Imagine you have a model such as

```
type user

type employee

type group
  relations
    define member: [employee]

type document
  relations
    define viewer: [user, group#member]
```

If you call `Check(document:1#viewer@user:jon)`, then the evaluation path for `group#member` relationships with `document#viewer` is an unecessary path in the search because there is no path in that subgraph that leads from `document#viewer` through `group#member` to a `user` object. `group#member` is only directly related to `employee` object types, and so there would not exist any path leading to a `user` object type from that subgraph.

[...]

PR #932, opened 2023-08-08
feat: optimizing check call by pruning user lookup

adriantam:
## Description

Do not get userset if only direct relations are possible

## References
First step for https://github.com/openfga/openfga/issues/852

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -465,6 +465,9 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		objectType := tuple.GetType(tk.GetObject())
 		relation := tk.GetRelation()
 
+		// directlyRelatedUsersetTypes could be "user:*" or "group#member"
+		directlyRelatedUsersetTypes, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
+
 		fn1 := func(ctx context.Context) (*ResolveCheckResponse, error) {
 			ctx, span := tracer.Start(ctx, "checkDirectUserTuple", trace.WithAttributes(attribute.String("tuple_key", tk.String())))
 			defer span.End()
@@ -500,9 +503,6 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			ctx, span := tracer.Start(ctx, "checkDirectUsersetTuples", trace.WithAttributes(attribute.String("userset", tuple.ToObjectRelationString(tk.Object, tk.Relation))))
 			defer span.End()
 
-			// allowedUsersetsTypeRestrictions could be "user:*" or "group#member"
-			allowedUsersetsTypeRestrictions, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
-
 			response := &ResolveCheckResponse{
 				Allowed: false,
 				ResolutionMetadata: &ResolutionMetadata{
@@ -513,7 +513,7 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			iter, err := c.ds.ReadUsersetTuples(ctx, storeID, storage.ReadUsersetTuplesFilter{
 				Object:                      tk.Object,
 				Relation:                    tk.Relation,
-				AllowedUserTypeRestrictions: allowedUsersetsTypeRestrictions,
+				AllowedUserTypeRestrictions: directlyRelatedUsersetTypes,
 			})
 			if err != nil {
 				return response, err
@@ -595,7 +595,6 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			checkFuncs = []CheckHandlerFunc{fn1}
 		}
 
-		directlyRelatedUsersetTypes, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
 		if len(directlyRelatedUsersetTypes) > 0 {
 			checkFuncs = append(checkFuncs, fn2)
 
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -218,7 +218,6 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
 // union implements a CheckFuncReducer that requires any of the provided CheckHandlerFunc to resolve
 // to an allowed outcome. The first allowed outcome causes premature termination of the reducer.
 func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*ResolveCheckResponse, error) {
-
 	ctx, cancel := context.WithCancel(ctx)
 	resultChan := make(chan checkOutcome, len(handlers))
 
@@ -496,21 +495,6 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			return response, nil
 		}
 
-		var checkFuncs []CheckHandlerFunc
-
-		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_0 {
-			checkFuncs = append(checkFuncs, fn1)
-		} else {
-			shouldCheckDirectTuple, _ := typesys.IsDirectlyRelated(
-				typesystem.DirectRelationReference(objectType, relation),                                         //target
-				typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
-			)
-
-			if shouldCheckDirectTuple {
-				checkFuncs = append(checkFuncs, fn1)
-			}
-		}
-
 		fn2 := func(ctx context.Context) (*ResolveCheckResponse, error) {
 			ctx, span := tracer.Start(ctx, "checkDirectUsersetTuples", trace.WithAttributes(attribute.String("userset", tuple.ToObjectRelationString(tk.Object, tk.Relation))))
 			defer span.End()
@@ -602,7 +586,25 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			return union(ctx, c.concurrencyLimit, handlers...)
 		}
 
-		checkFuncs = append(checkFuncs, fn2)
+		var checkFuncs []CheckHandlerFunc
+
+		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_0 {
+			checkFuncs = []CheckHandlerFunc{fn1, fn2}
+		} else {
+			shouldCheckDirectTuple, _ := typesys.IsDirectlyRelated(
+				typesystem.DirectRelationReference(objectType, relation),                                         //target
+				typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
+			)
+
+			if shouldCheckDirectTuple {
+				checkFuncs = []CheckHandlerFunc{fn1}
+			}
+
+			directlyRelatedUserTypesHasUsersetTuples, _ := typesys.DirectlyRelatedUserTypesHasUsersetTuples(objectType, relation)
+			if directlyRelatedUserTypesHasUsersetTuples {
+				checkFuncs = append(checkFuncs, fn2)
+			}
+		}
 
 		return union(ctx, c.concurrencyLimit, checkFuncs...)
 	}
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -588,22 +588,19 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 
 		var checkFuncs []CheckHandlerFunc
 
-		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_0 {
-			checkFuncs = []CheckHandlerFunc{fn1, fn2}
-		} else {
-			shouldCheckDirectTuple, _ := typesys.IsDirectlyRelated(
-				typesystem.DirectRelationReference(objectType, relation),                                         //target
-				typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
-			)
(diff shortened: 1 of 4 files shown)


pkg/typesystem/typesystem.go
```diff
@@ -262,6 +262,21 @@ func (t *TypeSystem) GetDirectlyRelatedUserTypes(objectType, relation string) ([
 	return r.GetTypeInfo().GetDirectlyRelatedUserTypes(), nil
 }
 
+// DirectlyRelatedUserTypesHasUsersetTuples returns whether any of the directly user related types requires lookup
+// for userset tuples
```
miparnisari, line 266:
```suggestion
// DirectlyRelatedUserTypesHasUsersetTuples returns whether any of the directly user related types is a userset
```

internal/graph/check.go
```diff
@@ -602,7 +586,25 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			return union(ctx, c.concurrencyLimit, handlers...)
 		}
 
-		checkFuncs = append(checkFuncs, fn2)
+		var checkFuncs []CheckHandlerFunc
+
+		if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_0 {
+			checkFuncs = []CheckHandlerFunc{fn1, fn2}
```
miparnisari, line 592: we don't need this code anymore

miparnisari:
In the PR description you say

> First step for https://github.com/openfga/openfga/issues/852

What does this mean? What is the next step?

internal/graph/check.go
adriantam: updated

pkg/typesystem/typesystem.go
adriantam: updated

adriantam:
> In the PR description you say
> 
> > First step for #852
> 
> What does this mean? What is the next step?

The ideal solution is to check with graph/ingress to see if there is a path and don't need to check userset if the types will not lead to any solution. However, this work, as seen in the TestCheckDatastoreQueryCount updates, already provide very good optimization.

internal/graph/check_test.go
```diff
@@ -279,8 +279,8 @@ func TestCheckDatastoreQueryCount(t *testing.T) {
 			name:       "union_or_ttu_no_access",
 			check:      tuple.NewTupleKey("document:x", "union_or_ttu", "user:unknown"),
 			allowed:    false,
-			minDBReads: 11, // union (4 reads) + ttu (3 reads) + union rewrite (4 reads)
-			maxDBReads: 11,
+			minDBReads: 6, // union (2 reads) + ttu (2 reads) + union rewrite (2 reads)
```
elbuo8, line 282: 🚀

internal/graph/check.go
```diff
@@ -602,7 +586,22 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+			typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
+		)
+
+		if shouldCheckDirectTuple {
+			checkFuncs = []CheckHandlerFunc{fn1}
+		}
+
+		directlyRelatedUserTypesHasUsersetTuples, _ := typesys.DirectlyRelatedUserTypesHasUsersetTuples(objectType, relation)
+		if directlyRelatedUserTypesHasUsersetTuples {
+			checkFuncs = append(checkFuncs, fn2)
```
miparnisari, line 602: This change is good because if the Check is done over, for example, `define viewer: [user, employee]`, we will not call fn2.

However, it's not enough... if you have a mixture of users and usersets, e.g. `define viewer: [user, group#member]`, then inside code `fn2` we are still sending `user` in the call to `ReadUsersetTuples` which is inefficient because `user` is not a userset.

I would change this code to

```go
listOfUsersets := typesys.DirectlyRelatedUserTypesThatAreUsersets(objectType, relation)
if len(listOfUsersets) > 0 {
	checkFuncs = append(checkFuncs, fn2)
}
```

And then inside `fn2` change 505 to be

```go
allowedUserTypeRestrictions := typesys.DirectlyRelatedUserTypesThatAreUsersets(objectType, relation)
```

miparnisari (changes requested):

internal/graph/check.go
miparnisari: Also, i'm not a fan of this code being outside `fn1` and `fn2` when really, they belong there.

How about we have

```go
var checkFuncs = []CheckHandlerFunc{fn1, fn2}
return union(ctx, c.concurrencyLimit, checkFuncs...)
```

and we move these optimization checks within each fn?

```go
fn1 := func(ctx context.Context) (*ResolveCheckResponse, error) {
			ctx, span := tracer.Start(ctx, "checkDirectUserTuple", trace.WithAttributes(attribute.String("tuple_key", tk.String())))
			defer span.End()

			response := &ResolveCheckResponse{
				Allowed: false,
					DatastoreQueryCount: req.GetResolutionMetadata().DatastoreQueryCount + 1,
				},
			}

			shouldCheckDirectTuple, _ := typesys.IsDirectlyRelated(
				typesystem.DirectRelationReference(objectType, relation),                                         //target
				typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
			)

			if !shouldCheckDirectTuple {
				return response, nil
			}
```

and

```go
fn2 := func(ctx context.Context) (*ResolveCheckResponse, error) {
			ctx, span := tracer.Start(ctx, "checkDirectUsersetTuples", trace.WithAttributes(attribute.String("userset", tuple.ToObjectRelationString(tk.Object, tk.Relation))))
			defer span.End()

			response := &ResolveCheckResponse{
				Allowed: false,
					DatastoreQueryCount: req.GetResolutionMetadata().DatastoreQueryCount + 1,
				},
			}

			allowedUserTypeRestrictions := typesys.DirectlyRelatedUserTypesThatAreUsersets(objectType, relation)
			if len(allowedUserTypeRestrictions) == 0 {
				return response, nil
			}
```

elbuo8: You needlessly launch goroutines, and memory allocations, empty spans instead of short circuiting early.

adriantam: > Also, i'm not a fan of this code being outside fn1 and fn2 when really, they belong there.

As noted by @elbuo8 , we do not want to launch the GoRoutine itself as it responds more slowly + require go routines allocation.

miparnisari: > That ends up launching goroutines, memory allocations

This is true, but we gain in readability because these checks happen where they are needed

> empty spans

How so? The code would be included in span `checkDirectUsersetTuples` which should be short-lived in these cases

internal/graph/check.go
```diff
@@ -496,30 +495,12 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
 			ctx, span := tracer.Start(ctx, "checkDirectUsersetTuples", trace.WithAttributes(attribute.String("userset", tuple.ToObjectRelationString(tk.Object, tk.Relation))))
 			defer span.End()
 
-			var allowedUserTypeRestrictions []*openfgav1.RelationReference
-			if typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 {
-				// allowedUserTypeRestrictions could be "user" or "user:*" or "group#member"
-				allowedUserTypeRestrictions, _ = typesys.GetDirectlyRelatedUserTypes(objectType, relation)
-			}
+			// allowedUserTypeRestrictions could be "user" or "user:*" or "group#member"
+			allowedUserTypeRestrictions, _ := typesys.DirectlyRelatedUserTypesHasUsersetTuples(objectType, relation)
```
miparnisari, line 503:
```suggestion
			// allowedUsersetsTypeRestrictions could be "user:*" or "group#member"
			allowedUsersetsTypeRestrictions, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
```

internal/graph/check.go
adriantam: > This is true, but so far the bottlenecks we've seen are only about DB connections, no?

We had seen evidence in the past where some "large/complex" check requests causes GoRoutine / Garbage collection spikes because of the huge number of concurrent DB requests.  If there is any way to reduce load on the system, why do we want to do it otherwise?

internal/graph/check.go
adriantam: updated

pkg/typesystem/typesystem.go
```diff
@@ -262,6 +262,21 @@ func (t *TypeSystem) GetDirectlyRelatedUserTypes(objectType, relation string) ([
 	return r.GetTypeInfo().GetDirectlyRelatedUserTypes(), nil
 }
 
+// DirectlyRelatedUserTypesHasUsersetTuples returns whether any of the directly user related types is a userset
+func (t *TypeSystem) DirectlyRelatedUserTypesHasUsersetTuples(objectType, relation string) ([]*openfgav1.RelationReference, error) {
```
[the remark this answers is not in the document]
adriantam: updated

internal/graph/check.go
jon-whit: > Also, i'm not a fan of this code being outside fn1 and fn2 when really, they belong there.

We have no need to resolve fn1 and/or fn2 if there is not a type restriction that would lead to a resolution path. It's wasted cycles and GC and memory etc..

My vote here is to move this code outside and queue up/append the function handlers IFF they are needed based on the type restrictions.

```go
fn1 := func(...)
fn2 := func(...)

var checkFuncs = []CheckHandlerFunc{fn1, fn2}
if (fn1Condition) {
  checkFuncs = append(checkFuncs, fn1)
}

if (fn2Condition) {
  checkFuncs = append(checkFuncs, fn2)
}

return union(..., checkFuncs...)
```

pkg/typesystem/typesystem.go
```diff
@@ -262,6 +262,21 @@ func (t *TypeSystem) GetDirectlyRelatedUserTypes(objectType, relation string) ([
 	return r.GetTypeInfo().GetDirectlyRelatedUserTypes(), nil
 }
 
+// DirectlyRelatedUsersets returns a list of the directly user related types that are usersets
+func (t *TypeSystem) DirectlyRelatedUsersets(objectType, relation string) ([]*openfgav1.RelationReference, error) {
+	refs, err := t.GetDirectlyRelatedUserTypes(objectType, relation)
+	var usersetRelationReferences []*openfgav1.RelationReference
+	if err != nil {
+		return usersetRelationReferences, err
+	}
```
jon-whit, line 271: Newline please

pkg/typesystem/typesystem.go
```diff
@@ -262,6 +262,21 @@ func (t *TypeSystem) GetDirectlyRelatedUserTypes(objectType, relation string) ([
 	return r.GetTypeInfo().GetDirectlyRelatedUserTypes(), nil
 }
 
+// DirectlyRelatedUsersets returns a list of the directly user related types that are usersets
+func (t *TypeSystem) DirectlyRelatedUsersets(objectType, relation string) ([]*openfgav1.RelationReference, error) {
```
jon-whit, line 266: Please add tests for this.

internal/graph/check.go
```diff
@@ -218,7 +218,6 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
 // union implements a CheckFuncReducer that requires any of the provided CheckHandlerFunc to resolve
 // to an allowed outcome. The first allowed outcome causes premature termination of the reducer.
 func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*ResolveCheckResponse, error) {
-
```
jon-whit, line 221: `intersection` and `exclusion` have a leading newline. Please either make consistent or add the visual whitespace back.

internal/graph/check.go
jon-whit: @adriantam sorry, I was reading the diff wrong. That's what we're doing below `fn2`. 👍

internal/graph/check.go
```diff
@@ -602,7 +583,22 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+		shouldCheckDirectTuple, _ := typesys.IsDirectlyRelated(
+			typesystem.DirectRelationReference(objectType, relation),                                         //target
+			typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
+		)
+
+		if shouldCheckDirectTuple {
+			checkFuncs = []CheckHandlerFunc{fn1}
+		}
+
+		directlyRelatedUserTypesHasUsersetTuples, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
```
jon-whit, line 597:
```suggestion
		directlyRelatedUsersetTypes, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
```

internal/graph/check.go
```diff
@@ -602,7 +584,22 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+		shouldCheckDirectTuple, _ := typesys.IsDirectlyRelated(
+			typesystem.DirectRelationReference(objectType, relation),                                         //target
+			typesystem.DirectRelationReference(tuple.GetType(tk.GetUser()), tuple.GetRelation(tk.GetUser())), //source
+		)
+
+		if shouldCheckDirectTuple {
+			checkFuncs = []CheckHandlerFunc{fn1}
+		}
+
+		directlyRelatedUsersetTypes, _ := typesys.DirectlyRelatedUsersets(objectType, relation)
```
jon-whit, line 598: Move this above `fn1` and `fn2` and reference `directlyRelatedUserTypes` in `fn2` instead of recalling `typesys.DirectlyRelatedUsersets(..)` inside of that handler. It's just duplicate work.

jon-whit (approved):

PR #948, opened 2023-08-17
refactor: have Check construct requestTupleReader internally instead of leaving it up to clients

miparnisari:
## Description
Address this TODO: https://github.com/openfga/openfga/blob/1f9c6370976a56dc0535309f02e3a12c4880a372/internal/graph/check_test.go#L305

Refactor: have Check construct requestTupleReader internally instead of leaving it up to clients.

With this change, we could (if we want) send a followup PR where we create `NewLocalChecker` on server creation and not once per Check request.


internal/graph/check.go
```diff
@@ -459,7 +460,9 @@ func (c *LocalChecker) ResolveCheck(
 		return nil, fmt.Errorf("relation '%s' undefined for object type '%s'", relation, objectType)
 	}
 
-	resp, err := union(ctx, c.concurrencyLimit, c.checkRewrite(ctx, req, rel.GetRewrite()))
+	reqDatastore := storagewrappers.NewRequestTupleReader(c.ds, req.GetContextualTuples())
```
miparnisari, line 463: I could also make this object a part of `ResolveCheckRequest` 🤔

jon-whit: I'm not sure I'm aligned with this change and/or the one you mentioned as an alternative (in the ResolveCheckRequest). The reason we introduced the `CombinedTupleReader` in the first place to was move this responsibility outside of Check.

We moved datastore specific code outside of Check in the first place so as to not conflate the implementation details with contextual tuples. As far as Check is concerned, it simply yields tuples from an iterator that comes from the database iterator provided. Those tuples may be due to contextual tuples or ones that have been persisted, but Check doesn't care. Tuples are tuples.

It's not Checks responsibility to construct a datastore derivative from the contextual tuples provided in the request, that's the upstream callers responsibility if that's the intent.

Moving this inside of Check is tangential to that initial intent.


So I'm a little torn on these changes..

miparnisari: >  As far as Check is concerned, it simply yields tuples from an iterator that comes from the database iterator provided. Those tuples may be due to contextual tuples or ones that have been persisted, but Check doesn't care. Tuples are tuples. 

That specific abstraction is provided by the `requestTupleReader`

> It's not Checks responsibility to construct a datastore derivative from the contextual tuples provided in the request, that's the upstream callers responsibility if that's the intent.

I don't agree. Check has a field in the request called `ContextualTuples`. It should know how to handle that.

Look at the test code we have today:

```go
checker := NewLocalChecker(storagewrappers.NewCombinedTupleReader(ds, test.contextualTuples))

res, err := checker.ResolveCheck(ctx, &ResolveCheckRequest{
	ContextualTuples:   test.contextualTuples,
	// ...
})
```

As a developer writing this, I ask myself: why am i given two places to pass in the contextual tuples? What happens if I forget to pass in one of them?

This is the ambiguity I want to get rid of.

pkg/storage/storagewrappers/requesttuplereader.go
```diff
@@ -8,18 +8,18 @@ import (
 	"github.com/openfga/openfga/pkg/tuple"
 )
 
-// NewCombinedTupleReader returns a TupleReader that reads from a persistent datastore and from the contextual
+// NewRequestTupleReader returns a TupleReader that reads from a persistent datastore and from the contextual
```
jon-whit, line 11: Can we call this `ContextualTupleReader` instead?

internal/graph/check.go
jon-whit: I redact my original comment. It's not unreasonable to expect the Check implementation to create a derivative iterator that yields the contextual tuples.  The original goal was to avoid manually iterating over the contextual tuples and validating/filtering them and doing evaluations against them etc.. We shouldn't have to treat the contextual tuples any different than tuples coming from the storage iterator. As long as we keep contextual tuples abstracted away from the primary Check code paths, then I'm happy 😄 

Since the contextual tuples are request specific and the derivative storage implementation has to be constructed per request, then I'm ok with this approach. But if we start seeing very specific contextual tuple code in critical code paths, then we should reconsider what we're doing.

internal/graph/check.go
```diff
@@ -459,7 +460,9 @@ func (c *LocalChecker) ResolveCheck(
 		return nil, fmt.Errorf("relation '%s' undefined for object type '%s'", relation, objectType)
 	}
 
-	resp, err := union(ctx, c.concurrencyLimit, c.checkRewrite(ctx, req, rel.GetRewrite()))
+	reqDatastore := storagewrappers.NewContextualTupleReader(c.ds, req.GetContextualTuples())
```
jon-whit, line 463:
```suggestion
	contextualDatastore := storagewrappers.NewContextualTupleReader(c.ds, req.GetContextualTuples())
```

internal/graph/check.go
```diff
@@ -109,6 +109,7 @@ type checkOutcome struct {
 }
 
 type LocalChecker struct {
+	// NOTE it doesn't have the request's contextual tuples.
```
jon-whit, line 112: The original design behind LocalChecker was not meant to have a LocalChecker constructed per request. A LocalChecker only needs to be constructed once for the lifetime of the server, and so the Checker doesn't have fields specific per request. It something should be enforced per request, then it should belong in the `ResolveCheck` implementation.

jon-whit:
The original `LocalChecker` was designed to be constructed once for the lifetime of the server, and on each request the contextual tuples were propagated through `context` to the storage layer and the combined tuple iterator was used under the covers. This design pattern allowed for the composition of a global CheckResolver that IMO made things more elegant and straightforward, but I recognize that it may not have been as apparent or easily understandable for other developers. 

As time has passed, we've moved completely away from a global LocalChecker to constructing a [LocalChecker _per request_](https://github.com/openfga/openfga/blob/main/pkg/server/server.go#L445). This has been the work of multiple PRs over time because of the distaste for passing things through context. I still believe this is the wrong approach, because it creates a lot of memory allocs and GC activity in a completely unnecessary manner, but if we're going to do this, then I ask that we rework the `LocalChecker` in general to match its intended usage/semantics with the current state of things.

You shouldn't have to modify `checkDirect` and `checkTupleToUserset` signatures in this PR, becuase the `LocalChecker` should encode the per request state that it needs to operate. For example, since we've effectively moved to the LocalChecker being constructed per request anyways, then I'd suggest that we change the semantics of LocalChecker to something like:

```go
type LocalChecker struct {
    contextualDatastore storage.RelationshipTupleReader
}

NewLocalChecker(WithCheckRequest(req), ...)
```
e.g. you construct the `contextualDatastore` when you construct the LocalChecker, and you create the LocalChecker from the incoming request.

Then in the code we can do
```go
func (l *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
    ...
    // use l.contextualDatastore instead of passing it through as a new parameter in the signature
}
```

Since we've evolved things beyond their original design intent, then I'd prefer we repurpose and refactor a bit as part of the scope of this work, because they are relevant to one another in this case.

internal/graph/check.go
jon-whit: See https://github.com/openfga/openfga/pull/948#issuecomment-1690422567

internal/graph/check.go
```diff
@@ -471,7 +474,7 @@ func (c *LocalChecker) ResolveCheck(
 // 'object#relation'. The first handler looks up direct matches on the provided 'object#relation@user',
 // while the second handler looks up relationships between the target 'object#relation' and any usersets
 // related to it.
-func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
+func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckRequest, reqDatastore *storagewrappers.ContextualTupleReader) CheckHandlerFunc {
```
jon-whit, line 477: See https://github.com/openfga/openfga/pull/948#issuecomment-1690422567

pkg/storage/storagewrappers/contextualtuplereader.go
```diff
@@ -8,18 +8,18 @@ import (
...
+// NewContextualTupleReader returns a TupleReader that reads from a persistent datastore and from the contextual
 // tuples specified in the request
-func NewCombinedTupleReader(ds storage.RelationshipTupleReader, contextualTuples []*openfgav1.TupleKey) storage.RelationshipTupleReader {
-	return &combinedTupleReader{RelationshipTupleReader: ds, contextualTuples: contextualTuples}
+func NewContextualTupleReader(ds storage.RelationshipTupleReader, contextualTuples []*openfgav1.TupleKey) *ContextualTupleReader {
+	return &ContextualTupleReader{RelationshipTupleReader: ds, contextualTuples: contextualTuples}
 }
 
-type combinedTupleReader struct {
+type ContextualTupleReader struct {
```
jon-whit, line 17:
```suggestion
type contextualTupleReader struct {
```

This doesn't need to be exported. We want to encourage upstream to depend on the interface `storage.RelationshipTupleReader` instead of this directly.

pkg/storage/storagewrappers/contextualtuplereader.go
```diff
@@ -8,18 +8,18 @@ import (
 	"github.com/openfga/openfga/pkg/tuple"
 )
 
-// NewCombinedTupleReader returns a TupleReader that reads from a persistent datastore and from the contextual
+// NewContextualTupleReader returns a TupleReader that reads from a persistent datastore and from the contextual
 // tuples specified in the request
-func NewCombinedTupleReader(ds storage.RelationshipTupleReader, contextualTuples []*openfgav1.TupleKey) storage.RelationshipTupleReader {
-	return &combinedTupleReader{RelationshipTupleReader: ds, contextualTuples: contextualTuples}
+func NewContextualTupleReader(ds storage.RelationshipTupleReader, contextualTuples []*openfgav1.TupleKey) *ContextualTupleReader {
```
jon-whit, line 13:
```suggestion
func NewContextualTupleReader(ds storage.RelationshipTupleReader, contextualTuples []*openfgav1.TupleKey) *contextualTupleReader {
```

miparnisari:
> ```go
> NewLocalChecker(WithContextualTuples(contextualTuples), ...)
> ```

But then we're right where we are now, no?  You'd have to pass the contextual tuples twice

```go
c := NewLocalChecker(WithContextualTuples(req.ContextualTuples.GetTupleKeys()), ...)
c.ResolveCheck(ctx, &graph.ResolveCheckRequest{
		...
		ContextualTuples:     req.ContextualTuples.GetTupleKeys(),
})
```

Or we could remove the 2nd parameter of ResolveCheck and pass the entire request to `NewLocalChecker`.

jon-whit:
> > ```go
> > NewLocalChecker(WithContextualTuples(contextualTuples), ...)
> > ```
> 
> But then we're right where we are now, no? You'd have to pass the contextual tuples twice
> 
> ```go
> c := NewLocalChecker(WithContextualTuples(req.ContextualTuples.GetTupleKeys()), ...)
> c.ResolveCheck(ctx, &graph.ResolveCheckRequest{
> 		...
> 		ContextualTuples:     req.ContextualTuples.GetTupleKeys(),
> })
> ```

If you construct the derivative datastore from the contextual tuples up front, then propagating the ContextualTuples down the call chain is redundant. You could omit this altogether. I don't see anything inherently wrong with passing the pointer around and propagating the request body, but yeah it's not necesssary because the datastore constructed up front will already have a reference to the contextual tuples.

miparnisari:
> I think for readability and understandability it makes sense to forward the request body in its entirety though.

I agree, but today we have to pass each field explicitly, so i argue that today it's not readable

https://github.com/openfga/openfga/blob/55f177adfeddc27323f2af5b0c55f4bb5b7d3191/pkg/server/server.go#L478-L493

pkg/storage/storagewrappers/contextualtuplereader.go
miparnisari: I need it to be exported so that I can reference it explicitly as function parameters within `check`: https://github.com/openfga/openfga/pull/948/files#diff-5074b7eeccca5be19ff0fa2a8f712335a6ed075628b66a46697ca9d8ca99fbe0R765

internal/graph/check.go
miparnisari: > LocalChecker was not meant to be constructed per request originally.

I could change this PR so that every request needs a distinct `LocalChecker`. Then, `LocalChecker.ds` would become `contextualTupleReader` and this comment would go away.

jon-whit:
Closing for now as this has become quite stale. We can revisit it at a later time.

tests/functional_test.go
```diff
@@ -103,235 +102,6 @@ condition conds(s: string) {
 	require.Nil(t, checkResp)
 }
 
-// TODO make a unit test from this
-func TestCheckWithQueryCacheEnabled(t *testing.T) {
```
miparnisari, line 107: I never liked this test introduced in https://github.com/openfga/openfga/pull/1059. The bug was in `check.go` but we wrote it as a functional tests. So I rewrote it in two parts.

Issue #949, opened 2023-08-18
Add label on datastore_bounded_read_delay_ms histogram


adriantam:
Currently, datastore_bounded_read_delay_ms history does not have label to distinguish whether it is from check or list object.  It will be nice to have labels distinguishing them.

Issue #947, opened 2023-08-17
Adding tags based on if datastore_query_count > configurable value


adriantam:
Now that we have [datastore query count](https://github.com/openfga/openfga/pull/924), it will be useful to add tags on whether the datastore_query_count is larger than a configurable value.  This allows observability into whether latency is caused by complex model or whether it is due to issues within the system.

PR #950, opened 2023-08-18
feat: adding check latency histogram with db query count label

adriantam:
## Description
Adding latency histogram that has label for method (only check for now) and database query count.

Sample Grafana:
Screenshot 2023-08-18 at 5 36 37 PM

## References
Close https://github.com/openfga/openfga/issues/949

## Review Checklist


cmd/run/run.go
```diff
@@ -185,6 +186,9 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Duration("check-query-cache-ttl", defaultConfig.CheckQueryCache.TTL, "if caching of Check and ListObjects is enabled, this is the TTL of each value")
 
+	// Unfortunately UintSlice/IntSlice does not work well when used as environment variable, we need to stick with string slice and convert back to integer
+	flags.StringSlice("latency-db-query-count-buckets", defaultConfig.LatencyDBQueryCountBuckets, "db query count buckets used in labelling latency histogram")
```
adriantam, line 190: See https://github.com/spf13/viper/issues/1611

pkg/server/server.go
```diff
@@ -68,6 +70,17 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method"})
+
+	latencyHistogramName = "latency"
```
adriantam, line 74: I am open to making this more specific to check latency only.  However, I envision we may want to have something similar for list object which makes me want to make this more flexible.

pkg/server/server.go
```diff
@@ -418,6 +447,7 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		attribute.KeyValue{Key: "user", Value: attribute.StringValue(tk.GetUser())},
 	))
 	defer span.End()
+	start := time.Now()
```
jon-whit, line 450: Put this before the trace. If the trace setup takes a while it should be representative.

pkg/server/server.go
jon-whit: latency is super generic. I'd recommend a more specific metric name such as `request_duration_by_query_count_ms`. That way the name is a little more clear and indicates the unit of the underlying value as well.

pkg/server/server.go
```diff
@@ -68,6 +70,17 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method"})
+
+	latencyHistogramName = "latency"
+
+	latencyHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:                            latencyHistogramName,
+		Help:                            "Observed latency (in milliseconds)",
```
jon-whit, line 78: Please be more specific here

> The request duration (in ms) labeled by method and buckets of datastore query counts. This allows for reporting percentiles based on the number of datastore queries required to resolve the request.

.config-schema.json
```diff
@@ -56,6 +56,16 @@
             "default": 1000,
             "x-env-variable": "OPENFGA_LIST_OBJECTS_MAX_RESULTS"
         },
+        "latencyDBQueryCountBuckets": {
```
jon-whit, line 59: To be consistent with vernacular elsewhere, and the label itself, please use `Datastore` instead of `DB`.

pkg/server/server.go
adriantam: If we are moving to request_duration_by_query_count_ms, we should also update the flag/config from latencyDBQueryCountBuckets to requestDurationDatastoreQueryCountBuckets

jon-whit:
@adriantam this doesn't relate to issue #949 right? That issue refers to the metric `datastore_bounded_read_delay_ms`, and this PR deals with adding a new metric for capturing a histgoram based on Checks involving a certain number of datastore queries. They are different issues in my book.

adriantam:
> 

No.  This is more as to allow operator to understand how check query latency is related to db query count.

jon-whit:
> > @adriantam this doesn't relate to issue #949 right? That issue refers to the metric `datastore_bounded_read_delay_ms`, and this PR deals with adding a new metric for capturing a histgoram based on Checks involving a certain number of datastore queries. They are different issues in my book.
> 
> No. This is more as to allow operator to understand how check query latency is related to db query count.

`datastore_bounded_read_delay_ms` is it's own metric though. I'm confused how that has anything to do with this work? Maybe that issue just needs to be fixed so that it mentions adding a new histogram metric or something with a `datastore_query_count` label. That way it is accurate with what this work is doing.

adriantam:
> > > @adriantam this doesn't relate to issue #949 right? That issue refers to the metric `datastore_bounded_read_delay_ms`, and this PR deals with adding a new metric for capturing a histgoram based on Checks involving a certain number of datastore queries. They are different issues in my book.
> > 
> > 
> > No. This is more as to allow operator to understand how check query latency is related to db query count.
> 
> `datastore_bounded_read_delay_ms` is it's own metric though. I'm confused how that has anything to do with this work? Maybe that issue just needs to be fixed so that it mentions adding a new histogram metric or something with a `datastore_query_count` label. That way it is accurate with what this work is doing.

My bad.  This should fix https://github.com/openfga/openfga/issues/947 instead of https://github.com/openfga/openfga/issues/948

pkg/server/server.go
```diff
@@ -480,6 +511,11 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 	}
 
 	span.SetAttributes(attribute.KeyValue{Key: "allowed", Value: attribute.BoolValue(res.GetAllowed())})
+	requestDurationByQueryHistogram.WithLabelValues(
+		"Check",
```
jon-whit, line 515: Can we keep the server package and service prefix for this so that it's a FQN for the RPC? That's more standard practice and leaves us well set for introducing a v2 potentially in the future.

Just like we do above in https://github.com/openfga/openfga/pull/950/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R505-R506

cmd/run/run_test.go
```diff
@@ -1146,15 +1180,19 @@ func TestRunCommandConfigIsMerged(t *testing.T) {
...
 		require.Equal(t, "postgres", viper.GetString(datastoreEngineFlag))
 		require.Equal(t, "postgres://postgres:PASS2@127.0.0.1:5432/postgres", viper.GetString(datastoreURIFlag))
 		require.Equal(t, "1", viper.GetString("max-types-per-authorization-model"))
 		require.True(t, viper.GetBool("check-query-cache-enabled"))
 		require.Equal(t, uint32(33), viper.GetUint32("check-query-cache-limit"))
 		require.Equal(t, 5*time.Second, viper.GetDuration("check-query-cache-ttl"))
+
+		require.Equal(t, []string{"33", "44"}, viper.GetStringSlice("request-duration-datastore-query-count-buckets"))
 		return nil
 	}
```
jon-whit, line 1197: Could you please add an assertion to  the `TestServerMetricsReporting` test to make sure that we're reporting the `request_duration_by_query_count_ms` metric when Check is called?

.config-schema.json
```diff
@@ -56,6 +56,16 @@
             "default": 1000,
             "x-env-variable": "OPENFGA_LIST_OBJECTS_MAX_RESULTS"
         },
+        "requestDurationDatastoreQueryCountBuckets": {
+            "description": "Datastore query count buckets used in labelling request duration by request histogram statistics",
+            "type": "array",
+            "items": {
+                "minimum": 0,
```
jon-whit, line 63: Isn't the minimum 1 if you provide the config?

If you omit the config (e.g. if the config is nil) then the defaults are used, but otherwise you have to provide at least 1 bucket.

.config-schema.json
```diff
@@ -56,6 +56,16 @@
             "default": 1000,
             "x-env-variable": "OPENFGA_LIST_OBJECTS_MAX_RESULTS"
         },
+        "requestDurationDatastoreQueryCountBuckets": {
+            "description": "Datastore query count buckets used in labelling request duration by request histogram statistics",
```
jon-whit, line 60: The "request duration by request histogram statistics" stutters

Maybe just 
> Datastore query count buckets used to label the histogram metric for measuring request duration.

.config-schema.json
adriantam: I do not understand what you mean.  Minimum 0 means you cannot put a negative number for the bucket (i.e., there will never be datastore query count of -1).  It does not imply the size of the array.

I think what you want is "minItems": 1,

pkg/server/server.go
adriantam: > Can we keep the server package and service prefix for this so that it's a FQN for the RPC? That's more standard practice and leaves us well set for introducing a v2 potentially in the future.
> 
> Just like we do above in https://github.com/openfga/openfga/pull/950/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R505-R506

I am not sure what you mean.  Currently the [datastoreQueryCount](https://github.com/openfga/openfga/blob/37f48d1b54aa1a69d7a16aa1d0deb4549aef4280/pkg/server/server.go#L473) uses "check" as the method name.  We can change this from "Check" to "check".  However, if we are going to use the FQN, we need to change datastoreQueryCount as well.

adriantam: See https://github.com/openfga/openfga/pull/924 for datastoreQueryCountHistogram grpc_method

jon-whit: Right, but we also use `grpc_service` in the other metric. So each observation is fully qualified by the grpc service and grpc method that the metric measures.

I vote we be consistent with what we have currently.

adriantam: So in this case we have 3 dimension then

.config-schema.json
jon-whit: Oh, yes. `minItems: 1` but each item's minimum can be no smaller than 0. Yep.

pkg/server/server.go
jon-whit: It's consistent with what we're doing elsewhere, but if we want to flatten it into a single dimension instead of two that's where I suggest we `fmt.Sprintf("%s/%s", openfgav1.OpenFGAService_ServiceDesc.ServiceName, "Check")` for example

Issue #956, opened 2023-08-22
ListObjects should log number of database reads in trace information


miparnisari:
Similar to how we do in Check:

<img width="1897" alt="image" src="https://github.com/openfga/openfga/assets/5374887/eeb8526a-d41d-4481-ba98-db8b0f919879">

For all ListObjects calls, add a new attribute `datastore_query_count` in the top-level span with the number of calls to `storage.ReadStartingWithUser` + the sum of all `checkResponse.GetResolutionMetadata().DatastoreQueryCount`.

Issue #938, opened 2023-08-14
Add `datastore_query_count` observations to histogram in ListObjects


jon-whit:
In https://github.com/openfga/openfga/pull/924 we added a histogram to observe the `datastore_query_count` per Check request. 

ListObjects can also dispatch Check calls if an object is contained under an exclusion or intersection, and we should also observe 
the `datastore_query_count` and add it to the histogram in the ListObjects code path.

PR #959, opened 2023-08-23
feat: add datastore query count for list object on trace and histogram

adriantam:
## Description
Adding trace information on datastore query count for list object
Screenshot 2023-08-23 at 3 47 14 PM
Screenshot 2023-08-23 at 3 47 52 PM

## References
Close https://github.com/openfga/openfga/issues/956
Close https://github.com/openfga/openfga/issues/938

## Review Checklist

### DIFF
--- a/pkg/server/commands/connectedobjects/connected_objects.go
+++ b/pkg/server/commands/connectedobjects/connected_objects.go
@@ -156,15 +156,25 @@ type ConnectedObjectsResult struct {
 	ResultStatus ConditionalResultStatus
 }
 
+type ResolutionMetadata struct {
+	QueryCount *uint32
+}
+
+func NewResolutionMetadata() *ResolutionMetadata {
+	return &ResolutionMetadata{
+		QueryCount: new(uint32),
+	}
+}
+
 // Execute yields all the objects of the provided objectType that the given user has, possibly, a specific relation with
 // and sends those objects to resultChan. It MUST guarantee no duplicate objects sent.
 func (c *ConnectedObjectsQuery) Execute(
 	ctx context.Context,
 	req *ConnectedObjectsRequest,
 	resultChan chan<- *ConnectedObjectsResult,
-	dsQueryCount *uint32,
+	resolutionMetadata *ResolutionMetadata,
 ) error {
-	return c.execute(ctx, req, resultChan, nil, dsQueryCount)
+	return c.execute(ctx, req, resultChan, nil, resolutionMetadata)
 }
 
 func (c *ConnectedObjectsQuery) execute(
@@ -172,7 +182,7 @@ func (c *ConnectedObjectsQuery) execute(
 	req *ConnectedObjectsRequest,
 	resultChan chan<- *ConnectedObjectsResult,
 	currentIngress *graph.RelationshipIngress,
-	dsQueryCount *uint32,
+	resolutionMetadata *ResolutionMetadata,
 ) error {
 	ctx, span := tracer.Start(ctx, "connectedObjects.Execute", trace.WithAttributes(
 		attribute.String("target_type", req.ObjectType),
@@ -268,7 +278,7 @@ func (c *ConnectedObjectsQuery) execute(
 
 			switch innerLoopIngress.Type {
 			case graph.DirectIngress:
-				return c.reverseExpandDirect(subgctx, r, resultChan, dsQueryCount)
+				return c.reverseExpandDirect(subgctx, r, resultChan, resolutionMetadata)
 			case graph.ComputedUsersetIngress:
 				// lookup the rewritten target relation on the computed_userset ingress
 				return c.execute(subgctx, &ConnectedObjectsRequest{
@@ -282,9 +292,9 @@ func (c *ConnectedObjectsQuery) execute(
 						},
 					},
 					ContextualTuples: r.contextualTuples,
-				}, resultChan, innerLoopIngress, dsQueryCount)
+				}, resultChan, innerLoopIngress, resolutionMetadata)
 			case graph.TupleToUsersetIngress:
-				return c.reverseExpandTupleToUserset(subgctx, r, resultChan, dsQueryCount)
+				return c.reverseExpandTupleToUserset(subgctx, r, resultChan, resolutionMetadata)
 			default:
 				return fmt.Errorf("unsupported ingress type")
 			}
@@ -306,7 +316,7 @@ func (c *ConnectedObjectsQuery) reverseExpandTupleToUserset(
 	ctx context.Context,
 	req *reverseExpandRequest,
 	resultChan chan<- *ConnectedObjectsResult,
-	dsQueryCount *uint32,
+	resolutionMetadata *ResolutionMetadata,
 ) error {
 	ctx, span := tracer.Start(ctx, "reverseExpandTupleToUserset", trace.WithAttributes(
 		attribute.String("ingress", req.ingress.String()),
@@ -345,7 +355,7 @@ func (c *ConnectedObjectsQuery) reverseExpandTupleToUserset(
 		return err
 	}
 	defer iter.Stop()
-	atomic.AddUint32(dsQueryCount, 1)
+	atomic.AddUint32(resolutionMetadata.QueryCount, 1)
 
 	subg, subgctx := errgroup.WithContext(ctx)
 	subg.SetLimit(int(c.resolveNodeBreadthLimit))
@@ -378,7 +388,7 @@ func (c *ConnectedObjectsQuery) reverseExpandTupleToUserset(
 				Relation:         targetObjectRel,
 				User:             sourceUserRef,
 				ContextualTuples: req.contextualTuples,
-			}, resultChan, req.ingress, dsQueryCount)
+			}, resultChan, req.ingress, resolutionMetadata)
 		})
 	}
 
@@ -389,7 +399,7 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
 	ctx context.Context,
 	req *reverseExpandRequest,
 	resultChan chan<- *ConnectedObjectsResult,
-	dsQueryCount *uint32,
+	resolutionMetadata *ResolutionMetadata,
 ) error {
 	ctx, span := tracer.Start(ctx, "reverseExpandDirect", trace.WithAttributes(
 		attribute.String("ingress", req.ingress.String()),
@@ -452,7 +462,7 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
 		return err
 	}
 	defer iter.Stop()
-	atomic.AddUint32(dsQueryCount, 1)
+	atomic.AddUint32(resolutionMetadata.QueryCount, 1)
 
 	subg, subgctx := errgroup.WithContext(ctx)
 	subg.SetLimit(int(c.resolveNodeBreadthLimit))
@@ -485,7 +495,7 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
 				Relation:         targetObjectRel,
 				User:             sourceUserRef,
 				ContextualTuples: req.contextualTuples,
-			}, resultChan, req.ingress, dsQueryCount)
+			}, resultChan, req.ingress, resolutionMetadata)
 		})
 	}
 
--- a/pkg/server/commands/connectedobjects/connected_objects.go
+++ b/pkg/server/commands/connectedobjects/connected_objects.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/openfga/openfga/internal/graph"
@@ -161,8 +162,9 @@ func (c *ConnectedObjectsQuery) Execute(
 	ctx context.Context,
 	req *ConnectedObjectsRequest,
 	resultChan chan<- *ConnectedObjectsResult,
+	dsQueryCount *uint32,
 ) error {
-	return c.execute(ctx, req, resultChan, nil)
+	return c.execute(ctx, req, resultChan, nil, dsQueryCount)
 }
 
 func (c *ConnectedObjectsQuery) execute(
@@ -170,6 +172,7 @@ func (c *ConnectedObjectsQuery) execute(
 	req *ConnectedObjectsRequest,
 	resultChan chan<- *ConnectedObjectsResult,
 	currentIngress *graph.RelationshipIngress,
+	dsQueryCount *uint32,
 ) error {
 	ctx, span := tracer.Start(ctx, "connectedObjects.Execute", trace.WithAttributes(
 		attribute.String("target_type", req.ObjectType),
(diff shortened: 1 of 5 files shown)


adriantam:
Will incorporate https://github.com/openfga/openfga/issues/938 in this PR as well

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -373,18 +371,14 @@ func (c *ConnectedObjectsQuery) reverseExpandTupleToUserset(
 				Relation:         targetObjectRel,
 				User:             sourceUserRef,
 				ContextualTuples: req.contextualTuples,
-			}, resultChan, req.ingress)
+			}, resultChan, req.ingress, currentQueryCount+1)
```
miparnisari, line 374: This only adds +1 if we found an object, but it could happen that `storage.ReadStartingWithUser` gave 0 tuples. In that case, we still need to count that read.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -152,6 +152,7 @@ const (
 
 type ConnectedObjectsResult struct {
 	Object       string
+	QueryCount   uint32
```
miparnisari, line 155: It's weird to me that we are adding this count `per object` and not as a total per request

pkg/server/commands/connectedobjects/connected_objects.go
adriantam: Good point - we need to make this as an overall count instead of individual object

pkg/server/commands/connectedobjects/connected_objects.go
adriantam: Updated to use overall count

pkg/server/commands/connectedobjects/connected_objects.go
adriantam: Modified algorithm to have overall count

pkg/server/commands/list_objects.go
```diff
@@ -296,7 +307,7 @@ func trySendObject(object string, objectsFound *uint32, maxResults uint32, resul
 func (q *ListObjectsQuery) Execute(
 	ctx context.Context,
 	req *openfgav1.ListObjectsRequest,
-) (*openfgav1.ListObjectsResponse, error) {
+) (*ListObjectsResponse, error) {
```
jon-whit, line 310: Why don't we leave this as is and move this down into the ConnectedObjects layer? Our pattern has always been to expose the same signature for the RPC in the commands package. For consistency it seems we should avoid changing that pattern.

A `ConnectedObjectsResult` should have metadata associated with the resolution info, and using that we can calculate the overall query count by taking the sum of all of the query counts from each result.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -161,15 +162,17 @@ func (c *ConnectedObjectsQuery) Execute(
 	ctx context.Context,
 	req *ConnectedObjectsRequest,
 	resultChan chan<- *ConnectedObjectsResult,
+	dsQueryCount *uint32,
```
jon-whit, line 165: Instead of passing this around everywhere, why don't we do something similar to what we did in Check and propagate some ResolutionMetadata in each ConnectedObjectsRequest? Then for each object we find we can include the resolution metadata in the ConnectedObjectsResult.

I'd prefer to encapsulate the resolution metadata as part of the request body because it avoids bloating the signature, is more consistent with code elsewhere, and it is more flexible in that it potentially allows us to distribute the ConnectedObjects expansion in a distributed way via dispatching. The way this is written right now doesn't allow for a distributed processing approach down since it assumes a single host is handling the request and makes use of shared memory between recursive calls. If we include this metadata in the request structure then we achieve the same thing but do so in a way that is more flexible.

adriantam: > Instead of passing this around everywhere, why don't we do something similar to what we did in Check and propagate some ResolutionMetadata in each ConnectedObjectsRequest? Then for each object we find we can include the resolution metadata in the ConnectedObjectsResult.
> 
> I'd prefer to encapsulate the resolution metadata as part of the request body because it avoids bloating the signature, is more consistent with code elsewhere, and it is more flexible in that it potentially allows us to distribute the ConnectedObjects expansion in a distributed way via dispatching. The way this is written right now doesn't allow for a distributed processing approach since it assumes a single host is handling the request and makes use of shared memory between recursive calls. If we include this metadata in the request structure then we achieve the same thing but do so in a way that is more flexible.

The problem is that the datastore query call should be per request rather than per returned result.  Otherwise, we may overcount as pointed out by @miparnisari  in https://github.com/openfga/openfga/pull/959#discussion_r1303577444

pkg/server/commands/list_objects.go
adriantam: For Check, we were able to "get away" with ResolveCheck response being different than the RPC signature name because the check code lives inside the server module.  However, list object has both server and commands package code. 

As noted below, we could not use  ConnectedObjectsResult because the resolution metadata needs to per request not per ListObject response.  Otherwise, it may overcount in the case of multiple objects or undercount if it does not have any object in the response.

pkg/server/commands/connectedobjects/connected_objects.go
adriantam: As discussed offline, we should use ResolutionMetadata instead of dsQueryCount to allow future expansion without bloating the signature.

pkg/server/commands/list_objects.go
```diff
@@ -379,7 +390,9 @@ func (q *ListObjectsQuery) ExecuteStreamed(
 		defer cancel()
 	}
 
-	err := q.evaluate(timeoutCtx, req, resultsChan, maxResults)
+	resolutionMetadata := connectedobjects.NewResolutionMetadata()
+
+	err := q.evaluate(timeoutCtx, req, resultsChan, maxResults, resolutionMetadata)
```
miparnisari, line 395: why don't we send the resolution metadata inside the `if !channelOpen` block?

pkg/server/test/connected_objects.go
```diff
@@ -1222,8 +1255,10 @@ func ConnectedObjectsTest(t *testing.T, ds storage.OpenFGADatastore) {
 			timeoutCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 			defer cancel()
 
+			resolutionMetadata := connectedobjects.NewResolutionMetadata()
+
 			go func() {
-				err = connectedObjectsCmd.Execute(timeoutCtx, test.request, resultChan)
+				err = connectedObjectsCmd.Execute(timeoutCtx, test.request, resultChan, resolutionMetadata)
```
miparnisari, line 1261: this change here is odd.. i think it'd be better if you called `resolutionMetadata := connectedobjects.NewResolutionMetadata()` inside of `Execute` and then change the return value of `Execute` to return the resolution metadata

pkg/server/commands/list_objects.go
miparnisari: > Why don't we leave this as is and move this down into the ConnectedObjects layer?

We could do this, but then the logging of the resolution metadata can't happen in the `server.go` file, it must happen within the `ListObjects` code (which i'm okay with, for me, the less code we have in `server.go`, the better)

miparnisari (changes requested):

pkg/server/test/connected_objects.go
adriantam: The problem with creating NewResolutionMetadata inside connected objects query's Execute function instead of passing in as parameter is that connectedObjectsCmd.Execute is used as go func in ListObjectsQuery's [evaluate function](https://github.com/openfga/openfga/blob/e7dc8c6e06946b6d8b98676b5f338add29f8dcf9/pkg/server/commands/list_objects.go#L229).  This means that we will either 

- have to create a separate channel in https://github.com/openfga/openfga/blob/e7dc8c6e06946b6d8b98676b5f338add29f8dcf9/pkg/server/commands/list_objects.go#L228, listen on the channel and sum all the values at the end  OR
- wait for connectedObjectsQuery.Execute to return all the resolution metadata value before [processing connectedObjectsResChan](https://github.com/openfga/openfga/blob/e7dc8c6e06946b6d8b98676b5f338add29f8dcf9/pkg/server/commands/list_objects.go#L254).

Both options are not ideal.

pkg/server/commands/list_objects.go
adriantam: Not only that, we will also need to move the trace inside the ListObjects code as well.  However, as I noted above, the biggest problem is the resolution metadata needs to be on a per request level, not on a per connected objects result level.  Otherwise, it is subject to overcount/undercount.

pkg/server/commands/list_objects.go
adriantam: Added for StreamedListObjects

pkg/server/test/connected_objects.go
miparnisari: Ah.... `evaluate` fires the call but doesn't wait for a response, because StreamedListObjects wouldn't work properly if we did wait.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -446,6 +462,7 @@ func (c *ConnectedObjectsQuery) reverseExpandDirect(
 		return err
 	}
 	defer iter.Stop()
+	atomic.AddUint32(resolutionMetadata.QueryCount, 1)
```
jon-whit, line 465: 😢 this adds a lot of concurrency contention.

pkg/server/commands/list_objects.go
```diff
@@ -148,6 +153,7 @@ func (q *ListObjectsQuery) evaluate(
 	req listObjectsRequest,
 	resultsChan chan<- ListObjectsResult,
 	maxResults uint32,
+	resolutionMetadata *connectedobjects.ResolutionMetadata,
```
jon-whit, line 156: Any reason why we add this as an argument here an not create it within the scope of this function? 

I don't see any usages that dictate this be passed in instead of created within.

Suggest changing the signature to `evaluate(...) (*ResolutionMetadata, error)` instead.

pkg/server/commands/list_objects.go
```diff
@@ -391,26 +400,26 @@ func (q *ListObjectsQuery) ExecuteStreamed(
...
 			if !channelOpen {
 				// Channel closed! No more results.
-				return nil
+				return resolutionMetadata, nil
 			}
 
 			if result.Err != nil {
 				if errors.Is(result.Err, serverErrors.AuthorizationModelResolutionTooComplex) {
-					return result.Err
+					return resolutionMetadata, result.Err
```
jon-whit, line 413: We should avoid returning non-nil here in the event of non-nil error.

pkg/server/server.go
```diff
@@ -362,6 +364,21 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 			User:                 req.User,
 		},
 	)
+	if err == nil {
```
jon-whit, line 367: Can we stay within standard Go idioms here and favor the `non-nil` check instead? That's what we mostly do elsewhere, and changing that pattern here makes it non-standard a tad.

```go
if err != nil {
    return nil, err
}

// otherwise do the other stuff and
return &openfgav1.ListObjectsResponse{...}, nil
```

pkg/server/server.go
```diff
@@ -318,9 +318,11 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 	))
 	defer span.End()
 
+	const methodName = "listObjects"
```
jon-whit, line 321: Here and below, can we follow existing conventions so as to not increase the cognitive impact due to the cardinality of different label values?

Existing labels for `grpc_method` and `method` use `listobjects`, `check`, `streamedlistobjects`, etc.. For example,

Screenshot 2023-08-30 at 2 22 28 PM

pkg/server/server.go
```diff
@@ -407,11 +426,24 @@ func (s *Server) StreamedListObjects(req *openfgav1.StreamedListObjectsRequest,
...
 
 	req.AuthorizationModelId = typesys.GetAuthorizationModelID() // the resolved model id
-	return q.ExecuteStreamed(
+
+	resolutionMetadata, err := q.ExecuteStreamed(
 		typesystem.ContextWithTypesystem(ctx, typesys),
 		req,
 		srv,
 	)
+	if err == nil {
```
jon-whit, line 435: Same comment as in https://github.com/openfga/openfga/pull/959/files#r1310778211

jon-whit:
Something worth pointing out with this work, and the work we did in Check, is that from a developer standpoint it's a bit confusing to see this:
Screenshot 2023-08-30 at 2 31 04 PM

The `datastore_query_count` is 1 but I see 2 overall database queries. `postgres.FindLatestAuthorizationModelID` and `postgres.ReadStartingWithUser`.

Instead of incrementnig a query counter in each Check and ListObjects resolution code path, maybe instead it makes more sense to increment the counter as a value in the storage context and the storage layer increments the counter on every read query. Just a thought..?

e.g.
```go

type QueryResolutionMetadata struct {
    DatastoreQueryCount *uint32
}

func (s *Server) Check(ctx context.Context, req openfgav1.CheckRequest) (...) {
    
    resolutionMetadata := &QueryResolutionMetadata{DatastoreQueryCount: 0}

    ctx = ContextWithDatastoreQueryCounter(ctx, resolutionMetadata.DatastoreQueryCount)

   ...
}
```
then in `pkg/storage/storagewrappers` we'd have a wrapped storage layer that does, for example,

```go
func (d *WrappedOpenFGADatastore) ReadUserTuple(ctx context.Context, ...) ... {
    datastoreQueryCounter := DatastoreQueryCounterFromContext(ctx)
    atomic.AddUint32(datastoreQueryCounter, 1)
}
```

In any case, something we may just consider improving a bit..

miparnisari:
^ I'm not opposed to that suggestion but if we do that, we might as well introduce that for Check as well, which right now has the same pattern as this PR

pkg/server/commands/connectedobjects/connected_objects.go
miparnisari: Why a lot? It's not like trying to send through a channel that might be full 🤔

pkg/server/commands/list_objects.go
miparnisari: `evaluate` fires the query but doesn't wait for the response. if we did, StreamedListObjects wouldn't work

adriantam: Yes, as @miparnisari pointed out, StreamdListObjects will not work as we lose the ability for partial results.  See [discussion](https://github.com/openfga/openfga/pull/959#discussion_r1309437519) above

adriantam:
> datastore_query_count

Agree that we may need to tweak a bit - however, we can do it as separate activity since we may need to consider how we want to handle the cache case.

pkg/server/commands/connectedobjects/connected_objects.go
```diff
@@ -341,6 +355,7 @@ func (c *ConnectedObjectsQuery) reverseExpandTupleToUserset(
...
```
miparnisari, line 358: Shouldn't we increase _before_ we call the database? Right now, if the db call errors out, we won't coun it

miparnisari (approved):

pkg/server/commands/connectedobjects/connected_objects.go
adriantam: Technically, we should increase before calling the database.  However, for now, we don't return query count in case of error.  But nonetheless, I will update the code to have counter before the error out path.

miparnisari (approved):

PR #962, opened 2023-08-23
feat: add conditions to WriteAuthorizationModel

jpadilla:
## Description

- Modify the `WriteAuthorizationModel` RPC implementation to support Conditions
- Validate the conditions referenced by the relations are included in the model.
- Validate the conditions provided in the model: define a well-formed expression and parameters, and compile to a valid CEL AST that can be evaluated.
- Validate that condition key names match condition names

### Database

- Introduce a new `serialized_protobuf` column in the `authorization_model` table
- `WriteAuthorizationModel` now stores the proto serialized authorization model in all rows.
- `ReadAuthorizationModel` now prefers to return a model from a "converged" row.

## References

https://github.com/openfga/api/pull/90

## Review Checklist


pkg/typesystem/typesystem.go
```diff
@@ -1066,11 +1073,33 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
...
 	return nil
 }
 
+// validateConditions validates the conditions provided in the model.
+//  1. Each condition should define a well-formed expression,
+//  2. Its parameters should be well-defined
+//  3. The expression and parameters should be compilable.
+func (t *TypeSystem) validateConditions() error {
+	for _, c := range t.conditions {
+		// TODO(jpadilla): cache the compiled condition in the typesystem?
```
miparnisari, line 1094: You can do this inside of `New`

pkg/typesystem/typesystem.go
```diff
@@ -186,6 +189,7 @@ func New(model *openfgav1.AuthorizationModel) *TypeSystem {
 		schemaVersion:   model.GetSchemaVersion(),
 		typeDefinitions: tds,
 		relations:       relations,
+		conditions:      model.GetConditions(),
```
miparnisari, line 192:
```suggestion
		conditions := make(map[string]*CompiledCondition, len(model.Conditions))
                for _, c := range model.conditions {
                         conditions[c.name] = condition.Compile(c.expression)
                  }
```

pkg/typesystem/typesystem.go
```diff
@@ -1066,11 +1073,33 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
...
+	for _, c := range t.conditions {
+		// TODO(jpadilla): cache the compiled condition in the typesystem?
+		_, err := condition.Compile(c)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
miparnisari, line 1102:
```suggestion
func validateConditions(model AuthorizationModel) error {
	for _, c := range model.conditions {
		_, err := condition.Compile(c)
		if err != nil {
			return err
		}
	}
	return nil
}

```

miparnisari:
Before merging this PR, please update this https://github.com/openfga/openfga/blob/4bc50537bbcb96baf70ab059d2566df433ea4ab9/.github/workflows/pull_request.yaml#L1-L6

so that the checks still run even if the target branch is not `main`

pkg/typesystem/typesystem.go
jpadilla: gotta think about this a bit more. `condition.Compile()` can currently return an error which wouldn't make sense in the context of `typesystem.New()`

jon-whit: It's assumed that by the time you're constructing a TypeSystem using `typesystem.New` that the underlying model has been validated. So if we see an error when compiling a condition when constructing a TypeSystem, then we can panic with the error because a serious invariant has not been met in the `WriteAuthorizationModel` API. 

See `typesystem.NewAndValidate(..)` instead.

pkg/typesystem/typesystem.go
```diff
@@ -1066,11 +1073,33 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
...
 }
 
+// validateConditions validates the conditions provided in the model.
+//  1. Each condition should define a well-formed expression,
+//  2. Its parameters should be well-defined
+//  3. The expression and parameters should be compilable.
+func (t *TypeSystem) validateConditions() error {
+	for _, c := range t.conditions {
+		// TODO(jpadilla): cache the compiled condition in the typesystem?
+		_, err := condition.Compile(c)
```
jon-whit, line 1095: We will need to validate the parameters types in here as well (e.g. #2). We only support a handful of parameter types, and if you provide a value other than one of those you should get an error.

pkg/typesystem/typesystem.go
```diff
@@ -181,11 +184,17 @@ func New(model *openfgav1.AuthorizationModel) *TypeSystem {
 		relations[typeName] = tdRelations
 	}
 
+	uncompiledConditions := make(map[string]*condition.EvaluableCondition, len(model.GetConditions()))
```
miparnisari, line 187: `uncompiledConditions`, `evaluableCondition`... can agree on one name please? i prefer `uncompiled` here because it's clear that there's an extra step that needs to happen before it's usable

pkg/typesystem/typesystem.go
```diff
@@ -1066,11 +1078,35 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
...
+		}
 	}
 
 	return nil
 }
 
+// validateConditions validates the conditions provided in the model.
+//  1. Each condition should define a well-formed expression,
+//  2. Its parameters should be well-defined
+//  3. The expression and parameters should be compilable.
```
miparnisari, line 1096: 1 and 3 seem to say the same thing? or is 1 that the expression must return a boolean?

Also where are we doing 2 in the code?

pkg/typesystem/typesystem.go
```diff
@@ -1066,11 +1078,35 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
 				}
 			}
 		}
+
+		if related.Condition != "" {
+			// Validate the conditions referenced by the relations are included in the model.
+			if _, ok := t.conditions[related.Condition]; !ok {
+				return ErrConditionUndefined
```
miparnisari, line 1085:
```suggestion
			if _, ok := t.conditions[related.Condition]; !ok {
				return fmt.Errorf(ErrConditionUndefined, "on relation %s", relationName)
```

pkg/typesystem/typesystem.go
```diff
@@ -1066,11 +1078,35 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
...
 
+// validateConditions validates the conditions provided in the model.
+//  1. Each condition should define a well-formed expression,
+//  2. Its parameters should be well-defined
+//  3. The expression and parameters should be compilable.
+func (t *TypeSystem) validateConditions() error {
+	for key, c := range t.conditions {
+		if key != c.Name {
+			return fmt.Errorf("condition key '%s' does not match condition name '%s'", key, c.Name)
+		}
```
miparnisari, line 1101: when would this happen? also, this isn't covered with tests and i'm not sure how you would write a test for it. i suggest removal

pkg/server/test/write_authzmodel.go
```diff
@@ -428,6 +433,166 @@ func WriteAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastor
 				fmt.Errorf("type 'user' defines a relation with an empty string for a name"),
 			),
 		},
+
+		// conditions
+		{
+			name: "condition_fails_undefined",
```
miparnisari, line 439: can you add a test that asserts that you can't have more than one condition with the same name?

pkg/condition/types/types_test.go
```diff
@@ -0,0 +1,144 @@
...
+		// 	repr:      "TYPE_NAME_LIST<string>",
+		// },
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			converted, err := test.paramType.ConvertValue(test.input)
+			require.Equal(t, test.output, converted)
+			require.Equal(t, test.celType, test.paramType.CelType())
+			require.Equal(t, test.repr, test.paramType.String())
```
miparnisari, line 115: i don't understand these assertions. You're not using the `converted` output at all

pkg/typesystem/typesystem.go
jpadilla: this protects agains the following:

```
map[string]*openfgav1.Condition{
	"condition1": {
		Name:       "condition2",
		Expression: "param1 == 'ok'",
		Parameters: map[string]*openfgav1.ConditionParamTypeRef{
			"param1": {
				TypeName: openfgav1.ConditionParamTypeRef_TYPE_NAME_STRING,
			},
		},
	},
}
```

pkg/typesystem/typesystem.go
jpadilla: the idea here is that: `uncompiledCondition` is an EvaluableCondition in its empty state (haven't called .Compiled()`) and `compiledCondition` after.

pkg/typesystem/typesystem.go
jpadilla: at this point its mostly in `(*EvaluableCondition).Compile` which already has a comment that i think covers it, will remove this bit.

pkg/server/test/write_authzmodel.go
jpadilla: there's one that asserts the case for `condition key '%s' does not match condition name '%s'`. because of this same guard, you can't really have conditions with the same name, since that'd imply a map with duplicate keys.

jon-whit:
@jpadilla we'll want to provide a mechanism to re-use previously compiled conditions so that we don't recompile the same condition again and again. Otherwise when we evaluate a tuple with a condition the performance will suffer.

pkg/condition/condition.go
```diff
@@ -0,0 +1,129 @@
...
+	"golang.org/x/exp/maps"
+)
+
+var emptyEvaluationResult = EvaluationResult{}
+
+type EvaluationResult struct {
+	ConditionMet bool
+}
+
+type EvaluableCondition struct {
```
jon-whit, line 19: Would you please add a godoc comment here with some of the semantic details of the lazy loading of the condition, and also note that it is not safe for concurrent use at this time.

pkg/typesystem/typesystem.go
```diff
@@ -862,6 +869,10 @@ func NewAndValidate(ctx context.Context, model *openfgav1.AuthorizationModel) (*
 		return nil, err
 	}
 
+	if err := t.validateConditions(); err != nil {
```
jon-whit, line 872: Should we add some tests to `typesystem_test.go` to ensure this new behavior continues to behave as expected, regardless of what happens upstream in WriteAuthorizationModel RPC?

Since many code paths use `typesystem.NewAndValidate` I'd like to make sure we have the coverage directly alongside the code instead of indirectly through some other related code path.

pkg/server/test/write_authzmodel.go
```diff
@@ -17,6 +17,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func conditionedRelation(rel *openfgav1.RelationReference, condition string) *openfgav1.RelationReference {
```
jon-whit, line 20: Mind adding this to the Typesystem package instead? 

See https://github.com/openfga/openfga/blob/main/pkg/typesystem/typesystem.go#L62-L144 for similar helpers.

pkg/server/test/write_authzmodel.go
jon-whit: Another +1 to using maps 😉

pkg/server/test/write_authzmodel.go
```diff
@@ -428,6 +433,253 @@ func WriteAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastor
...
+						Parameters: map[string]*openfgav1.ConditionParamTypeRef{
+							"param1": {
+								TypeName: openfgav1.ConditionParamTypeRef_TYPE_NAME_STRING,
+							},
+						},
+					},
+				},
+			},
+			err: serverErrors.InvalidAuthorizationModelInput(
+				fmt.Errorf("failed to compile condition expression: ERROR: <input>:1:1: Syntax error: mismatched input '<EOF>' expecting {'[', '{', '(', '.', '-', '!', 'true', 'false', 'null', NUM_FLOAT, NUM_INT, NUM_UINT, STRING, BYTES, IDENTIFIER}"),
```
jon-whit, line 515: We should sanitize the output here. This is a lot to consume IMO.

I may suggest we define some concrete errors in the `condition` package and return more succint errors here for the developer to consume.

e.g.

```go
if errors.Is(err, conditions.ErrCompilationFailed) {
    return serverErrors.InvalidAuthorizationModelnput(fmt.Errorf("failed to compile condition expression '%s'", expression))
}
```

If we want to be more specific I may suggest a different error model where we have the succint message and additional error details with the specific underlying cause "Syntax error: mismatched input '<EOF>' ..."

For example, maybe we consider using the standard gRPC error design:
https://grpc.io/docs/guides/error/#richer-error-model
https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto

pkg/server/test/write_authzmodel.go
```diff
@@ -428,6 +433,253 @@ func WriteAuthorizationModelTest(t *testing.T, datastore storage.OpenFGADatastor
...
+						},
+					},
+				},
+			},
+			err: serverErrors.InvalidAuthorizationModelInput(
+				fmt.Errorf("condition key 'condition2' does not match condition name 'condition3'"),
+			),
+		},
+		{
+			name: "condition_fails_missing_parameter",
```
jon-whit, line 648: Please add a second version of this same test where you define `Expression: "param1 == 'ok'"` and parameters `{"param2": ...}`. This way we're also testing the case where parameters is non-nil but the parameter is missing.

pkg/condition/types/types.go
```diff
@@ -0,0 +1,165 @@
...
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_ANY:       "any",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_BOOL:      "bool",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_STRING:    "string",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_INT:       "int",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_UINT:      "uint",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_DOUBLE:    "double",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_DURATION:  "duration",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_TIMESTAMP: "timestamp",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_MAP:       "map",
+	openfgav1.ConditionParamTypeRef_TYPE_NAME_LIST:      "list",
```
jon-whit, line 47: I don't see `IPADDRESS` type here.

pkg/condition/types/types_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/cel-go/cel"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPrimitives(t *testing.T) {
```
jon-whit, line 11: We should also have tests for conversions across types here. CEL interprets certain values across types.

e.g.

## Primitives
* string -> bool (should return error)
* number -> string (should return error)
* string -> float ("10" --> (float/double) 10 )
* float -> uint (10.0 --> uint64(10))
* float -> uint (10.1 --> return error, it cannot be represented by uint)
* negative float -> uint (-10.0 --> return error, it cannot be represented by uint)
* float -> int (10.0 --> int(10))
* string -> int ("10" --> int(10))

## Non-primitives
* invalid duration (e.g. "2sm")
* invalid timestamp (anything non-RFC3339, for example 2023-0914

jon-whit: Related to this bit of code for numeric conversions:
https://github.com/openfga/openfga/pull/962/files#diff-613c5ff6a77df837179a9e734de3e33ec4fa5e219145710921d2266d8738f58dR27-R65

pkg/condition/types/types_test.go
```diff
@@ -0,0 +1,142 @@
...
+			celType:   cel.TimestampType,
+			input:     "1972-01-01T10:00:20.021Z",
+			output:    time.Time(time.Date(1972, time.January, 1, 10, 0, 20, 21000000, time.UTC)),
+			repr:      "timestamp",
+		},
+		{
+			name:      "valid_ipaddress",
+			paramType: IPAddressType,
+			celType:   cel.ObjectType("IPAddress"),
+			input:     "127.0.0.1",
```
jon-whit, line 88: Please add a test for invalid ip

pkg/condition/parameters.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// castContextToTypedParameters takes a map of context and a map of condition parameter
+// types and casts/converts the map into a map of typed values that can be evaluated
+// with strong type guarantees.
+func castContextToTypedParameters(
+	contextMap map[string]any,
+	conditionParameterTypes map[string]*openfgav1.ConditionParamTypeRef,
+) (map[string]any, error) {
+	if len(contextMap) == 0 {
```
jon-whit, line 17: Can we add a test here and below to get coverage on these lines? I realize it's a bit overkill, but may as well make Codecov happy.

pkg/condition/types/ipaddress.go
```diff
@@ -0,0 +1,129 @@
...
+func (ip IPAddress) Value() any {
+	return ip
+}
+
+// ipaddressBinaryBinding implements a cel.BinaryBinding that is used as a receiver overload for
+// comparing an ipaddress value against a network CIDR defined as a string. If the ipaddress is
+// within the CIDR range this binding will return true, otherwise it will return false or an error.
+//
+// See https://pkg.go.dev/github.com/google/cel-go/cel#BinaryBinding
+func ipaddressCELBinaryBinding(lhs, rhs ref.Val) ref.Val {
```
jon-whit, line 112: Mind adding an explicit test for this in `ipaddress_test.go`?

These links should help with it:
https://pkg.go.dev/github.com/google/cel-go@v0.18.0/common/types/ref#Val
https://pkg.go.dev/github.com/google/cel-go@v0.18.0/common/types#String (implements ref.Val for string) - use this to mimic the CIDR string

pkg/server/test/write_authzmodel.go
jpadilla: took a simpler stab at this for now so we can continue moving forward.

> For example, maybe we consider using something more like the standard gRPC error design (or use it directly):
https://grpc.io/docs/guides/error/#richer-error-model
https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto

💯

pkg/condition/types/converters.go
```diff
@@ -0,0 +1,120 @@
...
+	if ok {
+		return v, nil
+	}
+
+	floatValue, ok := value.(float64)
+	bigFloat := big.NewFloat(floatValue)
+	if !ok {
+		stringValue, ok := value.(string)
+		if !ok {
+			return nil, fmt.Errorf("unexpected generic type value '%q', expected '%T'", reflect.TypeOf(value), *new(T))
```
jon-whit, line 30: The usage here isn't necessarily bound to generic types of `list<type>`, `map<type>`, etc..

We may avoid using the term 'generic' for this reason. I may suggest

> unexpected type value '%q', expected '%T'

instead

jon-whit:
Just need the database wired up and then let's 🚢 it! 🥳

assets/migrations/mysql/004_add_authorization_model_pbdata.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+ALTER TABLE authorization_model ADD COLUMN pbdata LONGBLOB;
```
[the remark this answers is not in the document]
jpadilla: done

assets/migrations/mysql/004_add_authorization_model_pbdata.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+ALTER TABLE authorization_model ADD COLUMN serialized_protobuf LONGBLOB;
```
jon-whit, line 2: We're assuming for the scope and purposes of this work that there has been prior migrations to move from a model stored across multiple rows to a single row per authorization model. The database schema changes and data migrations we expect to have been completed prior to releasing this work include:

* Converging all rows for each model's type definitions into a single row with `(store_id, authorization_model_id, serialized_protobuf)`
* Modifying the PK to be `(store_id, authorization_model_id)` instead of `(store_id, authorization_model_id, type)`

The code here largely assumes the "future state" where we store a model as a single row. Other work streams will have to execute this migration process before we release the ABAC work officially.

assets/migrations/mysql/004_add_authorization_model_pbdata.sql
jon-whit, line 1: Here and below, should we rename this suffix to `_serialized_protobuf.sql`?

jpadilla: done

jon-whit (approved):

PR #963, opened 2023-08-23
feat: add service and method label to datastore_bounded_read_delay_ms

adriantam:
## Description
Adding label to datastore_bounded_read_delay_ms histogram
Screenshot 2023-08-23 at 10 22 50 AM

## References
Close https://github.com/openfga/openfga/issues/949

## Review Checklist

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -25,6 +25,7 @@ import (
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
 	"github.com/openfga/openfga/pkg/storage"
 	"github.com/openfga/openfga/pkg/storage/storagewrappers"
+	"github.com/openfga/openfga/pkg/telemetry"
 	"github.com/openfga/openfga/pkg/typesystem"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -317,6 +318,8 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 	))
 	defer span.End()
 
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
+
 	storeID := req.GetStoreId()
 
 	typesys, err := s.resolveTypesystem(ctx, storeID, req.GetAuthorizationModelId())
@@ -366,6 +369,8 @@ func (s *Server) StreamedListObjects(req *openfgav1.StreamedListObjectsRequest, 
 	))
 	defer span.End()
 
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
+
 	storeID := req.GetStoreId()
 
 	typesys, err := s.resolveTypesystem(ctx, storeID, req.GetAuthorizationModelId())
@@ -410,6 +415,8 @@ func (s *Server) Read(ctx context.Context, req *openfgav1.ReadRequest) (*openfga
 	))
 	defer span.End()
 
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
+
 	q := commands.NewReadQuery(s.datastore, s.logger, s.encoder)
 	return q.Execute(ctx, &openfgav1.ReadRequest{
 		StoreId:           req.GetStoreId(),
@@ -422,6 +429,8 @@ func (s *Server) Read(ctx context.Context, req *openfgav1.ReadRequest) (*openfga
 func (s *Server) Write(ctx context.Context, req *openfgav1.WriteRequest) (*openfgav1.WriteResponse, error) {
 	ctx, span := tracer.Start(ctx, "Write")
 	defer span.End()
+
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
 
 	storeID := req.GetStoreId()
 
@@ -449,6 +458,8 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		attribute.KeyValue{Key: "user", Value: attribute.StringValue(tk.GetUser())},
 	))
 	defer span.End()
+
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
 
 	if tk.GetUser() == "" || tk.GetRelation() == "" || tk.GetObject() == "" {
 		return nil, serverErrors.InvalidCheckInput
@@ -530,6 +541,8 @@ func (s *Server) Expand(ctx context.Context, req *openfgav1.ExpandRequest) (*ope
 	))
 	defer span.End()
 
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
+
 	storeID := req.GetStoreId()
 
 	typesys, err := s.resolveTypesystem(ctx, storeID, req.GetAuthorizationModelId())
@@ -551,6 +564,8 @@ func (s *Server) ReadAuthorizationModel(ctx context.Context, req *openfgav1.Read
 	))
 	defer span.End()
 
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
+
 	q := commands.NewReadAuthorizationModelQuery(s.datastore, s.logger)
 	return q.Execute(ctx, req)
 }
@@ -558,6 +573,8 @@ func (s *Server) ReadAuthorizationModel(ctx context.Context, req *openfgav1.Read
 func (s *Server) WriteAuthorizationModel(ctx context.Context, req *openfgav1.WriteAuthorizationModelRequest) (*openfgav1.WriteAuthorizationModelResponse, error) {
 	ctx, span := tracer.Start(ctx, "WriteAuthorizationModel")
 	defer span.End()
+
+	ctx = telemetry.SaveRPCInfoInContext(ctx)
 
 	c := commands.NewWriteAuthorizationModelCommand(s.datastore, s.logger)
 	res, err := c.Execute(ctx, req)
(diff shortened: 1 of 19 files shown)


pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -79,7 +81,20 @@ func (b *boundedConcurrencyTupleReader) waitForLimiter(ctx context.Context) {
...
+	}
+
+	// grpc Method returns in the format "/service/method".
+	methods := strings.Split(method, "/")
+
+	boundedReadDelayMsHistogram.WithLabelValues(
+		openfgav1.OpenFGAService_ServiceDesc.ServiceName,
+		methods[len(methods)-1],
+	).Observe(float64(timeWaiting))
+
```
miparnisari, line 97:
```suggestion

        // grpc.Method returns in the format "/service/method". 
	serviceAndMethod, found := grpc.Method(ctx)
	if !found {
		method = "undefined"
	} else {
          	method = strings.Split(serviceAndMethod, "/")[2]
	}
	boundedReadDelayMsHistogram.WithLabelValues(
		openfgav1.OpenFGAService_ServiceDesc.ServiceName,
		method,
	).Observe(float64(timeWaiting))

```

adriantam: This is why I use len(methods)-1 as it will not panic if the method "/" is not as expected.

adriantam: Also, I believe @jon-whit does not want the grpc.Method to be used here. One thought is to add to context "Method" in the server method (i.e., Check/ListObject) and retrieve it here.  Want to hear your thought on that @miparnisari

miparnisari: > Also, I believe @jon-whit does not want the grpc.Method to be used here.

why?

jon-whit: It leaks transport specific details deeply in the storage layer. This should at minimum be moved down a level as a contract between the app layer and the storage layer, not the transport layer and the storage layer. e.g.

```go
// pkg/server/server.go

package server

func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openfgav1.CheckResponse, error) {
    ctx = ContextWithRPCMethod(fmt.Sprintf("%s/%s", openfgav1.OpenFGAService_ServiceDesc.ServiceName, "Check"

    ...
}
```
then in the storage layer
```go
// pkg/storage/storagewrappers/boundedconcurrency.go

package storagewrappers

func (b *boundedConcurrencyTupleReader) waitForLimiter(ctx context.Context) {
   ...
   method := RPCMethodFromContext(ctx)

   _ = method // use the method in the histogram observation
}
```

adriantam: I am ok with moving it up - but just want to get confirmation that we are good with using context to pass these info.

jon-whit: @miparnisari passing values through context has been met with resistance in the past. I think this is a good use of context value propagation, but what's your take?

adriantam: I took a stab and updated to pass through context.  @jon-whit / @miparnisari - please raise your voice if there are any objections.

jon-whit: We talked about not passing it through the transport layer, and we added more code and are still passing it through the transport layer. If we're going to extract context that is injected at the transport layer, then let's just go back to using `grpc.Method(ctx)`, as that is more standard and less code we have to manage. Otherwise, please move this to the application layer. Middleware interceptors are gRPC transport specific, so what we're doing right now is no different than what we had IMO. We already get the method context injected into the ctx from the grpc middleware, so no need to duplicate that work if we're going to inject that value at the transport middleware layer.

The suggestion above was to move the method context injection into the app layer directly, which is transport agnostic.

adriantam: As discussed offline, we will go back to adding the method via method in the app layer

pkg/telemetry/context.go
```diff
@@ -0,0 +1,50 @@
...
+
+func saveMethodServiceInContext(ctx context.Context, method string, service string) context.Context {
+	methods := strings.Split(method, "/")
+
+	ctx = context.WithValue(ctx, methodContextName, methods[len(methods)-1])
+	return context.WithValue(ctx, serviceContextName, service)
+}
+
+// SaveRPCInfoInContext will save the rpc method and service information in context
+func SaveRPCInfoInContext(ctx context.Context) context.Context {
```
jon-whit, line 26: It's a common idiom in Go to prefix context setters with `ContextWith...` . For example,

`ContextWithTimeout`
`ContextWithDeadline`
`ContextWithAuthClaims` (see https://github.com/openfga/openfga/blob/main/internal/authn/authn.go#L50-L52)

So, to follow status quo in Go world, can we refactor/rename as `ContextWithRPCInfo` convention.

pkg/telemetry/context.go
```diff
@@ -0,0 +1,50 @@
...
+
+const (
+	methodContextName  rpcContextName = "method"
+	serviceContextName rpcContextName = "service"
+)
+
+func saveMethodServiceInContext(ctx context.Context, method string, service string) context.Context {
+	methods := strings.Split(method, "/")
+
+	ctx = context.WithValue(ctx, methodContextName, methods[len(methods)-1])
```
jon-whit, line 21: Instead of enumerating per field just stuff a struct in here. That makes it more extensible in the future as well.

```go
type RPCInfo struct {
    Method string
    Service string
}

func WithRPCInfo(ctx context.Context, info RPCInfo) {
    return context.WithValue(ctx, rpcInfoContextKey, info)
}

func RPCInfoFromContext(ctx context.Context) RPCInfo {
    rpcInfo, ok := ctx.Value(rpcInfoContextKey).(RPCInfo)
    ...

    return rpcInfo
}
```

pkg/telemetry/context.go
jon-whit: Similar, the "getter" idiom is usually of the form `<thing>FromContext`. For example,

`AuthClaimsFromContext`
`PrincipalFromContext`
etc..

So let's refactor/rename to `RPCInfoFromContext(context.Context)` as well for the getter.

pkg/telemetry/context.go
```diff
@@ -0,0 +1,53 @@
...
+	return context.WithValue(ctx, serviceContextName, service)
+}
+
+// ContextWithRPCInfo will save the rpc method and service information in context
+func ContextWithRPCInfo(ctx context.Context) context.Context {
+	method, found := grpc.Method(ctx)
```
jon-whit, line 27: This is dependent upon a middleware upstream. The purpose of these context setters and getters is to eliminiate the codependency on the `grpc` transport.

This code should simply change to
```go
func ContextWithRPCInfo(ctx context.Context, info RPCInfo) {
    return context.WithValue(ctx, rpcInfoContextKey, info)
}
```

then there is no co-dependency on the grpc middleware at the transport layer above this point in the code.

pkg/telemetry/context.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type rpcContextName string
+
+const (
+	methodContextName  rpcContextName = "method"
+	serviceContextName rpcContextName = "service"
+)
+
+func saveMethodServiceInContext(ctx context.Context, method string, service string) context.Context {
```
jon-whit, line 18: No need for this function. This is the purpose of the context setter `ContextWithRPCInfo`

pkg/telemetry/context.go
```diff
@@ -0,0 +1,53 @@
...
+// ContextWithRPCInfo will save the rpc method and service information in context
+func ContextWithRPCInfo(ctx context.Context) context.Context {
+	method, found := grpc.Method(ctx)
+	if !found {
+		method = "unknown"
+	}
+	return saveMethodServiceInContext(ctx, method, openfgav1.OpenFGAService_ServiceDesc.ServiceName)
+}
+
+func getContextInfo(ctx context.Context, key rpcContextName) string {
```
jon-whit, line 34: This extra function is redundant. This is the responsibility of the context getter anyways `RPCInfoFromContext`. The extra function just adds indirection for no value.

pkg/server/server.go
```diff
@@ -317,6 +318,8 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 	))
 	defer span.End()
 
+	ctx = telemetry.ContextWithRPCInfo(ctx)
```
jon-whit, line 321: Here and elsewhere this should be, for example,
```go
ctx = telemetry.ContextWIthRPCInfo(
    ctx,
    telemetry.RPCInfo{
        Service: openfgav1.OpenFGAService_ServiceDesc.ServiceName,
        Method: "ListObjects",
    },
)
```

pkg/telemetry/context.go
adriantam: This method is to provide sane behaviour even if context is not set properly

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -94,7 +95,13 @@ func (b *boundedConcurrencyTupleReader) waitForLimiter(ctx context.Context) {
 
 	end := time.Now()
 	timeWaiting := end.Sub(start).Milliseconds()
-	boundedReadDelayMsHistogram.Observe(float64(timeWaiting))
+
+	rpcInfo := telemetry.RPCInfoFromContext(ctx)
+	boundedReadDelayMsHistogram.WithLabelValues(
+		rpcInfo.Service,
+		rpcInfo.Method,
+	).Observe(float64(timeWaiting))
```
miparnisari, line 103: don't we need to add `if rpcInfo != nil` here?

adriantam: Yes.  It is added

miparnisari (approved):

jon-whit (approved):
LGTM. Thanks for working through the proposed changes 👍

Issue #957, opened 2023-08-23
Address TODO in https://github.com/openfga/openfga/pull/921


miparnisari:
Address all the TODO comments in https://github.com/openfga/openfga/pull/921

PR #968, opened 2023-08-26
style: rename connected objects -> reverse expand

miparnisari:
## Description
Changes:
- Rename `ConnectedObjects -> ReverseExpand`
- Rename `ConnectedObjectsGraph -> RelationshipGraph`
- Rename `RelationshipIngress -> RelationshipEdge` (or sometimes `edge` when used as a variable)
- Updated a few comments
- Added a few TODOs (subject to discussion) which are added to https://github.com/openfga/openfga/issues/957

## References
This is a requisite of https://github.com/openfga/openfga/pull/934

### DIFF
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -19,15 +19,13 @@ const (
 
 var (
 	ErrResolutionDepthExceeded = errors.New("resolution depth exceeded")
-	ErrTargetError             = errors.New("graph: target incorrectly specified")
-	ErrNotImplemented          = errors.New("graph: intersection and exclusion are not yet implemented")
 )
 
-type findIngressOption int
+type findEdgeOption int
 
 const (
-	resolveAllIngresses findIngressOption = iota
-	resolveAnyIngress
+	resolveAllEdges findEdgeOption = iota
+	resolveAnyEdge
 )
 
 // ContextWithResolutionDepth attaches the provided graph resolution depth to the parent context.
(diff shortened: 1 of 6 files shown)


internal/graph/graph.go
```diff
@@ -52,120 +50,101 @@ type ResolutionMetadata struct {
...
-// 2. If `objectType#relation` is a type and `objectType#relation` can be a `source` then add `objectType#relation, direct` to the result.
-// 3. If computed userset, say `define viewer as writer`, then recurse on `document#writer, source`.
-// 4. If tuple-to-userset, say, `define viewer as viewer from parent`. Go to parent and find its types. In this case, `folder`. Go to `folder` and see if it has a `viewer` relation. If so, recurse on `folder#viewer, source`.
-//
-// To look up Ingresses(`document#viewer`, `folder#viewer`), look up the definition of relation `document#viewer` and:
-// 1. If `folder#viewer` is a directly related type then add `folder#viewer, direct` to the result.
-// 2. If computed userset, say `define viewer as writer`, then recurse on `document#writer, folder#viewer`.
-// 3. If tuple-to-userset, say, `define viewer as viewer from parent`. Go to parent and find its related types.
-//  1. If parent's types includes `folder` type, and `folder` contains `viewer` relation then this is exactly a ttu rewrite and....?
-//  2. Otherwise, suppose the types contains `objectType` which has a relation `viewer`, then recurse on `objectType#viewer, folder#viewer`
```
miparnisari, line 159: I feel it's 100% easier to understand by looking at a graph than by reading this chunk

internal/graph/graph.go
```diff
@@ -52,120 +50,101 @@ type ResolutionMetadata struct {
...
+	// NoFurtherEvalCondition indicates an edge condition whereby results found with ReverseExpandQuery are factual and
 	// known to be true and require no further evaluation before a determination of the outcome
 	// can be made.
 	NoFurtherEvalCondition
 )
 
-// RelationshipIngress represents a possible ingress point between some source object reference
-// and a target user reference.
-type RelationshipIngress struct {
+// RelationshipEdge represents a possible relationship some source object reference
```
jon-whit, line 98: > between some source object reference

internal/graph/graph.go
```diff
@@ -52,120 +50,101 @@ type ResolutionMetadata struct {
...
+// and a target user reference. The possibility is realized depending on the tuples and on the edge's type.
+type RelationshipEdge struct {
+	Type RelationshipEdgeType
 
-	// The type of the relationship ingress
-	Type RelationshipIngressType
+	// The edge is directed towards this node, which can be like group:*, or group, or group:member
+	TailUsersetNode *openfgav1.RelationReference
```
jon-whit, line 104: Let's find another name for this. "Tail" is ambiguous here.

This is formally the underlying reference to the target from the incoming "edge". So I may suggest simply calling this the `TargetRelationReference` or `TargetReference`

internal/graph/graph.go
```diff
@@ -52,120 +50,101 @@ type ResolutionMetadata struct {
...
-	// TuplesetRelation defines the tupleset relation reference that relates a source
-	// object reference with a target if the type of the relationship ingress is that
-	// of a TupleToUserset
-	// TODO this can be just a string for the relation (since the type will be the same as Ingress.Type)
+	// If the type is TupleToUsersetEdge, this defines the TTU condition
+	// TODO this can be just a string for the relation (since the type will be the same as TailUsersetNode.Type)
 	TuplesetRelation *openfgav1.RelationReference
 
-	Condition IngressCondition
+	// TODO this is leaking implementation details of ReverseExpand. This can be a boolean saying
```
jon-whit, line 110: Not sure I understand the motivation behind this comment?

Are you suggesting that instead of an `EdgeCondition` being `RequiresFurtherEvalCondition` or `NoFurtherEvalRequired` (which is leaky), that this should just change to boolean type here?

internal/graph/graph.go
```diff
@@ -52,120 +50,101 @@ type ResolutionMetadata struct {
...
 	return strings.ReplaceAll(val, "  ", " ")
 }
 
-// ConnectedObjectGraph represents a graph of relationships and the connectivity between
+// TypesystemGraph represents a graph of relationships and the connectivity between
 // object and relation references within the graph through direct or indirect relationships.
-// The ConnectedObjectGraph should be used to introspect what kind of relationships between
-// object types can exist.
-type ConnectedObjectGraph struct {
+type TypesystemGraph struct {
```
jon-whit, line 128: Can we keep the common theme here and name this `RelationshipGraph`.

We commonly refer to the graph as the "graph of relationships", so it seems appropriate to call this thing the `RelationshipGraph`.

internal/graph/graph.go
```diff
@@ -201,89 +180,88 @@ func (g *ConnectedObjectGraph) findIngresses(
 		return nil, err
 	}
 
-	return g.findIngressesWithTargetRewrite(
+	return g.getRelationshipEdgesWithTargetRewrite(
```
jon-whit, line 183: For consistency, please either use `relationshipEdgesWithTargetRewrite` here (like you did above) or switch what is above to `g.getRelationshipEdges(..)`.

jon-whit:
@miparnisari could you please rebase on `main` when you get a chance?

jon-whit (changes requested):
Left some minor renaming suggestions. Other than that, looks great!

internal/graph/graph.go
miparnisari: Yes

> Not sure I understand the motivation behind this comment?

It would separate concerns. Right now, `graph` knows a bit of how ReverseExpand works. I want `graph` to just tell me whether a node is `and` or `but not`.

And then in ReverseExpand we can do

```go
if (any of the edges we visited so far have this boolean = true) {
  resultStatus = RequiresFurtherEvalStatus
}

internal/graph/graph.go
miparnisari: > "Tail" is ambiguous here.

How? You have an edge that starts at A and points at B. Tail is B.

But i can rename

internal/graph/graph.go
jon-whit: Gotcha. Makes sense.

jon-whit (approved):

PR #973, opened 2023-08-29
test: add tests for datastore.Read

miparnisari:
## Description
Add more tests

### DIFF
--- a/pkg/storage/test/tuples.go
+++ b/pkg/storage/test/tuples.go
@@ -688,7 +688,25 @@ func ReadTest(t *testing.T, datastore storage.OpenFGADatastore) {
 	err := datastore.Write(ctx, storeID, nil, tuples)
 	require.NoError(t, err)
 
-	t.Run("user_and_relation_and_objectID", func(t *testing.T) {
+	t.Run("empty_filter_returns_all_tuples", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/storage/test/tuples.go
```diff
@@ -674,6 +674,105 @@ func ReadStartingWithUserTest(t *testing.T, datastore storage.OpenFGADatastore)
...
+		tuple.NewTupleKey("document:1", "reader", "user:anne"),
+		tuple.NewTupleKey("document:1", "reader", "user:bob"),
+		tuple.NewTupleKey("document:1", "writer", "user:bob"),
+	}
+
+	storeID := ulid.Make().String()
+
+	err := datastore.Write(ctx, storeID, nil, tuples)
+	require.NoError(t, err)
+
```
adriantam, line 690: Do we want to add a test to read all tuples?

jon-whit (dismissed):

pkg/storage/test/tuples.go
jon-whit: Ah yes, good call. Let's please add this test.

miparnisari: TIL /read supports not passing any tuple key 😆  i'll add that test.

adriantam (approved):

PR #990, opened 2023-09-05
chore: use proper tuples type

adriantam:
## Description
Update to use the proper tuples after updating the proto files

## Review Checklist


pkg/tuple/tuple.go
```diff
@@ -25,6 +25,22 @@ var (
 	relationRegex = regexp.MustCompile(`^[^:#@\s]+$`)
 )
 
+func ConvertCheckRequestTupleKeyToTupleKey(tk *openfgav1.CheckRequestTupleKey) *openfgav1.TupleKey {
```
jon-whit, line 28: Instead of converting to `TupleKey` from other `CheckTupleKey` and `ExpandTupleKey` etc.. I recommend that we change our storage interface to take filter inputs and avoid the TupleKey altogether.

e.g.

```go
package storage

Read(..., ReadFilter)
ReadUserTuple(..., ...)
ReadStartingWithuser(..., ReadStartingWithUserFilter)
```

instead of using a TupleKey and interpreting it different ways, it would be nice to have storage endpoint specific filters.

jon-whit: This is something we should follow up on, but not relevant for the scope of this work at this time.

Please create an issue to explore this approach.

jon-whit (approved):

miparnisari (approved):

pkg/tuple/tuple.go
adriantam: See https://github.com/openfga/openfga/issues/991

Issue #974, opened 2023-08-29
ListObjects is creating more goroutines than necessary


miparnisari:
See 
https://github.com/openfga/openfga/blob/1a846a4710fea8d87bb10a1dbee17b7194d3f23b/pkg/server/commands/reverseexpand/reverse_expand.go#L303-L323

In ListObjects, when there is a computed userset involved, we are creating a goroutine for no good reason (because a computed userset doesn't involve reading from the database like the rest of the cases).

The task is to make sure that we only create goroutines when we need to read from the database.

PR #997, opened 2023-09-08
refactor: ListObjects

miparnisari:
## Description
Simplify ListObjects

- Address two TODOs
- Remove duplicate code in `reverseExpandDirect` and `reverseExpandTupleToUserset`. Both methods read tuples and call `execute` on each result. What differs is the filter passed to the read call, and the `relation` value in the call to `execute`

## References
Closes https://github.com/openfga/openfga/issues/957


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -267,39 +258,29 @@ func (c *ReverseExpandQuery) execute(
...
+					ObjectRelation: &openfgav1.ObjectRelation{
+						Object:   sourceUserObj,
+						Relation: innerLoopEdge.TargetReference.GetRelation(),
 					},
-					ContextualTuples: r.contextualTuples,
-				}, resultChan, innerLoopEdge, resolutionMetadata)
-			case graph.TupleToUsersetEdge:
-				return c.reverseExpandTupleToUserset(subgctx, r, resultChan, resolutionMetadata)
+				}
+				return c.execute(subgctx, r, resultChan, intersectionOrExclusionInPreviousEdges, resolutionMetadata)
```
miparnisari, line 283: With this change it should be straightforward to do this https://github.com/openfga/openfga/issues/974

internal/graph/graph.go
```diff
@@ -104,19 +89,16 @@ type RelationshipEdge struct {
...
 
 	// If the type is TupleToUsersetEdge, this defines the TTU condition
-	// TODO this can be just a string for the relation (since the type will be the same as TargetReference.Type)
-	TuplesetRelation *openfgav1.RelationReference
+	TuplesetRelation string
 
-	// TODO this is leaking implementation details of ReverseExpand. This can be a boolean saying
-	// if `TargetReference` is intersection or exclusion.
-	Condition EdgeCondition
+	TargetReferenceIsIntersectionOrExclusion bool
```
jon-whit, line 94:
```suggestion
	TargetRefInvolvesIntersectionOrExclusion bool
```

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -267,39 +258,29 @@ func (c *ReverseExpandQuery) execute(
...
-					StoreID:    storeID,
-					ObjectType: req.ObjectType,
-					Relation:   req.Relation,
-					User: &UserRefObjectRelation{
-						ObjectRelation: &openfgav1.ObjectRelation{
-							Object:   sourceUserObj,
-							Relation: innerLoopEdge.TargetReference.GetRelation(),
-						},
+				// follow the computed_userset edge
+				r.User = &UserRefObjectRelation{
```
jon-whit, line 277: Why modify the incoming request pointer? I would advise we construct a derivative request given that we're rewriting the evaluation and starting a new subeval.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -267,39 +258,29 @@ func (c *ReverseExpandQuery) execute(
...
-				targetObjectRef:  targetObjRef,
-				sourceUserRef:    req.User,
-				contextualTuples: req.ContextualTuples,
 			}
 
 			switch innerLoopEdge.Type {
-			case graph.DirectEdge:
-				return c.reverseExpandDirect(subgctx, r, resultChan, resolutionMetadata)
+			case graph.DirectEdge, graph.TupleToUsersetEdge:
+				return c.readTuplesAndExecute(subgctx, r, resultChan, intersectionOrExclusionInPreviousEdges, resolutionMetadata)
```
jon-whit, line 274: Though this reduces some duplicate code paths, here are some downsides that I'm not particularly in favor of:

1. We lose trace clarity.

   Before this it was clear the particular resolution path we were taking by setting a trace named `reverseExpandDirect` or `reverseExpandTTU`. Now those independent paths are obfuscated.
   
   2. We introduce variations in evaluation similarities between Check and ListObjects related code.
   
       In Check resolution we have specific handlers for each terminal rewrite path we can encounter `checkDirect`, `checkComputedUserset`, `checkTTU`. The code before these changes mimicked that as well with `reverseExpandDirect`, `reverseExpandComputedUserset`, `reverseExpandTTU`. 
       
I would recommend we de-dup the shared code as a common helper, but reintroduce the prior composition for the aforementioned reasons above.

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: 🤔 the derivative request (`r`) was constructed in line 263. I'm not modifying the original request (`req`)

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: > Before this it was clear the particular resolution path we were taking by setting a trace named reverseExpandDirect or reverseExpandTTU. Now those independent paths are obfuscated.

The span `readTuplesAndExecute` contains the attribute `edge` which will tell if you if the edge is a Direct or a TTU

> We introduce variations in evaluation similarities between Check and ListObjects related code.

🤔 there are no similarities? ListObjects is doing BFS of a graph. Check isn't using the graph at all. The only similarities would be in the name of the functions 🤷‍♀️ 

My concern is that if we don't remove the duplicate code, when we come back to change something about the reads (for example), we may forget to change the other function too.

jon-whit:
Closing this for now until we want to revive.

miparnisari:
This PR is still relevant. If there's anything that you'd like me to postpone to a follow-up PR please let me know

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: @jon-whit, in the last commit I kept the original functions 👍

adriantam (approved):

PR #975, opened 2023-08-30
refactor(validation): move request validations to RPC handlers

jon-whit:
## Description
Move RPC request validation into the RPC handlers instead of as a top-level middleware on the standalone server.

This change allows developers who use OpenFGA as a library to get the same validations that the standalone server provides. Since middleware doesn't apply to the RPC handler, we move validation inside of the RPC handler.

## References
- Closes https://github.com/openfga/openfga/issues/965
- Relates to https://github.com/openfga/cli/issues/141 - this PR allows the CLI test command to have proper validations when the OpenFGA server is embedded
- https://github.com/openfga/openfga/issues/871

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #998, opened 2023-09-08
chore: rewire request validation

jon-whit:
## Description
Rewires the request validation behavior so that the validation(s) run early in the middleware chain if you're running OpenFGA as a standalone server, but if you're using OpenFGA as a library then the validation context won't be set and so the validation will run inline when the RPC handlers are invoked by the client application.

## References
This is a follow-up of https://github.com/openfga/openfga/pull/975

## Review Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,9 +9,9 @@
             "type": "go",
             "request": "launch",
             "mode": "auto",
-            "program": "${workspaceFolder}/cmd/openfga/openfga.go",
+            "program": "${workspaceFolder}/cmd/openfga/main.go",
+            "args": ["run"],
             "env": {
-                "OPENFGA_DATASTORE_ENGINE": "postgres"
             }
         }
     ]
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -21,6 +21,7 @@ import (
 	"github.com/cenkalti/backoff/v4"
 	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
 	grpc_ctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
+	grpc_validator "github.com/grpc-ecosystem/go-grpc-middleware/validator"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	grpc_prometheus "github.com/jon-whit/go-grpc-prometheus"
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
@@ -31,6 +32,7 @@ import (
 	"github.com/openfga/openfga/internal/build"
 	"github.com/openfga/openfga/internal/gateway"
 	authnmw "github.com/openfga/openfga/internal/middleware/authn"
+	"github.com/openfga/openfga/internal/middleware/validator"
 	"github.com/openfga/openfga/pkg/logger"
 	httpmiddleware "github.com/openfga/openfga/pkg/middleware/http"
 	"github.com/openfga/openfga/pkg/middleware/logging"
@@ -685,11 +687,15 @@ func (s *ServerContext) Run(ctx context.Context, config *Config) error {
 
 	unaryInterceptors := []grpc.UnaryServerInterceptor{
 		requestid.NewUnaryInterceptor(),
+		grpc_validator.UnaryServerInterceptor(),
+		validator.UnaryServerInterceptor(),
 		grpc_ctxtags.UnaryServerInterceptor(),
 	}
 
 	streamingInterceptors := []grpc.StreamServerInterceptor{
 		requestid.NewStreamingInterceptor(),
+		grpc_validator.StreamServerInterceptor(),
+		validator.StreamServerInterceptor(),
 		grpc_ctxtags.StreamServerInterceptor(),
 	}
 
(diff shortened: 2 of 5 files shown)


cmd/run/run.go
```diff
@@ -685,11 +687,15 @@ func (s *ServerContext) Run(ctx context.Context, config *Config) error {
 
 	unaryInterceptors := []grpc.UnaryServerInterceptor{
 		requestid.NewUnaryInterceptor(),
+		grpc_validator.UnaryServerInterceptor(),
+		validator.UnaryServerInterceptor(),
```
jon-whit, line 691: I debated on just copy/paste the `grpc_validator` code into the `validator` package and combining these middlewares into 1, because of the dependency on order between the two of them. But I ended up choosing this route to avoid code duplication.

If ya'll feel differently let me know and I'll change it.

.vscode/launch.json
```diff
@@ -9,9 +9,9 @@
             "type": "go",
             "request": "launch",
             "mode": "auto",
-            "program": "${workspaceFolder}/cmd/openfga/openfga.go",
+            "program": "${workspaceFolder}/cmd/openfga/main.go",
```
jon-whit, line 12: Included this change here because it's required to run OpenFGA in debug mode in VSCode.

internal/middleware/validator/validator.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type ctxKey string
+
+var (
+	requestIsValidatedCtxKey = ctxKey("request-validated")
+)
+
+func ContextWithRequestIsValidated(ctx context.Context) context.Context {
+	return context.WithValue(ctx, requestIsValidatedCtxKey, true)
```
miparnisari, line 16: what if we use `grpc_ctxtags.Extract(r.ctx).Set(requestIsValidatedCtxKey true)`? then we’d get the `request-validated` [ field in the logs for free](https://github.com/openfga/openfga/blob/84c25cf6015f8a604aa87982c7c79d8ea634ca00/pkg/middleware/logging/logging.go#L57) (and we can add a test for this [here](https://github.com/openfga/openfga/blob/1f9c6370976a56dc0535309f02e3a12c4880a372/tests/check/check_test.go#L201-L213))

cmd/run/run.go
miparnisari: I think i prefer having `grpc_validator.UnaryServerInterceptor(),` embedded so that we hide that complexity from here. 

We can also leave as is and put a comment that `validator` must always come after `grpc_validator`. But this is more flaky

internal/middleware/validator/validator.go
miparnisari: Never mind, this isn't necessary

cmd/run/run.go
jon-whit: I'm not sure how we can embed it, but we can copy the behavior into the `validator` package instead.

jon-whit: Refactored to embed it. See https://github.com/openfga/openfga/pull/998/commits/fb765da5eeaeef2673fc8fa09ab14cefcddbfb6e

internal/middleware/validator/validator.go
```diff
@@ -0,0 +1,64 @@
...
+
+func RequestIsValidatedFromContext(ctx context.Context) bool {
+	validated, ok := ctx.Value(requestIsValidatedCtxKey).(bool)
+	return validated && ok
+}
+
+// UnaryServerInterceptor returns a new unary server interceptor that injects a bool indicating if validation has been run.
+//
+// It is expected that this middleware is run after all request validation middleware. Failing to run this middleware after
+// request validation middleware could lead to downstream issues.
```
miparnisari, line 28:
```suggestion
// UnaryServerInterceptor returns a new unary server interceptor that runs request validations and injects a bool indicating if validation has been run.
```

internal/middleware/validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+			return handler(ContextWithRequestIsValidated(ctx), req)
+		})
+	}
+}
+
+// StreamServerInterceptor returns a new streaming server interceptor that injects a bool indicating if validation has been run.
+//
+// It is expected that this middleware is run after all request validation middleware. Failing to run this middleware after
+// request validation middleware could lead to downstream issues.
+func StreamServerInterceptor() grpc.StreamServerInterceptor {
```
miparnisari, line 41:
```suggestion
// StreamServerInterceptor returns a new streaming server interceptor that runs request validations and injects a bool in the contex indicating if validation has been run.
func StreamServerInterceptor() grpc.StreamServerInterceptor {
```

internal/middleware/validator/validator.go
```diff
@@ -0,0 +1,61 @@
...
+	return context.WithValue(ctx, requestIsValidatedCtxKey, true)
+}
+
+func RequestIsValidatedFromContext(ctx context.Context) bool {
+	validated, ok := ctx.Value(requestIsValidatedCtxKey).(bool)
+	return validated && ok
+}
+
+// UnaryServerInterceptor returns a new unary server interceptor that runs request validations and injects a bool indicating if validation has been run.
+func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
```
miparnisari, line 26:
```suggestion
// UnaryServerInterceptor returns a new unary server interceptor that runs request validations and injects a bool in the context indicating if validation has been run.
func UnaryServerInterceptor() grpc.UnaryServerInterceptor {
```

miparnisari (dismissed):

miparnisari (approved):

PR #999, opened 2023-09-09
feat: support conditions in tuples sent through Write API

miparnisari:
## Description
Support conditions in tuples sent through Write API

- [x] write validation tests and code
- [x] persist `condition` and `condition_context` in `changelog` and `tuple` tables
- [x] support conditions in tuples on Read* APIs

## References
Depends on https://github.com/openfga/api/pull/95 and https://github.com/openfga/openfga/pull/962


internal/validation/validation.go
```diff
@@ -166,6 +170,93 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+	} else {
+		condition, ok := typesys.GetConditions()[tk.Condition.Name]
+		if !ok {
+			return &tuple.InvalidConditionalTupleError{
+				Cause: fmt.Errorf("undefined condition"), TupleKey: tk,
+			}
+		}
+
+		validCondition := false
+		for _, userset := range typeRestrictions {
```
jon-whit, line 236:
```suggestion
		for _, directlyRelatedType := range typeRestrictions {
```

internal/validation/validation.go
```diff
@@ -166,6 +170,93 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+			if userset.Type == userType && userset.Condition == tk.Condition.Name {
+				validCondition = true
+				continue
```
jon-whit, line 239: No need to `continue`, we can break out of the containing loop, because we found at least 1 type restriction that allows the relationship with the condition.

pkg/condition/condition.go
```diff
@@ -75,6 +75,23 @@ func (c *EvaluableCondition) Compile() error {
 	return nil
 }
 
+func (c *EvaluableCondition) CastContextToTypedParameters(context map[string]any) (map[string]any, error) {
```
jon-whit, line 78: Mind adding a godoc comment here to note that this converts the provided context to typed condition parameters and returns an error if any additional context fields are provided that are not defined by the evaluable condition.

internal/validation/validation.go
```diff
@@ -166,6 +170,93 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+	typeRestrictions, err := typesys.GetDirectlyRelatedUserTypes(objectType, tk.Relation)
+	if err != nil {
+		return err
+	}
+
+	if tk.Condition == nil {
+		hasConditionedTypeRestriction := false
+		hasUnconditionedTypeRestriction := false
+
+		for _, userset := range typeRestrictions {
```
jon-whit, line 188:
```suggestion
		for _, directlyRelatedType := range typeRestrictions {
```

internal/validation/validation.go
```diff
@@ -166,6 +170,93 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+			}
+		}
+
+		if !validCondition {
+			return &tuple.InvalidConditionalTupleError{
+				Cause: fmt.Errorf("invalid condition for type restriction"), TupleKey: tk,
+			}
+		}
+
+		_, err := condition.CastContextToTypedParameters(tk.Condition.Context.AsMap())
```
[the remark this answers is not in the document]
jon-whit: 🤔 but can we..? If we see the same condition for the same authorization model id more than once in the lifetime of the server, then can we avoid conversion/casting?

internal/validation/validation.go
```diff
@@ -166,6 +170,93 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+				hasUnconditionedTypeRestriction = true
+				continue
+			}
+
+			if userset.RelationOrWildcard == nil {
+				hasConditionedTypeRestriction = true
+				continue
+			}
+
+			if _, ok := userset.RelationOrWildcard.(*openfgav1.RelationReference_Relation); ok {
```
jon-whit, line 203: `.GetRelation()` for shorthand

`.GetRelation() != ""`

internal/validation/validation.go
```diff
@@ -166,6 +170,93 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+	objectType := tuple.GetType(tk.Object)
+	userType := tuple.GetType(tk.User)
+	userRelation := tuple.GetRelation(tk.User)
+
+	typeRestrictions, err := typesys.GetDirectlyRelatedUserTypes(objectType, tk.Relation)
+	if err != nil {
+		return err
+	}
+
+	if tk.Condition == nil {
```
jon-whit, line 184: I think this whole block of code reads a little easier with this alternative implementation. The logic with the negation on the one variable and the non-negation on the other is hard to follow.

```go
if tk.Condition == nil {
	for _, userset := range typeRestrictions {
		if userset.Condition != "" {
			continue
		}

		if userset.Type != userType {
			continue
		}

		if userset.GetRelationOrWildcard() != nil {
			if userset.GetRelation() != "" && userset.GetRelation() != userRelation {
				continue
			}

			if userset.GetWildcard() != nil && !tuple.IsTypedWildcard(tk.User) {
				continue
			}
		}

		return nil
	}

	return &tuple.InvalidConditionalTupleError{
		Cause: fmt.Errorf("condition is missing"), TupleKey: tk,
	}

}
```

jon-whit: The idea with this code is to quickly skip over type restrictions that do not define a condition and otherwise require the type restriction to match.

If no such type restrictions are found we return the error, otherwise the first match we find we return. It avoids continuing to cycle through directly related types, and I think it reads a little more straightforward.

pkg/server/commands/write_test.go
```diff
@@ -197,12 +209,227 @@ func TestTransactionalWriteFailedError(t *testing.T) {
...
+	mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)
+	mockDatastore.EXPECT().MaxTuplesPerWrite().AnyTimes().Return(10)
+	mockDatastore.EXPECT().
+		ReadAuthorizationModel(gomock.Any(), gomock.Any(), gomock.Any()).
+		AnyTimes().
+		Return(model, nil)
+
+	cmd := NewWriteCommand(mockDatastore, logger)
+
+	tests := []test{
```
jon-whit, line 315: Would you add a test to cover the case where you reference an undefined condition. For example, `condition2`.

If a tuple references a condition that is undefined in that model, then it should return an error.

pkg/storage/mysql/mysql.go
```diff
@@ -187,9 +188,10 @@ func (m *MySQL) ReadUserTuple(ctx context.Context, store string, tupleKey *openf
 	objectType, objectID := tupleUtils.SplitObject(tupleKey.GetObject())
 	userType := tupleUtils.GetUserTypeFromUser(tupleKey.GetUser())
 
+	var conditionContext *[]byte
```
jon-whit, line 191: Any particular reason for `*[]byte` here? The default value for a slice is nil, so that seems to cover the case. I tested out `[]byte` on my end and it works just as good.

jpadilla: ah yeah good point 👍

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -285,14 +325,22 @@ func Write(ctx context.Context, dbInfo *DBInfo, store string, deletes storage.De
...
 		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
 
+		conditionName, conditionContext, err := marshalRelationshipCondition(tk.GetCondition())
+		if err != nil {
+			return err
+		}
+
 		res, err := deleteBuilder.
 			Where(sq.Eq{
 				"store":       store,
```
jon-whit, line 346: This new delete behavior brings some interesting nuances with it. What should be the expectations of the following:

```
Write(document:1#viewer@user:jon)
Delete(document:1#viewer@user:jon, {condition: "mycondition"})

Write(document:1#viewer@user:jon, {condition: "mycondition"})
Delete(document:1#viewer@user:jon)
```

If we still treat a tuple as `(object, relation, user)`, then the tuple keys that are to be deleted should be different than the tuple key that is to be inserted. That is,

```
Write({
    "writes": {
        "tuple_keys": [
            {"object": "document:1", "relation": "viewer", "user": "user:jon", "condition": {...}}
        ]
    },
    "deletes": {
        "tuple_keys": [
            {"object": "document:1", "relation": "viewer", "user": "user:jon"}
    }
})
```
The `condition` is not needed in the delete list.

pkg/storage/test/tuples.go
```diff
@@ -207,13 +207,44 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		require.EqualError(t, err, expectedError.Error())
```
jon-whit, line 207: Could you also add some tests to account for the comment above?

https://github.com/openfga/openfga/pull/999/files#r1366181079

jpadilla: [a8cf564](https://github.com/openfga/openfga/pull/999/commits/a8cf564222df38400e3f6b4bcb6602bf6dd02079)

internal/validation/validation.go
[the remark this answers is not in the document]
jon-whit: Re-posting what we discussed offline here.

We talke about memoizing the decoding of parameters type references. Since relationships will commonly see the same structure of context in the tuple, we can avoid re-decoding parameter type references each time we revisit one. For reference, I'm suggesting we can optimize and avoid this line if we're already decoded the same parameter type reference:

https://github.com/openfga/openfga/pull/999/files#diff-4e95dc26dacdca908d7eb173740544f27a74edf6474f9fbbeaa5461f57089fc7R100

We will target this work in a follow up PR though.

pkg/storage/sqlcommon/encoding.go
```diff
@@ -0,0 +1,21 @@
+package sqlcommon
+
+import (
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+func marshalRelationshipCondition(
+	rel *openfgav1.RelationshipCondition,
+) (string, *[]byte, error) {
```
jon-whit, line 10: Can we return `[]byte` here instead? `var marshaledBytes []byte = nil` can be treated the same as `var marshaledBytes *[]byte = nil`. The extra pointer appears to be redundant.

jpadilla: ah yep good catch forgot to change this one here too

internal/validation/validation.go
```diff
@@ -166,6 +170,76 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
...
+				continue
+			}
+
+			if directlyRelatedType.GetRelationOrWildcard() != nil {
+				if directlyRelatedType.GetRelation() != "" && directlyRelatedType.GetRelation() != userRelation {
+					continue
+				}
+
+				if directlyRelatedType.GetWildcard() != nil && !tuple.IsTypedWildcard(tk.User) {
+					continue
```
miparnisari, line 200: These two `continues` are never being hit by our tests

image

jon-whit:
I'm wondering if we should normalize `context: null` and `context: {}`. WDYT? For example,
```
Write({
    "writes": {
        "tuple_keys": [
            {
                "object": "document:1",
                "relation": "viewer",
                "user": "user:jon",
                "condition": {
                    "name":"condition1"
                }
            },
            {
                "object": "document:2",
                "relation": "viewer", 
                "user": "user:jon",
                "condition": {
                    "name": "condition1",
                    "context": {}
                }
            }
         ]
     }
})
```

`Read(objectType="document")` currently returns
```
{
    "tuples": [
        {
            "key": {
                "user": "user:jon",
                "relation": "viewer",
                "object": "document:2",
                "condition": {
                    "name": "condition1",
                    "context": {}
                }
            },
            "timestamp": "2023-10-24T20:01:55.752449Z"
        },
        {
            "key": {
                "user": "user:jon",
                "relation": "viewer",
                "object": "document:1",
                "condition": {
                    "name": "condition1",
                    "context": null
                }
            },
            "timestamp": "2023-10-24T20:02:14.285924Z"
        }
    ],
    "continuation_token": ""
}
```
but I wonder if we should normalize `null` and return
```
{
    "tuples": [
        {
            "key": {
                "user": "user:jon",
                "relation": "viewer",
                "object": "document:2",
                "condition": {
                    "name": "condition1",
                    "context": {}
                }
            },
            "timestamp": "2023-10-24T20:01:55.752449Z"
        },
        {
            "key": {
                "user": "user:jon",
                "relation": "viewer",
                "object": "document:1",
                "condition": {
                    "name": "condition1",
                    "context": {}
                }
            },
            "timestamp": "2023-10-24T20:02:14.285924Z"
        }
    ],
    "continuation_token": ""
}
```
So `null` context and empty context `{}` are treated identically. If you're evaluating a condition and the context is null, then you have to treat it as an empty context anyways, so I feel the normalization is warranted.

jpadilla:
> I'm wondering if we should normalize `context: null` and `context: {}`. WDYT? For example,
> 
> ```
> Write({
>     "writes": {
>         "tuple_keys": [
>             {
>                 "object": "document:1",
>                 "relation": "viewer",
>                 "user": "user:jon",
>                 "condition": {
>                     "name":"condition1"
>                 }
>             },
>             {
>                 "object": "document:2",
>                 "relation": "viewer", 
>                 "user": "user:jon",
>                 "condition": {
>                     "name": "condition1",
>                     "context": {}
>                 }
>             }
>          ]
>      }
> })
> ```
> 
> `Read(objectType="document")` currently returns
> 
> ```
> {
>     "tuples": [
>         {
>             "key": {
>                 "user": "user:jon",
>                 "relation": "viewer",
>                 "object": "document:2",
>                 "condition": {
>                     "name": "condition1",
>                     "context": {}
>                 }
>             },
>             "timestamp": "2023-10-24T20:01:55.752449Z"
>         },
>         {
>             "key": {
>                 "user": "user:jon",
>                 "relation": "viewer",
>                 "object": "document:1",
>                 "condition": {
>                     "name": "condition1",
>                     "context": null
>                 }
>             },
>             "timestamp": "2023-10-24T20:02:14.285924Z"
>         }
>     ],
>     "continuation_token": ""
> }
> ```
> 
> but I wonder if we should normalize `null` and return
> 
> ```
> {
>     "tuples": [
>         {
>             "key": {
>                 "user": "user:jon",
>                 "relation": "viewer",
>                 "object": "document:2",
>                 "condition": {
>                     "name": "condition1",
>                     "context": {}
>                 }
>             },
>             "timestamp": "2023-10-24T20:01:55.752449Z"
>         },
>         {
>             "key": {
>                 "user": "user:jon",
>                 "relation": "viewer",
>                 "object": "document:1",
>                 "condition": {
>                     "name": "condition1",
>                     "context": {}
>                 }
>             },
>             "timestamp": "2023-10-24T20:02:14.285924Z"
>         }
>     ],
>     "continuation_token": ""
> }
> ```
> 
> So `null` context and empty context `{}` are treated identically. If you're evaluating a condition and the context is null, then you have to treat it as an empty context anyways, so I feel the normalization is warranted.

@jon-whit Yep I think that's fair. Should our storage layer normalize `{}` to `NULL` though?

jon-whit:
> > ...
> Yep I think that's fair. Should our storage layer normalize `{}` to `NULL` though?

@jpadilla NULL is more efficient to store than `{}`, so if we care more about efficiency then I would say yes. If we care more about simplicity then I would say that we just write the context through directly as provided (if provided and it is empty). If context is omitted (e.g. nil) then we can choose to write `NULL` to save on efficiency. I feel comfortable with either one, but it may be safer to avoid working will NULL types on the off chance that we mishandle them. Working with empty context is safer than a potentially nillable value.

jon-whit (approved):

PR #1015, opened 2023-09-21
chore: add docker compose override file for local development

### LINKED ISSUE #896: Add a Docker-compose.yaml for development purposes
The official docker-compose has

https://github.com/openfga/openfga/blob/9e69c9ebacc7de16b4068cc096d6dafab55796e4/docker-compose.yaml#L32-L38

If I want to test a change locally, I need to remember to change `image: ...` to `build: .` and then run `docker-compose up --build openfga`.

sanketrai1:
Closes #896

## Description
Adds a `docker-compose.override` file that can be used in conjunction with the existing `docker-compose.yaml` file for applying local development specific build instructions and configuration for services.

## References
https://docs.docker.com/compose/multiple-compose-files/merge/

## Review Checklist


sanketrai1:
Hi @miparnisari, could you please check if these changes could help in overcoming the issue mentioned in the corresponding issue? Over time more local development-specific changes can be added to the compose override file.

rhamzeh (commented):
@sanketrai1 May I ask if you can add a third option to the "Build from Source" documentation in the README (we can call it: build using docker-compose) which documents how to build it for people other than us who may want to follow?

e.g. `docker compose up -f ... -f ...`

jon-whit:
> @sanketrai1 May I ask if you can add a third option to the "Build from Source" documentation in the README (we can call it: build using docker-compose) which documents how to build it for people other than us who may want to follow?
> 
> e.g. `docker compose up -f ... -f ...`

From my understanding, and it appears to work this way now as well, Docker looks for `docker-compose.yml` and `docker-compose.override.yml` by default. So nothing changes with the existing documentation around `docker compose up`.

sanketrai1:
@rhamzeh, @jon-whit is correct. If `docker-compose.override.yml` is present, then `docker-compose.yml` and `docker-compose.override.yml` are merged automatically when running `docker compose up`. If a compose override file with any other name than `docker-compose.override.yml` or `compose.override.yml` needs to be merged with the default compose file, only then multiple `-f` arguments syntax is needed.

jon-whit:
@sanketrai1 could you please rebase on `main`?

rhamzeh (approved):

jpadilla:
I think the biggest difference when using an override file is that there's no way to skip the overriding. Why not just modify the current docker-compose.yaml file to not use the latest image? Is the intent of the docker-compose setup to get started quick with the latest stable version or with the current version in the filesystem?

miparnisari:
@rhamzeh our current instructions in the README regarding `docker-compose.yaml` don't really work... see https://github.com/openfga/openfga/issues/1047. Based on the fact that we've had that instruction for a long time now, and we've only noticed now, I think it's fair to assume that nobody is actually using this for setup? 🤷‍♀️

miparnisari:
@sanketrai1 thanks for submitting this PR! But it doesn't seem to work for me? I removed my OpenFGA containers to start with a clean slate, and I ran `docker-compose up`, and it is using `openfga:openfga/latest`...

miparnisari (changes requested):
Does this work? See https://github.com/docker/docs/issues/17732

This works for me

```
services:
  migrate:
    image: !reset null
    build: .

  openfga:
    image: !reset null
    build: .
```

But it needs two more changes:

1. Override file must be `.yml`, not `.yaml`
2. I need to run `docker-compose up --build` to trigger a fresh build every time

rhamzeh:
> `docker-compose up --build` to trigger a fresh build every time (otherwise, first `docker-compose up` triggers a build but subsequent ones do not

This is how docker compose works unfortunately, there is nothing we can do about that. By default it uses the cached build unless you trigger a rebuild.

rhamzeh (approved):

docker-compose.override.yaml
```diff
@@ -0,0 +1,6 @@
+services:
+  migrate:
+    build: .
+
+  openfga:
+    build: .
```
miparnisari, line 6:
```suggestion
services:
  migrate:
    image: !reset null
    build: .

  openfga:
    image: !reset null
    build: .
```

miparnisari (changes requested):

miparnisari (approved):

jon-whit (approved):

Issue #944, opened 2023-08-15
Let cache be agnostic of the order of contextual tuples


adriantam:
I suggested we take a checksum of a the contextual tuples to produce a cache key that is stable independent of order.

If we're going to save that work for a subsequent PR, would you please add an issue to track that work and cross reference it here? We don't need to address it in this PR, but we should follow up promptly on that work.

Same contextual tuples but different order should result in a cache hit.

_Originally posted by @jon-whit in https://github.com/openfga/openfga/pull/891#discussion_r1294831411_

PR #1018, opened 2023-09-22
feat: evaluate conditions in ListObjects and Check

jpadilla:
## Description

- Evaluation conditions in ListObjects
- Evaluate conditions in Check
- Track missing condition parameters. This will allow exposing partial evaluation metadata in the future.
- Address #944 and give us a stable hash of Check requests so that we can generate stable cache keys.

  > Previous implementation of `checkRequestCacheKey` had two major flaws:
  > 
  > 1. `req.GetTupleKey().String()` is not guaranteed to produce a stable formatted string. If you follow the call chain, it utlimately calls `prototext.MarshalOptions{Multiline: false}.Format(m)`. According to the godoc for `.Format(..)`
  > 
  > > Format formats the message as a string.
  > > This method is only intended for human consumption and ignores errors.
  > > Do not depend on the output being stable. It may change over time across
  > > different versions of the program.
  > 
  > Depending on this for a cache key would not be recommended as it is not stable.
  > 
  > 2. Contextual tuples would cause cache misses if the same contextual tuples were present but in different order. Taking a deterministic Hash64 of the context and contextual tuples produces a stable cache key regardless of order. Previously the following would lead to a cache miss.
  > 
  > ```
  > Check(
  >   document:1#viewer@user:jon,
  >   contextual_tuples=["group:1#member@user:jon", "group:2#member@user:jon"]
  > ) --> {allowed: true}, cache the result
  > 
  > Check(
  >   document:1#viewer@user:jon,
  >   contextual_tuples=["group:2#member@user:jon", "group:1#member@user:jon"]
  > ) --> cache miss (because of contextual tuple order) --> {allowed: true}, cache the result under a different cache key, causing greater cache contention
  > ```
  > 
  > You can actually verify the latter by running the added tests. The code before wouldn't produce a stable cache key given two different orders (but same contents) of contextual tuples.

## Review Checklist


internal/graph/check.go
```diff
@@ -522,6 +533,33 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			err = validation.ValidateTuple(typesys, tk)
 
 			if t != nil && err == nil {
+				typeRestrictions, err := typesys.GetDirectlyRelatedUserTypes(objectType, relation)
+				if err != nil {
+					return nil, err
+				}
+
+				var condition *condition.EvaluableCondition
+				for _, typeRestriction := range typeRestrictions {
```
jon-whit, line 542: We will need to evaluate a condition anywhere we read a tuple from the storage layer, not just on a direct relationship evaluations. Here are a few good examples,

```
type user

type group
  relations
    define member: [user]

type document
  relations
    define viewer: [group#member with condition1]

condition condition1(x int)
   x < 100

- group:eng#member@user:jon

Check(document:1#viewer@user:jon) should return {allowed: false} because the group membership relationship is conditioned on x < 100, and that condition is not met so the userset (not direct) relationship should be discredited. 
```

jpadilla: In https://github.com/openfga/openfga/pull/999/commits/088389df22eda7819e7d8634568201fd9bcfe8ca I moved validation logic used in Write into ValidateTuple which would be reused wherever we're filtering out invalid tuples. Haven't yet tried those changes against these.

internal/graph/check.go
miparnisari, line 1: In this file, look for every call to `Dispatch` and pass `req.GetContext` in each one. See https://github.com/openfga/openfga/pull/1059 to see why

pkg/server/test/list_objects.go
```diff
@@ -48,7 +57,7 @@ type listObjectsTestCase struct {
 	useCheckCache          bool
 }
 
-func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore) {
+func TestListObjects(t *testing.T, ds storage.OpenFGADatastore) {
```
miparnisari, line 60: Why change the test name? The assertions haven't changed

miparnisari (changes requested):

aaguiarz:
When using dictionaries as parameters, if we reference an non existing key, we return an internal server error:

```
model: |
  model
    schema 1.1

  type user

  type document
    relations
      define can_access: [user with condition1] 

  condition condition1(document_attributes: map<string>){
    document_attributes["status"] == "draft" 
  }

tuples:
  - user: "user:bob"
    relation: "can_access"
    object: "document:1"

tests:
  - check:
    - user: "user:bob"
      object: "document:1"
      context: 
        document_attributes :
          status1: "draft"
      assertions:
        can_access: true

```

It would be great if we can return an error saying that the "status" key is not present.

internal/keys/hasher.go
jon-whit, line 1: The changes here address https://github.com/openfga/openfga/issues/944 and give us a stable hash of Check requests so that we can generate stable cache keys.

If you run the tests that were added in `internal/graph/cached_resolver_test.go` for contextual tuples, they actually fail on `main` today. I thought that `gob` produced a stable ordering on a slice of tuple keys, but it doesn't. This fixes that, and it provides a mechanism for caching context for ABAC as well.

internal/graph/check.go
jon-whit: They are accounted for. There are 3 places we dispatch with a rewritten ResolveCheckRequest. Each of the 3 have the context propagated. If you see otherwise, please let me know.

pkg/server/test/list_objects.go
jon-whit: The only test cases that existed herein before were ones specific to result limit checking. That is no longer the case. The test is more generic to generally testing ListObjects, and so that was the motive behind the name change.

internal/graph/check.go
```diff
@@ -519,9 +528,27 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			}
 
 			// filter out invalid tuples yielded by the database query
-			err = validation.ValidateTuple(typesys, tk)
+			tupleKey := t.GetKey()
+			err = validation.ValidateTuple(typesys, tupleKey)
 
 			if t != nil && err == nil {
+				tupleCondition := tupleKey.GetCondition()
```
jon-whit, line 535: note to self (@jon-whit ) - we can consolidate this since we use it in a couple of places

pkg/condition/condition.go
```diff
@@ -110,14 +133,6 @@ func (c *EvaluableCondition) CastContextToTypedParameters(contextMap map[string]
 		converted[key] = convertedParam
 	}
 
-	// validate against extraneous parameters
-	for key := range contextMap {
```
jon-whit, line 114: Dropped this here since it is strictly used for validations when applied in the `Write` API. When casting context parameters to typed params in query evaluations we don't want to apply this behavior.

pkg/server/test/list_objects.go
```diff
@@ -31,15 +32,23 @@ func (x *mockStreamServer) Send(m *openfgav1.StreamedListObjectsResponse) error
 	return nil
 }
 
+func mustNewStruct(t *testing.T, v map[string]interface{}) *structpb.Struct {
```
jon-whit, line 35: note self (@jon-whit ) - this should move up more generally to a test helper. It may be used in multiple places.

internal/graph/cached_resolver_test.go
```diff
@@ -515,3 +513,214 @@ func TestCachedCheckDatastoreQueryCount(t *testing.T) {
...
+
+	tuples1 := []*openfgav1.TupleKey{
+		tuple.NewTupleKey("document:1", "viewer", "user:jon"),
+		tuple.NewTupleKey("document:2", "viewer", "user:jon"),
+	}
+
+	tuples2 := []*openfgav1.TupleKey{
+		tuple.NewTupleKey("document:2", "viewer", "user:jon"),
+		tuple.NewTupleKey("document:1", "viewer", "user:jon"),
+	}
```
miparnisari, line 571: Is this `tuples1` but with the order inverted? So, test that the order of the ctx tuples don't impact the cache key?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -503,13 +522,31 @@ func (c *ReverseExpandQuery) reverseExpandDirect(
...
 
-		foundObject := tk.GetObject()
+		tupleCondition := tk.GetCondition()
+		conditionName := tupleCondition.GetName()
+		if conditionName != "" {
+			evaluableCondition := c.typesystem.GetCondition(conditionName)
+			conditionEvalResult, err := evaluableCondition.Evaluate(
+				req.context.AsMap(),
+				tupleCondition.GetContext().AsMap(),
+			)
```
miparnisari, line 532: Can we use `TupleConditionMet` here, like we are doing inside `reverseExpandTupleToUserset`

miparnisari (changes requested):

pkg/server/test/list_objects.go
miparnisari: I would prefer that we keep all the correctness tests in `consolidated_1_1_tests.yaml`

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -392,13 +400,23 @@ func (c *ReverseExpandQuery) reverseExpandTupleToUserset(
 
 		tk := t.GetKey()
 
-		foundObject := tk.GetObject()
+		conditionMet, err := eval.TupleConditionMet(tk, c.typesystem, req.context.AsMap())
+		if err != nil {
+			return fmt.Errorf("failed to evaluate relationship condition: %v", err)
```
jpadilla, line 405: there's some duplicate error "wrapping" going on here and elsewhere that `eval.TupleConditionMet()` is being used.

```
failed to evaluate relationship condition: failed to evaluate relationship condition: failed to convert context parameter 'param1': for string: unexpected type value '\"bool\"', expected 'string'
```

pkg/typesystem/typesystem.go
```diff
@@ -861,6 +865,10 @@ func NewAndValidate(ctx context.Context, model *openfgav1.AuthorizationModel) (*
 		return nil, err
 	}
 
+	if err := t.validateConditions(); err != nil {
```
jpadilla, line 868: this can be removed, likely bad artifact from resolving a merge.

internal/graph/cached_resolver.go
```diff
@@ -170,32 +170,43 @@ func (c *CachedCheckResolver) ResolveCheck(
...
-// cache key. If the contextual tuples are different order, it is possible that a different
-// cache key will be produced. This will result in duplicate entries.
+// used for Check resolution cache key lookups in a stable way.
+//
+// The same tuple provided with the same contextual tuples and context should produce the same
+// cache key. Contextual tuple order and context parameter order is ignored, only the contents
+// are compared.
 func checkRequestCacheKey(req *ResolveCheckRequest) (string, error) {
-	var contextualTuplesCacheKey string
+	hasher := keys.NewCacheKeyHasher(xxhash.New())
```
jpadilla, line 179: curious to know more about what issues we had with the previous implementation and how this one solves it.

internal/graph/cached_resolver_test.go
```diff
@@ -515,3 +513,214 @@ func TestCachedCheckDatastoreQueryCount(t *testing.T) {
...
+
+	require.Equal(t, key1, key2)
+	require.NotEqual(t, key1, key3)
+	require.NotEqual(t, key1, key4)
+	require.NotEqual(t, key3, key4)
+}
+
+var checkCacheKey string
+
+func BenchmarkCheckRequestCacheKey(b *testing.B) {
```
jpadilla, line 660: @jon-whit it'd be nice to lift this benchmark up to our main branch just for comparison sake.

internal/graph/cached_resolver.go
jon-whit: Previous implementation had two major flaws:

1. `req.GetTupleKey().String()` is not guaranteed to produce a stable formatted string. If you follow the call chain, it utlimately calls `prototext.MarshalOptions{Multiline: false}.Format(m)`. According to the godoc for `.Format(..)`

> Format formats the message as a string.
> This method is only intended for human consumption and ignores errors.
> Do not depend on the output being stable. It may change over time across
> different versions of the program.

Depending on this for a cache key would not be recommended as it is not stable.

2. Contextual tuples would cache cache misses if the same contextual tuples were present but in different order. Taking a deterministic Hash64 of the context and contextual tuples produces a stable cache key regardless of order. Previously the following would lead to a cache miss.

```
Check(
  document:1#viewer@user:jon,
  contextual_tuples=["group:1#member@user:jon", "group:2#member@user:jon"]
) --> {allowed: true}, cache the result

Check(
  document:1#viewer@user:jon,
  contextual_tuples=["group:2#member@user:jon", "group:1#member@user:jon"]
) --> cache miss (because of contextual tuple order) --> {allowed: true}, cache the result under a different cache key, causing greater cache contention
```

`Check

pkg/server/test/list_objects.go
jon-whit: I would too, but we don't have ABAC DSL models in the scope of this PR 😄

pkg/server/commands/reverseexpand/reverse_expand.go
jon-whit: Definitely. That was just an oversight. All of these bits were meant to use `TupleConditionMet`.

internal/graph/cached_resolver_test.go
jon-whit: `[]*openfgav1.TupleKey` doesn't implement https://pkg.go.dev/sort#Interface, so it's not a valid target. If we were using primitive data types I could do that easily, but to do this I'd have to write more code than we already have here.

I renamed the test though.

internal/graph/cached_resolver_test.go
jon-whit: https://github.com/openfga/openfga/pull/1095

miparnisari (approved):

PR #1023, opened 2023-09-26
feat(storage): track db metrics

matoous:
Exports DB metrics (if metrics are enabled) via prometheus.

Question: shuld this be added behind additional config field under `metrics`?

## Description

Exports DB metrics, such as time spent waiting for free connection, pool stats, query stats, and more if the metrics are enabled. For full list of metrics you can see [prometheus/client_golang](https://github.com/prometheus/client_golang/blob/main/prometheus/collectors/dbstats_collector.go).

## Review Checklist
  - Didn't find any section where this would need to be added.

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -175,6 +175,12 @@
                     "type": "duration",
                     "default": "connections are not closed due to connection's age - database/sql default",
                     "x-env-variable": "OPENFGA_DATASTORE_CONN_MAX_LIFETIME"
+                },
+                "metricsEnabled": {
+                    "description": "enabled or disable sql metrics",
+                    "type": "boolean",
+                    "default": false,
+                    "x-env-variable": "OPENFGA_DATASTORE_METRICS_ENABLED"
                 }
             }
         },
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -176,11 +176,16 @@
                     "default": "connections are not closed due to connection's age - database/sql default",
                     "x-env-variable": "OPENFGA_DATASTORE_CONN_MAX_LIFETIME"
                 },
-                "metricsEnabled": {
-                    "description": "enabled or disable sql metrics",
-                    "type": "boolean",
-                    "default": false,
-                    "x-env-variable": "OPENFGA_DATASTORE_METRICS_ENABLED"
+                "metrics": {
+                    "type": "object",
+                    "properties": {
+                        "enabled": {
+                            "description": "enable/disable sql metrics for the datastore",
+                            "type": "boolean",
+                            "default": false,
+                            "x-env-variable": "OPENFGA_DATASTORE_METRICS_ENABLED"
+                        }
+                    }
                 }
             }
         },
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -136,6 +136,8 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Duration("datastore-conn-max-lifetime", defaultConfig.Datastore.ConnMaxLifetime, "the maximum amount of time a connection to the datastore may be reused")
 
+	flags.Bool("datastore-metrics-enabled", defaultConfig.Datastore.MetricsEnabled, "enable/disable sql metrics")
+
 	flags.Bool("playground-enabled", defaultConfig.Playground.Enabled, "enable/disable the OpenFGA Playground")
 
 	flags.Int("playground-port", defaultConfig.Playground.Port, "the port to serve the local OpenFGA Playground on")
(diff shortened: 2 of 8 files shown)


cmd/run/run.go
```diff
@@ -652,12 +653,12 @@ func (s *ServerContext) Run(ctx context.Context, config *Config) error {
...
 		datastore, err = mysql.New(config.Datastore.URI, dsCfg)
 		if err != nil {
-			return fmt.Errorf("failed to initialize mysql datastore: %w", err)
+			return fmt.Errorf("initialize mysql datastore: %w", err)
 		}
 	case "postgres":
 		datastore, err = postgres.New(config.Datastore.URI, dsCfg)
 		if err != nil {
-			return fmt.Errorf("failed to initialize postgres datastore: %w", err)
+			return fmt.Errorf("initialize postgres datastore: %w", err)
```
jon-whit, line 661: What motivated this change?

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -31,6 +31,8 @@ type Config struct {
 	MaxIdleConns    int
 	ConnMaxIdleTime time.Duration
 	ConnMaxLifetime time.Duration
+
+	Metrics bool
```
jon-whit, line 35:
```suggestion
	ExportMetrics bool
```

jon-whit:
@matoous we should put this behind a `--datastore-metrics-enabled` flag.

env: `OPENFGA_DATASTORE_METRICS_ENABLED`
flag: `datastore-metrics-enabled
config: `datastore.metrics.enabled`

jon-whit:
@matoous any updates on this? Anything we can help with?

matoous:
@jon-whit sorry, no time previous week and now I am on vacation. As of now - this doesn't work. When I tried this locally the `/metrics` endpoint didn't return the database metrics, reason unknown so far. Will investigate further when I am back to work.

miparnisari:
> As of now - this doesn't work. When I tried this locally the /metrics endpoint didn't return the database metrics

I tried this two days ago and took a screenshot, it worked for me 🤔 
image

matoous:
@miparnisari thank you! Maybe I used wrong image by accident. Will take a look 👀

matoous:
@jon-whit / @miparnisari fixed based on feedback and added configuration.

.config-schema.json
```diff
@@ -175,6 +175,12 @@
                     "type": "duration",
                     "default": "connections are not closed due to connection's age - database/sql default",
                     "x-env-variable": "OPENFGA_DATASTORE_CONN_MAX_LIFETIME"
+                },
+                "metricsEnabled": {
```
jon-whit, line 179: One last thing here - could we nest this so that when/if we allow more selective metric collection it can be configured more similarly to the other configs we have. For example, we have:

`metrics.enabled`, `metrics.addr`, etc..

I can forsee a scenario where you only want to export a handful of the available metrics reported by a DB stats collector, and so having this as a nested field would be preferable in that regard. We may not always choose to use the default prometheus DBStatsCollector, and so this very well could have more configuration options.

So we'd have

`OPENFGA_DATASTORE_METRICS_ENABLED` and `datastore.metrics.enabled` for now, but in the future we may also have `datastore.metrics.include=["open_conns", "idle_conns"]` which would only track those metrics.

matoous: @jon-whit done, would you please take a look? 🙏 I will update the documentation website / PR once this gets green light.

miparnisari (approved):

cmd/run/run.go
```diff
@@ -327,7 +329,13 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 		sqlcommon.WithMaxIdleConns(config.Datastore.MaxIdleConns),
 		sqlcommon.WithConnMaxIdleTime(config.Datastore.ConnMaxIdleTime),
 		sqlcommon.WithConnMaxLifetime(config.Datastore.ConnMaxLifetime),
-	)
+	}
+
+	if config.Datastore.Metrics.Enabled {
+		datastoreOptions = append(datastoreOptions, sqlcommon.WithMetrics())
+	}
```
miparnisari, line 336: There is a test called `TestServerMetricsReporting`. Can you turn on this flag in that test and then assert that one of the metrics is emitted?

miparnisari (changes requested):

cmd/run/run.go
miparnisari: Ah... that test is using the `memory` datastore. We'd need to change it to spin up a Postgres datastore. Hmmm

jon-whit (approved):

cmd/run/run.go
miparnisari: Let's do that in a follow-up PR

miparnisari (approved):

PR #1024, opened 2023-09-26
fix: http server tracing

matoous:
## Description

Make sure to propagate the trace headers from HTTP server to gRPC such that the spans stay connected from the origin to OpenFGA.

At the moment, the proxying of HTTP requests to the gRPC server isn't configured with optentelemtry resulting in the OpenFGA spans not being connected to parent (if there's one).

Tested this on our deployment and it's working. I tried to write tests but it proved to be more complicated than I expected so this PR is without tests.

### Before

Screenshot 2023-09-27 at 16 58 31

(the trace isn't attached to parent)

### After

Screenshot 2023-09-27 at 16 59 29

(the trace is attached to parent, in this case the trace originates higher up the stack in application called `permissions`)

## Review Checklist
- [ ] ~I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]~


cmd/run/run.go
```diff
@@ -867,11 +872,19 @@ func (s *ServerContext) Run(ctx context.Context, config *Config) error {
...
 		mux := runtime.NewServeMux(muxOpts...)
 		if err := openfgav1.RegisterOpenFGAServiceHandler(ctx, mux, conn); err != nil {
 			return err
 		}
 
+		handler := http.Handler(mux)
+		if config.Trace.Enabled {
+			// if tracig is enabled wrap the handler with otel to extract the
+			// trace context from incoming requests
+			handler = otelhttp.NewHandler(mux, "proxy")
```
jon-whit, line 885: The `proxy` intermediary is quite confusing. If we could annotate the endpoint handler with the method itself then it would make sense to me (e.g. the grpc method), but the operation 'proxy' doesn't mean anything to me.

I wonder if it makes more sense to just create a lightweight middleware that extracts the parent span from the incoming HTTP request context (if it exists) and inject it if it does. That would solve the same problem, remove the extra nested span with an odd name, and reduce one additional unecessary dependency.

What do you think?

matoous: Yes, good idea. Adjusted the code with tiny wrapper, please check if that is what you meant.

cmd/run/run.go
```diff
@@ -1016,3 +1037,15 @@ func (s *ServerContext) Run(ctx context.Context, config *Config) error {
 
 	return nil
 }
+
+// injectTraceMetadata is lightweight [grpc.UnaryClientInterceptor] that injects trace metadata into outgoing grpc unary call
+// without wrapping the call in new span.
+// This method is used for proxying HTTP calls to the gRPC server implementation without adding unnecessary nesting.
+func injectTraceMetadata(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
```
jon-whit, line 1044: Please move this into an internal telemetry middleware package `internal/middleware/telemetry`. We'll also need to verify the streaming endpoint exhibits the same behavior. We may need to implement the [StreamClientInterceptor](https://pkg.go.dev/google.golang.org/grpc#StreamClientInterceptor)

cmd/run/run.go
jon-whit: Looks better. Thanks!

cmd/run/run.go
matoous: Files restructured and strea middleware added.

miparnisari:
> I tried to write tests but it proved to be more complicated than I expected so this PR is without tests.

I am hesitant to merge without unit tests or integration tests because it will be hard for me to notice if this feature breaks

miparnisari:
Let's merge this one first: https://github.com/openfga/openfga/pull/1199

pkg/telemetry/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+	otelgrpc.Inject(ctx, &md)
+	return streamer(ctx, desc, cc, method, opts...)
+}
+
+// HTTPServerTraceExtractor is a custom middleware that extracts
+// the tracing context and sets the corrent tracing context.
+// It can be used in place of [otelhttp.NewHandler] to avoid creating
+// additional span.
+func HTTPServerTraceExtractor(h http.Handler) http.Handler {
+	propagator := otel.GetTextMapPropagator()
```
jon-whit, line 45: The OpenTelemetry specification leaves the propagator implementation open-ended and supports standards including W3C, B3, and TextMap propagators (others potentially as well).

We probably don't want to be this strictly opinionated and should instead allow for different propagators as allowed by the OpenTelemetry specification.

See https://opentelemetry.io/docs/specs/otel/context/api-propagators/ for more info.

jon-whit: We can follow up with that body of work though. We don't need to block this to account for that.

jon-whit:
Confirmed it works as expected with, for example:

```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"os/signal"
	"syscall"

	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
	"go.opentelemetry.io/otel"
	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
	"go.opentelemetry.io/otel/propagation"
	"go.opentelemetry.io/otel/sdk/resource"
	"go.opentelemetry.io/otel/sdk/trace"
	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
)

func main() {
	res, err := resource.Merge(
		resource.Default(),
		resource.NewSchemaless(semconv.ServiceNameKey.String("some-app")),
	)
	options := []otlptracegrpc.Option{
		otlptracegrpc.WithDialOption(grpc.WithBlock()),
		otlptracegrpc.WithInsecure(),
	}

	var exp trace.SpanExporter
	exp, err = otlptracegrpc.New(context.Background(), options...)
	if err != nil {
		panic(fmt.Sprintf("failed to establish a connection with the otlp exporter: %v", err))
	}

	tp := trace.NewTracerProvider(
		trace.WithSampler(trace.TraceIDRatioBased(1.0)),
		trace.WithResource(res),
		trace.WithSpanProcessor(trace.NewBatchSpanProcessor(exp)),
	)

	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(propagation.TraceContext{}, propagation.Baggage{}))

	otel.SetTracerProvider(tp)
	tracer := tp.Tracer("openfga-client")

	conn, err := grpc.NewClient("localhost:8081",
		grpc.WithStatsHandler(otelgrpc.NewClientHandler()),
		grpc.WithTransportCredentials(insecure.NewCredentials()),
	)
	if err != nil {
		log.Fatal(err)
	}

	var ctx context.Context
	ctx, span := tracer.Start(context.Background(), "OpenFGAClientOperation")
	defer span.End()

	fgaClient := openfgav1.NewOpenFGAServiceClient(conn)
	resp, err := fgaClient.CreateStore(ctx, &openfgav1.CreateStoreRequest{
		Name: "openfga-demo",
	})
	if err != nil {
		log.Fatal(err)
	}

	_ = resp

	done := make(chan os.Signal, 1)
	signal.Notify(done, syscall.SIGINT, syscall.SIGTERM)

	<-done
}

```
Screenshot 2024-04-19 at 4 04 39 PM

You can see the app `some-app` starts the span and propagates it to the OpenFGA server.

miparnisari:
I tested with this demo app: https://github.com/miparnisari/jaegerGoTest/tree/41261fcdf1480b639a86b6ce057eb1d0db3e60fe

I see the inner spans without this PR: 

image

pkg/telemetry/middleware.go
jcchavezs: This is the right call. Imposing a propagation format limits the adoption of openfga.

pkg/telemetry/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+var _ grpc.UnaryClientInterceptor = GRPCUnaryTraceMetadataInjector
+
+// GRPCUnaryTraceMetadataInjector is lightweight [grpc.UnaryClientInterceptor] that injects trace metadata into outgoing grpc unary call
+// without wrapping the call in new span.
```
jcchavezs, line 17: Wonder about the reason for this. Why not just use Otel interceptors?

Zach-Johnson:
Hey @jon-whit I think there is a misunderstanding here and I think this issue should be re-opened. The original author of this issue is concerned about the propagation of trace context from the openfga HTTP server to the openfga gRPC server. The various fixes noted in this thread are only using the gRPC server. If I use, for example, the Go SDK (which hits the http server) and pass a trace context along, that trace context is dropped by openfga in the grpc-gateway. The grpc-gateway code needs to be updated to respect the trace context and pass it along to the grpc server backend.

Issue #1025, opened 2023-09-26
Store authorization models in a single database row


jpadilla:
# Overview

Currently the postgres and mysql storage adapters expect an authorization model to be split into multiple rows, one per each type definition. We can likely converge these into a single row in a series of releases.

## Proposal

In order to introduce this change with zero downtime, we’ll need to make some staggered releases:

### Release 0
- Migration: Add new `serialized_protobuf` column
- WriteAuthorizationModel: create rows for each type definition but including `serialized_protobuf`
- ReadAuthorizationModel: if a row containing a non-null value for `serialized_protobuf`, return this as the model.

### Release 1
- WriteAuthorizationModel: instead of creating a row per each type definition, create a single row where `type='' type_definition=null` and including `serialized_protobuf`
- ReadAuthorizationModel: no changes are required

### Release 2
- Migration: New data migration that creates a new row where `type='' type_definition=null` and includes `serialized_protobuf` for each authorization model where a previously converged row doesn't yet exist.
- Migration: This step deletes all rows except the converged row.
- Migration: Concurrently add new index (`store`, `authorization_model_id`)
- ReadAuthorizationModel: Modify so that read always happens from the single converged row.
- WriteAuthorizationModel: no changes are required

### Release 3
- Migration: Swap primary key index to the new one created in Release 2.
- WriteAuthorizationModel: Modify so that we no longer insert values into the `type` and `type_definition` columns.

### Release 4
- Migration: Drop columns: `type` and `type_definition`

miparnisari:
@jpadilla can we add in the scope of this work 

1. documentation in the storage interface
2. test [here](https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/storage/test/authz_models.go#L45) 

that defines the behavior when someone tries to write a model with 0 types? should the write work or not? and if yes, what should ReadModel return?

IMO: WriteAuthModel should work, and ReadAuthModel too, it will return an empty model. So should ReadAuthModels.

miparnisari:
Also add to the scope of this issue: if we store one model per row, then `ReadAuthorizationModels` (plural) becomes simply `SELECT * FROM authorization_model WHERE store_id=<id>` and then the following code becomes irrelevant:

- https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/storage/mysql/mysql.go#L394-L408
[...]

PR #1028, opened 2023-09-27
feat: persist authz model serialized protobuf

jpadilla:
## Description

- Migration: Add new `serialized_protobuf` column
- WriteAuthorizationModel: create rows for each type definition but including `serialized_protobuf`
- ReadAuthorizationModel: if a row containing a non-null value for `serialized_protobuf`, return this as the model.

## References

https://github.com/openfga/openfga/issues/1025

## Review Checklist

### DIFF
--- a/pkg/storage/mysql/mysql_test.go
+++ b/pkg/storage/mysql/mysql_test.go
@@ -12,9 +12,7 @@ import (
 	"github.com/openfga/openfga/pkg/storage/test"
 	storagefixtures "github.com/openfga/openfga/pkg/testfixtures/storage"
 	"github.com/openfga/openfga/pkg/tuple"
-	"github.com/openfga/openfga/pkg/typesystem"
 	"github.com/stretchr/testify/require"
-	"google.golang.org/protobuf/proto"
 )
 
 func TestMySQLDatastore(t *testing.T) {
@@ -25,6 +23,7 @@ func TestMySQLDatastore(t *testing.T) {
 	require.NoError(t, err)
 	defer ds.Close()
 	test.RunAllTests(t, ds)
+	test.ReadAuthorizationModelUnmarshallErrorTest(t, "mysql", uri, ds)
 }
 
 // TestReadEnsureNoOrder asserts that the read response is not ordered by ulid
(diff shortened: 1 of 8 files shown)


pkg/storage/postgres/postgres_test.go
```diff
@@ -143,3 +143,28 @@ func TestReadPageEnsureOrder(t *testing.T) {
 	require.Equal(t, firstTuple, tuples[1].Key)
 
 }
+
+func TestReadAuthorizationModelUnmarshallError(t *testing.T) {
```
jon-whit, line 147: It'd be nice to see this test move into a common place and the only thing that we do differently between postgres and mysql is the `RunDatastoreTestContainer(t, dbEngine)`.

pkg/storage/mysql/mysql_test.go
```diff
@@ -25,6 +23,7 @@ func TestMySQLDatastore(t *testing.T) {
 	require.NoError(t, err)
 	defer ds.Close()
 	test.RunAllTests(t, ds)
+	test.ReadAuthorizationModelUnmarshallErrorTest(t, "mysql", uri, ds)
```
jon-whit, line 26: Instead of adding this here, would you please put it in the `RunAllTests` function so that it gets run in that group of tests?

pkg/storage/test/authz_models.go
```diff
@@ -161,3 +163,43 @@ func FindLatestAuthorizationModelIDTest(t *testing.T, datastore storage.OpenFGAD
 		require.Equal(t, newModel.Id, latestID)
 	})
 }
+
+func ReadAuthorizationModelUnmarshallErrorTest(t *testing.T, engine string, uri string, ds storage.OpenFGADatastore) {
```
jon-whit, line 167: Up until now the abstraction has been to pass any implementation of the OpenFGADatastore interface to these test methods. The reason for that is so that we can run the storage conformance tests against anything implementing the interface.

I originally said that it would be nice to remove the duplicate code in postgres and mysql, but now I'm wondering if removing the duplicate code is just leaving other non-idealisms behind. It'd be good to avoid implementation specific code in the general storage conformance test suite.

I'm wondering if there is a way we can get the best of both worlds (avoid the implementation specific behavior in the general storage test suite AND not have the duplicate code) 🤔 . Do you have any ideas? If not, should we revert to what you originally had (I'm sorry 😄 )?

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -351,3 +352,95 @@ func Write(ctx context.Context, dbInfo *DBInfo, store string, deletes storage.De
 
 	return nil
 }
+
+func WriteAuthorizationModel(ctx context.Context, dbInfo *DBInfo, store string, model *openfgav1.AuthorizationModel) error {
```
jon-whit, line 356: Seems DBInfo is overkill for this method given we only need the statement builder out of it, and we could probably do a pass over `sqlcommon` to improve things a bit, but for consistency with the other signatures for now we can move forward.

jon-whit (approved):

Issue #1022, opened 2023-09-25
Reduce use of GOB in encoding cache key


adriantam:
Performance test shows that CPU cycles are spent in encoding GOB.

In https://github.com/openfga/openfga/blob/main/internal/graph/cached_resolver.go#L180C1-L186C1, we see that we use GOB to encode cache key.  The tuple keys should be stabled within the same OpenFGA build and should not require GOB for stability.

PR #1029, opened 2023-09-28
feat: reduce use of GOB in encoded cache key

adriantam:
## Description
Only use gob in the case of there are contextual tuples as it is CPU intensive.

## References
Close https://github.com/openfga/openfga/issues/1022

## Review Checklist


internal/graph/cached_resolver.go
```diff
@@ -175,24 +175,25 @@ func (c *CachedCheckResolver) ResolveCheck(
...
-	contextualTuplesCacheKey := c.String()
+		// only use gob if there are contextual tuples as it is CPU intensive
+		if err := gob.NewEncoder(&c).Encode(req.GetContextualTuples()); err != nil {
+			return "", err
+		}
+
+		contextualTuplesCacheKey = c.String()
+	}
 
 	key := fmt.Sprintf("%s/%s/%s/%s",
```
jon-whit, line 193: Seems we should avoid including the trailing `/` if the contextual tuples slice is empty.

`store/model/tuple` vs `store/model/tuple/`

adriantam: Keeping slash allows us to add more fields (maybe ABAC's condition) if needed.

jon-whit: We can add the slash at that time then. For now it's not needed.

jon-whit (approved):

PR #1030, opened 2023-09-29
feat: write authz model in a single row

jpadilla:
## Description

**Note: This should only be released after #1028 has been released**

- WriteAuthorizationModel: instead of creating a row per each type definition, create a single row where `type='' type_definition=null` and including `serialized_protobuf`

## References
https://github.com/openfga/openfga/issues/1025

## Review Checklist


jon-whit:
@jpadilla - We will want to be very clear that there is an upgrade path that should strictly be followed to avoid problems. If developer teams don’t update to `v1.3.3` first and they jump from `v1.3.2` to `v1.3.4` then this will break them. `v1.3.4` should strictly come after the upgrade to `v1.3.3`.

miparnisari:
@jpadilla @jon-whit 

> @jpadilla - We will want to be very clear that there is an upgrade path that should strictly be followed to avoid problems. If developer teams don’t update to `v1.3.3` first and they jump from `v1.3.2` to `v1.3.4` then this will break them. `v1.3.4` should strictly come after the upgrade to `v1.3.3`.

I made a change in this PR https://github.com/openfga/openfga/pull/702/files#diff-d4415e74b94ab18e955f1b14cfd07a462889a57f9dcc3770a4ef549ee48e10e0 to avoid this warning. Basically the server doesn't boot unless you ran the latest migration available. Maybe we should merge that (or a separate PR that only has that change) so that the migration is smooth :)

jon-whit (approved):

PR #1036, opened 2023-10-02
refactor: move server config defaults to pkg/server/config

jpadilla:
## Description

Move server related default config values out from `cmd/run` into `pkg/server/config`, preventing import cycles and hard coded values in various places.

## Review Checklist

### DIFF
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"html/template"
-	"math"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -38,6 +37,7 @@ import (
 	"github.com/openfga/openfga/pkg/middleware/storeid"
 	"github.com/openfga/openfga/pkg/middleware/validator"
 	"github.com/openfga/openfga/pkg/server"
+	serverconfig "github.com/openfga/openfga/pkg/server/config"
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
 	"github.com/openfga/openfga/pkg/server/health"
 	"github.com/openfga/openfga/pkg/storage"
@@ -77,7 +77,7 @@ func NewRunCommand() *cobra.Command {
 		Args:  cobra.NoArgs,
 	}
 
-	defaultConfig := DefaultConfig()
+	defaultConfig := serverconfig.DefaultConfig()
 	flags := cmd.Flags()
 
 	flags.StringSlice("experimentals", defaultConfig.Experimentals, "a list of experimental features to enable")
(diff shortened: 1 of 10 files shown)


pkg/server/config/config.go
```diff
@@ -0,0 +1,326 @@
...
+func (cfg *Config) Verify() error {
+	if cfg.ListObjectsDeadline > cfg.HTTP.UpstreamTimeout {
+		return fmt.Errorf("config 'http.upstreamTimeout' (%s) cannot be lower than 'listObjectsDeadline' config (%s)", cfg.HTTP.UpstreamTimeout, cfg.ListObjectsDeadline)
+	}
+
+	if cfg.Log.Format != "text" && cfg.Log.Format != "json" {
+		return fmt.Errorf("config 'log.format' must be one of ['text', 'json']")
+	}
+
+	if cfg.Log.Level != "none" &&
```
adriantam, line 210: maybe use switch/case?

pkg/server/config/config.go
```diff
@@ -0,0 +1,326 @@
...
+}
+
+// GRPCConfig defines OpenFGA server configurations for grpc server specific settings.
+type GRPCConfig struct {
+	Addr string
+	TLS  *TLSConfig
+}
+
+// HTTPConfig defines OpenFGA server configurations for HTTP server specific settings.
+type HTTPConfig struct {
```
jon-whit, line 57: Thinking out loud here, it's a bit odd to have standalone server configurations in `pkg/server/config`. For example, `GRPCConfig`, `HTTPConfig`, `TLSConfig`, `AuthnConfig`, `TraceConfig`, `ProfilerConfig`, `MetricsConfig`, etc..

These are all configurations that are used to configure features of OpenFGA when it is running as a standalone server as opposed to being imported as a library. All of these are specific to the server as a library:

```
	DefaultChangelogHorizonOffset           = 0
	DefaultResolveNodeLimit                 = 25
	DefaultResolveNodeBreadthLimit          = 100
	DefaultListObjectsDeadline              = 3 * time.Second
	DefaultListObjectsMaxResults            = 1000
	DefaultMaxConcurrentReadsForCheck       = math.MaxUint32
	DefaultMaxConcurrentReadsForListObjects = math.MaxUint32
	DefaultCheckQueryCacheLimit             = 10000
	DefaultCheckQueryCacheTTL               = 10 * time.Second
	DefaultCheckQueryCacheEnable            = false
```

Should we keep standalone server configs internal to OpenFGA since there is no use to import those publicly in another Go app (maybe `internal/server/config`)? What do you think would be the better way to structure these differences?

jpadilla: ok, needed to play out what it looked to us openfga as a lib and setup a server. i think moving this over to `internal` is a practical approach. Defaults are already encapsulated when using the available server constructors, all the composing structs for Config are also nicely tucked away into our builtin run command and such

internal/server/config/config.go
```diff
@@ -0,0 +1,355 @@
...
+		MaxTuplesPerWrite:                         DefaultMaxTuplesPerWrite,
+		MaxTypesPerAuthorizationModel:             DefaultMaxTypesPerAuthorizationModel,
+		MaxAuthorizationModelSizeInBytes:          DefaultMaxAuthorizationModelSizeInBytes,
+		MaxConcurrentReadsForCheck:                DefaultMaxConcurrentReadsForCheck,
+		MaxConcurrentReadsForListObjects:          DefaultMaxConcurrentReadsForListObjects,
+		ChangelogHorizonOffset:                    DefaultChangelogHorizonOffset,
+		ResolveNodeLimit:                          DefaultResolveNodeLimit,
+		ResolveNodeBreadthLimit:                   DefaultResolveNodeBreadthLimit,
+		Experimentals:                             []string{},
+		ListObjectsDeadline:                       DefaultListObjectsDeadline, // there is a 3-second timeout elsewhere
```
miparnisari, line 295: is this comment still relevant? i _think_ it was referring to `HTTP.UpstreamTimeout`, but that's 5 seconds, so 🤷‍♀️

jpadilla: i'll remove it, its not useful as it stands.

pkg/server/test/write_assertions.go
```diff
@@ -89,13 +96,17 @@ func TestWriteAssertions(t *testing.T, datastore storage.OpenFGADatastore) {
 		t.Run(test._name, func(t *testing.T) {
 			model := githubModelReq
 
-			modelID, err := commands.NewWriteAuthorizationModelCommand(datastore, logger, 256*1_024).Execute(ctx, model)
+			writeAuthzModelCmd := commands.NewWriteAuthorizationModelCommand(
```
jon-whit, line 99: Can we follow up and replace this kind of stuff (here and elsewhere) with the options builder pattern? The default should be used if the option isn't explicitly set. You shouldn't have to explicitly set it if it's a default 😄

jon-whit: e.g.

```go
NewWriteAuthorizationModelCommand(datastore, WithLogger(logger), WithAuthorizationModelSizeInBytes(2048)) // if non-defaults are desired, otherwise

NewWriteAuthorizationModelCommand(datastore)
```

cmd/run/run.go
```diff
@@ -508,7 +253,7 @@ func ReadConfig() (*Config, error) {
 	return config, nil
 }
 
-func VerifyConfig(cfg *Config) error {
+func VerifyConfig(cfg *serverconfig.Config) error {
 	if int(cfg.MaxConcurrentReadsForCheck) > cfg.Datastore.MaxOpenConns {
```
jon-whit, line 257: Please drop this and the one below it for the ListObjects max concurrent reads. They are currently emitted every time you start the server because the defaults changed without changing this somehow 🤷 

The default is unbounded/unthrottled (math.MaxUint32) reads per Check and ListObjects requests, which will obviously be higher than the default max open conns. So let's drop this.

jon-whit (dismissed):
Looks good, but left a couple of follow up comments.

pkg/server/test/write_assertions.go
jpadilla: https://github.com/openfga/openfga/issues/1039

jon-whit (approved):

PR #1041, opened 2023-10-04
test: add test for ReadChanges when writes happen concurrently

miparnisari:
## Description
Add a test that ensures that concurrent tuple writes do not impact ReadChanges API (as in, return duplicate results).

### DIFF
--- a/pkg/server/test/read_changes.go
+++ b/pkg/server/test/read_changes.go
@@ -257,19 +257,23 @@ func TestReadChangesReturnsSameContTokenWhenNoChanges(t *testing.T, datastore st
 
 func TestReadChangesAfterConcurrentWritesReturnsUniqueResults(t *testing.T, datastore storage.OpenFGADatastore) {
 	store := testutils.CreateRandomString(10)
-	ctx, backend := writeFourTuplesConcurrently(t, store, datastore)
+
+	tuplesToWriteOne := []*openfgav1.TupleKey{tkMaria, tkCraig}
+	tuplesToWriteTwo := []*openfgav1.TupleKey{tkYamil}
+	totalTuplesToWrite := len(tuplesToWriteOne) + len(tuplesToWriteTwo)
+	ctx, backend := writeTuplesConcurrently(t, store, datastore, tuplesToWriteOne, tuplesToWriteTwo)
 
 	readChangesQuery := commands.NewReadChangesQuery(backend, logger.NewNoopLogger(), encoder.NewBase64Encoder(), 0)
 
 	// without type
 	res1, err := readChangesQuery.Execute(ctx, newReadChangesRequest(store, "", "", storage.DefaultPageSize))
 	require.NoError(t, err)
-	require.Len(t, res1.Changes, 4)
+	require.Len(t, res1.Changes, totalTuplesToWrite)
 
 	// with type
 	res2, err := readChangesQuery.Execute(ctx, newReadChangesRequest(store, "repo", "", storage.DefaultPageSize))
 	require.NoError(t, err)
-	require.Len(t, res2.Changes, 3)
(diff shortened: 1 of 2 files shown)


pkg/server/test/read_changes.go
```diff
@@ -265,3 +283,34 @@ func setup(store string, datastore storage.OpenFGADatastore) (context.Context, s
 
 	return ctx, datastore, nil
 }
+
+// writeFourTuplesConcurrently to expose potential race issues when reading changes
+func writeFourTuplesConcurrently(t *testing.T, store string, datastore storage.OpenFGADatastore) (context.Context, storage.ChangelogBackend) {
```
adriantam, line 288: Instead of harding coding to only write four tuples, perhaps better to accept two groups of tuples keys as parameters.  Each group will be in its own go func?

miparnisari: What is the benefit of doing that?

adriantam: It is better to tie the expectation to where the input is... in line 267/272, I have no idea why I should expect 4 or 3 unless I see the implementation of the function here in line 288.  Not a blocker but something to keep in mind to improve readability

pkg/server/test/read_changes.go
```diff
@@ -265,3 +287,32 @@ func setup(store string, datastore storage.OpenFGADatastore) (context.Context, s
...
+	t.Helper()
+	ctx := context.Background()
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go func() {
+		err := datastore.Write(ctx, store, []*openfgav1.TupleKey{}, tupleGroupOne)
+		if err != nil {
+			t.Logf("failed to write tuples: %s", err)
```
adriantam, line 302: Instead of Logf, should it be Errorf?

pkg/server/test/read_changes.go
```diff
@@ -254,7 +255,28 @@ func TestReadChangesReturnsSameContTokenWhenNoChanges(t *testing.T, datastore st
 	require.Equal(t, res1.ContinuationToken, res2.ContinuationToken)
 }
 
-func setup(store string, datastore storage.OpenFGADatastore) (context.Context, storage.ChangelogBackend, error) {
+func TestReadChangesAfterConcurrentWritesReturnsUniqueResults(t *testing.T, datastore storage.OpenFGADatastore) {
+	store := testutils.CreateRandomString(10)
+
+	tuplesToWriteOne := []*openfgav1.TupleKey{tkMaria, tkCraig}
+	tuplesToWriteTwo := []*openfgav1.TupleKey{tkYamil}
```
adriantam, line 262: Do we want to have tkMariaOrg in order to test the effect of having repo filter?

pkg/server/test/read_changes.go
miparnisari: No, `Errorf` calls `FailNow` and 

> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

[source](https://pkg.go.dev/testing#:~:text=c%20*B)%20FailNow()-,FailNow%20marks%20the%20function%20as%20having%20failed%20and%20stops%20its%20execution,during%20the%20test.%20Calling%20FailNow%20does%20not%20stop%20those%20other%20goroutines.,-func%20(*B)

pkg/server/test/read_changes.go
miparnisari: I removed it because i felt it's less cognitive load, otherwise the assertion has to say `len == 3`

pkg/server/test/read_changes.go
adriantam: If write fails, does it render test invalid and we should fail?

jpadilla (approved):

Issue #954, opened 2023-08-22
ListObjects internally isn't respecting the deadline


miparnisari:
As a follow-up of https://github.com/openfga/openfga/pull/704

We need `ReverseExpandQuery` to stop doing unnecessary work:
- listen to the `ctx.Done()` event so that if the ListObjects call exceeds its deadline (> 3s), then we abort the underlying `ReverseExpandQuery`
- if we already returned `List_objects_max_results`, also abort `ReverseExpandQuery`, i.e. stop looking for more objects

miparnisari:
This was done in https://github.com/openfga/openfga/pull/1064

miparnisari:
Fixed in 1.3.5

PR #1051, opened 2023-10-07
fix: reverse expand query respects context cancellations and deadlines

miparnisari:
## Description
Close https://github.com/openfga/openfga/issues/954

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -177,9 +177,8 @@ func (c *ReverseExpandQuery) Execute(
 	resultChan chan<- *ReverseExpandResult,
 	resolutionMetadata *ResolutionMetadata,
 ) error {
-	err := c.execute(ctx, req, resultChan, nil, resolutionMetadata)
-	close(resultChan)
-	return err
+	defer close(resultChan)
+	return c.execute(ctx, req, resultChan, nil, resolutionMetadata)
 }
 
 func (c *ReverseExpandQuery) execute(
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -177,7 +177,9 @@ func (c *ReverseExpandQuery) Execute(
 	resultChan chan<- *ReverseExpandResult,
 	resolutionMetadata *ResolutionMetadata,
 ) error {
-	return c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	err := c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	close(resultChan)
+	return err
 }
 
 func (c *ReverseExpandQuery) execute(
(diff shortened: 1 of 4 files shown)


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -334,6 +336,9 @@ func (c *ReverseExpandQuery) reverseExpandTupleToUserset(
 
 	combinedTupleReader := storagewrappers.NewCombinedTupleReader(c.datastore, req.contextualTuples)
 
+	if ctx.Err() != nil {
```
jon-whit, line 339: Checking context error between every line of code has got to be an anti-pattern. Instead, the check should be at the iterator level (as in https://github.com/openfga/openfga/pull/1055).

We should be avoiding dispatching more work if the context has been cancelled, but checking it between every function call is extreme.

miparnisari: > Checking context error between every line of code has got to be an anti-pattern.

Ah yes i didn't like it either, i only put it before the expensive database calls. 

> Instead, the check should be at the iterator level (as in #1055).

This sounds more reasonable 😆

jon-whit: I'm going to see if I can resolve the issue that @adriantam reported in #1055 and then we can rebase on that. The other changes are awesome.

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -0,0 +1,119 @@
...
+				  relations
+				    define allowed: [user] as self
+				    define viewer: [user] as self and allowed
+				`),
+	})
+	mockController := gomock.NewController(t)
+	defer mockController.Finish()
+
+	mockDatastore := mocks.NewMockOpenFGADatastore(mockController)
+	mockDatastore.EXPECT().ReadStartingWithUser(gomock.Any(), store, gomock.Any()).
```
jon-whit, line 95: Should be `.MaxTimes(2)` since we may at most see ReadStartingWithUser for `document#allowed` and `document#viewer`. If the context is not cancelled by the other goroutine before `ReadStartingWithUser` is invoked, then there will be a non-zero call count.

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -0,0 +1,119 @@
...
+		Times(0)
+
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), time.Nanosecond)
+	defer cancel()
+	resultChan := make(chan *ReverseExpandResult)
+	reverseExpandQuery := NewReverseExpandQuery(mockDatastore, typeSystem)
+	err := reverseExpandQuery.Execute(timeoutCtx, &ReverseExpandRequest{
+		StoreID:    ulid.Make().String(),
```
jon-whit, line 103: This should be `store`. The mocked called `ReadStartingWithUser(gomock.Any(), store, ...)` won't match at index 1 otherwise.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -177,7 +177,9 @@ func (c *ReverseExpandQuery) Execute(
 	resultChan chan<- *ReverseExpandResult,
 	resolutionMetadata *ResolutionMetadata,
 ) error {
-	return c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	err := c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	close(resultChan)
```
jon-whit, line 181:
```suggestion
	defer close(resultChan)
	return c.execute(...)
```

PR #1055, opened 2023-10-09
add context to iterator interface to cancel iteration if context is cancelled or timed out

jon-whit:
## Description
Add context to storage iterator interface to allow iterators to be promptly interrupted and avoid potentially spawning more work.

This is particularly important to avoid potentially dispatching more Check subproblem resolutions. Consider the following scenario:

```
type user

type group
  relations

type document
  relations
    define viewer: [group#member]
```

- document:1#viewer@group:x#member
- group:x#member@user:jon
- group:x#member@group:1#member
- group:x#member@group:2#member
- ...
- group:x#member@group:N#member

When you call `Check(document:1#viewer@user:jon)`, then two Check subproblems get evaluated. One evaluation path for the direct lookup on `group:x#member@user:jon` and one for the other usersets. These subproblems get evaluated concurrently relative to one another. If the evaluation path for the usersets races faster than the direct lookup, then the iterator that iterates over the usersets and dispatches new Check subproblems could fan-out more work that could otherwise be promptly terminated from the context cancellation that will arise as a result of resolving the direct resolution path (in the other concurrent path of eval).

graphviz

This work will help us to reduce additional goroutine creation and will help increase efficiency in doing so.

## Review Checklist


adriantam:
Running with list object sees the following panic
```
"peer.address": "127.0.0.1:54629", "store_id": "01HCD6RYT1MFS46VAAPMPBXSZW", "authorization_model_id": "01HCD6SCCN1D58YJBDET52Y1PP", "datastore_query_count": 3, "grpc_code": 0}
panic: send on closed channel

goroutine 134 [running]:
github.com/openfga/openfga/pkg/server/commands.(*ListObjectsQuery).evaluate.func1.1()
	/Users/adrian.tam/src/github.com/openfga/openfga/pkg/server/commands/list_objects.go:232 +0x2bc
created by github.com/openfga/openfga/pkg/server/commands.(*ListObjectsQuery).evaluate.func1 in goroutine 133
	/Users/adrian.tam/src/github.com/openfga/openfga/pkg/server/commands/list_objects.go:218 +0x4ec
make: *** [run-postgres] Error 2
```

jon-whit:
> Running with list object sees the following panic
> 
> ```
> "peer.address": "127.0.0.1:54629", "store_id": "01HCD6RYT1MFS46VAAPMPBXSZW", "authorization_model_id": "01HCD6SCCN1D58YJBDET52Y1PP", "datastore_query_count": 3, "grpc_code": 0}
> panic: send on closed channel
> 
> goroutine 134 [running]:
> github.com/openfga/openfga/pkg/server/commands.(*ListObjectsQuery).evaluate.func1.1()
> 	/Users/adrian.tam/src/github.com/openfga/openfga/pkg/server/commands/list_objects.go:232 +0x2bc
> created by github.com/openfga/openfga/pkg/server/commands.(*ListObjectsQuery).evaluate.func1 in goroutine 133
> 	/Users/adrian.tam/src/github.com/openfga/openfga/pkg/server/commands/list_objects.go:218 +0x4ec
> make: *** [run-postgres] Error 2
> ```

Are you able to reproduce this on every run? I cannot produce it on my end yet.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -220,7 +220,11 @@ func (t *SQLTupleIterator) ToArray(opts storage.PaginationOptions) ([]*openfgav1
 	return res, contToken, nil
 }
 
-func (t *SQLTupleIterator) Next() (*openfgav1.Tuple, error) {
+func (t *SQLTupleIterator) Next(ctx context.Context) (*openfgav1.Tuple, error) {
+	if errors.Is(ctx.Err(), context.Canceled) {
```
miparnisari, line 224: also check for `context.DeadlineExceeded`

pkg/storage/memory/memory_test.go
```diff
@@ -28,12 +29,33 @@ func TestStaticTupleIteratorNoRace(t *testing.T) {
...
 	}()
 
 	go func() {
-		_, err := iter.Next()
+		_, err := iter.Next(context.Background())
 		require.NoError(t, err)
 	}()
 }
+
+func TestStaticTupleIteratorContextCanceled(t *testing.T) {
```
miparnisari, line 42: Can you add a test `TestStaticTupleIteratorContextDeadlineExceeded`?

miparnisari:
`refactor` commits will not be mentioned in the changelog since they are assumed to not contain changes that customers will care for.

> This work will help us to reduce additional goroutine creation and will help increase efficiency in doing so.

Can you rename this PR to `perf: add context to iterator interface to cancel iteration if context is cancelled or timedout` #1055`

miparnisari (commented):
Instead of changing the `iterator` interfaces, why don't we update https://github.com/openfga/openfga/blob/afe0b3d4f9edb83c97bd83b19c41b69b703dc4d3/pkg/storage/storagewrappers/context.go#L38-L62 so that they all do

```go
if errors.Is(ctx.Err(), context.Canceled) {
    return nil, ctx.Err()
}
queryCtx := queryContext(ctx)
```

?

jon-whit:
> Instead of changing the `iterator` interfaces, why don't we update
> 
> https://github.com/openfga/openfga/blob/afe0b3d4f9edb83c97bd83b19c41b69b703dc4d3/pkg/storage/storagewrappers/context.go#L38-L62
> 
> so that they all do
> ```go
> if errors.Is(ctx.Err(), context.Canceled) {
>     return nil, ctx.Err()
> }
> queryCtx := queryContext(ctx)
> ```
> 
> ?

Because that's merely a point-in-time check before querying the database. Once the database has been queried the iterator returned from it can still yield results since it is not cancelled, and this can lead to more work being dispatched after the context has been cancelled. If we want to add a point-in-time check on context cancellation before the databse query we can consider doing so, but that would be _in addition to_ this work. Cancelling the ongoing iterator is critical to avoid potentially dispatching more work and continuing to consume results on an iterator created prior to cancellation.

miparnisari:
Reminder: https://github.com/openfga/openfga/pull/1055#issuecomment-1756394082

pkg/storage/memory/memory_test.go
jon-whit: Yep, added 👍

jon-whit:
> Reminder: [#1055 (comment)](https://github.com/openfga/openfga/pull/1055#issuecomment-1756394082)

It's more reliability than performance.

miparnisari:
> This work will help us to reduce additional goroutine creation and will help increase efficiency in doing so.

But in the PR description you wrote "This work will help us to reduce additional goroutine creation and will help increase efficiency in doing so.". How is this not performance?

Issue #1058, opened 2023-10-10
Check with contextual tuple exhibits non-deterministic behavior while running openfga with query cache enabled 


padhikari-godaddy:
**Summary**

Check with contextual tuple exhibits non-deterministic behavior while running openfga with query cache enabled 

- Steps to reproduce
   - Run openfga with check query cache enabled
      `./openfga run  --experimentals check-query-cache --check-query-cache-enabled=true`
   -  Use the model and tuples provided from the Store data section below.
   - Run those three checks in order.

- What did you expect to happen?
    - The `check-1` request should respond with `allowed`: `true`
    - The `check-2` request should respond with `allowed`: `false`
    - The `check-3` request should respond with `allowed`: `true`

- What actually happened?
    - The `check-1` request should respond with `allowed`: `true`
    - The `check-2` request should respond with `allowed`: `false`
    - The `check-3` request should respond with `allowed`: `false`
   If you wait for some time (basically for the cache to expire) and make `check-3` request again, it responds with  `allowed`: `true`.

**OpenFGA version or commit**

Issue exists in the `2b04c251e5f99ccd5bd39eb0485f987deaa9f9ef` or v1.3.3 release.

**Store data**

If applicable, provide information about your authorization model, tuples, and the Check or ListObjects calls you're making. For example:

```yaml
model_file: |
  model
  schema 1.1
type fga_user
type timeslot
  relations
    define user: [fga_user]
type commerce_store
  relations
    define approved_hourly_access: user from approved_timeslot and hourly_employee
    define approved_timeslot: [timeslot]
    define hourly_employee: [fga_user]

tuples:
  - user: fga_user:anne
    relation: hourly_employee
    object: commerce_store:0
  - user: fga_user:anne
    relation: hourly_employee
    object: commerce_store:1
  - user: timeslot:11_12
    relation: approved_timeslot
    object: commerce_store:0
   - user: timeslot:12_13
    relation: approved_timeslot
    object: commerce_store:1
tests: # remove this if not a bug in Check or ListObjects API
  - name: check-1
    check:
      {
        "tuple_key": {
          "user": "fga_user:anne",
          "relation": "approved_hourly_access",
          "object": "commerce_store:0"
        },
        "contextual_tuples": {
          "tuple_keys": [
            {
              "user": "fga_user:anne",
              "relation": "user",
              "object": "timeslot:11_12"
            }
          ]
        }
      }
 - name: check-2
    check:
      {
        "tuple_key": {
[...]

PR #1059, opened 2023-10-11
fix: propagate contextual tuples to dispatched Check subproblems

jon-whit:
## Description
Fixes an issue where contextual tuples were not being propagated to Check sub-evaluations.

## References
https://github.com/openfga/openfga/issues/1058

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -604,6 +604,7 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 							StoreID:              storeID,
 							AuthorizationModelID: req.GetAuthorizationModelID(),
 							TupleKey:             tupleKey,
+							ContextualTuples:     req.GetContextualTuples(),
 							ResolutionMetadata: &ResolutionMetadata{
 								Depth:               req.GetResolutionMetadata().Depth - 1,
 								DatastoreQueryCount: response.GetResolutionMetadata().DatastoreQueryCount,
@@ -661,6 +662,7 @@ func (c *LocalChecker) checkComputedUserset(_ context.Context, req *ResolveCheck
 				StoreID:              req.GetStoreID(),
 				AuthorizationModelID: req.GetAuthorizationModelID(),
 				TupleKey:             rewrittenTupleKey,
+				ContextualTuples:     req.GetContextualTuples(),
 				ResolutionMetadata: &ResolutionMetadata{
 					Depth:               req.GetResolutionMetadata().Depth - 1,
 					DatastoreQueryCount: req.GetResolutionMetadata().DatastoreQueryCount,
@@ -751,6 +753,7 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 					StoreID:              req.GetStoreID(),
 					AuthorizationModelID: req.GetAuthorizationModelID(),
 					TupleKey:             tupleKey,
+					ContextualTuples:     req.GetContextualTuples(),
 					ResolutionMetadata: &ResolutionMetadata{
 						Depth:               req.GetResolutionMetadata().Depth - 1,
 						DatastoreQueryCount: req.GetResolutionMetadata().DatastoreQueryCount, // add TTU read below


cmd/openfga/main_test.go
```diff
@@ -203,6 +203,99 @@ func newOpenFGATester(t *testing.T, args ...string) OpenFGATester {
 	}
 }
 
+func TestCheckWithContextualTuplesAndQueryCacheEnabled(t *testing.T) {
```
jon-whit, line 206: This test is the same issue reported in #1058

cmd/openfga/main_test.go
```diff
@@ -203,6 +203,99 @@ func newOpenFGATester(t *testing.T, args ...string) OpenFGATester {
...
+				tuple.NewTupleKey("commerce_store:0", "hourly_employee", "fga_user:anne"),
+				tuple.NewTupleKey("commerce_store:1", "hourly_employee", "fga_user:anne"),
+				tuple.NewTupleKey("commerce_store:0", "approved_timeslot", "timeslot:11_12"),
+				tuple.NewTupleKey("commerce_store:1", "approved_timeslot", "timeslot:12_13"),
+			},
+		},
+	})
+	require.NoError(t, err)
+
+	checkResp, err := client.Check(context.Background(), &openfgav1.CheckRequest{
```
adriantam, line 259: not a blocker - wondering if we want to add these tests as test harness (i.e., for loop of test/expected values) for better readability and maintainability

jon-whit: I can refactor it to use a test table, sure. It's a very specific test, but making it more general purpose could be useful in the future.

adriantam (dismissed):

cmd/openfga/main_test.go
```diff
@@ -203,6 +204,123 @@ func newOpenFGATester(t *testing.T, args ...string) OpenFGATester {
 	}
 }
 
+func TestCheckWithQueryCacheEnabled(t *testing.T) {
```
miparnisari, line 207: can we rename this test to better explain what it is testing? e.g. `TestCheckPassesAlongContextualTuplesInDispatchedChecks`?

miparnisari (changes requested):

cmd/openfga/main_test.go
miparnisari: Also curious why did we make this an integration test? It can be made a unit test that runs much faster

jon-whit: Integration test is higher fidelity here, it was easier to write, and the same unit-test setup to exercise this test was proving to be awkward and more difficult as a unit test because of how we construct a LocalChecker per request.

jon-whit: > can we rename this test to better explain what it is testing? e.g. TestCheckPassesAlongContextualTuplesInDispatchedChecks?

> Otherwise this makes it sound like the subject under test is the cache, when the problem is actually within Check

The problem is with Check, specifically the cached check resolver, and since the test was re-written using a test table pattern I felt it was reasonble to name the test TestCheckWithQueryCacheEnabled, because there will be other tests we add for testing query cache behavior I'm sure.. This isn't the only one that will ever exist.

miparnisari: > The problem is with Check, specifically the cached check resolver,

To be more clear, by "check" I mean the LocalChecker. The cached check resolved made an assumption about LocalChecker that was not being maintained, this is why I say it's a problem within LocalChecker

jon-whit: I could pass in a mocked delegate to the LocalChecker and assert that the contextual tuples are propagated. We already have a test that exercises the original reported issue with high fidelity though. If we feel that doing this is worth it in addition to what's been done then I will add another test for that as well in `interna/graph/check_test.go`.

internal/graph/check.go
```diff
@@ -751,6 +753,7 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 					StoreID:              req.GetStoreID(),
 					AuthorizationModelID: req.GetAuthorizationModelID(),
 					TupleKey:             tupleKey,
+					ContextualTuples:     req.GetContextualTuples(),
```
miparnisari, line 756: I hate to be _that_person, but this is the only change needed to make `TestCheckWithQueryCacheEnabled` pass. Can we add tests for the other two lines?

jon-whit: Ok, I pushed some new tests that fail without these changes. These new tests also assert that the code paths for `checkDirect` (in the case of a userset lookup) and `checkComputedUserset` behave as expected when the Check query cache is enabled.

cmd/openfga/main_test.go
```diff
@@ -203,6 +204,210 @@ func newOpenFGATester(t *testing.T, args ...string) OpenFGATester {
...
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			createResp, err := client.CreateStore(context.Background(), &openfgav1.CreateStoreRequest{
+				Name: test.name,
+			})
+			require.NoError(t, err)
+			require.NotPanics(t, func() { ulid.MustParse(createResp.GetId()) })
```
miparnisari, line 362: 🤔

cmd/openfga/main_test.go
miparnisari: > proving to be awkward and more difficult as a unit test because of how we construct a LocalChecker per request.

This is part of why I sent https://github.com/openfga/openfga/pull/948 :) can you give it another look?

The next step to make it easier is to not have to do `context.WithTypesystem`.

miparnisari: My concern with integ tests vs unit tests: i am trying to verify that commenting out each change breaks one test. And to do that, i need to comment all lines, run docker build, run the test. Uncomment a line, run docker build, run the test, etc.

Whereas with a unit test i don't need to run docker build. And I can put breakpoints easily.

miparnisari (approved):

cmd/openfga/main_test.go
jon-whit: It's not, but why not assert expectations of the API? If it were to ever fail that'd be good to know anyways.

cmd/openfga/main_test.go
jon-whit: I don't disagree, but test fidelity and confidence is more important than convenience. We're testing to increase confidence not just for the sake of getting some rough coverage of changed code, and so I don't see anything wrong with favoring _functional tests_ in scenarios like this. If you're manually setting up the structures being tested, then you're not testing the system with as much fidelity as running the actual standalone server that end-users are going to be interacting with at the end of the day.

miparnisari: > We're testing to increase confidence not just for the sake of getting some rough coverage of changed code

? Coverage gives you a metric on confidence. The only way i noted that our tests were not good enough was when i ran them after making commenting out the fixes.

Issue #1033, opened 2023-10-02
Stuck with timeouts on v1.3.2


KlausVii:
**Summary**

We recently upgraded from 1.3.0 to 1.3.2, and since then after a while of use all of our calls to openFGA start failing with 
`"internal_error": "failed to FindLatestAuthorizationModelID: sql error: context deadline exceeded"` logs. The only temporary remedy was to restart the pods. We've rolled back to 1.3.0 and have not seen the error for a few hours now.

I am at a loss as to what could be causing this. There is no abnormalities being reported by the db and the load is minimal, this is in our testing environment. 

Considering how simple the FindLatestAuthorizationModelID query is I don't see how it could be timing out, unless its being provided with dead context or something. 

**OpenFGA version or commit**

"build.version":"v1.3.2"
"build.commit":"725296025fd81227c89525808652c6acd4a605f6"

If you know it, enter here.

**Store data**

**Other data**

- How are you running OpenFGA? (As a binary, in Docker, in Kubernetes): K8s
- What datastore are you using? (In memory, MySQL, Postgres): Postgres
- Are you running OpenFGA with any configuration overrides or with any of the flags mentioned in `./openfga run --help`?
- Do you have any logs or traces that could help us debug the problem?
```
{
  "level": "error",
  "timestamp": 1696256450.1516082,
  "msg": "rpc error: code = Code(4000) desc = Internal Server Error",
  "build.version": "v1.3.2",
  "build.commit": "725296025fd81227c89525808652c6acd4a605f6",
  "grpc_service": "openfga.v1.OpenFGAService",
  "grpc_method": "Check",
  "grpc_type": "unary",
  "request_id": "a880433e-ff9c-448a-b171-05e725597a19",
  "user_agent": "openfga-sdk js/0.2.8",
  "raw_request": {
    "store_id": "01H66ETS41XMTKPCCQ2TV5GZ5V",
    "tuple_key": {
      "object": "project:67c94dbb-03c5-4e6f-85e2-305c95d317bc",
      "relation": "can_view",
      "user": "user:fe999a2c-6554-4492-9ef6-9ddcdd3cc385"
    },
    "contextual_tuples": {
      "tuple_keys": []
    },
    "authorization_model_id": "",
    "trace": false
  },
  "raw_response": {
    "allowed": false,
    "resolution": ""
  },
  "peer.address": "127.0.0.1:59956",
  "store_id": "01H66ETS41XMTKPCCQ2TV5GZ5V",
[...]

PR #1061, opened 2023-10-13
fix: avoid forgotten sender with context cancellation in ReverseExpand

jon-whit:
## Description
This fixes a problem where dangling goroutines in ReverseExpand are all blocked on a channel that no longer has a listener. Instead, we listen on context and if it is cancelled we avoid trying to send on the channel.

## References
https://github.com/openfga/openfga/issues/1033

## Review Checklist

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -215,7 +215,12 @@ func (q *ListObjectsQuery) evaluate(
 			reverseexpand.WithResolveNodeBreadthLimit(q.resolveNodeBreadthLimit),
 		)
 
+		wg := sync.WaitGroup{}
+
+		wg.Add(1)
 		go func() {
+			defer wg.Done()
+
 			err = reverseExpandQuery.Execute(ctx, &reverseexpand.ReverseExpandRequest{
 				StoreID:          req.GetStoreId(),
 				ObjectType:       targetObjectType,
@@ -230,10 +235,8 @@ func (q *ListObjectsQuery) evaluate(
 				}
 
 				resultsChan <- ListObjectsResult{Err: err}
+				return
 			}
-
-			// this is necessary to terminate the range loop below
-			close(reverseExpandResultsChan)
 		}()
 
 		checkResolver := graph.NewLocalChecker(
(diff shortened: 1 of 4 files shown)


pkg/server/commands/list_objects.go
```diff
@@ -229,6 +229,10 @@ func (q *ListObjectsQuery) evaluate(
 					return
 				}
 
+				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
jon-whit, line 232: This still doesn't feel right to me. If we experience a cancellation or deadline error we should still surface upwards. It's the callers responsibility to choose what to do with it, not the callee's responsibility here.

Without this code though you get a panic because you. may try to send on the channel below meanwhile the caller has returned.

jon-whit: I'm going to move the WaitGroup up above this and add the ReverseExpand bit to the WaitGroup. That should avoid the closure of the channel before this function may yield on it. I believe that is the best course of action.

pkg/server/commands/list_objects.go
```diff
@@ -355,7 +352,11 @@ func (q *ListObjectsQuery) Execute(
 					return nil, result.Err
 				}
 
-				return nil, serverErrors.HandleError("", result.Err)
+				if !(errors.Is(result.Err, context.Canceled) || errors.Is(result.Err, context.DeadlineExceeded)) {
+					return nil, serverErrors.HandleError("", result.Err)
+				}
+
+				continue
```
jon-whit, line 359: The assumption here is that if any of the sub-Checks or the ReverseExpand query returns a cancellation or deadline exceeded error, then we assume that either a) the operation was cancelled because this code path has already met the limit and has since returned (e.g. this would be a context.Canceled) or b) the deadline wasn't met.

In these cases we return everything we could find up to that point.

pkg/server/server_test.go
```diff
@@ -200,6 +201,80 @@ func TestCheckDoesNotThrowBecauseDirectTupleWasFound(t *testing.T) {
...
+
+	timeoutCtx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+
+	checkResp, err := s.Check(timeoutCtx, &openfgav1.CheckRequest{
+		StoreId:              storeID,
+		AuthorizationModelId: modelID,
+		TupleKey:             tuple.NewTupleKey("document:0", "editor", "user:jon"),
+	})
+	require.NoError(t, err)
```
jon-whit, line 274: This Check fails without the code changes in ReverseExpand, because it timesout due to there not being an available connection in the database connection pool, since it is being hogged by the ListObjects call still.

adriantam (dismissed):

miparnisari (approved):

PR #1064, opened 2023-10-17
fix(list-objects): promptly return if max results are met before deadline

jon-whit:
## Description
Follow up to https://github.com/openfga/openfga/pull/1061. This PR ensures that if the max results are met before the deadline, then we return promptly instead of waiting up until the deadline.

## Review Checklist


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -259,8 +270,9 @@ func (c *ReverseExpandQuery) execute(
 		return err
 	}
 
-	subg, subgctx := errgroup.WithContext(ctx)
-	subg.SetLimit(int(c.resolveNodeBreadthLimit))
+	pool := pool.New().WithContext(ctx)
```
jon-whit, line 273: The sourcegraph concurrency pool works like I would have expected `errgroup` to behave. What I noticed is that when ctx is cancelled, the errgroup still waits for the goroutines to return, and thus it can hang. However, using the sourcegraph concurrency pool the pool is immediately interrupted and returns promptly.

I made this change because a) it was necessary to get the concurrency behavior that I expected when context is cancelled on group of goroutines, and b) it minimized the complexity of the change as compared to introducing waitgroups etc..

pkg/server/commands/list_objects.go
```diff
@@ -208,20 +208,22 @@ func (q *ListObjectsQuery) evaluate(
 		}
 
 		reverseExpandResultsChan := make(chan *reverseexpand.ReverseExpandResult, 1)
-		var objectsFound = new(uint32)
+		objectsFound := atomic.Uint32{}
```
miparnisari, line 211: is this change necessary in this PR?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -177,8 +178,14 @@ func (c *ReverseExpandQuery) Execute(
...
-	defer close(resultChan)
-	return c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	err := c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	if err != nil {
+		resultChan <- &ReverseExpandResult{Err: err}
+	} else {
+		close(resultChan)
+	}
+
+	return err
```
miparnisari, line 188: The error case looks very odd... we send `err` through the channel _and_ we return it synchronously at the very least we need to document this so that clients (right now, only ListObjects) know how to handle it.

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: > using the sourcegraph concurrency pool the pool is immediately interrupted and returns promptly.

does this mean that no more elements will be sent through the channel? 

if so, why don't we `close(resultsChan)` in all cases?

pkg/server/commands/list_objects.go
miparnisari: addressed by https://github.com/openfga/openfga/pull/1067

pkg/server/commands/reverseexpand/reverse_expand.go
jon-whit: I've reworked the error handling to produce an error on the channel and close the channel if Execute returns without any error. Let me know what you think 👍

pkg/server/commands/list_objects.go
```diff
@@ -294,6 +299,7 @@ func (q *ListObjectsQuery) evaluate(
 			}(res)
 		}
 
+		cancel()
 		wg.Wait()
 		close(resultsChan)
```
miparnisari, line 304: can we put tracing around this code? so that if in the future someone reports an issue about ListObjects being slow, we'll be able to see if time is being spent waiting?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -171,14 +172,23 @@ func NewResolutionMetadata() *ResolutionMetadata {
...
 	req *ReverseExpandRequest,
 	resultChan chan<- *ReverseExpandResult,
 	resolutionMetadata *ResolutionMetadata,
-) error {
-	defer close(resultChan)
-	return c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+) {
+	err := c.execute(ctx, req, resultChan, nil, resolutionMetadata)
+	if err != nil {
+		resultChan <- &ReverseExpandResult{Err: err}
```
miparnisari, line 187: I ran `BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkListObjects/oneResult` manually with a few print statements added. This line becomes blocked because `resultChan` is only of size 1. If we increase it to size 100, the benchmark runs much faster

jon-whit: I checked out `v1.3.4` and tested this and it also occasionally fails, so I believe this is unrelated to this PR directly. We should create an issue and track separately.

miparnisari: But then the title of the PR "promptly return if max results are met before deadline" is misleading 🤔

jon-whit: Why? In `v1.3.4` we are not promptly returning. We wait all the way up to the deadline. In this PR we do promptly return as soon as we've been cancelled, the deadline has been reached, and/or we found all remaining results.

pkg/server/commands/list_objects.go
jon-whit: Not in this PR, but if you want to create a follow up issue we can consider it. We already have tracing in ReverseExpand and that should give a very good indication if we're blocked waiting on a channel send or if we're sluggish because of database queries being blocked based on concurrency settings etc.. I'm not sure I see the value of adding a span here. In either case, it's outside the scope of the original reason this PR was created.

miparnisari (approved):

PR #1066, opened 2023-10-18
fix: ensure usersets are self defining

jon-whit:
## Description
This PR ensures that the variant that a userset defines itself it met. This defines the reflexive property of sets.

A userset such as 'document:1#viewer' always defines itself. That is, `Check(document:1#viewer@document:1#viewer)` should always return true, regardless of the tuples in the system and/or the type restrictions.

One of the motivating factors for this work is based on basic set theory and practical use cases we have for indexing relationships. To take one example, in the Zanzibar paper they described the mechanics behind the Leopard indexing system. In particular they mentioned building an index by taking the intersection between two sets - a Member2Group (M2G) and Group2Group (G2G) index which stored direct user to group relationships and group to group relationships, respectively. 

Let's assume that we have an FGA model such as the following:
```
type user

type group
  relations
```
and the following relationship tuples:
- group:eng#member@group:fga#member
- group:fga#member@user:jon

The M2G index would be defined as:
| object_type | object_id | relation | subject_type | subject_id | subject_relation |
|-------------|-----------|----------|--------------|------------|------------------|
| group       | fga       | member   | user         | jon        |                    |

And the G2G index, naively, would be defined as:
| object_type | object_id | relation | subject_type | subject_id | subject_relation |
|-------------|-----------|----------|--------------|------------|------------------|
| group       | eng       | member   | group        | fga        | member           |

We take the intersection of M2G and G2G, namely
```
SELECT g2g.object_type, g2g.object_id, g2g.relation, m2g.subject_type, m2g.subject_id
FROM M2G m2g 
INNER JOIN G2G on m2g.object_type=g2g.object_type AND m2g.relation=g2g.relation
```
which returns
| object_type | object_id | relation | subject_type | subject_id |
|-------------|-----------|----------|--------------|------------|
| group       | eng       | member   | user         | jon        |

Notice that we account for the `group:eng#member` relationship for `user:jon`, but the `group:fga#member` relationship is missing. This is because we naively overlooked the fact that `group:fga#member` defines itself. That is, every subject who is a `member` of `group:fga` is definitely a subject of `group:fga#member`. A subject set defines itself, because it is reflexive. So our system should represent this and reflect this natively in our APIs, and what we should have had for the G2G index as a side-effect of this property is the following:

| object_type | object_id | relation | subject_type | subject_id | subject_relation |
|-------------|-----------|----------|--------------|------------|------------------|
| group       | eng       | member   | group        | fga        | member           |
| group       | fga       | member   | group        | fga        | member           |

Then our join would return:
| object_type | object_id | relation | subject_type | subject_id |
|-------------|-----------|----------|--------------|------------|
| group       | eng       | member   | user         | jon        |
| group       | fga       | member   | user         | jon        |

## Review Checklist


miparnisari (approved):

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5090,4 +5090,30 @@ tests:
...
+
+          type document
+            relations
+              define viewer: [user] as self
+        tuples:
+        checkAssertions:
+          - tuple:
+              user: document:1#viewer
+              relation: viewer
+              object: document:1
```
miparnisari, line 5107: should we be rejecting these tuples on writes since they are redundant?

internal/graph/check.go
```diff
@@ -450,6 +450,19 @@ func (c *LocalChecker) ResolveCheck(
 	object := req.GetTupleKey().GetObject()
 	relation := req.GetTupleKey().GetRelation()
 
+	userObject, userRelation := tuple.SplitObjectRelation(req.GetTupleKey().GetUser())
+	if userRelation != "" {
+		if object == userObject && relation == userRelation {
+			return &ResolveCheckResponse{
+				Allowed: true,
+				ResolutionMetadata: &ResolutionMetadata{
+					Depth:               defaultResolveNodeLimit,
```
miparnisari, line 459:
```suggestion
					Depth:               req.GetResolutionMetadata().Depth,
```

since you can't be sure that you encountered the tupleKey at the first call

miparnisari (changes requested):

internal/graph/check.go
```diff
@@ -450,6 +450,19 @@ func (c *LocalChecker) ResolveCheck(
...
+					DatastoreQueryCount: 0,
```
miparnisari, line 460: Why 0 and not `req.GetResolutionMetadata().DatastoreQueryCount`?

assets/tests/consolidated_1_1_tests.yaml
jon-whit: This tuple wouldn't be a valid tuple at Write time, because the type restrictions doesn't allow writing `document#viewer` relationships to `document#viewer`. This is currently how the system behaves, and that should remain unchanged. This behavior is a reflexive property of a set of users in general. A set of objects related to the object itself is self defining. That is

`document:1#viewer` reflects itself `document:1#viewer`. So if you Check it, then it is logical that it should be {allowed: true}.

willvedd:
@jon-whit Is this still in a reviewable state?  Also, is this TODO still outstanding? 

> update the tests within `typesystem_test.go` to account for any changes in the entrypoint calculation

I'm trying to understand if this PR can be realistically pushed over the finish or if we close-out.

assets/tests/consolidated_1_1_tests.yaml
miparnisari: rephrasing my question: if the model did allow such tuples, e.g.

model
  schema 1.1
type doc
	 viewer: [doc#viewer]


then should the write of `doc:1 # viewer @ doc:1#viewer`  be allowed? we're going to consider this tuple as `true` and never look it up 🤷‍♀️

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -267,6 +267,16 @@ func (c *ReverseExpandQuery) execute(
 		sourceUserObj = val.ObjectRelation.GetObject()
 		sourceUserRef = typesystem.DirectRelationReference(sourceUserType, val.ObjectRelation.GetRelation())
 
+		sourceUserRel := val.ObjectRelation.GetRelation()
+
+		if req.edge == nil {
```
miparnisari, line 272: **question**. With this line we are only consider this "fact" as true on the first (base) call of the recursion. Whereas in Check, you are treating it as true on the base case and every subsequent recursion call. Is this intentional?

willvedd: **Another question:** Is this check required? Couldn't you have a self-referencing tuple as a result of another expansion or itself expand further?

Ex:

```
model
  schema 1.1
type user
  relations
    define editor: [user]
    define viewer: [user] or editor
```
```
Check(user:1#editor@user:1#editor) // True
Check(user:1#editor@user:1#viewer) // ??? 
```

Another case to ponder:
```
model
  schema 1.1
type user
  relations
    define blocked: [user]
    define viewer: [user] but not blocked
```
```
user:1#blocked@user:1
```
```
Check(user:1#viewer@user:1#viewer) // ??
```

willvedd (commented):
Trying to wrap my head around the implications of this. Its intuitive that a tuple can reference itself but does that have knock on effects with expansion?

assets/tests/consolidated_1_1_tests.yaml
jon-whit: Yes, there's nothing restricting that from occurring and nor should there be. If you want to establish a concrete relationship with a userset of the same type and relation, then you should be able to do so.  For example, if you want to establish that all subjects who can view `doc:Z` can view `doc:1` (e.g. `doc:1#viewer@doc:Z#viewer`) then you should be able to do that. By extension, if you want to establish a self-defining tuple of the form `doc:1#viewer@doc:1#viewer` then writing that tuple is reasonable so long as the type restrictions allow it as well.

The invariants for Check are different than the invariants for Write. Namely, you can Check if a subject has a relationship who would otherwise be unrelated based on the type-restrictions, and you'll get an `{allowed: false}` outcome. Whereas with Write you can only write that which the type restrictions permit.

If you want to determine if the set of subjects of some relationship contain the same set of subjects from another relationship, then that is also reasonable to ask OpenFGA (e.g. `Check(document:1#viewer@group:eng#member)`) . By extension, if you want to Check if the set of subjects of some relationship are also related to that same relationship, then the answer should always be `{allowed: true}` because of the reflexive property of sets. A set of subjects of some relationships is equal to the same set of subjects of the same relationship. For example, `Check(document:1#viewer@document:1#viewer) --> {allowed: true}`.

jon-whit:
@willvedd in response to your questions above, I'll answer inline:

```
model
  schema 1.1

type user
  relations
    define editor: [user]
    define viewer: [user] or editor
```

1. Check(user:1#editor@user:1#editor) // True
2. Check(user:1#editor@user:1#viewer) // ??? 

For Check (2) what is being asked is "Is the set of the subjects `user:1#viewer` (the set itself) equivalent to the set of subjects represented by `user:1#editor`"? 

Mathematically, no they are not equal. The set of subjects who are viewers of a particular user are not the same as the set of subjects who are editors. Thinking in terms of reference equality, the two sets are different references. Another way to look at it, there can exist sets of subjects for `user:1#viewer` and `user:1#editor` where `user:1#viewer` does not have the same elements as `user:1#editor`. They not only can contain different set elements, but they also do not refer to the same physical set in the first place. A set X is referentially equal to the same set reference X, but not necessarily some set reference Y.

```
model
  schema 1.1
type user
  relations
    define blocked: [user]
    define viewer: [user] but not blocked
```

- user:1#blocked@user:1

3. Check(user:1#viewer@user:1#viewer) // ??

Following along with the explanation above, `Check(user:1#viewer@user:1#viewer) --> {allowed: true}`. A set of subjects is definitively equal to itself. The set of subjects of a relationship are equal to the same set of subjects of the same relationship.

However, given the model constraint on `viewer but not blocked`, an individual subject cannot be present within the subject set `user:1#viewer` unless that subject is a viewer and is not blocked. So in the example `user:1` cannot be a 'viewer' member of the set of subjects who can view `user:1` (e.g. `user:1#viewer`) because that user is blocked. But that is different than the reference to the set of subjects who can view `user:1`.

> Trying to wrap my head around the implications of this. Its intuitive that a tuple can reference itself but does that have knock on effects with expansion?

It's not so much a tuple as it is a computed relationship. A relationship expands to 0 or more subjects/objects of 1 or more subject/object types, and the set of subjects/objects a relationship expands to is the subject set (userset). If we are comparing two subject sets that refer to the same set of subjects, then they are obviously the same, because any subject in one is also in the other. On the other hand, two different subject sets are not equal, though the subject contained in their sets may be overlapping depending on the invariants of the FGA model.

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: But we don't have a top-level ResolveCheck. ResolveCheck is recursive

jon-whit: @miparnisari could you be more specific with what you are wanting to see? Are you suggesting that we move this up to the `Execute` function instead of inside of the `execute` function? If so, I can do that, but it's functionally the same. `execute` where `req.edge==nil` only happens on the first invocation, because any subsequent recursive call will have a non-nil edge.

miparnisari: I'm not sure what is the correct approach here, whether it's okay to do the check on the top-level call only, or we should also do it in each sub-problem. I haven't thought about the implications. What i'm pointing out is that in your PR this isn't consistent. 

Since ResolveCheck is for both the top-level call and the sub-problems, until we figure out the implications, IReverseExpand should behave the same and we should get rid of `if req.edge == nil` check

willvedd: > whether it's okay to do the check on the top-level call only, or we should also do it in each sub-problem. 

That's what I was trying to get at above. It would be odd if a reflexive userset only returned itself at the top-level call but not for all other subproblems. @jon-whit is that even possible?

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5295,3 +5295,30 @@ tests:
               relation: viewer
             expectation:
               - document:1
+  - name: userset_defines_itself
```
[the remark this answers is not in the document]
miparnisari: Yes please :D 

```
model
  schema 1.1

type user

type document
  relations
    define viewer: writer or editor
    define editor: [user]
    define writer: [user]
```

`Check(document:1 # viewer @ document:1#writer)` should return `allowed:true`

willvedd:
A wonderful side-effect of this work:

```
model
  schema 1.1
type user

type document
  relations
    define writer: [user]
    define viewer: writer
```
_No tuples required_

**Before:**
```
Check(document:1#viewer, writer, document:1)// False
```

**After:**
```
Check(document:1#viewer, writer, document:1)// True
```

^ Seeing this answers the questions I had above and demonstrates the value to me. It'll enable a lot of work in `ListUsers` too. 👍

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: Discussed on Zoom. Let's remove `if req.edge == nil` :)

jon-whit: We talked about this offline, but just reposting for broader visibility and historical context.

I've added some tests that exhibit some of the behaviors around the questions asked in this thread. The `tl;dr` is all based around set theory. If every subject of some subject set would also be guaranteed to be a subset of the other, then we should be considering the subgraphs as equivalent subgraphs. As just one simple example,

```
type user

type document
  relations
    define writer: [user]
    define editor: [user]
    define viewer: writer or editor
```

`Check(document:1#viewer@document:1#writer)` --> `{allowed: true}`, because every subject who is a subject of the set `document:1#writer` is also a subject of the set `document:1#viewer`. Similarly `Check(document:1#viewer@document:1#writer)` --> `{allowed: true}`.

However, this isn't necessarily true in all cases, for example,

```
type user

type document
  relations
    define allowed: [user]
    define viewer: [user] and allowed
```

`Check(document:1#viewer@document:1#allowed)` --> {allowed: false} because not all subjects of `document:1#allowed` are also part of the set `document:1#viewer`.

assets/tests/consolidated_1_1_tests.yaml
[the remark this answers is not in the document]
jon-whit: Added a broader set of test cases. Let me know if there are some more specific tests you want to add. Going to resolve this thread for now. Just post inline for any additional tests you want to add.

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5295,3 +5295,294 @@ tests:
...
+        - tuple:
+            user: document:1#restricted
+            relation: viewer
+            object: document:1
+          expectation: false
+        - tuple:
+            user: document:1#viewer
+            relation: viewer
+            object: document:1
+          expectation: true
```
miparnisari, line 5522: missing the equivalent ListObejcts assertion

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5295,3 +5295,294 @@ tests:
...
+    stages:
+      - model: |
+          model
+            schema 1.1
+
+          type user
+
+          type group
+            relations
+              define member: [user, group#member] but not other
```
miparnisari, line 5544: nit: can we rename `other` to `blocked`?

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5295,3 +5295,294 @@ tests:
...
+            user: document:1#writer
+            relation: viewer
+            object: document:1
+          expectation: true
+        listObjectsAssertions:
+          - request:
+              user: document:1#writer
+              type: document
+              relation: viewer
+            expectation: ["document:1"]
```
miparnisari, line 5588: can we add another test? (it passes on main BTW)

```
  - name: userset_defines_itself_9
    stages:
      - model: |
          model
            schema 1.1
          type user
          type group
            relations
              define member: [user]

          type document
            relations
              define blocked: [user]
              define viewer: [group#member] but not blocked
        tuples:
          - user: group:fga#member
            relation: viewer
            object: document:1
          - user: user:maria
            relation: blocked
            object: document:1
        checkAssertions:
          - tuple:
              user: group:fga#member
              relation: viewer
              object: document:1
            expectation: true #the group as a whole has access, but individual members may not
        listObjectsAssertions:
          - request:
              user: group:fga#member
              type: document
              relation: viewer
            expectation: [document:1]
```

miparnisari (approved):

willvedd (dismissed):
Those new tests do a good job highlighting the power of this change! Looks good to me.

assets/tests/consolidated_1_1_tests.yaml
jon-whit: I have the test case `Check(x#y@x#y)` tested in various other places, but I'll add it here for good measure too.

willvedd (approved):

pkg/server/test/expand.go
```diff
@@ -865,6 +866,37 @@ func TestExpandQuery(t *testing.T, datastore storage.OpenFGADatastore) {
...
+			},
+			expected: &openfgav1.ExpandResponse{
+				Tree: &openfgav1.UsersetTree{
+					Root: &openfgav1.UsersetTree_Node{
+						Name: "document:1#viewer",
+						Value: &openfgav1.UsersetTree_Node_Leaf{
+							Leaf: &openfgav1.UsersetTree_Leaf{
+								Value: &openfgav1.UsersetTree_Leaf_Users{
+									Users: &openfgav1.UsersetTree_Users{
+										Users: []string{},
```
miparnisari, line 891: We need to return `document:1#viewer` here

Issue #1044, opened 2023-10-05
When OPENFGA_LIST_OBJECTS_MAX_RESULTS=0, 0 results are returned


miparnisari:
Hi @miparnisari  and @jon-whit.

We've been experimenting with OpenFGA at Cochlear and we'd like to appreciate the amazing work you've been doing. It's awesome to see this how this project's evolving.

On a more codebase related question: How are you ensuring you return all data when `OPENFGA_LIST_OBJECTS_MAX_RESULTS` is set to `0`?

I couldn't find any test case that validates that flow.

Also, that setting doesn't seem to work for me. I'm happy to help and contribute with a fix if that's the case.

Thank you!

_Originally posted by @jcbertolucci in https://github.com/openfga/openfga/issues/757#issuecomment-1747893509_

miparnisari:
Seems like it broke in this commit https://github.com/openfga/openfga/commit/5c7c5c88dd07c85666275621503f867428a59693 because the previous commit respects the 0 value

miparnisari:
@jcbertolucci this was released as part of 1.3.5. Thanks for the report!

jcbertolucci:
@miparnisari all good! You guys are the rockstars here.

Thank you for the quick turnaround ♥️

PR #1067, opened 2023-10-18
fix: return all results when `OPENFGA_LIST_OBJECTS_MAX_RESULTS=0`

miparnisari:
## Description
- Test: Update benchmarks for ListObjects
- Fix: When OPENFGA_LIST_OBJECTS_MAX_RESULTS=0, return all results

## References
- Issue https://github.com/openfga/openfga/issues/1044

- https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -304,13 +304,11 @@ func (q *ListObjectsQuery) evaluate(
 }
 
 func trySendObject(object string, objectsFound *atomic.Uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
-	if getAllResults(maxResults) {
-		goto send
+	if !getAllResults(maxResults) {
+		if objectsFound.Add(1) > maxResults {
+			return
+		}
 	}
-	if objectsFound.Add(1) > maxResults {
-		return
-	}
-send:
 	resultsChan <- ListObjectsResult{ObjectID: object}
 }
 
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -208,7 +208,7 @@ func (q *ListObjectsQuery) evaluate(
 		}
 
 		reverseExpandResultsChan := make(chan *reverseexpand.ReverseExpandResult, 1)
-		var objectsFound = new(uint32)
+		objectsFound := atomic.Uint32{}
 
 		reverseExpandQuery := reverseexpand.NewReverseExpandQuery(q.datastore, typesys,
 			reverseexpand.WithResolveNodeLimit(q.resolveNodeLimit),
@@ -247,13 +247,13 @@ func (q *ListObjectsQuery) evaluate(
 		concurrencyLimiterCh := make(chan struct{}, q.resolveNodeBreadthLimit)
 
 		for res := range reverseExpandResultsChan {
-			if atomic.LoadUint32(objectsFound) >= maxResults {
+			if !getAllResults(maxResults) && objectsFound.Load() >= maxResults {
 				break
 			}
 
 			if res.ResultStatus == reverseexpand.NoFurtherEvalStatus {
 				noFurtherEvalRequiredCounter.Inc()
-				trySendObject(res.Object, objectsFound, maxResults, resultsChan)
+				trySendObject(res.Object, &objectsFound, maxResults, resultsChan)
 				continue
 			}
 
@@ -289,7 +289,7 @@ func (q *ListObjectsQuery) evaluate(
 				atomic.AddUint32(resolutionMetadata.QueryCount, resp.GetResolutionMetadata().DatastoreQueryCount)
 
 				if resp.Allowed {
-					trySendObject(res.Object, objectsFound, maxResults, resultsChan)
+					trySendObject(res.Object, &objectsFound, maxResults, resultsChan)
 				}
 			}(res)
 		}
@@ -303,11 +303,19 @@ func (q *ListObjectsQuery) evaluate(
 	return nil
 }
 
-func trySendObject(object string, objectsFound *uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
-	if objectsFound != nil && atomic.AddUint32(objectsFound, 1) > maxResults {
+func trySendObject(object string, objectsFound *atomic.Uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
+	if getAllResults(maxResults) {
+		goto send
+	}
+	if objectsFound.Add(1) > maxResults {
 		return
 	}
+send:
 	resultsChan <- ListObjectsResult{ObjectID: object}
+}
+
+func getAllResults(maxResults uint32) bool {
+	return maxResults == 0
 }
 
 // Execute the ListObjectsQuery, returning a list of object IDs up to a maximum of q.listObjectsMaxResults
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -247,7 +247,7 @@ func (q *ListObjectsQuery) evaluate(
 		concurrencyLimiterCh := make(chan struct{}, q.resolveNodeBreadthLimit)
 
 		for res := range reverseExpandResultsChan {
-			if !getAllResults(maxResults) && objectsFound.Load() >= maxResults {
+			if !(maxResults == 0) && objectsFound.Load() >= maxResults {
 				break
 			}
 
@@ -304,16 +304,12 @@ func (q *ListObjectsQuery) evaluate(
 }
 
 func trySendObject(object string, objectsFound *atomic.Uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
-	if !getAllResults(maxResults) {
+	if !(maxResults == 0) {
 		if objectsFound.Add(1) > maxResults {
 			return
 		}
 	}
 	resultsChan <- ListObjectsResult{ObjectID: object}
-}
-
-func getAllResults(maxResults uint32) bool {
-	return maxResults == 0
 }
 
 // Execute the ListObjectsQuery, returning a list of object IDs up to a maximum of q.listObjectsMaxResults
--- a/pkg/server/test/list_objects.go
+++ b/pkg/server/test/list_objects.go
@@ -386,8 +386,6 @@ func BenchmarkListObjects(b *testing.B, ds storage.OpenFGADatastore) {
 		User:                 "user:maria",
 	}
 
-	b.ResetTimer()
-
 	var r *commands.ListObjectsResponse
 
 	var oneResultIterations, allResultsIterations int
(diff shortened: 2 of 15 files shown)


pkg/server/commands/list_objects.go
```diff
@@ -247,13 +247,13 @@ func (q *ListObjectsQuery) evaluate(
 		concurrencyLimiterCh := make(chan struct{}, q.resolveNodeBreadthLimit)
 
 		for res := range reverseExpandResultsChan {
-			if atomic.LoadUint32(objectsFound) >= maxResults {
+			if !getAllResults(maxResults) && objectsFound.Load() >= maxResults {
```
miparnisari, line 250: this is the fix

pkg/server/commands/list_objects.go
```diff
@@ -303,13 +303,21 @@ func (q *ListObjectsQuery) evaluate(
 	return nil
 }
 
-func trySendObject(object string, objectsFound *uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
-	if objectsFound != nil && atomic.AddUint32(objectsFound, 1) > maxResults {
+func trySendObject(object string, objectsFound *atomic.Uint32, maxResults uint32, resultsChan chan<- ListObjectsResult) {
+	if getAllResults(maxResults) {
+		goto send
```
jon-whit, line 308: Avoid the goto here.

```go
if !getAllResults(maxResults) {
    if objectsFound.Add(1) > maxResults {
        return
    }
}

resultsChan <- ListObjectsResult{ObjectID: object}
```

pkg/server/test/server.go
```diff
@@ -66,6 +66,8 @@ func RunAllBenchmarks(b *testing.B, ds storage.OpenFGADatastore) {
 }
 
 func RunListObjectsBenchmarks(b *testing.B, ds storage.OpenFGADatastore) {
-	b.Run("BenchmarkListObjectsWithReverseExpand", func(b *testing.B) { BenchmarkListObjectsWithReverseExpand(b, ds) })
-	b.Run("BenchmarkListObjectsWithConcurrentChecks", func(b *testing.B) { BenchmarkListObjectsWithConcurrentChecks(b, ds) })
+	b.Run("BenchmarkListObjectsWithMaxConcurrentReadsEqualToOne", func(b *testing.B) { BenchmarkListObjectsWithMaxConcurrentReadsEqualToOne(b, ds) })
```
jon-whit, line 69: Are we actually benchmarking performance on these things? It feels like these are more of tests than benchmarks. What are we trying to show with these benchmarks?

jon-whit (changes requested):
Left a question and a code change recommendation.

pkg/server/test/server.go
miparnisari: > Are we actually benchmarking performance on these things?

What do you mean? Yes, the PR build should fail if performance gets worse. (I may need to fine tune the `bench` github action in the future + look into Check benchmarks)

> What are we trying to show with these benchmarks?

Not trying to "show" anything, the goal is for PR builds to fail if the benchmark results get worse. I feel like we would have caught the regression fixed in https://github.com/openfga/openfga/pull/1064 if we had had them 🤔

miparnisari:
I'm getting a failure locally :( 

```go
BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkListObjectsWithMaxConcurrentReadsEqualToOne          panic: send on closed channel

goroutine 3386 [running]:
github.com/openfga/openfga/pkg/server/commands/reverseexpand.(*ReverseExpandQuery).trySendCandidate(0xc000398000, {0x1dd9cf8, 0xc0009bf5c0}, 0xc000098420, {0xc000989ad0, 0xd}, 0xc000354240)
        /Users/mparnisari/GitHub/openfga/pkg/server/commands/reverseexpand/reverse_expand.go:509 +0x697
github.com/openfga/openfga/pkg/server/commands/reverseexpand.(*ReverseExpandQuery).execute(0xc000398000, {0x1dd9d30, 0xc0003b2c80}, 0xc0009bae40, 0xc000354240, 0xc000098420, 0xc00012c470)
        /Users/mparnisari/GitHub/openfga/pkg/server/commands/reverseexpand/reverse_expand.go:246 +0xde8
github.com/openfga/openfga/pkg/server/commands/reverseexpand.(*ReverseExpandQuery).reverseExpandDirect.func1()
        /Users/mparnisari/GitHub/openfga/pkg/server/commands/reverseexpand/reverse_expand.go:482 +0x1b2
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /Users/mparnisari/go/pkg/mod/golang.org/x/sync@v0.4.0/errgroup/errgroup.go:75 +0x56
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1930
        /Users/mparnisari/go/pkg/mod/golang.org/x/sync@v0.4.0/errgroup/errgroup.go:72 +0x96
exit status 2
FAIL    github.com/openfga/openfga/pkg/server   18.862s
```

pkg/server/test/server.go
miparnisari: Ok so for example, in this PR:

`BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListObjectsWithMaxResultsEqualToOne                           2        3015844262 ns/op        17364208 B/op     389812 allocs/op`

On your PR https://github.com/openfga/openfga/pull/1064:

`BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListObjectsWithMaxResultsEqualToOne                         320          18427689 ns/op         4534970 B/op      90500 allocs/op`

:)

pkg/server/commands/list_objects.go
```diff
@@ -303,13 +303,19 @@ func (q *ListObjectsQuery) evaluate(
...
+	if !getAllResults(maxResults) {
+		if objectsFound.Add(1) > maxResults {
+			return
+		}
 	}
 	resultsChan <- ListObjectsResult{ObjectID: object}
 }
 
+func getAllResults(maxResults uint32) bool {
+	return maxResults == 0
```
jpadilla, line 316: imho this function doesn't add any more clarity than just doing the check where needed

jon-whit: Agree

pkg/server/test/list_objects.go
```diff
@@ -308,55 +334,40 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
...
+	listObjectsResponse = r
+}
+
+func BenchmarkListObjectsWithMaxConcurrentReadsEqualToThirty(b *testing.B, ds storage.OpenFGADatastore) {
+	ctx := context.Background()
+	store := ulid.Make().String()
+
+	model, modelID, numberObjectsAccesible := setupListObjectsBenchmark(b, ds, store)
+
+	listObjectsQuery := commands.NewListObjectsQuery(ds, commands.WithMaxConcurrentReads(30))
```
adriantam, line 370: For these tests (BenchmarkListObjectsWithMaxConcurrentReadsEqualToOne, BenchmarkListObjectsWithMaxConcurrentReadsEqualToThirty, BenchmarkListObjectsWithMaxResultsEqualToZero, etc.), other than setting commands.WithMaxConcurrentReads different, are there any difference?  If not, can we have common function to make things easier to read?

miparnisari: will do

pkg/server/commands/list_objects.go
```diff
@@ -247,13 +247,13 @@ func (q *ListObjectsQuery) evaluate(
 		concurrencyLimiterCh := make(chan struct{}, q.resolveNodeBreadthLimit)
 
 		for res := range reverseExpandResultsChan {
-			if atomic.LoadUint32(objectsFound) >= maxResults {
+			if !(maxResults == 0) && objectsFound.Load() >= maxResults {
```
jpadilla, line 250: nit: isn't this just `maxResults != 0`?

jpadilla (dismissed):

pkg/server/commands/list_objects.go
miparnisari: Yes. See, this is why I had written `getAllResults`, LOL

miparnisari: I can't do math.

miparnisari: I'll try to make this simpler to read in a future PR. For now I am keeping as is, to be consistent with line 307

adriantam (approved):

pkg/server/test/list_objects.go
```diff
@@ -308,132 +334,88 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
...
-			Relation:             "viewer",
-			User:                 "user:999",
-		})
-	}
+		listObjectsResponse = r
+		allResultsIterations = b.N
+	})
 
-	listObjectsResponse = r
+	require.Greater(b, oneResultIterations, allResultsIterations)
```
miparnisari, line 420: Not comparing `b.Elapsed()` here because `benchtime=5s` for all benchmarks, so the comparison is useless and sometimes misleading

pkg/server/test/list_objects.go
```diff
@@ -297,7 +321,9 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
 
 				require.NotNil(t, res)
 				require.NoError(t, err)
-				require.LessOrEqual(t, len(res.Objects), int(test.maxResults))
+				if test.maxResults != 0 { // don't get all results
```
jon-whit, line 324: If `text.maxResults == 0` then we should just assert they are exactly equal (ignoring order). Otherwise, then we can assert that the length of the results is less than or equal to max results AND greather than or equal to minimum results. That is,

```go
if test.maxresults == 0 {
    require.ElementsMatch(...)
} else {
    require.LessOrEqual(t, len(res.Objects), int(test.maxResults))
    require.GreaterOrEqual(t, len(res.Objects), int(test.minimumResultsExpected))
    require.Subset(t, test.allResults, res.Objects)
}
```

pkg/server/test/list_objects.go
```diff
@@ -308,132 +334,88 @@ func TestListObjectsRespectsMaxResults(t *testing.T, ds storage.OpenFGADatastore
...
+		Relation:             "viewer",
+		User:                 "user:maria",
 	}
-	err := ds.WriteAuthorizationModel(ctx, store, model)
-	require.NoError(b, err)
 
-	n := 0
-	for i := 0; i < 100; i++ {
-		var tuples []*openfgav1.TupleKey
+	b.ResetTimer()
```
jon-whit, line 389: Shouldn't `b.ResetTimer` also be nested within the sub-benchmarks before the bechnmark loop. e.g.

```go
b.Run("oneResult", func(b *testing.T) {
    ...
    b.ResetTimer()
    for i := o; i < b.N; i++ {
       // benchmark under test
    }
})
```

Otherwise we'll be timer benchmark setup time for the sub-benchmarks runs between `b.Run("oneResult", ...)` and `b.Run("allResults", ...)`.

jon-whit (approved):

PR #1069, opened 2023-10-20
ci: fix benchmark reporting

miparnisari:
## Description

How it works after this PR: on push to `main` we generate and save benchmarks and on every commit to a PR we compare against `main`.

Tested in a personal repo: https://github.com/miparnisari/benchtest/actions/runs/6589243682/job/17903264164?pr=22

image

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -6,7 +6,7 @@ on:
       - main
 
 permissions:
-  contents: read
+  contents: write # to be able to write comments
 
 jobs:
   lint:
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request.yaml
```diff
@@ -6,7 +6,7 @@ on:
       - main
 
 permissions:
-  contents: read
+  contents: write # to be able to write comments
```
miparnisari, line 9: Need to check if this is safe

miparnisari: removed for now. Opened for the future: https://github.com/benchmark-action/github-action-benchmark/issues/197

jpadilla (approved):

adriantam (approved):

Issue #1078, opened 2023-10-24
Changes pagination some requests returning a tuple from previous request


collisj1996:
**Summary**

Testing pagination on the changes endpoint I chose a page size of 2, some of the subsequent requests would have duplicate values from the previous request

- Steps to reproduce
  - Using the Python openfga_sdk
  - Using a page size of 2
  - Using 5 tuples (see below)
  - calling the change endpoint 3 times in succession each using the previous requests continuation token
  - The second request has a duplicate tuple
  - This results in 6 tuples in total with one duplicate
  
- What did you expect to happen?

That each request using the previous request continuation token would have newer distinct tuples

- What actually happened?

Some subsequent request using the previous requests continuation token would include the duplicate values

**OpenFGA version or commit**

Latest OpenFga docker image as of 24th Oct

**Store data**

If applicable, provide information about your authorization model, tuples, and the Check or ListObjects calls you're making. For example:

```yaml
model_file: |
model
  schema 1.1
type user
type domain
  relations
    define member: [user] or admin
    define admin: [user]
    define can_create_folder: member or admin
    define can_create_playlist: member or admin
    define can_view_item_type: member or admin
    define can_view_tag: member or admin
    define can_check_action: member or admin
    define can_create_item_type: admin
    define can_delete_item_type: admin
    define can_rename_item_type: admin
    define can_view_role: admin
    define can_edit_role: admin
    define can_create_tag: admin
    define can_update_tag: admin
    define can_delete_tag: admin
type playlist
  relations
    define can_view: base_can_view and member from domain
    define can_update_description: base_can_update_description and member from domain
    define can_delete: base_can_delete and member from domain
    define can_change_item_type: base_can_change_item_type and member from domain
    define can_rename: base_can_rename and member from domain
    define can_update_items: base_can_update_items and member from domain
    define base_can_view: view from policy or owner
    define base_can_update_description: edit from policy or owner
    define base_can_delete: delete from policy or owner
    define base_can_change_item_type: admin from policy or owner
    define base_can_rename: admin from policy or owner
    define base_can_update_items: admin from policy or owner
    define policy: [policy]
    define owner: [user]
    define domain: [domain]
type folder
  relations
    define can_view_in_list: base_can_view_in_list and member from domain
    define can_view: base_can_view and member from domain
    define can_edit_content: base_can_edit_content and member from domain
    define can_delete: base_can_delete and member from domain
    define can_change_item_type: base_can_change_item_type and member from domain
[...]

PR #1079, opened 2023-10-25
fix: ReadChanges sort order

miparnisari:
## Description
Fixes https://github.com/openfga/openfga/issues/1078.

Currently: when we Write tuples, they all get the same `inserted_at` field. So when calling ReadChanges, we are sorting by `inserted_at` which does not give guarantees on the order. See https://stackoverflow.com/questions/19304579/postgresql-limit-order-returning-inconsistent-results

This PR changes the sort column so that we have a deterministic tie-break.

TO DO:
- [x] if possible, add an automated test that showcases the bug + the fix

## Testing

Before this change:
image

After:

```shell
[24/10/23 5:12:06] ~ $ curl --location 'http://localhost:8080/stores/01HDH8NMN0TCRRQR0VMDRD4VV2/changes?page_size=2' \
--header 'Accept: application/json' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   431  100   431    0     0  28664      0 --:--:-- --:--:-- --:--:-- 71833
{
  "changes": [
    {
      "tuple_key": {
        "object": "file:file_one",
        "relation": "owner",
        "user": "user:user_one"
      },
      "operation": "TUPLE_OPERATION_WRITE",
      "timestamp": "2023-10-24T16:38:54.049097Z"
    },
    {
      "tuple_key": {
        "object": "file:file_two",
        "relation": "owner",
        "user": "user:user_one"
      },
      "operation": "TUPLE_OPERATION_WRITE",
      "timestamp": "2023-10-24T16:38:54.049097Z"
    }
  ],
  "continuation_token": "eyJ1bGlkIjoiMDFIREg4VkJCMEZYUzM1MVZTTlQwVlpESE0iLCJPYmplY3RUeXBlIjoiIn0="
}
[24/10/23 5:13:11] ~ $ curl --location 'http://localhost:8080/stores/01HDH8NMN0TCRRQR0VMDRD4VV2/changes?page_size=2&continuation_token=eyJ1bGlkIjoiMDFIREg4VkJCMEZYUzM1MVZTTlQwVlpESE0iLCJPYmplY3RUeXBlIjoiIn0%3D' \
--header 'Accept: application/json' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   437  100   437    0     0  31623      0 --:--:-- --:--:-- --:--:-- 72833
{
  "changes": [
    {
      "tuple_key": {
        "object": "file:file_three",
        "relation": "owner",
        "user": "user:user_two"
      },
      "operation": "TUPLE_OPERATION_WRITE",
      "timestamp": "2023-10-24T16:38:54.049097Z"
    },
    {
      "tuple_key": {
        "object": "domain:domain_one",
        "relation": "admin",
        "user": "user:user_one"
      },
      "operation": "TUPLE_OPERATION_WRITE",
      "timestamp": "2023-10-24T16:38:54.049097Z"
    }
  ],
  "continuation_token": "eyJ1bGlkIjoiMDFIREg4VkJCMEZYUzM1MVZTTllDN00wQlYiLCJPYmplY3RUeXBlIjoiIn0="
}
```


pkg/storage/postgres/postgres.go
```diff
@@ -612,7 +612,7 @@ func (p *Postgres) ReadChanges(
 		From("changelog").
 		Where(sq.Eq{"store": store}).
 		Where(fmt.Sprintf("inserted_at < NOW() - interval '%dms'", horizonOffset.Milliseconds())).
-		OrderBy("inserted_at asc")
+		OrderBy("ulid asc")
```
jon-whit, line 615: Instead of using `ulid` here, which is not guaranteed to be strictly ordered, we should use the transaction commit timestamp.

See https://pgpedia.info/p/pg_xact_commit_timestamp.html. This is the time the transaction commits, not the time that the transaction starts. We are currently using `NOW()` which returns the time that `NOW()` is called, which could return a timestamp before the time at which the data is actually committed, and thus it could lead to ordering inconsistencies.

miparnisari: Even if i change `inserted_at` to use `pg_xact_commit_timestamp`, it will not fix the problem because every item writen in the transaction would share this value.
image

jon-whit: Gotcha 👍

jon-whit:
I'm going to try and see if I can come up with a test(s).

jon-whit:
@miparnisari please add a test like the following:
https://gist.github.com/jon-whit/bed86b6e46a2ea16128505c17ff56d2d

This produces the problem without the change and demonstrates the fix with the change.

jpadilla (approved):

Issue #1076, opened 2023-10-23
Update `TestServerMetricsReporting`


miparnisari:
There is a test called [`TestServerMetricsReporting`](https://github.com/openfga/openfga/blob/6fcc0752ffcc8918c1a61d6c97f3a7a10c1a94b7/cmd/run/run_test.go#L735). Can you turn on this flag in that test and then assert that one of the metrics is emitted?

_Originally posted by @miparnisari in https://github.com/openfga/openfga/pull/1023#discussion_r1369080675_

PR #1088, opened 2023-10-27
test: modify TestServerMetricsReporting

miparnisari:
## Description
Closes https://github.com/openfga/openfga/issues/1076


pkg/storage/mysql/mysql.go
```diff
@@ -32,6 +32,7 @@ type MySQL struct {
 	stbl                   sq.StatementBuilderType
 	db                     *sql.DB
 	logger                 logger.Logger
+	dbStatsCollector       prometheus.Collector
```
jon-whit, line 35: Why add the collector explicitly here instead of just using the default collector registration?

jpadilla (approved):

jon-whit (changes requested):
What is the problem in https://github.com/openfga/openfga/issues/1076? It's not clear what the objective is with this PR? The reason I ask is that I strongly discourage us from explicitly passing around collectors unless we have a very strong reason to not use default collector registration.

miparnisari:
> What is the problem in https://github.com/openfga/openfga/issues/1076?

It works but we don't have proper tests for it 
image

I wouldn't want to introduce a regression in the publishing of metrics because it'd lead to a gap of valuable data.

> It's not clear what the objective is with this PR? The reason I ask is that I strongly discourage us from explicitly passing around collectors unless we have a very strong reason to not use default collector registration.

We use the default collector registration. But for the tests I needed to properly de-register once done using them because otherwise i would get "attempt to register duplicate collector"

pkg/storage/mysql/mysql.go
miparnisari: We use the default, but I need to deregister after closing to have the tests work properly

cmd/run/run.go
```diff
@@ -465,8 +465,9 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 		s.Logger.Info(fmt.Sprintf("📈 starting metrics server on '%s'", config.Metrics.Addr))
 
 		go func() {
-			http.Handle("/metrics", promhttp.Handler())
-			if err := http.ListenAndServe(config.Metrics.Addr, nil); err != nil {
+			mux := http.NewServeMux()
+			mux.Handle("/metrics", promhttp.Handler())
+			if err := http.ListenAndServe(config.Metrics.Addr, mux); err != nil {
```
miparnisari, line 470: This change prevents

```
2023-10-27T11:07:41.557-0700	INFO	🚀 starting openfga service...	{"version": "dev", "date": "unknown", "commit": "none", "go-version": "go1.21.3"}
panic: http: multiple registrations for /metrics

goroutine 447 [running]:
net/http.(*ServeMux).Handle(0x31cac80, {0x25d3d41, 0x8}, {0x27fd440?, 0xc0005632f0})
	/usr/local/opt/go/libexec/src/net/http/server.go:2530 +0x3bd
net/http.Handle(...)
	/usr/local/opt/go/libexec/src/net/http/server.go:2573
github.com/openfga/openfga/cmd/run.(*ServerContext).Run.func2()
	/Users/mparnisari/GitHub/openfga/cmd/run/run.go:468 +0x88
created by github.com/openfga/openfga/cmd/run.(*ServerContext).Run in goroutine 440
	/Users/mparnisari/GitHub/openfga/cmd/run/run.go:467 +0x33c5
```

jon-whit: Taking a look. There used to be a good prometheus test package to de-register between tests

jon-whit (approved):

PR #1096, opened 2023-11-01
fix: error handling in evaluating conditions

miparnisari:
- Add unit tests

### DIFF
--- a/pkg/condition/eval/eval.go
+++ b/pkg/condition/eval/eval.go
@@ -16,7 +16,10 @@ func TupleConditionMet(
 	tupleCondition := tupleKey.GetCondition()
 	conditionName := tupleCondition.GetName()
 	if conditionName != "" {
-		evaluableCondition := typesys.GetCondition(conditionName)
+		evaluableCondition, ok := typesys.GetCondition(conditionName)
+		if !ok {
+			return false, fmt.Errorf("failed to evaluate relationship condition: condition '%s' was not found", conditionName)
+		}
 		conditionResult, err := evaluableCondition.Evaluate(
 			context,
 			tupleCondition.GetContext().AsMap(),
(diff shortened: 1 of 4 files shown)


jon-whit:
We haven't agreed on this change of behavior yet.

My opinion - we're likely going to be changing our response structure in a way that accommodates exposing more information to the developer about why a Check failed. In the case of ABAC Conditions, we may have one or more paths of eval that are conditional evaluation paths (e.g. dependent upon an ABAC Condition being evaluated) and if at least one path cannot be evaluated due to missing context, then we will want to surface that the outcome is a "conditional" outcome and expose information about which paths were non-evaluable and why (e.g. what were the missing parameters and the expression governing the relationship).

With this ☝️ , it's unlikely we'll ever be surfacing an error because of missing context. IMO the only time we should be surfacing an error explicitly for ABAC Condition evaluation is a) if the expression itself somehow fails to compile or b) the expression is not well formed and/or does not explicitly produce a boolean output. Both a) and b) are validated at the time of WriteAuthorizationModel, so its highly unlikely they will occur, but it is possible corrupt data gets in with a bug in the validation code. Otherwise, if a) or b) are _not_ the case, then I believe that we should be skipping over it for now, and in the very near future handling it with partial evaluation semantics by consuming the missing parameter information and surfacing it to the developer in a structured way if no other evaluation paths are evaluable.

pkg/condition/eval/eval.go
```diff
@@ -16,11 +17,19 @@ func TupleConditionMet(
...
+		}
+
+		// merge both contexts
+		contextSlice := []map[string]interface{}{context}
+		tupleContext := tupleCondition.GetContext()
+		if tupleContext != nil {
+			contextSlice = append(contextSlice, tupleContext.AsMap())
+		}
+
+		conditionResult, err := evaluableCondition.Evaluate(contextSlice...)
```
jpadilla, line 32: fwiw `.Evaluate()` already merges the passed in context maps

miparnisari: That's ok, the change here is to prevent panic when `tupleCondition.GetContext()` is `nil`, since calling `.AsMap(` on it will panic

jon-whit (approved):

PR #1098, opened 2023-11-01
chore: add counter on expired cache entry

adriantam:
## Description
Add counter on number of cache miss due to entry being expired

## Review Checklist

### DIFF
--- a/internal/graph/cached_resolver.go
+++ b/internal/graph/cached_resolver.go
@@ -31,6 +31,11 @@ var (
 		Name: "check_cache_hit_count",
 		Help: "The total number of cache hits for ResolveCheck.",
 	})
+
+	checkCacheExpiredCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Name: "check_cache_expired_count",
+		Help: "The total number of cache misses for ResolveCheck due to entry expired.",
+	})
 )
 
 // CachedResolveCheckResponse is very similar to ResolveCheckResponse except we
@@ -158,6 +163,8 @@ func (c *CachedCheckResolver) ResolveCheck(
 	if cachedResp != nil && !cachedResp.Expired() {
 		checkCacheHitCounter.Inc()
 		return cachedResp.Value().convertToResolveCheckResponse(), nil
+	} else if cachedResp != nil && cachedResp.Expired() {
+		checkCacheExpiredCounter.Inc()
 	}
 
 	resp, err := c.delegate.ResolveCheck(ctx, req)


internal/graph/cached_resolver.go
```diff
@@ -158,6 +163,8 @@ func (c *CachedCheckResolver) ResolveCheck(
 	if cachedResp != nil && !cachedResp.Expired() {
 		checkCacheHitCounter.Inc()
 		return cachedResp.Value().convertToResolveCheckResponse(), nil
+	} else if cachedResp != nil && cachedResp.Expired() {
+		checkCacheExpiredCounter.Inc()
```
jon-whit, line 167: What's the plan for having this metric? What's the intention behind it?

miparnisari: I assume it's to decide whether we need a longer TTL

miparnisari (approved):

internal/graph/cached_resolver.go
miparnisari: Wait but umm.. will `cachedResp != nil && cachedResp.Expired()` return true a lot? Won't the entry be evicted quickly?

jon-whit: But if it's expired it won't be a cache hit anyways. I want to know how this metric is intended to be useful beyond what we have today.

jon-whit (changes requested):
See thread of discussion.

internal/graph/cached_resolver.go
adriantam: The ccache library does not have hooks / counters on eviction.  Thus, we need to observe it indirectly (miss - old) will give us ideas on the number of entries that are either new or evicted.

jon-whit: The cache keys are constantly changing though as a result of LRU, so calculating `(miss - old)` doesn't seem appropriate. Eviction, expiry, and cache misses cannot, and should not, be treated identically.

Lets explore what libraries may be out there that could expose more information on this. For example, one that comes to mind is [`dgraph-io/ristretto`](https://pkg.go.dev/github.com/dgraph-io/ristretto). It's a high performance in-memory cache built for high concurrency (low cache contention). It exposes a method we can register as a callback on cache eviction. See `Config#OnEvict` https://pkg.go.dev/github.com/dgraph-io/ristretto#Config

miparnisari: > The ccache library does not have hooks / counters on eviction. 

https://github.com/karlseguin/ccache/issues/12#issuecomment-586666024 

Why aren't we using `cache.GetDropped`?

jon-whit: @miparnisari that isn't super useful without a background poller ticking at some pre-defined frequency. Seems overly complicated for what we need. I'd still recommend a cache that has a callback for key eviction. That way we can adjust a metric dynamically on the fly in that callback instead of having to do some overly clever background poller.

miparnisari: This is how i'd do it

```go
func NewCachedCheckResolver(delegate CheckResolver, opts ...CachedCheckResolverOpt) *CachedCheckResolver {
	// ....

	go checker.updateRateOfEviction()

	return checker
}

func (c *CachedCheckResolver) updateRateOfEviction() {
	select {
	case c.evictionPollerExit <- struct{}{}:
		return
	default:
		if c.cache != nil {
			total := c.cache.GetDropped()
			checkCacheExpiredCounter.Add(float64(total))
		}
	}
}

func (c *CachedCheckResolver) Close() {
	c.evictionPollerExit <- struct{}{}
	if c.allocatedCache {
		c.cache.Stop()
		c.cache = nil
	}
}
```

jon-whit: You'll need a ticker in there, because you shouldn't be polling infinitely fast, otherwise you're burning tons of CPU cycles. So you'd need to modify the `select` to select similar to:

```go
ticker := time.NewTicker(1 * time.Second)

for {
    select {
    case <- ticker.C:
       c.cache.GetDropped()
       ...
    }
}
```

But again, getting back to my original comment, this polling strategy is inferior as it is more complex and easier to get wrong and burns more cycles potentially unecessarily as opposed to a callback driven paradigm. If a key doesn't get evicted then there's no reason to poll, and yet the poller will still be burning cycles. This is why callback driven eviction is preferable.

miparnisari: Yeah you're right about my code being inefficient. I'm still hesitant about replacing our caching library that has been working OK so far just to add a metric.

jon-whit: We can benchmark it. We may very well find its better, and we'll get a better strategy for capturing the metric. I think it's worth the investigation.

PR #1038, opened 2023-10-03
feat: attribute-based access control

jpadilla:
## Description

- [x]  https://github.com/openfga/openfga/pull/962
- [x] https://github.com/openfga/openfga/pull/999
- [x] https://github.com/openfga/openfga/pull/1031
- [x] https://github.com/openfga/openfga/pull/1018
- [x] https://github.com/openfga/openfga/pull/1108
- [x] https://github.com/openfga/openfga/pull/1114
- [x] https://github.com/openfga/openfga/pull/1115
- [x] https://github.com/openfga/openfga/pull/1162
- [x] https://github.com/openfga/api/pull/98
- [x] https://github.com/openfga/language/pull/86
- [x] https://github.com/openfga/cli/pull/169
- [x] Bounding CEL evaluation complexity using CEL cost estimations - https://github.com/openfga/openfga/pull/1137
    * https://pkg.go.dev/github.com/google/cel-go/cel#CostTracking
    * https://pkg.go.dev/github.com/google/cel-go/cel#CostLimit
    * https://pkg.go.dev/github.com/google/cel-go/cel#ValidateComprehensionNestingLimit
    * https://pkg.go.dev/github.com/google/cel-go/cel#Env.EstimateCost
    * https://pkg.go.dev/github.com/google/cel-go/cel#EvalDetails.ActualCost
- [ ] tests all the things (please add reference to this PR and associated PR with each subtask here)
  - [x] Conditions on direct relationships @jon-whit WIP https://github.com/openfga/openfga/pull/1100
  - [x] Conditions through computed relationships @jon-whit WIP https://github.com/openfga/openfga/pull/1100
  - [x] Conditions defined on tupleset relations (e.g. `define viewer: viewer from parent` with `parent` having a condition). @jon-whit  WIP https://github.com/openfga/openfga/pull/1100
  - [ ] Testing parameter type conversion and error codes returned with them
  - [ ] Expressions with odd character escape sequences
  - [x] Expressions that are very large and/or parameters that are very lage (benchmarks and tests) @miparnisari WIP https://github.com/openfga/openfga/pull/1107
- [x] add context size constraints - https://github.com/openfga/openfga/pull/1163

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
[...]

PR #1108, opened 2023-11-02
chore(abac): return errors instead of deny decision with ABAC conditions

jon-whit:
## Description
These changes modify the server behavior to return errors instead of deny decisions when we attempt to evaluate an ABAC Condition that we cannot fully evaluate. For example,

```
type user

type document
  relations
    define viewer: [user with condxy]

condition condxy(x: int, y: int) {
  x < 10 || y < 10
}

- document:1#viewer@user:jon, {condition: "condxy"}
```

Evaluating `Check(document:1#viewer@user:jon, { context: { } })` would return an error instead of `{allowed: false}`.

The changes also are strict about requiring every parameter of an condition to be specified even if it is not necessary to fully evaluate the condition. We chose to do this to be explicit about the type system. We may change this at a later time. Using the example above,

`Check(document:1#viewer@user:jon, { context: { "x": 10 } })` would return an error, because we want the developer to be explicit about providing `x` and `y` along with the context to evaluate the condition `condxy`.

## References

https://github.com/openfga/openfga/pull/1038

## Review Checklist

### DIFF
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -1,0 +1,78 @@
+package errors
+
+import (
+	"errors"
+	reflectlite "reflect"
+)
+
+// With returns an error that represents top wrapped on top of the base error.
+func With(base, top error) error {
+	if base == nil && top == nil {
+		return nil
+	}
+	if top == nil {
+		return base
+	}
+	if base == nil {
+		return top
+	}
+	return union{error: base, top: top}
+}
+
+type union struct {
+	error
+	top error
+}
+
+func (u union) Is(target error) bool {
+	// Copied from errors.Is, but without iterative unwrapping.
+	// If top doesn't match, errors.Is will Unwrap, which does the right thing.
+	if target == nil {
+		return false
+	}
+
+	isComparable := reflectlite.TypeOf(target).Comparable()
+	if isComparable && u.top == target {
+		return true
+	}
+	if x, ok := u.top.(interface{ Is(error) bool }); ok && x.Is(target) {
+		return true
+	}
+	return false
+}
+
+func (u union) As(target any) bool {
+	// copied from errors.As, but without the iterative unwrapping.
+	// If top doesn't match, errors.Is will Unwrap, which does the right thing.
+
+	if target == nil {
+		panic("errors: target cannot be nil")
+	}
+	val := reflectlite.ValueOf(target)
+	typ := val.Type()
+	if typ.Kind() != reflectlite.Ptr || val.IsNil() {
+		panic("errors: target must be a non-nil pointer")
+	}
+	targetType := typ.Elem()
+	if targetType.Kind() != reflectlite.Interface && !targetType.Implements(errorType) {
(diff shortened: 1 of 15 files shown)


pkg/condition/condition.go
```diff
@@ -82,6 +83,7 @@ func (c *EvaluableCondition) compile() error {
 
 	prgopts := []cel.ProgramOption{
 		cel.EvalOptions(cel.OptPartialEval),
+		cel.EvalOptions(cel.OptExhaustiveEval),
```
jon-whit, line 86: todo(@jon-whit ) : remove. residual

pkg/condition/condition.go
```diff
@@ -94,6 +96,7 @@ func (c *EvaluableCondition) compile() error {
 	}
 
 	c.celEnv = env
+	c.celAst = ast
```
jon-whit, line 99: todo(@jon-whit )): remove. residual

pkg/condition/condition.go
```diff
@@ -163,6 +166,15 @@ func (c *EvaluableCondition) Evaluate(contextMaps ...map[string]any) (Evaluation
 		return emptyEvaluationResult, fmt.Errorf("failed to construct condition partial vars: %v", err)
 	}
 
+	var missingParameters []string
```
jon-whit, line 169: todo(@jon-whit): note this will produce duplicates if the partial bindings cannot be evaluated to produce a `bool`. The bit here https://github.com/openfga/openfga/pull/1108/files#diff-4e95dc26dacdca908d7eb173740544f27a74edf6474f9fbbeaa5461f57089fc7R187 will append to the list in this case and we'll get duplicates.

pkg/condition/condition.go
jpadilla, line 86: > // OptExhaustiveEval causes the runtime to disable short-circuits and track state.

why do we need this?

jon-whit: We don't. I removed it. It was residual.

internal/test/check/check.go
```diff
@@ -2,13 +2,12 @@ package checktest
...
 import (
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
-	"google.golang.org/protobuf/types/known/structpb"
 )
 
 type Assertion struct {
 	Tuple            *openfgav1.TupleKey
 	ContextualTuples []*openfgav1.TupleKey `yaml:"contextualTuples"`
-	Context          *structpb.Struct
+	Context          map[string]interface{}
```
jon-whit, line 10: todo: change this back to `*structpb.Struct`

jpadilla (approved):

internal/errors/errors.go
```diff
@@ -0,0 +1,78 @@
+package errors
```
miparnisari, line 1: @jon-whit this code looks like a copy of https://www.reddit.com/r/golang/comments/ue73jz/comment/i6ob6os/

Can we use the library instead and give attribution? https://github.com/natefinch/wrap

PR #1113, opened 2023-11-03
test: fix timeouts

miparnisari:
## Description

- Prevent builds from failing when tests take longer than 5 minutes to run: https://github.com/openfga/openfga/actions/runs/6738684281/job/18318778494
- Fix flaky test

image


Makefile
```diff
@@ -67,7 +67,7 @@ unit-test: go-generate ## Run unit tests
 			-coverprofile=coverageunit.tmp.out \
 			-covermode=atomic \
 			-count=1 \
-			-timeout=5m \
+			-timeout 0 \
```
adriantam, line 70: Does it mean that the test will never timeout? I feel more comfortable increasing this to a larger value but definitely ends.  Otherwise, we can run into a situation where test will never complete.

miparnisari: The timeout is at the Github action level: https://github.com/openfga/openfga/blob/main/.github/workflows/pull_request.yaml#L28-L30

If we forget to set that, it defaults to 1 hour.

jon-whit: But this is specific to our local boxes too, not just CI. I'd prefer not to have `make unit-test` hang forever in local development.

miparnisari: Yeah true, I can also write a new `make` target that is specific for CI/CD?

But then again in local development we also don't want such a strict timeout right? As we continue to keep adding tests, the possibility of them extending over 5 minutes grows larger. 

And if you introduced a bug somewhere, you will notice because the tests never stop running 🤷‍♀️

jon-whit: Panics terminate the process. If it hangs I have to kill my terminal. We should have a timeout set, and it should be reasonable. If our tests are taking too long, lets fix the root cause.

miparnisari: OK, updated to 10 minutes and opened https://github.com/openfga/openfga/issues/1140.

adriantam (approved):

jon-whit (approved):

PR #1133, opened 2023-11-07
Add env vars and username/password overwrite for migrate tool

### LINKED ISSUE #1125: Migration doesn't support --datastore-username and --datastore-password
**Summary**

- Reproduction steps:
  1. running `openfga migrate` the same way as `openfga run`, namely with --datastore-username and --datastore-password set.
- What did you expect to happen?
  1. Expected migrations to work
- What actually happened?
  1. Migrations failed with `unknown flag: --datastore-password`

**OpenFGA version or commit**: v1.3.5 (via helm chart)

**Store data**

not relevant.

martin31821:
Before, openfga run had the possibility to overwrite the username / password present in the run command, but not in the migrate command.

This commit fixes it by adding the flags and the corresponding behavior to the migrate command. Furthermore it adds the env variable handling from run also to migrate

## References

Fixes #1125 

## Review Checklist

### DIFF
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/go-sql-driver/mysql"
-	_ "github.com/go-sql-driver/mysql"
 	_ "github.com/jackc/pgx/v5/stdlib"
 	"github.com/openfga/openfga/assets"
 	"github.com/pressly/goose/v3"
--- a/cmd/migrate/migrate.go
+++ b/cmd/migrate/migrate.go
@@ -86,20 +86,20 @@ func runMigration(_ *cobra.Command, _ []string) error {
 	}
 
 	// Parse the database uri with url.Parse() and update username/password, if set via flags
-	dbUrl, err := url.Parse(uri)
+	dbURI, err := url.Parse(uri)
 	if err != nil {
 		log.Fatalf("invalid database uri: %v\n", err)
 	}
-	if username == "" && dbUrl.User != nil {
-		username = dbUrl.User.Username()
+	if username == "" && dbURI.User != nil {
+		username = dbURI.User.Username()
 	}
-	if password == "" && dbUrl.User != nil {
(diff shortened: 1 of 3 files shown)


cmd/migrate/migrate.go
```diff
@@ -78,6 +85,22 @@ func runMigration(_ *cobra.Command, _ []string) error {
 		return fmt.Errorf("unknown datastore engine type: %s", engine)
 	}
 
+	// Parse the database uri with url.Parse() and update username/password, if set via flags
+	dbURI, err := url.Parse(uri)
```
jon-whit, line 89: This is specific to the postgres implementation, and does not work for the MySQL variant.

See https://github.com/openfga/openfga/blob/main/pkg/storage/postgres/postgres.go#L44-L68
vs
https://github.com/openfga/openfga/blob/main/pkg/storage/mysql/mysql.go#L42-L56

jon-whit (changes requested):
Left an important comment

cmd/migrate/migrate.go
martin31821: my bad, assumed that mysql would use the same format string, just with mysql:// prefix. 

Moved the code to the datastore specific part and added a case handler for mysql as well.

cmd/migrate/migrate_test.go
jon-whit, line 1: One last thing before approval. Would you please add a test for this new behavior?

There are some example tests for this same thing in `pkg/server/server_test.go`. Specifically, `TestServerWithPostgresDatastoreAndExplicitCredentials` and `TestServerWithMySQLDatastoreAndExplicitCredentials`. We should create a similar test for the migrate command.

jon-whit (changes requested):
One final comment - requested at least 2 tests be added.

cmd/migrate/migrate_test.go
martin31821: Added tests.

jon-whit (approved):

jon-whit:
@martin31821 please rebase on `main` and then we can merge

martin31821:
Updated

jon-whit (approved):

PR #1134, opened 2023-11-07
Healthcheck in `OpenFGA` Dockerfile 

### LINKED ISSUE #1117: Health Check should be in the Dockerfile and not docker-compose
https://github.com/openfga/openfga/blob/0d38ae05f3de40773f6ba6afc66b18e59a5b9651/docker-compose.yaml#L51

Can we move this into the Dockerfile?

Siddhant-K-code:
## Description

This pull request adds a gRPC health probe to the Dockerfile, which will be used to check the health of the container. The healthcheck configuration in `docker-compose.yaml` file has been removed and replaced with the new gRPC health probe. This change ensures that the container is considered healthy only if the gRPC health probe returns a successful response.

## References

fixes #1117

## Review Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,4 +23,8 @@ EXPOSE 3000
 COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.22 /ko-app/grpc-health-probe /user/local/bin/grpc_health_probe
 COPY --from=builder /bin/openfga /openfga
 
+# Healthcheck configuration for the container using grpc_health_probe
+# The container will be considered healthy if the gRPC health probe returns a successful response.
+HEALTHCHECK --interval=5s --timeout=30s --retries=3 CMD ["/usr/local/bin/grpc_health_probe", "-addr=:8081"]
+
 ENTRYPOINT ["/openfga"]
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -48,8 +48,3 @@ services:
       - "8081:8081" #grpc
       - "3000:3000" #playground
       - "2112:2112" #prometheus metrics
-    healthcheck:
-      test: ["CMD", "/usr/local/bin/grpc_health_probe", "-addr=openfga:8081"]
-      interval: 5s
-      timeout: 30s
-      retries: 3
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -11,7 +11,7 @@ services:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=password
     healthcheck:
-      test: [ "CMD-SHELL", "pg_isready -U postgres" ]
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
       interval: 5s
       timeout: 5s
       retries: 5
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -23,4 +23,8 @@ EXPOSE 3000
 COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.22 /ko-app/grpc-health-probe /user/local/bin/grpc_health_probe
 COPY --from=builder /bin/openfga /openfga
 
+# Healthcheck configuration for the container using grpc_health_probe
+# The container will be considered healthy if the gRPC health probe returns a successful response.
+HEALTHCHECK --interval=5s --timeout=30s --retries=3 CMD ["/usr/local/bin/grpc_health_probe", "-addr=:8081"]
```
Siddhant-K-code, line 28: 🤔 Not sure, whether we should define after entrypoint with 

```suggestion
HEALTHCHECK --interval=5s --timeout=30s --retries=3 CMD ["/usr/local/bin/grpc_health_probe", "-addr=openfga:8081"]
```

Or just keep as it is.

Siddhant-K-code: But, Currently this assumes that your `openfga` service is listening on 8081 (as defined on Line 19) and is accessible from within the container at that port. This is the standard and simplest way to set it up, and I would recommend this unless you have a specific reason to use the service name in the health check command.

jon-whit: >  Currently this assumes that your openfga service is listening on 8081 (as defined on Line 19) and is accessible from within the container at that port

That's a big downside to this, and a reason to avoid this IMO. If you can configure your server with whatever host:port you want, then this is counter productive to that. How does docker-compose operate if you set a health check on the docker compose service definition even if the HEALTHCHECK is defined in the dockerfile? Which one overrides?

Siddhant-K-code: > How does docker-compose operate if you set a health check on the docker compose service definition even if the HEALTHCHECK is defined in the dockerfile? Which one overrides?

When it comes to Docker health checks, the settings defined in a `docker-compose.yml` (I need to revert this file change for this) will override the `HEALTHCHECK` instruction in the Dockerfile. This means that if users want to customize the health check when using Docker Compose, they can specify their health check parameters directly in the `docker-compose.yml`, and it'll take precedence over any `HEALTHCHECK` defined in the Dockerfile.

For instance, if the Dockerfile has a health check like so:

```Dockerfile
HEALTHCHECK CMD curl --fail http://localhost/ || exit 1
```

And `docker-compose.yml` contains:

```yaml
healthcheck:
  test: ["CMD", "curl", "--fail", "http://localhost/some-other-path"]
  interval: 1m30s
  timeout: 10s
  retries: 3
  start_period: 40s
```

Then, the health check from `docker-compose.yml` will be used when containers are run with Docker Compose.

This allows for flexibility and lets users tailor the health check to match their deployment configuration, which is particularly useful for services that may have configurable ports and hosts.

jon-whit: Ok, so the summary is as follows:
1. The Dockerfile HEALTHCHECK is the default
2. If a docker-compose service definition defines a health check for the service then it takes precendence over the HEALTHCHECK directive in the built image.
3. For other consuming container orchestration platforms, e.g. Kubernetes, they choose how to interpret the HEALTHCHECK directive from the container. In Kubernetes' case they don't make use of it whatsoever.

Sounds good. Thanks for the info!

jon-whit:
@Siddhant-K-code mind rebasing on main?

Siddhant-K-code:
Also, Let me revert the `docker-compose.yml` change. Brb!

Siddhant-K-code:
@jon-whit We are good to go :shipit: 🚀

jon-whit (approved):

PR #1137, opened 2023-11-08
chore(abac): add CEL evaluation cost enforcement

jon-whit:
## Description
Adds protections for ABAC Condition evaluation by bounding CEL evaluation to a maximum cost factor.

## Review Checklist


pkg/typesystem/typesystem.go
```diff
@@ -190,7 +192,9 @@ func New(model *openfgav1.AuthorizationModel) *TypeSystem {
 
 	uncompiledConditions := make(map[string]*condition.EvaluableCondition, len(model.GetConditions()))
 	for name, cond := range model.GetConditions() {
-		uncompiledConditions[name] = condition.NewUncompiled(cond)
+		uncompiledConditions[name] = condition.NewUncompiled(cond).
+			WithMaxEvaluationCost(defaultMaxEvaluationCost).
```
jon-whit, line 196: Not sure the best composition for this. I'm open to suggestions.

This should be configurable, but in order to make it configurable in a dynamic way with our current `EvaluableCondition` design we have to provide it before we compile the condition. If we rework our `Evaluate` method we could construct the CEL program in the Evaluate method and provide this more dynamic at evaluation time instead of compile time.

It didn't feel too out of place to put this here because we're controlling model behavior, and so having it in the Typesystem package didn't seem too unreasonable, but it doesn't' feel "perfect" to me.

jon-whit:
Just wanted to get this started so we're tracking. Still need to finish tests.

internal/condition/eval/eval_test.go
```diff
@@ -90,3 +92,77 @@ condition correct_ip(ip: string) {
...
 	}
 }
+
+// TestDefaultCELEvaluationCost is used to ensure we don't decreasee the default evaluation cost
+// of CEL expressions, which would break API compability.
+//
+// Critical paths involving ABAC Condition evaluations use the EvaluateTupleCondition function,
+// and so we test that directly to give us higher confidence we're not introducing a compability
+// issue.
+func TestDefaultCELEvaluationCost(t *testing.T) {
```
jon-whit, line 102: The goal with this test was not to try and test CEL itself but just to make sure we don't introduce an API incompatibility with CEL evaluation cost enforcement. So the tests included herein are just to give some confidence around that, and not testing CEL directly. However, in order to construct some of these tests you have to have a deeper understanding of how CEL cost calculation works.

pkg/typesystem/typesystem.go
jpadilla: i see what you mean, but i don't dislike it tbh. NewUncompiled is just a constructor and in that sense this kinda fits in there nicely.

jpadilla (approved):

PR #1151, opened 2023-11-13
perf(cel): construct base CEL environment at package level

jon-whit:
## Description
Eagerly validate type declarations and construct CEL base environment up front instead of per condition. Each `EvaluableCondition` now extends the base env with the type declarations specific to the typed parameters of the condition that the EvaluableCondition defines.

## Review Checklist


internal/condition/condition.go
```diff
@@ -14,6 +14,24 @@ import (
...
+
+func init() {
+	var envOpts []cel.EnvOption
+	for _, customTypeOpts := range types.CustomParamTypes {
+		envOpts = append(envOpts, customTypeOpts...)
+	}
+
+	envOpts = append(envOpts, types.IPAddressEnvOption(), cel.EagerlyValidateDeclarations(true))
+
+	env, err := cel.NewEnv(envOpts...)
```
jpadilla, line 27: haven't fully traced this, but I'm not exactly sure if `EagerlyValidateDeclarations(true)` is doing something here.

https://github.com/google/cel-go/blob/cfbf821f1b458533051306305a39b743db7c4bdb/cel/env.go#L160-L171
https://github.com/google/cel-go/blob/cfbf821f1b458533051306305a39b743db7c4bdb/cel/env.go#L778-L784

jon-whit: It looks like the implementation is odd, but that could be subject to change. Looks like the base std env registers eager validation of declarations, but it may not always. So I think it makes sense to be explicit about it in the env options.

jpadilla (approved):

Issue #871, opened 2023-07-14
Usage of internal?


mxplusb:
There are a lot of core runtime packages for openfga under the `internal` path. Why are those packages there and not part of `pkg`? It seems to be the opposite of "open"

rhamzeh:
Hi @mxplusb 

In general we try to follow the [Go guidelines towards internal](https://go.dev/doc/go1.4#internalpackages), and try not to put too many things outside internal b/c we don't want people to assume that the interfaces are stable.

However if there are particular functionalities you'd like exposed, let us know and we can either move them out as is or work on an interface that we believe is more stable for external use. PRs are also always appreciated, but let us know what you are planning to do before spending the effort, just in case we have other options in mind.

mxplusb:
I’ve worked with Go for a long time and understand `internal`, thanks for the refreshers since it had been awhile 🙂

For my use case, I’m evaluating embedding OpenFGA in my [database](https://github.com/mxplusb/pleiades) so I can provide native FGA functionality at the lowest levels of data operations. The embedding would essentially be a complete rewrite of the OpenFGA runtime architecture to fit into my database’s runtime architecture, and I would also be providing my own storage implementation for the graph. I know I would be taking on a ton of ownership if I went that route, so I was looking through the OpenFGA code base and wanted to ask about the `internal` usages because it felt like stuff was inconsistently placed there.

Ory’s Zanzibar impl, Keto, puts almost everything in `internal` and that really took their implementation off the table for me. I like OpenFGA because it’s less of that, and it feels less “corporate” (even though I know there’s corporate backing for the project). Is y’all’s usage of `internal` focused on safe early iterations of unstable APIs, or is there a long-term plan for putting more stuff in it? If I decide to embed OpenFGA, would I risk losing access to the existing libraries in `pkg` in later releases? Should I create a one-time fork and then port updates to the internal libraries as needed?

Personally, partially because of Keto’s egregiously offensive usage of `internal`, I’m finding the landscape for FGA solutions is very limited, and embedding what exists is very risky. The only path forward for my use case is embedding because I can’t take an external dependency on another database cuz I’m building a database 🤷🏼‍♀️

[...]

PR #1155, opened 2023-11-15
docs: document usage as library (#871)

miparnisari:
Follow-up of https://github.com/openfga/openfga/issues/871


pkg/server/server_test.go
```diff
@@ -45,6 +45,66 @@ func init() {
 	}
 }
 
+func ExampleNewServerWithOpts() {
+	openfga, err := NewServerWithOpts(WithDatastore(memory.New()))
```
adriantam, line 49: The interesting part is here, should we document what options that are typical to be used as library?  Also, do we need to note what sort of middleware we expect user to setup?

miparnisari: > should we document what options that are typical to be used as library

I'm not sure if this is the best place to put that, i don't want to be too verbose, or to repeat documentation. We could point to https://github.com/openfga/openfga/blob/c461c42a3993af58fa2b4c60242ef97cb7765d09/.config-schema.json? I'm open to ideas

> need to note what sort of middleware we expect user to setup?

🤔 You don't need middleware to use it as a library. At one point in the past we realized that if you used it as a library you had to add the gRPC validation middleware, but that is not the case anymore.

README.md
```diff
@@ -19,7 +19,7 @@ OpenFGA is designed to make it easy for developers to model their application pe
...
 
-It offers an [HTTP API](https://openfga.dev/api/service) and a [gRPC API](https://buf.build/openfga/api/file/main:openfga/v1/openfga_service.proto). It has SDKs for [Node.js/JavaScript](https://www.npmjs.com/package/@openfga/sdk), [GoLang](https://github.com/openfga/go-sdk), [Python](https://github.com/openfga/python-sdk) and [.NET](https://www.nuget.org/packages/OpenFga.Sdk). Look in our [Community section](https://github.com/openfga/community#community-projects) for third-party SDKs and tools.
+It offers an [HTTP API](https://openfga.dev/api/service) and a [gRPC API](https://buf.build/openfga/api/file/main:openfga/v1/openfga_service.proto). It has SDKs for [Java](https://central.sonatype.com/artifact/dev.openfga/openfga-sdk), [Node.js/JavaScript](https://www.npmjs.com/package/@openfga/sdk), [GoLang](https://github.com/openfga/go-sdk), [Python](https://github.com/openfga/python-sdk) and [.NET](https://www.nuget.org/packages/OpenFga.Sdk). Look in our [Community section](https://github.com/openfga/community#community-projects) for third-party SDKs and tools. It can also be used as a library (see [example](https://pkg.go.dev/github.com/openfga/openfga/pkg/server#pkg-examples)).
```
aaguiarz, line 22: That 'example' does not look like an example :) but a documentation page for the Server 

Should we have an example somewhere?

miparnisari: Once we merge this PR that link will be automatically updated to include the example I posted here. Similar to https://pkg.go.dev/github.com/openfga/openfga@v1.3.7/pkg/storage#pkg-examples

pkg/server/server_test.go
jon-whit: The godoc should serve as the list 😄 . That's the point.

Focus on good godoc and we'll get good documentation on go.pkg.dev

miparnisari: Are we good to merge as is then?

rhamzeh:
@miparnisari  can we add docuumentation in how to 
* specify the pg store
* override the default config
* preload the in mem store with data?

rhamzeh (approved):

willvedd (approved):

PR #1163, opened 2023-11-20
chore(abac): add context constraints and avoid superfluous context copies

jon-whit:
## Description
Add context size limit enforcement to the Write API and rework the condition evaluation to avoid superfluous copies of and iteration over potentially large context input.

Design points that motivated these changes:

* By the time you're seeing a protobuf structure in the RPC layer its already decoded into memory. The decoding stage has already occurred, and you've already afforded that penalty of decoding a costly `context` input. The best way to reduce the impact of the maximum RPC message size is to set a stricter maximum message size limit on the gRPC server. See [grpc#MaxCallRecvMsgSize](https://pkg.go.dev/google.golang.org/grpc#MaxCallRecvMsgSize). We should consider setting a bit stricter limit on this as the default is 4MB.

* Enforcing any further size constraints on top of the already decoded context parameter would merely be useful to avoid further copying and/or further duplication of the provided context which would "amplify" the memory footprint. Any usages of context.ToMap() or context.ToMap().Clone or copying of the struct, especially in critical paths, would potentially amplify the large input even worse, especially if the copy is a deeper copy of the nested protobuf struct values (e.g. avoid decoding to native Go type `map[string]any`).

* If the transformations above ☝️  are avoided, and selective pruning of the input context is done in favor of cloning or copying the input value, then that is more preferrable as it reduces superfulous mem copies and visitation/decoding of unneeded context values.

* Care should also be taken not to persist arbitrarily large context to the database as well, so we should only be persisting the context needed for the ABAC condition. We currently do this, and we also will limit the size of the context that can be written to the database, because a database row/item has a practical limit.

* A context body can be very lean in structure but very large in size. The concern here would be in the memory footprint of the value (e.g. memory)
  e.g.

```
  {
    "x": "...." // very large string here ....
  }
```

* A context body can be very nested but otherwise lean and small in value size. The concern here would be in the marshaling of deeply nested values (e.g. CPU).
  e.g.

```
  {
    "a": {
        "b": {
            "c": {
                "d": {
                  "e": {
                    "f": {
                      "g": {
                         "x": "1"
                      }
                    }
                  }
                }
            }
        }
    }
  }
```

* Practically speaking the database has a limit on the size of a value in a column (SQL stores) or the size of a single item (DynamoDB). No matter how large we allow the context input to be, the size of the value written to the database cannot exceed that value.

  - postgres column max size is up to 1GB
  - dynamodb item max size is 400KB
  - mysql column max size is 4GB
---
Based on the points above, the changes I have made:
  - Avoid the context size check overhead for queries that included context, e.g. Check, ListObjects, etc.. We focus attention on pruning context needed for evaluation, not on verification of context input constraints. You've already decoded the context if you're computing its size, so you're just adding more latency fuel to the latency fire. To do this I've reworked the usages of context.ToMap(), context.ToMap().Clone, and copy stages of context to avoid superfulous copies of potentially large context. Instead, we do only a shallow copy of the context (map key copy but value pointers move over 1:1), we avoid iteratively converting the structpb Values to native Go types, and we prune the context only to the working map that is needed for evaluation instead of iterating over all provided context keys.

  - For write inputs, we check the context size and make sure it is within the acceptable limits of the most limited implementation (DynamoDB today). The lowest supported value is DynamoDB with 400KB, so recommend a size of about 32KB to start which is well within the limits and still reasonably large, we can always increase it later if needs be.

## References
https://github.com/openfga/openfga/issues/1192

## Review Checklist


jon-whit:
- [ ] todo: add new test structure to `pkg/middleware/validator/validator_test.go`

pkg/middleware/validator/rpc_with_context.go
miparnisari, line 1: suggestion: rename this file to `rpc_with_context_param.go` so that we don't confuse it with the `ctxt` struct

miparnisari (commented):
Re: 

>  todo: wire up Write API to enforce context size limits as well. They should share the same underlying validation code.

Why do you need to modify Write API especially? By putting it in the middleware, all APIs will receive the validation

jon-whit:
> Re:
> 
> > todo: wire up Write API to enforce context size limits as well. They should share the same underlying validation code.
> 
> Why do you need to modify Write API especially? By putting it in the middleware, all APIs will receive the validation

The middleware operates on a request-body structure that assumes `context` is a top-level field. In the Write API `context` is an embedded field per tuple being written. If N tuples are being written, N context inputs have to be validated. That's what I mean..

jpadilla (approved):

PR #1169, opened 2023-11-22
feat: reject conditions in models and tuples

jpadilla:
## Description

For the upcoming v1.3.8 release we want to push ABAC code but without official support for ABAC for this specific release. OpenFGA should continue to operate as it always has, but it will be with the ABAC code changes in place.

The v1.3.8 release will not allow ABAC models or tuples, but it will contain the ABAC code. The purpose of this ticket is to add the code changes to enforce  behavior.

```
< HTTP/1.1 404 Not Found
< Openfga-Store-Id: 01HG8S8A7PXD6Q8NF15DR0S7T6
< Vary: Origin
< X-Request-Id: c403158f-32b1-41eb-996f-916ee2fe4ceb
< Date: Mon, 27 Nov 2023 16:31:12 GMT
< Content-Length: 62
<
{"code":"unimplemented","message":"conditions not supported"}
```

## References

https://github.com/openfga/openfga/pull/1038

## Review Checklist


miparnisari (commented):
Can you post a cURL command showing how the error looks from a client perspective? I'm interested in the HTTP status code

cmd/run/run.go
```diff
@@ -223,6 +223,7 @@ func MustDefaultConfigWithRandomPorts() *serverconfig.Config {
 	// Since this is used for tests, turn the following off:
 	config.Playground.Enabled = false
 	config.Metrics.Enabled = false
+	config.Experimentals = []string{}
```
jon-whit, line 226: This is so that our ABAC tests still work internally?

pkg/server/server.go
```diff
@@ -684,6 +695,12 @@ func (s *Server) WriteAuthorizationModel(ctx context.Context, req *openfgav1.Wri
 		}
 	}
 
+	if slices.Contains(s.experimentals, ExperimentalRejectConditions) {
+		if req.Conditions != nil {
```
jon-whit, line 699: What if a type restriction references a condition but the condition is undefined? What should be the server response behavior there? Should we return the validation error because we reference a condition that is undefined or should we prioritize the `codes.Unimplemented` status code due to conditions not being supported?

Personally, if you've disabled conditions I'm leaning toward the default favoring `codes.Unimplemented`. If we go this route will need a little more validation checking here on the type restrictions.

jpadilla: took a stab at this in [658aafc](https://github.com/openfga/openfga/pull/1169/commits/658aafcf78aa410421834a9ac4577b401aa72ec4)

cmd/run/run.go
jpadilla: Correct, made it more explicit so intent is clear.

jpadilla:
> Can you post a cURL command showing how the error looks from a client perspective? I'm interested in the HTTP status code

@miparnisari its a 404

```
< HTTP/1.1 404 Not Found
< Openfga-Store-Id: 01HG8S8A7PXD6Q8NF15DR0S7T6
< Vary: Origin
< X-Request-Id: c403158f-32b1-41eb-996f-916ee2fe4ceb
< Date: Mon, 27 Nov 2023 16:31:12 GMT
< Content-Length: 62
<
{"code":"unimplemented","message":"conditions not supported"}
```

pkg/server/commands/write_authzmodel.go
```diff
@@ -53,6 +57,10 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 		Conditions:      req.GetConditions(),
 	}
 
+	if w.rejectConditions && model.Conditions != nil {
```
jon-whit, line 60: We'll probably want to guard the case where `model.Conditions` is an empty slice as well (e.g. `[]*openfgav1.RelationshipCondition{}`.

If the conditions are unspecified (e.g. nil) or empty then it should be fine. Tooling may normalize no conditions to an empty slice.

pkg/server/commands/write.go
```diff
@@ -12,24 +12,37 @@ import (
...
-	logger    logger.Logger
-	datastore storage.OpenFGADatastore
+	logger           logger.Logger
+	datastore        storage.OpenFGADatastore
+	rejectConditions bool
 }
 
 // NewWriteCommand creates a WriteCommand with specified storage.TupleBackend to use for storage.
-func NewWriteCommand(datastore storage.OpenFGADatastore, logger logger.Logger) *WriteCommand {
+func NewWriteCommand(datastore storage.OpenFGADatastore, logger logger.Logger, rejectConditions bool) *WriteCommand {
```
jon-whit, line 27: I'd recommend using the options builder pattern here for 2 reasons:

1. We're trying to use it elsewhere. I know this is temporary for the `v1.3.8` release, but it's inline with what we're already doing.

2. We can default `rejectConditions` to `false` so we can avoid all of the boilerplate cleanup in tests after the `v1.3.8` release. If the experimental is set then we set it to true, but otherwise it's defaulted to false.

jon-whit (approved):

jon-whit (approved):

Issue #1104, opened 2023-11-01
Make `typesystem.IsTuplesetRelation` more efficient


miparnisari:
https://github.com/openfga/openfga/blob/f211089a2e40c90ad7342b4e11f6f27bc36a22cc/pkg/typesystem/typesystem.go#L1301-L1314

This is the only usage https://github.com/openfga/openfga/blob/f211089a2e40c90ad7342b4e11f6f27bc36a22cc/pkg/typesystem/typesystem.go#L1324

Can we have `getAllTupleToUsersetsDefinitions` accept `objectType` and void the extra `for` loop?

Additionally, maybe it's more efficient to store this as a field of `typesystem` since it's used to validate within `ValidateTuple` on every single tuple read in a Check or ListObjects call.

PR #1171, opened 2023-11-24
perf: improve tuple validation by precomputing TTUs

miparnisari:
## Description
Closes https://github.com/openfga/openfga/issues/1104

### DIFF
--- a/pkg/typesystem/typesystem.go
+++ b/pkg/typesystem/typesystem.go
@@ -148,7 +148,10 @@ type TypeSystem struct {
 	// [objectType] => typeDefinition
 	typeDefinitions map[string]*openfgav1.TypeDefinition
 	// [objectType] => [relationName] => relation
-	relations     map[string]map[string]*openfgav1.Relation
+	relations map[string]map[string]*openfgav1.Relation
+	// [objectType] => [relationName] => TTU relation
+	ttuRelations map[string]map[string][]*openfgav1.TupleToUserset
+
 	modelID       string
 	schemaVersion string
 }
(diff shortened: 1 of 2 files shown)


pkg/typesystem/typesystem.go
```diff
@@ -177,6 +182,8 @@ func New(model *openfgav1.AuthorizationModel) *TypeSystem {
 			}
 
 			tdRelations[relation] = r
+			ttus := make([]*openfgav1.TupleToUserset, 0)
```
adriantam, line 185: Are we able to know the size and preallocate instead of allocating per slice?

miparnisari: Hmmm we'd have to recursively walk through the relation definition, which we do in the next line. I could preallocate a fixed size like 100, but i don't see much benefit. The problem with not preallocating is that if you add many items, Go will have to copy all elements into bigger and bigger arrays, but here, we won't have many elements. And if we do, we only pay the penalty once when building the model

adriantam: If we don't have many items, that's fine.  I was worried that if we have lots of items, there is a lot of copying / moving elements into larger array.

adriantam (dismissed):
Are we able to measure the perf. difference between the two implementations?

pkg/typesystem/typesystem.go
miparnisari: I mean, we don't have a limit of number of statements in a relation definition, but I can't imagine someone writing a very long line? `define viewer as a1 or a2 or a3 .... or a100`?

miparnisari:
@adriantam it's in the Pr summary

image

The last column is the one that matters, less than 1 means improvement over `main`, more than 1 means regression.

🤔 I need a new benchmark that runs lots of tuple validations, which is where I expect this PR would show an improvement...

miparnisari:
@adriantam  it looks like double as fast based on number of iterations made :D 

```
On main:

[24/11/23 11:43:47] ~/GitHub/openfga (main) $ go test ./internal/validation -bench=BenchmarkValidateTuple  -cpu 1 -benchmem -benchtime 5s -timeout 0 -run=^a
goos: darwin
goarch: amd64
pkg: github.com/openfga/openfga/internal/validation
cpu: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
BenchmarkValidateTuple   1292979              4639 ns/op            1232 B/op         11 allocs/op
PASS
ok      github.com/openfga/openfga/internal/validation  11.159s
[24/11/23 11:45:04] ~/GitHub/openfga (main) $ 


This PR:

[24/11/23 11:45:22] ~/GitHub/openfga (perf-improv) $  go test ./internal/validation -bench=BenchmarkValidateTuple  -cpu 1 -benchmem -benchtime 5s -timeout 0 -run=^a
goos: darwin
goarch: amd64
pkg: github.com/openfga/openfga/internal/validation
cpu: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
BenchmarkValidateTuple   2022118              3015 ns/op             120 B/op          3 allocs/op
PASS
ok      github.com/openfga/openfga/internal/validation  9.404s

```

adriantam (approved):

PR #1176, opened 2023-11-27
refactor(write): modify Write API to match new protobuf definitions

jon-whit:
## Review Checklist

### DIFF
--- a/pkg/storage/sqlcommon/sqlcommon.go
+++ b/pkg/storage/sqlcommon/sqlcommon.go
@@ -311,11 +311,6 @@ func Write(
 		id := ulid.MustNew(ulid.Timestamp(now), ulid.DefaultEntropy()).String()
 		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
 
-		conditionName, conditionContext, err := marshalRelationshipCondition(tk.GetCondition())
-		if err != nil {
-			return err
-		}
-
 		res, err := deleteBuilder.
 			Where(sq.Eq{
 				"store":       store,
@@ -346,7 +341,7 @@ func Write(
 		changelogBuilder = changelogBuilder.Values(
 			store, objectType, objectID,
 			tk.GetRelation(), tk.GetUser(),
-			conditionName, conditionContext,
+			nil, nil,
 			openfgav1.TupleOperation_TUPLE_OPERATION_DELETE,
 			id, dbInfo.sqlTime,
 		)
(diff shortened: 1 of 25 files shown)


pkg/storage/storage.go
```diff
@@ -35,7 +35,7 @@ func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
 
 // Writes and Deletes are typesafe aliases for Write arguments.
 type Writes = []*openfgav1.TupleKey
-type Deletes = []*openfgav1.TupleKey
```
jon-whit, line 38: This update highlights that our storage interface is a bit to coupled with our protobufs. Ideally we'd take some time to refactor our storage interface so that it’s more agnostic of the protobuf layer. We can make better use of interfaces and abstract storage filter structs there to avoid protobuf impacts.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -311,11 +311,6 @@ func Write(
 		id := ulid.MustNew(ulid.Timestamp(now), ulid.DefaultEntropy()).String()
 		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
 
-		conditionName, conditionContext, err := marshalRelationshipCondition(tk.GetCondition())
```
jon-whit, line 314: - [ ] todo: add test for NULL on condition and condition context when a tuple is deleted from the store.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -346,7 +341,7 @@ func Write(
 		changelogBuilder = changelogBuilder.Values(
 			store, objectType, objectID,
 			tk.GetRelation(), tk.GetUser(),
-			conditionName, conditionContext,
+			"", nil,
```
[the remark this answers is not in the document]
jon-whit: Added

jpadilla (approved):

pkg/tuple/tuple.go
```diff
@@ -10,6 +10,18 @@ import (
...
+type TupleWithCondition interface {
+	TupleWithoutCondition
+	GetCondition() *openfgav1.RelationshipCondition
+}
+
+type TupleWithoutCondition interface {
+	GetUser() string
+	GetObject() string
+	GetRelation() string
+	String() string
```
miparnisari, line 22: I don't think you should put `String` here.  I haven't tested it but i think`TupleWithCondition.String()` will delegate to `TupleWithoutCondition.String()`

jon-whit: That's intended. If you're composing tuples a TupleWithCondition is a TupleWithoutCondition + the GetCondition bit. So the TupleWithoutCondition is the base triplet and we have a string method for it. See it's usage. It's only used where we print `(object, relation, user)`, and its default implementation comes from the protobuf generated `String` method fo the `TupleKeyWithoutCondition`.

miparnisari (approved):

PR #1180, opened 2023-11-28
chore: memory storage stores TupleRecords internally

miparnisari:
## Description
Bring these changes over to `main`: 
- https://github.com/openfga/openfga/pull/999/files#diff-76e1d9c2639fc6c407f987f2698d74723cbe632c79cb31a5e0c1140dd33783b7
- https://github.com/openfga/openfga/pull/999/files#diff-bae0ac15d10001bd5dd35974bc2075e71b6dbc7d590a3157b40ded4480ace797

(I removed all the ABAC-related stuff in this PR)

The reason for this change is to lay the groundwork for `feat/abac`. Currently, the `feat/abac` branch has a regression for the memory storage adapter. See https://github.com/openfga/openfga/actions/runs/7009608156/job/19068427402?pr=1038. Specifically, the regression is that we are introducing lots of calls to `AsTuple`:

```shell
git checkout feat/abac
go test -run=XXX -bench=BenchmarkCheckMemory/BenchmarkCheckWithDirectResolution -benchtime 5s -memprofile heapprofileabac.out 
git checkout origin/main
go test -run=XXX -bench=BenchmarkCheckMemory/BenchmarkCheckWithDirectResolution -benchtime 5s -memprofile heapprofilemain.out

go tool pprof -http=localhost:8084 -diff_base heapprofilemain.out heapprofileabac.out
```

image

### DIFF
--- a/pkg/storage/memory/memory.go
+++ b/pkg/storage/memory/memory.go
@@ -27,6 +27,31 @@ type staticIterator struct {
 	mu                sync.Mutex
 }
 
+// Match returns true if all the fields in *TupleRecord are equal to the same field in the target *TupleKey.
+// If the input Object doesn't specify an ID, only the Object Types are compared.
+// If a field in the input parameter is empty, it is ignored in the comparison.
+func Match(t *storage.TupleRecord, target *openfgav1.TupleKey) bool {
+	if target.Object != "" {
+		td, objectid := tupleUtils.SplitObject(target.Object)
+		if objectid == "" {
+			if td != t.ObjectType {
+				return false
+			}
+		} else {
+			if td != t.ObjectType || objectid != t.ObjectID {
+				return false
+			}
+		}
+	}
+	if target.Relation != "" && t.Relation != target.Relation {
+		return false
+	}
+	if target.User != "" && t.User != target.User {
+		return false
+	}
+	return true
+}
+
 func (s *staticIterator) Next() (*openfgav1.Tuple, error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -224,7 +249,7 @@ func (s *MemoryBackend) read(ctx context.Context, store string, tk *openfgav1.Tu
 		copy(matches, s.tuples[store])
 	} else {
 		for _, t := range s.tuples[store] {
-			if t.Match(tk) {
+			if Match(t, tk) {
 				matches = append(matches, t)
 			}
 		}
@@ -270,7 +295,7 @@ func (s *MemoryBackend) Write(ctx context.Context, store string, deletes storage
 Delete:
 	for _, t := range s.tuples[store] {
 		for _, k := range deletes {
-			if t.Match(k) {
+			if Match(t, k) {
 				s.changes[store] = append(s.changes[store], &openfgav1.TupleChange{TupleKey: t.AsTuple().Key, Operation: openfgav1.TupleOperation_TUPLE_OPERATION_DELETE, Timestamp: now})
 				continue Delete
 			}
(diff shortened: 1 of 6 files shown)


pkg/storage/record.go
```diff
@@ -0,0 +1,67 @@
...
+	return &openfgav1.Tuple{
+		Key:       tk,
+		Timestamp: timestamppb.New(t.InsertedAt),
+	}
+}
+
+// Match returns true if all the fields in *TupleRecord are equal to the same field in the target *TupleKey.
+// If the input Object doesn't specify an ID, only the Object Types are compared.
+// If a field in the input parameter is empty, it is ignored in the comparison.
+func (t *TupleRecord) Match(target *openfgav1.TupleKey) bool {
```
jpadilla, line 37: @miparnisari I _really_ think `Match` and `Find` belong in that memory adapter instead of here. We have no other use case for these in other adapters at the moment.

miparnisari: Ok I can move them

pkg/storage/record.go
```diff
@@ -0,0 +1,67 @@
+package storage
```
jon-whit, line 1: This file belongs in `pkg/storage/memory` at this time.

pkg/storage/memory/memory.go
```diff
@@ -21,57 +22,48 @@ import (
...
-	if key.Relation != "" && key.Relation != target.Relation {
-		return false
-	}
-	if key.User != "" && key.User != target.User {
-		return false
-	}
-	return true
-}
-
 func (s *staticIterator) Next() (*openfgav1.Tuple, error) {
```
jon-whit, line 30: It'd be nice if we could get rid of this duplicate `staticIterator` in favor of a usage pattern that uses the other generic static iterator in `pkg/storage/iterators.go`.

Do you see an elegant way to do so? If so, let's go that direction.

miparnisari: I can do that in a separate PR

pkg/storage/memory/memory.go
```diff
@@ -21,28 +22,31 @@ import (
...
 	mu                sync.Mutex
 }
 
-func match(key *openfgav1.TupleKey, target *openfgav1.TupleKey) bool {
-	if key.Object != "" {
-		td, objectid := tupleUtils.SplitObject(key.Object)
+// Match returns true if all the fields in *TupleRecord are equal to the same field in the target *TupleKey.
+// If the input Object doesn't specify an ID, only the Object Types are compared.
+// If a field in the input parameter is empty, it is ignored in the comparison.
+func Match(t *storage.TupleRecord, target *openfgav1.TupleKey) bool {
```
jon-whit, line 33: This shouldn't be exported.

pkg/storage/memory/memory.go
```diff
@@ -274,49 +291,62 @@ func (s *MemoryBackend) Write(ctx context.Context, store string, deletes storage
...
 		}
 	}
 	return nil
 }
 
-func find(tuples []*openfgav1.Tuple, tupleKey *openfgav1.TupleKey) bool {
-	for _, tuple := range tuples {
-		if match(tuple.Key, tupleKey) {
+// Find returns true if there is any *TupleRecord for which storage.Match returns true
+func Find(records []*storage.TupleRecord, tupleKey *openfgav1.TupleKey) bool {
```
jon-whit, line 347: This shouldn't be exported.

pkg/storage/memory/memory.go
miparnisari: Oops 😅

jon-whit (approved):

jpadilla (approved):

PR #1187, opened 2023-11-29
perf: different order of contextual tuples results in cache hit, not miss

miparnisari:
## Description
Closes https://github.com/openfga/openfga/issues/944.

The code here was mostly copied from https://github.com/openfga/openfga/pull/1018 (which targeted feat/abac branch), with a few changes:

- I added more unit tests
- I changed the implementation of the TupleKeyHasher a bit so that it's faster (by not conditionally appending at the end of every tuple)
- I fixed a bug

:) See the improvement below

image

### DIFF
--- a/internal/graph/cache_key_hasher.go
+++ b/internal/graph/cache_key_hasher.go
@@ -1,0 +1,87 @@
+package graph
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/cespare/xxhash/v2"
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+)
+
+type hasher interface {
+	WriteString(value string) error
+}
+
+// CacheKeyHasher implements a key hash using Hash64 for computing cache keys in a stable way.
+type CacheKeyHasher struct {
+	hasher *xxhash.Digest
+}
+
+// NewHasher returns a hasher for string values.
+func NewHasher(xhash *xxhash.Digest) *CacheKeyHasher {
+	return &CacheKeyHasher{hasher: xhash}
+}
+
+// WriteString writes the provided string to the hash.
+func (c *CacheKeyHasher) WriteString(value string) error {
+	_, err := c.hasher.WriteString(value)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *CacheKeyHasher) Key() uint64 {
+	return c.hasher.Sum64()
+}
+
+// NewTupleKeysHasher returns a hasher for an array of *openfgav1.TupleKey.
+// It sorts the tuples first to guarantee that two arrays that are identical except for the ordering
+// return the same hash.
+func NewTupleKeysHasher(tupleKeys ...*openfgav1.TupleKey) *TupleKeysHasher {
+	return &TupleKeysHasher{tupleKeys}
+}
+
+type TupleKeysHasher struct {
+	tupleKeys []*openfgav1.TupleKey
+}
+
+func (t *TupleKeysHasher) Append(h hasher) error {
+	sortedTupleKeys := append([]*openfgav1.TupleKey(nil), t.tupleKeys...) // Copy input to avoid mutating it
+
+	sort.SliceStable(sortedTupleKeys, func(i, j int) bool {
+		if sortedTupleKeys[i].Object > sortedTupleKeys[j].Object {
+			return false
+		}
+
+		if sortedTupleKeys[i].Relation > sortedTupleKeys[j].Relation {
+			return false
+		}
+
+		if sortedTupleKeys[i].User > sortedTupleKeys[j].User {
+			return false
+		}
+
+		return true
+	})
+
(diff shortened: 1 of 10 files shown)


internal/graph/cache_key_hasher.go
jon-whit, line 1: Please move this out a layer into `internal/keys` like it was in the original `feat/abac` PR. The reason for this is it starts to establish a base package for managing stable keys used for various purposes, including cache keys. The `hasher`, and `cacheKeyHasher` impl should not be so specific to the `graph` package.

internal/graph/cache_key_hasher.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/cespare/xxhash/v2"
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+)
+
+type hasher interface {
+	WriteString(value string) error
+}
+
+// CacheKeyHasher implements a key hash using Hash64 for computing cache keys in a stable way.
+type CacheKeyHasher struct {
```
jon-whit, line 16: No need for this to be publicly exported from the containing package.

https://github.com/openfga/openfga/pull/1038/files#diff-7dedee4631643575bf44454c0194c373e5f67dde51a28b4a4ce4b816380e48dbR6-R12

internal/graph/cache_key_hasher.go
```diff
@@ -0,0 +1,84 @@
...
+	hasher *xxhash.Digest
+}
+
+// NewHasher returns a hasher for string values.
+func NewHasher(xhash *xxhash.Digest) *CacheKeyHasher {
```
jon-whit, line 21:
```suggestion
func NewCacheKeyHasher(xhash *xxhash.Digest) *CacheKeyHasher {
```

internal/graph/cache_key_hasher.go
miparnisari: I get this warning if I don't export the struct: 
image

internal/graph/cache_key_hasher.go
miparnisari: I can do that, but then this idea goes against your other comment against exporting. Do we want to be able to use this code from other packages, or not? Right now it's only used for Check, so it makes sense to have it co-located with the Check code.

miparnisari: Also I believe we should do a re-org in another PR :)  `graph` no longer has just `graph` code, it also has just Check code, but the Check code doesn't use `graph.go`, ListObjects uses it (which is in another package)

jon-whit: Create this PR and added a related issue with a `refactor` label:
https://github.com/openfga/openfga/pull/1193

internal/graph/cache_key_hasher.go
jon-whit: It works fine on my end if you move the relevant symbols into `internal/keys` and unexport the publicly exported symbols in favor of the constructors. I'd advise we reintroduce the changes that were originally included in `feat/abac`, because these changes made are honestly just scope creep at the moment.

The hasher and hashable value interfaces exist in their own right without the `graph` package and should be separated as such. Composing hashers together should also be possible, such is the case with the `tupleKeysHasher` and the `cacheKeyHasher`. When `feat/abac` is merged we introduce other hashable values that we can append as another form of a hasher to the hash chain. Point is, different kinds of keys can be hashed using different hashable values. The interface is simple and composable and exhibits the use case.

internal/keys/hasher.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// NewTupleKeysHasher returns a hasher for an array of *openfgav1.TupleKey.
+// It sorts the tuples first to guarantee that two arrays that are identical except for the ordering
+// return the same hash.
+func NewTupleKeysHasher(tupleKeys ...*openfgav1.TupleKey) *TupleKeysHasher {
+	return &TupleKeysHasher{tupleKeys}
+}
+
+type TupleKeysHasher struct {
```
jon-whit, line 21: Lets not export this. Intention is to use the constructor.

```go
func NewTupleKeysHasher(...) *tupleKeysHasher {...}
```

internal/graph/cache_key_hasher.go
miparnisari: > Please move this out a layer into internal/keys

Done 👍

internal/graph/cache_key_hasher.go
miparnisari: Hmm ok. For now it works but the moment we try to use this code from another package, it won't

jon-whit: I'm not sure I understand what you mean. This code already being used outside of the package that defines it and it works fine.

`internal/graph/cached_resolver.go` uses the constructor to construct a `cacheKeyHasher` and then subsequently uses it.

jon-whit (approved):

PR #1191, opened 2023-11-30
fix: inverse experimental flag for conditions

jpadilla:
## Description

In #1169, I introduced the default `reject-conditions` experimental flag. In order to run server and enable conditions you have to do something like `OPENFGA_EXPERIMENTALS="-"` which is clunky.

> Since we default reject-conditions and it’s the only experimental, it’s a negative negative, which means it can never be true.

This PR, changes that in favor of `enable-conditions` experimental flag which is not set by default.

## References

Follow up to https://github.com/openfga/openfga/pull/1169

## Review Checklist

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -80,6 +80,7 @@
                 "type": "string",
                 "enum": ["enable-conditions"]
             },
+            "default": [],
             "x-env-variable": "OPENFGA_EXPERIMENTALS"
         },
         "playground": {
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -78,9 +78,8 @@
             "type": "array",
             "items": {
                 "type": "string",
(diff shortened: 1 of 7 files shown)


jon-whit:
Just a sanity check here - If you attempt to Write a model with conditions when `enable-conditions=false`, then you get a 404 NotFound for the Unimplemented code. This is the default gRPC status code mapping between Unimplemented and the HTTP code. I'm wondering if it makes more sense to return a 400 BadRequest (e.g. INVALID_ARGUMENT code) instead? The endpoint exists, and the functionality is defined, but the input provided is invalid given the server configurations. So it feels a bit more like an input issues, not a server routing or resource missing issue.

WDYT?

Screenshot 2023-11-30 at 12 17 06 PM

PR #1197, opened 2023-11-30
fix: filter out tuples with non-matching conditions from reverse expand

jon-whit:
## Description
Prior to ABAC Conditions we'd only reverse expand edges that were relevant to the model. Consequently, we had no need for similar tuple [filter iterator code](https://github.com/openfga/openfga/blob/main/internal/graph/check.go#L573-L577) like we have in check.

Now that we introduce conditions into the mix, between two models we can have the same relationship edges but with incompatible conditions defined for type restrctions. This means that we must also now filter the tuples yielded by `ReadStartingWithUser`.

These changes fixes this test:
https://github.com/openfga/openfga/blob/13a1dec951623513056a5c894f6ed650d4d1c449/assets/tests/abac_tests.yaml#L589

## Review Checklist

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -426,7 +426,9 @@ func (c *ReverseExpandQuery) readTuplesAndExecute(
 	// filter out invalid tuples yielded by the database iterator
 	filteredIter := storage.NewFilteredTupleKeyIterator(
 		storage.NewTupleKeyIteratorFromTupleIterator(iter),
-		validation.FilterInvalidTuples(c.typesystem),
+		func(tupleKey *openfgav1.TupleKey) bool {
+			return validation.ValidateCondition(c.typesystem, tupleKey) == nil
+		},
 	)
 	defer filteredIter.Stop()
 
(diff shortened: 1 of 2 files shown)


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -421,7 +422,13 @@ func (c *ReverseExpandQuery) readTuplesAndExecute(
 	if err != nil {
 		return err
 	}
-	defer iter.Stop()
+
+	// filter out invalid tuples yielded by the database iterator
+	filteredIter := storage.NewFilteredTupleKeyIterator(
+		storage.NewTupleKeyIteratorFromTupleIterator(iter),
+		validation.FilterInvalidTuples(c.typesystem),
```
miparnisari, line 429: I think that, for safety and for speed, we should only remove tuples that are invalid because of the condition.

```go
func FilterInvalidConditionedTuples(typesys *typesystem.TypeSystem) storage.TupleKeyFilterFunc {
	return func(tupleKey *openfgav1.TupleKey) bool {
		err := ValidateCondition(typesys, tupleKey)
		return err == nil
	}
}
```

Also, in the future i think it's worth revisiting `ReadStartingWithUser` to include another filter by `condition`. This way we avoid the unnecessary DB roundtrips in the first place.

jon-whit: >  in the future i think it's worth revisiting ReadStartingWithUser to include another filter by condition. This way we avoid the unnecessary DB roundtrips in the first place.

Agreed on this point.

> I think that, for safety and for speed, we should only remove tuples that are invalid because of the condition.

Fair point - my goal was to try and keep it more consistent with already tested behavior we depend on in Check, but a quick and dirty check directly on the defined condition is just as good. Will updated.

miparnisari (approved):

PR #1213, opened 2023-12-05
test(storage): add conformance test for normalized Condition behavior

jon-whit:
## Description
Add storage conformance tests for normalization behavior around TupleKeys with conditions and condition context. Namely, if you write a tuple with a `nil` condition it should come back as a nil condition. Conditions are nullable. Likewise,  if you specify a non-empty condition in a tuple we normalize condition context to an empty context if the context is `nil`.

## Review Checklist

### DIFF
--- a/pkg/storage/test/tuples.go
+++ b/pkg/storage/test/tuples.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openfga/openfga/pkg/storage"
 	"github.com/openfga/openfga/pkg/testutils"
 	"github.com/openfga/openfga/pkg/tuple"
+	tupleutils "github.com/openfga/openfga/pkg/tuple"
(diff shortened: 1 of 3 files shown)


pkg/storage/test/tuples.go
```diff
@@ -710,6 +710,152 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		_, err = iter.Next(ctx)
 		require.ErrorIs(t, err, storage.ErrIteratorDone)
 	})
+
+	t.Run("tuples_with_nil_condition", func(t *testing.T) {
```
jon-whit, line 714: Here and in the added test below, we should also cover `ReadPage`.

Issue #1225, opened 2023-12-11
Extend request_duration_by_query_count_ms to list objects


adriantam:
Currently request_duration_by_query_count_ms is only available for [check API](https://github.com/openfga/openfga/blob/96b9c61a5a1616a35d717fc1ad78956cb04193b5/pkg/server/server.go#L66).  We want to extend this to list objects.

PR #1230, opened 2023-12-12
feat: extend request_duration_by_query_count_ms to list objects

adriantam:
## Description
Extend request_duration_by_query_count_ms to list objects.

## References
Close https://github.com/openfga/openfga/issues/1225

## Review Checklist


pkg/server/server.go
```diff
@@ -384,6 +386,12 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 		methodName,
 	).Observe(queryCount)
 
+	requestDurationByQueryHistogram.WithLabelValues(
+		openfgav1.OpenFGAService_ServiceDesc.ServiceName,
+		methodName,
+		utils.Bucketize(uint(*result.ResolutionMetadata.QueryCount), s.requestDurationByQueryHistogramBuckets),
+	).Observe(float64(time.Since(start).Milliseconds()))
```
Siddhant-K-code, line 393: **Suggestion (non-blocking)**: I think, we should define the OpenFGA Desc `serviceName`. As we are re-using it again & again in the whole file (9 occurrences -- not all, but we can optimize for some of them), and we can define it as a constant like we did for `methodName`. Maybe something to refactor to later on, something like:

```go
// Deinfing & extracting service name for clarity 
serviceName := openfgav1.OpenFGAService_ServiceDesc.ServiceName

// Calculate the duration in milliseconds -- Optional suggestion for better readability.
durationMs := float64(time.Since(start).Milliseconds())

requestDurationByQueryHistogram.WithLabelValues(
    serviceName,
    methodName,
    utils.Bucketize(uint(*result.ResolutionMetadata.QueryCount), s.requestDurationByQueryHistogramBuckets),
).Observe(durationMs)

```

pkg/server/server.go
```diff
@@ -384,6 +386,12 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 		methodName,
 	).Observe(queryCount)
 
+	requestDurationByQueryHistogram.WithLabelValues(
+		openfgav1.OpenFGAService_ServiceDesc.ServiceName,
+		methodName,
```
jon-whit, line 391: We really should try to follow the naming standards of the gRPC service definition. Looking at `openfgav1.OpenFGAService_ServiceDesc.Methods`, the names are of the form `Check`, `ListObjects`, etc..

pkg/server/server.go
```diff
@@ -306,6 +306,8 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 }
 
 func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequest) (*openfgav1.ListObjectsResponse, error) {
```
jon-whit, line 308: What of `StreamedListObjects`? We should have the same metric observed in `StreamedListObjects` only under that RPC name instead of `ListObjects`.

pkg/server/server.go
adriantam: openfgav1.OpenFGAService_ServiceDesc.Methods returns array of [MethodDesc](https://pkg.go.dev/google.golang.org/grpc@v1.59.0#MethodDesc), which I don't is what we want.  As for whether "ListObjects" vs. "listobjects", I believe that was brought into by [comment](https://github.com/openfga/openfga/pull/959#discussion_r1310781858) made in the [PR 959](https://github.com/openfga/openfga/pull/959).

pkg/server/server.go
adriantam: Added metrics for streamed list object

pkg/server/server.go
adriantam: Thanks for the suggestion @Siddhant-K-code .  Added common serviceName as suggested

jon-whit (approved):

PR #1231, opened 2023-12-13
fix: record span errors in Check, ListObjects and streaming ListObjects

miparnisari:
## Description
I was testing OpenFGA manually. I noticed that it'd be nice to know at what point in the Check resolution (i.e. node in the tree) we encountered an error. I do in this PR via span errors. If the error is terminal, it propagates to the root span. Otherwise, it only records in the span in which it happened.

(In the specific test I was doing, the Check response was `allowed=true` but I had entered a `context` that would have led to an evaluation error. I couldn't figure out why I wasn't getting the error. Turns out that my model had an `or` and one of the branches led to `allowed=true`.)

This PR should also help us identify if we are doing unnecessary work after an error has happened.

To be honest, I am not sure how much of a performance penalty this will be. It will only add extra processing in the case of errors.

## Testing

Check with a terminal error
image

Check with a non-terminal error
image

ListObjects
image

Streaming ListObjects

image


internal/graph/check.go
```diff
@@ -282,6 +283,9 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 		}
 	}
 
+	if err != nil {
+		telemetry.TraceError(trace.SpanFromContext(ctx), err)
```
jon-whit, line 287: IMO these shouldn't leak outside of the function defining the span itself. If a functionA creates a span and has multiple functions it calls functionB, functionC, etc.. and if any of called functions return an error, then the caller functionA should record the error event.

I believe we should prioritize more `defer` based capture here to avoid code smells having this scattered everywhere.

e.g.

```go
func functionA(ctx context.Context) ... {
    ctx, span := tracer.Start(ctx, "functionA")
    
    var err error

    defer func() {
        span.SetError(err) // only sets the error event if err is non-nil
        span.End()
    }()
   
    err = functionB(ctx)
    err = functionC(ctx)
}
```

miparnisari: I tried this change inside `checkSetOperation`:

```
// ...
		var err error
		var resp *ResolveCheckResponse

		ctx, span := tracer.Start(ctx, reducerKey)
		defer func() {
			span.End()
			if err != nil {
			}
		}()

		resp, err = reducer(ctx, c.concurrencyLimit, handlers...)
		return resp, err
	}
```

But it doesn't propagate all the way to the root span:

image

I'll spend some time later to see if it can be improved

miparnisari: I'm an idiot. I put the `span.RecordError(...)` _after_ the `span.End()`

pkg/telemetry/tracing.go
```diff
@@ -100,7 +101,12 @@ func MustNewTracerProvider(opts ...TracerOption) *sdktrace.TracerProvider {
 	return tp
 }
 
+// TraceError marks the span as having an error, except if the error is context.Canceled,
+// in which case it does nothing.
 func TraceError(span trace.Span, err error) {
+	if errors.Is(err, context.Canceled) {
+		return
+	}
 	span.RecordError(err)
```
jon-whit, line 110: Wondering if this is the right choice? My intuition suggests that it would be good to see the spans which have been cancelled as that will help us identify ongoing bodies of work that were cancelled due to upstream resolution having occurred already.

miparnisari: The reason I added this is because I don't want to pollute the trace with errors when the behavior you mention is a feature of how the code works. I want to mark the span as errored when it is an actual error

jon-whit (approved):

PR #1233, opened 2023-12-14
fix: missing span in ListObjects

miparnisari:
Add missing span:

image

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -364,6 +364,9 @@ func (c *ReverseExpandQuery) readTuplesAndExecute(
 		return ctx.Err()
 	}
 
+	ctx, span := tracer.Start(ctx, "readTuplesAndExecute")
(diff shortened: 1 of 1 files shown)


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -432,6 +432,9 @@ func (c *ReverseExpandQuery) readTuplesAndExecute(
 	)
 	defer filteredIter.Stop()
 
+	ctx, span := tracer.Start(ctx, "readTuplesAndExecute")
```
jon-whit, line 435: What's the rationale for putting this here as opposed to at the top of the function body? It's more standard to put this at the top of the function body, and this is what we've done elsewhere as well.

If there's a rationale reason, it should be commented in the code to indicate as such.

miparnisari: Good point 😆  no reason, updated

jon-whit (approved):

PR #1235, opened 2023-12-14
chore: add prometheus histogram metric for Condition evaluation cost

jon-whit:
## Description
Add a prometheus histogram metric to calculate the distribution of Condition evaluation costs. 

Additionally, I've wired up the `Namespace` field so it reports the project name `openfga`.

## Review Checklist

### DIFF
--- a/cmd/run/run_test.go
+++ b/cmd/run/run_test.go
@@ -870,6 +870,7 @@ func testServerMetricsReporting(t *testing.T, engine string) {
 	require.Contains(t, stringBody, "list_objects_further_eval_required_count")
 	require.Contains(t, stringBody, "list_objects_no_further_eval_required_count")
 	require.Contains(t, stringBody, "go_sql_idle_connections")
+	require.Contains(t, stringBody, "condition_evaluation_cost")
 }
 
 func TestHTTPServerDisabled(t *testing.T) {
--- a/internal/condition/eval/eval.go
+++ b/internal/condition/eval/eval.go
@@ -19,7 +19,7 @@ var conditionEvaluationCostHistogram = promauto.NewHistogram(prometheus.Histogra
 	Namespace:                       build.ProjectName,
 	Name:                            "condition_evaluation_cost",
 	Help:                            "A histogram of the CEL evaluation cost of a Condition in a Relationship Tuple",
-	Buckets:                         utils.LinearBuckets(1, config.DefaultMaxConditionEvaluationCost, 10),
+	Buckets:                         utils.LinearBuckets(0, config.DefaultMaxConditionEvaluationCost, 10),
 	NativeHistogramBucketFactor:     1.1,
 	NativeHistogramMaxBucketNumber:  config.DefaultMaxConditionEvaluationCost,
 	NativeHistogramMinResetDuration: time.Hour,
--- a/internal/condition/eval/eval.go
+++ b/internal/condition/eval/eval.go
@@ -2,12 +2,26 @@ package eval
 
 import (
 	"fmt"
+	"time"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"github.com/openfga/openfga/internal/build"
 	"github.com/openfga/openfga/internal/condition"
 	"github.com/openfga/openfga/pkg/typesystem"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"google.golang.org/protobuf/types/known/structpb"
 )
+
+var conditionEvaluationCostHistogram = promauto.NewHistogram(prometheus.HistogramOpts{
+	Namespace:                       build.ProjectName,
+	Name:                            "condition_evaluation_cost",
+	Help:                            "A histogram of the CEL evaluation cost of a Condition in a Relationship Tuple",
+	Buckets:                         []float64{1, 5, 15, 20, 35, 50, 75, 90, 100},
+	NativeHistogramBucketFactor:     1.1,
+	NativeHistogramMaxBucketNumber:  100,
+	NativeHistogramMinResetDuration: time.Hour,
+})
 
 // EvaluateTupleCondition returns a bool indicating if the provided tupleKey's condition (if any) was met.
 func EvaluateTupleCondition(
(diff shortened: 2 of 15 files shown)


internal/condition/eval/eval.go
```diff
@@ -2,13 +2,27 @@ package eval
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+var conditionEvaluationCostHistogram = promauto.NewHistogram(prometheus.HistogramOpts{
+	Namespace:                       build.ProjectName,
+	Name:                            "condition_evaluation_cost",
+	Help:                            "A histogram of the CEL evaluation cost of a Condition in a Relationship Tuple",
+	Buckets:                         []float64{1, 5, 15, 20, 35, 50, 75, 90, 100},
```
jon-whit, line 20: - [ ] todo: the upper bucket here should be conditionally defined by the max evaluation cost variable. Ideally it shouldn't be hard coded.

internal/condition/eval/eval.go
```diff
@@ -2,13 +2,27 @@ package eval
...
 	"github.com/openfga/openfga/internal/condition"
 	"github.com/openfga/openfga/pkg/typesystem"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+var conditionEvaluationCostHistogram = promauto.NewHistogram(prometheus.HistogramOpts{
+	Namespace:                       build.ProjectName,
+	Name:                            "condition_evaluation_cost",
```
jon-whit, line 18: - [ ] todo: add a test to make sure we're emitting this.

cmd/run/run_test.go
```diff
@@ -870,6 +870,7 @@ func testServerMetricsReporting(t *testing.T, engine string) {
 	require.Contains(t, stringBody, "list_objects_further_eval_required_count")
 	require.Contains(t, stringBody, "list_objects_no_further_eval_required_count")
 	require.Contains(t, stringBody, "go_sql_idle_connections")
```
jon-whit, line 872: Turns out these assertions are a bit misleading. `promauto` automatically registers and emits the zero value for a metric. Even if a metric doesn't have an observed value prometheus will report it. Just by virtue of having

```
var myCounter = promauto.NewCounter(...)
```
the metric will show up. It doesn't mean you ever call `myCounter.Inc()`, for example.

So this test is a bit of a misleading test. It's not actually making sure the metric usage is correct, it's just making sure the metric is somehow defined. Not sure if we care 🤷 ? The problem could be that we break code that stops reporting the value but this test still passes because the metric is still defined but not used correctly.

jon-whit: @miparnisari @jpadilla do you two have an opinion here?

internal/utils/bucket.go
```diff
@@ -12,3 +12,18 @@ func Bucketize(value uint, buckets []uint) string {
 	}
 	return ">" + strconv.Itoa(int(buckets[len(buckets)-1]))
 }
+
+// LinearBuckets returns an evenly distributed range of buckets in the closed interval
+// [min...max]. The min and max count toward the bucket count since they are included
+// in the range.
+func LinearBuckets(min, max float64, count int) []float64 {
```
jon-whit, line 19: I looked into using https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#LinearBuckets but it doesn't allow me to define the closed interval I wish to build the bucket range around. It only allows me to specify a start and width and count of buckets. I wanted to define a start and end and count of buckets and distribute the buckets evenly around that close-interval range.

This is why I added this function, and the function name is inspired by the prometheus one.

cmd/run/run_test.go
miparnisari: This is fine as is

internal/condition/eval/eval.go
```diff
@@ -2,13 +2,29 @@ package eval
...
+	Help:                            "A histogram of the CEL evaluation cost of a Condition in a Relationship Tuple",
+	Buckets:                         utils.LinearBuckets(1, config.DefaultMaxConditionEvaluationCost, 10),
```
adriantam, line 22: Will it be an interesting point for 0 CEL?

internal/server/config/config.go
```diff
@@ -23,10 +23,11 @@ const (
...
 
-	DefaultWriteContextByteLimit = 32 * 1_024 // 32KB
-	DefaultCheckQueryCacheLimit  = 10000
-	DefaultCheckQueryCacheTTL    = 10 * time.Second
-	DefaultCheckQueryCacheEnable = false
+	DefaultWriteContextByteLimit      = 32 * 1_024 // 32KB
+	DefaultCheckQueryCacheLimit       = 10000
+	DefaultCheckQueryCacheTTL         = 10 * time.Second
+	DefaultCheckQueryCacheEnable      = false
+	DefaultMaxConditionEvaluationCost = 100 // care should be taken here - decreasing can cause API compatibility problems with Conditions
```
adriantam, line 30: In this case, will it make sense to have smaller default max condition evaluation cost and increase on an as-needed basis?

internal/condition/eval/eval.go
miparnisari, line 22:
```suggestion
	Buckets:                         utils.LinearBuckets(0, config.DefaultMaxConditionEvaluationCost, 10),
```

since I thought some expressions could have cost 0?

internal/server/config/config.go
adriantam: I see that 100 is used.  So never mind.

internal/condition/eval/eval.go
jon-whit: I assumed that the expression `true` had cost 1. I just tested out some expressions and here's what I found.

```
condition cond1() {
  true
}
```
evaluates to a cost of 0.

```
condition cond2() {
  1 < 2
}
```
evaluates to a cost of 1.

```
condition cond3(x: int) {
  x < 2
}
```
with `x: 1` evaluates to a cost of 2.

etc..

So my assumption was wrong. I'll set the lower bound here to 0 to account for the first case `cond1`.

jon-whit: https://github.com/openfga/openfga/pull/1235#discussion_r1428250705

miparnisari (approved):

PR #1237, opened 2023-12-15
feat: support cancellation/timeouts on condition evaluation

jpadilla:
## Description

- Use `ContextEval` instead of `Eval` https://pkg.go.dev/github.com/google/cel-go/cel#Program.ContextEval
- Set a default of 100 for `InterruptCheckFrequency` https://pkg.go.dev/github.com/google/cel-go/cel#InterruptCheckFrequency

In https://github.com/openfga/openfga/pull/1237/commits/aa117f1db7f93726b2c024caceae4ccc8c8eefb4, I added two temporary benchmark suites to get an idea of how long evaluations might take based on the expression + parameters supplied. Notice the `p99(ns)` column.

```
$ go test ./internal/condition -bench ^BenchmarkEvaluate -test.count=10 -timeout 0 -run=XXX -benchmem
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/internal/condition
BenchmarkEvaluate-10                           	 1497010	       811.8 ns/op	     66392 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1488909	       797.9 ns/op	     62086 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1515388	       793.4 ns/op	     59103 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1519147	       792.3 ns/op	     60268 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1495501	       791.6 ns/op	     61239 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1519246	       793.6 ns/op	     59887 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1508084	       805.4 ns/op	     65923 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1502704	       797.7 ns/op	     63425 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1487602	       795.0 ns/op	     61058 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluate-10                           	 1498320	       799.1 ns/op	     62269 p99(ns)	     440 B/op	       7 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   28660	     38706 ns/op	    331240 p99(ns)	   24792 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31036	     38456 ns/op	    242090 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31383	     38306 ns/op	    227997 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31207	     38479 ns/op	    302734 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31215	     38321 ns/op	    194209 p99(ns)	   24792 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31353	     38387 ns/op	    218893 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31309	     39251 ns/op	   1090966 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31428	     39416 ns/op	   1257040 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   30876	     39960 ns/op	   1729277 p99(ns)	   24791 B/op	     732 allocs/op
BenchmarkEvaluateComprehensionIterations-10    	   31045	     38503 ns/op	    271264 p99(ns)	   24791 B/op	     732 allocs/op
PASS
ok  	github.com/openfga/openfga/internal/condition	36.263s
```

## Review Checklist


internal/condition/condition.go
```diff
@@ -217,7 +219,10 @@ func (e *EvaluableCondition) Evaluate(contextMaps ...map[string]*structpb.Value)
 		missingParameters = append(missingParameters, key)
 	}
 
-	out, details, err := e.celProgram.Eval(activation)
+	evalCtx, cancel := context.WithTimeout(ctx, time.Millisecond)
```
jon-whit, line 222: side note - if we're going to set an aggressive timeout here, then we should have some metrics calculating the percentiles so that we can make a more informed decision what this value should be. I'll capture this work in a follow up PR since I'm focusing on some metrics right now.

However, in the spirit of making a more informed decision here, can we do a short and sweet analysis of how long a CEL evaluation takes in this code path `EvaluableCondition#Evaluate` if we evaluate a condition that is of the maximum evaluation cost (100 by default right now)?

That is, if we evaluate a condition with cost 100 thousands of times, what is the p99 value of that? Then we should use that value here. Otherwise this value may be more aggressive than it should be and could lead to server errors for conditions even if they are well below the 100 cost limit.

jon-whit: You may consider using a TDigest in a benchmark to compute these quantiles/percentiles in a Go benchmark itself.

https://github.com/influxdata/tdigest


That would probs be the quickest/easiest way to get this data quickly.

internal/condition/condition.go
```diff
@@ -181,7 +183,7 @@ func (e *EvaluableCondition) CastContextToTypedParameters(contextMap map[string]
 // context provided. If more than one source of context is provided, and if the
 // keys provided in those context(s) are overlapping, then the overlapping key
 // for the last most context wins.
-func (e *EvaluableCondition) Evaluate(contextMaps ...map[string]*structpb.Value) (EvaluationResult, error) {
+func (e *EvaluableCondition) Evaluate(ctx context.Context, contextMaps ...map[string]*structpb.Value) (EvaluationResult, error) {
```
miparnisari, line 186: You can also add

```
if ctx.Err() != nil {
   return nil, ctx.Err()
}
```

jpadilla: @miparnisari why do you think we should do this here?

internal/condition/condition.go
jpadilla: Updated PR description with benchmark results.

internal/condition/condition.go
jon-whit: In this case though I don’t see the value. We’re not potentially invoking I/O heavy workloads between the top of the function body and where the context cancellation would actually cancel the costly operation `ContextEval` . I wouldn’t encourage it in this scenario.

This opinion differs from the usage of this similar pattern in `internal/graph/check.go` because in that usage we're potentially avoiding several I/O heavy expansion workloads by doing the point in time context cancellation check.

pkg/typesystem/typesystem.go
```diff
@@ -187,8 +189,10 @@ func New(model *openfgav1.AuthorizationModel) *TypeSystem {
 	uncompiledConditions := make(map[string]*condition.EvaluableCondition, len(model.GetConditions()))
 	for name, cond := range model.GetConditions() {
 		uncompiledConditions[name] = condition.NewUncompiled(cond).
-			WithMaxEvaluationCost(config.DefaultMaxConditionEvaluationCost). // care should be taken here - decreasing can cause API compatibility problems
-			WithTrackEvaluationCost()
+			WithTrackEvaluationCost().
+			// care should be taken here - decreasing can cause API compatibility problems
+			WithMaxEvaluationCost(config.DefaultMaxConditionEvaluationCost).
+			WithInterruptCheckFrequency(defaultInterruptCheckFrequency)
```
jon-whit, line 195: Should we expose this from `internal/server/config` as well like the one above? That way server specific configurations/behavior are more centralized.

internal/condition/condition.go
```diff
@@ -301,6 +292,16 @@ func (e *EvaluableCondition) WithMaxEvaluationCost(cost uint64) *EvaluableCondit
 	return e
 }
 
+// WithInterruptCheckFrequency enables CEL evaluation enforcement of the number of iterations
```
jon-whit, line 295:
```suggestion
// WithInterruptCheckFrequency defines the upper limit on the number of iterations within a CEL comprehension to evaluate before CEL will interrupt evaluation and check for cancellation.
```

internal/condition/condition_test.go
```diff
@@ -500,6 +504,120 @@ func TestCastContextToTypedParameters(t *testing.T) {
...
+						TypeName: openfgav1.ConditionParamTypeRef_TYPE_NAME_LIST,
+						GenericTypes: []*openfgav1.ConditionParamTypeRef{
+							{
+								TypeName: openfgav1.ConditionParamTypeRef_TYPE_NAME_INT,
+							},
+						},
+					},
+				},
+			},
+			checkFrequency: 1000,
```
jon-whit, line 584: What if we modify these tests to account for more of the boundary cases? What happens?

e.g. `checkFrequency: 100` and list comprehends over `101` items. `checkFrequency: 100` and list comprehends over `99` items. etc..

Have we tested those cases?

jpadilla: updated these

internal/condition/condition.go
miparnisari: > @miparnisari why do you think we should do this here?

in case the evaluation takes up a lot of CPU or memory, we can prevent it from happening

jon-whit: That's why we're adding `e.celProgram.ContextEval` though. The evaluation of the program is the costly bit, and that's what needs to be interrupted.

miparnisari: Oh lol, never mind

miparnisari (approved):

PR #1239, opened 2023-12-15
fix: resolve exclusion more deterministically in Check resolution

### LINKED ISSUE #1238: Check gives two responses at random if one of the branches of evaluation errors
**Summary**

Check gives two responses, depending on which branch of the `but not` gets evaluated first:
- allowed=false, or
- "failed to evaluate relationship condition: 'condX' - context is missing parameters '[x]'"

**OpenFGA version or commit**

1.4.0

**Store data**

```yaml
model_file: |
model
  schema 1.1
type user
type document
  relations
    define admin: [user:*]
    define viewer: [user with condX] but not admin
condition condX(x: int) {
   x > 100
}
tuples:
  - user: user:*
    relation: admin
    object: document:budget
  - user: user:maria
    relation: viewer
    object: document:budget
    condition: 
      name: condX
tests:
  - name: test-1
    check:
      - user: user:maria
        object: document:budget
        assertions:
          viewer: false
    list_objects:
      - user: user:maria
        type: document
        assertions:
          viewer:
```

jon-whit:
## Description
Resolve non-deterministic resolution of Check queries with relations involving exclusion.

Given a definition `a but not b` and concurrent resolution of both branches `a` and `b`, then the following should occur:
* If `a` resolves to an `{allowed: false}` outcome we should short-circuit and return. No further resolution of `b` is necessary.
* If `b` resolves to an `{allowed: true}` outcome we should short-circuit and return. No further resolution of `a` is necessary.
* If `a` returns an error first, we must wait for `b` to resolve, because if `b` resolves to `{allowed: true}` the outcome should be `{allowed: false}`, because we can resolve this with a deterministic outcome.
* If `b` returns an error first, we must wait for `a` to resolve, because of `a` resolves to `{allowed: false}` the outcome should be `{allowed: false}`, because we can resolve this with a deterministic outcome.

## References
Closes #1238 

## Review Checklist


internal/graph/check.go
```diff
@@ -422,6 +428,11 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 		}
 	}
 
+	_ = errs
```
jon-whit, line 431: Remove after uncommenting the lines above. Uncommenting the lines above should resolve the reported issue.

assets/tests/abac_tests.yaml
```diff
@@ -786,4 +786,43 @@ tests:
               user: user:jon
             context:
               "x": -1.79769313486231570814527423731704356798070e+309
-            errorCode: 2000
\ No newline at end of file
+            errorCode: 2000
+  - name: exclusion_resolves_deterministically
```
miparnisari, line 790: This should not be an integration test. It should be a unit test that runs a reasonable number of times to guarantee that we were not just lucky if it passed

miparnisari (changes requested):

assets/tests/abac_tests.yaml
```diff
@@ -786,4 +786,43 @@ tests:
...
+          type user
+
+          type document
+            relations
+              define admin: [user:*]
+              define viewer: [user with condX] but not admin
+
+          condition condX(x: int) {
+            x < 100
+          }
```
miparnisari, line 805: Can you also add a test to prove that when the error happens in the right hand side of the `but not`, we also get a deterministic response?

Ie.

```
 model
  schema 1.1
 type user
 type document
    relations
       define admin: [user with condX]
        define viewer: [user:*] but not admin
 condition condX(x: int) {
    x < 100
 }
```

assets/tests/abac_tests.yaml
jon-whit: Added a unit test case to the `TestResolveCheckDeterministic` test  in the `internal/graph/check_test.go` file

assets/tests/abac_tests.yaml
jon-whit: Added one here and also a unit style in the `internal/graph/check_test.go` file.

miparnisari: The test in this file will only add to the overall test time and it is useless because it's only run once, so you will gain little to no confidence that the fix worked. I'd keep the unit test only.

internal/graph/check_test.go
```diff
@@ -50,29 +51,89 @@ type document
...
+		require.Nil(t, resp)
 	})
-	require.NoError(t, err)
-	require.True(t, resp.Allowed)
 
-	resp, err = checker.ResolveCheck(ctx, &ResolveCheckRequest{
-		StoreID:            storeID,
-		TupleKey:           tuple.NewTupleKey("document:2", "editor", "user:x"),
-		ResolutionMetadata: &ResolutionMetadata{Depth: 2},
+	t.Run("exclusion_resolves_deterministically", func(t *testing.T) {
```
miparnisari, line 79: This test doesn't seem to be triggering the changed code:

image

internal/graph/check.go
```diff
@@ -411,6 +415,10 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 		}
 	}
 
+	if errs != nil {
```
sergiught, line 418: If you're using `multierror` there's a chance that you don't actually have any errors but `errs` is not nil, so you'll enter this condition erroneously. This is due to the fact that `errs.Errors` could be instantiated as an empty slice. 

To prevent this, use `errs.ErrorOrNil() != nil `. 

However we could also remove this dependency entirely and use https://pkg.go.dev/errors#Join instead in the future.

jon-whit: That would be good 👍

internal/graph/check_test.go
```diff
@@ -50,29 +51,125 @@ type document
...
+
+type document
+  relations
+	define admin: [user:*]
+	define viewer: [user with condX] but not admin
+
+condition condX(x: int) {
+	x < 100
+}
+`).TypeDefinitions
```
miparnisari, line 102:
```suggestion
		model := parser.MustTransformDSLToProto(`model
	schema 1.1
type user

type document
  relations
	define admin: [user:*]
	define viewer: [user with condX] but not admin

condition condX(x: int) {
	x < 100
}
`)```

File: internal/graph/check_test.go [hunk 7]
```diff
@@ -50,29 +51,125 @@ type document
...
+
+		checker := NewLocalChecker(ds)
+
+		ctx := typesystem.ContextWithTypesystem(context.Background(), typesystem.New(
+			&openfgav1.AuthorizationModel{
+				Id:              ulid.Make().String(),
+				TypeDefinitions: typedefs,
```
miparnisari, line 109: Here and in the other test you forgot to add `Conditions:      model.Conditions,`

We have this problem elsewhere in our code, too 😓 

https://github.com/openfga/openfga/blob/cfa2b13611f7509d2d58f51bdff3382aee991585/pkg/server/test/list_objects.go#L578-L582

We should find a way of making this kind of sneaky bug not possible... not sure how, though.

internal/graph/check_test.go
```diff
@@ -50,29 +51,125 @@ type document
...
+				SchemaVersion:   typesystem.SchemaVersion1_1,
+			},
+		))
+
+		for i := 0; i < 2000; i++ {
```
miparnisari, line 161: We should fine-tune this number so that you consistently hit the code that you changed at least one time per test run. In my machine, 2000 iterations is not enough.

You can also make these individual tests parallel

miparnisari (changes requested):

internal/graph/check_test.go
miparnisari: Please see https://github.com/openfga/openfga/pull/1239#discussion_r1439820571

internal/graph/check_test.go
jon-whit: > We should fine-tune this number so that you consistently hit the code that you changed at least one time per test run. In my machine, 2000 iterations is not enough.

How would you propose we do that? There isn't a definitively way to guarantee the probability. What was enough on your machine? 

> You can also make these tests parallel

👍

internal/graph/check_test.go
jon-whit: I suppose the underlying problem is that in our tests we want to pass in an AuthorizationModel that includes a test generated `Id`, but `parser.MustTransformDSLToProto` doesn't generate an `Id` for us. 

It may be a nice helper to have a test utils package in the parser that produces ☝️ . cc'ing @rhamzeh to get his opinion on that idea.

internal/graph/check_test.go
miparnisari: > How would you propose we do that? There isn't a definitively way to guarantee the probability. What was enough on your machine?

Never mind, with the latest changes both lines changed get hit pretty frequently

internal/graph/check_test.go
jon-whit: Added a helper func and refactored to use it.

internal/graph/check.go
jon-whit: Fixed up. Thanks!

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -3008,7 +3008,7 @@ tests:
             relations
               define blocked: [user]
               define owner: [user] but not blocked
-        assertions:
+        listObjectsAssertions:
```
miparnisari, line 3011: Thank you :)

miparnisari (approved):

PR #1240, opened 2023-12-16
style: delegates in check

miparnisari:
This is just something minor that I noticed while doing something else

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -143,19 +143,10 @@ func WithMaxConcurrentReads(limit uint32) LocalCheckerOption {
 	}
 }
 
-func WithDelegate(delegate CheckResolver) LocalCheckerOption {
-	return func(d *LocalChecker) {
-		d.delegate = delegate
-	}
-}
-
 func WithCachedResolver(opts ...CachedCheckResolverOpt) LocalCheckerOption {
 	return func(d *LocalChecker) {
-		cachedCheckResolver := NewCachedCheckResolver(
-			d,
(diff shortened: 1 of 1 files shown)


internal/graph/check.go
```diff
@@ -143,19 +143,10 @@ func WithMaxConcurrentReads(limit uint32) LocalCheckerOption {
...
-}
-
 func WithCachedResolver(opts ...CachedCheckResolverOpt) LocalCheckerOption {
 	return func(d *LocalChecker) {
-		cachedCheckResolver := NewCachedCheckResolver(
-			d,
-			opts...,
-		)
-		d.SetDelegate(cachedCheckResolver)
+		cachedCheckResolver := NewCachedCheckResolver(d, opts...)
```
miparnisari, line 148: I've just noticed how odd this looks. `LocalChecker` can have a pointer to `CachedChecker`.

But `CachedChecker` _must_ receive a Checker, too.

https://github.com/openfga/openfga/blob/9e1f5366ca9ba672f11528c016f3e60a3b997a45/internal/graph/cached_resolver.go#L112-L123

So, these two structs depend on each other.

And then in this test `TestCachedCheckDatastoreQueryCount` we actually never call `NewCachedCheckResolver` 😆  and make assertions on `LocalChecker`.

https://github.com/openfga/openfga/blob/aa0c8162003a389263ee27c6be6731773bda0147/internal/graph/cached_resolver_test.go#L504-L506

jon-whit (approved):

PR #1236, opened 2023-12-14
fix: #935 (log request validation errors)

miparnisari:
## Description

Close https://github.com/openfga/openfga/issues/918 and https://github.com/openfga/openfga/issues/935

Now, all **unary** requests that are invalid (for example, a missing parameter) will be logged in the server.

## Testing

With the tests in this PR + manual tests:

Logs in Main vs this PR:

image

Traces in this PR:

image

Response headers in this PR 

image

PR #1241, opened 2023-12-16
test: assert on response headers (store-id, model-id, request-id)

miparnisari:
Follow-up of https://github.com/openfga/openfga/pull/1236


tests/check/check_test.go
```diff
@@ -43,7 +43,7 @@ func TestCheckMySQL(t *testing.T) {
 	testRunAll(t, "mysql")
 }
 
-func TestCheckLogs(t *testing.T) {
+func TestCheckLogsAndHeaders(t *testing.T) {
```
jon-whit, line 46: Since headers are a second-class citizen of the public API contract, it feels more appropriate that these tests for headers should be part of a functional API test instead of here in the test that asserts certain logs are being emitted from the server.

Can we relocate these headers to a more appropriate location that coincides with the HTTP grpc-gateway suite of functional tests?

cmd/run/run_test.go
```diff
@@ -1166,3 +1167,102 @@ func TestRunCommandConfigIsMerged(t *testing.T) {
...
+
+	for name, test := range testCases {
+		t.Run(name, func(t *testing.T) {
+			req, err := retryablehttp.NewRequest(test.httpVerb, test.httpPath, strings.NewReader(test.httpJsonBody))
+			require.NoError(t, err, "Failed to construct request")
+
+			httpResponse, err := httpClient.Do(req)
+			require.NoError(t, err)
+
+			// These are set in the server RPCs
```
miparnisari, line 1257: To be more precise: these are set here https://github.com/openfga/openfga/blob/9e1f5366ca9ba672f11528c016f3e60a3b997a45/pkg/server/server.go#L981

`ReadAssertions` and `WriteAssertions` are using the function, so they get the header set, too, even though they don't need it because both APIs require a model ID 🤷‍♀️

tests/check/check_test.go
miparnisari: Moved

jpadilla (dismissed):

jpadilla (dismissed):

cmd/run/run_test.go
```diff
@@ -1167,3 +1169,102 @@ func TestRunCommandConfigIsMerged(t *testing.T) {
...
+			require.Len(t, httpResponse.Header["Openfga-Authorization-Model-Id"], 1)
+			require.Equal(t, authorizationModelID, httpResponse.Header["Openfga-Authorization-Model-Id"][0])
+
+			// These are set in middlewares
+			require.Len(t, httpResponse.Header["Openfga-Store-Id"], 1)
```
jon-whit, line 1264: Can we use consts for these instead of hard coding them? We should be using const values from where they are defined, ideally.

pkg/middleware/requestid/requestid.go
```diff
@@ -15,7 +15,7 @@ import (
 const (
 	requestIDCtxKey   = "request-id-context-key"
 	requestIDTraceKey = "request_id"
-	requestIDHeader   = "x-request-id"
+	RequestIDHeader   = "X-Request-Id"
```
jpadilla, line 18: These should get canonicalized at some point, but curious what lead to the change?

miparnisari: There is no functional change. This is how headers are set:

image

cmd/run/run_test.go
```diff
@@ -1167,3 +1173,102 @@ func TestRunCommandConfigIsMerged(t *testing.T) {
 	rootCmd.SetArgs([]string{"run"})
 	require.NoError(t, rootCmd.Execute())
 }
+
+func TestHTTPHeaders(t *testing.T) {
+	cfg := MustDefaultConfigWithRandomPorts()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
sergiught, line 1180: **(non-blocking, suggestion):** I'm noticing we're using defer a lot in tests across the code base. Have we considered switching to `t.Cleanup` instead?

The advantage that it brings is that it's guaranteed to run after the test or benchmark is finished, even if `t.FailNow()`,` t.Fatal()`, `t.SkipNow()`, or `t.Skip() `is called, which is not the case with `defer`.

miparnisari: 🤔  it is called

image

sergiught: Oh that's interesting! Thanks for checking 👍🏻 . I'll probably need to get updated on this. I'll research it a bit more.

sergiught: @miparnisari alright did a bit of digging to brush off some of the wrong assumptions I had on this.

The main benefit is around execution timing: `defer` executes when the function it is called in returns. `t.Cleanup` executes when the test and all its subtests have completed.

So we can think of `t.Cleanup()` as an enhanced and extended version of `defer`. It also documents that the passed functions are for cleanup purposes.

Here's also a good article on the topic: https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests.

pkg/middleware/requestid/requestid.go
jpadilla: oh i wonder if assertions should just be against `response.Header.Get(...)`

jpadilla (approved):

Issue #1045, opened 2023-10-05
Ensure that every API reads the model at most once.


miparnisari:
Task: Add unit tests (if not already existing.. I couldn't find any) for each API that accepts a `modelID` parameter, that use a mock datastore and:

1. Asserts that, when `modelID` is passed, exactly ONE call to `datastore. ReadAuthorizationModel` is made.
1. Asserts that, when `modelID` is not passed, exactly ONE call to `datastore. ReadAuthorizationModel` is made.

PR #1246, opened 2023-12-18
refactor: use typesystem resolver in favor of storage wrapper

jon-whit:
## Description
Drop the storage wrapper that caches models and refactor usage with the cached Typesystem resolver that is used elsewhere.

## References
Closes https://github.com/openfga/openfga/issues/1045

## Review Checklist


pkg/server/commands/read_authzmodel.go
```diff
@@ -36,16 +39,22 @@ func NewReadAuthorizationModelQuery(backend storage.AuthorizationModelReadBacken
...
 
-func (q *ReadAuthorizationModelQuery) Execute(ctx context.Context, req *openfgav1.ReadAuthorizationModelRequest) (*openfgav1.ReadAuthorizationModelResponse, error) {
+func (q *ReadAuthorizationModelQuery) Execute(
+	ctx context.Context,
+	req *openfgav1.ReadAuthorizationModelRequest,
+) (*openfgav1.ReadAuthorizationModelResponse, error) {
 	modelID := req.GetId()
-	azm, err := q.backend.ReadAuthorizationModel(ctx, req.GetStoreId(), modelID)
+
+	typesys, err := q.typesystemResolver.ResolveTypesystem(ctx, req.GetStoreId(), req.GetId(), false)
```
miparnisari, line 48: This is changing the behavior of the API, because `ResolveTypesystem` is also doing model validation, whereas `backend.ReadAuthorizationModel` is not

cmd/run/run.go
```diff
@@ -504,6 +506,14 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 		server.WithRequestDurationByQueryHistogramBuckets(convertStringArrayToUintArray(config.RequestDurationDatastoreQueryCountBuckets)),
 		server.WithMaxAuthorizationModelSizeInBytes(config.MaxAuthorizationModelSizeInBytes),
 		server.WithExperimentals(experimentals...),
+		server.WithTypesystemResolver(
+			typesystem.NewCachedTypesystemResolver(
+				datastore,
+				typesystem.WithTypesystemResolverCache(
+					ccache.New(ccache.Configure[*typesystem.TypeSystem]().MaxSize(int64(config.ModelCacheMaxSizeBytes))),
```
miparnisari, line 513: This reads very weird. 

```go
typesystem.NewCachedTypesystemResolver(
....typesystem.WithTypesystemResolverCache(
```

pkg/typesystem/resolver.go
```diff
@@ -19,73 +19,109 @@ const (
...
+	}
 
-		key := fmt.Sprintf("%s/%s", storeID, modelID)
+	return resolver
+}
 
-		item := cache.Get(key)
-		if item != nil {
-			return item.Value(), nil
+// ResolveTypesystem resolves
```
miparnisari, line 61: We should document here all this function does:

- If model Id is given, try to fetch that model
- If model is not given, try to fetch the latest model for that store
- it validates the model

pkg/server/server.go
```diff
@@ -242,6 +242,12 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
 	}
 }
 
+func WithTypesystemResolver(resolver typesystem.TypesystemResolver) OpenFGAServiceV1Option {
+	return func(s *Server) {
+		s.typesystemResolver = resolver
+	}
+}
+
```
miparnisari, line 250: Why do we give an option here? This should be hidden from consumers, the server cannot function without the typesystem resolver

pkg/server/commands/read_authzmodel.go
jon-whit: Model validation is skipped.

See `, false)` at the end of the function call.

cmd/run/run.go
jon-whit: Constructing a CachedTypesystemResolver and providing a cache to do so.

Would `typesystem.NewCachedTypesystemResolver(typesystem.WithCache(...)` read better?

pkg/server/server.go
jon-whit: The server cannot function without a logger and datastore as well and yet we provide those as overridable options you can provide to the server.

The purpose is to expose ways to inject/override the dependency based on a common contract. If you want to show up and provide a custom logger or datastore or TypesystemResolver that implements the interfaces expected, then that's the caller's choice. The server should provide a way for those to be specified, and if not specified should provide reasonable defaults.

pkg/server/commands/read_authzmodel.go
```diff
@@ -7,13 +7,13 @@ import (
...
-	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/typesystem"
 )
 
 // ReadAuthorizationModelQuery retrieves a single type definition from a storage backend.
 type ReadAuthorizationModelQuery struct {
-	backend storage.AuthorizationModelReadBackend
-	logger  logger.Logger
+	logger             logger.Logger
+	typesystemResolver typesystem.TypesystemResolver
```
miparnisari, line 16: If in the future this command needs to read something else aside from the model from the DB, we will have to undo this change and bring back the `backend` field.

This comment applies to every other command where we make this update.

pkg/server/commands/read_authzmodel.go
miparnisari: Ah sorry my bad.

My concern with this approach is that it's not obvious that this call is doing model validations. A `WithValidate(true)` would be more obvious

pkg/server/server.go
miparnisari: > The server cannot function without a logger and datastore as well and yet we provide those as overridable options you can provide to the server.

Yes but there are multiple possible implementations of a logger or a datastore, but there is only one implementation of the typesystem resolver

pkg/storage/storagewrappers/caching.go
miparnisari, line 1: By removing this file, now every command that is doing `storage.ReadModel` will go strictly to the database and not use the cache. Is this PR going to be ammended so that every command that needs the model will use `typesystemResolver` only?

More importantly... how are we going to ensure that, in the future, someone doesn't mistakenly write `storage.ReadModel`,  and uses the `typesystemResolver` only? The `caching` wrapper here is 100% transparents to the users. The `typesystemResolver` is not.

pkg/typesystem/resolver.go
```diff
@@ -19,73 +19,109 @@ const (
...
-// then the earlier TypeSystem that was constructed will be used.
-//
-// The memoized resolver function is safe for concurrent use.
-func MemoizedTypesystemResolverFunc(datastore storage.AuthorizationModelReadBackend) TypesystemResolverFunc {
-	lookupGroup := singleflight.Group{}
-
-	cache := ccache.New(ccache.Configure[*TypeSystem]())
+type TypesystemResolver interface {
+	ResolveTypesystem(ctx context.Context, storeID, modelID string, validate bool) (*TypeSystem, error)
+}
```
miparnisari, line 26: Why an interface if, today, there is only one possible implementation? What other implementations do you foresee?

miparnisari (changes requested):

cmd/run/run.go
miparnisari: > Constructing a CachedTypesystemResolver and providing a cache to do so.

Then why not simply `NewCachedTypesystemResolver(size)` and

```go
func NewCachedTypesystemResolver(size int) {
    ccache.New(ccache.Configure[*typesystem.TypeSystem]().MaxSize(size))
}
```

?

pkg/server/commands/read_authzmodel.go
jon-whit: What is it specifically that is _too much_? If it's the additional validation bit, then we can work through an alternative strategy around that. Keep in mind that the reason we have `typesystem.NewAndValidate` in the resolver is because of a historical reasoning. At some point we realized we had invalid models because of a bug we found in a prior release, and so we added validation to the model resolver so that we can holistically inherit that functionality whenever and wherever we resolved a model.

The typesystem resolver _may_ be doing too much, and you may be right there, but don't forget the reasoning behind why it's currently doing what it's doing. I think the typesystem resolver still stands as an abstraction. We have a `storeID` and `modelID` and we want to resolve to a `*TypeSystem`. Seems logical to have an interface/signature representing as such.

cmd/run/run.go
jon-whit: Seperation of concerns really. The TypesystemResolver doesn't care about the size of the cache or the kind of cache provided etc.. It just _needs a_ cache of a specific type. Since the server has the size available to it as server-level config, then it felt reasonable to have the server construct the cache it needs to provide the resolver with a cache.

pkg/server/server.go
jon-whit: > but there is only one implementation of the typesystem resolver that I can think of. The typesystem resolver could be thought of as a reusable command.

Not sure I agree here, because the whole intent is that the server and related code that needs a TypeSystem doesn't care how it gets it. It just needs it. The TypesystemResolver abstraction succinctly establishes a contractual mechanism to do so that is shared/common across usages. Implementations of a TypesystemResolver can choose how to resolve a model (e.g. with caching, singleflight, or all of the above), but callers of the resolver ultimately don't care, they just need a TypeSystem.

pkg/server/commands/read_authzmodel.go
jon-whit: > If in the future this command needs to read something else aside from the model from the DB, we will have to undo this change and bring back the backend field.

This is a bottomless point. You could say the same thing of so many other design choices. Future proofing the decision based on what future things this may need to do with a database or not is irrelevant to the composition here at this time. The command for `ReadAuthorizationModel` needs to be able to resolve a TypeSystem. It doesn't care how that is done. It could be done via a database, a file, stdin, or some remote RPC. It doesn't care. What it needs is something that can resolve a TypeSystem, and the TypesystemResolver signature provides that functionality uniquely.

pkg/typesystem/resolver.go
jon-whit: We can drop the interface in favor of the explicit `TypesystemResolverFunc`. They just as easily could be used interchangeably, though the interface allows for more mock generation.

pkg/storage/storagewrappers/caching.go
jon-whit: > Is this PR going to be ammended so that every command that needs the model will use typesystemResolver only?

That was the intent, yeah. Anything that needs to resolve a TypeSystem given a `storeID` and `modelID` would use the shared TypesystemResolver that is created once when we first construct the server.

> More importantly... how are we going to ensure that, in the future, someone doesn't mistakenly write storage.ReadModel, and uses the typesystemResolver only? The caching wrapper here is 100% transparent to the users. The typesystemResolver is not.


How do we know that we're always using the caching wrapper if it's transparent to us? Hopefully we're testing for that if it is so transparent... The same would go for the TypesystemResolver would it not?

miparnisari: I don't think we have a test for that, no. 

The thing about the TypesystemResolver that is good and that the caching wrapper doesn't have is a good name.  We've been always operating with the assumption that models are immutable, so they are very safe to cache indefinitely. The caching wrapper isn't explicit about what it caches and for how long. I guess we could change that.

PR #1248, opened 2023-12-18
chore: add more Condition metrics and specific `metrics` package

jon-whit:
## Description
Add more Condition metrics and refactor said metrics into a common internal condition metrics package.

## Review Checklist


internal/condition/metrics/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+			Help:      "A histogram measuring the compilation time (in milliseconds) of a Condition.",
+			Buckets:   []float64{1, 5, 15, 50, 100, 250, 500, 1000},
+		}),
+
+		evaluationTime: promauto.NewHistogram(prometheus.HistogramOpts{
+			Namespace: build.ProjectName,
+			Subsystem: subsystem,
+			Name:      "condition_evaluation_duration_ms",
+			Help:      "A histogram measuring the evaluation time (in milliseconds) of a Condition.",
+			Buckets:   []float64{0.1, 0.25, 0.5, 1, 5, 15, 50, 100, 250, 500},
```
jon-whit, line 36: We can tune these overtime as we see more practical distributions of them, but given the benchmarks we've already done around evaluation I thought this was a decent distribution range.

internal/condition/metrics/metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		}),
+
+		evaluationCost: promauto.NewHistogram(prometheus.HistogramOpts{
```
jon-whit, line 39: I moved this into this package instead to colocate the Condition metrics into a singular package.

miparnisari (approved):

jpadilla (approved):

jpadilla (approved):

internal/condition/metrics/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+			Name:                            "condition_evaluation_cost",
+			Help:                            "A histogram of the CEL evaluation cost of a Condition in a Relationship Tuple",
+			Buckets:                         utils.LinearBuckets(0, config.DefaultMaxConditionEvaluationCost, 10),
+			NativeHistogramBucketFactor:     1.1,
+			NativeHistogramMaxBucketNumber:  config.DefaultMaxConditionEvaluationCost,
+			NativeHistogramMinResetDuration: time.Hour,
+		}),
+	}
+
+	Metrics = m
```
adriantam, line 51: Do we intend this to be singleton pattern?  If so, should we only overwrite Metrics if it is nil?

jon-whit: `init` is called once upon package initialization, so the intent is the `Metrics` variable gets assigned once at package init, which as you point out is a singleton.

We have similar usage patterns in `internal/condition/condition.go`.

adriantam: In this case, should we only set Metrics only if Metrics is nil?  Otherwise, someone accidentally calling init more than once will cause unexpected behaviour?

miparnisari: `init` is a function that gets called automatically, it's not exported so you can't call it

miparnisari (approved):

PR #1252, opened 2023-12-20
feat: read model succeeds even if model had 0 types

miparnisari:
Pre-requisite: https://github.com/openfga/openfga/issues/1025


jon-whit:
We haven’t fully dropped multi-row models in OpenFGA yet, and so that leaves this functionality a bit challenging since we can’t clearly distinguish between a model that exists but has 0 typedefs vs a model that does not physically exist.

With a single row model we can distinguish this, because if the row does not exist then the model does not exist, but with multi-row models it’s more nuanced.

We should finish the work to converge to single row models in OpenFGA before we go this route don’t you think?

PR #1258, opened 2023-12-21
feat: remove duplicate model caches (#1045)

miparnisari:
## Description
Proposal to close https://github.com/openfga/openfga/issues/1045.  Alternative proposal: https://github.com/openfga/openfga/pull/1246

Major changes:
- Added New internal command `Read Auth Model or Latest`. Rationale:
  - Reusability. This is needed by Check, ListObjects, StreamedListObjects, Expand, Write
  - Unit testability
- Updated `pkg/server/server.go`: Removal of `typesystemResolver` in favor of `Read Auth Model or Latest` + the existing `cachingOpenFGA` decorator. Rationale:
  -  `cachingOpenFGA` uses the well-known decorator design pattern. We've been using this pattern a lot within package `storagewrappers`. Design patterns will make it easier for newcomers (and us) to read the codebase
  - reading a model from the DB, or from the cache, or from the deduplication library `singleflight` will be 100% transparent to callers of `storage.ReadAuthorizationModel`. (Today, you need to remember - which we haven't in some cases [1] - that calling `storage.ReadAuthorizationModel` bypasses the `typesystemresolver` cache)
  -  Separation of concerns. `typesystemResolver` currently has a lot of responsibilities. I've split the concerns to: the new command, the caching decorator, and in `server.logModelUsed`.
- Updated `pkg/storage/storage.go` (breaking change ⚠️ ). `storage.ReadAuthorizationModel` now returns `typesystem` instead of a protobuf `model`. Rationale:
  - Every API that needs a model also needs the `typesystem` out of it (with the only exception of `ReadAuthModel`) and sometimes call `typesystem.Validate` on it. It makes sense to update the interface to accomodate for this. [2] Also, the raw protobuf object cannot be extended [3]

## Pending work

StreamedListObjects is a bit different to all the other APIs, in that it needs to set the `auth_model_id` header multiple times (once per response send). 

- [ ] Communicate from StreamedListObjects to the Server through a channel. The Server can then figure out how to send the individual responses.

## References
[1] See [this cache read](https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/server/server.go#L537-L540) and then [this DB read](https://github.com/openfga/openfga/blob/main/pkg/server/commands/write.go#L90) for one Write API call
[2] https://softwareengineering.stackexchange.com/q/448643/109252
[3] https://protobuf.dev/getting-started/cpptutorial/#parsing-serialization:~:text=model.%20If%20you%20want%20to%20add%20richer%20behavior%20to%20a%20generated%20class%2C%20the%20best%20way%20to%20do%20this%20is%20to%20wrap%20the%20generated%20protocol%20buffer%20class%20in%20an%20application%2Dspecific%20class.


internal/commands/read_authzmodel_or_latest.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+type ReadAuthorizationModelOrLatestQuery struct {
+	backend storage.AuthorizationModelReadBackend
+}
+
+// NewReadAuthorizationModelOrLatestQuery is a reusable command meant for other commands
+// where model ID parameter is optional.
+func NewReadAuthorizationModelOrLatestQuery(backend storage.AuthorizationModelReadBackend) *ReadAuthorizationModelOrLatestQuery {
```
jon-whit, line 22: Existing commands in the `commands` package separate the RPC server definitions and their implementations. This seems to be breaking that mold. We're introducing a command that has no backing RPC and merely using it for shared functionality. 

It's not necessarily a problem, but I'm just pointing out that it breaks the status-quo of what we've been calling "commands". My fear is we're overloading code into commands and so the purpose of the commands in the first place is losing value.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -107,7 +107,7 @@ type UserRef struct {
 }
 
 type ReverseExpandQuery struct {
-	datastore               storage.RelationshipTupleReader
+	datastore               storage.Reader
```
jon-whit, line 110: This expands the scope of what the `ReverseExpandQuery` needs. `ReverseExpandQuery` specifically needs a `RelationshipTupleReader` because it needs to be able to read tuples. It doesn't care about reading authorization model from storage.

If you want to pass a struct that implements the `storage.Reader` interface when constructing one of these queries than that would be fine because it implements the `RelationshipTupleReader` interface, but this type definition here is over scoped.

internal/graph/check.go
```diff
@@ -123,7 +123,7 @@ type checkOutcome struct {
 }
 
 type LocalChecker struct {
-	ds                 storage.RelationshipTupleReader
+	ds                 storage.Reader
```
jon-whit, line 126: Same comment as in https://github.com/openfga/openfga/pull/1258/files#r1441147756

pkg/storage/storagewrappers/boundedconcurrency.go
```diff
@@ -32,17 +32,17 @@ var (
 )
 
 type boundedConcurrencyTupleReader struct {
-	storage.RelationshipTupleReader
+	storage.Reader
```
jon-whit, line 35: Same comment as in https://github.com/openfga/openfga/pull/1258/files#r1441147756 - overly constrictive type restriction here.

pkg/storage/storage.go
```diff
@@ -184,6 +186,11 @@ type ChangelogBackend interface {
 	ReadChanges(ctx context.Context, store, objectType string, paginationOptions PaginationOptions, horizonOffset time.Duration) ([]*openfgav1.TupleChange, []byte, error)
 }
 
+type Reader interface {
```
jon-whit, line 189: Entities we have in FGA:

* Stores
* Assertions
* Models
* Tuples
* Changelog Entries (changes)

`Reader` is ambiguous and not specific enough. `ModelAndTupleReader` maybe?

jon-whit:
Two main differences I see between this PR and the alternative proposal in #1246 include:

* Storage interface has changed to return a `*TypeSystem` specifically instead of an `*openfgav1.AuthorizationModel`. This can be viewed as a positive if you look at it from the perspective of where that code is used most. Most places don't need a model directly they need a TypeSystem that is the helper/wrapper around the model. Or this change can be viewed as a negative if you look at from the perspective of the storage layer is no longer returning values normalized to/from the persistent storage medium, and it also adds the overhead of having to use the TypeSystem abstraction in the storage layer implementations, which is non-ideal but not a deal-breaker necessarily. It does impact existing storage implementations which has a negative side-effect a bit.

* Model resolution has been refactored into a shared `command` and that command is passed around and/or constructed as needed - this seems fine, but it does undermine what we've historically been using `commands` for. I fear the overloaded nature of the commands package would start to get a little unwieldy if we put various shared code bits into the commands package(s). The usage pattern around this new command necessitated a common `storage.Reader` interface which expands the scope of needed signatures for various implementations (easily changeable but with the side-effect of less elegant composition I suppose).

internal/commands/read_authzmodel_or_latest.go
miparnisari: See https://github.com/openfga/openfga/issues/1382

curfew-marathon:
**We're closing this PR and will circle back when the work is prioritized.  Do not delete the branch**

PR #1266, opened 2023-12-22
chore: bump api and update corresponding tests

miparnisari:
## Description

- Bumped API to bring two fixes: https://github.com/openfga/api/pull/130 and https://github.com/openfga/api/pull/131
- Added new API validation tests in `cmd/openfga/main_test.go`
- Test reorganization:
  - Moved existing API validation tests from the YAML file to `cmd/openfga/main_test.go`
  - Moved existing functional tests from `cmd/openfga/main_test.go` to the YAML

`cmd/openfga/main_test.go` should be used to test validations defined in https://github.com/openfga/api


tests/check/check.go
```diff
@@ -73,36 +67,6 @@ func testCheck(t *testing.T, client ClientInterface) {
 	})
 }
 
-func testBadAuthModelID(t *testing.T, client ClientInterface) {
```
miparnisari, line 76: Moved to https://github.com/openfga/openfga/pull/1266/files#diff-13b105826200b5ebee7b7a8308e113e492ce11f13ad0aaf5652fa78fe0a4d416R1012

rhamzeh (dismissed):

cmd/openfga/main_test.go
```diff
@@ -1562,15 +1677,23 @@ func GRPCReadAuthorizationModelTest(t *testing.T, tester OpenFGATester) {
...
 
 			s, ok := status.FromError(err)
 			require.True(t, ok)
 			require.Equal(t, test.output.errorCode.String(), s.Code().String())
 
-			if test.output.errorCode == codes.OK {
-				_ = response // use response for assertions
-				// require.Equal(t, test.output.resp.Allowed, response.Allowed)
+			if s.Code() == codes.OK {
+				require.NoError(t, err)
```
adriantam, line 1696: for the != codes.OK case, should we assert require.Error?

miparnisari: You mean assert on the exact error message? Yes we can do that, but at this moment i don't really care about those, only the response status code (as a first pass).

We can submit a follow up PR to assert on all the error messages

adriantam: No, what I meant is at least that we are returning err

adriantam (approved):

cmd/openfga/main_test.go
miparnisari: Oh! Let me try

adriantam (approved):

jpadilla (approved):

PR #1267, opened 2023-12-22
test: reorganize

miparnisari:
## Description

See historical context in https://github.com/openfga/openfga/pull/83#discussion_r923558659

This PR:

- Updates `Make` target names.
     - `make test` runs all unit + integration/functional + server flag tests
     - `make test-docker-local` runs sanity tests against Docker
- Run `tests/functional_test.go` as part of `make test`, and run them against a server, not against Docker. Rationale:
   - Package location: `tests` is where we have existing integration tests for Check, ListObjects, and WriteModel. I felt it is a good place for all the other functional tests (but we can change this)
   - Code coverage: Functional tests against Docker cannot measure code coverage. Our code coverage percentage has been lower because of this. **New overall coverage is +1%** 🥳 
   - Developer speed: 
      - tests against Docker are hard to debug because (a) need to rebuild the image every time you need to change Go code, (b) hard to put breakpoints.
      - Usually, we run `make test` locally, and not `make functional-test`. If a functional test fails, we will only notice after we push the commit and let CI/CD run it for us, which is slower feedback.
- Dedicate `cmd/openfga/main_test.go` to sanity tests against our Docker image. (Note: i am not sure if this is the best place for this file, we can change this)

`cmd/run/run_test.go` is still dedicated to test the server with its various configuration flags. Example: https://github.com/openfga/openfga/blob/68dfc32b7d8a7bd122dc006be46e4f2f4c2fee58/cmd/run/run_test.go#L190-L197


Dockerfile
```diff
@@ -20,7 +20,7 @@ EXPOSE 8081
 EXPOSE 8080
 EXPOSE 3000
 
-COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.23 /ko-app/grpc-health-probe /user/local/bin/grpc_health_probe
+COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.24 /ko-app/grpc-health-probe /usr/local/bin/grpc_health_probe
```
ahawker, line 23: 👍

jpadilla (approved):

PR #1268, opened 2023-12-22
Implement Concurrent Model Fetching in Database Access

### LINKED ISSUE #1256: Make ReadAuthorizationModels faster for Postgres and MySQL
https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/storage/mysql/mysql.go#L394-L408

https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/storage/postgres/postgres.go#L409-L421

Siddhant-K-code:
**tl;dr;** This PR Implemented concurrency in the `ReadAuthorizationModels` functions for both MySQL and Postgres. This is achieved by introducing a worker pool pattern with a maximum of 5 goroutines, which concurrently process model fetching. The change optimizes performance by parallelizing database reads while maintaining the existing structure for error handling and pagination. This update is expected to reduce response times for queries involving multiple models, thereby improving the efficiency of our database interactions.

fixes #1256

---

**Key Changes:**

1. **Concurrent Goroutines:** In both the `ReadAuthorizationModels` functions for MySQL and Postgres, I have introduced a concurrency pattern that utilizes a fixed number of worker goroutines (set to 5). These workers concurrently fetch authorization models based on the provided model IDs.

2. **Error Handling:** The implementation includes robust error handling within the goroutines. Any error encountered during the fetching process is immediately propagated back, ensuring that the function's behavior remains consistent with its non-concurrent version.

3. **Integration with Existing Logic:** The concurrency implementation is seamlessly integrated with the existing pagination and token management logic. This ensures that the enhancements do not disrupt the existing functionalities but rather complement them.

**Benefits:**

- **Performance Improvement:** By fetching models in parallel, the response time for large queries is expected to be reduced.
- **Resource Efficiency:** The use of a fixed number of goroutines prevents excessive resource usage while still offering the benefits of parallel processing.

---

💭 This enhancement aligns with our ongoing efforts to optimize our database interactions and improve the overall performance of our system. I look forward to any feedback or suggestions on this implementation.

### DIFF
--- a/pkg/storage/mysql/mysql.go
+++ b/pkg/storage/mysql/mysql.go
@@ -391,16 +391,41 @@ func (m *MySQL) ReadAuthorizationModels(ctx context.Context, store string, opts 
 		}
 	}
 
-	// TODO: make this concurrent with a maximum of 5 goroutines. This may be helpful:
-	// https://stackoverflow.com/questions/25306073/always-have-x-number-of-goroutines-running-at-any-time
+	// Initialize channels and worker pool
+	modelIDChan := make(chan string, numModelIDs)
+	modelsChan := make(chan *openfgav1.AuthorizationModel, numModelIDs)
+	errChan := make(chan error, 1)
+
+	// Start a fixed number of goroutines
+	numWorkers := 5
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			for id := range modelIDChan {
+				model, err := m.ReadAuthorizationModel(ctx, store, id)
+				if err != nil {
+					errChan <- err
+					return
+				}
+				modelsChan <- model
+			}
+		}()
+	}
+
+	// Send model IDs to the workers
+	for _, id := range modelIDs[:numModelIDs] {
+		modelIDChan <- id
+	}
+	close(modelIDChan)
+
+	// Collect results from the workers
 	models := make([]*openfgav1.AuthorizationModel, 0, numModelIDs)
-	// We use numModelIDs here to avoid retrieving possibly one extra model.
 	for i := 0; i < numModelIDs; i++ {
-		model, err := m.ReadAuthorizationModel(ctx, store, modelIDs[i])
-		if err != nil {
+		select {
+		case model := <-modelsChan:
+			models = append(models, model)
+		case err := <-errChan:
 			return nil, nil, err
 		}
-		models = append(models, model)
 	}
 
 	return models, token, nil
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -406,16 +406,41 @@ func (p *Postgres) ReadAuthorizationModels(ctx context.Context, store string, op
 		}
 	}
 
-	// TODO: make this concurrent with a maximum of 5 goroutines. This may be helpful:
-	// https://stackoverflow.com/questions/25306073/always-have-x-number-of-goroutines-running-at-any-time
+	// Initialize channels and worker pool
+	modelIDChan := make(chan string, numModelIDs)
+	modelsChan := make(chan *openfgav1.AuthorizationModel, numModelIDs)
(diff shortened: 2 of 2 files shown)


Siddhant-K-code:
Anything i should change more to pass the CI for unit-test? 🤔

pkg/storage/mysql/mysql.go
```diff
@@ -391,16 +391,41 @@ func (m *MySQL) ReadAuthorizationModels(ctx context.Context, store string, opts
...
+
+	// Collect results from the workers
 	models := make([]*openfgav1.AuthorizationModel, 0, numModelIDs)
-	// We use numModelIDs here to avoid retrieving possibly one extra model.
 	for i := 0; i < numModelIDs; i++ {
-		model, err := m.ReadAuthorizationModel(ctx, store, modelIDs[i])
-		if err != nil {
+		select {
+		case model := <-modelsChan:
+			models = append(models, model)
```
miparnisari, line 425: Compared to `main`, does this implementation change the order of the models in the response? I'd assume that in `main` responses are sorted by model_id whereas here they are not.

We should try to keep the response sorted by descending ID. (Personally, I don't think it matters, but we may have customers that rely on the order)

1. Add a test here https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/storage/test/storage.go#L26-L49 
2. Document here https://github.com/openfga/openfga/blob/78422af76964abbfe524002a7dbf1f01e8ec8979/pkg/storage/storage.go#L140-L141

If it's not possible we can always document in the release notes that this is breaking change

miparnisari:
> Anything i should change more to pass the CI for unit-test? 🤔

Our tests are flaky 😢 I personally have not had time to look into why.

pkg/storage/mysql/mysql.go
Siddhant-K-code: Regarding sorting, I think we would need something to check about the concurrency sync is completed or not (i.e, all channels are done or not). For that, we can use `sync.waitGroup` (but it would require a separate import). Or we could the old-school way (like following), then sort it by descending order.

```go
// Initialize channels
    modelIDChan := make(chan string, numModelIDs)
    doneChan := make(chan bool, numModelIDs)
    models := make([]*openfgav1.AuthorizationModel, numModelIDs)

    // Start a fixed number of worker goroutines
    numWorkers := 5
    for i := 0; i < numWorkers; i++ {
        go func() {
            for id := range modelIDChan {
                model, err := m.ReadAuthorizationModel(ctx, store, id)
                if err != nil {
                    errChan <- err
                    return
                }
                for index, modelID := range modelIDs[:numModelIDs] {
                    if modelID == id {
                        models[index] = model
                        break
                    }
                }
                doneChan <- true
            }
        }()
    }

    // Send model IDs to the workers
    for _, id := range modelIDs[:numModelIDs] {
        modelIDChan <- id
    }
    close(modelIDChan)

    // Wait for all goroutines to finish
    for i := 0; i < numModelIDs; i++ {
        select {
        case <-doneChan:
            // One goroutine finished
        case err := <-errChan:
            // Return early on error
            return nil, nil, err
        }
    }

    // Sort models in descending order by their ID
    sort.Slice(models, func(i, j int) bool {
        return models[i].GetId() > models[j].GetId()
    })

    return models, token, nil
```


wdyt @miparnisari ?

miparnisari:
@Siddhant-K-code i forgot that we have this task in our backlog: https://github.com/openfga/openfga/issues/1025.

Since we plan on tackling that work soon, I feel like if we merge this PR we will have to undo it anyways as part of https://github.com/openfga/openfga/issues/1025#issuecomment-1874622118.

I greatly appreciate your contribution, but i'm going to close this one. Feel free to open a similar PR for the issue I mentioned above :)

PR #1278, opened 2024-01-02
test: pass in conditions in tests

miparnisari:
- Until we agree on a more permanent solution https://github.com/openfga/openfga/pull/1239#discussion_r1439888499, this PR tries to ensure that in existing tests we don't forget to pass in the `conditions` when building models by taking advantage of [`testutils.MustTransformDSLToProtoWithID`](https://github.com/openfga/openfga/blob/fa080db300e37f775dff62317239a321a2079409/pkg/testutils/testutils.go#L102)
- Make test `TestHTTPServerDisabled` less flaky when running locally


cmd/run/run_test.go
```diff
@@ -839,9 +839,11 @@ func TestHTTPServerDisabled(t *testing.T) {
 		}
 	}()
 
-	_, err := http.Get("http://localhost:8080/healthz")
```
miparnisari, line 842: This fails locally if you have openfga running on docker 🤣

jon-whit (approved):

PR #1281, opened 2024-01-03
perf: remove redundant map.Clone calls in model validation

miparnisari:
- Remove redundant `map.Clone()` calls inside recursive calls within `hasEntrypoints`. The cloning already happens at the beginning of the function. This should reduce memory usage. https://github.com/openfga/openfga/blob/b7e4e662ae13714baa1d7f5eea2b71c0cdb3059a/pkg/typesystem/typesystem.go#L670-L676
- Pass in server logger to `cachedCheckResolver` logger


pkg/typesystem/typesystem.go
```diff
@@ -770,7 +770,7 @@ func hasEntrypoints(
 
 		loop := false
 		for _, child := range rw.Union.Child {
-			hasEntrypoints, childLoop, err := hasEntrypoints(typedefs, typeName, relationName, child, maps.Clone(visitedRelations))
```
jon-whit, line 773: I need a little more time to think on this one, but I'm hesitant to drop the `maps.Clone` since it avoids sharing visited relations among children of the same immediate ancestor. I know for cycle detection this would be a deal-breaker, but I'm trying to comprehend a scenario where thay _may_ be an issue with entrypoint detection as well.

Can you think of any side-effects off of the top of your head?

miparnisari: In the first line of `hasEntrypoints` we are already cloning the input parameter

jon-whit: > In the first line of hasEntrypoints we are already cloning the input parameter

Right, but that cloned map is shared by all children of the inner union, exclusion, or intersection, which would be mutating each other's maps. That's the concern.

miparnisari: 🤔 I don't understand the concern

jon-whit (approved):

PR #1290, opened 2024-01-05
docs: location of tests

miparnisari:
## Description

Adding documentation for 
- contributors: on where the various kinds of tests should be added, 
- maintainers: some personal notes as to how we got to the current state.


docs/tests.md
```diff
@@ -0,0 +1,27 @@
+## Location of tests
+
+Take the Write API. It has:
+
+1. Unit tests in pkg/server/commands/write_test.go. These mock all the external dependencies. The tests here should test all business logic that is independent of the external dependencies.
```
jon-whit, line 5: I wouldn't say it's necessarily independent of the external dependencies, but any test within these units should be targeting mocked dependencies to hit certain business logic paths.

docs/tests.md
```diff
@@ -0,0 +1,27 @@
...
+2. **Exported** integration tests in pkg/server/test/write.go. These take in a `datastore` dependency. This is so that anyone that writes their own implementation of the datastore can run these tests by doing
```
jon-whit, line 6: I think it's also worth explicitly calling out (in another bullet point) that `pkg/server/test` is a suite of server conformance tests that are publicly exported so that external projects can test the Server with a provided datastore implementation.

jon-whit:
This is going to be a nice addition for contributors and such. I wonder though if it would make more sense to make this more of a formal overview and part of the contributing guidelines instead of more of a free-formed document.

It'd be nice if we had an overview section that outlines the overall overview of our tests. For example, an outline of the difference(s) between unit, functional, and publicly exported tests, for example. A section with test guidance that has a decision chart or something to help contributors decide when/where to write specific kinds of tests would also be helpful. 

I think if we structured/styled this content in more of a guideline instead of free-formed document it may be more valuable to the community.

willvedd:
> I wonder though if it would make more sense to make this more of a formal overview and part of the contributing guidelines instead

Agreed. This could even be the testing section in a `CONTRIBUTORS.md` file. It's already oriented towards contributors.

I'm hesitant to approve as a standalone doc because  its very specific and come with maintenance overhead. I also feat that we'll forget about the when refactors happen.

docs/tests.md
```diff
@@ -0,0 +1,52 @@
...
+    check.RunAllTests(t, client)
+    ```
+
+    The reason these tests live here is because:
+      - the test cases are defined using easy-to-write YAML files. This enables us to write test cases that apply to all query APIs. (If one Check returns `allowed=true`, the corresponding ListObjects call will, depending on server parameters and other factors, also include it in the response.)
+      - since these tests are exported, it allows authors of datastores to get extra confidence that their implementations of some methods related to Read of Tuples are correct.
+
+> NOTE TO MAINTAINERS: ideally, these tests need not be exported, since they are testing for correctness and should be independent of the storage layer. They are the result of not having enough coverage with unit tests.
+
+2. One test that asserts on the log fields that are emmitted: `TestCheckLogs` in tests/check/check_test.go.
```
ttrzeng, line 46: spellcheck: emitted

docs/tests.md
```diff
@@ -0,0 +1,52 @@
...
+
+2. Functional tests in tests/functional_test.go. These are tests for validating the API inputs. The inputs are defined in the proto files (https://github.com/openfga/api).
+3. Some APIs' behavior can be fine-tuned with server flags. If this is the case, the tests for the correct wiring of the flags live in pkg/server/server_test.go.
+4. Some APIs set specific response headers. If this is the case, update the test `TestHTTPHeaders` in cmd/run/run_test.go.
+ 
+### Query APIs (Check, ListObjects, etc.)
+
+Take the Check API. Aside from the tests mentioned in "All APIs", it has:
+
+1. **Exported** integration tests in tests/check/check.go. These take in a `client` dependency. This is so that anyone that writes their own version of the server and/or datastore can run these tests by doing
```
ttrzeng, line 33: spellcheck: that writes -> who writes

ttrzeng (commented):
minor nitpick

adriantam (dismissed):

miparnisari:
> Agreed. This could even be the testing section in a CONTRIBUTORS.md file. It's already oriented towards contributors.

@willvedd the contributors guideline (https://github.com/openfga/.github/blob/main/CONTRIBUTING.md?plain=1) is not specific to the openfga repo. However, that doc says:

> * All changes must have adequate test coverage. For instruction on writing and running the various test suites, see Testing your changes.

I'm not sure what "testing your changes" is supposed to be (a section in that doc? a new doc? where?). I could rename this doc to `testing_your_changes.md`. WDYT?

willvedd (approved):

docs/testing_your_changes.md
```diff
@@ -0,0 +1,52 @@
+## Location of tests
```
jon-whit, line 1: Can we put this under a top-level "Testing Guidelines" header please?

docs/testing_your_changes.md
```diff
@@ -0,0 +1,52 @@
...
+
+OpenFGA strives to maintain good code coverage with tests so that the possibility of regressions is reduced.
+
+The purpose of this guide is to inform contributors where to write new tests.
+
+## 1. Storage layer
+
+1. Any change to the OpenFGA storage interface must have an integration test in pkg/storage/test/storage.go. This is so that anyone writing their own storage adapter can have confidence that it is correct.
```
jon-whit, line 9:
```suggestion
1. Any change to the OpenFGA storage interface must have an integration test in `pkg/storage/test/storage.go`. This is so that anyone writing their own storage adapter can have confidence that it is correct.
```

docs/testing_your_changes.md
```diff
@@ -0,0 +1,52 @@
...
+
+## 2. API layer
+
+1. All APIs use an underlying `command` object. These commands must have
+   1. unit tests. For example, Write API is backed by Write Command, which has unit tests in pkg/server/commands/write_test.go. These should target mocked dependencies to cover the majority of the business logic. If you are changing business logic, add a test here.
+   2. **Exported** integration tests in pkg/server/test/server.go. These take in an unmocked `datastore` dependency. The goal of these is to test that the command and the underlying datastore work well. This is so that anyone who writes their own implementation of the datastore can run these tests by doing
```
jon-whit, line 15:
```suggestion
   2. **Exported** integration tests in `pkg/server/test/server.go`. These take in an unmocked `datastore` dependency. The goal of these is to test that the command and the underlying datastore work well. This is so that anyone who writes their own implementation of the datastore can run these tests by doing
```

docs/testing_your_changes.md
```diff
@@ -0,0 +1,52 @@
...
+
+   ```go
+    import openfgatest "github.com/openfga/openfga/pkg/server/test"
+    
+    openfgatest.RunAllTests(t, datastore)
+    ```
+
+      > NOTE TO MAINTAINERS: ideally, these integration tests shouldn't exist. They are the result of not having enough coverage at the storage interface layer.
+
+2. Functional tests in tests/functional_test.go. These are tests for validating the API inputs. The inputs are defined in the proto files (https://github.com/openfga/api).
```
jon-whit, line 25:
```suggestion
2. Functional tests in `tests/functional_test.go`. These are tests for validating the API inputs and functional API behavior. The inputs are defined in the proto files (https://github.com/openfga/api).
```

docs/testing_your_changes.md
```diff
@@ -0,0 +1,52 @@
...
+1. **Exported** integration tests in tests/check/check.go. These take in a `client` dependency. This is so that anyone who writes their own version of the server and/or datastore can run these tests by doing
+    ```go
+    import "github.com/openfga/openfga/tests/check"
+    
+    check.RunAllTests(t, client)
+    ```
+
+    The reason these tests live here is because:
+      - the test cases are defined using easy-to-write YAML files. This enables us to write test cases that apply to all query APIs. (If one Check returns `allowed=true`, the corresponding ListObjects call will, depending on server parameters and other factors, also include it in the response.)
+      - since these tests are exported, it allows authors of datastores to get extra confidence that their implementations of some methods related to Read of Tuples are correct.
```
jon-whit, line 42:
```suggestion
      - since these tests are exported, it allows authors of datastores to get extra confidence that their implementations of the storage interface are correct.
```

docs/testing_your_changes.md
```diff
@@ -0,0 +1,52 @@
...
+
+> NOTE TO MAINTAINERS: ideally, these tests need not be exported, since they are testing for correctness and should be independent of the storage layer. They are the result of not having enough coverage with unit tests.
```
jon-whit, line 44: My vote - drop these callouts. Every one of these goes without saying. Assuming we had perfect coverage for all time and eternity, then we'd never need these. But having these tests at various layers give us confidence. Even if we had better storage layer coverage we'd still have value and benefit in having these, because they give us better ensures that we haven't broken an internal code contract.

docs/testing_your_changes.md
miparnisari: It shouldn't be the goal of those tests to test the _entire_ interface.

docs/testing_your_changes.md
miparnisari: People will notice this either way, when we ask them to import five suites of tests. When they ask us about it, we can show them why with this doc. 

I agree it makes us look bad, but we're being honest here. If they don't like it, we can tell them we haven't had time to improve and ask them to send a PR. WDYT?

jon-whit (approved):

Issue #1296, opened 2024-01-08
GoRoutine leak for list objects when there are more than 1000 items to return


adriantam:
### Checklist

- [X] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [X] I have searched the [Discord](https://discord.gg/8naAwJfWN6) community and have not found a suitable solution or answer.
- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

PProf shows that there are goroutine leaks in ListObjects when the number of items > max items for list objects

### Expectation

No goroutine leaks

### Reproduction

1. Set profiler enabled by specifying `--profiler-enabled`.  It does not matter what the datastore is
2. Create new model 
```
model
  schema 1.1

type user

type group
  relations
    define member: [user]
```
3. Adding tuples `group:1#member@user:A` to `group:10000#member@user:A`
4. Call list objects with `group#member@user:A`
5. Inspect profiler http://localhost:3001/debug/pprof/goroutine?debug=1 .  See that the number of goroutine increases per list object call.

### Store data

```yaml
model_file: |
model
  schema 1.1

type user

type group
  relations
    define member: [user]

tuples:
  - user: user:A
    relation: member
    object: group:1
...
  - user: user:A
    relation: member
    object: group:10000
tests: # remove this if not a bug in Check or ListObjects API
  - name: test-1
    list_objects:
      - user: user:A
        type: group
        relation: member
```

### OpenFGA version

1.4.1

### How are you running OpenFGA?

As a binary

### What datastore are you using?

In-Memory

### OpenFGA Flags

--profiler-enabled

PR #1297, opened 2024-01-08
fix: goroutine leaks in reverse expand

adriantam:
## Description
Fix goroutine leak due to paths in reverse expand not calling pool.Wait()

## References
Close https://github.com/openfga/openfga/issues/1296

## Review Checklist


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -312,14 +312,29 @@ func (c *ReverseExpandQuery) execute(
 			}
 			err = c.execute(ctx, r, resultChan, intersectionOrExclusionInPreviousEdges, resolutionMetadata)
 			if err != nil {
-				return err
+				var returnedErrors *multierror.Error
```
jon-whit, line 315: I would recommend we normalize and simplify this code to avoid duplication. May I suggest as an alternative the following snippet?

```go
var errs *multierror.Error

func _, edge := range edges {
    ...
    switch {
    ...
    case graph.ComputedUsersetEdge:
        ...
        err = c.execute(...)
        if err != nil {
          errs = multierror.Append(errs, err)
          break // this bit!
       }
    }
}

if err := pool.Wait(); err != nil {
    errs = multierror.Append(errs, err)
}

if errs.ErrorOrNil() != nil {
    return errs
}

return nil
```

This way we can avoid the duplications with the span handling (e.g. `span.End()`) as well as the telemetry and pool duplicated logic.

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -144,4 +144,120 @@ type document
 	case <-done:
 		// OK!
 	}
+
+	select {
+	case <-done:
+		t.Log("OK!")
```
jon-whit, line 150: Do we need this residual `t.Log`? Seems too non-specific to be worth keeping.

pkg/server/commands/reverseexpand/reverse_expand.go
adriantam: Will do.  The only caveat is that the break needs to be labeled as per https://go.dev/ref/spec#Break_statements

pkg/server/commands/reverseexpand/reverse_expand_test.go
adriantam: cleaned up

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -141,7 +141,116 @@ type document
...
 			require.Nil(t, res)
 		}
+		<-done
 	case <-done:
 		// OK!
 	}
 }
+
+// errorIterator is a mock iterator that returns error when calling next on the second Next call
+type errorIterator[T any] struct {
```
jon-whit, line 151: Can we please move this to `internal/mocks` instead of inline here?

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -141,7 +141,116 @@ type document
...
+	for i := 0; i < 100; i++ {
+		obj := fmt.Sprintf("document:%s", strconv.Itoa(i))
+		tuples = append(tuples, &openfgav1.Tuple{Key: tuple.NewTupleKey(obj, "viewer", "user:maria")})
+	}
+
+	mockDatastore := mocks.NewMockOpenFGADatastore(mockController)
+	mockDatastore.EXPECT().ReadStartingWithUser(gomock.Any(), store, gomock.Any()).
+		DoAndReturn(func(_ context.Context, _ string, _ storage.ReadStartingWithUserFilter) (storage.TupleIterator, error) {
+			iterator := newErrorIterator(tuples)
+			t.Logf("returning tuple iterator")
```
jon-whit, line 214: Is this log necessary?

adriantam: no, i can remove.

pkg/server/commands/reverseexpand/reverse_expand_test.go
adriantam: done

jon-whit (approved):

internal/mocks/mock_error_iterator.go
```diff
@@ -0,0 +1,50 @@
...
+	next, rest := s.items[0], s.items[1:]
+	s.items = rest
+
+	return next, nil
+}
+
+func (s *errorIterator[T]) Stop() {}
+
+// NewErrorIterator mocks case where Next will return error after the first Next()
+func NewErrorIterator(tuples []*openfgav1.Tuple) storage.TupleIterator {
```
miparnisari, line 43: We should probably add a `var _ TupleIterator = (*errorIterator)(nil)` ?

internal/mocks/mock_error_iterator.go
```diff
@@ -0,0 +1,50 @@
+package mocks
```
miparnisari, line 1: I'm not convinced by this package name. This isn't a mock, this is an implementation of a tupleiterator that does a specific thing. The fact that it should be only used in tests should probably be mentioned in the godoc of `NewErrorIterator`

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -312,22 +314,29 @@ func (c *ReverseExpandQuery) execute(
...
+				errs = multierror.Append(errs, err)
+				break LoopOnEdges
 			}
 		case graph.TupleToUsersetEdge:
 			pool.Go(func(ctx context.Context) error {
 				return c.reverseExpandTupleToUserset(ctx, r, resultChan, intersectionOrExclusionInPreviousEdges, resolutionMetadata)
 			})
 		default:
-			return fmt.Errorf("unsupported edge type")
+			errs = multierror.Append(errs, fmt.Errorf("unsupported edge type"))
```
miparnisari, line 325: I think this should be a panic. The `switch` already covers all possible graph edge types, if we messed this up then something went very wrong.

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -141,7 +141,75 @@ type document
 		} else {
 			require.Nil(t, res)
 		}
+		<-done
```
miparnisari, line 144: can you add a comment here that explains why we need this? this seems to be waiting for the goroutine above to avoid a leak, but you didn't add `defer goleak.VerifyNone(t)` so i'm not sure

miparnisari (commented):
Sorry for the late review 😓

internal/mocks/mock_error_iterator.go
adriantam: Are there mocks that are not used for tests?

PR #1299, opened 2024-01-09
Fix: contextual tuple sorting for cache key 

willvedd:
## Description
Fixing the sorting of contextual tuples when generating a cache key during check. Specifically, ensuring that the sorting accounts for all properties (object, user, relation).

This bug can result in caching misses for the same effective payload when multiple contextual tuples are provided and sorted differently. 

This was not picked up in previous unit tests because each set of unit tests only covered the sorting of a single property, not the combination of multiple properties.

The precedence of sorting will be: object, user and relation, in that order. This is an attempt to preserve the existing sorting. **Example:**

-  user:A relationA document:A
-  user:B relationA document:A
-  user:A relationB document:A
-  user:B relationB document:A
-  user:A relationA document:B
-  user:B relationA document:B
-  user:A relationB document:B
-  user:B relationB document:B

## Review Checklist


jon-whit (changes requested):
Please add a test in `internal/graph/cached_resolver_test.go` that demonstrates that a duplicate cache key can be generated with alternative ordering of contextual tuples? Specifically, add a test for `graph#checkRequestCacheKey` in that file.

internal/graph/cached_resolver_test.go
```diff
@@ -590,13 +590,13 @@ func TestCheckCacheKey_ContextualTuplesOrdering(t *testing.T) {
 	modelID := ulid.Make().String()
 
 	tuples1 := []*openfgav1.TupleKey{
-		tuple.NewTupleKey("document:1", "viewer", "user:jon"),
-		tuple.NewTupleKey("document:2", "viewer", "user:jon"),
+		tuple.NewTupleKey("document:1", "viewer", "user:anne"),
+		tuple.NewTupleKey("document:2", "admin", "user:jon"),
```
willvedd, line 594: Existing unit test only tested the sorting order of the object property but did not cover when the relation and user are differing values.

jon-whit: I'm realizing now there's another test gap in our cache key computation. Can we also add a test or two asserting alternative order of tuples with conditions are also ok?

For example,

```
document:1#viewer@user:anne, condition: "mycond"
document:2#admin@user:jon

vs

document:2#admin@user:jon
document:1#viewer@user:anne, condition: "mycond"
```

We have tests that ensure two different check requests with the same context are the same, but not necessarily tests with tuples involving a condition like the one above.

internal/keys/hasher.go
```diff
@@ -45,16 +45,16 @@ func (t tupleKeysHasher) Append(h hasher) error {
 	sortedTupleKeys := append([]*openfgav1.TupleKey(nil), t.tupleKeys...) // Copy input to avoid mutating it
 
 	sort.SliceStable(sortedTupleKeys, func(i, j int) bool {
-		if sortedTupleKeys[i].Object > sortedTupleKeys[j].Object {
-			return false
+		if sortedTupleKeys[i].Object != sortedTupleKeys[j].Object {
+			return sortedTupleKeys[i].Object < sortedTupleKeys[j].Object
```
miparnisari, line 49: For a followup PR: do you mind checking the rest of the codebase for uses of `sort.SliceStable`? We may have this problem elsewhere too

internal/graph/cached_resolver_test.go
willvedd: @jon-whit AFAICT, the cache key generation does not account for condition. Which explains the absence of testing but begs the question if it should be accounted for. If so, I can look into adding this capability.

Added the requested test for this anyway.

jon-whit: I’ll need to think about this one more tomorrow.

You can’t have the same base triplet (e.g. `(object, relation, user)`) in the database at the same time where one is conditioned and the other is not. In other words, you can’t have

```
- document:1#viewer@user:will
- document:1#viewer@user:will, condition: "somecondition"
```
at the same time in the same store.

If we cache a Check subproblem with a tuple that is not conditioned, then no matter what you do to the tuples (deletions etc) the subproblem will be cached for the TTL regardless if you delete it and re-introduce it with a condition. So I don’t think omitting the condition from the cache key is an issue per-se, but I need to think more about it tomorrow 🤔

willvedd: If there is something actionable, can that be done in a follow-up? AFAICT, the spirit of this PR has been fulfilled and its ready to go.

jon-whit (approved):

Issue #1309, opened 2024-01-22
GoRoutine leak in case of list object timeout


adriantam:
### Checklist

- [X] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [X] I have searched the [Discord](https://discord.gg/8naAwJfWN6) community and have not found a suitable solution or answer.
- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

There is a go routine leak at ListObjectsQuery.evaluate in case of list object timeout

### Expectation

There should be no goroutine leak

### Reproduction

1.  Enable go profiler.  Disable check query cache and using postgres makes the problem easier to to reproduce
```
./openfga run --datastore-engine postgres --datastore-uri 'postgres://postgres:password@localhost:5432/postgres' --profiler-enabled=true --check-query-cache-enabled=false --listObjects-deadline=1s 


```
2. Use the specify file and populate with the specified tuples
3. Run list object (easier to reproduce when running in parallel)
```
fga query list-objects --store-id=01HKMSRHKACCYQ9KJDFX83QE37 user:A can_read doc & fga query list-objects --store-id=01HKMSRHKACCYQ9KJDFX83QE37 user:A can_read doc & fga query list-objects --store-id=01HKMSRHKACCYQ9KJDFX83QE37 user:A can_read doc
```

Inspect 
http://localhost:3001/debug/pprof/goroutine?debug=1
and see that there is a leak every time there is a list object timeout
```
5 @ 0x104295b98 0x1042610c4 0x104260c94 0x104d22264 0x1042ca674
#	0x104d22263	github.com/openfga/openfga/pkg/server/commands.(*ListObjectsQuery).evaluate.func1+0x643	/Users/adrian.tam/src/github.com/openfga/openfga/pkg/server/commands/list_objects.go:252
```

### Store data

```yaml
model
  schema 1.1

type user

type group
  relations
    define member: [user]

type folder
  relations
    define can_create_file: owner
    define owner: [user]
[...]

PR #1310, opened 2024-01-22
fix: reverse expand closes channel in all cases

miparnisari:
## Description

Before this PR we had in `ReverseExpand`:

```go
if err != nil {
		select {
		case <-ctx.Done():
			return
		case resultChan <- &ReverseExpandResult{Err: err}:
			return
		}
	}
close(resultChan)
```

Notice how if we hit the first `return`s, we are not sending anything through the channel (which is reserved for when we saw no errors)

So I changed it to

```go
if err != nil {
		select {
		case <-ctx.Done():
			reverseExpandResultsChan <- &ReverseExpandResult{Err: ctx.Err()}
			return
		case reverseExpandResultsChan <- &ReverseExpandResult{Err: err}:
			return
		}
	}
close(resultChan)
```
so that if we see an error (any error) we send it through the channel.

 The remaining changes I can remove them from this PR depending on how reviewers feel, I added them because I used those to help me debug.

## References

Closes https://github.com/openfga/openfga/issues/1309.

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -16,6 +16,8 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/protobuf/types/known/structpb"
 
+	"github.com/openfga/openfga/pkg/logger"
+
 	"github.com/openfga/openfga/internal/condition"
 	"github.com/openfga/openfga/internal/condition/eval"
 	"github.com/openfga/openfga/internal/graph"
@@ -107,6 +109,7 @@ type UserRef struct {
 }
 
 type ReverseExpandQuery struct {
+	logger                  logger.Logger
 	datastore               storage.RelationshipTupleReader
 	typesystem              *typesystem.TypeSystem
 	resolveNodeLimit        uint32
@@ -132,8 +135,15 @@ func WithResolveNodeBreadthLimit(limit uint32) ReverseExpandQueryOption {
 	}
 }
 
+func WithLogger(logger logger.Logger) ReverseExpandQueryOption {
+	return func(d *ReverseExpandQuery) {
+		d.logger = logger
+	}
+}
+
 func NewReverseExpandQuery(ds storage.RelationshipTupleReader, ts *typesystem.TypeSystem, opts ...ReverseExpandQueryOption) *ReverseExpandQuery {
 	query := &ReverseExpandQuery{
+		logger:                  logger.NewNoopLogger(),
 		datastore:               ds,
 		typesystem:              ts,
 		resolveNodeLimit:        serverconfig.DefaultResolveNodeLimit,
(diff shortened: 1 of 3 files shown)


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -173,36 +183,38 @@ func NewResolutionMetadata() *ResolutionMetadata {
...
 	req *ReverseExpandRequest,
-	resultChan chan<- *ReverseExpandResult,
+	reverseExpandResultsChan chan<- *ReverseExpandResult,
 	resolutionMetadata *ResolutionMetadata,
 ) {
-	err := c.execute(ctx, req, resultChan, false, resolutionMetadata)
+	err := c.execute(ctx, req, reverseExpandResultsChan, false, resolutionMetadata)
 	if err != nil {
 		select {
 		case <-ctx.Done():
```
jon-whit, line 203: Why not handle this case the same as the one below? It's the callers choice how to handle cancellations of the underlying ReverseExpand call.

e.g.

```go
if err != nil {
    select {
        case err := <-ctx.Done()
            reverseExpandResultsChan <- &ReverseExpandResult{Err: err}
        case reverseExpandResultsChan <- &ReverseExpandResult{Err: err}
    }
}
```

Usually `ctx.Err()` isn't swallowed or treated differently than `err != nil` cases, so it's odd we're swallowing that specifically.

miparnisari: We need to consider the scenario that the channel might be full already, so trying to send anything to it will block indefinitely

miparnisari: Update: before this PR we had

```go
if err != nil {
		select {
		case <-ctx.Done():
			return
		case resultChan <- &ReverseExpandResult{Err: err}:
			return
		}
	}
```

I noticed while writing a new unit test that, unless you call `cancel()` on the context, this will block forever too, which means the test never ends.

So I changed it to

```go
if err != nil {
		select {
		case resultChan <- &ReverseExpandResult{Err: err}:
		default:
		}
	}
```

so that we don't need context cancellation for ReverseExpand to finish.

miparnisari: Updated the PR description with more details

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -21,6 +21,78 @@ import (
...
+				Object: &openfgav1.Object{
+					Type: "user",
+					Id:   "maria",
+				},
+			},
+			ContextualTuples: []*openfgav1.TupleKey{},
+		}, resultChan, NewResolutionMetadata())
+		t.Logf("after execute reverse expand")
+		doneProducing <- struct{}{}
+	}()
```
jon-whit, line 76: nit pick - but do you mind if we add some newlines between these goroutines?

miparnisari:
Superseded by https://github.com/openfga/openfga/pull/1315

PR #1315, opened 2024-01-23
fix: handle ReverseExpand channel closure correctly

jon-whit:
## Description
Avoid the forgotten consumer problem between ListObjects and ReverseExpand which, under some query outcomes, could lead to memory leaks. In particular, if we hit the `--listObjects-deadline` while actively reverse expanding, then the ListObjects consumer listening on the ReverseExpand results channel would be left dangling and thus not release the resources back to the Go garbage collector. The changes herein fix that issue particularly and thus avoid memory leaks.

## References
Closes https://github.com/openfga/openfga/issues/1309.

## Review Checklist

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -218,7 +218,9 @@ func (q *ListObjectsQuery) evaluate(
 		reverseExpandResultsChan := make(chan *reverseexpand.ReverseExpandResult, 1)
 		objectsFound := atomic.Uint32{}
 
-		reverseExpandQuery := reverseexpand.NewReverseExpandQuery(q.datastore, typesys,
+		reverseExpandQuery := reverseexpand.NewReverseExpandQuery(
+			q.datastore,
+			typesys,
 			reverseexpand.WithResolveNodeLimit(q.resolveNodeLimit),
 			reverseexpand.WithResolveNodeBreadthLimit(q.resolveNodeBreadthLimit),
 			reverseexpand.WithLogger(q.logger),
@@ -228,11 +230,13 @@ func (q *ListObjectsQuery) evaluate(
 
 		wg := sync.WaitGroup{}
 
+		errChan := make(chan error, 1)
+
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 
-			reverseExpandQuery.Execute(cancelCtx, &reverseexpand.ReverseExpandRequest{
+			err := reverseExpandQuery.Execute(cancelCtx, &reverseexpand.ReverseExpandRequest{
 				StoreID:          req.GetStoreId(),
 				ObjectType:       targetObjectType,
 				Relation:         targetRelation,
@@ -240,6 +244,9 @@ func (q *ListObjectsQuery) evaluate(
 				ContextualTuples: req.GetContextualTuples().GetTupleKeys(),
 				Context:          req.GetContext(),
 			}, reverseExpandResultsChan, resolutionMetadata)
+			if err != nil {
+				errChan <- err
+			}
 		}()
 
 		checkResolver := graph.NewLocalChecker(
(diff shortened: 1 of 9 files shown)


pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -201,20 +271,87 @@ type document
...
+	model := testutils.MustTransformDSLToProtoWithID(`model
+  schema 1.1
+type user
+type document
+  relations
+    define viewer: [user]`)
+
+	mockDatastore := mocks.NewMockSlowDataStorage(memory.New(), 1*time.Second)
+
+	// if the error can be sent through the channel, this test should pass
```
jon-whit, line 311: drop this comment - it's not longer relevant

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -195,18 +194,14 @@ func (c *ReverseExpandQuery) Execute(
 	req *ReverseExpandRequest,
 	resultChan chan<- *ReverseExpandResult,
 	resolutionMetadata *ResolutionMetadata,
-) {
+) error {
```
miparnisari, line 197: Can you update the godoc?

```
// Execute yields all the objects of the provided objectType that the given user has, possibly, a specific relation with
// and sends those objects to resultChan. It MUST guarantee no duplicate objects sent.
// This function respects context timeouts and cancellations.
// If an error is encountered (e.g. context timeout) before resolving all objects: the provided channel will NOT be closed, 
// and it will send the error through the channel.
// If no errors, Execute will yield all of the objects on the provided channel and then close the channel
// to signal that it is done.
```

pkg/testfixtures/storage/postgres.go
```diff
@@ -98,7 +99,13 @@ func (p *postgresTestContainer) RunPostgresTestContainer(t testing.TB) (Datastor
 	cont, err := dockerClient.ContainerCreate(context.Background(), &containerCfg, &hostCfg, nil, nil, name)
 	require.NoError(t, err, "failed to create postgres docker container")
 
+	var db *sql.DB
+
 	stopContainer := func() {
+		if db != nil {
+			db.Close()
+		}
```
miparnisari, line 107: why do we need this? by the time this function runs, `db` is nil always

pkg/testfixtures/storage/mysql.go
```diff
@@ -100,7 +100,13 @@ func (m *mySQLTestContainer) RunMySQLTestContainer(t testing.TB) (DatastoreTestC
 	cont, err := dockerClient.ContainerCreate(context.Background(), &containerCfg, &hostCfg, nil, nil, name)
 	require.NoError(t, err, "failed to create mysql docker container")
 
+	var db *sql.DB
+
 	stopContainer := func() {
+		if db != nil {
+			db.Close()
+		}
```
miparnisari, line 108: same here

pkg/server/test/list_objects.go
```diff
@@ -520,11 +520,22 @@ condition condition1(x: int) {
...
+						select {
+						case objectID, open := <-server.channel:
+							if !open {
+								done <- struct{}{}
+								return
+							}
+
+							streamedObjectIds = append(streamedObjectIds, objectID)
+
+						// for tests whose deadline is sooner than the latency of the storage layer
```
miparnisari, line 533: 👍

pkg/server/commands/list_objects.go
```diff
@@ -424,37 +426,25 @@ func (q *ListObjectsQuery) ExecuteStreamed(ctx context.Context, req *openfgav1.S
...
-			q.logger.WarnWithContext(
-				ctx, fmt.Sprintf("list objects timeout after %s", q.listObjectsDeadline.String()),
-			)
-			return resolutionMetadata, nil
-
-		case result, channelOpen := <-resultsChan:
-			if !channelOpen {
-				// Channel closed! No more results.
-				return resolutionMetadata, nil
+	for result := range resultsChan {
```
miparnisari, line 429: ❤️

pkg/testfixtures/storage/postgres.go
miparnisari: actually, if you move `var db *sql.DB` to 139, you don't need this

jon-whit: True, because we `defer db.Close()` above any points of potential error below it. It felt like an added safeguard as it guarantees if even if miss a point herein the caller calling `stop` will close it. That's why I had it.

miparnisari (approved):

Issue #1172, opened 2023-11-24
Speed up Check benchmarks


miparnisari:
## The problem

https://github.com/openfga/openfga/blob/ca5dde2b9574c1b7137881a852610d5bf8aa55e6/tests/check/check_test.go#L279-L294

Benchmarks should test as little code as possible to be more consistent across runs. 

Here, we are running a full blown server which has stuff running in the background by default, like the playground and the metrics endpoint. This can make the  benchmarks take longer. The time it takes to spin up the server non-trivial and it will impact the outcome. We want the benchmarks to measure the API itself, not the networking code required.

## The task

- Move the existing benchmarks https://github.com/openfga/openfga/blob/19bc3294a34d35561a9ac78f3ed1fa0fa35b93cf/tests/check/check_test.go#L364-L381 to a new file: `pkg/server/test/check.go`
- Include those benchmarks here: https://github.com/openfga/openfga/blob/07794019d080bf629a6408d9917ea1492e1eaba6/pkg/server/test/server.go#L51-L54
- Modify the benchmarks:
  - Remove the `server` object that is created on every benchmark and use the `Check` command instead: https://github.com/openfga/openfga/blob/main/pkg/server/commands/check_command.go. As guidance, you can see what the ListUsers benchmark looks like: https://github.com/openfga/openfga/blob/ca44adbe218a54cfcc05dfc61a9eb7ef6d935837/pkg/server/test/list_users.go#L182-L198. (You don't need to copy the table structure).

⚠ make sure that all the parameters passed to the `query` object (for example: MaxConcurrentReads), are the same as they are today, or else the benchmarks could alarm.
- Remove these two comments: https://github.com/openfga/openfga/blob/fe5eb52f531a94cb67d84dfbee2f3837d1980442/.github/workflows/pull_request.yaml#L103 and here https://github.com/openfga/openfga/blob/fe5eb52f531a94cb67d84dfbee2f3837d1980442/.github/workflows/pull_request.yaml#L103, as benchmarks can take ~17 minutes to run.

miparnisari:
See https://github.com/openfga/openfga/pull/1317#discussion_r1465524541

PR #1317, opened 2024-01-24
test: amend #1316

miparnisari:
We need to stop the benchmark timer when tear down is costly in terms of time, so that time spent tearing down is not added to the benchmark itself.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,7 +10,7 @@ permissions:
 jobs:
   go-bench:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -10,7 +10,7 @@ permissions:
 jobs:
   go-bench:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30 # until we do https://github.com/openfga/openfga/issues/1172
```
jon-whit, line 13: Cross reference this issue with #1172 so that we remember this 😄

jon-whit (approved):

.github/workflows/main.yaml
miparnisari: done

PR #1199, opened 2023-12-01
refactor: allow better testing of tracing telemetry

jpadilla:
## References

#1196 

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #1318, opened 2024-01-24
feat: new server.Stop() to close all server resources

miparnisari:
This is a follow-up of https://github.com/openfga/openfga/pull/1317.

Whether we use OpenFGA as a library or as a server, we need to provide an API to clean up Server resources.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -99,7 +99,8 @@ type Server struct {
 	experimentals                    []ExperimentalFeatureFlag
 	serviceName                      string
 
-	typesystemResolver typesystem.TypesystemResolverFunc
+	typesystemResolver     typesystem.TypesystemResolverFunc
+	typesystemResolverStop func()
 
 	checkOptions           []graph.LocalCheckerOption
 	checkQueryCacheEnabled bool
@@ -306,9 +307,17 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		return nil, fmt.Errorf("request duration datastore count buckets must not be empty")
 	}
 
-	s.typesystemResolver = typesystem.MemoizedTypesystemResolverFunc(s.datastore)
+	s.typesystemResolver, s.typesystemResolverStop = typesystem.MemoizedTypesystemResolverFunc(s.datastore)
 
 	return s, nil
+}
+
+func (s *Server) Close() {
+	if s.checkCache != nil {
+		s.checkCache.Stop()
+	}
+	s.datastore.Close()
+	s.typesystemResolverStop()
 }
 
 func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequest) (*openfgav1.ListObjectsResponse, error) {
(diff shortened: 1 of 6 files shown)


pkg/server/server.go
```diff
@@ -99,7 +99,8 @@ type Server struct {
 	experimentals                    []ExperimentalFeatureFlag
 	serviceName                      string
 
-	typesystemResolver typesystem.TypesystemResolverFunc
+	typesystemResolver     typesystem.TypesystemResolverFunc
+	typesystemResolverStop func()
```
miparnisari, line 103: this doesn't look super nice to me but it's the fastest way to remove the leak. We can make this nicer later

pkg/server/server.go
```diff
@@ -306,11 +307,19 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+	s.typesystemResolver, s.typesystemResolverStop = typesystem.MemoizedTypesystemResolverFunc(s.datastore)
 
 	return s, nil
 }
 
+func (s *Server) Close() {
+	if s.checkCache != nil {
+		s.checkCache.Stop()
+	}
+	s.datastore.Close()
```
jon-whit, line 319: Doesn't feel right to have the Server be the responsible party to close the datastore. The datastore is a dependency of the Server and is passed in. It should be the callers responsibility to close it because it is the callers responsibility to construct it. This approach leaks the responsibility of the caller vs callee.

pkg/server/server.go
jon-whit: Makes sense. I think it's fine. Nice addition 👍

pkg/server/server.go
jon-whit: cc @willvedd @jpadilla @adriantam @sergiught for your opinions. Should the constructor of the datastore be the party responsible for closing it? If not, why do you feel the Server should be the responsible party?

adriantam: IMO, whoever create the object is responsible for closing it.  Otherwise, you will have a situation where the caller will still have reference to the object while the callee had close it.

miparnisari: Oh! I forgot that the server doesn't actually instantiate the `datastore`, it merely takes a reference to it :) good point, i'll leave this one out

willvedd: I think its a well intentioned DX improvement but creator should be expected to close. Might be unnecessary risk otherwise.

I think an easy n+1 is to include the close in any public examples we have documented.

tests/check/check_test.go
```diff
@@ -45,6 +45,9 @@ func TestCheckMySQL(t *testing.T) {
 }
 
 func TestCheckLogs(t *testing.T) {
+	// uncomment after https://github.com/openfga/openfga/pull/1199 is done. the span exporter needs to be closed properly
```
jon-whit, line 48: Cross referencing with #1199 so we don't forget this.

tests/listobjects/listobjects_test.go
```diff
@@ -25,6 +25,8 @@ func TestListObjectsMySQL(t *testing.T) {
 }
 
 func testRunAll(t *testing.T, engine string) {
+	// uncomment in https://github.com/openfga/openfga/pull/1315
```
jon-whit, line 28: Cross referencing with #1315 so we don't forget.

jon-whit (approved):

PR #1325, opened 2024-01-30
test: add tests for storage.ReadPage

miparnisari:
I am adding one test that is meant to simulate the case where ReadPage is supposed to skip over tuples that don't match the filter.

In the future we need to consider revisiting our entire storage interface tests, see what code paths aren't well covered, and write tests for those.

### DIFF
--- a/internal/graph/graph_test.go
+++ b/internal/graph/graph_test.go
@@ -6,9 +6,9 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/stretchr/testify/require"
-	"google.golang.org/protobuf/testing/protocmp"
 
 	"github.com/openfga/openfga/pkg/testutils"
 
@@ -262,8 +262,8 @@ type organization
 			require.NoError(t, err)
 
 			cmpOpts := []cmp.Option{
+				cmpopts.IgnoreUnexported(openfgav1.RelationReference{}),
 				RelationshipEdgeTransformer,
-				protocmp.Transform(),
 			}
 			if diff := cmp.Diff(test.expected, edges, cmpOpts...); diff != "" {
 				t.Errorf("mismatch (-want +got):\n%s", diff)
@@ -1490,8 +1490,8 @@ type document
 			require.NoError(t, err)
 
 			cmpOpts := []cmp.Option{
+				cmpopts.IgnoreUnexported(openfgav1.RelationReference{}),
 				RelationshipEdgeTransformer,
-				protocmp.Transform(),
 			}
 			if diff := cmp.Diff(test.expected, edges, cmpOpts...); diff != "" {
 				t.Errorf("mismatch (-want +got):\n%s", diff)
--- a/internal/graph/graph_test.go
+++ b/internal/graph/graph_test.go
@@ -6,9 +6,9 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/testing/protocmp"
 
 	"github.com/openfga/openfga/pkg/testutils"
 
@@ -21,16 +21,16 @@ var (
 
 		// Sort by Type and then by edge and then by tupleset relation
 		sort.SliceStable(out, func(i, j int) bool {
-			if out[i].Type > out[j].Type {
-				return false
+			if out[i].Type != out[j].Type {
+				return out[i].Type < out[j].Type
 			}
 
-			if typesystem.GetRelationReferenceAsString(out[i].TargetReference) > typesystem.GetRelationReferenceAsString(out[j].TargetReference) {
-				return false
+			if typesystem.GetRelationReferenceAsString(out[i].TargetReference) != typesystem.GetRelationReferenceAsString(out[j].TargetReference) {
+				return typesystem.GetRelationReferenceAsString(out[i].TargetReference) < typesystem.GetRelationReferenceAsString(out[j].TargetReference)
 			}
 
-			if out[i].TuplesetRelation > out[j].TuplesetRelation {
-				return false
+			if out[i].TuplesetRelation != out[j].TuplesetRelation {
+				return out[i].TuplesetRelation < out[j].TuplesetRelation
 			}
 
 			return true
@@ -262,8 +262,8 @@ type organization
 			require.NoError(t, err)
 
 			cmpOpts := []cmp.Option{
-				cmpopts.IgnoreUnexported(openfgav1.RelationReference{}),
 				RelationshipEdgeTransformer,
+				protocmp.Transform(),
 			}
 			if diff := cmp.Diff(test.expected, edges, cmpOpts...); diff != "" {
 				t.Errorf("mismatch (-want +got):\n%s", diff)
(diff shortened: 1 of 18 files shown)


pkg/storage/test/tuples.go
```diff
@@ -1215,6 +1216,64 @@ func ReadTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+		}
+
+		cmpOpts := []cmp.Option{
+			testutils.TupleKeyCmpTransformer,
+			protocmp.Transform(),
+		}
+		if diff := cmp.Diff(expectedTuples, receivedTuples, cmpOpts...); diff != "" {
+			t.Errorf("mismatch (-want +got):\n%s", diff)
+		}
+		require.Empty(t, contToken)
```
sergiught, line 1273: We can remove the `if` if we make use of the assert and leverage its ability to provide accurate error messages:

```suggestion
		diff := cmp.Diff(expectedTuples, receivedTuples, cmpOpts...)
		
		assert.Empty(t, diff)
		assert.Empty(t, contToken)
```

internal/graph/graph_test.go
```diff
@@ -21,16 +21,16 @@ var (
 
 		// Sort by Type and then by edge and then by tupleset relation
 		sort.SliceStable(out, func(i, j int) bool {
-			if out[i].Type > out[j].Type {
-				return false
+			if out[i].Type != out[j].Type {
+				return out[i].Type < out[j].Type
 			}
```
miparnisari, line 26: this problem was identified 3 weeks ago 😅 https://github.com/openfga/openfga/pull/1299#discussion_r1446703961

rhamzeh (approved):

jon-whit:
It seems we're doing 2 things in the scope of this PR

1. Changing the tuple validation logic in the `graph` package to make sure we filter out invalid tuples correctly and
2. Adding tests for `storage#ReadPage`

#2 can presumably be in its own PR. I don't see why its relevant to lump it in with #1. What was the reasoning behind combining these two code changes together?

pkg/storage/postgres/postgres.go
```diff
@@ -158,7 +158,7 @@ func (p *Postgres) ReadPage(ctx context.Context, store string, tupleKey *openfga
 	return iter.ToArray(opts)
 }
 
-func (p *Postgres) read(ctx context.Context, store string, tupleKey *openfgav1.TupleKey, opts *storage.PaginationOptions) (*sqlcommon.SQLTupleIterator, error) {
+func (p *Postgres) read(ctx context.Context, store string, tupleKey *openfgav1.TupleKey, opts *storage.PaginationOptions, sortByUlid bool) (*sqlcommon.SQLTupleIterator, error) {
```
jon-whit, line 161: `sortByUlid` belongs as an option to the method. 5 function parameters is starting to get a bit wieldy.

pkg/storage/mysql/mysql.go
```diff
@@ -161,9 +161,11 @@ func (m *MySQL) read(ctx context.Context, store string, tupleKey *openfgav1.Tupl
 		).
 		From("tuple").
 		Where(sq.Eq{"store": store})
-	if opts != nil {
+
+	if sortByUlid {
```
miparnisari, line 165: no functional change here but it's more clear when we sort now (in ReadPage, not Read)

pkg/storage/storage.go
```diff
@@ -75,9 +76,9 @@ type RelationshipTupleReader interface {
 	// There is NO guarantee on the order returned on the iterator.
 	Read(ctx context.Context, store string, tupleKey *openfgav1.TupleKey) (TupleIterator, error)
 
-	// ReadPage functions similarly to Read but includes support for pagination. It takes additional
-	// pagination parameters and returns a slice of tuples along with a continuation token, which may
-	// not be empty. This token can be used for retrieving subsequent pages of data.
+	// ReadPage functions similarly to Read but includes support for pagination. It takes
+	// mandatory pagination options (pageSize can be zero :/)
```
miparnisari, line 80: I just noticed this in this PR that the interface of ReadPage is really weird, if you do `ReadPage(.... &storageOptions{From: "cont-token"})` then the storage gets `pageSize = 0`. I think implementations should handle this, in a different PR

pkg/storage/test/tuples.go
```diff
@@ -1215,6 +1217,57 @@ func ReadTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+			{Key: tuple.NewTupleKey("document:1", "admin", "user:anne")},
+			{Key: tuple.NewTupleKey("document:1", "reader", "user:anne")},
+		}
+
+		cmpOpts := []cmp.Option{
+			protocmp.IgnoreFields(protoadapt.MessageV2Of(&openfgav1.Tuple{}), "timestamp"),
+			testutils.TupleCmpTransformer,
+			protocmp.Transform(),
+		}
+		diff := cmp.Diff(expectedTuples, tuplesRead, cmpOpts...)
```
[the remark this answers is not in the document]
miparnisari: Does this warrant a helper? For me, it's fast enough to find usages of `cmp.Diff` and then copy-paste the options 🤷‍♀️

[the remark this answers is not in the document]
jon-whit: We're comparing proto tuples in more than one place, and we have 5 lines of boilerplate every time we have to because of this. It most certainly warrants a helper, for the same reason we have helpers similarly to reduce test boilerplate.

internal/graph/graph_test.go
willvedd: 🙈

jon-whit: Ideally we'd delegate to a shared test utility here that shares the logic. Otherwise this may bite us in the ass again.

pkg/storage/test/tuples.go
```diff
@@ -1215,6 +1217,92 @@ func ReadTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+				PageSize: 50, // fetch the remainder
+				From:     string(contToken),
+			},
+		)
+		require.NoError(t, err)
+
+		require.Len(t, secondRead, 1)
+		require.Equal(t, "document:1", secondRead[0].Key.Object)
+		require.Equal(t, "user:anne", secondRead[0].Key.User)
+		require.NotEqual(t, firstRead[0].Key.Relation, secondRead[0].Key.Relation)
```
willvedd, line 1301: If the ordering is deterministic, can't we make an exact assertion of the relation for both reads?

willvedd (approved):
One small question about an assertion but otherwise good set of additions 👍

pkg/storage/test/tuples.go
miparnisari: I removed the documentation saying that this function returns sorted by ULID. It's true for Mysql and Postgres, but may not be true for other storage implementations.

willvedd (approved):

pkg/storage/test/tuples.go
```diff
@@ -1215,6 +1217,92 @@ func ReadTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+			storeID,
+			tuple.NewTupleKey("document:1", "", "user:anne"),
+			storage.PaginationOptions{
+				PageSize: 2,
+			},
+		)
+		require.NoError(t, err)
+
+		expectedTuples := []*openfgav1.Tuple{
+			{Key: tuple.NewTupleKey("document:1", "admin", "user:anne")},
```
jon-whit, line 1255: `tuple.NewTupleKey(..)`

jon-whit, line 1255:
```suggestion
			tuple.NewTupleKey("document:1", "admin", "user:anne"),
```

pkg/storage/test/tuples.go
```diff
@@ -1215,6 +1217,92 @@ func ReadTest(t *testing.T, datastore storage.OpenFGADatastore) {
...
+			{Key: tuple.NewTupleKey("document:1", "reader", "user:anne")},
```
jon-whit, line 1256:
```suggestion
			tuple.NewTupleKey("document:1", "reader", "user:anne"),
```

miparnisari: i assume this is a nitpick and will come back to it in a different PR since i already have approvals

pkg/storage/test/tuples.go
miparnisari: i assume this is a nitpick and will come back to it in a different PR since i already have approvals

pkg/storage/storage.go
miparnisari: @adriantam

PR #1329, opened 2024-01-31
Change default time format to ISO8601 in the logger

raj-saxena:
## Description
Changes the default time format of the logger to ISO8601.
Verified visually (since there was no test for this)
image

## References
Fixes https://github.com/openfga/openfga/issues/1326

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

Issue #1326, opened 2024-01-30
Support ISO8601 time format in JSON logs.


raj-saxena:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

We are running OpenFGA in our Kubernetes cluster using the helm-chart. There are 2 logging formats that are supported - `text` and `json` and [it is advised](https://github.com/openfga/helm-charts/blob/d3f80da94057b294570f1a2b314f760ac8ca9bd0/charts/openfga/values.schema.json#L287) to use `json` for `production`. 

The default timestamp format for `json` is Unix time format i.e., the number of _seconds_ since Epoch (Example - `1706609988.788818`).
<img width="1512" alt="image" src="https://github.com/openfga/openfga/assets/6253803/0b34d6a5-fdfd-4f8f-bbf0-90cc6998decc">

However, the issue is that [Datadog expects ](https://docs.datadoghq.com/logs/guide/logs-not-showing-expected-timestamp/#supported-date-formats)_milliseconds_ for this format. 
<img width="874" alt="image" src="https://github.com/openfga/openfga/assets/6253803/c555286c-d85e-4571-82a5-3b041494daa6">
I have been working with their support and also found [this issue ](https://github.com/DataDog/datadog-agent/issues/3926) that confirms the limitation.

There's no easy way to fix it in Datadog and without this, the logs don't seem to be ingested in their system. Since Datadog is a big enough observability platform, I come to you to propose a possible solution.

### Describe the ideal solution

I see that the [`text` log format supports `ISO8601` as timestamp format](https://github.com/openfga/openfga/blob/ad04038afbd58890cb65b409780b0cbbf85d5103/pkg/logger/logger.go#L125-L129)
However, I can't use them for my purposed because of the weird format of the generated logs. 
<img width="1501" alt="image" src="https://github.com/openfga/openfga/assets/6253803/b3cfef33-6fab-4d26-8aeb-87ea9a30e6cc"> 
(By weird, I mean the color and the json-in-text that breaks parsing)
<img width="729" alt="image" src="https://github.com/openfga/openfga/assets/6253803/66b2a080-a1a0-4fee-9042-1154386c7626">

The simplest solution seems like either:
[...]

PR #1330, opened 2024-01-31
Add an optional way to override timestamp format in json logs

raj-saxena:
## Description
I opened an issue earlier where @jon-whit agreed (https://github.com/openfga/openfga/issues/1326#issuecomment-1917797974) that moving towards ISO8601 is a better option.
As such, I raised a PR earlier https://github.com/openfga/openfga/pull/1329 where @jon-whit raised a valid point (https://github.com/openfga/openfga/pull/1329#issuecomment-1919242763) that we should not break compatibility for any existing users. 

~I have made the changes accordingly in this PR making ISO8601 as the default format while giving an option to enable the older option too if someone wants that. I understand that this could be a breaking change for some users but in the long run, I believe, this would be a step in the right direction.~ (adjusted to keep `Unix` as default after review comments)

Verified visually (since there was no test for this)
image

## References
Fixes https://github.com/openfga/openfga/issues/1326 while addressing the comment from @jon-whit 
https://github.com/openfga/openfga/pull/1329#issuecomment-1919242763

## Review Checklist
- [x] I have added tests to validate that the change in functionality is working as expected - provided visual evidence


cmd/run/run.go
```diff
@@ -267,7 +269,7 @@ func run(_ *cobra.Command, _ []string) {
 		panic(err)
 	}
 
-	logger := logger.MustNewLogger(config.Log.Format, config.Log.Level)
+	logger := logger.MustNewLogger(config.Log)
```
raj-saxena, line 272: It didn't feel right to add another method parameter when we have a dedicated struct already

pkg/logger/logger.go
```diff
@@ -93,13 +94,13 @@ func NewNoopLogger() *ZapLogger {
 	}
 }
 
-func NewLogger(logFormat, logLevel string) (*ZapLogger, error) {
-	if logLevel == "none" {
+func NewLogger(logConfig config.LogConfig) (*ZapLogger, error) {
```
raj-saxena, line 97: Possibly a better alternative than
```go
func NewLogger(logFormat, logLevel, logTimestampFormat string) (*ZapLogger, error) {
```

.config-schema.json
```diff
@@ -318,6 +318,13 @@
                     "enum": ["none", "debug", "info", "warn", "error", "panic", "fatal"],
                     "default": "info",
                     "x-env-variable": "OPENFGA_LOG_LEVEL"
+                },
+                "timestampFormat": {
+                    "description": "The timestamp format to use for the log output.",
+                    "type": "string",
+                    "enum": ["Unix", "ISO8601"],
+                    "default": "ISO8601",
```
jon-whit, line 326: Default should be `Unix`. That's the current default today. Otherwise we'll break compability.

.config-schema.json
```diff
@@ -318,6 +318,13 @@
                     "enum": ["none", "debug", "info", "warn", "error", "panic", "fatal"],
                     "default": "info",
                     "x-env-variable": "OPENFGA_LOG_LEVEL"
+                },
+                "timestampFormat": {
```
jon-whit, line 322:
```suggestion
                "logTimestampFormat": {
```

jon-whit:
Going to close #1329 in favor of this.

miparnisari (approved):

miparnisari (changes requested):
Like Jon said, let's keep the backward compatibility. Otherwise lgtm. We'll have to think where to add a test for this. `TestCheckLogs` already asserts on some fields logged, but I fear that test is becoming too bloated

jon-whit (changes requested):
Default needs to be Unix.

raj-saxena:
@jon-whit @miparnisari I have pushed [a commit](https://github.com/openfga/openfga/pull/1330/commits/5513f8173feb2287a2e06e81988efe7ea987d6e1) defaulting to `Unix` format for json. I have updated the description and the screenshot in the PR too. 
Please note that there's no way with the current change to have `Unix` for `text` format, but it also isn't something in the scope of the linked issue or being asked for right now.

internal/server/config/config.go
```diff
@@ -126,6 +126,9 @@ type LogConfig struct {
 
 	// Level is the log level to use in the log output (e.g. 'none', 'debug', or 'info')
 	Level string
+
+	// Format of the timestamp in the log output (e.g. 'Unix' or 'ISO8601'(default))
```
jon-whit, line 130: This comment needs to be updated. Default is 'Unix'

pkg/logger/logger.go
jon-whit: This logger is a publicly exported signature (e.g. `pkg/logger`). The server config is not a publicly exported signature (e.g. `internal/server/config`).

This change will break consumers of the OpenFGA library and make it impossible for anyone to construct a Logger externally as they cannot construct a `config.LogConfig` object since it is an internal package.

raj-saxena: Hi @jon-whit, sorry about missing that I am using an `internal` package.
I tried to take a stab at the code by changing to the options pattern like you suggested, but I am not comfortable enough to do such refactoring. As such, I have reverted to using expanded parameters. I hope this is okay and we can move this PR forward.

pkg/logger/logger.go
```diff
@@ -127,6 +127,12 @@ func NewLogger(logFormat, logLevel string) (*ZapLogger, error) {
 		cfg.DisableCaller = true
 		cfg.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
 		cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	} else { // Json
+		if logTimestampFormat == "ISO8601" {
+			cfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+		} else {
```
jon-whit, line 133: Redundant else. Just set the default and then override if `ISO8601`.

pkg/logger/logger.go
```diff
@@ -93,7 +93,7 @@ func NewNoopLogger() *ZapLogger {
 	}
 }
 
-func NewLogger(logFormat, logLevel string) (*ZapLogger, error) {
+func NewLogger(logFormat, logLevel, logTimestampFormat string) (*ZapLogger, error) {
```
jon-whit, line 96: We should refactor this signature to use the options builder pattern and if an option is not provided we just default it.

For example,

```go
NewLogger(
  WithFormat("text"),
  WithLevel("info"),
  WithTimestampFormat("ISO8601"),
)
```

jon-whit: Added issue #1355 to follow up with this suggestion.

jon-whit:
@raj-saxena left one last comment. Looks good otherwise. Thanks!

raj-saxena:
Updated the PR @jon-whit. Let's move this forward 🙏🏼

jon-whit (approved):

miparnisari (approved):

PR #1331, opened 2024-01-31
Adding tracing + attributes to all resolvers

willvedd:
## Description
Improving observability of the check resolution process by adding tracing and an appropriate set of attributes for each resolver. Previously, only the local check resolver had tracing but now that we have three separate resolvers, it makes sense to observe those too. 

Admittedly, the `resolver_type` attribute is not a best practice and should have been accomplished by the built-in `package` attribute but that would have required a messy refactor.

## Review Checklist

### DIFF
--- a/internal/graph/cached_resolver.go
+++ b/internal/graph/cached_resolver.go
@@ -10,6 +10,7 @@ import (
 	"github.com/karlseguin/ccache/v3"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 
 	"github.com/openfga/openfga/internal/build"
@@ -150,6 +151,11 @@ func (c *CachedCheckResolver) ResolveCheck(
 	ctx context.Context,
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	defer span.End()
+	span.SetAttributes(attribute.String("resolver_type", "cache"))
+	span.SetAttributes(attribute.String("tuple_key", req.GetTupleKey().String()))
+
 	checkCacheTotalCounter.Inc()
 
 	cacheKey, err := CheckRequestCacheKey(req)
@@ -161,8 +167,10 @@ func (c *CachedCheckResolver) ResolveCheck(
 	cachedResp := c.cache.Get(cacheKey)
 	if cachedResp != nil && !cachedResp.Expired() {
 		checkCacheHitCounter.Inc()
+		span.SetAttributes(attribute.Bool("is_cached", true))
 		return cachedResp.Value().convertToResolveCheckResponse(), nil
 	}
+	span.SetAttributes(attribute.Bool("is_cached", false))
 
 	resp, err := c.delegate.ResolveCheck(ctx, req)
 	if err != nil {
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -458,7 +458,7 @@ func (c *LocalChecker) ResolveCheck(
 
 	ctx, span := tracer.Start(ctx, "ResolveCheck")
 	defer span.End()
-
+	span.SetAttributes(attribute.String("resolver_type", "local"))
 	span.SetAttributes(attribute.String("tuple_key", req.GetTupleKey().String()))
 
 	if req.GetResolutionMetadata().Depth == 0 {
--- a/internal/graph/singleflight_resolver.go
+++ b/internal/graph/singleflight_resolver.go
@@ -3,6 +3,7 @@ package graph
 import (
 	"context"
 
+	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 	"golang.org/x/sync/singleflight"
 
@@ -53,6 +54,11 @@ func (s *singleflightCheckResolver) ResolveCheck(
 		return nil, err
 	}
 
+	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	defer span.End()
+	span.SetAttributes(attribute.String("resolver_type", "singleflight"))
+	span.SetAttributes(attribute.String("tuple_key", req.GetTupleKey().String()))
+
 	isUnique := false
 	singleFlightResp, err, shared := s.group.Do(key, func() (interface{}, error) {
 		isUnique = true
@@ -77,6 +83,9 @@ func (s *singleflightCheckResolver) ResolveCheck(
 		deduplicatedDBQueryCount.Add(float64(resp.GetResolutionMetadata().DatastoreQueryCount))
 		resp.ResolutionMetadata.DatastoreQueryCount = 0
 	}
+
+	span.SetAttributes(attribute.Bool("is_shared_request", shared))
+	span.SetAttributes(attribute.Bool("is_ultimately_processed_request", isUnique))
 
 	return &resp, err
 }


internal/graph/cached_resolver.go
```diff
@@ -150,6 +151,11 @@ func (c *CachedCheckResolver) ResolveCheck(
 	ctx context.Context,
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	defer span.End()
+	span.SetAttributes(attribute.String("resolver_type", "cache"))
```
jon-whit, line 156:
```suggestion
	span.SetAttributes(attribute.String("resolver_type", "CachedCheckResolver"))
```

internal/graph/check.go
```diff
@@ -458,7 +458,7 @@ func (c *LocalChecker) ResolveCheck(
 
 	ctx, span := tracer.Start(ctx, "ResolveCheck")
 	defer span.End()
-
+	span.SetAttributes(attribute.String("resolver_type", "local"))
```
jon-whit, line 461:
```suggestion
	span.SetAttributes(attribute.String("resolver_type", "LocalChecker"))
```

internal/graph/singleflight_resolver.go
```diff
@@ -53,6 +54,11 @@ func (s *singleflightCheckResolver) ResolveCheck(
 		return nil, err
 	}
 
+	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	defer span.End()
+	span.SetAttributes(attribute.String("resolver_type", "singleflight"))
```
jon-whit, line 59:
```suggestion
	span.SetAttributes(attribute.String("resolver_type", "SingleflightCheckResolver"))
```

internal/graph/singleflight_resolver.go
```diff
@@ -78,6 +84,9 @@ func (s *singleflightCheckResolver) ResolveCheck(
 		resp.ResolutionMetadata.DatastoreQueryCount = 0
 	}
 
+	span.SetAttributes(attribute.Bool("is_shared_request", shared))
```
jon-whit, line 87: Can we rename to `deduplicated`?

willvedd: `shared` != deduplicated, it also includes the unique request that was ultimately resolved. 

If we want to know if a request was deduplicated, we would check `shared && !unique`. There are three states to communicate:
- Not shared, resolved as normal
- Shared and ultimately resolved
- Shared but deduplicated

Instead of communicating this as two booleans, how do you feel about combining into a single field?

jon-whit: 🤔 I guess `deduplicated` may imply the work encapsulated by this span was de-duplicated, but it actually wasn't unless `isUnique` is true.

So I think I'd lean toward setting an attr `deduplicated: false` if `isUnque==true` and setting `deduplicated: true` if `isUnique == false && shared`. Just capture the semantics of the de-dupe in a single attr.

internal/graph/singleflight_resolver.go
```diff
@@ -78,6 +84,9 @@ func (s *singleflightCheckResolver) ResolveCheck(
...
+	span.SetAttributes(attribute.Bool("is_ultimately_processed_request", isUnique))
```
jon-whit, line 88: What's the importance of this for our usage in our telemetry? Can we approach this with the suggestion in https://github.com/openfga/openfga/pull/1331/files#r1474847198 instead?

internal/graph/cached_resolver_test.go
```diff
@@ -47,10 +47,10 @@ func TestResolveCheckFromCache(t *testing.T) {
 				TupleKey:             tuple.NewTupleKey("document:abc", "reader", "user:XYZ"),
 			},
 			setInitialResult: func(mock *MockCheckResolver, request *ResolveCheckRequest) {
-				mock.EXPECT().ResolveCheck(ctx, request).Times(1).Return(result, nil)
+				mock.EXPECT().ResolveCheck(gomock.Any(), request).Times(1).Return(result, nil)
```
willvedd, line 50: Had a tough time trying to get this pass with the tracer wrapping the context. Know `gomock.Any()` isn't a strong assertion but the context does not play a critical role in these tests anyway.

internal/graph/singleflight_resolver.go
```diff
@@ -90,3 +98,14 @@ func copyResolveResponse(original ResolveCheckResponse) ResolveCheckResponse {
 		},
 	}
 }
+
+func singleflightRequestStateAttribute(isShared bool, isUnique bool) attribute.KeyValue {
```
willvedd, line 102: I separated this function out to not convolute the inside of the resolver itself; it became messy with the nested if/elses.

willvedd:
**Demo:**
_Note: traces show that nested resolver delegation doesn't work as intended. Behavior currently depends if cached is enabled or not. This will be fixed in future PR. These are just to prove that traces are working._


**Cache disabled:**
Screenshot 2024-02-02 at 11 29 25 AM


**Cache enabled:**
Screenshot 2024-02-02 at 11 16 33 AM

internal/graph/cached_resolver_test.go
```diff
@@ -397,6 +397,8 @@ func TestResolveCheckExpired(t *testing.T) {
 	defer ctrl.Finish()
 
 	ctx := context.Background()
+
+	gomock.AssignableToTypeOf(ctx)
```
miparnisari, line 401: ??

internal/graph/singleflight_resolver.go
```diff
@@ -53,6 +54,11 @@ func (s *singleflightCheckResolver) ResolveCheck(
 		return nil, err
 	}
 
+	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	defer span.End()
+	span.SetAttributes(attribute.String("resolver_type", "SingleflightCheckResolver"))
+	span.SetAttributes(attribute.String("tuple_key", req.GetTupleKey().String()))
+
```
miparnisari, line 61: Why is this new span being created after the `if`? in the Cached resolver, it's right at the start. 


If you put it at the start, it gives you the opportunity of marking the spans as errored if there was an error.

```go
        ctx, span := tracer.Start(ctx, "ResolveCheck")
	defer span.End()
	span.SetAttributes(attribute.String("resolver_type", "SingleflightCheckResolver"))
	span.SetAttributes(attribute.String("tuple_key", req.GetTupleKey().String()))
	
	key, err := CheckRequestCacheKey(req)
	if err != nil {
		telemetry.TraceError(span, err) // <------ this line
		s.logger.Error("singleflight cache key computation failed with error", zap.Error(err))
		return nil, err
	}
```

That aside, we should also be doing `telemetry.TraceError` here:

```go
    singleFlightResp, err, shared := s.group.Do(key, func() (interface{}, error) {
		// ...
	})
	if err != nil {
		return nil, err
	}
```

So that this

image

becomes this

image

internal/graph/cached_resolver.go
```diff
@@ -161,8 +167,10 @@ func (c *CachedCheckResolver) ResolveCheck(
 	cachedResp := c.cache.Get(cacheKey)
 	if cachedResp != nil && !cachedResp.Expired() {
 		checkCacheHitCounter.Inc()
+		span.SetAttributes(attribute.Bool("is_cached", true))
 		return cachedResp.Value().convertToResolveCheckResponse(), nil
 	}
+	span.SetAttributes(attribute.Bool("is_cached", false))
 
 	resp, err := c.delegate.ResolveCheck(ctx, req)
 	if err != nil {
```
miparnisari, line 176:
```suggestion
	resp, err := c.delegate.ResolveCheck(ctx, req)
	if err != nil {
	     telemetry.TraceError(span, err)
	     return nil, err
```

internal/graph/singleflight_resolver.go
willvedd: No particular reason for placement but great suggestion, will change.

internal/graph/singleflight_resolver.go
```diff
@@ -58,11 +65,13 @@ func (s *singleflightCheckResolver) ResolveCheck(
 		isUnique = true
 		resp, err := s.delegate.ResolveCheck(ctx, req)
 		if err != nil {
+			telemetry.TraceError(span, err)
```
miparnisari, line 68: nit: this one is redundant, it will be caught by 74 :)

miparnisari (approved):

internal/graph/singleflight_resolver.go
willvedd: Yes, good catch. Also noticed that the resolver state should be moved up before the error check so that it can properly set that info on the trace if the error is non-nil.

miparnisari (approved):

PR #1335, opened 2024-02-02
fix: wire up SingleflightCheckResolver properly in Server

jon-whit:
## Description
Prior to this PR if an external library constructed a new OpenFGA Server with [NewServerWithOpts](https://pkg.go.dev/github.com/openfga/openfga@v1.4.3/pkg/server#NewServerWithOpts) and provided the [WithCheckQueryCacheEnabled(true)](https://pkg.go.dev/github.com/openfga/openfga@v1.4.3/pkg/server#WithCheckQueryCacheEnabled), then the Server would bypass the [SingleflightCheckResolver](https://github.com/openfga/openfga/blob/f2ddc07cf598c3bae174012cf8cbaa2d87cc164f/internal/graph/singleflight_resolver.go#L36) and go straight to the [CachedCheckResolver](https://github.com/openfga/openfga/blob/f2ddc07cf598c3bae174012cf8cbaa2d87cc164f/internal/graph/cached_resolver.go#L65).

The problem stemmed from the fact that the way we were composing our CheckResolver delegates assumed two delegates at most and so we could create a delegate chain. What we want is:

`SingleflightCheckResolver` 
--> `CachedCheckResolver`
&nbsp;&nbsp;&nbsp;--> `LocalChecker`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;--> `SingleflightCheckResolver`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;--> etc...

Doing that was not going to be possible with the builder pattern composition we previously had, and so I took the liberty to make the delegate composition more explicit with `.SetDelegate` pattern. 

Additionally, I noticed that if we construct a new `LocalChecker` on every ListObjects and Check request, then that would imply we would have to construct a new `SingleflightCheckResolver` on every request as well, because otherwise we'd be mutating the `SingleflightCheckResolver`'s delegate to a new object reference on every `CheckRequest` and they would overlap with one another and cause race conditions. Not to mention, we don't get cross-request de-duplication if we construct a `SingleflightCheckResolver` per request. So I also took the liberty to wire up a single `CheckResolver` that is shared across ListObjects and Check at the server level. The server has a single `CheckResolver` instance (default outer most delegate is the singleflight resolver) and it is used anywhere we have to resolve a Check query.

☝️ implies that we must be able to contextually construct a `storage.RelationshipTupleReader` per `ResolveCheck` query. Previously we would construct a new `LocalChecker` per request and pass in a new `RelationshipTupleReader` with the bounded concurrency wrapper and contextual tuples storage wrapper, but since we're now sharing a single `LocalChecker` reference I took the liberty to pass this per-request specific object reference in through the context, like we're doing with the TypeSystem, and then extracting it in `ResolveCheck` and using it appropriately. I recognize this usage of context may be controversial but a) it was the easiest way for me to achieve the goal at hand without much larger refactors and function signature changes b) it's inline with other usages of context passing today for methods that share a single object reference but whose behavior changes based on parent request behavior (e.g. setting per-request concurrency limits or per-request contextual tuples and the like..). So these were the two primary motivators for the reason behind why I did what I did. This can be revisited if we want, but these changes fix the problem at hand and also will improve our deduplication across the whole server (not just per request). So please don't fixate on the usage of context so much.

## References
- Original `SingleFlightCheckResolver` PR: #1313 

## Review Checklist


assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -4772,253 +4772,6 @@ tests:
               relation: owner
             expectation:
               - repo:1
-      - model: | #complex model
```
jon-whit, line 4775: Fun fact, this sub-stage within this test fails currently on `main` if you comment out the two stages above it. We have something weird going on with our multi-stage tests in `tests/check` and `tests/listobjects`. I didn't have time to investigate the full issue.

That being said, this test is actually invalid, because when we run this test with passing all of these tuples as contextual tuples, the number of tuples here well exceeds the 20 tuple limit for contextual tuples (see code links below). As a result of this, I decided to remove this test from the suite. It may be pertinent to add it as just a regular test somewhere else, but it doesn't belong here because it's invalid.

https://github.com/openfga/openfga/blob/main/tests/listobjects/listobjects.go#L105

https://github.com/openfga/openfga/blob/main/tests/listobjects/listobjects.go#L177

internal/graph/check.go
```diff
@@ -123,7 +122,6 @@ type checkOutcome struct {
 }
 
 type LocalChecker struct {
-	ds                 storage.RelationshipTupleReader
```
jon-whit, line 126: Not needed as we contextuall provide the datastore per ResolveCheck. Each time ResolveCheck is invoked we may be dealing with a) different contextual tuples and thus a CombinedTupleReader with a different set of tuples from the parent Check or b) a new incoming request which implies that we need to had a BoundedConcurrencyTupleReader that resets at the per-request Check concurrency limit.

pkg/server/server.go
```diff
@@ -298,23 +298,30 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		opt(s)
 	}
 
-	s.checkOptions = []graph.LocalCheckerOption{
+	singleflightCheckResolver := graph.NewSingleflightCheckResolver()
```
jon-whit, line 301: This is where the explicit delegate composition takes place with explicit `.SetDelegate` calls. It makes it abundantly clear which resolver is delegate to the other.

pkg/server/server_test.go
```diff
@@ -236,6 +237,17 @@ func TestServerWithMySQLDatastoreAndExplicitCredentials(t *testing.T) {
 	test.RunAllTests(t, ds)
 }
 
+func TestSingleflightCheckResolverDefault(t *testing.T) {
```
jon-whit, line 240: This will save our bacon because we can make sure now that the entrypoint for `Server#Check` starts at a SingleflightCheckResolver. This test gives us the fidelity we would have liked to have had prior when we were constructing a LocalChecker per request and therefore couldn't easily determine the precise resolver entrypoint the server was going to make when evaluating a Check.

willvedd:
FWIW, I ran the tracing proposed in #1331 against this branch and the result proves that the nested delegation works as expected:

Screenshot 2024-02-02 at 12 20 49 PM

assets/tests/consolidated_1_1_tests.yaml
miparnisari: The test is not invalid, please do not remove it. It failed when you commented out the previous stages because that triggers the contextualTuple harness, which fails in this case because we didn't account for the 20 tuple limit that this test in particular exceeds. Please see the fix in https://github.com/openfga/openfga/pull/1336

pkg/server/server.go
```diff
@@ -332,9 +339,14 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
 func (s *Server) Close() {
-	if s.checkCache != nil {
-		s.checkCache.Stop()
+	if s.cachedCheckResolverCloser != nil {
+		s.cachedCheckResolverCloser()
 	}
+
+	if s.checkResolver != nil {
+		s.checkResolver.Close()
+	}
```
miparnisari, line 348:
```suggestion
	if s.checkQueryCacheEnabled {
		s.cachedCheckResolverCloser()
		s.checkResolver.Close()
	}
```

pkg/server/server.go
```diff
@@ -370,25 +382,14 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
...
-			graph.WithExistingCache(s.checkCache),
-			graph.WithCacheTTL(s.checkQueryCacheTTL),
-			graph.WithLogger(s.logger),
-		))
-	}
-
-	q := commands.NewListObjectsQuery(s.datastore,
+	q := commands.NewListObjectsQuery(
+		s.datastore,
+		s.checkResolver,
```
miparnisari, line 387: i like this

assets/tests/consolidated_1_1_tests.yaml
willvedd: If I'm reading correctly, this is just an extra change thrown into the PR. Invalid or not, I'd prefer we keep limit the scope of this PR to the check resolver delegation.

pkg/server/server.go
```diff
@@ -620,14 +610,17 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
-		storagewrappers.NewCombinedTupleReader(s.datastore, req.ContextualTuples.GetTupleKeys()),
-		s.checkOptions...,
+	ctx = storage.ContextWithRelationshipTupleReader(ctx,
+		storagewrappers.NewBoundedConcurrencyTupleReader(
+			storagewrappers.NewCombinedTupleReader(
+				s.datastore,
+				req.GetContextualTuples().GetTupleKeys(),
+			),
+			s.maxConcurrentReadsForCheck,
+		),
```
miparnisari, line 620: There is no reason to do this setup here. It can happen inside `LocalChecker.ResolveCheck`. LocalChecker already has `maxConcurrentReadsForCheck` stored internally. Again, https://github.com/openfga/openfga/pull/948 :)

assets/tests/consolidated_1_1_tests.yaml
miparnisari: I agree, but i add that this test is valid, if it failed in the scope of building this PR i'd like to know that

internal/graph/check.go
```diff
@@ -448,6 +432,14 @@ func (c *LocalChecker) dispatch(_ context.Context, req *ResolveCheckRequest) Che
 
 // ResolveCheck resolves a node out of a tree of evaluations. If the depth of the tree has gotten too large,
 // evaluation is aborted and an error is returned. The depth is NOT increased on computed usersets.
+//
+// It is expected that callers pass in, contextually, a [[storage.RelationshipTupleReader]] using
```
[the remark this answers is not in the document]
willvedd: I agree that we should not use context for passing-in the tuple reader dependency, the resulting code obscurity is not a worthy tradeoff.

internal/graph/check.go
```diff
@@ -174,11 +157,12 @@ func NewLocalChecker(ds storage.RelationshipTupleReader, opts ...LocalCheckerOpt
...
+	return checker
+}
 
-	// Depending on whether cached check resolver is used,
-	// we either return the newly created checker or the delegate (i.e., cached check resolver).
-	return checker.delegate
+// SetDelegate sets this LocalChecker's dispatch delegate.
+func (c *LocalChecker) SetDelegate(delegate CheckResolver) {
+	c.delegate = delegate
 }
```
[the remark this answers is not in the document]
willvedd: `SetSubproblemDelegate` is certainly more descriptive. Though when talking about delegation in the context of the codebase, there is no ambiguity. 

Even though the local checker is the smartest of them all, it still delegates its work in many cases.

willvedd (commented):
See three main things bundled:

**1.** Explicit delegation definition of resolvers to fix `SingleflightCheckResolver` and promote sustainable composition practices in the future.
**2.** A single, shared check resolver attached to the server to be shared across all requests. This is in contrast of instantiating a new check resolver per request.
**3.** Passing-in `RelationshipTupleReader` dependency via context where appropriate to facilitate point 2.


Points 1 and 2 are great, no objections.

Point 3 seems to enabled point 2 but is a bit controversial; I do not necessarily agree with the implementation. For this reason, I think this PR should be split into at least two, one to establish the `SetDelegate` pattern, which is most pressing IMO. The other PR would be to share a single check resolver instance across the codebase, much deserved but a little less pressing.

pkg/server/server_test.go
willvedd: Love this test, simple yet effective. Though I think it should be recontextualized to be about the composition rather than the 

In practice, I see this as renaming to `TestCheckResolverComposition` and then checking the type of the top-level delegate and it's chain of sub-delegates for both a cache and uncached server configuration.

pkg/server/server.go
```diff
@@ -620,14 +610,17 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 	}
 
 	ctx = typesystem.ContextWithTypesystem(ctx, typesys)
-
-	checkResolver := graph.NewLocalChecker(
-		storagewrappers.NewCombinedTupleReader(s.datastore, req.ContextualTuples.GetTupleKeys()),
-		s.checkOptions...,
+	ctx = storage.ContextWithRelationshipTupleReader(ctx,
+		storagewrappers.NewBoundedConcurrencyTupleReader(
+			storagewrappers.NewCombinedTupleReader(
```
miparnisari, line 615: shouldn't we invert the order of these? otherwise i think we will try to bound concurrency for the contextual tuples 🤔

assets/tests/consolidated_1_1_tests.yaml
willvedd: I'm not observing the test failure. I say we should revert for this PR.

pkg/server/server_test.go
willvedd: I'll hold a vote on what the name should be. Until that is concluded, I don't think this PR is an appropriate forum to discuss renaming.

pkg/server/test/list_objects.go
```diff
@@ -491,25 +490,27 @@ condition condition1(x: int) {
...
-			checkOptions := []graph.LocalCheckerOption{
+			var checkResolver graph.CheckResolver
+			localCheckResolver := graph.NewLocalChecker(
 				graph.WithResolveNodeBreadthLimit(100),
 				graph.WithMaxConcurrentReads(30),
-			}
+			)
+			checkResolver = localCheckResolver
 
 			if test.useCheckCache {
```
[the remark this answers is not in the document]
willvedd: Good catch. I've created a ticket in the backlog that captures the information in your comment.

pkg/server/commands/list_objects.go
```diff
@@ -118,7 +116,7 @@ func NewListObjectsQuery(ds storage.RelationshipTupleReader, opts ...ListObjects
 		resolveNodeLimit:        serverconfig.DefaultResolveNodeLimit,
 		resolveNodeBreadthLimit: serverconfig.DefaultResolveNodeBreadthLimit,
 		maxConcurrentReads:      serverconfig.DefaultMaxConcurrentReadsForListObjects,
-		checkOptions:            []graph.LocalCheckerOption{},
+		checkResolver:           checkResolver,
```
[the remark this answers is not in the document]
jon-whit: Practical usages of this never allow for `checkResolver` to be nil. We always pass the server's check resolver, which is instantiated at server startup. See usages of `NewListObjectsQuery`. Adding a test for it and instantiating a default check resolver would solely be for testing purposes itself. What I'll do is change the signature for `NewListObjectsQuery` so that it returns an error, and if the datastore or check resolver are nil then we can return an error.

I think this raises a point though - commands should be internal packages, not part of `pkg/server`. A Go project shouldn't be importing the commands, they should be importing the Server and invoking the RPC handlers on it.

internal/graph/check_test.go
```diff
@@ -376,9 +397,16 @@ type document
 					t.Parallel()
 
 					checker := NewLocalChecker(
-						// TODO build this wrapper inside ResolveCheck so that we don't need to construct a new Checker per test
```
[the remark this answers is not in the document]
willvedd: As I'm reading it, this TODO is no longer resolvable because the check resolvers are instantiated at the server-level now. It wouldn't make sense to construct inside `ResolveCheck`. However it does appear that the `NewLocalChecker` call here might be able to be pulled outside the loop, but I haven't looked at the test itself super closely.

willvedd (commented):
**Update:** I've tried myself to implement this without passing-in tuple reader via context and it was a huge PITA. Considering further, using context in this way is not a black/white thing, and this particular usage has its fair share of upsides. I think for the sake of preventing a major refactoring overhaul and keeping the function signatures clean for a little, we keep this PR largely as-is. We can evaluate the tradeoffs and review our usage of context and maybe set some guidelines. 

But aside from that and a few minor pieces of feedback above, I intend to approve.

pkg/server/commands/list_objects.go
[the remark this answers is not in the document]
jon-whit: I have changed the signature and added a unit test to account for the specific case where you pass in nil, but to the point above that shouldn't be exposed to the public anyways.

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: They all delegate. It makes no difference who is the delegate vs the delegator. The only difference is the LocalChecker has more responsibility of the core algorithm if it is the delegate that gets invoked. For this reason, and for uniformity across the CheckResolver composition, I think it makes sense to leave uniformly as `SetDelegate`.

internal/graph/cached_resolver_test.go
```diff
@@ -374,17 +374,20 @@ func TestResolveCheckFromCache(t *testing.T) {
...
 
-			dut2 := NewCachedCheckResolver(dut, WithExistingCache(dut.cache))
+			dut2 := NewCachedCheckResolver(WithExistingCache(dut.cache))
 			defer dut2.Close()
 
+			dut2.SetDelegate(dut)
+
 			actualResult, err := dut2.ResolveCheck(ctx, test.subsequentReq)
 			require.Equal(t, result.Allowed, actualResult.Allowed)
 			require.NoError(t, err)
```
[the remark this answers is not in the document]
jon-whit: I suppose this test ensures that if you pass in `WithExistingCache` it actually uses it instead of constructing an internal one and using it instead AND that we return the previously computed cache result. It's possible for a bug to creep into the CachedCheckResolver where the `dut2` could return `allowed: false` even though the initial request returned `allowed: true`. It's unlikely, but I guess that's what we're protected against.

The `test.setTestExpectations(mockResolver, test.subsequentReq)` ensures we don't end up calling `ResolveCheck` a subsequent time, but it doesn't ensure that we get the prior computed result back. So having this covers our bases a bit better.

pkg/server/server.go
jon-whit: The check `s.cachedCheckResolverCloser != nil` and `s.checkResolver != nil` are safer. Assuming a bug upstream, for example if we didn't construct the check resolver or check resolver closer properly when `checkQueryCacheEnabled=true`, then that wouldn't be a safer bet.

If the query cache is not enabled both _should be_ nil, but a direct check on nil is safer.

pkg/server/server.go
jon-whit: No, because the counded concurrency tuple reader just bounds concurrent calls to the storage methods. The combined reader just merges the remote storage with local in-memory contextual tuples. The outer layer should be the bounded concurrency reader. This is how it is today.

We construct the BoundedConcurrencyTupleReader around what we pass into `NewLocalChecker`, and we pass in the CombinedTupleReader into `NewLocalChecker`. See:
* https://github.com/openfga/openfga/blob/main/pkg/server/server.go#L625
* https://github.com/openfga/openfga/blob/main/internal/graph/check.go#L177

So effectively what we have is:

```
NewBoundedConcurrencyTupleReader(
    NewCombinedTupleReader(baseDatastore, req.GetContextualTuples()),
)
```

These changes mimic that behavior in a functionally equivalent manner.

pkg/server/server.go
jon-whit: I provided the context for this design decision in the original PR description, but I'll reiterate the main motivation for this pattern:

1. It was the easiest way for me to achieve the goal at hand without much larger refactors and function signature changes
2. It's inline with other usages of context passing today for methods that share a single object reference but whose behavior changes based on parent request behavior (e.g. setting per-request concurrency limits or per-request contextual tuples or changing behavior based on baggage propagation and the like..)
3. It hinders intentional dispatch composition design objectives (see below).


If we change the signature we hinder more uniform dispatch composition. You can't pass Typesystem or an OpenFGADatastore reference across an RPC boundary, and so doing a refactor such as the one suggested would be hurting the very intentional uniform loopback design. Not all CheckResolver implementations are localized in-memory function calls, because some may delegate externally over an RPC implementation. So for those that delegate locally we can modify the behavior contextually.

internal/graph/check_test.go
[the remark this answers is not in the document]
jon-whit: @willvedd is correct. The CheckResolver is instantiated at the top-level, not per request and so the commented TODO is no longer applicable. I have moved the LocalChecker construction up a level in the test though (as Will pointed out).

assets/tests/consolidated_1_1_tests.yaml
jon-whit: Something odd is going on, I'll investigate further. I'll revert the change. We should follow up on this though, because it is an invalid test. Maybe validation or something else is not working somewhere in the call stack.

jon-whit:
Here's a simple test scenario that demonstrates the issue with deadlocking with Singleflight and why cycle detection doesn't work. Consider the following model, tuples, and Check query.

```
type user

type document
  relations
    define parent: [document]
    define owner: [user] or owner from parent
    define viewer: owner or viewer from parent

- document:1#owner@user:anne
- document:2#parent@document:1
- document:1#parent@document:2
```

If you call `Check(document:1#viewer@user:anne)` you get the following callstack:
```
Check(document:1#viewer@user:anne)
|--> SingleflightCheckResolver#ResolveCheck(document:1#viewer@user:anne)
|-----> LocalChecker#ResolveCheck(document:1#viewer@user:anne) visited=["document:1#viewer@user:anne"]
|--------> union
|-----------> LocalChecker#checkComputedUserset(document:1#owner@user:anne)
|--------------> LocalChecker#dispatch(document:1#owner@user:anne)
|-----------------> SingleflightCheckResolver#ResolveCheck(document:1#owner@user:anne) (1a)
|--------------------> LocalChecker#ResolveCheck(document:1#owner@user:anne) visited=["document:1#viewer@user:anne", "document:1#owner@user:anne"]
|-----------------------> union
|--------------------------> LocalChecker#checkDirect(document:1#owner@user:anne)
|-----------------------------> storage#ReadUserTuple(document:1#owner@user:anne) -> ["document:1#owner@user:anne"]
|-----------------------------> return {allwoed: true}
|--------------------------> LocalChecker#checkTTU(document:1#owner@user:anne)
|-----------------------------> storage#Read(document:1#parent) --> ["document:2"]
|-----------------------------> LocalChecker#dispatch(document:2#owner@user:anne)
|--------------------------------> SingleflightCheckResolver#ResolveCheck(document:2#owner@user:anne) (2a)
|-----------------------------------> LocalCheckResolver#ResolveCheck(document:2#owner@user:anne) visited=["document:1#viewer@user:anne", "document:1#owner@user:anne", "document:1#owner@user:anne"]
|--------------------------------------> union
|-----------------------------------------> LocalCheckResolver#checkDirect(document:2#owner@user:anne)
|--------------------------------------------> storage#ReadUserTuple(document:2#owner@user:anne) --> ["document:2#owner@user:anne"]
|--------------------------------------------> return {allowed: true}
|
|-----------------------------------------> LocalCheckResolver#checkTTU(document:2#owner@user:anne)
|--------------------------------------------> storage#Read(document:2#parent) --> []
|--------------------------------------------> return {allowed: false}
|--------------------------------------> return {allowed: true}
|--------------------------> return {allowed: true}
|
|-----------> LocalChecker#checkTTU(document:1#viewer@user:anne)
|--------------> storage#Read(document:1#parent) --> ["document:2"]
|--------------> LocalChecker#dispatch(document:2#viewer@user:anne)
|-----------------> SingleflightCheckResolver#ResolveCheck(document:2#viewer@user:anne)
|--------------------> LocalChecker#ResolveCheck(document:2#viewer@user:anne)  visited=["document:1#viewer@user:anne", "document:2#viewer@user:anne"]
|-----------------------> union
|--------------------------> LocalChecker#checkComputedUserset(document:2#owner@user:anne)
|-----------------------------> LocalChecker#dispatch(document:2#owner@user:anne)
|--------------------------------> SingleflightCheckResolver#ResolveCheck(document:2#owner@user:anne) (2b)
|-----------------------------------> LocalChecker#ResolveCheck(document:2#owner@user:anne) visited=["document:1#viewer@user:anne", "document:2#viewer@user:anne", "document:2#owner@user:anne"]
|--------------------------------------> union
|-----------------------------------------> LocalChecker#checkDirect(document:2#owner@user:anne)
|--------------------------------------------> storage#ReadUserTuple(document:2#owner@user:anne)
|--------------------------------------------> return {allowed: false}
|-----------------------------------------> LocalChecker#checkTTU(document:2#owner@user:anne)
|--------------------------------------------> storage#Read(document:2#parent) --> ["document:1"]
|--------------------------------------------> LocalChecker#dispatch(document:1#owner@user:anne)
|-----------------------------------------------> SingleflightCheckResolver#ResolveCheck(document:1#owner@user:anne) (1b)
|--------------------------> LocalChecker#checkTTU(document:2#viewer@user:anne)
|
|-----> return {allowed: true}

|--> return {allowed: true}
```

I've omitted some of the branches as they are irrelevant for proving the point, but as you can see, we don't necessarily see a cycle in either of the invidual branches of the top-most union (or any of the subproblems therein). However, based on dispatch loopback, if (1a) wins the race to `SingleflightCheckResolver#ResolveCheck` and (2b) wins the race (in the other branch), then (1a) and (2b) are locked co-dependently upon each other and cannot resolve. Put differently, the top-level union cannot resolve because both fo the two bfranches are co-dependent upon resolution relative to one another (e.g. a deadlock).

pkg/server/server.go
miparnisari: > This is how it is today.

So this is broken in main today too?

pkg/server/server.go
miparnisari: We don't have to change the signature in this PR.

That being said, and i'm sorry to beat a dead horse here but I want to stop using the context for everything!

> It was the easiest way for me to achieve the goal at hand without much larger refactors and function signature changes

I agree

> It's inline with other usages of context passing today

I agree

> It hinders intentional dispatch composition design objectives (see below). If we change the signature we hinder more uniform dispatch composition. 

Do you really believe that just because we have this ResolveCheck signature we can easily introduce remote dispatching? 😆  We will need a **ton** more things, and if this PR shows anything, is that we will probably need to pass other metadata around for cycle detection.

> Not all CheckResolver implementations are localized in-memory function calls

You used this same argument a year ago and here we are, they are all in memory still.

internal/graph/singleflight_resolver.go
```diff
@@ -2,10 +2,12 @@ package graph
 
 import (
 	"context"
+	"sync/atomic"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
-	"golang.org/x/sync/singleflight"
+
+	"resenje.org/singleflight"
```
jon-whit, line 10: This change was actually critical to fix the broken tests across the board. In [resenje.org/singleflight#Group.Do](https://pkg.go.dev/resenje.org/singleflight#Group.Do) 

> if there are multiple callers, context passed to one caller does not affect the execution and returned values of others except if the function result is dependent on the context values.

This ☝️  was the exact problem manifesting itself. We'd de-duplicate with different contexts passed in (see [this test](https://github.com/openfga/openfga/pull/1335/files#diff-7031b8db92a3ea494a5f79c70542afa86c66750a16601d78a4cdedf3bddc4536R154)) at the top-level and yet one would return a context cancellation error. 

The regular [x/sync/singleflight#Group.Do](https://pkg.go.dev/golang.org/x/sync/singleflight#Group.Do) creates a closure over the context that wins the singleflight race, and thus causes some perplexing context cancellation errors in the de-duplicated paths.

The `resenje.org/singleflight` package avoids this issue with better support for context for singleflight invocations. Notice that we pass in `ctx` to `Group.Do(ctx, ...)` but the function that gets executed once for each of the shared invocations gets it's own unique `innerCtx context.Context`.

```
Do(ctx, ..., func(innerCtx context.Context) ... {
    _ = innerCtx // we use innerCtx to avoid shared closure over `ctx`
})
```

pkg/server/server.go
jon-whit: I don't understand why it's broken. Please elaborate more with an example or test if possible. That would help direct the conversation more concretely.

The way I see it is that the inner combined tuple reader just merges in the contextual tuples, but the bounded concurrency reader limits how many invocations of `Read`, `ReadUserTuple`, `ReadUsersetTuples`, etc.. can get called concurrently. I don't see that being an issue. The combined tuple reader still has to hit the database to join the tuples in remote storage with the contextual tuples in memory. So you still want to bound that concurrency to the remote storage layer.

pkg/server/server.go
jon-whit: @miparnisari  I know you deeply care about the quality of our code, the maintainability and testability of it, etc... You are arguably the most influential backend member keeping us inline with higher standards of code quality and practices, and I appreciate that. It's easy to let those things slip through the cracks at times.

I recognize you have distaste for how our CheckResolver composition and RPC motivated dispatch composition works in our implementation today. And I also recognize it's probably obnoxious to continue to hear things like "when we do remote dispatching in the near future". But we've made various design decisions around various things including scalability and consistency tradeoffs and other architecture bits (both in OpenFGA and Okta FGA) for future thinking reasons that we have little to no practical reason for at this time. I'd be happy to enumerate that list of things further 1:1. 

But, to focus on a constructive feedback loop here, I guess I'd ask that if there is so much distaste for the current state of the code, then please take the liberty to discuss what you'd like to fix, how you'd like to fix it (maybe some sample code snippets and such), and brief reason why. I don't think myself or anyone else on the @openfga/backend team would oppose you on that, but we need to focus on constructive progress and feedback here, otherwise we'll keep beating the dead horse. Am I generally opposed to going in and refactoring the Check code and breaking our ability to more easily do remote dispatch? Yes. But am I willing to budge on that? Yes. We have yet to have an actual constructive conversation discussing _what_ we would do differently. Let's start there.

> Do you really believe that just because we have this ResolveCheck signature we can easily introduce remote dispatching? 😆

I do believe so, yes. In fact, my [dispatch-demo](https://github.com/jon-whit/dispatch-demo) demonstrated to myself that it _would_ be quite straightforward of a drop in (compositionally speaking). Figuring out how to do peer discovery and consistent hashing and all of that other stuff that comes with remote dispatch will obviously have it's own set of challenges. I have never said implementing remote dispatching will be easy, but I still stand by my statement that with our current compositional structure it will be much <u>easier</u> to make that move. That POC proved it to me, and I'd encourage everyone on the @openfga/backend team to go and take a look at that code and get an idea of where I'm coming from with the compositional design.

And to your other point,
> We will need a ton more things, and if this PR shows anything, is that we will probably need to pass other metadata around for cycle detection.

If we continue to keep an eye on what composition would look like if we did dispatch over a remote interface instead of in-memory (across package and function boundaries), then this just underpins this point even more. If we build everything with the idea that we can just pass a pointer to something in-memory, then it's going to be a really really painful transition requiring a much much larger refactor.

> Which implementations today (and in the foreseeable future) are not in memory function calls?! https://github.com/openfga/openfga/pull/476#discussion_r1093892700 and here we are, they are all in memory still.

Yep, can't argue with that. If you don't invest in the architecture that the composition is specifically made for, then I guess you just wrote some code that may have some side-effects that have no value at this time. Not sure I'm sold on if that negates the value of thinking about the problem in a way that could be extendable though.. In any case though, you're right that we don't have remote dispatch at this time. I wonder how the tides may turn if we just took a little bit of time to explore how remote dispatch could help our platform, because maybe we'd be more inclined to spend the time there instead of a refactoring code. If remote dispatch has promising hypothesis' around it, it's probably worth exploring that before we go hacking at the current composition that is more well-tailored for it.

So here are my final parting thoughts  - if we really have enough distaste for the current code composition and don't feel it's valuable enough to more easily support remote dispatching adoption down the road, and us as a maintainer team prefer to refactor things to assume purely in-memory composition semantics, then let's start identifying those bits and get refactoring. It's going to be a pretty sizable refactor, but if that's the direction we feel is right at this time, then let's do it.

internal/graph/singleflight_resolver.go
jon-whit: In fact, I have found that what @willvedd and I were exploring in https://github.com/openfga/openfga/pull/1335#issuecomment-1942938328 was also a bit misleading. I have pretty high confidence now that we were wrong in our theory.

The root cause still boils down to context propagation across the inflight singleflight resolvers. All of our hypothesis' around tracking locked subproblems didn't work (tests failed) when I ran large iterations of our tests. They led to more stable tests, but I still saw intermittent test failures quite frequently.

The changes as they are now are the latest most changes that fix the problem in a stable way. I can run our test suites for ListObjects and Check with a large number of iterations and they still pass. The changes also make the most sense to me, becase prior to using the `resenje.org/singlelfight` package we were always going to have some problems with independent context overlapping on de-duplicated subproblems. I wish the standard `x/sync/singleflight` package has `DoWithContext` that operated like this alternative package does.

go.mod
```diff
@@ -49,6 +49,7 @@ require (
 	golang.org/x/sync v0.6.0
 	google.golang.org/grpc v1.61.0
 	google.golang.org/protobuf v1.32.0
+	resenje.org/singleflight v0.4.1
```
willvedd, line 52: Double checking here, I think you mean to use the x/sync version of SF.

jon-whit: Actually this was very purposeful. See https://github.com/openfga/openfga/pull/1335#discussion_r1488920233

willvedd:
**Update:** I'm opting to close this PR out. It has gotten quite large, the discussion quite noisy and we as a team have learned so much about implementing a check request deduplication optimization layer since first opened. We've also broken-down the problems in this PR into more discrete PRs: #1371 , #1372 , #1375 . The final PR for implementing the singleflight check resolver will be opened shortly.

PR #1343, opened 2024-02-07
Dispatch count to resolution metadata

willvedd:
## Description
What makes a check query expensive in both resources and time? Ostensibly, a high number of datastore queries is probably the first answer. However, that isn't the full picture and there are other aspects of a check query that could affect the cost. For example, the depth of the relationships, the number of tuples, the number of dispatches, etc. We have decent intuitions but ultimately we need data to understand the relationships between those facets and the bottom-line performance. 

This PR is a first-pass at trying to unearth some check resolution metrics through metadata so that they can be reported on and analyzed.

## Review Checklist


internal/graph/check.go
```diff
@@ -485,7 +499,12 @@ func (c *LocalChecker) Close() {
 // was constructed with.
 func (c *LocalChecker) dispatch(_ context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
-		return c.delegate.ResolveCheck(ctx, req)
+		resp, err := c.delegate.ResolveCheck(ctx, req)
+		if err != nil {
+			return resp, err
+		}
+		resp.GetResolutionMetadata().DispatchCount++
+		return resp, nil
```
willvedd, line 507: **Note:** I know this ruins the purity of the dispatch function but its a logical place to put this code.

jon-whit:
@willvedd would you mind adding this test as well. It just exercises direct userset dispatches and is a good test to have. It's a little easier to reason about the direct userset lookups and their dispatches.

https://gist.github.com/jon-whit/d59ea63d26f5e509ec61d3776185ea2b

internal/graph/check_test.go
```diff
@@ -996,6 +996,75 @@ condition condition1(param1: string) {
...
+	storeID := ulid.Make().String()
+
+	model := parser.MustTransformDSLToProto(`model
+  schema 1.1
+
+type user
+
+type folder
+  	relations
+		define viewer: [user] or viewer from parent
```
jon-whit, line 1012: Please add a test that exercises `computed_userset` as well. For example:

```
type doc
  relations
     define owner: [user]
     define editor: [user] or owner
```

- document:1#owner@user:jon
- document:2#editor@user:will

`Check(document:1#owner@user:jon)` should require 1 dispatch.
`Check(document:2#editor@user:will)` should require 0 dispatches.

Evaluating a rewritten relation through computed_userset incurs a dispatch.

internal/graph/check_test.go
```diff
@@ -996,6 +996,75 @@ condition condition1(param1: string) {
 	require.False(t, resp.Allowed)
 }
 
+func TestCheckDispatchCount(t *testing.T) {
```
jon-whit, line 999: Can we add a subtest that just ensures we don't start counting a direct lookup as a dispatch with a code change in the future?

e.g.

```
type user

type document
  relations
    define viewer: [user]
```

`Check(document:1#viewer@user:jon)` should never have a positive dispatch count. It should always be 0, but with a datastore query count of 1.

internal/graph/graph.go
```diff
@@ -50,6 +50,8 @@ type ResolutionMetadata struct {
 	// evaluated and potentially discarded
 	// If the solution is "allowed=false", no paths were found. This is the sum of all the reads in all the paths that had to be evaluated
 	DatastoreQueryCount uint32
+
+	DispatchCount uint32
```
jon-whit, line 54:
```suggestion
        // The number of Check subproblems that had to be dispatched to resolve the parent subproblem this response is associated with.
	DispatchCount uint32
```

willvedd:
**Note:** Only added dispatch count metrics for check for now. Listobjects will require a follow-up PR once this is approved.

tests/check/check_test.go
```diff
@@ -250,7 +250,8 @@ type document
 			require.NotEmpty(t, fields["trace_id"])
 			if !test.expectedError {
 				require.NotEmpty(t, fields["datastore_query_count"])
-				require.Len(t, fields, 13)
+				require.NotNil(t, fields["dispatch_count"])
```
willvedd, line 253: The dispatch count for this specific example is zero so asserting `require.NotEmpty` would result in an error.

internal/graph/check_test.go
```diff
@@ -996,6 +996,230 @@ condition condition1(param1: string) {
...
+	ds := memory.New()
+	ctx := storage.ContextWithRelationshipTupleReader(context.Background(), ds)
+
+	t.Run("dispatch_count_tuples_to_usersets", func(t *testing.T) {
```
jon-whit, line 1003:
```suggestion
	t.Run("dispatch_count_ttu", func(t *testing.T) {
```

internal/graph/check_test.go
```diff
@@ -996,6 +996,230 @@ condition condition1(param1: string) {
...
+	  relations
+	    define member: [user, group#member]
+
+	type document
+	  relations
+		define viewer: [group#member]
+	`)
+
+		err := ds.Write(ctx, storeID, nil, []*openfgav1.TupleKey{
+			{
```
jon-whit, line 1099: Can we reduce this and the others to the shorthand `tuple.NewTupleKey("group:1", "member", "user:jon")`? We strive to do that elsewhere to cut down on the visual boilerplate.

internal/graph/check_test.go
```diff
@@ -996,6 +996,230 @@ condition condition1(param1: string) {
...
+		})
+		require.NoError(t, err)
+		require.True(t, resp.Allowed)
+
+		require.Zero(t, resp.GetResolutionMetadata().DispatchCount)
+
+		resp, err = checker.ResolveCheck(ctx, &ResolveCheckRequest{
+			StoreID:              storeID,
+			AuthorizationModelID: model.GetId(),
+			TupleKey:             tuple.NewTupleKey("document:2", "editor", "user:will"),
```
jon-whit, line 1212: And `ResolveCheck(document:1#editor@user:jon)` should always require exactly 1 dispatch (because `user:jon` is a `document:1#owner`). It may be good to just add that one as well. The `user:will` example below covers the overlap with the race on `editor or owner`. But for `user:jon` it should always be deterministic with 1 dispatch.

tests/check/check_test.go
jon-whit: nit (non-blocking) - If you're consuming this field you probably don't want it to change from numeric to string accidentally as well, so it may be worth adding that type assertion on the value type as well 🤷 ? This in addition to the `NotNil` check already here.

internal/graph/check_test.go
willvedd: The shorthand leaves ambiguity between user and object. Even as I've been making regular contributions, I still have trouble groking it. 

Can we add a linting rule or enforce this style decision?

jon-whit: > Can we add a linting rule or enforce this style decision automatically?
That'd be great, but we'd have to build the linter for it. Up until now we've been operating on keeping each other in check (human lint rules 😄 ).

> The shorthand leaves ambiguity between user and object. Even as I've been making regular contributions, I still have trouble groking it.
That's fair, but if you pass the wrong arguments to a function that's just an unintended side-effect, which the unit under test should keep you in check with anyways. For example, if you have a function that does`fmt.Sprintf("%s/%s", var1, var2)` - if you meant `("%s/%s", var2, var2)` then presumably a test should have caught. Similarly, if you inverse `tuple.NewTupleKey(user, relation, object)` instead of `tuple.NewTupleKey(object, relation, user)` then hopefully the thing you're testing would have accounted for that anyways, and if not then the order probably didn't truly matter.

jon-whit (approved):

jon-whit (approved):

PR #1349, opened 2024-02-09
SingleflightCheckResolver testing – Improving reliability and strength

willvedd:
## Description
Tests for the `SingleflightCheckResolver` were fairly consistent but failed occasionally due to the non-deterministic nature of the feature; it is never guaranteed that identical subproblems are executed concurrently. 

This PR incorporates two main strategies for improving reliability. First, multiple iterations are run. By doing so, the intermittent inconsistencies are smoothed-out. Secondly, the tuples and model are intentionally written in a way to force full graph traversal. Specifically, ensuring that the resolution is false to exhaust all subproblems, yielding a more consistent result.

Proving that the tests are highly reliable:
```
go test ./internal/graph/... -race -run "^TestSingleflightResolver$" -count=5000

ok      github.com/openfga/openfga/internal/graph       19.692s
```

## References
 - Original PR: #1313 

## Review Checklist

### DIFF
--- a/internal/graph/singleflight_resolver_test.go
+++ b/internal/graph/singleflight_resolver_test.go
@@ -2,7 +2,6 @@ package graph
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	"github.com/oklog/ulid/v2"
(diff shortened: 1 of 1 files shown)


internal/graph/singleflight_resolver_test.go
```diff
@@ -18,156 +17,146 @@ import (
...
+			})
+			require.NoError(t, err)
+			require.False(t, resp.GetAllowed())
+			dbReadsWithout += resp.GetResolutionMetadata().DatastoreQueryCount
+		}
+
+		require.Less(t, dbReadsWith, dbReadsWithout) //singleflight resolver will always result in fewer DB reads than without
+
+		require.LessOrEqual(t, dbReadsWith, uint32(testIterations+2)) // A buffer of two DB reads to ensure test reliability
+		require.GreaterOrEqual(t, dbReadsWith, uint32(testIterations))
```
miparnisari, line 120: You are comparing the outcome of the sum of 2 experiments that ran N times each. I think this is mathematically equivalent to looking at the outcome of each individual experiment, but I feel like the latter is easier to read. I suggest moving the assertions to within each loop. 

```go
// for each iteration of singleflight
require.GreaterOrEqual(t, res.ResolutionMetadata.DatastoreQueryCount, 1)
require.LessOrEqual(t, res.ResolutionMetadata.DatastoreQueryCount, 2)
```

`testIterations := 3` i also recommend making this larger to increase the chance that you'll see both results. say 1000

willvedd: The problem here is that 99% is roughly equivalent to 100%. That's the whole point here, the tests aren't deterministic. 

Taking the sum of several iterations is more likely to result in a higher pass-rate because the aggregation will regress to the mean. Your suggestion of running 1000 iterations and testing the result of each assertion is the opposite, it will fail if 1/1000 trials fails, which is an almost certainty.

miparnisari: Sorry i'm confused,

> Your suggestion of running 1000 iterations and testing the result of each assertion is the opposite, it will fail if 1/1000 trials fails,

but why would it fail? If the run for singleflight isn't 1 or 2, then that is a legit bug and we should know about it

willvedd: > If the run for singleflight isn't 1 or 2

Evaluating the DB query count at face value doesn't prove that it works. If a single test requires two DB reads, so what? That's the same as not using it. Granted, we could add that assertion to compliment the existing assertions, but it would only prove that there aren't more than two DB queries for a single check.

The test is much stronger and reliable when comparing singleflight resolver usage vs not. And the only reliable way of testing this is over multiple iterations because the results are probabilistic.

internal/graph/singleflight_resolver_test.go
```diff
@@ -18,156 +17,146 @@ import (
...
-
-type folder  
-    relations  
-    define parent1: [folder]  
-    define parent2: [folder]  
-    define parent3: [folder]  
-    define viewer: [user] or viewer from parent1 or viewer from parent2 or viewer from parent3`)
+	type user
+	type doc
+		relations
```
jon-whit, line 27: Indentation is pretty extreme here. Can you change the spacing please?

miparnisari (dismissed):

miparnisari (dismissed):

miparnisari (approved):

PR #1359, opened 2024-02-12
fix: forbid conditions that share the same parameter name

### LINKED ISSUE #1286: If two conditions share the same parameter name but with different types, you get evaluation errors when calling Check or ListObjects
### Checklist

- [X] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [X] I have searched the [Discord](https://discord.gg/8naAwJfWN6) community and have not found a suitable solution or answer.

### Description

We should forbid two or more conditions in a model from sharing the same parameter name.

### Expectation

We should forbid conditions that share the same parameter names.

### Reproduction

If two conditions share the same parameter name and type, issuing a Check or ListObjects call with `context {x = 1}` (for example) will be ambiguous, since `x` can have different meaning for each condition. E.g.

```yaml
model_file: |
  model
    schema 1.1
  type user
  type document
    relations
      define viewer: [user with cond1, user with cond2]

condition cond1(x: string) {
  x == "19/2/2024" // here "x" means a date
}
condition cond2(x: string) {
   x == "maria" // here "x" means a name
}
```

If two conditions share the same parameter name but with different types, you get type cast errors when calling Check or ListObjects

```yaml
model_file: |
  model
    schema 1.1
  type user
  type document
    relations
      define viewer: [user with cond1, user with cond2]

condition cond1(x: int) {
  x > 100
}
condition cond2(x: string) {
   x == "hello"
}
```

For example if you call Check with `context: "hello", you can get
```
{
   "code": "validation_error",
   "message": "1 error occurred:\n\t* failed to evaluate relationship condition: parameter type error on condition 'cond1' - failed to convert context parameter 'x': expected type value 'int64', but found 'string'"
}
```

or if you call Check with `context: 1`, you can get

```
{
   "code": "validation_error",
   "message": "1 error occurred:\n\t* failed to evaluate relationship condition: parameter type error on condition 'cond' - failed to convert context parameter 'x': expected type value 'string', but found 'int64'"
}
```

### Store data

_No response_

### OpenFGA version

1.4.0

As a binary

In-Memory

### OpenFGA Flags

none

### Logs

_No response_

edwin-Marrima:
Closes #1286 

## Description
Forbid conditions that share the same parameter names

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -264,6 +264,7 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 	}()
 
 	var dbReads uint32
+	var dispatchCount uint32
 	var err error
 	for i := 0; i < len(handlers); i++ {
 		select {
@@ -276,9 +277,11 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 				continue
 			}
 			dbReads += result.resp.GetResolutionMetadata().DatastoreQueryCount
+			dispatchCount += result.resp.GetResolutionMetadata().DispatchCount
 
 			if result.resp.GetAllowed() {
 				result.resp.GetResolutionMetadata().DatastoreQueryCount = dbReads
+				result.resp.GetResolutionMetadata().DispatchCount = dispatchCount
 				return result.resp, nil
 			}
 		case <-ctx.Done():
@@ -290,6 +293,7 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 		Allowed: false,
 		ResolutionMetadata: &ResolutionMetadata{
 			DatastoreQueryCount: dbReads,
+			DispatchCount:       dispatchCount,
 		},
 	}, err
 }
@@ -318,6 +322,7 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 	}()
 
 	var dbReads uint32
+	var dispatchCount uint32
 	var err error
 	for i := 0; i < len(handlers); i++ {
 		select {
@@ -339,8 +344,11 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 			}
 
 			dbReads += result.resp.GetResolutionMetadata().DatastoreQueryCount
+			dispatchCount += result.resp.GetResolutionMetadata().DispatchCount
+
 			if !result.resp.GetAllowed() {
 				result.resp.GetResolutionMetadata().DatastoreQueryCount = dbReads
+				result.resp.GetResolutionMetadata().DispatchCount = dispatchCount
 				return result.resp, nil
 			}
 		case <-ctx.Done():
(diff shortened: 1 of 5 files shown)


pkg/typesystem/typesystem.go
```diff
@@ -1123,14 +1123,26 @@ func (t *TypeSystem) validateTypeRestrictions(objectType string, relationName st
...
+
 		if key != c.Name {
 			return fmt.Errorf("condition key '%s' does not match condition name '%s'", key, c.Name)
 		}
 
 		if err := c.Compile(); err != nil {
 			return err
 		}
+
+		//validate duplicated conditions parameter
```
jon-whit, line 1139: nit - mind adding a space between `// validate`?

pkg/typesystem/error.go
```diff
@@ -47,6 +47,9 @@ var (
 
 	// ErrNoConditionForRelation is returned when no condition is defined for a relation in the authorization model.
 	ErrNoConditionForRelation = errors.New("no condition defined for relation")
+
+	//ErrDuplicatedConditionParameterName is return when two or more conditions share the same parameter name.
```
jon-whit, line 51: nit - add space between `// Err` please

jon-whit (changes requested):
This is a breaking change - we need more discussion and/or coordination before we proceed forward with this.

cc'ing @openfga/backend  more broadly.

miparnisari:
I don't think it's a breaking change, i would think of it as a bug fix 🤷‍♀️ . If you have a model with clashing parameters today, you are getting errors when calling Check or ListObjects anyway.

edwin-Marrima:
[miparnisari](https://github.com/miparnisari) With this change in place, If two conditions share the same parameter name with same types it will fail, due this change doesn't check the type (just the parameter name). 

Maybe we should consider checking whether parameters with same name has different types

jon-whit:
> I don't think it's a breaking change, i would think of it as a bug fix 🤷‍♀️ . If you have a model with clashing parameters today, you are getting errors when calling Check or ListObjects anyway (or you will if you add tuples)

Not necessarily though. Here's an example:

```
model
  schema 1.1

type user

type document
  relations
    define viewer: [user with cond1, user with cond2]

condition cond1(x: int) {
    x < 100
}

condition cond2(x: string) {
    x == "hello"
}
```

If you have the following tuples:
- document:1#viewer@user:jon, condition: cond1
- document:1#viewer@user:maria, condition: cond2

The following Checks are totally acceptable:
`Check(document:1#viewer@user:jon, context: {x: 1})` --> {allowed: true}
`Check(document:1#viewer@user:maria, context: {x: "hello"})` --> {allowed: true}

If we did what this PR says it is doing then we'd be breaking the use case above, and I'm not sure we want to do that.

miparnisari:
Ok yes in that case it's a breaking change. I'm still okay with this, as long as we call it out on the release notes.

jon-whit:
> Ok yes in that case it's a breaking change. I'm still okay with this, as long as we call it out on the release notes.

I left a more detailed description of the concern [here](https://github.com/openfga/openfga/issues/1286#issuecomment-1946395136). We just need to make a decision on which approach we want to go. cc'ing @aaguiarz for context.

PR #1368, opened 2024-02-15
chore: move gateway to pkg/

miparnisari:
## Description

One can do `new OpenFGAServer(WithTransport(....))`.  The parameter, the transport, needs to be a publicly exported interface.

In this PR I am making it public so consumers can inject their own implementation (or a mock).


sergiught (changes requested):
Thanks for doing this @miparnisari 🙏🏻 

Could we ensure we have comments on publicly exported types within the gateway pkg as well? 🙏🏻

pkg/gateway/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package gateway provides implementations to send data through the transport (e.g. headers).
+package gateway
```
miparnisari, line 2: i don't know why we called this gateway (probably because of https://github.com/grpc-ecosystem/grpc-gateway?). But this word means nothing to me personally and i would like to rename this to `transport`.

pkg/server/server.go
```diff
@@ -1007,7 +1006,7 @@ func (s *Server) resolveTypesystem(ctx context.Context, storeID, modelID string)
 
 	span.SetAttributes(attribute.KeyValue{Key: authorizationModelIDKey, Value: attribute.StringValue(resolvedModelID)})
 	grpc_ctxtags.Extract(ctx).Set(authorizationModelIDKey, resolvedModelID)
-	_ = grpc.SetHeader(ctx, metadata.Pairs(AuthorizationModelIDHeader, resolvedModelID))
+	s.transport.SetHeader(ctx, AuthorizationModelIDHeader, resolvedModelID)
```
miparnisari, line 1009: We already have a test for this change 👍 https://github.com/openfga/openfga/blob/a6e000905c3947355f138fe459ae9c48536823f4/cmd/run/run_test.go#L1149

sergiught (approved):

PR #1371, opened 2024-02-15
Ignore cycle errors in union reducer

willvedd:
## Description
The `union()` function is a reducer that accepts any number of handlers and makes a check decision based on the combination of all those reducers. It is how the union operator is implemented in models with the `or` syntax.

Generally, if one or more handler within the reducer is truthy, the whole evaluation returns true. If all handlers are falsey, then the union reducer will return false. Errors generally override any truthy or falsey resolution. If one handler errors for any reason, even if other handlers are truthy, the reducer will return that error. 

However, the `ErrCycleDetected` is not an error that should be considered a critical error. It should be considered the same as a falsey result because it signals that a resolution branch was exhausted without any truthy result. **Ignoring this error within the union reducer will result in a more correct and fault-tolerant check resolution.** 

**Note: ** Similar changes may need to be made in the intersection and exclusion reducers. This PR is meant to serve as a starting point for those conversations.

## Review Checklist


internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
 	require.NoError(t, err)
 	require.False(t, resp.Allowed)
 }
+
+func TestUnionReducer(t *testing.T) {
```
jon-whit, line 586:
```suggestion
func TestUnionCheckFuncReducer(t *testing.T) {
```

internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
...
+		require.True(t, resp.GetAllowed())
+	})
+
+	t.Run("if a handler errors but other handler is falsy, return error and allowed:false", func(t *testing.T) {
+		depthExceededHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			return nil, ErrResolutionDepthExceeded
+		}
+
+		resp, err := union(ctx, concurrencyLimit, depthExceededHandler, falseHandler)
+		require.Error(t, err)
```
jon-whit, line 645:
```suggestion
		require.ErrorIs(t, ErrResolutionDepthExceeded, err))
```

internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
...
+		cyclicErrorHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			return nil, ErrCycleDetected
+		}
+
+		resp, err := union(ctx, concurrencyLimit, cyclicErrorHandler, falseHandler)
+		require.NoError(t, err)
+		require.False(t, resp.GetAllowed())
+	})
+
+	t.Run("should only aggregate DB queries of truthy handlers", func(t *testing.T) {
```
jon-whit, line 659: Misleading test name. Should be something like

> aggregate DatastoreQueryCount of any non-error handlers

internal/graph/check.go
```diff
@@ -269,7 +269,9 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 		select {
 		case result := <-resultChan:
 			if result.err != nil {
-				err = result.err
+				if result.err != ErrCycleDetected {
```
jon-whit, line 272:
```suggestion
				if error.Is(result.err, ErrCycleDetected) {
```

jon-whit: Just for consistency and to ensure any potential error wrapping upstream (not that it exists today).

internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
...
+		depthExceededHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			return nil, ErrResolutionDepthExceeded
+		}
+
+		resp, err := union(ctx, concurrencyLimit, depthExceededHandler, falseHandler)
+		require.ErrorIs(t, ErrResolutionDepthExceeded, err))
+		require.False(t, resp.GetAllowed())
+	})
+
+	t.Run("if a handler errors with cycle but other handler is falsy, return allowed:false with a nil error", func(t *testing.T) {
```
jon-whit, line 649: Please add a corresponding test for the case where one handler returns an ErrCycleDetected while the other returns `allowed: true`. The outcome should be `allowed: true`.

internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
...
+			}, nil
+		}
+
+		resp, err := union(ctx, concurrencyLimit, handler, handler, handler) // three handlers
+		require.NoError(t, err)
+		require.False(t, resp.GetAllowed())
+		require.Equal(t, uint32(3*3), resp.GetResolutionMetadata().DatastoreQueryCount)
+	})
+
+	t.Run("should return allowed:true if truthy handler evaluated before handler cancels via context", func(t *testing.T) {
```
jon-whit, line 700: This test is a bit misleading because the handlers can get invoked in different order and they are run concurrently relative to one another. I think it may make more sense to do something like:

```
ctx, cancel := context.WithTimeout(50*time.Millisecond)

resp, err := union(ctx, concurrencyLimit, trueHandler)
require.NoError(t, err)
require.True(t, resp.GetAllowed())
```

This way the context cancellation isn't somehow bound/hidden by the concurrent resolver in the reducer.

internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
...
+			time.Sleep(50 * time.Millisecond)
+			cancel()
+			return nil, nil
+		}
+
+		resp, err := union(ctx, concurrencyLimit, cancelHandler, trueHandler)
+		require.NoError(t, err)
+		require.True(t, resp.GetAllowed())
+	})
+	t.Run("should handle cancellations through context", func(t *testing.T) {
```
jon-whit, line 713: Can simply be
```go
ctx, cancel := context.WithTimeout(10*time.Millisecond)

slowHandler := func(context.Context) (*ResolveCheckResponse, error) {
	time.Sleep(50*time.Millisecond)
	return nil, nil
}

resp, err := union(ctx, concurrencyLimit, slowHandler)
require.ErrorIs(t, err, context.DeadlineExceeded)
```

internal/graph/check_test.go
```diff
@@ -581,3 +582,150 @@ condition condition1(param1: string) {
...
+		stuckHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			time.Sleep(1 * time.Hour)
+			return nil, nil
+		}
+
+		resp, err := union(ctx, concurrencyLimit, cancelHandler, stuckHandler)
+		require.ErrorIs(t, err, context.Canceled)
+		require.False(t, resp.GetAllowed())
+	})
+}
```
jon-whit, line 731: Can we add a test case where we assert what the behavior should be if we see a context deadline and context cancellation prior to seeing an `allowed: true` outcome. In that case we should get an error straight up.

internal/graph/check_test.go
```diff
@@ -478,15 +479,11 @@ type resource
...
 
-			// if the branch producing the cycle is reached first, then an error is returned, otherwise
-			// a result is returned if some other terminal path of evaluation was reached before the cycle
-			if err != nil {
-				require.ErrorIs(t, err, ErrCycleDetected)
-			} else {
-				require.False(t, resp.GetAllowed())
-				require.GreaterOrEqual(t, resp.ResolutionMetadata.DatastoreQueryCount, uint32(1)) // min of 1 (x) if x isn't found and it returns quickly
-				require.LessOrEqual(t, resp.ResolutionMetadata.DatastoreQueryCount, uint32(3))    // max of 3 (x, y, z) before the cycle
-			}
```
willvedd, line 489: Removal of non-deterministic test cases demonstrates the power of this small change.

internal/graph/check_test.go
```diff
@@ -581,3 +578,175 @@ condition condition1(param1: string) {
...
+		resp, err := union(ctx, concurrencyLimit, falseHandler, falseHandler, falseHandler)
+		require.NoError(t, err)
+		require.False(t, resp.GetAllowed())
+	})
+
+	t.Run("if_all_handlers_are_falsy,_return_allowed:false", func(t *testing.T) {
+		resp, err := union(ctx, concurrencyLimit, falseHandler, falseHandler, falseHandler)
+		require.NoError(t, err)
+		require.False(t, resp.GetAllowed())
+	})
```
miparnisari, line 623: duplicate test?

internal/graph/check_test.go
```diff
@@ -581,3 +578,175 @@ condition condition1(param1: string) {
...
+		t.Cleanup(cancel)
+
+		trueHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			time.Sleep(25 * time.Millisecond)
+			return &ResolveCheckResponse{
+				Allowed: true,
+			}, nil
+		}
+
+		resp, err := union(ctx, concurrencyLimit, trueHandler, falseHandler)
```
miparnisari, line 742: do you need to pass the falseHandler? i expect that because of the deadline exceeded we are already setting false

internal/graph/check_test.go
```diff
@@ -581,3 +578,175 @@ condition condition1(param1: string) {
...
+	t.Run("should_handle_cancellations_through_context", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+		t.Cleanup(cancel)
+
+		slowHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			time.Sleep(50 * time.Millisecond)
+			return nil, nil
+		}
+
+		resp, err := union(ctx, concurrencyLimit, slowHandler, falseHandler)
```
miparnisari, line 726: do you need to pass the falseHandler? i expect that because of the deadline exceeded we are already setting false

internal/graph/check_test.go
willvedd: Yes, good catch!

internal/graph/check_test.go
```diff
@@ -581,3 +578,175 @@ condition condition1(param1: string) {
...
+		cyclicErrorHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			return nil, ErrCycleDetected
+		}
+
+		resp, err := union(ctx, concurrencyLimit, cyclicErrorHandler, falseHandler)
+		require.NoError(t, err)
+		require.False(t, resp.GetAllowed())
+	})
+
+	t.Run("if_a handler_errors_with_cycle_but_other_handler_is_truthy,_return_allowed:true_with_a_nil_error", func(t *testing.T) {
```
miparnisari, line 649:
```suggestion
	t.Run("if_a_handler_errors_with_cycle_but_other_handler_is_truthy,_return_allowed:true_with_a_nil_error", func(t *testing.T) {
```

You can use this regex `t.run.*?\s.*?"` to find test names with missing underscores

internal/graph/check_test.go
```diff
@@ -581,3 +578,175 @@ condition condition1(param1: string) {
...
+		require.NoError(t, err)
+		require.True(t, resp.GetAllowed())
+	})
+	t.Run("should_handle_cancellations_through_context", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+		t.Cleanup(cancel)
+
+		slowHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			time.Sleep(50 * time.Millisecond)
+			return nil, nil
```
miparnisari, line 723: is this `nil, nil `intended or should it be 

```
return &ResolveCheckResponse{
				Allowed: false,
			}, nil
```

for this test to be the opposite of the one that follows it, where the handler returns true?

internal/graph/check_test.go
```diff
@@ -581,3 +578,169 @@ condition condition1(param1: string) {
...
+		depthExceededHandler := func(context.Context) (*ResolveCheckResponse, error) {
+			return nil, ErrResolutionDepthExceeded
+		}
+
+		resp, err := union(ctx, concurrencyLimit, depthExceededHandler, falseHandler)
+		require.ErrorIs(t, ErrResolutionDepthExceeded, err)
+		require.False(t, resp.GetAllowed())
+	})
+
+	t.Run("if_a_handler_errors_with_cycle_but_other_handler_is_falsy,_return_allowed:false_with_a_nil_error", func(t *testing.T) {
```
miparnisari, line 639: this test and the one below it are the only one that have handlers throwing `ErrCycleDetected`. but they both have _two_ handlers.

can you add a test for where there is only _one_ handler, and it throws `ErrCycleDetected`? in that case, following what @rhamzeh said here: https://github.com/openfga/openfga/pull/1372#pullrequestreview-1885724198

we should return the error, no?

miparnisari: never mind - if one handler returns ErrCycleDetected the final outcome is falsy

same if you have N handlers, if they all return ErrCycleDetected,  the final outcome is falsy

internal/graph/check_test.go
willvedd: It doesn't matter what `slowHandler` returns because the context will time-out long before the handler resolves.

internal/graph/check_test.go
willvedd: Not necessarily but just serves to demonstrate that you could pass another deterministic handler and it wouldn't affect the outcome; the union waits for the `slowHandler`.

internal/graph/check_test.go
willvedd: No, not necessarily but just serves to demonstrate that you can pass additional handlers. Can remove too.

internal/graph/check.go
```diff
@@ -269,6 +269,9 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 		select {
 		case result := <-resultChan:
 			if result.err != nil {
+				if result.err == ErrCycleDetected {
```
jon-whit, line 272:
```suggestion
				if errors.Is(result.err, ErrCycleDetected) {
```

jon-whit (dismissed):

internal/graph/singleflight_resolver_test.go
jon-whit, line 1: This file should be removed. It's not there in `main` today.

jon-whit (approved):

PR #1372, opened 2024-02-16
treat cycles as falsy outcome in intersection and exclusion CheckFuncReducers

jon-whit:
## Description
The changes in this PR adjust the intersection and exclusion CheckFuncReducers to treat a loopback cycle effectively as `{allowed: false}`. The idea being that if you could have resolved something in the resolution path then you would have, otherwise you can return `{allowed: false}` explicitly. `ErrCycleDetected` is not an error that should be considered a critical error like transient errors and/or `ErrResolutionDepthExceeded`. It should be considered the same as a falsy result because it signals that a resolution branch was exhausted without any explicitly truthy or falsy outcome. Treating this error in this specific way within the intersection and exclusion reducers will result in more correct and fault-tolerant Check resolution.

I've also added some tests for the intersection and exclusion reducers so that this behavior is more apparent in the future as well. The tests I added also include behavior around context cancellation, so that will help us avoid accidentally introducing regression in cancellation of in-progress subproblem resolution.

## References
Related to https://github.com/openfga/openfga/pull/1371

## Review Checklist


assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5212,12 +5212,18 @@ tests:
...
+              relation: viewer
+              object: document:1
+            expectation: false
         listObjectsAssertions:
           - request:
               user: user:jon
               type: document
               relation: viewer
-            errorCode: 2002
+            expectation: []
```
jon-whit, line 5226: These test changes are the 1 thing I'm not too sure about. I'd like to discuss this with the @openfga/backend team more.

If you have the following model and tuples:

```
type user

type document
  relations
    define allowed: [user, document#viewer]
    define viewer: [user, document#editor] and allowed

- document:1#viewer@user:jon
- document:1#viewer@document:1#editor
- document:1#editor@document:1#viewer
```
Before this PR `Check(document:1#viewer@user:jon)` would return an ErrCycleDetected. With this PR we return `{allowed: false}, nil` instead. The idea being that if you could have resolved a branch you would have, but otherwise you looped back around and thus didn't find a resolution path.

jon-whit: Just to be abundantly clear as well, this is a breaking change of behavior. If we agree it makes sense to proceed forward with it that is ok, but I wanted to at least point it out.

internal/graph/check.go
```diff
@@ -311,7 +318,7 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
 		select {
 		case result := <-resultChan:
 			if result.err != nil {
-				err = result.err
+				err = errors.Join(err, result.err)
 				continue
 			}
```
willvedd, line 323: Instead of checking for `ErrCycleDetected` twice below, can we instead check once here? 

```
case result := <-resultChan:
   if result.err == ErrCycleDetected {
      return &ResolveCheckResponse{
		Allowed: false,
			DatastoreQueryCount: dbReads,
		},
	}, nil
   }
   if result.err != nil {
      err = errors.Join(err, result.err)
      continue
   }
```

Then much of the error handling below can be removed, right?

internal/graph/check.go
```diff
@@ -422,12 +437,23 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
...
+			Allowed: false,
+			ResolutionMetadata: &ResolutionMetadata{
+				DatastoreQueryCount: 0,
+			},
+		}, errors.Join(baseErr, subErr)
+	}
+
+	allowed := true
+	if errors.Is(baseErr, ErrCycleDetected) {
+		allowed = false
```
willvedd, line 452: IMO it's hard to reason about this. Can we not just handle the `ErrCycleDetected` in the channel handlers above?

assets/tests/consolidated_1_1_tests.yaml
willvedd: **Note** The above model is missing the `editor` relationship.

But otherwise, I softly disagree about this being a breaking change because the results are now completely deterministic and not relying on race conditions. So I see this more as a bug fix 😄 .

willvedd (commented):
Agree in principal to these changes.

assets/tests/consolidated_1_1_tests.yaml
elbuo8: I'm pro this. 

But we should start considering a way to produce events so folks can debug/improve their models accordingly. Unsure if logging would cut it.

rhamzeh (commented):
I would caution against ignoring errors in `Check`. We did the same mistake in `ListRelations` in the SDK and been wanting to revert since then.

You are assuming all checks are of the form:
```
can user:anne read document 1?
```

Which implies if we cannot say that Anne can read, it implies that Anne cannot.

But this is not the only possibility. For example, a check may be on a relation with negative connotations, for example:
```
is user:anne blocked from document 1?
```

Now imagine someone calling FGA from a different context (e.g. OPA), so FGA is functioning as an information point instead of a decision point. In OPA the policy might be:
```
Anne can access the document, if it is tagged "marketing" and "published" and not blocked in FGA
```

By us returning false, we are indicating that we know 100% that anne is not blocked, so OPA will assume that and the user will be granted access.

indeterimite answers (such as errors) should be ignored ONLY if another path 100% guarantees a determinate answer, for example (true or error is always true, and false AND error is always false), because no matter what the value of error is, we know the final outcome.

For the full cases where error can be ignored:
```py
true OR true = true
true OR false = true
true OR error = true # error can be ignored
false OR true = true
false OR false = false
false OR error = error # error CANNOT be ignored
error OR error = error # error CANNOT be ignored
true AND true = true
true AND false = false
true AND error = error # error CANNOT be ignored
false AND true = false
false AND false = false
false AND error = false # error can be ignored
error AND error = error # error cannot be ignored
true BUT NOT true = false
true BUT NOT false = true
true BUT NOT error = error # error CANNOT be ignored
false BUT NOT true = false
false BUT NOT false = false
false BUT NOT error = false # error can be ignored
error BUT NOT error = error # error CANNOT be ignored
```

willvedd:
@rhamzeh This is not ignoring _all_ errors but effectively considering `ErrCycleDetected` as `allowed=false`. In fact, I'd argue that it's not even an error at all during check evaluations because the graph does indeed exhaust all possible nodes along the cyclical branch. I'd even go one step further and say that cycles are determinate especially because their behavior is consistent and repeatable. This is in contrast to other errors that are intermittent and do not exhausted the nodes along a resolution branch.

assets/tests/consolidated_1_1_tests.yaml
miparnisari: > Unsure if logging would cut it.

`if err == cycle { span.setError(cycle) }`

please :)

internal/graph/check_test.go
```diff
@@ -18,6 +21,381 @@ import (
...
+
+	t.Run("base_handler_with_cycle_subtract_handler_is_falsy_return_allowed:false_with_a_nil_error", func(t *testing.T) {
+		resp, err := exclusion(ctx, concurrencyLimit, cyclicErrorHandler, falseHandler)
+		require.NoError(t, err)
+		require.False(t, resp.GetAllowed())
+		require.LessOrEqual(t, resp.GetResolutionMetadata().DatastoreQueryCount, uint32(2+1))
+	})
+
+	t.Run("base_handler_with_cycle_subtract_handler_is_truthy_return_allowed:false_with_a_nil_error", func(t *testing.T) {
+		resp, err := exclusion(ctx, concurrencyLimit, cyclicErrorHandler, falseHandler)
```
miparnisari, line 125:
```suggestion
		resp, err := exclusion(ctx, concurrencyLimit, cyclicErrorHandler, trueHandler)
```

internal/graph/check.go
jon-whit: Updated - does that look better?

willvedd (dismissed):
Great! Agree with the decisions here.

internal/graph/check.go
jon-whit: Changed to simplify.

assets/tests/consolidated_1_1_tests.yaml
jon-whit: @miparnisari now we also record an error if we see a branch with a cycle (or error in general).

Screenshot 2024-02-16 at 2 55 46 PM

willvedd (approved):

PR #1375, opened 2024-02-16
chore: remove SingleflightCheckResolver and add CycleDetectionCheckResolver

jon-whit:
## Description
This PR rips out the SingleflightCheckResolver for now as we are working through some fixes in #1335 and it introduces a building block to the fix for #1335 which introduces a layered `CycleDetectionCheckResolver` and wires this up in a compositional way at the top-level Server layer.

The plan is to follow up with this PR by finishing #1335, but we're introducing this PR to help split out the changes and make it easier to digest.

## References
#1372
#1371 
#1335

## Review Checklist


willvedd (approved):
Very much agree with the idea of the cycle detection being its own resolver. Great work, lgtm!

internal/graph/graph.go
```diff
@@ -385,3 +385,43 @@ func (g *RelationshipGraph) getRelationshipEdgesWithTargetRewrite(
...
+//			LocalChecker                   |
+//				CycleDetectionCheckResolver -|
+//
+// The returned CheckResolverCloser should be used to close all resolvers involved in the
+// composition after you are done with the CheckResolver.
+func NewLayeredCheckResolver(
+	localResolverOpts []LocalCheckerOption,
+	cacheEnabled bool,
+	cachedResolverOpts []CachedCheckResolverOpt,
+) (CheckResolver, CheckResolverCloser) {
```
miparnisari, line 403: This looks like a code smell. `CheckResolver` interface includes a `Close` method, so why do we need `CheckResolverCloser`??

internal/graph/graph.go
```diff
@@ -385,3 +385,43 @@ func (g *RelationshipGraph) getRelationshipEdgesWithTargetRewrite(
...
+//		CachedCheckResolver              |
+//			LocalChecker                   |
+//				CycleDetectionCheckResolver -|
+//
+// The returned CheckResolverCloser should be used to close all resolvers involved in the
+// composition after you are done with the CheckResolver.
+func NewLayeredCheckResolver(
+	localResolverOpts []LocalCheckerOption,
+	cacheEnabled bool,
+	cachedResolverOpts []CachedCheckResolverOpt,
```
miparnisari, line 402: nit: if `cacheEnabled` is false, having to pass an empty `cachedResolverOpts` is... weird? I recommend making `cacheEnabled` part of `CachedCheckResolverOpt`

internal/graph/check.go
```diff
@@ -145,25 +143,14 @@ func WithMaxConcurrentReads(limit uint32) LocalCheckerOption {
...
-		d.delegate = singleflightResolver
-	}
-}
-
 // NewLocalChecker constructs a LocalChecker that can be used to evaluate a Check
 // request locally.
-func NewLocalChecker(ds storage.RelationshipTupleReader, opts ...LocalCheckerOption) CheckResolver {
+//
+// The constructed LocalChecker is not wrapped with cycle detection. Developers
+// wanting a LocalChecker without other wrapped layers (e.g caching and others)
```
miparnisari, line 150: `// wanting a LocalChecker WIT other wrapped layers (e.g caching and others)`

pkg/server/server.go
```diff
@@ -297,23 +297,30 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+		cachedCheckResolver := graph.NewCachedCheckResolver(
+			graph.WithMaxCacheSize(int64(s.checkQueryCacheLimit)),
+			graph.WithLogger(s.logger),
 			graph.WithCacheTTL(s.checkQueryCacheTTL),
-		))
+		)
+		s.cachedCheckResolverCloser = cachedCheckResolver.Close
+
+		cachedCheckResolver.SetDelegate(localChecker)
+		cycleDetectionCheckResolver.SetDelegate(cachedCheckResolver)
```
miparnisari, line 323: Why are we not using `NewLayeredCheckResolver` here, which appears to be made just for this use case?

internal/graph/check.go
```diff
@@ -145,25 +143,14 @@ func WithMaxConcurrentReads(limit uint32) LocalCheckerOption {
...
+// are encouraged to use [[NewLocalCheckerWithCycleDetection]] instead.
+func NewLocalChecker(opts ...LocalCheckerOption) *LocalChecker {
```
miparnisari, line 152: This is confusing

```
wanting a LocalChecker without other wrapped layers (e.g caching and others)
// are encouraged to use [[NewLocalCheckerWithCycleDetection]] instead.
```

but NewLocalCheckerWithCycleDetection doesn't include support for caching. Did you mean to say `NewLayeredCheckResolver`?

pkg/server/server.go
```diff
@@ -619,14 +615,17 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 	}
 
 	ctx = typesystem.ContextWithTypesystem(ctx, typesys)
-
-	checkResolver := graph.NewLocalChecker(
-		storagewrappers.NewCombinedTupleReader(s.datastore, req.ContextualTuples.GetTupleKeys()),
-		s.checkOptions...,
+	ctx = storage.ContextWithRelationshipTupleReader(ctx,
+		storagewrappers.NewBoundedConcurrencyTupleReader(
+			storagewrappers.NewCombinedTupleReader(
```
miparnisari, line 620: Following what I said in https://github.com/openfga/openfga/pull/1335#discussion_r1480652748, if this is the order of instantation done in `main` today too, then it's incorrect. 

When you call `Read` on a `b=boundedConcurrencyTupleReader` you are effectively doing this:

```go
       b.waitForLimiter(ctx)

	defer func() {
		<-b.limiter
	}()

	return b.RelationshipTupleReader.ReadUserTuple(ctx, store, tupleKey)
```

which is:

```go
        b.waitForLimiter(ctx)

	defer func() {
		<-b.limiter
	}()

	iter1 := storage.NewStaticTupleIterator(filterTuples(c.contextualTuples, tk.Object, tk.Relation))

	iter2, err := c.RelationshipTupleReader.Read(ctx, storeID, tk)
	if err != nil {
		return nil, err
	}

	return storage.NewCombinedIterator(iter1, iter2), nil
```

so I think in practice we are rate-limiting reading of things that are in memory.

The composition of functions should be this

```
storagewrappers.NewCombinedTupleReader(
			storagewrappers.NewBoundedConcurrencyTupleReader(
```

internal/graph/cycle_check_resolver.go
```diff
@@ -0,0 +1,55 @@
...
+func (c *CycleDetectionCheckResolver) ResolveCheck(
+	ctx context.Context,
+	req *ResolveCheckRequest,
+) (*ResolveCheckResponse, error) {
+	key := tuple.TupleKeyToString(req.GetTupleKey())
+
+	if req.VisitedPaths == nil {
+		req.VisitedPaths = map[string]struct{}{}
+	}
+
```
miparnisari, line 35: we should do `maps.Clone(req.VisitedPaths)` here. It should not be the caller's responsibility to do this.

miparnisari (commented):
Sorry for the late review. I have no major concerns with the code, but I do have concerns around introducing regressions in the future..

internal/graph/graph.go
jon-whit: `cacheEnabled` controls not only if the CachedCheckResolver gets constructed but also if if re-wire the other delegates based on if it is enabled. If we refactor `WithCacheEnabled` as a `CachedCheckResolverOpt` we lose the ability to make that determination.

With that in mind, how would you suggest the composition of the code look like to behave the same but remove the stutter?

internal/graph/graph.go
jon-whit: The intent was that anything that implements the `Close` method of the CheckResolver implements the `CheckResolverCloser` signature. It's a way to share/communicate that shared behavior.

We can refactor to simply `func()` though. I just prefer more strongly typed signatures, and giving that signature a typed name makes it as such.

miparnisari: i don't follow. Why do we need the `CheckResolverCloser` interface in the first place?

PR #1388, opened 2024-02-23
Adding tracing to CycleDetectionCheckResolver

willvedd:
## Description
Adding tracing to the `CycleDetectionCheckResolver` because we forgot to add it during its recent introduction into the codebase. This tracing provides important insight into the graph traversal algorithm during check resolution. Similar tracing is already implemented on the `LocalChecker` and `CachedCheckResolver` (see: #1331).

## References
Original PR: #1375 
Similar PR: #1331 

## Review Checklist


internal/graph/cycle_check_resolver.go
```diff
@@ -27,13 +29,20 @@ func (c *CycleDetectionCheckResolver) ResolveCheck(
 	ctx context.Context,
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "ResolveCheck")
+	defer span.End()
```
miparnisari, line 33: I am starting to get worried about the number of spans that we are generating. It looks as though if the Check involves N subproblems we are generating, N * (number of resolvers) spans. If you have caching & singleflight turned on, that will be 4N. 

And this doesn't include database reads spans, which could be thousands.

internal/graph/cycle_check_resolver.go
```diff
@@ -27,13 +29,20 @@ func (c *CycleDetectionCheckResolver) ResolveCheck(
...
 	if req.VisitedPaths == nil {
 		req.VisitedPaths = map[string]struct{}{}
 	}
 
-	if _, ok := req.VisitedPaths[key]; ok {
+	_, cycleDetected := req.VisitedPaths[key]
+	span.SetAttributes(attribute.Bool("cycle_detected", cycleDetected))
+	if cycleDetected {
 		return nil, ErrCycleDetected
 	}
```
miparnisari, line 47:
```suggestion
	if cycleDetected {
		span.SetAttributes(attribute.Bool("cycle_detected", true))
		return nil, ErrCycleDetected
	}
```

internal/graph/cycle_check_resolver.go
willvedd: Yes we need a new span for this because otherwise it is inconsistent with the other resolvers. 

The concern is fair but I'm not sure why it is arising now; a single check has always had the potential to create a large number of spans. I'd prefer if we focused on just the changes in this PR. If you're able to quantify the cost of your concerns maybe we can address as a team soon.

miparnisari: > Yes we need a new span for this because otherwise it is inconsistent with the other resolvers. 

Then i'd argue that we need to update the other resolvers too so that they do not create new spans. The number of spans has been a concern forever, specifically since we executed the [1 million RPS test](https://auth0.com/blog/getting-unlimited-scalability-with-okta-fine-grained-authorization/).

jon-whit: If we want to holistically address the concern of too many spans per Check, then we should embark on a focused effort for that and come up with what we want our trace generation to look like under the umbrella of that effort.

If the objective of our traces is to visualize everything, then I think we're going to be hard pressed for reducing our span generation. However, if the purpose of our tracing is to gain insights into latent paths, then maybe should be more selective about where we capture trace information. In any case, this conversation seems better had outside of the scope of this PR, because it is already affecting us today.

jon-whit:
@willvedd should we add a test for this?

```
// cycle_check_resolver_test.go

func TestResolveCheck_SpanCreation(t *testing.T) {
    	exporter := tracetest.NewInMemoryExporter()
	processor := trace.NewSimpleSpanProcessor(exporter)
	provider := trace.NewTracerProvider(
		trace.WithSampler(trace.AlwaysSample()),
		trace.WithSpanProcessor(processor),
	)

	otel.SetTracerProvider(provider)
	
	var cycleDetectionCheckResolver CheckResolver

        cycleDetectionCheckResolver.ResolveCheck(...)

        spans := exporter.GetSpans()
        for _, span := range spans {
            // ensure the CycleDetectionCheckResolver#ResolveCheck span is in the slice
        }

        ...
}
```

willvedd:
Re: a test for the spans, I think its too heavy-handed for what's being added. There isn't precedent for it anyway.

The whole intent of this PR is to bring consistency and observability to all our resolvers. This is something we forgot to do when we introduced the `CycleDetectionCheckResolver`.

internal/graph/cycle_check_resolver.go
willvedd: Right. The concern is fair but if we want observability into our check algorithm, we're going to need spans. I'm not convinced that the cost of this is not trivial and if it becomes a performance issue, we can address holistically within another body of work.

jon-whit (approved):

PR #1405, opened 2024-02-29
fix: HasEntrypoints issues #1385 and #1260

miparnisari:
## Description

Fix `typesystem.HasEntrypoints`.

## References
- Close https://github.com/openfga/openfga/issues/1385
- Close https://github.com/openfga/openfga/issues/1260


pkg/typesystem/typesystem.go
```diff
@@ -667,167 +667,208 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
 
+		// At least one type must have an entrypoint.
 		for _, assignableType := range tuplesetRelation.GetTypeInfo().GetDirectlyRelatedUserTypes() {
 			assignableTypeName := assignableType.GetType()
 
 			if assignableRelation, ok := typedefs[assignableTypeName][computedRelationName]; ok {
-				if _, ok := v[assignableTypeName][computedRelationName]; ok {
-					continue
+				if r, ok := v[assignableTypeName][computedRelationName]; ok {
+					return r, nil
```
miparnisari, line 781: This fixes #1260

pkg/typesystem/typesystem.go
```diff
@@ -667,167 +667,208 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
 		computedRelation, ok := typedefs[typeName][computedRelationName]
 		if !ok {
-			return false, false, fmt.Errorf("undefined type definition for '%s#%s'", typeName, computedRelationName)
+			return nil, fmt.Errorf("undefined type definition for '%s#%s'", typeName, computedRelationName)
 		}
 
-		if _, ok := v[typeName][computedRelationName]; ok {
-			return false, true, nil
+		if r, ok := v[typeName][computedRelationName]; ok {
+			return r, nil
```
miparnisari, line 754: this fixes #1385

pkg/typesystem/typesystem.go
```diff
@@ -667,167 +667,214 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
+	// hasEntrypoints is true when a particular object+relation relates to at least one assignable object type.
+	// A relation such as `define document#viewer: [document#viewer]` implies that `hasEntrypoints=false`.
+	// A relation such as `define document#viewer: [user, document#viewer]` implies that `hasEntrypoints=true`.
+	hasEntrypoints bool
+
+	// hasCycle is true when a particular object+relation introduces a cycle in the model.
+	// A relation such as `define document#viewer: [document#viewer]` does NOT imply a cycle.
+	// hasEntrypoints=true implies that hasCycle=false.
+	// If hasEntrypoints=false, nothing can be said about hasCycle.
+	hasCycle bool
```
miparnisari, line 681: I still don't know the difference between this and https://github.com/openfga/openfga/blob/b4f33fd3d7ec140e5ed4b3f94745656f5f930e03/pkg/typesystem/typesystem.go#L1294

That function returns cycles that `GetRelationDetails` does not. 🤷‍♀️ 

Ideally we only keep one of the two...

pkg/typesystem/typesystem.go
```diff
@@ -667,167 +667,212 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
+	// hasEntrypoints is true when a particular object+relation relates to at least one assignable object type.
+	// A relation such as `define document#viewer: [document#viewer]` implies that `hasEntrypoints=false`.
+	// A relation such as `define document#viewer: [user, document#viewer]` implies that `hasEntrypoints=true`.
+	hasEntrypoints bool
+
+	// hasLoop is true when a particular object+relation introduces a loop in the model.
+	// hasEntrypoints=true implies that hasLoop=false.
+	// If hasEntrypoints=false, nothing can be said about hasLoop.
+	// Loops are introduced by computed usersets.
+	hasLoop bool
```
miparnisari, line 681: I still don't know the difference between this and https://github.com/openfga/openfga/blob/b4f33fd3d7ec140e5ed4b3f94745656f5f930e03/pkg/typesystem/typesystem.go#L1294

That function finds cycles that `GetRelationDetails` can't. 🤷‍♀️

pkg/typesystem/typesystem.go
```diff
@@ -667,167 +667,212 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
 	case *openfgav1.Userset_This:
+		// At least one type must have an entrypoint.
 		for _, assignableType := range relation.GetTypeInfo().GetDirectlyRelatedUserTypes() {
-			if assignableType.GetRelationOrWildcard() == nil || assignableType.GetWildcard() != nil {
-				return true, false, nil
+			assignableTypeName := assignableType.GetType()
+
+			if _, ok := typedefs[assignableTypeName]; !ok {
+				return nil, fmt.Errorf("undefined type '%s'", assignableTypeName)
+			} else {
```
jon-whit, line 718: `else { if ` is redundant. 

```go
if _, ok := typedefs[assignableTypeName]; !ok {
    return nil, fmt.Errorf("undefined type '%s', assignableTypeName)
}

if assignableType.GetRelationOrWildcard() == nil || assignableType.GetWildcard() != nil {
    v[typeName][relationName].hasEntrypoints = true
    break
}
```
☝️ easier on the reader.

pkg/typesystem/typesystem.go
```diff
@@ -667,167 +667,212 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
+				if assignableType.GetRelationOrWildcard() == nil || assignableType.GetWildcard() != nil {
+					v[typeName][relationName].hasEntrypoints = true
+					break
+				}
 			}
 
-			assignableTypeName := assignableType.GetType()
 			assignableRelationName := assignableType.GetRelation()
 
 			assignableRelation, ok := typedefs[assignableTypeName][assignableRelationName]
```
jon-whit, line 727: note (non-blocking): this kind of code adds a lot of noise to the logic herein and it really should be validated ahead of time. If we were to hit this code block the model should have been invalid in the first place, because this would imply that we had a type restriction that referenced an undefined type and/or relation.

I suggest we add a `validated` flag to the `TypeSystem` structure to indicate if a model has already been validated. The assumption is that it should have already been validated by the time we are seeing it in this code, but if it hasn't then before doing any code herein we can do

```go
func (t *TypeSystem) getRelationDetails(...) (*RelationDetails, error) {
    if !t.validated {
        t.Validate() // lazy model validation
    }

    // assume the model is now valid from here below
}
```

I suggest we use a similar lazy build/validate pattern like the wonderful @jpadilla introduced for the Conditional Relationship Tuple [EvaluableCondition](https://github.com/openfga/openfga/blob/61b87fb6cd234261308f82fda253c062d2264906/internal/condition/condition.go#L54)

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,204 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
 			assignableTypeName := assignableType.GetType()
 
 			if assignableRelation, ok := typedefs[assignableTypeName][computedRelationName]; ok {
-				if _, ok := v[assignableTypeName][computedRelationName]; ok {
+				if r, ok := v[assignableTypeName][computedRelationName]; ok {
+					// We entered a potential cycle, so keep on looking.
+					// Note that, unlike the "This" case above, one TTU can add a parallel edge in the graph.
+					// For example: `define viewer: can_view from start and can_view from end`.
+					// So we need to remember if any of the edges led to "hasEntrypoints=true"
+					hasEntrypoints = hasEntrypoints || r.hasEntrypoints
```
miparnisari, line 795: this fixes #1260. The model in that scenario is special because there is more than one edge between two nodes:

1. `state#can_view` to `transition#can_apply` via `can view from start`
2. `state#can_view` to `transition#can_apply` via `can view from end`

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,204 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
-			return false, false, fmt.Errorf("undefined type definition for '%s#%s'", typeName, computedRelationName)
+			return nil, fmt.Errorf("undefined type definition for '%s#%s'", typeName, computedRelationName)
 		}
 
-		if _, ok := v[typeName][computedRelationName]; ok {
-			return false, true, nil
+		if r, ok := v[typeName][computedRelationName]; ok {
+			// we entered a definitive cycle, so return
+			r.hasLoop = true
+			return r, nil
```
miparnisari, line 766: Line 765 fixes #1385. If we found a loop, we return the correct value for `hasEntrypoints`, and not just `false`.

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,205 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
-		v[typeName] = map[string]struct{}{
-			relationName: {},
-		}
+	if _, ok := v[typeName]; !ok {
+		// first time visiting this type
+		v[typeName] = map[string]*relationDetails{}
+	}
+	if _, ok := v[typeName][relationName]; !ok {
+		// first time visiting this relation
+		v[typeName][relationName] = &relationDetails{false, false}
```
miparnisari, line 710: The map was updated not just to track which usersets were visited, but also their details.

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,205 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
+	// hasEntrypoints=true implies that hasLoop=false.
+	// If hasEntrypoints=false, nothing can be said about hasLoop.
+	// Loops are introduced by computed usersets.
+	hasLoop bool
+}
+
+// getRelationDetails returns information about a particular type and relation.
 // This function assumes that all other model validations have run.
-func hasEntrypoints(
+func (t *TypeSystem) getRelationDetails(
```
jon-whit, line 686:
```suggestion
func (t *TypeSystem) getEntrypointDetails(
```

jon-whit: `getRelationDetails` is quite ambiguous as there are many things related to a relation including: type restrictions, entrypoints, rewrites, etc.. For this reason I think it's more appropriate to name this function something like "entrypoint details"

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,205 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
-			if hasEntrypoint {
-				return true, false, nil
+			if relDetails.hasEntrypoints {
+				v[typeName][relationName].hasEntrypoints = true
+				break
 			}
 		}
 
-		return false, false, nil
+		return v[typeName][relationName], nil
```
jon-whit, line 754: Can we refactor/rename `v` to `visitedCopy` or `visitedRelationsCopy`?

pkg/typesystem/typesystem.go
jon-whit: Could you elaborate why this fixes #1385? I don't see a loop whatsoever in that example, so it's not clear to me why this logic herein would pertain to that specific issue.

At first glance it would seem that the visited map, specifically for the `but not` case would be more applicable.

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,205 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
...
 		if !ok {
-			return false, false, fmt.Errorf("undefined type definition for '%s#%s'", typeName, tuplesetRelationName)
+			return nil, fmt.Errorf("undefined type definition for '%s#%s'", typeName, tuplesetRelationName)
 		}
 
+		hasEntrypoints := false
+
+		// At least one type must have an entrypoint.
 		for _, assignableType := range tuplesetRelation.GetTypeInfo().GetDirectlyRelatedUserTypes() {
 			assignableTypeName := assignableType.GetType()
```
jon-whit, line 788: Testing 1, 2, 3

jon-whit (commented):
@miparnisari we need revisit entrypoints in general.

Consider the following test case:

```
model
	schema 1.1

type user
type employee

type document
	relations
		define a: [user]
		define b: [employee]
		define c: a and b
`,
inputType:     "document",
inputRelation: "c",
```
Today this returns `{hasEntrypoints: true}` when in reality it shouldn't. `document#c` only has entrypoints if `a` and `b` can resolve to at least one of the same subject/user types.

pkg/typesystem/typesystem.go
```diff
@@ -667,163 +667,205 @@ func (t *TypeSystem) relationInvolvesExclusion(objectType, relation string, visi
 	return false, nil
 }
 
-// hasEntrypoints recursively walks the rewrite definition for the given relation to determine if there is at least
-// one path in the rewrite rule that could relate to at least one concrete object type. If there is no such path that
-// could lead to at least one relationship with some object type, then false is returned along with an error indicating
-// no entrypoints were found. If at least one relationship with a specific object type is found while walking the rewrite,
-// then true is returned along with a nil error.
+type relationDetails struct {
```
jon-whit, line 670:
```suggestion
type relationEntrypointDetails struct {
```

jon-whit:
For historical context - I am remembering why we have cycle detection logic in the entrypoint code and how it differs from the `HasCycle` method in the TypeSystem. The term "cycle" when used in the entrypoint code should probably be renamed, because it's a little misleading.

When searching for an entrypoint in an FGA model we're trying to find a terminal subject/user that is reachable from some source object type and relation. If at least one relationship can lead to a terminal subject then we consider that relation to have non-zero entrypoints. Consider the following model:

```
type user

type document
  relations
    define viewer: [document#viewer, user]
```

As our code is written today (this is up for debate long term), determining the entrypoints for `document#viewer` involves looking at the type type restrictions `document#viewer` and `user`. Let's say we process these in the order in which they are defined. This means that while determining entrypoints for `document#viewer` we will find the type restriction for `document#viewer`. We mark this as a "cycle", but we don't return an error until we've search other potential terminal paths. That is, we move on and find the `user` type restriction. Since this type restriction is a terminal subject type we found at least 1 entrypoint. Consequently, the model above is a valid model. If we change the model ever so slight the model becomes invalid. Namely,
```
type document
  relations
    define viewer: [document#viewer]
```
In this case we find the `document#viewer` type restriction while processing the `document#viewer` entrypoints, and we mark it as a potential cycle. Since there are no other possible entrypoints to terminal subjects, then we return the cycle error.


This whole approach seems a bit off to me, and I think we need to revisit it. An entrypoint should not just represent a relationship to a terminal subject, because the model above can totally be valid assuming a particular usage pattern. If a develop plans on writing `document:1#viewer@document:2#viewer` and also intends to call `Check(document:1#viewer@document:2#viewer)`, then that is a reasonable expectation. The model above has an entrypoint, but the entrypoint is to a set of subjects, not just a terminal subject. I don't see anything wrong with that, and I believe our model validation is too strict today.

pkg/typesystem/typesystem.go
miparnisari: Note to self: they detect (or try to) the same kinds of cycles (those defined by computed relations only)

miparnisari:
> I am remembering why we have cycle detection logic in the entrypoint code and how it differs from the HasCycle method in the TypeSystem. The term "cycle" when used in the entrypoint code should probably be renamed, because it's a little misleading.

`hasEntrypoint` has cycle detection code to avoid stack overflows. We could get rid of cycle detection in there if `HasCycle` ran first always. Today, it runs after.

PR #1407, opened 2024-02-29
docs: Check

miparnisari:
Adding a few docs about Check in general, motivated by me not understanding the difference between `Depth` and `DispatchCount` (which was introduced in https://github.com/openfga/openfga/pull/1343)


internal/graph/check.go
```diff
@@ -508,16 +508,10 @@ func (c *LocalChecker) dispatch(_ context.Context, req *ResolveCheckRequest) Che
 	}
 }
 
-// ResolveCheck resolves a node out of a tree of evaluations. If the depth of the tree has gotten too large,
-// evaluation is aborted and an error is returned. The depth is NOT increased on computed usersets.
```
miparnisari, line 512: > The depth is NOT increased on computed usersets.

it is, see how in this test we expect an error: https://github.com/openfga/openfga/blob/1843929c1e76af581754ffe65bd244316c477b82/assets/tests/consolidated_1_1_tests.yaml#L5067-L5110

internal/graph/graph.go
```diff
@@ -42,6 +42,9 @@ func ResolutionDepthFromContext(ctx context.Context) (uint32, bool) {
 }
 
 type ResolutionMetadata struct {
+	// Thinking of a Check as a tree of evaluations:
```
jon-whit, line 45:
```suggestion
	// Thinking of a Check as a tree of evaluations, depth is ...
```

internal/graph/graph.go
```diff
@@ -52,6 +55,8 @@ type ResolutionMetadata struct {
 	DatastoreQueryCount uint32
 
 	// The number of Check subproblems that had to be dispatched to resolve the parent subproblem this response is associated with.
+	// Thinking of a Check as a tree of evaluations:
```
jon-whit, line 58:
```suggestion
	// Thinking of a Check as a tree of evaluations, DispatchCount will be the sum of all of the subproblems the parent subproblem has to evaluate.
```

Let's keep using already coined terms. We define subproblems above. Node isn't too clear in this context.

internal/graph/interface.go
```diff
@@ -6,9 +6,24 @@ import "context"
 
 type CheckResolverCloser func()
 
-// CheckResolver represents an interface that can be implemented to provide recursive resolution
-// of a Check.
 type CheckResolver interface {
+	// ResolveCheck resolves a node out of a tree of evaluations. If the depth of the tree has gotten too large,
```
jon-whit, line 10:
```suggestion
	// ResolveCheck resolves a single subproblem out of a tree of evaluations. If the depth of the tree has gotten too large,
```

jon-whit:
Nice 👍 . Thank you, thank you!

internal/graph/graph.go
miparnisari: I feel like tree & node go well together, since they belong to graph theory

willvedd: IMO both terms have their place in the docs. I do think "node" specifically fits when talking about graphs in the abstract. Also might be good to connect nodes and subproblems when defining subproblems.

willvedd (approved):

PR #1424, opened 2024-03-06
chore(break): remove some duplicate code and unused code

miparnisari:
## Description

Changes in the `typesystem` file:

1. `RewriteContainsIntersection` and `RewriteContainsExclusion` are dead code because they are not referenced anywhere outside of tests. Also I'm pretty sure that they are incorrectly implemented. The correct versions are `RelationInvolvesIntersection` and `RelationInvolvesExclusion`, respectively. These latter versions follow indirect relationships correctly.

2. The code for `relationInvolvesIntersection` is exactly the same as `relationInvolvesExclusion`, with the only difference in the case where they return `true`. I simplified it to a new function `relationInvolves(.... target)`

⚠️ Change 1 is a breaking change, though I expect little to zero outside callers of those functions.


jon-whit:
> RewriteContainsIntersection and RewriteContainsExclusion are dead code because they are not referenced anywhere outside of tests. Also I'm pretty sure that they are incorrectly implemented. The correct versions are RelationInvolvesIntersection and RelationInvolvesExclusion, respectively. 

There's history here. They served different purposes, but yeah the old ones can be remove. Thanks 👍

jon-whit (approved):

PR #1427, opened 2024-03-07
feat: add dispatch count for list objects

poovamraj:
## Description

As a continuation of adding `DispatchCount` metrics to the `Check` endpoint, we have added it to `ListObjects` as well to identify the amount of reverse expands and checks that are done for each request.

ListObjects works by reverse expanding from a subject to a list of objects with particular relation. There will be recursion involved in cases union and inheritance uses cases. And we run the check on each object that is retrieved if there is intersection or exclusion present so as to identify the related list of objects. We are counting these dispatches to identify the complexity of queries.

## References

Continuation of Dispatch Count in Check - https://github.com/openfga/openfga/pull/1343

## Review Checklist


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -167,11 +167,16 @@ type ReverseExpandResult struct {
 
 type ResolutionMetadata struct {
 	QueryCount *uint32
+
+	// The number of times we are expanding from each node to find set of objects
+	// In case of intersection and exclusion, it also includes the dispatch counts done for Check calls to exclude Objects
+	DispatchCount *uint32
```
miparnisari, line 173: ReverseExpand code is unaware that Checks are happening, that happens outside, in ListObjects.
```suggestion
	// The number of times we are expanding from each node to find set of objects
	DispatchCount *uint32
```

miparnisari, line 173:
```suggestion
	// The number of times we are expanding from each node to find set of objects
	DispatchCount *uint32
```

ReverseExpand is not aware that any Checks are happening. That's ListObjects's responsibility.

We should introduce a new `ResolutionMetadata` specifically for ListObjects, so we can distinguish between the two:

```go
type ListObjectsResponse struct {
	Objects            []string
	ResolutionMetadata ListObjectsResolutionMetadata
}

type ListObjectsResolutionMetadata struct {
	QueryCount *uint32

	// The number of times we are expanding from each node to find set of objects, 
       // AND the dispatch counts done for Check calls to exclude Objects
	DispatchCount *uint32
}

func NewResolutionMetadata() *ListObjectsResolutionMetadata {
	return &ListObjectsResolutionMetadata{
		QueryCount:    new(uint32),
		DispatchCount: new(uint32),
	}
}
```

you will need to make some changes, for example:

1. `evaluate`'s last parameter needs to change its type to `ListObjectsResolutionMetadata`

https://github.com/openfga/openfga/blob/b351a720134d323a5d64b9e62cffca9dd95b3e94/pkg/server/commands/list_objects.go#L165-L171

2. here, at the end of the goroutine, https://github.com/openfga/openfga/blob/b351a720134d323a5d64b9e62cffca9dd95b3e94/pkg/server/commands/list_objects.go#L255-L269 you'll need to add `atomic.AddUint32(listObjectsResolutionMetadata.DispatchCount, *reverseexpandResolutionMetadata.DispatchCount)`

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -321,6 +326,7 @@ LoopOnEdges:
 					Relation: innerLoopEdge.TargetReference.GetRelation(),
 				},
 			}
+			atomic.AddUint32(resolutionMetadata.DispatchCount, 1)
```
miparnisari, line 329: Why not put this as the first line of `execute()`?

miparnisari (changes requested):

pkg/server/commands/reverseexpand/reverse_expand.go
poovamraj: @miparnisari I spoke about this with @jon-whit so basically we want to add only the consecutive count of dispatches and not the first one i.e root. The difference is we have all the dispatch count higher by 1 but for the correctness of definition we are adding it right before dispatch happens instead of counting it from first.

pkg/server/commands/reverseexpand/reverse_expand.go
poovamraj: Agreed, but is there a future use case where reverse expand will be reused for another operation? In that case it is better to leave it here to be reused. 

Or else we should move it to ListObjects

poovamraj: Also looks like Check does the same and has it inside graph.go - https://github.com/openfga/openfga/blob/main/internal/graph/graph.go#L44

willvedd: Maria can you explain why you want to separate? Is it just because of the code comment?

pkg/server/commands/reverseexpand/reverse_expand.go
willvedd: Am I understanding correctly that the initial (root) invocation of `execute` doesn't technically count as a dispatch? And because of this, we only want to count subsequent invocations?

If so, that makes technical sense but highlights the inconsistency of recursive behavior between all the operations. I really like the idea of a standard dispatch concept but we're not very close.

poovamraj: >Am I understanding correctly that the initial (root) invocation of execute doesn't technically count as a dispatch?

Yes, that was my conversation with Jon

For me having it at top of execute means, even if we add a new call to execute, or a new recursion line, the one who is implementing doesn't need to remember to increase the dispatch count there and having all values off by just 1 didn't feel that big of a miscount but maybe it is. We do miss out on correctness.

willvedd: And that's fine. No further action is needed for this PR IMO. But just speaking generally, this highlights the inconsistency in behavior between execute and dispatch.

pkg/server/commands/list_objects_test.go
```diff
@@ -22,3 +31,124 @@ func TestNewListObjectsQuery(t *testing.T) {
...
+		name                  string
+		model                 string
+		tuples                []*openfgav1.TupleKey
+		objectType            string
+		relation              string
+		user                  string
+		expectedDispatchCount uint32
+	}{
+		{
+			name: "Test direct relation",
```
willvedd, line 49:
```suggestion
			name: "test_direct_relation",
```

For consistency and easier test identification, we underscore test names. Though I admit it's not obvious and there _should_ be some way to programmatically enforce.

pkg/server/commands/list_objects_test.go
```diff
@@ -22,3 +31,124 @@ func TestNewListObjectsQuery(t *testing.T) {
...
+				tuple.NewTupleKey("folder:B", "viewer", "user:jon"),
+				tuple.NewTupleKey("folder:A", "viewer", "user:jon"),
+			},
+			objectType:            "folder",
+			relation:              "can_delete",
+			user:                  "user:jon",
+			expectedDispatchCount: 2,
+		},
+	}
+
```
willvedd, line 114: Maybe add a couple zero cases to ensure accuracy. 
```suggestion
{
			name: "no_tuples",
			model: `model
schema 1.1

type user

type folder
  	relations
		define editor: [user]
		define can_delete: [user] and editor 
`,
			tuples:                []*openfgav1.TupleKey{},
			objectType:            "folder",
			relation:              "can_delete",
			user:                  "user:jon",
			expectedDispatchCount: 0,
		},
```

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: > Also looks like Check does the same and has it inside graph.go - [main/internal/graph/graph.go#L44](https://github.com/openfga/openfga/blob/main/internal/graph/graph.go?rgh-link-date=2024-03-08T11%3A11%3A34Z#L44)

Yeah that is okay, the resolution metadata of Check only includes Check resolutions

> Maria can you explain why you want to separate? Is it just because of the code comment?

Single responsibility principle. I think the DispatchCount of ReverseExpand should only include the number of dispatches, not the Checks that happen outside of it

pkg/server/commands/list_objects_test.go
```diff
@@ -22,3 +31,141 @@ func TestNewListObjectsQuery(t *testing.T) {
 		require.Error(t, err)
 	})
 }
+
+func TestCheckDispatchCount(t *testing.T) {
```
jon-whit, line 35:
```suggestion
func TestListObjectsDispatchCount(t *testing.T) {
```

pkg/server/commands/reverseexpand/reverse_expand.go
jon-whit: As for the godoc comment, can we just simply say 

> The number of subproblems that had to be dispatched to resolve the parent subproblem.

We're defining `Dispatch` in terms of nodes and other vernacular. It's confusing at best.

poovamraj: > Single responsibility principle. I think the DispatchCount of ReverseExpand should only include the number of dispatches, not the Checks that happen outside of it

I agree with this. Guess I misunderstood to use the same ResolutionMetadata. It would be better to use 2. 1 for reverse_expand and another for ListObjects 👍 I'll make those changes

pkg/server/commands/list_objects_test.go
```diff
@@ -22,3 +31,141 @@ func TestNewListObjectsQuery(t *testing.T) {
...
+		tuples                []*openfgav1.TupleKey
+		objectType            string
+		relation              string
+		user                  string
+		expectedDispatchCount uint32
+	}{
+		{
+			name: "test_direct_relation",
+			model: `model
+schema 1.1
```
jon-whit, line 51: Here are below, can we please format these models so they are indented inline with the field that defines them?

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: i don't think being off by 1 is a big deal, but i won't block the pr for it

willvedd: It's not off by one if we're defining a list objects dispatch the same as a check dispatch.

miparnisari: i didn't understand 🤔 

list objects dispatch = (reverse expand dispatches) + (Check dispatches)

Versus

list objects dispatch = (reverse expand dispatches - 1) + (Check dispatches)

i personally don't think it matters that much

jon-whit: > If so, that makes technical sense but highlights the inconsistency of recursive behavior between all the operations. I really like the idea of a standard dispatch concept but we're not very close.

☝️ this +1. This is one of the reasons why I think it's important to establish a clear and consistent Dispatch pattern across the code base. Today we have:

* `ReverseExpand#execute`
* `Expand#Execute`
* `CheckResolver#ResolveCheck`
* etc..

The invocation patterns and recursion loopback behaviors differ between these as well, and they really don't need to. I proposed to @adriantam and @poovamraj as part of this lager body of work to introduce more uniform dispatch composition through better interface composition, namely:

```go
type Dispatcher interface {
  CheckDispatcher
  ReverseExpandDispatcher
  ListUsersDispatcher
  ExpandDispatcher
}

// replaces CheckResolver today
type CheckDispatcher interface {

  // replaces ResolveCheck today
  DispatchCheck(context.Context, *DispatchCheckRequest) (...)
}

// replaces ReverseExpandQuery today
type ReverseExpandDispatcher interface {

  // replaces ReverseExpandQuery#execute today
  DispatchReverseExpand(...)
}

etc..
```

The purpose of this is to force us to more consciously consider the dispatch semantics and pattern, because doing so makes everything more uniform in general.

pkg/server/commands/list_objects_test.go
poovamraj: I tried to, was getting some model parsing errors. Saw check had same way of implementing these so just kept it. Let me check for someother example

jon-whit:
@poovamraj please add some more tests, specifically these cases:

## Test Case 1
```
model
  schema 1.1

type user

type group
  relations
    define member: [user]
```
- `group:eng#member@user:jon`
- `group:fga#member@user:jon`

`ListObjects(group#member, user:jon)`

This returns 2 dispatches - **note that this is non-ideal and points out an inefficiency in our algorithm**, because `user` is directly related to `group#member` and so this should be a single database lookup without any dispatch required. The only time dispatch should be required is with expansion of usersets and/or nested group/hierarchy expansion. More on this coming soon 😄 . I'm working on alternative algorithm proposals to avoid this and be more efficient.

## Test Case 2
```
model
  schema 1.1

type user

type group
  relations
```

- `group:eng#member@group:fga#member`
- `group:fga#member@user:jon`

`ListObjects(group#member, user:jon)`

This should return 2 dispatches today - note: it should only require 1 but because of the inefficiency in our algorithm today this returns 2. 

## Test Case 3
```
model
  schema 1.1

type user

type document
  relations
    define editor: [user]
    define viewer: editor
```

- `document:1#editor@user:jon`

`ListObjects(document#viewer, user:jon)`

This should return 2 dispatches today - again, it should only require 1 dispatch but it won't until we improve our algorithm a bit.

pkg/server/commands/reverseexpand/reverse_expand.go
poovamraj: @miparnisari made the changes you suggested. Check it out when you have sometime

pkg/server/commands/list_objects_test.go
jon-whit: Historically the model parser was finicky about whitespace. So that's why you see old references to models with poor indentation. I thought we updated our model parser to the latest one, which isn't so picky about whitespace.. Maybe not though? Please confirm.

pkg/server/commands/list_objects.go
```diff
@@ -57,9 +57,23 @@ type ListObjectsQuery struct {
 	checkResolver graph.CheckResolver
 }
 
+type ResolutionMetadata struct {
+	QueryCount *uint32
```
miparnisari, line 61:
```suggestion
        // The total number of database reads from reverse_expand and Check (if any) to complete the ListObjects request
	QueryCount *uint32
```

pkg/server/commands/list_objects.go
```diff
@@ -57,9 +57,23 @@ type ListObjectsQuery struct {
...
+
+	// The total number of dispatch counts from reverse_expand and check resolution to complete the ListObjects request
+	DispatchCount *uint32
```
miparnisari, line 64:
```suggestion
	// The total number of dispatch counts from reverse_expand and check resolutions (if any) to complete the ListObjects request
	DispatchCount *uint32
```

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari, line 173:
```suggestion
	// The number of times we are expanding from each node to find set of objects
	DispatchCount *uint32
```

😅

pkg/server/commands/list_objects_test.go
poovamraj: Well looks like we need schema to be the first line with no spaces or empty lines above. I don't know whether that is necessary

poovamraj:
@jon-whit added the test cases except No.1 as it is already the first test case

pkg/server/commands/list_objects_test.go
```diff
@@ -22,3 +31,181 @@ func TestNewListObjectsQuery(t *testing.T) {
...
+					  define can_delete: [user] and editor 
+			`,
+			tuples:                []*openfgav1.TupleKey{},
+			objectType:            "folder",
+			relation:              "can_delete",
+			user:                  "user:jon",
+			expectedDispatchCount: 0,
+		},
+		{
+			name: "no_tuples",
```
jon-whit, line 132: Please update this name. This test has tuples. My suggestion - `direct_userset_dispatch` or something like that.

pkg/server/commands/list_objects_test.go
```diff
@@ -22,3 +31,181 @@ func TestNewListObjectsQuery(t *testing.T) {
...
+				tuple.NewTupleKey("group:eng#member", "member", "group:fga#member"),
+				tuple.NewTupleKey("group:fga", "member", "user:jon"),
+			},
+			objectType:            "group",
+			relation:              "member",
+			user:                  "user:jon",
+			expectedDispatchCount: 2,
+		},
+		{
+			name: "no_tuples",
```
jon-whit, line 152: More descriptive name please. `computed_userset_dispatch` or something similar

pkg/server/commands/list_objects.go
```diff
@@ -57,9 +57,24 @@ type ListObjectsQuery struct {
 	checkResolver graph.CheckResolver
 }
 
+type ResolutionMetadata struct {
+	// The total number of database reads from reverse_expand and Check (if any) to complete the ListObjects request
+	QueryCount *uint32
+
+	// The total number of dispatch counts from reverse_expand and check resolutions (if any) to complete the ListObjects request
```
jon-whit, line 64:
```suggestion
	// The total number of dispatches aggregated from reverse_expand and check resolutions (if any) to complete the ListObjects request
```

pkg/server/commands/list_objects.go
```diff
@@ -57,9 +57,24 @@ type ListObjectsQuery struct {
 	checkResolver graph.CheckResolver
 }
 
+type ResolutionMetadata struct {
```
jon-whit, line 60:
```suggestion
type ListObjectsResolutionMetadata struct {
```

pkg/server/commands/list_objects.go
```diff
@@ -57,9 +57,24 @@ type ListObjectsQuery struct {
...
+	// The total number of database reads from reverse_expand and Check (if any) to complete the ListObjects request
+	QueryCount *uint32
+
+	// The total number of dispatch counts from reverse_expand and check resolutions (if any) to complete the ListObjects request
+	DispatchCount *uint32
+}
+
+func NewResolutionMetadata() *ResolutionMetadata {
```
jon-whit, line 68:
```suggestion
func NewListObjectsResolutionMetadata() *ListObjectsResolutionMetadata {
```

pkg/server/commands/list_objects.go
jon-whit: Because `ListObjectsQuery` is not scoped inside of it's own pkg. So let's be clear what the metadata applies to.

pkg/server/commands/list_objects.go
```diff
@@ -167,7 +182,7 @@ func (q *ListObjectsQuery) evaluate(
 	req listObjectsRequest,
 	resultsChan chan<- ListObjectsResult,
 	maxResults uint32,
-	resolutionMetadata *reverseexpand.ResolutionMetadata,
+	resolutionMetadata *ResolutionMetadata,
```
jon-whit, line 185: (nit non-blocking) Opinion piece here - resolution metadata is trailer metadata included and passed along as part of the request. For uniformity, dispatch composition should be consistent between ListObjects and Check, so in the near term I'd like to see this trailer metadata included in the request and not a tag-along as a parameter in the function signature. It's inconsistent.

miparnisari: > in the near term I'd like to see this trailer metadata included in the request

the response you mean? since the metadata applies to the response, not the request

but i agree..

the problem is how are we going to implement it? with the streaming variant this becomes complicated 🤔

poovamraj: In the request 🤔 I don't understand why should it be part of the request.

willvedd (approved):
Nice job.

miparnisari (approved):

pkg/server/commands/list_objects.go
jon-whit: Yeah, you guys get my drift 😄 . Sorry, it's late afternoon on a Friday, haha.

jon-whit (approved):

PR #1440, opened 2024-03-11
feat: dispatch throttling

adriantam:
## Description
Adding throttling based on configurable number of dispatches.  System will throttle request > configurable level

## Review Checklist


pkg/server/server.go
```diff
@@ -269,6 +279,48 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
...
+}
+
+// WithRateLimitedCheckResolverLowPriorityShaper sets the number of tickers required to dispatch a low priority work
+func WithRateLimitedCheckResolverLowPriorityShaper(shaper uint32) OpenFGAServiceV1Option {
+	return func(s *Server) {
+		s.rateLimitedCheckResolverLowPriorityShaper = shaper
+	}
+}
+
+// WithRateLimitedCheckResolverMediumPriorityLevel sets the number of dispatches a request incur before being placed on medium priority shaper
```
miparnisari, line 310: Can you explain here what is the difference between the low and the medium priority shaper?

And why is there no "high" priority shaper?

pkg/server/server.go
```diff
@@ -350,6 +434,10 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 
 // Close releases the server resources.
 func (s *Server) Close() {
+	if s.rateLimitedCheckResolverCloser != nil {
+		s.rateLimitedCheckResolverCloser()
+	}
```
miparnisari, line 439:
```suggestion
	if s.rateLimitedCheckResolver != nil {
		s.rateLimitedCheckResolver.Close()
	}
```

FWIW i also don't like that we have 

```go
if s.cachedCheckResolverCloser != nil {
	s.cachedCheckResolverCloser()
}
```

I know it's not your code to change, but i don't really like this new pattern that has started to creep where we save a pointer to a function unnecessarily.

cmd/run/run.go
```diff
@@ -200,6 +200,18 @@ func NewRunCommand() *cobra.Command {
 	// Unfortunately UintSlice/IntSlice does not work well when used as environment variable, we need to stick with string slice and convert back to integer
 	flags.StringSlice("request-duration-datastore-query-count-buckets", defaultConfig.RequestDurationDatastoreQueryCountBuckets, "datastore query count buckets used in labelling request duration by query count histogram")
 
+	flags.Bool("dispatch-throttling-enabled", defaultConfig.DispatchThrottling.Enabled, "enable throttling when request's number of dispatches is high")
```
miparnisari, line 203: Similar to my other comment on `server.go`, this doesn't give enough information to people to decide whether they want this on or off

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,166 @@
...
+type RateLimitedCheckResolverConfig struct {
+	TimerTickerFrequency time.Duration
+	MediumPriorityLevel  uint32
+	MediumPriorityShaper uint32
+	LowPriorityLevel     uint32
+	LowPriorityShaper    uint32
+}
+
+// RateLimitedCheckResolver will prioritize high priority queue (requests with smaller number of dispatches) over
+// medium and low priority queue (requests with larger number of dispatches)
```
miparnisari, line 24: This is the doc I was looking for :) 

why do we need two queues to represent `(requests with larger number of dispatches)`?

internal/graph/check.go
```diff
@@ -38,6 +39,7 @@ type ResolveCheckRequest struct {
 	Context              *structpb.Struct
 	ResolutionMetadata   *ResolutionMetadata
 	VisitedPaths         map[string]struct{}
+	DispatchCounter      *atomic.Uint32
```
miparnisari, line 42: We already keep track of this number within the ResolutionMetadata https://github.com/openfga/openfga/blob/d7f77478e8024f3f299c66d7279f9a0ccd84ba28/internal/graph/graph.go#L57-L61

Can we remove line 42? If not, can you leave a godoc explaining how these counters differ?

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,166 @@
...
+	for {
+		select {
+		case <-r.done:
+			r.ticker.Stop()
+			close(r.done)
+			close(r.medPriorityQueue)
+			close(r.lowPriorityQueue)
+			return
+		case <-r.ticker.C:
+			count = r.handleTimeTick(count)
```
miparnisari, line 118: i don't understand what `count` is for

miparnisari (commented):
I'm giving some early feedback, but i have not finished reviewing the changes entirely

willvedd (commented):
Really great to see this come together. I like the usage of the ticker and think that it's a modest improvement over the more naive sleep approach initially discussed.

I think its worth reconsidering wether the multiple arbitrary priority-levels is worth the overhead and complexity. Both internally and externally. Considering that 0.2% of check requests gobble ~75% of dispatches, we're really just trying to limit the tail-end of offenders here. I would wager that a single, harsher threshold would have similar performance characteristics. We could always introduce more ways of "shaping" the throttling as the need arises or per customer request.

cmd/run/flags.go
```diff
@@ -182,5 +182,23 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
...
 		util.MustBindEnv("requestDurationDatastoreQueryCountBuckets", "OPENFGA_REQUEST_DURATION_DATASTORE_QUERY_COUNT_BUCKETS")
+
+		util.MustBindPFlag("dispatchThrottling.enabled", flags.Lookup("dispatch-throttling-enabled"))
+		util.MustBindEnv("dispatchThrottling.enabled", "OPENFGA_DISPATCH_THROTTLING_ENABLED")
+
+		util.MustBindPFlag("dispatchThrottling.timeTickerFrequency", flags.Lookup("dispatch-throttling-time-ticker-frequency"))
+		util.MustBindEnv("dispatchThrottling.timeTickerFrequency", "OPENFGA_DISPATCH_THROTTLING_TIME_TICKER_FREQUENCY")
+
+		util.MustBindPFlag("dispatchThrottling.lowPriorityLevel", flags.Lookup("dispatch-throttling-low-priority-level"))
+		util.MustBindEnv("dispatchThrottling.lowPriorityLevel", "OPENFGA_DISPATCH_THROTTLING_LOW_PRIORITY_LEVEL")
```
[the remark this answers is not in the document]
adriantam: See [internal/server/config/config.go](https://github.com/openfga/openfga/pull/1440/files#diff-d2ecce123fd0776f269e6d6542f943ddecfa7e284169ce2db8467111801d06f2)

pkg/server/server.go
adriantam: High priority means no shaping.  Perhaps we relabel medium priority to high priority?

pkg/server/server.go
```diff
@@ -269,6 +279,48 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
 	}
 }
 
+// WithRateLimitedCheckResolverEnabled sets whether dispatch throttling is enabled
+func WithRateLimitedCheckResolverEnabled(enabled bool) OpenFGAServiceV1Option {
```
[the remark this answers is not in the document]
adriantam: updated comments

pkg/server/server.go
```diff
@@ -269,6 +279,48 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
...
+	return func(s *Server) {
+		s.rateLimitedCheckResolverEnabled = enabled
+	}
+}
+
+// WithRateLimitedCheckResolverTimerTickerFrequency sets how frequent dispatch throttling is enabled
```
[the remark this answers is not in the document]
adriantam: updated comments

pkg/server/server.go
```diff
@@ -269,6 +279,48 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
...
+func WithRateLimitedCheckResolverTimerTickerFrequency(frequency time.Duration) OpenFGAServiceV1Option {
+	return func(s *Server) {
+		s.rateLimitedCheckResolverTimerTickerFrequency = frequency
+	}
+}
+
+// WithRateLimitedCheckResolverLowPriorityLevel sets the number of dispatches a request incur before being placed on low priority shaper
+func WithRateLimitedCheckResolverLowPriorityLevel(level uint32) OpenFGAServiceV1Option {
```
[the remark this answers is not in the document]
adriantam: updated comments

cmd/run/run.go
adriantam: Added in more descriptions

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,166 @@
...
+	delta := currentNumDispatch - r.config.MediumPriorityLevel
+	if currentNumDispatch < r.mediumPriorityLevel2 {
+		return delta%(r.config.MediumPriorityShaper) == 0
+	}
+	if currentNumDispatch < r.mediumPriorityLevel3 {
+		return delta%(r.mediumPriorityShaperFreq2) == 0
+	}
+	if currentNumDispatch < r.mediumPriorityLevel4 {
+		return delta%(r.mediumPriorityShaperFreq3) == 0
+	}
```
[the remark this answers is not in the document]
willvedd: To be clear, I'm actually advocating for fewer knobs. In practice, we don't know how or if developers will configure (or misconfigure) their throttling. Removing the concept of various priority queues and shaping coefficients will actually make the API and code much simpler. And frankly, I'm still unconvinced that there's a need to "shape" the throttling at all. 

> Perhaps we can tackle this later when there is a need?

This is my stance on all the config that's being introduced in this PR. Let's add it when it's needed! 😄 

**Note:** Just my two cents. This is _not_ meant to be blocking feedback so long as there's a general consensus on API.

.config-schema.json
```diff
@@ -419,6 +419,47 @@
...
+                    "default": "50",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_MEDIUM_PRIORITY_LEVEL"
+                },
+                "mediumPriorityShaper": {
+                    "description": "initial frequency on un-throttled dispatches for medium priority jobs",
+                    "type": "integer",
+                    "default": "8",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_MEDIUM_PRIORITY_SHAPER"
+                }
+            }
```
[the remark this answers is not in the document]
poovamraj: Personally I feel that such minute way to configure the system is a good thing. If we consider tools like RDBMS (postgres) or file systems, they have such fine tuned way to configure things as the ones who will be tweaking these are engineers and they might want to change these values for their specific edge cases.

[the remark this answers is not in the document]
poovamraj: > could we not have a singular throttling threshold that limits this huge disparity.

I agree with this though, instead of having 2 queues. Why not have a single queue? Having a single DispatchThrottledQueue makes it much simpler than a LowPriorityQueue and a MediumPriorityQueue

About the requests not being gradual, why not we a single "THRESHOLD_VALUE" after which we send the request again to the back of the queue

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+func (r *RateLimitedCheckResolver) ResolveCheck(ctx context.Context,
+	req *ResolveCheckRequest,
+) (*ResolveCheckResponse, error) {
+	start := time.Now()
+	queueName := "high_priority"
+
+	currentNumDispatch := req.DispatchCounter.Add(1)
+	if currentNumDispatch > r.config.MediumPriorityLevel {
```
poovamraj, line 154: We currently cannot find the root request (the first check) when resolving a check. If we are able to individually identify root request, we can put them at the back of the queue each time they cross a threshold. This way we wouldn't need multiple queues.

I might be missing context here on why above suggestion wouldn't work but would love to understand it.

poovamraj:
Will it be possible to put the dispatches in queue depending on available system resource or current number of dispatches being executed?

Why should we throttle a high dispatch request by exponential time if that is the only request that is being executed? 

Wouldn't this reduce the request duration for tail end of high dispatch requests?

docker-compose.override.yaml
```diff
@@ -10,12 +10,17 @@ services:
       otel-collector:
         condition: service_started
     environment:
-      - OPENFGA_TRACE_ENABLED=true
+      - OPENFGA_TRACE_ENABLED=false
```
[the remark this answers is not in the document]
adriantam: This is a mistake - I will re-enable tracing.

cmd/run/run.go
```diff
@@ -202,6 +202,18 @@ func NewRunCommand() *cobra.Command {
...
+
+	flags.Uint32("dispatch-throttling-low-priority-level", defaultConfig.DispatchThrottling.LowPriorityLevel, "define the number of dispatches to be considered in low priority throttling queue. Requests with dispatch count above this threshold will be placed into the low priority queue. Low priority queue requests are processed less frequently than other requests.")
+
+	flags.Uint32("dispatch-throttling-low-priority-shaper", defaultConfig.DispatchThrottling.LowPriorityShaper, "number of tickers required to dispatch a low priority work. System will release one job from the low priority queue every Nth tick (configured via this variable).")
+
+	flags.Uint32("dispatch-throttling-medium-priority-level", defaultConfig.DispatchThrottling.MediumPriorityLevel, "define the number of dispatches to be considered in medium priority throttling queue. Check and list objects requests with dispatches higher than the medium priority level but lower than the low priority level will be placed in the medium priority queue. The medium priority queue are processed more frequently than the low priority queue.")
```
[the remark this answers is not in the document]
adriantam: Here, I really meant "medium priority level" as in the config parameter "dispatch-throttling-medium-priority-level".

internal/graph/check.go
```diff
@@ -38,6 +39,12 @@ type ResolveCheckRequest struct {
 	Context              *structpb.Struct
 	ResolutionMetadata   *ResolutionMetadata
 	VisitedPaths         map[string]struct{}
+	// This is different from ResolutionMetadata dispatch counter in that this counter needs to be
+	// incremented before dispatches are processed (where the ResolutionMetadata dispatch counter
+	// are incremented after the dispatches are processed). As this counter needs to be updated
+	// synchronously are multiple concurrent go-routine, it needs to use shared counter within
+	// the same request.
+	DispatchCounter *atomic.Uint32
```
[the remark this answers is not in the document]
adriantam: Updating the dispatch counter to avoid using atomic is the next item we'll look into.  At that time, we should have a single counter instead of two counters.

cmd/run/run.go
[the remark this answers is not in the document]
poovamraj: Makes sense 👍

.config-schema.json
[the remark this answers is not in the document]
adriantam: Even if we have a single "threshold_value", we will still need three knobs 
- level where we start shaping
- how frequently we shape at the beginning (we need to do that, otherwise, there is a sharp rise in latency between requests just over the  threshold vs. requests just below the threshold
- this single "threshold value" where we send the request again to the back of the queue

One concern with the single queue model is that the medium dispatches  (not too bad) are competing against the large dispatches.  We had seen in some cases where some model yield number of dispatches > 1000.  In this case, if we want to start shaping at 50 dispatches, requests with 60 dispatches will be competing against requests with 900 dispatches. This will make it unfair even when you send the large requests back to the line.  An analogy is that you have a bus with two groups (one group with 10 folks and another group with 1000 folks) and the folks line up randomly.  If you have a single line but penalize group 2 members by having them to go back to the line again, group 1's member will still have a hard time getting in the bus by virtue they have to compete against group 2.  If you have two lines, however, group 1 will have a better chance of getting into the bus because they are only competing against yourself.

For operators who only need a single queue, they can set the medium threshold to be equal low threshold.  In this way, they will effectively have a single queue. Perhaps I can update the configuration to point out that if only one queue is desired, set medium threshold level to be equal to low threshold level.

adriantam:
> Will it be possible to put the dispatches in queue depending on available system resource or current number of dispatches being executed?

It "could" be a future direction.  However, this current PR is a +1 moving to that direction.  If we scope this change in, the amount of work will grow exponentially as we also start work on monitoring these resources in real-time.

> 
> Why should we throttle a high dispatch request by exponential time if that is the only request that is being executed?


We are not throttling high dispatch by exponential time.  If you look at low priority queue, the decay is not exponential.  You are just competing against other low priority requests.  If you are the only "low priority" request, you will be dispatched every clock ticker (or every Nth clock ticker depending on your configuration).  The benefit of spreading out the dispatch is to avoid thundering herd effect and cause issues such as connection contention.
 
> Wouldn't this reduce the request duration for tail end of high dispatch requests?

I am not sure what you meant by "reduce request duration".  Do you mean increase latency of high dispatch request?  If so, yes, this is to improve latency of low number dispatches request at the expense of requests with high number dispatches.  Again, if the operator does not desire this dispatch throttling (and consider all requests equal in priority), they should not use this feature.  This feature is meant to have a differentiated service based on some form of criteria (in this case, favouring requests with low number of dispatches over requests with high number of dispatches).

pkg/server/server.go
```diff
@@ -327,8 +402,33 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+			zap.Duration("TimerTickerFrequency", s.rateLimitedCheckResolverTimerTickerFrequency),
+			zap.Uint32("MediumPriorityLevel", s.rateLimitedCheckResolverMediumPriorityLevel),
+			zap.Uint32("MediumPriorityShaper", s.rateLimitedCheckResolverMediumPriorityShaper),
+			zap.Uint32("LowPriorityLevel", s.rateLimitedCheckResolverLowPriorityLevel),
+			zap.Uint32("LowPriorityShaper", s.rateLimitedCheckResolverLowPriorityShaper),
+		)
+
+		rateLimitedCheckResolver := graph.NewRateLimitedCheckResolver(rateLimitConfig)
+		rateLimitedCheckResolver.SetDelegate(cycleDetectionCheckResolver)
+		s.rateLimitedCheckResolver = rateLimitedCheckResolver
```
jon-whit, line 424: Outer most should still remain cycle detection. If we've detected a cycle, then there is no further work to do because we can return the error promptly. We shouldn't incur a potential dispatch delay just to find a cycle.

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,176 @@
...
+		Buckets:                         []float64{1, 3, 5, 10, 25, 50, 100, 1000, 5000}, // Milliseconds. Upper bound is config.UpstreamTimeout.
+		NativeHistogramBucketFactor:     1.1,
+		NativeHistogramMaxBucketNumber:  100,
+		NativeHistogramMinResetDuration: time.Hour,
+	}, []string{"grpc_service", "grpc_method", "queue_name"})
+)
+
+func NewRateLimitedCheckResolver(
+	config RateLimitedCheckResolverConfig) *RateLimitedCheckResolver {
+	ticker := time.NewTicker(config.TimerTickerFrequency)
```
miparnisari, line 67: nit: you don't need this ticker to be a struct field, it can just be a local variable inside `runTicker` since that is the only place where it is used

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,176 @@
...
+		medPriorityQueue: make(chan bool),
+		done:             make(chan bool),
+	}
+	rateLimitedCheckResolver.setMediumConfig(config)
+	rateLimitedCheckResolver.delegate = rateLimitedCheckResolver
+	go rateLimitedCheckResolver.runTicker()
+	return rateLimitedCheckResolver
+}
+
+func (r *RateLimitedCheckResolver) setMediumConfig(config RateLimitedCheckResolverConfig) {
```
miparnisari, line 81: So, it looks like this function is taking the ranges

[0, mediumlevel, lowlevel]

and further dividing them into

[0, mediumlevel2, mediumlevel3, mediumlevel4, lowlevel]

is my understanding correct? if so can you add godocs for it? please also explain _why_ we need to further divide it

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,176 @@
...
+// This allows a check / list objects request to be gradually throttled.
+type RateLimitedCheckResolver struct {
+	delegate         CheckResolver
+	config           RateLimitedCheckResolverConfig
+	ticker           *time.Ticker
+	lowPriorityQueue chan bool
+	medPriorityQueue chan bool
+	done             chan bool
+
+	// these are helper config value to reduce calculation needed for medium shaper
```
miparnisari, line 43: nit: this comment isn't adding any value to me, i suggest removing it and adding godocs in lines 17 and 18 because i still don't know what "shaper" means

pkg/server/server_test.go
```diff
@@ -1373,13 +1423,58 @@ type repo
...
+		require.False(t, cfg.DispatchThrottling.Enabled)
+
+		ds := memory.New()
+		t.Cleanup(ds.Close)
+		s := MustNewServerWithOpts(
+			WithDatastore(ds),
+		)
+		t.Cleanup(s.Close)
+		require.Nil(t, s.rateLimitedCheckResolver)
+		require.False(t, s.rateLimitedCheckResolverEnabled)
```
jon-whit, line 1457: Please also add an assertion that the server's CheckResolver is of type `CycleDetectionCheckResolver` in addition to the assertions above.

pkg/server/server_test.go
```diff
@@ -1373,13 +1423,58 @@ type repo
...
+	})
+
+	t.Run("enable_rate_limited_check_resolver_enabled", func(t *testing.T) {
```
jon-whit, line 1460:
```suggestion
	t.Run("rate_limited_check_resolver_enabled", func(t *testing.T) {
```

pkg/server/server_test.go
```diff
@@ -1373,13 +1423,58 @@ type repo
...
+			WithRateLimitedCheckResolverEnabled(true),
+			WithRateLimitedCheckResolverMediumPriorityShaper(20),
+			WithRateLimitedCheckResolverMediumPriorityLevel(30),
+			WithRateLimitedCheckResolverLowPriorityShaper(40),
+			WithRateLimitedCheckResolverLowPriorityLevel(50),
+		)
+		t.Cleanup(s.Close)
+
+		require.NotNil(t, s.rateLimitedCheckResolver)
+		require.True(t, s.rateLimitedCheckResolverEnabled)
```
jon-whit, line 1474: Please also add an assertion that asserts the server's CheckResolver is of type `CycleDetectionCheckResolver`.

```go
_, ok := s.checkResolver.(*graph.CycleDetectionCheckResolver)
require.True(t, ok)
```

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	end := time.Now()
+	timeWaiting := end.Sub(start).Milliseconds()
+	rpcInfo := telemetry.RPCInfoFromContext(ctx)
+	rateLimitedCheckResolverDelayMsHistogram.WithLabelValues(
+		rpcInfo.Service,
+		rpcInfo.Method,
+		queueName,
+	).Observe(float64(timeWaiting))
```
miparnisari, line 173: what do you think about moving this block inside the `if`? otherwise, you will be calculating `end - start` even for high priority queue, which I imagine will be zero most of the time... so, we will be emitting lots of metric for it, which is kinda pointless

internal/graph/check.go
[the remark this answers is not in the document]
adriantam: I want to defer this part of work into separate PR to avoid expanding this PR's scope.  The work for this is already ongoing.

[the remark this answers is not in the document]
miparnisari: But it wouldn't expand this PR scope. Instead of doing `currentNumDispatch := req.DispatchCounter.Add(1)` you would do 

```go
resp, err := r.delegate.ResolveCheck(ctx, req)
if err != nil {
    // ...
}
currentNumDispatch := resp.GetResolutionMetadata().GetDispatchCount()
```

would that not work?

.config-schema.json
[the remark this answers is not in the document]
jon-whit: @miparnisari @jpadilla I'm curious what you two also think of this expansion of configuration and the implementation overhead?

As a reminder for us all reading this thread, the original premise and goal was simple - **throttle requests that reach a dispatch count more than some threshold, because after some threshold we cannot guarantee our SLO will be met**. 

A single dispatch leads to at least 1 database query, but our SLO targets 50ms. Subproblems of a dispatch are evaluated concurrently with some maximum bound of concurrency. So I leave a question to the reader - On average, how many dispatches does it take to exceed a 50ms budget? If we don't have an answer to that, then we should evaluate that. This metric will obviously depend on the model and the tuple data. 

My hot take on these changes - I'm very hesitant to introduce this level of configuration to the server without comparing the alternative implementation which was proposed originally prior to this work kicking off.  Namely, throttle proportionally to some configured threshold using an exponentially delaying ticker. Such an implementation would be much easier to maintain as a team and easier to configure and to explain to the community how to configure. I recognize this kind of implementation _could_ have some implications, but both implementations have implications. It's our job to compare the implications relative to one another, and I don't think we have the information needed to make that determination yet.

I wouldn't know how to tune these existing settings, and I also wouldn't even know where to start to explain how to configure these settings to an external OpenFGA operator. These default values were based on a distribution that is unique to Okta FGA traffic, which isn't necessarily representative of the average case. Would these defaults work well for the average case?


One last thing - a lot of these configs are oriented around trying to manage the "shape" of the distribution, but instead of observing the distribution and apply the throttle we are trying to approximate the shape with synthentics represented as `OPENFGA_DISPATCH_THROTTLING_LOW_PRIORITY_SHAPER` and `OPENFGA_DISPATCH_THROTTLING_MEDIUM_PRIORITY_SHAPER`. If we're going to do multi-priority throttling, then I may suggest we compute the quantiles of the dispatches in real-time using an online approximation of them (t-digest or the like) and then assign requests whose dispatch count is greater than p99 to the low priority and any others >threshold but <= p99 to the high priority queue. This would achieve the shaping we're seeking and would only require 2 queues and would also avoid the overhead of all of these configs.

internal/graph/check.go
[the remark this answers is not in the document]
adriantam: Not quite: We need to increment the counter when we decide whether we want to do dispatch throttling or not.  For example, imagine you have two simultaneous dispatches.

Normally, you would have expect one dispatch will have dispatch count of X and the other to be X + 1.  However, in the sample code above, it is possible that you will observe both to be X+1 because currentNumDispatch is measured after it is incremented. This is not necessary a bad thing - but this warrants consideration of where we want to really increment that counter.  

This is why I propose to defer this to a subsequent PR

pkg/server/server_test.go
adriantam: Done

pkg/server/server_test.go
adriantam: done

internal/graph/rate_limited_check_resolver.go
```diff
@@ -0,0 +1,176 @@
...
+	start := time.Now()
+	queueName := "high_priority"
+
+	currentNumDispatch := req.DispatchCounter.Add(1)
+	if currentNumDispatch > r.config.MediumPriorityLevel {
+		if currentNumDispatch > r.config.LowPriorityLevel {
+			queueName = "low_priority"
+			<-r.lowPriorityQueue
+		} else {
+			if r.shouldWait(currentNumDispatch) {
```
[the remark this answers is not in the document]
adriantam: this should be clearer now that there is only one level

internal/graph/rate_limited_check_resolver.go
adriantam: comment obsolete as we only have 1 queue

internal/graph/rate_limited_check_resolver.go
adriantam: Rename to rate

internal/graph/rate_limited_check_resolver.go
adriantam: done

internal/graph/rate_limited_check_resolver.go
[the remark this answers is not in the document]
adriantam: Since there is only 1 level - this comment is now obsolete.

.config-schema.json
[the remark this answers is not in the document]
adriantam: Updated to use only 1 level.  We can add the multi-level back when it is needed.

.config-schema.json
```diff
@@ -430,6 +430,35 @@
...
+                    "default": "10µs",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_TIME_TICKER_FREQUENCY"
+                },
+                "level": {
+                    "description": "define the number of dispatches to be throttled",
+                    "type": "integer",
+                    "default": "100",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_LEVEL"
+                },
+                "rate": {
```
jon-whit, line 455: What's the difference between the `timeTickerFrequency` and the `rate`? Aren't they derivatives of one another? e.g. if I configure a rate of 10 dispatches/sec then the ticker interval is defined as 1/10. The ticker frequency is always the inverse of the rate.

Shouldn't the rate be an exponentially decaying function of the delta between the `level` and the current observed dispatch_count (e.g. `(X - dispatch.level)*exp`)?

adriantam: timeTickerFrequency = how frequent time.Ticker will run to evaluate dispatch
rate = how many ticks to launch a job

Maybe we use a different word to represent "rate"?

>Shouldn't the rate be an exponentially decaying function of the delta between the level and the current observed dispatch_count (e.g. (X - dispatch.level)*exp)?

No, there is no notion of exponential decay.  We have removed all exponential decay when we remove the multi-level design.

internal/graph/rate_limited_check_resolver.go
adriantam: updated

internal/graph/check.go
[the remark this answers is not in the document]
willvedd: Separate counters are necessary to complete the two tasks at hand – throttling and reporting. Count on request is to signal to determine wether to throttle as the dispatches expand recursively. Those dispatches could fan-out into hundreds more. That count is necessary at the time of dispatch to make that decision. Response on the other hand is returning dispatch count for reporting/metrics. By the time the response happens, its too late for throttling to occur. 

In short, it makes sense to me to keep these separate. I'm not sure how you'd combine.

internal/graph/check.go
```diff
@@ -38,6 +39,12 @@ type ResolveCheckRequest struct {
 	Context              *structpb.Struct
 	ResolutionMetadata   *ResolutionMetadata
 	VisitedPaths         map[string]struct{}
+	// This is different from ResolutionMetadata dispatch counter in that this counter needs to be
+	// incremented before dispatches are processed (where the ResolutionMetadata dispatch counter
+	// are incremented after the dispatches are processed). As this counter needs to be updated
+	// synchronously are multiple concurrent go-routine, it needs to use shared counter within
```
miparnisari, line 45: > As this counter needs to be updated synchronously by multiple concurrent go-routine

This is not happening. Every time we call `dispatch` inside `LocalChecker`, the request object is a copy of the parent object, so you lose the "shared" aspect of it.

internal/graph/check.go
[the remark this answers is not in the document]
miparnisari: I understand now. This is a global counter. I have a proposal in https://github.com/openfga/openfga/pull/1453/files so that we put this counter within the request metadata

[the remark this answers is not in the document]
adriantam: >Separate counters are necessary to complete the two tasks at hand – throttling and reporting. Count on request is to signal to determine wether to throttle as the dispatches expand recursively. Those dispatches could fan-out into hundreds more. That count is necessary at the time of dispatch to make that decision. Response on the other hand is returning dispatch count for reporting/metrics. By the time the response happens, its too late for throttling to occur.

Exactly.  Not that it cannot be done - but let's defer this to a different PR so that we can focus on the dispatch throttling piece.

.config-schema.json
jon-whit: > timeTickerFrequency = how frequent time.Ticker will run to evaluate dispatch
rate = how many ticks to launch a job

So if `timeTickerFrequency=1ms` and `rate=10` then we'll dispatch at a rate of `10ms`. What I'm saying is that we don't need 2 configs to define that. `timeTickerFrequency` defines the unit, rate defines the value. Instead, how about we just define `rate=10ms` and/or maybe a better name is `interval=10ms`.

> No, there is no notion of exponential decay. We have removed all exponential decay when we remove the multi-level design.

This is my bigger concern - what is stopping us from doing expo decay without the multi-level design, that's not clear to me? The delay is a function of the delta between the dispatch threshold and the current dispatch count multiplied by some exponential.

internal/graph/dispatch_throttling_check_resolver.go
```diff
@@ -0,0 +1,123 @@
...
+		NativeHistogramMinResetDuration: time.Hour,
+	}, []string{"grpc_service", "grpc_method"})
+)
+
+func NewDispatchThrottlingCheckResolver(
+	config DispatchThrottlingCheckResolverConfig) *DispatchThrottlingCheckResolver {
+	dispatchThrottlingCheckResolver := &DispatchThrottlingCheckResolver{
+		config:          config,
+		ticker:          time.NewTicker(config.TimerTickerFrequency),
+		throttlingQueue: make(chan bool),
```
jon-whit, line 55: No need for a `bool` here and below for the `done` channel. We don't need a value, just a signal. So use the empty struct instead `struct{}`, because it has no overhead as compared to a bool type which takes a byte.

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c *countResolveCheck) ResolveCheck(ctx context.Context, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
+	c.count.Add(1)
+	return &ResolveCheckResponse{Allowed: true}, nil
+}
+
+func (c *countResolveCheck) Close() {}
+
+func (c *countResolveCheck) NumResolveCheck() uint32 {
```
jon-whit, line 31: It feels odd to wrap a single function call `c.count.Load()` into a function. Can we just `c.count.Load()` where needed instead?

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -0,0 +1,105 @@
...
+		// the first time tick will not release the job
+		newCount := dut.handleTimeTick(0)
+		time.Sleep(1 * time.Millisecond)
+		require.Equal(t, uint32(0), c.NumResolveCheck())
+		require.Equal(t, uint32(1), newCount)
+
+		// the second time tick will release the job
+		newCount = dut.handleTimeTick(newCount)
+		goFuncDone.Wait()
+		require.Equal(t, uint32(1), c.NumResolveCheck())
```
jon-whit, line 102: This test is confusing at first glance, because if the tick frequency is 1hr and we require 2 ticks, then I wouldn't expect ResolveCheck to increment the dispatch count until 2h later.

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (c *countResolveCheck) NumResolveCheck() uint32 {
+	return c.count.Load()
+}
+
+func TestDispatchThrottlingCheckResolver(t *testing.T) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+	t.Run("below_shaping_level_should_not_be_shaped", func(t *testing.T) {
```
jon-whit, line 39: This test name and the one below don't make sense to me - it's not obvious what shaping is referring to here.

I may suggest `dispatch_below_threshold_resolves_immediately`.

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
+)
+
+type countResolveCheck struct {
```
jon-whit, line 14: I'd lean in favor of removing this code as it is a pseudo-mock. We already have a mocked CheckResolver.

```go
defer mockController.Finish()
mockCheckResolver := NewMockedCheckResovler(mockController)
```

We can make the appropriate assertions on calls to `ResolveCheck` on the mock. For example,
```go
mockCheckResolver.EXPECT().
	ResolveCheck(gomock.Any(), gomock.Any()).
	Return(&ResolveCheckResponse{Allowed: true}, nil).
	MaxTimes(1)
```

See the example below in my other comment for a full example.

internal/graph/dispatch_throttling_check_resolver_test.go
jon-whit, line 39: We can drop the use of the pseudo-mock and use our existing `MockCheckResolver`. For example,

```go
t.Run("below_shaping_level_should_not_be_shaped", func(t *testing.T) {
	dispatchThrottlingCheckResolverConfig := DispatchThrottlingCheckResolverConfig{
		TimerTickerFrequency: 1 * time.Hour,
		Level:                200,
		Rate:                 8,
	}
	dut := NewDispatchThrottlingCheckResolver(dispatchThrottlingCheckResolverConfig)
	defer dut.Close()

	mockCtrl := gomock.NewController(t)
	t.Cleanup(mockCtrl.Finish)

	mockCheckResolver := NewMockCheckResolver(mockCtrl)
	dut.SetDelegate(mockCheckResolver)

	mockCheckResolver.EXPECT().
		ResolveCheck(gomock.Any(), gomock.Any()).
		Return(&ResolveCheckResponse{Allowed: true}, nil).
		MaxTimes(1)

	dispatchCounter := &atomic.Uint32{}
	_, err := dut.ResolveCheck(context.Background(), &ResolveCheckRequest{DispatchCounter: dispatchCounter})
	require.NoError(t, err)
	require.Equal(t, uint32(1), dispatchCounter.Load())
})
```

internal/server/config/config_test.go
```diff
@@ -132,6 +132,45 @@ func TestVerifyConfig(t *testing.T) {
...
+			TimeTickerFrequency: 10 * time.Microsecond,
+			Rate:                30,
+			Level:               0,
+		}
+
+		err := cfg.Verify()
+		require.Error(t, err)
+	})
+
+	t.Run("non_positive_dispatch_level", func(t *testing.T) {
```
jon-whit, line 162:
```suggestion
	t.Run("non_positive_dispatch_rate", func(t *testing.T) {
```

pkg/server/server.go
```diff
@@ -291,6 +299,45 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
 	}
 }
 
+// WithDispatchThrottlingCheckResolverEnabled sets whether dispatch throttling is enabled.
+// Enable this feature will prioritize processing simple check and list objects requests
```
jon-whit, line 303: Enabling this feature will prioritize dispatched requests requiring less than the configured dispatch level threshold over requests whose dispatch count exceeds the configured level.

pkg/server/server.go
```diff
@@ -291,6 +299,45 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
...
+	return func(s *Server) {
+		s.dispatchThrottlingCheckResolverEnabled = enabled
+	}
+}
+
+// WithDispatchThrottlingCheckResolverTimerTickerFrequency defines how frequent dispatch throttling will be evaluated.
+// TimeTickerFrequency controls how frequently throttled dispatch requests are evaluated to determine whether
+// it can be processed.
+// This value should not be too small (i.e., in the ns ranges) as i) there are limitation in timer resolution
+// and ii) very small value will result in frequent processing of
```
jon-whit, line 316: Dangling sentence. 

> ii) very small value will result in a higher frequency of processing dispatches, which diminishes the value of the throttling.

pkg/server/server.go
```diff
@@ -291,6 +299,45 @@ func WithMaxAuthorizationModelSizeInBytes(size int) OpenFGAServiceV1Option {
...
+func WithDispatchThrottlingCheckResolverTimerTickerFrequency(frequency time.Duration) OpenFGAServiceV1Option {
+	return func(s *Server) {
+		s.dispatchThrottlingCheckResolverTimerTickerFrequency = frequency
+	}
+}
+
+// WithDispatchThrottlingCheckResolverLevel define the number of dispatches to be considered in
```
jon-whit, line 323: > WithDispatchThrottlingCheckResolverLevel defines the level or threshold that has to be met for dispatch throttling to apply.

I feel we can drop the rest of this godoc comment, because we don't have a notion of high priority or low priority. We either have requests that are throttled or we dont.

.config-schema.json
jpadilla: suggestion: for starters, the only two config knobs we should expose is enabled/level. Let's keep the defaults for the rate/interval in code for now.

pkg/server/server.go
```diff
@@ -339,8 +391,29 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+			zap.Uint32("LowPriorityRate", s.dispatchThrottlingCheckResolverRate),
+		)
+
+		dispatchThrottlingCheckResolver := graph.NewDispatchThrottlingCheckResolver(dispatchThrottlingConfig)
+		dispatchThrottlingCheckResolver.SetDelegate(localChecker)
+		s.dispatchThrottlingCheckResolver = dispatchThrottlingCheckResolver
+
+		cycleDetectionCheckResolver.SetDelegate(dispatchThrottlingCheckResolver)
+		localChecker.SetDelegate(cycleDetectionCheckResolver)
+	} else {
```
jon-whit, line 413: This else is redunadnt as these delegate assignments remain unchanged and are the defaults if dispatch throttling is disabled. I recommend an alternative setup that looks like this, for example:

```go
cycleDetectionCheckResolver := graph.NewCycleDetectionCheckResolver()
s.checkResolver = cycleDetectionCheckResolver

localChecker := graph.NewLocalChecker(
)

cycleDetectionCheckResolver.SetDelegate(localChecker)
localChecker.SetDelegate(cycleDetectionCheckResolver)

var dispatchThrottlingCheckResolver *graph.DispatchThrottlingCheckResolver
if s.dispatchThrottlingCheckResolverEnabled {
    ...

	dispatchThrottlingCheckResolver = graph.NewDispatchThrottlingCheckResolver(...)
	dispatchThrottlingCheckResolver.SetDelegate(localChecker)
	s.dispatchThrottlingCheckResolver = dispatchThrottlingCheckResolver

	cycleDetectionCheckResolver.SetDelegate(dispatchThrottlingCheckResolver)
}

if s.checkQueryCacheEnabled {
	...

	cachedCheckResolver := graph.NewCachedCheckResolver(...)
	s.cachedCheckResolver = cachedCheckResolver

	cachedCheckResolver.SetDelegate(localChecker)

	if dispatchThrottlingCheckResolver != nil {
		dispatchThrottlingCheckResolver.SetDelegate(cachedCheckResolver)
	} else {
		cycleDetectionCheckResolver.SetDelegate(cachedCheckResolver)
	}
}
```

internal/graph/check.go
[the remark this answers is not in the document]
jon-whit: If you're actively in `DispatchThrottlingCheckResolver#ResolveCheck`, then you'e already dispatched the subproblem. You've gone through 2 layers of the onion already `CycleDetectionCheckResolver` and `DispatchThrottlingCheckResolver`. In the future this could even mean that you've distributed the problem to another remote server. Consequently, incrementing the dispatch count should happen as you are dispatching it, not after you've dispatched.

internal/server/config/config.go
```diff
@@ -31,6 +31,10 @@ const (
 	// care should be taken here - decreasing can cause API compatibility problems with Conditions
 	DefaultMaxConditionEvaluationCost = 100
 	DefaultInterruptCheckFrequency    = 100
+
+	DefaultDispatchThrottlingEnabled             = false
+	DefaultDispatchThrottlingTimeTickerFrequency = 10 * time.Microsecond
```
jon-whit, line 36: This is likely too aggressive. Most OS' won't be able to provide this level of precision on the timer. 1ms is about the highest precision resolution in general on Unix systems, but on windows it's much worse.

See https://pkg.go.dev/time#hdr-Timer_Resolution

We may consider moving this number up to 1.5-5ms range?

adriantam: Based on my POC test, if this is in the ms range, the throttling becomes very aggressive and "bad" requests will timeout very quickly.  Imagine in the case where 
- check requests timeout in 3 seconds (i.e., 3000ms) 
- you have 7 check request with 500 dispatches 
- throttling start at 50 dispatches

450 * 7 =   3150 (or 3.15s) 
which means some of these requests will timeout even without counting other delays.

.config-schema.json
adriantam: updated to remove rate and have the frequency instead.

internal/graph/dispatch_throttling_check_resolver.go
adriantam: updated

internal/graph/dispatch_throttling_check_resolver_test.go
adriantam: obsolete when we use the MockCheckResolver

internal/graph/dispatch_throttling_check_resolver_test.go
adriantam: added comment to clarify

internal/graph/dispatch_throttling_check_resolver_test.go
adriantam: updated

internal/graph/dispatch_throttling_check_resolver_test.go
adriantam: change to MockCheckResolver

internal/server/config/config_test.go
adriantam: obsolete as this test is removed as we no longer have rate

pkg/server/server.go
adriantam: updated comment

.config-schema.json
```diff
@@ -430,6 +430,29 @@
...
+        "dispatchThrottling": {
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "description": "enable throttling when request's number of dispatches is high",
+                    "type": "bool",
+                    "default": "false",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_ENABLED"
+                },
+                "timeTickerFrequency": {
```
jon-whit, line 443:
```suggestion
                "frequency": {
```

internal/graph/dispatch_throttling_check_resolver.go
```diff
@@ -0,0 +1,112 @@
...
+	TimerTickerFrequency time.Duration
+	Level                uint32
+}
+
+// DispatchThrottlingCheckResolver will prioritize requests with fewer dispatches over
+// requests with more dispatches.
+// Initially, request's dispatches will not be throttled and will be processed
+// immediately. When the number of request dispatches is above the Level, the dispatches are placed
+// in the throttling queue. One item form the throttling queue will be processed every Nth ticker
+// (configured via config's Rate parameter).
```
jon-whit, line 25: Where's the Rate parameter?

pkg/server/server.go
```diff
@@ -352,7 +409,7 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 			graph.WithLogger(s.logger),
 			graph.WithCacheTTL(s.checkQueryCacheTTL),
 		)
-		s.cachedCheckResolverCloser = cachedCheckResolver.Close
+		s.cachedCheckResolver = cachedCheckResolver
 
 		cachedCheckResolver.SetDelegate(localChecker)
```
jon-whit, line 414: You're missing a setup of the `dispatchThrottlingCheckResolver` in here.

There should be a:
```go
if dispatchThrottlingChecResolver != nil {
    disptachThrottlingCheckResolver.SetDelegate(cachedCheckResolver)
}
```
in this block of code somewhere.

The order is this `CycleDetectionCheckResolver` --> (optional) `DispatchThrottlingCheckResolver` --> (optional) `CachedCheckResolver` -->` LocalChecker` --> `CycleDetectionCheckResolver`

.config-schema.json
```diff
@@ -430,6 +430,29 @@
...
+                    "description": "defines how frequent dispatch throttling will be evaluated",
```
willvedd, line 444: **Slightly pedantic:**The dispatch throttling isn't being evaluated, the deprioritized queue is.

```suggestion
                    "description": "defines the frequency period that the deprioritized throttling queue is evaluated. A higher value will result in more aggressive throttling.",
```

.config-schema.json
```diff
@@ -430,6 +430,29 @@
...
+                    "type": "duration",
+                    "default": "10µs",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_TIME_TICKER_FREQUENCY"
+                },
+                "level": {
+                    "description": "define the number of dispatches to be throttled",
```
willvedd, line 450: "Dispatch" is not a term that we use externally. I'm not sure if it makes sense to introduce it now or use a term that conveys a similar meaning. 

Perhaps something like:

```suggestion
                    "description": "define the number of recursive operations to occur before getting throttled",
```

.config-schema.json
```diff
@@ -430,6 +430,29 @@
...
+                    "default": "false",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_ENABLED"
+                },
+                "timeTickerFrequency": {
+                    "description": "defines how frequent dispatch throttling will be evaluated",
+                    "type": "duration",
+                    "default": "10µs",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_TIME_TICKER_FREQUENCY"
+                },
+                "level": {
```
willvedd, line 449: IMO "level" doesn't convey what this is. Perhaps:

```suggestion
                "limit": {
```
or
```suggestion
                "throttling_threshold": {
```
or

```suggestion
                "threshold": {
```

willvedd (commented):
Really nice to see this take shape. I think this simplified, linear throttling makes a lot of sense, especially as an initial implementation. Great work!

A couple of minor terminology requests but otherwise approve of this PR in principle.

internal/graph/dispatch_throttling_check_resolver.go
adriantam: corrected comment

pkg/server/server.go
adriantam: test added

.config-schema.json
adriantam: Changed to threshold

.config-schema.json
adriantam: updated

.config-schema.json
adriantam: taken suggestion

jon-whit (approved):

poovamraj (dismissed):

.config-schema.json
```diff
@@ -430,6 +430,29 @@
...
+        "dispatchThrottling": {
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "description": "enable throttling when request's number of dispatches is high",
+                    "type": "bool",
+                    "default": "false",
+                    "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_ENABLED"
+                },
+                "frequency": {
```
jon-whit, line 443: nit (non-blocking): my opinion is that we shouldn't expose this as a flag yet. I think we should just choose a sensible value and if we find we need to expose it then we can do so at a later time.

adriantam: I want to give us the flexibility to adjust the level of shaping in the field without having to spin new builds. This value does not need to be set by the user and we can update the default if we find some other optimal values

pkg/server/server.go
```diff
@@ -352,10 +408,14 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
-		cycleDetectionCheckResolver.SetDelegate(cachedCheckResolver)
+		if s.dispatchThrottlingCheckResolver != nil {
+			s.dispatchThrottlingCheckResolver.SetDelegate(cachedCheckResolver)
+		} else {
```
miparnisari, line 416: I pulled your PR and annotated the code with comments showing how the circular linked list of resolvers is wired and it looks like this

image

The `if` branch looks incorrect.

adriantam: Please see https://github.com/openfga/openfga/blob/5810589f38dc5ba91b5a871e75050171a2927aef/pkg/server/server_test.go#L1532

miparnisari: Never mind - it is correct, i forgot that in this `if`, the other `if` is also running

Issue #1428, opened 2024-03-08
ListUsers API


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

https://github.com/openfga/roadmap/issues/16

### Describe the ideal solution

New ListUsers and StreamingListUsers API

meldsza:
Will be fixed by: #1433 
(just tagging it because I see that it is not tagged here)

PR #1441, opened 2024-03-12
test: add more integration tests for ListUsers API

miparnisari:
## Description

Suggestion for reviewers: hit "hide whitespace" when reviewing file changes.

- Adding more assertions for ListUsers API. 
    - We have internal tickets tracking the extension of the tests to include coverage for ABAC (DXAZC-1916), contextual tuples (DXAZC-1915) and `and, but not` (DXAZC-1921)

Some reason for tests failures:
- contextual tuples need to be parsed
- duplicate results need to be cleaned
- if the filter is `type#relation`, we need to include `relation` in the response
- cycles need to be detected
- etc..

image

## References

### DIFF
--- a/assets/tests/consolidated_1_1_tests.yaml
+++ b/assets/tests/consolidated_1_1_tests.yaml
@@ -44,7 +44,7 @@ tests:
         listUsersAssertions:
           - request:
               filters:
-              - user
+                - user
               object: document:1 # exists in store
               relation: viewer
             expectation:
@@ -224,6 +224,21 @@ tests:
               relation: viewer
             expectation:
               - document:2
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            expectation:
+              - user:aardvark
+          - request:
+              filters:
+                - user
+              object: document:2
+              relation: viewer
+            expectation:
+              - user:badger
 
   - name: this_and_intersection
     stages:
@@ -370,6 +385,14 @@ tests:
               relation: viewer
             expectation:
               - document:1
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            expectation:
+              - user:aardvark
 
   - name: computed_userset_and_union
     stages:
@@ -414,6 +437,21 @@ tests:
               relation: viewer
             expectation:
               - document:2
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            expectation:
+              - user:aardvark
+          - request:
+              filters:
+                - user
+              object: document:2
+              relation: viewer
+            expectation:
+              - user:badger
 
   - name: simple_computeduserset_indirect_ref
     stages:
@@ -452,6 +490,27 @@ tests:
             expectation:
               - folder:a
               - folder:b
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: folder:a
+              relation: viewer
+            expectation:
+              - user:anne
+          - request:
+              filters:
+                - user
+              object: folder:b
+              relation: viewer
+            expectation:
+              - user:anne
+          - request:
+              filters:
+                - folder
+              object: folder:a
+              relation: viewer
+              expectation:
 
   - name: computed_userset_and_intersection
     stages:
@@ -607,6 +666,20 @@ tests:
               relation: viewer
             expectation:
               - document:1
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            expectation:
+              - user:aardvark
+          - request:
+              filters:
+                - folder
+              object: document:1
+              relation: viewer
+            expectation:
 
   - name: tuple_to_userset_and_tuple_to_userset
     stages:
@@ -651,6 +724,26 @@ tests:
               relation: viewer
             expectation:
               - document:1
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            expectation:
+              - user:aardvark
+          - request:
+              filters:
+                - group
+              object: document:1
+              relation: viewer
+            expectation:
+          - request:
+              filters:
+                - folder
+              object: document:1
+              relation: viewer
+            expectation:
 
   - name: tuple_to_userset_and_union
     stages:
@@ -703,6 +796,21 @@ tests:
               relation: viewer
             expectation:
               - document:1
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            expectation:
+              - user:aardvark
+              - user:badger
+          - request:
+              filters:
+                - user
+              object: document:2
+              relation: viewer
+            expectation:
 
   - name: tuple_to_userset_and_intersection
     stages:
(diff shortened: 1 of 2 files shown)


willvedd (commented):
Looks great. Only thing I ask is that we annotate the cases that are skipped with why they're skipped. That way, when we fix the underlying functionality, we can be sure to uncomment the appropriate tests.

jon-whit:
Are we not including the ABAC test files in the test suite? Or are we just choosing to neglect them at this time?

Referring to https://github.com/openfga/openfga/pull/1441/files#diff-ea7dfe8d6bd844454c46e92cedf2894359bcd2b613b76d9af0d7b008037b8675R53

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -452,6 +490,27 @@ tests:
             expectation:
               - folder:a
               - folder:b
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: folder:a
+              relation: viewer
```
jon-whit, line 498:
```suggestion
              relation: can_view
```

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -452,6 +490,27 @@ tests:
...
+            expectation:
+              - user:anne
+          - request:
+              filters:
+                - user
+              object: folder:b
+              relation: viewer
```
jon-whit, line 505:
```suggestion
              relation: can_view
```

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -452,6 +490,27 @@ tests:
...
+            expectation:
+              - user:anne
+          - request:
+              filters:
+                - folder
+              object: folder:a
+              relation: viewer
```
jon-whit, line 512:
```suggestion
              relation: can_view
```

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -651,6 +724,26 @@ tests:
...
+            expectation:
+              - user:aardvark
+          - request:
+              filters:
+                - group
+              object: document:1
+              relation: viewer
+            expectation:
+          - request:
```
jon-whit, line 741: We should consider adding one more request assertion just to ensure we don't expand any of our parent ancestors.

e.g.
```yaml
- request:
     filters:
       - group
     object: document:1
     relation: viewer
  expectation:
```

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -2196,7 +2355,14 @@ tests:
...
               - document:public
-
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:public
+              relation: viewer
+            expectation:
+              - user:*
```
jon-whit, line 2365: Would you please add a tuple `user:jon` above and assert we return the wildcard along with the explicit user as well? e.g.

```
expectation:
  - user:*
  - user:jon
```

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -2318,7 +2514,27 @@ tests:
...
```
jon-whit, line 2524: Consider adding a tuple `document:public#viewer@user:jon` and also asserting you get `[user:*, user:jon]` here. This gives us confidence we follow both paths and yield both the wildcard and the explicit subject.

miparnisari:
> Are we not including the ABAC test files in the test suite? Or are we just choosing to neglect them at this time?
> And what about contextual tuple variants of the same tests? Shouldn't we be running those alongside this suite as we grow it?

@jon-whit updated PR description with tracking tickets for those scenarios.

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -2437,7 +2674,28 @@ tests:
...
```
jon-whit, line 2684: This should be `[user:*, user:bob]`.

user:bob is part of the FGA group which is assigned to document:public#viewer in addition to the wildcard being assigned to `document:public#viewer` as well.

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -2787,6 +3159,17 @@ tests:
...
+        listUsersAssertions:
+          - request:
+              filters:
+                - user
+              object: document:1
+              relation: viewer
+            contextualTuples:
+              - object: document:1
+                relation: parent
+                user: user:* #invalid
```
jon-whit, line 3171: Add another test elsewhere where `user: group#undefined` for example. I don't see a test where we pass in a contextual tuple to ListUsers that defines a user that has an undefined relation on the userset relation.

I may have overlooked a test though..

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -2886,6 +3282,7 @@ tests:
               relation: owner
             expectation:
               - repo:1
+            errorCode: 2027 # ErrorCode_invalid_tuple
```
jon-whit, line 3285: Why do we have an errorCode and expectation above? The request and tuples looks fine to me.

Also, if these tests are passing, there is a false positive somehow as a result of this. We shouldn't be able to have an `expectation` and `errorCode` simultaneously and have the tests pass.

jon-whit:
I may have missed it, but I thought I noticed we didn't have any added or modified test cases where we added expectations of nested usersets. For example:

```
type user

type group
  relations
```
* `group:eng#member@group:fga#member`
* `group:fga#member@group:fga-backend#member`

```
ListUsers(group:eng#member, user_filters: [{"type": "group", "relation": "member"}]
```

should return ["group:fga#member", "group:fga-backend#member"]. In other words, we should continue to expand if there are more potentially more users/subjects of the same target objective (e.g. loopback on the same objectType#relation).

jon-whit (changes requested):
One main change around a test where there is both an error code and expectation. That seems erroneous.

assets/tests/consolidated_1_1_tests.yaml
miparnisari: 🤔 how is that different from lines 735-740?

assets/tests/consolidated_1_1_tests.yaml
miparnisari: woops

yeah the test harness should assert that if you defined `errorCode` you didn't define `expectation` and vice-versa

assets/tests/consolidated_1_1_tests.yaml
miparnisari: added: `validation_invalid_userset_in_contextual_tuple`

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -3537,7 +4129,49 @@ tests:
             expectation:
               - module:a
               - module:b
-
+  #        listUsersAssertions:
```
willvedd, line 4132: It's important to annotate the tests that are skipped with _why_ they're skipped. This way, when that functionality is added/patched, we can uncomment in those respective PRs. 

```suggestion
  #  listUsersAssertions are commented-out because ....
  #        listUsersAssertions:
```

willvedd (approved):
Looks good to me. I'm sure we might discover some small issues or want to add additional assertions but this serves as a good start IMO.

PR #613, opened 2023-03-07
feat: allow schema 1.0 flag

adriantam:
## Description
Disallow model schema 1.0 with WriteAuthorizationModel, WriteAssertions, Check, Expand, ListObjects, and Write API unless override flag is specified.

## References

https://openfga.dev/docs/modeling/migrating/migrating-schema-1-1#deprecation-timeline

## Review Checklist
- [X] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [X] The correct base branch is being used, if not `main`
- [X] I have added tests to validate that the change in functionality is working as expected

PR #1443, opened 2024-03-12
feat: support schema version 1.2 for modular models

### LINKED ISSUE #1402: Support Schema 1.2
### Checklist

As part of our [Modular Models](https://github.com/openfga/rfcs/pull/14/files) work, we are introducing a new schema version: 1.2, which should be accepted by the API.

- Update to a version of the API that supports module and filename in the metadata
- Add the enum: https://github.com/openfga/openfga/blob/314357d3e5434e18ee65134b21e32093228687a9/pkg/typesystem/typesystem.go#L28
- Add it as supported: https://github.com/openfga/openfga/blob/314357d3e5434e18ee65134b21e32093228687a9/pkg/typesystem/typesystem.go#L36
- Mark it as has type info: https://github.com/openfga/openfga/blob/314357d3e5434e18ee65134b21e32093228687a9/pkg/typesystem/typesystem.go#L392, https://github.com/openfga/openfga/blob/314357d3e5434e18ee65134b21e32093228687a9/pkg/typesystem/typesystem.go#L1018
- Ensure additional 1.1+ validations run on it: https://github.com/openfga/openfga/blob/main/internal/validation/validation.go#L338, https://github.com/openfga/openfga/blob/main/internal/validation/validation.go#L372
- Make any other necessary changes such as: https://github.com/openfga/openfga/blob/main/internal/graph/check.go#L694

_No response_

### Additional context

_No response_

ewanharris:
## Description

This PR adds support for using the schema version `1.2` behind the `enable-modular-models` experimental flag, and brings in the versions of openfga/api and openfga/language that support modular models.

As I was writing the additions I was thinking that the process of adding a new version isn't really scalable with the checks but I'm not familiar enough with the codebase to suggest any potential improvements here. Would love any feedback/prior consideration folks have had on this and would be happy to integrate into this PR.

Opening as a draft for now mostly just so the PR is here.

## References

Closes #1402

## Review Checklist


miparnisari (commented):
We had a similar need in the past: https://github.com/openfga/openfga/pull/613

We introduced two flags, one that allowed you to write models, and one that allowed you to call Check/ListObjects.

I think we should do something similar here.

poovamraj (commented):
Should we wrap the below line into it's own method so that it can be reused. Any future versions to be supported can be added here. Any version to be dropped from can be removed from here.

```go
t.GetSchemaVersion() == SchemaVersion1_1 || t.GetSchemaVersion() == SchemaVersion1_2)
```

internal/graph/check.go
```diff
@@ -704,7 +704,7 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 
 				// if the user value is a typed wildcard and the type of the wildcard
 				// matches the target user objectType, then we're done searching
-				if tuple.IsTypedWildcard(usersetObject) && typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 {
+				if tuple.IsTypedWildcard(usersetObject) && (typesys.GetSchemaVersion() == typesystem.SchemaVersion1_1 || typesys.GetSchemaVersion() == typesystem.SchemaVersion1_2) {
```
jon-whit, line 707: This block of code is residual and dead code given that we don't support 1.0 models anymore.

I recommend changing this boolean eval to simply:
```go
if typesystem.IsSchemaVersionSupported(typesys.GetSchemaVersion()) {
  ...
}
```

rhamzeh: I think having that check remain and do: `t.GetSchemaVersion() == SchemaVersion1_1 || t.GetSchemaVersion() == SchemaVersion1_2)` as @poovamraj suggested above might be worth it.

In this case it would return true for 1.2 only if experimental flag is set

.config-schema.json
```diff
@@ -78,7 +78,7 @@
             "type": "array",
             "items": {
                 "type": "string",
-                "enum": []
+                "enum": ["enable-modular-models"]
```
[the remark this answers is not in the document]
rhamzeh: The main reason I'm asking we keep it experimental is to let it be used for a week or two and gather feedback in case we need to change the API, or we missed a place where we're not properly handling 1.2

ewanharris:
Ah thanks for raising that @miparnisari, I'll refactor the existing one to be `enable-modular-models-write` and add in a new `enable-modular-models-evaluate`.

@poovamraj Yeah I was thinking simplifying it would make sense, @jon-whit's [suggestion](https://github.com/openfga/openfga/pull/1443#discussion_r1523977515) makes sense to me and allows us handle if the feature is enabled a little easier.

I'll work on the above and once those are done I think this should be ready for review

rhamzeh:
> We had a similar need in the past: #613
> 
> We introduced two flags, one that allowed you to write models, and one that allowed you to call Check/ListObjects.
> 
> I think we should do something similar here.

I don't think we need to. Unlike Schema 1.0, we're not planning on dropping support for 1.1.

All this new flag is doing is allowing support for 1.2. We expect to have a release out and remove that flag in a couple of weeks.

So as far as the server is concerned, only `WriteAuthorizationModel` will be concerned with rejecting `1.2` if the experimental flag is enabled. All the other pieces will treat 1.2 as supported.

That means if you can write the model, you can take all actions using it.

pkg/server/server_test.go
```diff
@@ -1383,3 +1383,33 @@ func TestDefaultMaxConcurrentReadSettings(t *testing.T) {
...
+			TypeDefinitions: []*openfgav1.TypeDefinition{
+				{
+					Type: "user",
+				},
+			},
+		})
+
+		require.ErrorIs(t, err, status.Error(codes.InvalidArgument, "modular models (schema version 1.2) is not supported"))
+	})
+}
```
miparnisari, line 1415: can you also make this a unit test in the `commands` pkg?

```go
func TestWriteAuthorizationModelWithExperimentalEnableModularModels(t *testing.T) {
	ctx := context.Background()

	defer mockController.Finish()
	mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)

	testCases := map[string]struct {
		enableModularModules bool
		inputSchema          string
		expectAllowed        bool
	}{
		`allow`: {
			enableModularModules: true,
			inputSchema:          typesystem.SchemaVersion1_2,
			expectAllowed:        true,
		},
		`forbid`: {
			enableModularModules: false,
			inputSchema:          typesystem.SchemaVersion1_2,
			expectAllowed:        false,
		},
	}

	for testName, test := range testCases {
		t.Run(testName, func(t *testing.T) {
			cmd := commands.NewWriteAuthorizationModelCommand(mockDatastore).
				WithWriteAuthModelEnableModularModels(test.enableModularModules)
			_, err := cmd.Execute(ctx, &openfgav1.WriteAuthorizationModelRequest{
				StoreId: storeID,
				TypeDefinitions: []*openfgav1.TypeDefinition{
					{
						Type: "user",
					},
				},
				SchemaVersion: test.inputSchema,
			})
			if test.expectAllowed {
				require.NoError(t, err)
			} else {
				require.ErrorIs(t, err, status.Error(codes.InvalidArgument, "modular models (schema version 1.2) is not supported"))
			}
		})
	}
}
```

the test you wrote we will keep it to test that the actual server flag is passed downstream

ewanharris: Added into 4245759b5bd252234c557bfed49c7ca6ff5f4321

miparnisari (dismissed):

rhamzeh (approved):

pkg/server/commands/write_authzmodel.go
```diff
@@ -38,11 +39,18 @@ func WithWriteAuthModelMaxSizeInBytes(size int) WriteAuthModelOption {
 	}
 }
 
+func WithWriteAuthModelEnableModularModels(enable bool) WriteAuthModelOption {
```
jon-whit, line 42:
```suggestion
func WithEnableModularModels(enable bool) WriteAuthModelOption {
```

pkg/server/commands/write_authzmodel.go
```diff
@@ -63,6 +71,10 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 		req.SchemaVersion = typesystem.SchemaVersion1_1
 	}
 
+	if !w.enableModularModels && req.GetSchemaVersion() == typesystem.SchemaVersion1_2 {
+		return nil, status.Error(codes.InvalidArgument, "modular models (schema version 1.2) is not supported")
```
jon-whit, line 75:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "modular models (schema version 1.2) are not supported")
```

pkg/server/server_test.go
```diff
@@ -1383,3 +1383,33 @@ func TestDefaultMaxConcurrentReadSettings(t *testing.T) {
...
+	mockController := gomock.NewController(t)
+	defer mockController.Finish()
+
+	mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)
+
+	s := MustNewServerWithOpts(
+		WithDatastore(mockDatastore),
+	)
+
+	t.Run("rejects_request_with_schema_version_1.2", func(t *testing.T) {
```
jon-whit, line 1400: Please add one more with enables it using the `With..` option and make sure it passes at the server level

pkg/server/commands/write_authzmodel.go
ewanharris: I agree this name is stupendously long, but it's in keeping with the existing `WithWriteAuthModelLogger` and `WithWriteAuthModelMaxSizeInBytes`, as well all other options throughout this folder follow the same `With<Command><Option>` pattern.

I'm assuming they are all intentionally prefixed to help differentiate usage as they're all under the `commands` namespace

ewanharris: nvm, I didn't refresh the page this morning and hadn't noticed this change was made

pkg/server/server_test.go
```diff
@@ -1563,3 +1563,60 @@ func TestDelegateCheckResolver(t *testing.T) {
...
+func TestWriteAuthorizationModelWithExperimentalEnableModularModels(t *testing.T) {
+	ctx := context.Background()
+	storeID := ulid.Make().String()
+
+	mockController := gomock.NewController(t)
+	defer mockController.Finish()
+
+	mockDatastore := mockstorage.NewMockOpenFGADatastore(mockController)
+
+	s := MustNewServerWithOpts(
```
adriantam, line 1576: nitpick: will it be better to move this inside the test "rejects_request_with_schema_version_1.2" to avoid shadow variable in the test "accepts_request_with_schema_version_1.2_if_experimental_flag_enabled"?

pkg/server/server_test.go
```diff
@@ -1563,3 +1563,60 @@ func TestDelegateCheckResolver(t *testing.T) {
 		require.True(t, ok)
 	})
 }
+
+func TestWriteAuthorizationModelWithExperimentalEnableModularModels(t *testing.T) {
+	ctx := context.Background()
```
adriantam, line 1568:
```suggestion
	t.Cleanup(func() {
		goleak.VerifyNone(t)
	})
	
	ctx := context.Background()
```
Please assert no leak.  Your changes should not cause leak - but by convention, please add the assertion to ensure no goroutine leak elsewhere

pkg/server/commands/write_authzmodel_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"go.uber.org/mock/gomock"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	mockstorage "github.com/openfga/openfga/internal/mocks"
+	"github.com/openfga/openfga/pkg/typesystem"
+)
+
+func TestWriteAuthorizationModelWithExperimentalEnableModularModels(t *testing.T) {
+	ctx := context.Background()
```
adriantam, line 19:
```suggestion
	t.Cleanup(func() {
		goleak.VerifyNone(t)
	})
	ctx := context.Background()
```

pkg/server/commands/write_authzmodel_test.go
```diff
@@ -0,0 +1,70 @@
...
+			enableModularModules: true,
+			inputSchema:          typesystem.SchemaVersion1_2,
+			expectAllowed:        true,
+		},
+		`forbid`: {
+			enableModularModules: false,
+			inputSchema:          typesystem.SchemaVersion1_2,
+			expectAllowed:        false,
+		},
+	}
```
adriantam, line 43: will it make sense to add test to ensure schema 1.1 is still allowed if enableModularModules = false?

pkg/server/server_test.go
ewanharris: Given the mock calls on `mockDatastore` do you think it makes sense to also move that into each test?

pkg/server/server_test.go
ewanharris: Adding the call looks to catch a leak, however I added to another test (`invalid_schema_error_in_write_model`) that uses a mocked datastore in conjunction with `WriteAuthorizationModel` and it also errors there on main.

Given that other tests are using `WriteAuthorizationModel` with a non-mocked datastore is there any chance that could be the cause? If not, any tips how I can go about debugging this?

```
found unexpected goroutines:
        [Goroutine 130149 in state select, with github.com/karlseguin/ccache/v3.(*Cache[...]).worker on top of the stack:
        github.com/karlseguin/ccache/v3.(*Cache[...]).worker(0x103e960e0)
                /Users/ewan/go/pkg/mod/github.com/karlseguin/ccache/v3@v3.0.5/cache.go:243 +0x14c
        created by github.com/karlseguin/ccache/v3.New[...] in goroutine 130148
                /Users/ewan/go/pkg/mod/github.com/karlseguin/ccache/v3@v3.0.5/cache.go:65 +0x56c
         Goroutine 130151 in state select, with github.com/karlseguin/ccache/v3.(*Cache[...]).worker on top of the stack:
        github.com/karlseguin/ccache/v3.(*Cache[...]).worker(0x103e960e0)
                /Users/ewan/go/pkg/mod/github.com/karlseguin/ccache/v3@v3.0.5/cache.go:243 +0x14c
        created by github.com/karlseguin/ccache/v3.New[...] in goroutine 130150
                /Users/ewan/go/pkg/mod/github.com/karlseguin/ccache/v3@v3.0.5/cache.go:65 +0x56c
        ]
```

pkg/server/commands/write_authzmodel_test.go
ewanharris: Yeah that makes sense, will add

rhamzeh: I think this is covered by the existing tests. We should add one that ensures 1.1 is accepted if enableModularModules = true

pkg/server/server_test.go
ewanharris: Moved in e7679cdab99fce275ceedfaf1a137458b08de86a

pkg/server/server_test.go
ewanharris: Added in e7679cdab99fce275ceedfaf1a137458b08de86a

pkg/server/commands/write_authzmodel_test.go
ewanharris: Added in 3d14bf43dc9e2d0b6ad2a4f93503c0d4cf2f0de6

pkg/server/commands/write_authzmodel_test.go
ewanharris: Tests added in 3d14bf43dc9e2d0b6ad2a4f93503c0d4cf2f0de6

miparnisari (approved):

rhamzeh (approved):

PR #1444, opened 2024-03-12
feat: Add Dispatch metrics as dimension to request duration metrics

poovamraj:
This PR adds `dispatch_count` dimension to request duration metrics

## Description
- Configurable buckets to measure request duration for dispatch count
- Config is provided with the default value of 
- Added these metrics to Check, ListObjects and StreamedListObjects
- Default values for these buckets are set to `[]string{"50", "200"}`

### Screenshots
Screenshot 2024-03-13 at 16 31 51

Screenshot 2024-03-13 at 16 31 43

## Review Checklist


cmd/run/run.go
```diff
@@ -200,6 +200,8 @@ func NewRunCommand() *cobra.Command {
 	// Unfortunately UintSlice/IntSlice does not work well when used as environment variable, we need to stick with string slice and convert back to integer
 	flags.StringSlice("request-duration-datastore-query-count-buckets", defaultConfig.RequestDurationDatastoreQueryCountBuckets, "datastore query count buckets used in labelling request duration by query count histogram")
 
+	flags.StringSlice("request-duration-datastore-dispatch-count-buckets", defaultConfig.RequestDurationDatastoreDispatchCountBuckets, "datastore query count buckets used in labelling request duration by dispatch count histogram")
+
```
miparnisari, line 204: Can we update the description for both of these? We don't track this metric for every API, only for the APIs tagged with "relationship queries" (https://openfga.dev/api/service#/Relationship%20Queries). _Note to self... we need to track this for Expand..._ 

Also, people reading these descriptions will probably not know what "dispatch count" means. I suggest: `buckets used in labelling request duration by dispatch count (i.e. Check subproblem count) histogram"` or something similar.

pkg/server/server.go
```diff
@@ -343,6 +353,10 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		return nil, fmt.Errorf("request duration datastore count buckets must not be empty")
 	}
 
+	if len(s.requestDurationByDispatchCountHistogramBuckets) == 0 {
+		return nil, fmt.Errorf("request duration datastore count buckets must not be empty")
```
miparnisari, line 357:
```suggestion
		return nil, fmt.Errorf("request duration by dispatch count buckets must not be empty")
```

pkg/server/server.go
```diff
@@ -86,7 +86,7 @@ var (
 		NativeHistogramBucketFactor:     1.1,
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
-	}, []string{"grpc_service", "grpc_method", "datastore_query_count"})
+	}, []string{"grpc_service", "grpc_method", "datastore_query_count", "dispatch_count"})
```
miparnisari, line 89: can you please update the description of this histogram? since it doesn't label by datastore query count alone anymore

pkg/server/server.go
```diff
@@ -709,6 +735,12 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		utils.Bucketize(uint(resp.GetResolutionMetadata().DatastoreQueryCount), s.requestDurationByQueryHistogramBuckets),
 	).Observe(float64(time.Since(start).Milliseconds()))
 
+	requestDurationByQueryHistogram.WithLabelValues(
+		s.serviceName,
+		methodName,
+		utils.Bucketize(uint(resp.GetResolutionMetadata().DispatchCount), s.requestDurationByDispatchCountHistogramBuckets),
+	).Observe(float64(time.Since(start).Milliseconds()))
```
miparnisari, line 742: i'm not very sure about how histograms work in prometheus so take this with a grain of salt, but shouldn't this be 

```go
requestDurationByQueryHistogram.WithLabelValues(
		s.serviceName,
		methodName,
                utils.Bucketize(uint(resp.GetResolutionMetadata().DatastoreQueryCount), s.requestDurationByQueryHistogramBuckets),
		utils.Bucketize(uint(resp.GetResolutionMetadata().DispatchCount), s.requestDurationByDispatchCountHistogramBuckets),
	).Observe(float64(time.Since(start).Milliseconds()))
```

?

since I assume that a) you need to specify a value for _all_ labels, and b) the label that you added `dispatch_count` comes last in the list? (is there a way to say more explicitly "this label has this value"??)

internal/server/config/config.go
```diff
@@ -232,7 +232,8 @@ type Config struct {
 	Metrics         MetricConfig
 	CheckQueryCache CheckQueryCache
 
-	RequestDurationDatastoreQueryCountBuckets []string
+	RequestDurationDatastoreQueryCountBuckets    []string
+	RequestDurationDatastoreDispatchCountBuckets []string
```
adriantam, line 236: datastore dispatch seems a bit misleading - as dispatch is focused on resolving subgraphs, not datastore lookup.

pkg/server/server.go
adriantam: Agree with @miparnisari - the way we add the dimension appears to be incorrect.

adriantam (commented):
Would it be possible to test out the changes by seeing how the requests duration measurement will look in Grafana?

cmd/run/run.go
poovamraj: I have updated the description but calling it "Check subproblem count" feels misleading as 
1. We will do this for ListObjects as well
2. We should have a common consistent term used everywhere and just dispatch count works for that.

Modified this a little bit

.config-schema.json
```diff
@@ -73,6 +73,17 @@
             "default": [50, 200],
             "x-env-variable": "OPENFGA_REQUEST_DURATION_DATASTORE_QUERY_COUNT_BUCKETS"
         },
+        "requestDurationDispatchCountBuckets": {
+            "description": "Datastore dispatch count buckets used to label the histogram metric for measuring request duration.",
```
adriantam, line 77:
```suggestion
            "description": "Dispatch count buckets used to label the histogram metric for measuring request duration.",
```

Dispatch is not related to datastore

pkg/server/server.go
```diff
@@ -343,6 +354,10 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		return nil, fmt.Errorf("request duration datastore count buckets must not be empty")
 	}
 
+	if len(s.requestDurationByDispatchCountHistogramBuckets) == 0 {
```
adriantam, line 357: Please add the corresponding test - similar to https://github.com/openfga/openfga/blob/5039e0d38af21e1179bc28b289f62bc9645150d9/pkg/server/server_test.go#L156

adriantam (changes requested):

pkg/server/server.go
```diff
@@ -76,17 +76,18 @@ var (
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method"})
 
+	//The name of the histogram is not referring to dispatch count, but we are not changing this as it will be a breaking change.
 	requestDurationByQueryHistogramName = "request_duration_by_query_count_ms"
 
-	requestDurationByQueryHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+	requestDurationByQueryAndDispatchHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace:                       build.ProjectName,
 		Name:                            requestDurationByQueryHistogramName,
```
jon-whit, line 84: The name of this metric has become too overloaded and is too specific and misleading. We should rename to `request_duration_ms` instead of `request_duration_by_query_count_ms`

poovamraj: I discussed about this with @adriantam since it is breaking change. Ended up changing only the internal names. Should we go ahead with this?

jon-whit: Let’s introduce it in a compatible way.

Release `v1.5.1`  we introduce `request_duration_ms` and announce deprecation notice of `request_duration_by_query_count_ms`.

* `request_duration_ms`
* `request_duration_by_query_count_ms`

These report the same values in release `v1.5.1`.

(wait 2-4 weeks)

Release `v1.5.X` we drop `request_duration_by_query_count_ms` in favor of `request_duration_ms`.

miparnisari (dismissed):

pkg/server/server.go
miparnisari: Wait where is this test? 😆

pkg/server/server.go
poovamraj: Great, I will do this in another PR

pkg/server/server.go
```diff
@@ -255,14 +257,22 @@ func WithCheckQueryCacheTTL(ttl time.Duration) OpenFGAServiceV1Option {
...
-// WithRequestDurationByQueryHistogramBuckets sets the buckets used in labelling the requestDurationByQueryHistogram
+// WithRequestDurationByQueryHistogramBuckets sets the buckets used in labelling the requestDurationByQueryAndDispatchHistogram
 func WithRequestDurationByQueryHistogramBuckets(buckets []uint) OpenFGAServiceV1Option {
 	return func(s *Server) {
 		sort.Slice(buckets, func(i, j int) bool { return buckets[i] < buckets[j] })
 		s.requestDurationByQueryHistogramBuckets = buckets
 	}
 }
 
+// WithRequestDurationByDispatchCountHistogramBuckets sets the buckets used in labelling the requestDurationByDispatchCountHistogramBuckets
```
adriantam, line 268:
```suggestion
// WithRequestDurationByDispatchCountHistogramBuckets sets the buckets used in labelling the requestDurationByQueryAndDispatchHistogram
```

pkg/server/server.go
poovamraj: I pushed it to wrong branch 😢 - `rename-query-count-to-datastore-query-count`. Just cherry picked it to this one

poovamraj: https://github.com/openfga/openfga/blob/4cbb08dfde9bdc9b3c7df1200e7b7b1a00b98a00/pkg/server/server_test.go#L168-L178

adriantam (approved):

PR #1448, opened 2024-03-13
test: ListUsers unit tests

willvedd:
## Description
This PR introduces a fairly comprehensive set of unit tests for the upcoming `ListUsers` API. These tests will aide in development and identify current functionality gaps and bugs in the current PoC. 

**Tested Scenarios**
- Direct Relationships
- Computed Relationships
- Usersets
- Tuple-to-usersets
- Wildcards
- Union, intersection, negation
- Cycle detection
- Contextual tuples
- Conditions

By no means is the above list going to be _every_ case we need for this feature but it serves as a solid foundation.

I introduce a temporary mechanism that allows tests to skip with the `TemporarySkipReason` property. As functionality gets added, we will incrementally remove these skips. There will be no skipped tests when #1428 is merged into main.

## References
- Parent PR - #1428 

## Review Checklist


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -216,7 +217,8 @@ func (l *listUsersQuery) expandDirect(
 		return err
 	}
 
-	iter, err := l.ds.Read(ctx, req.GetStoreId(), &openfgav1.TupleKey{
+	ds := storagewrappers.NewCombinedTupleReader(l.ds, req.GetContextualTuples().GetTupleKeys())
+	iter, err := ds.Read(ctx, req.GetStoreId(), &openfgav1.TupleKey{
```
willvedd, line 221: This was a quick change to support contextual tuples.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,272 +2,1091 @@ package listusers
...
 	"github.com/stretchr/testify/require"
-	"google.golang.org/protobuf/protoadapt"
-	"google.golang.org/protobuf/testing/protocmp"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func TestListUsers(t *testing.T) {
+type ListUsersTests []struct {
+	name                  string
+	TemporarilySkipReason string // Temporarily skip test until functionality is fixed
```
willvedd, line 20: This is a temporary way to skip tests and document why they're being skipped. As we address functionality gaps, we can repeal these skip statements.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,272 +2,1091 @@ package listusers
...
+			},
+			expectedUsers: []string{"group:*"},
+		},
 	}
+	tests.runListUsersTestCases(t)
+}
 
-	for _, test := range tests {
+func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
+	storeID := ulid.Make().String()
```
adriantam, line 1084: Please add the following to ensure there are no leaks
```
	t.Cleanup(func() {
		goleak.VerifyNone(t)
	})
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,272 +2,1091 @@ package listusers
...
+	model                 string
+	tuples                []*openfgav1.TupleKey
+	expectedUsers         []string
+	expectedError         error
+}
 
-	storeID := ulid.Make().String()
-	modelID := ulid.Make().String()
+func TestListUsersDirectRelationship(t *testing.T) {
+	model := `model
```
miparnisari, line 29:
```suggestion
func TestListUsersDirectRelationship(t *testing.T) {
	t.Cleanup(func() {
		goleak.VerifyNone(t)
	})
	model := `model
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -285,18 +1104,33 @@ func TestListUsers(t *testing.T) {
 		ctx := typesystem.ContextWithTypesystem(context.Background(), typesys)
 
 		t.Run(test.name, func(t *testing.T) {
```
miparnisari, line 1106: i would move all the code related to each test (lines 1091-1102) inside `t.Run`, otherwise if some of those lines throw an error it will be harder to say to which test they belong

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: Good call! I should look into adding a linting rule for all tests.

pkg/server/commands/listusers/list_users_rpc.go
jon-whit, line 221: note (non-blocking): @willvedd this needs to be a filtered read, otherwise the `FilteredTupleKeyIterator` below will be doing a lot of filtering in memory.

This doesn't need to be done in this PR, but before we close out this work we need a filtered read that only returns all possible related users/subjects that match the target subject types we need to expand. Otherwise we'll be filtering tons of shit out. This is particularly important because we can also have multi-model stores that leave residual tuples in the system that may have to get filtered out. We should strive as much as possible to avoid the in-memory filtering below this line of code.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+	}
+	tests.runListUsersTestCases(t)
+}
+
+func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
+	storeID := ulid.Make().String()
+
+	for _, test := range testCases {
 		ds := memory.New()
 		defer ds.Close()
```
jon-whit, line 1084: It seems more recently we're trying to encourage more usage of `t.Cleanup` instead of `defer` in test functions. Can we switch this to `t.Cleanup(ds.Close)`?

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+			test.req.StoreId = storeID
 
 			resp, err := l.ListUsers(ctx, test.req)
 			require.ErrorIs(t, err, test.expectedError)
 
-			ignoredFieldsOpts := protocmp.IgnoreFields(protoadapt.MessageV2Of(&openfgav1.Object{}), "type", "id")
+			actualUsers := resp.GetUsers()
+
+			actualCompare := make([]string, len(actualUsers))
+			for i, u := range resp.GetUsers() {
```
jon-whit, line 1116: Can we please make the conversion of `*openfgav1.User` to `string` a utility of the `testutils` package and delegate to it here?

```go
package testutils

func UserProtobufsToStrings(users *openfgav1.User...) []string {
   ...
}
```

then we have
```go
actual := testutils.UserProtobufsToStrings(resp.GetUsers()...)
require.ElementsMatch(t, actual, test.expectedUsers)
```

miparnisari: Agree on the helper, but i would put it in the `tuple` pkg because that has been the standard place for all converters, e.g. https://github.com/openfga/openfga/blob/7bb7fe40ab28610994cfdd75744edb0ae6523d8e/pkg/tuple/tuple.go#L151-L168

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
 
-	tests := []struct {
-		name            string
-		req             *openfgav1.ListUsersRequest
-		typeDefinitions []*openfgav1.TypeDefinition
-		tuples          []*openfgav1.TupleKey
-		expectedUsers   []*openfgav1.User
-		expectedError   error
-	}{
+	tests := ListUsersTests{
```
jon-whit, line 40: Please add 2 more test cases:

1. One where the filter's relation is inapplicable. e.g.

```
type user

type document
  relations
    define viewer: [user]
```
`ListUsers(document:1#viewer, user_filters=[{"type": "document", "relation": "undefined"}]`

2. Direct relationships with userset subjects and a userset filter. e.g.

```
type user

type group
  relations
```

- group:eng#member@group:fga#member
- group:fga#member@group:fga-backend#member

`ListUsers(group:eng#member, user_filters=[{"type": "group", "relation": "member"}]` should return ["group:eng#member", "group:fga#member", "group:fga-backend#member"]

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
 			},
-			expectedUsers: []*openfgav1.User{
-				{
-					User: &openfgav1.User_Object{
-						Object: &openfgav1.Object{Type: "user", Id: "jon"},
+			expectedUsers: []string{"user:will", "user:maria"},
+		},
+		{
+			name:                  "userset_group_granularity",
+			TemporarilySkipReason: "because `group:eng` is being returned instead of `group:fga#member`",
```
jon-whit, line 263: Did you mean "because `group:eng` is being returned instead of `group:eng#member`?

I don't see `group:fga` referenced anywhere in this specific test case.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
 			  relations
-			    define viewer: [group#member] as self
-			`),
+			    define viewer: [group#member]`,
 			tuples: []*openfgav1.TupleKey{
-				tuple.NewTupleKey("document:1", "viewer", "group:eng#member"),
 				tuple.NewTupleKey("group:eng", "member", "user:hawker"),
+				tuple.NewTupleKey("group:fga", "member", "user:jon"),
 				tuple.NewTupleKey("group:eng", "member", "group:fga#member"),
+				tuple.NewTupleKey("document:1", "viewer", "group:eng#member"),
```
jon-whit, line 361: Just as a sanity check in addition to the nested group, can you add one more group at the `document:1#viewer` level?

e.g.
* document:1#viewer@group:eng#member
* document:1#viewer@group:other#member
* group:eng#member@group:fga#member
* group:eng#member@user:hawker
* group:fga#member@user:jon
* group:other#member@user:will

expected [`user:jon`, `user:hawker`, `user:will`]

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+			model: `model
+            schema 1.1
+			type user
+			type group
+			  relations
+			    define member: [user, group#member]
+			type document
+			  relations
+			    define viewer: [group#member]`,
+			tuples: []*openfgav1.TupleKey{
```
jon-whit, line 387: Similar comment to that above, can we also add one more at the `document:1#viewer` level just for good measure?

* document:1#viewer@group:eng#member
* document:1#viewer@group:other#member
* ...

expected: [`group:eng#member`, `group:fga#member`, `group:other#member`]

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+						Type: "user",
+					},
+				},
+			},
+			model: `model
+			schema 1.1
+			type user
+			type group
+				relations
+					define member: [user]
```
jon-whit, line 1035: This extra relation definitions is extraneous and misleading in this test. Let's just drop it in favor of:

```
type user
type group

type document
  relations
    define viewer: [group:*, user:*]
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+					define viewer: [ group:*, user:*]`,
+
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "viewer", "user:*"),
+				tuple.NewTupleKey("document:1", "viewer", "group:*"),
+			},
+			expectedUsers: []string{"user:*"},
+		},
+		{
+			name: "multiple_possible_wildcards_group_granularity",
```
jon-whit, line 1048: Same test as the one directly above it. I vote we drop this since it's a duplicate.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+			type document
+				relations
+					define viewer: [ group:*, user:*]`,
+
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "viewer", "user:*"),
+				tuple.NewTupleKey("document:1", "viewer", "group:*"),
+			},
+			expectedUsers: []string{"group:*"},
+		},
```
jon-whit, line 1074: Can we add a test that follows an indirect relationship that leads to wildcard?

```
type user

type group
  relations
    define member: [user:*]

type document
  relations
    define viewer: [group#member]
```
- group:eng#member@user:*

ListUsers(document:1#viewer, user_filters=[{"type":"user"}]) should return ["user:*"]

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+				tuple.NewTupleKey("folder:x", "editor", "user:will"),
+				tuple.NewTupleKey("folder:x", "owner", "user:jon"),
+				tuple.NewTupleKey("folder:x", "unrelated_not_computed", "user:poovam"),
+
+				tuple.NewTupleKey("folder:no-doc", "viewer", "user:maria"),
+				tuple.NewTupleKey("document:1", "parent", "folder:no-user"),
+			},
+			expectedUsers: []string{"user:maria", "user:will", "user:jon"},
+		},
+	}
```
jon-whit, line 565: Can we also add one more test that establishes at least a hierarchy of depth 2. e.g.

```
type user

type folder
  relations
    define parent: [folder]
    define viewer: [user] or viewer from parent
```

- folder:b#parent@folder:1a
- folder:c#parent#folder:b
- folder:c#parent@folder:1b
- folder:1a#viewer@user:will
- folder:1b#viewer@user:jon

ListUsers(folder:c#viewer, user_filters=[{"type": "user"}]) should return ["user:will", "user:jon"]

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+
+	tests.runListUsersTestCases(t)
+}
+
+func TestListUsersCycles(t *testing.T) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+	tests := ListUsersTests{
```
jon-whit, line 574: It would also be good to add one more that establishes a cycle through a TTU hierarchy. e.g.

```
type user

type folder
  relations
    define viewer: [user, document#viewer]

type document
  relations
    define parent: [folder]
```

- document:1#parent@folder:x
- folder:x#viewer@document:1#viewer

ListUsers(document:1#viewer, user_filters=[{"type": "user"}])

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -338,6 +341,7 @@ func (l *listUsersQuery) expandTTU(
 				Object:               &openfgav1.Object{Type: userObjectType, Id: userObjectID},
 				Relation:             computedRelation,
 				UserFilters:          req.GetUserFilters(),
+				ContextualTuples:     req.GetContextualTuples(),
```
jon-whit, line 344: If we're going to add the tests for ABAC Conditions in the `_test.go` file at this time, then we should also do our diligence to wire up the `Context` here and everywhere else in this file.

Otherwise we should post pone the support for Conditions and the tests for them until a subsequent PR.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1132 @@ package listusers
...
+		},
+	}
+	tests.runListUsersTestCases(t)
+}
+
+func TestListUsersUnion(t *testing.T) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+	tests := ListUsersTests{
```
jon-whit, line 826: Just to cover the bases, let's add one more variant where we test all 3 base rewrite cases in the same union. e.g.

```
type user

type folder
  relations
    define viewer: [user]

type document
  relations
    define editor: [user]
    define viewer: [user] or editor or viewer from parent
```

The tests we have in here cover TTU and computed joined by union, but not also the directly assignable relationship included in the union.

pkg/server/commands/listusers/list_users_rpc.go
willvedd: Good call. Won't do here but created a ticket to capture that task.

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: Yep. Just a typo.

pkg/server/commands/listusers/list_users_rpc.go
willvedd: We have a ticket specific to conditions work so would prefer that be done then. The tests in this PR were a first pass to confirm that yes, conditions don't work. We can always add more tests when ^ work is done.

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: 100% agree with a dedicated conversion utility function. Are we ok with that being a quick follow-up PR? This PR is already quite large and the aforementioned work will also itself require tests.

miparnisari (approved):

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -60,7 +61,7 @@ func (l *listUsersQuery) ListUsers(
 	foundUsersCh := make(chan *openfgav1.User, 1)
 	expandErrCh := make(chan error, 1)
 
-	var foundUsers []*openfgav1.User
+	foundUsers := make([]*openfgav1.User, 0)
```
adriantam, line 64: curious why you don't use   
```
var foundUsers []*openfgav1.User
```
instead

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -216,7 +217,8 @@ func (l *listUsersQuery) expandDirect(
 		return err
 	}
 
-	iter, err := l.ds.Read(ctx, req.GetStoreId(), &openfgav1.TupleKey{
+	ds := storagewrappers.NewCombinedTupleReader(l.ds, req.GetContextualTuples().GetTupleKeys())
```
adriantam, line 220: suggestion: (non-blocking) Looks like there is a lot of duplicate code between expandDirect and expandTTU.  Anyway we can make some common function to avoid code duplication?

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1245 @@ package listusers
...
-	"google.golang.org/protobuf/protoadapt"
-	"google.golang.org/protobuf/testing/protocmp"
+	"go.uber.org/goleak"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func TestListUsers(t *testing.T) {
+type ListUsersTests []struct {
+	name                  string
+	TemporarilySkipReason string // Temporarily skip test until functionality is fixed
```
adriantam, line 21: why is `TemporarilySkipReason` exported?  (vs. temporarilySkipReason)

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1245 @@ package listusers
...
+            schema 1.1
+			type user
+			type document
+			  relations
+			    define viewer: [user]
+			`,
+			tuples:        []*openfgav1.TupleKey{},
+			expectedUsers: []string{"document:1#viewer"},
+		},
+	}
```
adriantam, line 491: will be an interesting case if you have the case where a user belongs to multiple group and multiple groups as viewer access to the document

For example]
```
model
	schema 1.1
type user
type group
	relations
		define member: [user]
type document
	relations
		define viewer: [ user, group#member ]			
```

with tuples
```
tuple.NewTupleKey("group:eng", "member", "user:will"),
tuple.NewTupleKey("group:eng", "member", "user:maria"),
tuple.NewTupleKey("group:core", "member", "user:will"),
tuple.NewTupleKey("document:1", "viewer", "group:core#member"),
```
You should see `user:will` and `user:maria` appear once in the return.  (try to prevent `user:will` from appearing twice)

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2,301 +2,1245 @@ package listusers
...
+			type document
+				relations
+					define viewer: [group#member]`,
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "viewer", "group:eng#member"),
+				tuple.NewTupleKey("group:eng", "member", "user:*"),
+			},
+			expectedUsers: []string{"user:*"},
+		},
+	}
```
adriantam, line 1188: should we have wildcard test for computed wildcard?   TTU?

pkg/server/commands/listusers/list_users_rpc.go
willvedd: Because don't want to return nil if no users found.

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: To stand-out against the other, more permanent struct properties. This will be removed before merging into main so just want to make abundantly sure we don't miss.

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: Good idea. Just added!

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: Lack of deduplication can be observed in a few of the other test cases but this has standalone value. Just added.

PR #1451, opened 2024-03-15
listusers: avoid stack overflows

miparnisari:
## Description

Avoid stack overflows caused by loops in ListUsers API.

## References

### DIFF
--- a/pkg/server/commands/listusers/list_users.go
+++ b/pkg/server/commands/listusers/list_users.go
@@ -84,8 +84,8 @@ func fromListUsersRequest(o listUsersRequest) *internalListUsersRequest {
 }
 
 // clone creates a copy of the request. Note that some fields are not deep-cloned.
-func clone(o *internalListUsersRequest) *internalListUsersRequest {
-	v := fromListUsersRequest(o)
-	v.visitedUsersetsMap = maps.Clone(o.visitedUsersetsMap)
+func (r *internalListUsersRequest) clone() *internalListUsersRequest {
+	v := fromListUsersRequest(r)
+	v.visitedUsersetsMap = maps.Clone(r.visitedUsersetsMap)
 	return v
 }
--- a/pkg/server/commands/listusers/list_users.go
+++ b/pkg/server/commands/listusers/list_users.go
@@ -16,13 +16,7 @@ type listUsersRequest interface {
 }
 
 type internalListUsersRequest struct {
-	StoreID              string
-	AuthorizationModelID string
-	Object               *openfgav1.Object
-	Relation             string
-	UserFilters          []*openfgav1.ListUsersFilter
-
-	ContextualTuples *openfgav1.ContextualTupleKeys
+	*openfgav1.ListUsersRequest
 
 	// visitedUsersetsMap keeps track of the "path" we've made so far.
 	// It prevents stack overflows by preventing visiting the same userset twice.
@@ -36,7 +30,7 @@ func (r *internalListUsersRequest) GetStoreId() string {
 	if r == nil {
 		return ""
 	}
-	return r.StoreID
+	return r.StoreId
 }
 
 //nolint:stylecheck
@@ -44,7 +38,7 @@ func (r *internalListUsersRequest) GetAuthorizationModelId() string {
 	if r == nil {
 		return ""
 	}
-	return r.AuthorizationModelID
+	return r.AuthorizationModelId
 }
 
 func (r *internalListUsersRequest) GetObject() *openfgav1.Object {
@@ -77,24 +71,21 @@ func (r *internalListUsersRequest) GetContextualTuples() *openfgav1.ContextualTu
 
 func from(o listUsersRequest) *internalListUsersRequest {
 	return &internalListUsersRequest{
-		StoreID:              o.GetStoreId(),
-		AuthorizationModelID: o.GetAuthorizationModelId(),
-		Object:               o.GetObject(),
-		Relation:             o.GetRelation(),
-		UserFilters:          o.GetUserFilters(),
-		ContextualTuples:     o.GetContextualTuples(),
-		visitedUsersetsMap:   make(map[string]struct{}),
+		ListUsersRequest: &openfgav1.ListUsersRequest{
+			StoreId:              o.GetStoreId(),
+			AuthorizationModelId: o.GetAuthorizationModelId(),
+			Object:               o.GetObject(),
+			Relation:             o.GetRelation(),
+			UserFilters:          o.GetUserFilters(),
+			ContextualTuples:     o.GetContextualTuples(),
+		},
+		visitedUsersetsMap: make(map[string]struct{}),
 	}
 }
 
 // clone creates a copy of the request. Note that some fields are not deep-cloned.
 func clone(o *internalListUsersRequest) *internalListUsersRequest {
-	return &internalListUsersRequest{
-		StoreID:            o.GetStoreId(),
-		Object:             o.GetObject(),
-		Relation:           o.GetRelation(),
-		UserFilters:        o.GetUserFilters(),
-		ContextualTuples:   o.GetContextualTuples(),
-		visitedUsersetsMap: maps.Clone(o.visitedUsersetsMap),
-	}
+	v := from(o)
+	v.visitedUsersetsMap = maps.Clone(o.visitedUsersetsMap)
+	return v
 }
--- a/pkg/server/commands/listusers/list_users.go
+++ b/pkg/server/commands/listusers/list_users.go
@@ -25,7 +25,7 @@ type internalListUsersRequest struct {
 
 var _ listUsersRequest = (*internalListUsersRequest)(nil)
 
-//nolint:stylecheck
+//nolint:stylecheck // it should be GetStoreID, but we want to satisfy the interface listUsersRequest
 func (r *internalListUsersRequest) GetStoreId() string {
 	if r == nil {
 		return ""
@@ -33,7 +33,7 @@ func (r *internalListUsersRequest) GetStoreId() string {
 	return r.StoreId
 }
 
-//nolint:stylecheck
+//nolint:stylecheck // it should be GetAuthorizationModelID, but we want to satisfy the interface listUsersRequest
 func (r *internalListUsersRequest) GetAuthorizationModelId() string {
 	if r == nil {
 		return ""
--- a/pkg/server/commands/listusers/list_users.go
+++ b/pkg/server/commands/listusers/list_users.go
@@ -1,6 +1,8 @@
 package listusers
 
 import (
+	"maps"
+
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/server/commands/listusers/list_users.go
```diff
@@ -2,6 +2,7 @@ package listusers
 
 import (
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"golang.org/x/exp/maps"
```
adriantam, line 5: why do we use `golang.org/x/exp/maps` vs. `maps`?  [Clone](https://pkg.go.dev/maps#Clone) is available as in standard go map

adriantam (changes requested):

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -595,6 +595,59 @@ func TestListUsersCycles(t *testing.T) {
 			},
 			expectedUsers: []string{},
 		},
+		{
```
willvedd, line 598: **FYI:** I updated the upstream unit tests branch to include one more test cyclical test case.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -135,10 +135,20 @@ func (l *listUsersQuery) ListUsers(
 // 	return nil
 // }
 
-func (l *listUsersQuery) expand(
+func (l *listUsersQuery) beginExpand(
```
willvedd, line 138: Can we consolidate to a single recursive function? This keeps it consistent with Check and doesn't introduce another function. Can just initialize `VisitedUsersets` if empty.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -195,8 +203,9 @@ func (l *listUsersQuery) expandRewrite(
 
 		children := rewrite.Union.GetChild()
 		for _, childRewrite := range children {
+			childRewrite1 := childRewrite
```
willvedd, line 206:
```suggestion
			childRewriteCopy := childRewrite
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -195,8 +203,9 @@ func (l *listUsersQuery) expandRewrite(
...
 			pool.Go(func(ctx context.Context) error {
-				return l.expandRewrite(ctx, req, childRewrite, foundUsersChan)
+				return l.expandRewrite(ctx, req, childRewrite1, foundUsersChan)
```
willvedd, line 208:
```suggestion
				return l.expandRewrite(ctx, req, childRewriteCopy, foundUsersChan)
```

pkg/server/commands/listusers/list_users.go
```diff
@@ -12,3 +14,87 @@ type listUsersRequest interface {
...
+	AuthorizationModelID string
+	Object               *openfgav1.Object
+	Relation             string
+	UserFilters          []*openfgav1.ListUsersFilter
+
+	ContextualTuples *openfgav1.ContextualTupleKeys
+
+	// visitedUsersetsMap keeps track of the "path" we've made so far.
+	// It prevents stack overflows by preventing visiting the same userset twice.
+	visitedUsersetsMap map[string]struct{}
```
willvedd, line 29: **Question:** Am I to understand that cycles are _only_ caused by usersets? And if so, can you explain why?

pkg/server/commands/listusers/list_users.go
```diff
@@ -12,3 +14,87 @@ type listUsersRequest interface {
 	GetUserFilters() []*openfgav1.ListUsersFilter
 	GetContextualTuples() *openfgav1.ContextualTupleKeys
 }
+
+type internalListUsersRequest struct {
```
willvedd, line 18: **Thinking out loud:** This mostly-redundant struct seems clunky but I'm wondering maybe if that's because we haven't wired this up to the server yet. Once we do that, we'll have abstracted the server layer away from the underlying algorithm, at which point we can rename this to something more elegant. 

WDYT?

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -595,6 +595,59 @@ func TestListUsersCycles(t *testing.T) {
...
+			type transition
+				relations
+					define start: [state]
+					define end: [state]
+					define can_view: can_view from start or can_view from end
+					define can_view_2: can_view
+					define can_view_3: can_view_2`,
+			tuples:        []*openfgav1.TupleKey{},
+			expectedUsers: []string{},
+		},
```
willvedd, line 650: **Request:** Can we also add a unit test for `expand` that focuses purely on the `visitedUsersets` aspect?

willvedd (changes requested):

pkg/server/commands/listusers/list_users.go
miparnisari: i don't think it's clunky, i agree it's not pretty, but we need to keep track of internal state somewhere. Since we cannot extend protobuf objects, i created my own. I could also do

```go
type internalListUsersRequest struct {
     openfgav1.ListUsersRequest

     visitedUsersetsMap map[string]struct{}

}
```

pkg/server/commands/listusers/list_users.go
miparnisari: No, usersets don't _cause_ the cycles, they are merely the way of detecting if we've encountered one.

Some cycles are caused by the model, doesn't matter which tuples you have (in red below), e.g.

```
define a: b
define b: a
```

If the request is ListUsers(document:1 , a, filter = [user]), we start expanding `document:1#a`, so we mark it as visited, then we go to `document:1#b`, we mark it as visited, then we will go back to `document:1#a`, loop!


image

The other category of cycles are caused by the model _and_ the tuples in the store (in yellow above).

If the request is ListUsers(document:1 , viewer, filter = [user]), we start expanding `document:1#viewer`, so we mark it as visited, then by virtue of the computed userset we go to `document:1#editor`, we mark it as visited, then assume we have a tuple `document:1#viewer@document:1#editor`, we will go back to `document:1#viewer`, loop!

pkg/server/commands/listusers/list_users_rpc_test.go
miparnisari: sorry i don't follow, a unit test that tests which aspect?

pkg/server/commands/listusers/list_users.go
jon-whit: If this were implemented as an RPC, what would we do? Assume the ListUsersRequest is an internal RPC definition and would otherwise be dispatchable over any interface.

In this case, we'd "extend" the base request. Namely,

```
type DispatchListUsersRequest struct {
    openfgav1.ListUsersRequest
    ... // other metadata applicable to the internal dispatch
}
```

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: A unit test for the `expand` function that detects cycles solely through the visited paths attribute. Similar to what we did in https://github.com/openfga/openfga/blob/343adc7926436630e19dc8c32cc021a46af275fe/internal/graph/cycle_check_resolver_test.go#L29

pkg/server/commands/listusers/list_users.go
willvedd: I'm getting confused by usage of usersets in `visitedUsersetsMap`. The first example doesn't have user sets. Can we keep consistent with check's `VisitedPaths`?

miparnisari: > The first example doesn't have user sets

Yes and no... there are no tuples, true, but with the query being `document:1 , a, filter = [user]`, I need to expand the userset `document:1#a`.

> Can we keep consistent with check's VisitedPaths?

No, Check uses full tuples as key of the map, but I don't have that here. I am using the same key as in ListObjects: 

https://github.com/openfga/openfga/blob/0efd7244c43197369337ce288d6af6b6245f9fe5/pkg/server/commands/reverseexpand/reverse_expand.go#L271-L275

pkg/server/commands/listusers/list_users.go
miparnisari: I ended up doing ☝️

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -335,16 +327,21 @@ func (l *listUsersQuery) expandTTU(
...
+			newReq.Object = &openfgav1.Object{Type: userObjectType, Id: userObjectID}
+			newReq.Relation = computedRelation
+			return l.expand(ctx, newReq, foundUsersChan)
 		})
 	}
 
 	return pool.Wait()
 }
+
+func enteredLoop(req *internalListUsersRequest) bool {
```
adriantam, line 340: nitpick: perhaps `visitedPath` is more descriptive than enteredLoop?

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1305,3 +1333,67 @@ func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
...
+				UserFilters: []*openfgav1.ListUsersFilter{{
+					Type: "user",
+				}},
+			},
+			visitedUsersetsMap: visitedUsersets,
+		}, channelWithResults)
+
+		require.NoError(t, err)
+		select {
+		case <-channelWithResults:
```
adriantam, line 1393: question: do we need go func to wrap around the expand call?  It feels weird to see us listening on a channel but the originator is not a go routine.

pkg/server/commands/listusers/list_users.go
```diff
@@ -12,3 +14,78 @@ type listUsersRequest interface {
...
+
+	// visitedUsersetsMap keeps track of the "path" we've made so far.
+	// It prevents stack overflows by preventing visiting the same userset twice.
+	visitedUsersetsMap map[string]struct{}
+}
+
+var _ listUsersRequest = (*internalListUsersRequest)(nil)
+
+//nolint:stylecheck
+func (r *internalListUsersRequest) GetStoreId() string {
```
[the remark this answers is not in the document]
miparnisari: because the linter complains that it should be `GetStoreID` :( 

image

however, the interface `listUsersRequest` is meant to be satisfied by the proto objects, which has the auto-generated `GetStoreId`

image

pkg/server/commands/listusers/list_users_rpc.go
miparnisari: i feel the opposite, actually 🤔

pkg/server/commands/listusers/list_users_rpc_test.go
miparnisari: good point!

pkg/server/commands/listusers/list_users.go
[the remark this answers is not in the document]
adriantam: perhaps include a comment to explain in the code.

pkg/server/commands/listusers/list_users_rpc.go
willvedd: Only other thing here is "loop" vs "cycle". We only use loop to describe model cycles and even then they're seemingly synonymous. Can we land on `enteredCycle`?

adriantam (approved):

pkg/server/commands/listusers/list_users.go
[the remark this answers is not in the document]
jon-whit: For context here - the interface it has to satisfy is purposeful. Protobuf generated getters for id fields produce `GetId()` not `GetID()`. So that is the reason for the siganture.

pkg/server/commands/listusers/list_users.go
```diff
@@ -12,3 +14,78 @@ type listUsersRequest interface {
...
+}
+
+func (r *internalListUsersRequest) GetContextualTuples() *openfgav1.ContextualTupleKeys {
+	if r == nil {
+		return nil
+	}
+	return r.ContextualTuples
+}
+
+func fromListUsersRequest(o listUsersRequest) *internalListUsersRequest {
```
jon-whit, line 72: It could be convienent to define this as a Clone on the InternalListUsersRequest directly. 

e.g.
```
var req *openfgav1.ListUsersRequest

var internalReq *InternalListUsersRequest
internalReq.Clone(req)
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -691,19 +693,43 @@ func TestListUsersCycles(t *testing.T) {
 			model: `model
 			schema 1.1
 			type user
-
-			type folder
-			relations
-				define viewer: [user, document#viewer]
-
 			type document
```
jon-whit, line 696: How does this test pass? The model is invalid in the first place.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1264,7 +1290,10 @@ func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
 			}
 
 			typesys, err := typesystem.NewAndValidate(context.Background(), model)
-			require.NoError(t, err)
+			if err != nil {
+				t.Log("warn! model is invalid")
```
jon-whit, line 1294: What's the rationale here? It feels odd to have expectations of system behavior on invalid models. If the model is invalid and we're not detecting, then that's the responsibility of other code. It shouldn't be this test and the component it is testing's responsibility to ensure correct behavior given invalid constraints.

pkg/server/commands/listusers/list_users_rpc_test.go
jon-whit: Oh I see, we added a warn and still run it.

See https://github.com/openfga/openfga/pull/1451#discussion_r1532383675

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: AFAICT, this is to run tests against old models that exhibited cycles, prior to the appropriate model validation. My assumption is that `typesystem.NewAndValidate` should enforce these new rules but we still need to be able to test older models too.

jon-whit: 1.0 models are no longer. We wouldn't evaluate Check or ListUsers on old models as it is no longer a supported schema version.

miparnisari: Apologies for the confusion. I updated the PR with a comment explaining why I did this. Let me know what you think 🙇‍♀️

miparnisari: As per offline discussion: removed test for the model with no entrypoints.

jon-whit (approved):

PR #1452, opened 2024-03-15
fix(check): count of datastore reads

miparnisari:
## Description

While reviewing https://github.com/openfga/openfga/pull/1440 locally I found that we are not counting datastore reads properly on Check.

On main, these tests are broken:
- `userset_no_access_1` returns 0 DB reads instead of 1
- `wildcard_no_access` returns 0 DB reads instead of 1
- `multiple_userset_no_access` returns 4 DB reads instead of 3


internal/graph/check.go
```diff
@@ -728,7 +728,7 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 							ContextualTuples:     req.GetContextualTuples(),
 							ResolutionMetadata: &ResolutionMetadata{
 								Depth:               req.GetResolutionMetadata().Depth - 1,
-								DatastoreQueryCount: response.GetResolutionMetadata().DatastoreQueryCount,
+								DatastoreQueryCount: req.GetResolutionMetadata().DatastoreQueryCount, // add userset read at the end
```
miparnisari, line 731: we use the same pattern when resolving TTUs: https://github.com/openfga/openfga/blob/d7f77478e8024f3f299c66d7279f9a0ccd84ba28/internal/graph/check.go#L920-L923

internal/graph/check.go
```diff
@@ -768,7 +768,13 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		resp, err := union(ctx, c.concurrencyLimit, checkFuncs...)
 		if err != nil {
 			telemetry.TraceError(span, err)
+			return nil, err
 		}
+		if len(directlyRelatedUsersetTypes) > 0 {
+			// if we had N userset checks, that was 1 read, not N
+			resp.GetResolutionMetadata().DatastoreQueryCount++
```
adriantam, line 775: Feels a bit weird to add it here rather than closer to the source we call the datastore query

miparnisari: I agree, but i don't know how to make it better. This is counting + 1 in the case where we didn't find a path that led to `allowed=true`.

internal/graph/check_test.go
jon-whit, line 1: @miparnisari could you point out which tests which were added specifically failed prior to this code? Or did all of these added tests fail?

internal/graph/check_test.go
```diff
@@ -731,6 +741,41 @@ type document
...
+			allowed:    false,
+			minDBReads: 3, // 1 userset read (2 found) follow by 2 direct tuple check (not found)
+			maxDBReads: 3,
+		},
+		{
+			name:       "multiple_userset_access",
+			check:      tuple.NewTupleKey("document:x", "multiple_userset", "user:maria"),
+			allowed:    true,
+			minDBReads: 2, // 1 userset read (2 found) follow by 1 direct tuple check (found, returns immediately)
+			maxDBReads: 2,
```
jon-whit, line 779: This should be 3. In the worst case the read of the second userset branches occurs first. There's a race on the userset evaluations.

The resolved branch will require 2 database queries, but the max database reads is 3.

miparnisari: we may execute 3 reads overall but we will only return 2 because we return the count in the path where we found `allowed=true`, which is not going to be aware of the read happening in the other goroutine

internal/graph/check_test.go
miparnisari: I wrote in the PR description which tests fail on `main` 👍

internal/graph/check_test.go
jon-whit: Right - we agree. But this test assumes the max db reads is the upper bound of how many total may get executed. That's my point.

internal/graph/check_test.go
jon-whit: Oh, gotcha. Sorry I overlooked that and just jumped right into the file changes. Thanks.

internal/graph/check_test.go
miparnisari: no, this test asserts on the response field's datastore count, which will never be 3. I'd get test failures otherwise

internal/graph/check.go
```diff
@@ -768,8 +768,19 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
 		}
-		return resp, err
+
+		// count db reads after they happen in the case that we didn't find 'allowed=false' but we still incurred reads
+		if len(directlyRelatedUsersetTypes) > 0 {
+			// if we had N userset checks, that was 1 read, not N
+			resp.GetResolutionMetadata().DatastoreQueryCount++
+		}
+		if shouldCheckDirectTuple {
+			resp.GetResolutionMetadata().DatastoreQueryCount++
```
adriantam, line 780: Similar to the comment above, this should be part of responsibility of fn1 to increment the datastore query count

internal/graph/check.go
jon-whit: I think it would help composition and readability here to keep track of parent and child aggregates independently, so we don't stomp on each other. A single parent spawns multiple child userset evals, but they are overlapping on a shared resolution metadata field.

What we probably should do is the following:
```go
  ParentRequest *ResolveCheckRequest
  ...
  ResolutionMetadata ResolveCheckRequestMetadata
}

type ResolveCheckRequestMetadata struct {
   DatastoreQueryCount int
}

func ... checkDirect(..., req *ResolveCheckRequest) {
    fn2 := ... {
      iter, err := ds.ReadUsersetTuples(...)
      if err != nil {
        // handle error
      }

      req.GetResolutionMetadata().DatastoreQueryCount += 1

      for {
        iter = iter.Next()
        ...

        if usersetRelation != "" {
          ...

          handlers = append(handlers, c.dispatch(
            ctx,
            &ResolveCheckRequest{
              ParentRequest: req,
              ...
                Depth:               req.GetResolutionMetadata().Depth - 1,
                DatastoreQueryCount: 0, // start at 0 for the subproblem, we'll aggregate as we query
              },
              ...
            }))
        }
      }

      resp, err := union(ctx, c.concurrencyLimit, handlers...)
      if err != nil {
        ...
      }

      // req is the parent, remember
      // add the aggregate sum of the children datastore query counts, which are aggreated
      // in the response.
      req.GetResolutionMetadata().DatastoreQueryCount += resp.GetResolutionMetadata().DatastoreQueryCount
    }
}
```

internal/graph/check.go
miparnisari: undid this change

internal/graph/check.go
miparnisari: > The datastore query count increment should be part of fn2 (only when fn2 is invoked - as in the case of line 765), the datastore query count is incremented.

we can't do that because the outcome is incorrect.

- `union(allowed=true, allowed=false)` should return the result of the reads for `allowed=true` branch only, which we _can_ count inside `fn2`
- however, `union(allowed=false, allowed=false)` should return the sum of _all_ reads that we had to do. if we add +1 to each individual branch, we will be over-counting

internal/graph/check_test.go
jon-whit: What's the purpose of the `maxDBReads` anymore then? In this example it's possible at most 3 DB reads will occur - 1 for the initial userset read, and 2 direct queries for each of the usersets found. If we're only ever returning the "resolved" DB query count, then what's the purpose of defining the `maxDBReads`? The number of DB queries required to resolve is different than the total number of DB queries invoked per request, so that's also not clear to me here.

Could we clarify that?

jon-whit (approved):

internal/graph/check.go
miparnisari: Tracking original comment in https://github.com/openfga/openfga/issues/1477

PR #1453, opened 2024-03-15
refactor: distinguish Check request metadata and response metadata

miparnisari:
## Description

This PR is a proposal to lay the groundwork for https://github.com/openfga/openfga/pull/1440.

Changes:
1. `ResolveCheckRequest` and `ResolveCheckResponse` both, currently, have an inner `Metadata` struct. This is confusing to me because, for example, `Depth` is only used in the `Request`. In this PR I am separating it:

```go
	RequestMetadata      *ResolveCheckRequestMetadata
}
	Allowed            bool
	ResolutionMetadata *ResolveCheckResponseMetadata
}
```
2. Introduce field `DispatchCounter` in `CheckRequestMetadata`. This field is not written, its value is always 0 (that is where #1440 comes in)
3. Simplify callers of `LocalChecker.dispatch`. Now, `dispatch` is the one that updates all the metadata accordingly.


jon-whit:
@miparnisari @poovamraj ya'll are overlapping work with this right now. Make sure you sync up. @poovamraj is working on uniform dispatch, which impacts this work as well.

internal/graph/graph.go
```diff
@@ -41,13 +42,32 @@ func ResolutionDepthFromContext(ctx context.Context) (uint32, bool) {
 	return depth, ok
 }
 
-type ResolutionMetadata struct {
+type CheckRequestMetadata struct {
```
jon-whit, line 45:
```suggestion
type ResolveCheckRequestMetadata struct {
```

This to distinguish between the parent Check RPC. We should strive to stay consistent with the internal dispatch RPC signature, which is `ResovleCheck`.

internal/graph/graph.go
```diff
@@ -41,13 +42,32 @@ func ResolutionDepthFromContext(ctx context.Context) (uint32, bool) {
...
+
+func NewCheckRequestMetadata(maxDepth uint32) *CheckRequestMetadata {
+	return &CheckRequestMetadata{
+		Depth:               maxDepth,
+		DatastoreQueryCount: 0,
+		DispatchCounter:     new(atomic.Uint32),
+	}
+}
+
+type CheckResponseMetadata struct {
```
jon-whit, line 69:
```suggestion
type ResolveCheckResponseMetadata struct {
```

internal/graph/check.go
```diff
@@ -36,13 +36,29 @@ type ResolveCheckRequest struct {
 	TupleKey             *openfgav1.TupleKey
 	ContextualTuples     []*openfgav1.TupleKey
 	Context              *structpb.Struct
-	ResolutionMetadata   *ResolutionMetadata
+	RequestMetadata      *CheckRequestMetadata
```
jon-whit, line 39:
```suggestion
	Metadata      *CheckRequestMetadata
```

RequestMetadata stutters - we can simplify to just Metadata since it is attached to the request to begin with.

internal/graph/check.go
```diff
@@ -495,11 +511,15 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
...
 
-// dispatch dispatches the provided Check request to the CheckResolver this LocalChecker
-// was constructed with.
-func (c *LocalChecker) dispatch(_ context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
+// dispatch clones the parent request, modifies its metadata and tupleKey, and dispatches the new request
+// to the CheckResolver this LocalChecker was constructed with.
+func (c *LocalChecker) dispatch(_ context.Context, parentReq *ResolveCheckRequest, tk *openfgav1.TupleKey) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
-		resp, err := c.delegate.ResolveCheck(ctx, req)
+		newReq := clone(parentReq)
```
jon-whit, line 518:
```suggestion
		childReq := clone(parentReq)
```

internal/graph/check.go
miparnisari: I don't want to do this change now because it will blow up this PR with too many lines changed

jon-whit: Ok, let's follow up semi-promptly at least.

miparnisari: Sounds good!

jon-whit (approved):

PR #1433, opened 2024-03-08
ListUsers API

willvedd:
## Description
Introducing the `ListUsers` API which answers the question "what users are related to a specific object?". To learn more about this feature, refer to the [ListUsers RFC](https://github.com/openfga/rfcs/pull/15).

A ListUsers request is composed of three main properties:

- `user_filter` – the type of user (subject) targeted for the response. This can either be a terminal type (e.g. `user`) or a userset (e.g. `group#member`). Only one user filter can be provided at this time.
- `relation` – the target relation that subjects need to the specified object
- `object` – the target resource that subjects need to be related to

**Example:**
Consider this model and tuples:
```
model
  schema 1.1

type user

type group
  relations
    define member: [user]

type document
  relations
    define viewer: [user, group#member]
    
#   document:1, viewer, user:maria
#   document:1, viewer, user:will
#   document:2, viewer, user:jon
#   group:A, member, user:christy
#   document:1, viewer, group:A#member
```
If we want to return all subjects of type `user` that have the `viewer` relation on `document:1` we would call 
`ListUsers(user_filter=user, viewer, document:1)` which would return the following:
```
{
    "users": [ "user:maria", "user:will", "user:christy" ]
    "excluded_users": []
}
```

**Note:** This API will initially be released in an experimental capacity. It can be enabled by setting `OPENFGA_EXPERIMENTALS=enable-list-users`

## References
- [ListUsers API Docs](https://openfga.dev/api/service#/Relationship%20Queries/ListUsers)
- [ListUsers RFC](https://github.com/openfga/rfcs/pull/15)
- [ListUsers Roadmap Item](https://github.com/openfga/roadmap/issues/16)
- [API Protobuf Definitions](https://github.com/openfga/api/pull/140)
- Related Issues:
   - https://github.com/openfga/openfga/issues/1215
   - https://github.com/openfga/openfga/issues/406

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #1464, opened 2024-03-18
`ListUsers` edge pruning

willvedd:
## Description
Introducing the process of "edge pruning" to `ListUsers`. I evaluates the model, `user_filters` and the target document type to understand if a given query can has resolvable "edges" on the graph. And if not, the query will exit early. This is not only an optimization but ensures correctness. It only needed to happen once up-front for each request; it does not need to be done on a recursive basis.

## References
- Parent PR: #1433 

## Review Checklist


pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1282,7 +1414,12 @@ func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
 			test.req.StoreId = storeID
 
 			resp, err := l.ListUsers(ctx, test.req)
-			require.ErrorIs(t, err, test.expectedError)
+
+			actualErrorMsg := ""
+			if err != nil {
+				actualErrorMsg = err.Error()
+			}
+			require.Equal(t, test.expectedErrorMsg, actualErrorMsg)
```
willvedd, line 1422: I had issue wrapping the error here. Really we just want to assert a specific message, not that literal error itself is equal.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1250,6 +1249,139 @@ func TestListUsersWildcards(t *testing.T) {
...
+		  type document
+			relations			  
+			  define viewer: [user]`,
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "viewer", "user:maria"),
+			},
+			expectedUsers: []string{"user:maria"},
+		},
+		{
+			name: "impossible_edge_several_computed_relations_away",
```
miparnisari, line 1278: 🤔 why does this say "impossible edge"?

willvedd: Because there is no edge between those two nodes, it is impossible.

miparnisari: there is no edge connecting `user` and `document#viewer`, but if you look at the entire model, there is a set of edges connecting them. i would remove `impossible_edge` from the test name because to me it implies that there is no connectivity possible, but it's not the case here.

willvedd: Ah. This is a mistake, the edge is indeed possible. Updated the test name.

miparnisari (approved):

PR #1471, opened 2024-03-20
`ListUsers` integration tests for intersection

miparnisari:
## Description
Add assertions for ListUsers API in models that contain `intersection`.

## References

### DIFF
--- a/assets/tests/consolidated_1_1_tests.yaml
+++ b/assets/tests/consolidated_1_1_tests.yaml
@@ -297,6 +297,14 @@ tests:
               type: document
               relation: viewer
             expectation:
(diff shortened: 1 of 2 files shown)


willvedd (changes requested):
These look good. My only request is that we add cases for `relation_with_wildcard_involving_intersection`? That's a pretty important one that's top-of-mind currently.

miparnisari:
> My only request is that we add cases for relation_with_wildcard_involving_intersection? That's a pretty important one that's top-of-mind currently.

Do you mean add assertions? I pushed another one, but i can't think of any others 🤔

willvedd (approved):

PR #1473, opened 2024-03-20
`ListUsers` exclusion

willvedd:
## Description
Adding support for exclusion rewrites in `ListUsers` via the `but not` operator.

To achieve this, we need to expand both parts of the rewrite, the base and subtract. .The base is the left-portion of the `but not` operator that must evaluate true. The subtract is the right-portion that much evaluate false. First, all the subjects that evaluate true to the base are iterated through and cross-checked against the subtract subjects. If subjects that are only present in the base and not the subtract are returned on the channel.

## References
 - Sibling PR (intersection): #1467 
 - Parent PR: #1428 

## Review Checklist


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -370,6 +372,48 @@ func (l *listUsersQuery) expandIntersection(
...
+		if _, isSubtracted := subtractFoundUsersCountMap[key]; !isSubtracted {
+			// Iterate over base users because at minimum they need to pass
+			// but then they are further compared to the subtracted users map.
+			// If users exist in both maps, they are excluded. Only users that exist
+			// solely in the base map will be returned.
+			foundUsersChan <- tuple.StringToUserProto(key)
+		}
+	}
+
+	return <-errChan
```
miparnisari, line 414: This will only return the last error.

i think we should do

```go
var errs *multierror.Error
	go func() {
		err1 := l.expandRewrite(ctx, req, rewrite.Difference.GetBase(), baseFoundUsersCh)
                 if err1 != nil {
			multierror.Append(errs, err1)
	        }
		close(baseFoundUsersCh)

		err2 := l.expandRewrite(ctx, req, rewrite.Difference.GetSubtract(), subtractFoundUsersCh)
                if err2 != nil {
			multierror.Append(errs, err2)
	        }
		close(subtractFoundUsersCh)
	}()

...
return errs
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1033,8 +1105,7 @@ func TestListUsersExclusion(t *testing.T) {
 	})
 	tests := ListUsersTests{
```
jon-whit, line 1106: These tests look great, nice work! Can we add one or two more though that tests block lists through self-referntial cases, e.g.

```
model
  schema 1.1

type user

type group
  relations
    define member: [user, group#member] but not other
    define other: [user, group#member] or member
```

**Test Case 1**
- group:1#member@user:jon

`ListUsers(group:1#member, user_filters=[{"type": "user"}])` --> [] (should always return empty, no error)

**Test Case 2**
- group:2#member@group:1#member
- group:1#member@user:will

`ListUsers(group:2#member, user_filters=[{"type": "user"}])` --> [] (should always return empty, no error)

**Test Case 3**
```
model
  schema 1.1

type user

type group
  relations
    define member: [user, group#member] but not other
    define other: [user, group#member]
```
- group:1#other@group:1#member
- group:1#member@user:will

`ListUsers(group:1#member, user_filters=[{"type": "user"}])` --> [] (should always return empty, no error)

willvedd: Added case 3. Case 1 and 2 produced cycles.

jon-whit (approved):

miparnisari (approved):

PR #1479, opened 2024-03-21
fix: correct default dispatch throttle enabled

adriantam:
## Description

Initialization of server dispatch throttle enabled should use default dispatch throttle enabled

## Review Checklist


pkg/server/server.go
```diff
@@ -363,7 +363,7 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		requestDurationByDispatchCountHistogramBuckets: []uint{50, 200},
 		serviceName: openfgav1.OpenFGAService_ServiceDesc.ServiceName,
 
-		dispatchThrottlingCheckResolverEnabled:   serverconfig.DefaultCheckQueryCacheEnable,
+		dispatchThrottlingCheckResolverEnabled:   serverconfig.DefaultDispatchThrottlingEnabled,
```
jpadilla, line 366: uff good catch.

jpadilla (approved):

pkg/server/server.go
miparnisari: how did we miss this? should we update the test `dispatch_throttling_check_resolver_enabled`?

adriantam: Default values are constant and they have the same value (false).  I don't know if we can override constants in the UT.

miparnisari: ah, i see. yeah not sure how to test for that..

miparnisari (approved):

PR #1458, opened 2024-03-17
CI: Add workflow to update NOTICE file

Siddhant-K-code:
## Description

This pull request updates the NOTICE file based on the FOSSA analysis. The FOSSA CLI is used to analyze the project and generate an updated NOTICE file. The changes include including unused dependencies and generating attribution information. For more details, please refer to the FOSSA analysis report at [FOSSA][1].

[1]: https://app.fossa.com/projects/git%2Bgithub.com%2Fopenfga%2Fopenfga

## Action items

- It requires `FOSSA_API_KEY` API token be to added and add them as secrets to be consumed by GitHub Action. (ref [docs](https://docs.fossa.com/docs/api-reference))

It requires a FOSSA subscription (see [ref](https://github.com/Siddhant-K-code/openfga/actions/runs/8316135416/job/22755235252#step:4:190))

image

## References

fixes #1319

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

Issue #1319, opened 2024-01-25
Automated process to update NOTICE file


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

We want to keep the NOTICE file up to date on every dependency change.

### Describe the ideal solution

Update [NOTICE](https://github.com/openfga/openfga/blob/main/NOTICE) file using some kind of automated process (Github actions?)

The file is in https://app.fossa.com/projects/git%2Bgithub.com%2Fopenfga%2Fopenfga

Siddhant-K-code:
@miparnisari For the first iteration, we can go with this:

We can schedule a cron job to update `NOTICE` file using GitHub Actions (requires FOSSA API key) using their CLI to get the latest report and update it with the latest report w/ `NOTICE` file.

Currently, In `NOTICE` file, it seems it also has Custom Copyrights at top. I think, the person who is creating FOSSA API token or configured `OpenFGA` project on FOSSA needs to enable below left option (Copyrights section -- requires premium)

<img width="648" alt="image" src="https://github.com/openfga/openfga/assets/55068936/a98384f9-ebb6-4298-a7fa-915f1cc311c2">



Wdyt? 💭 

_// cc: @jon-whit_

Siddhant-K-code:
> I created an issue directly with FOSSA: https://github.com/fossas/fossa-action/issues/38

Until, then we can go the manual way (using CLI in GitHub action) ?

miparnisari:
@Siddhant-K-code I think so 👍

Siddhant-K-code:
Sure, I will take a look on it.

Siddhant-K-code:
I created a rough workflow [here](https://github.com/openfga/openfga/compare/main...Siddhant-K-code:openfga:main). Challenge is that, how we wanted to push the changed NOTICE file, it could be a direct commit (with custom branch protection rules to force push to `main`) Or, GitHub Action bot should raise the PR with the updated NOTICE file (when we run the workflow or weekly). Wdyt?

miparnisari:
@Siddhant-K-code thanks for that! We think a weekly PR is good 👍  it should only open if the NOTICE files has had changes.

Siddhant-K-code:
> it should only open if the NOTICE file has had changes.

Not sure if we can detect this 🤔

[...]

PR #1480, opened 2024-03-21
Revert "CI: Add workflow to update NOTICE file"

willvedd:
Reverts openfga/openfga#1458 because it didn't succeed, even when a valid API token is applied. 

# References

- Failing Job: https://github.com/openfga/openfga/actions/runs/8379924182
- Original PR: #1458 
- Related Issue: #1319


miparnisari:
Let's follow-up with FOSSA support and once they give a clear path forward we can resurrect this 👍

PR #1485, opened 2024-03-22
`ListUsers` intersection with wildcards

willvedd:
## Description
The initial implementation of intersection (#1467) in `ListUsers` missed an important case during testing, public typed wildcard tuples. Wildcards require special consideration when it comes to intersection. 

The initial implementation of intersection relied on a counter per user object, and if that count equaled the number of `and` clauses, then the user would be returned on the main results channel. However the presence of wildcards in a clause needs to be considered; you do not want to double-count users as shown in this example below:

```
model
  schema 1.1
type user
type document
  relations
    define public_1: [user:*,user]
    define public_2: [user:*,user]
    define is_public: public_1 and public_2
```
```
"document:1", "public_1", "user:maria"
"document:1", "public_1", "user:*",
```
```
ListUsers("document:1", user_filter: ["user"], relation: "is_public") // should return empty list
```

The solution provided in this PR is to individually evaluate a separate channel per clause to be able to consider the wildcard case.

## References
- Related PR: #1467 
- Related comment: https://github.com/openfga/openfga/pull/1467#discussion_r1533089380
- Parent PR: #1428 

## Review Checklist

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -350,7 +350,7 @@ func (l *listUsersQuery) expandIntersection(
 	wg.Add(len(childOperands))
 
 	wildcardCount := atomic.Uint32{}
-	wildcardKey := fmt.Sprintf("%s:*", req.GetUserFilters()[0].GetType())
+	wildcardKey := tuple.TypedPublicWildcard(req.GetUserFilters()[0].GetType())
 	foundUsersCountMap := make(map[string]uint32, 0)
 	for _, foundUsersChan := range intersectionFoundUsersChans {
 		go func(foundUsersChan chan *openfgav1.User) {
@@ -430,7 +430,7 @@ func (l *listUsersQuery) expandExclusion(
 		subtractFoundUsersMap[key] = struct{}{}
 	}
 
-	wildcardKey := fmt.Sprintf("%s:*", req.GetUserFilters()[0].GetType())
+	wildcardKey := tuple.TypedPublicWildcard(req.GetUserFilters()[0].GetType())
 	_, subtractWildcardExists := subtractFoundUsersMap[wildcardKey]
 	for key := range baseFoundUsersMap {
 		if _, isSubtracted := subtractFoundUsersMap[key]; !isSubtracted && !subtractWildcardExists {


pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1447,6 +1447,243 @@ func TestListUsersEdgePruning(t *testing.T) {
...
+
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "required_1", "user:maria"),
+				tuple.NewTupleKey("document:1", "required_1", "user:jon"),
+
+				tuple.NewTupleKey("document:1", "required_1", "user:will"),
+				tuple.NewTupleKey("document:1", "required_2", "user:will"),
+
+				tuple.NewTupleKey("document:1", "required_1", "user:*"),
+				tuple.NewTupleKey("document:1", "required_2", "user:*"),
```
miparnisari, line 1651: these two tuples aren't allowed by the model you defined

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1447,6 +1447,243 @@ func TestListUsersEdgePruning(t *testing.T) {
...
+		  type user
+		  type document
+		    relations
+		      define required_1: [user]
+		      define required_2: [user]
+		      define can_view: required_1 and required_2`,
+
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "required_1", "user:*"),
+				tuple.NewTupleKey("document:1", "required_2", "user:*"),
```
miparnisari, line 1589: these two tuples aren't allowed by the model you defined

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1447,6 +1447,243 @@ func TestListUsersEdgePruning(t *testing.T) {
...
+					},
+				},
+			},
+			model: `model
+            schema 1.1
+          type user
+          type document
+            relations
+              define public_1: [user:*,user]
+			  define public_2: [user:*,user]
```
miparnisari, line 1499: nit: indent

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: That's exactly what I'm trying to test. Or is it safe to assume that this is disallowed by the datastore?

miparnisari: The datastore doesn't check for correctness, the API has to do that. So you have make sure that the API filters out invalid tuples with respect to the model being used. There are two approaches for this:

1. Use a filtered tuple iterator `filteredIter := storage.NewFilteredTupleKeyIterator(storage.NewTupleKeyIteratorFromTupleIterator(tupleIter), validation.FilterInvalidTuples(typesys))`
2. Use the graph of the model to pre-filter what tuples to fetch (like we do in ReverseExpand)

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -324,38 +318,70 @@ func (l *listUsersQuery) expandIntersection(
...
+		// If this summed value equals the number of clauses, the user satisfies
 		// the intersection expression and can be sent on `foundUsersChan`
-		if c == uint32(len(children)) {
-			foundUsersChan <- tuple.StringToUserProto(key)
+		if (count + wildcardCount.Load()) < uint32(len(clauses)) {
+			continue
 		}
+
+		foundUsersChan <- tuple.StringToUserProto(key)
 	}
```
miparnisari, line 385: I'm not a huge fan of this approach. It probably works but it looks complex, and you'll have to repeat it for exclusion. 

I feel like we're making our lives harder by not re-using the ReverseExpansion approach of using the graph traversal 🤔

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: We currently employ strategy 1. This test is ensuring that it works correctly. I have a strong preference to maintain this test case because it is easily forgotten and can wreak havoc on our system.

pkg/server/commands/listusers/list_users_rpc_test.go
willvedd: Discussed in similar comment below.

pkg/server/commands/listusers/list_users_rpc.go
willvedd: > I'm not a huge fan of this approach. It probably works but it looks complex, and you'll have to repeat it for exclusion.

I'm not in love either but it works and it makes sense why it works, although this is hugely subjective. 

Fortunately we don't need to repeat for exclusion. The complexity stems from supporting any number of clauses whereas exclusion always deal with two clauses.

> I feel like we're making our lives harder by not re-using the ReverseExpansion approach of using the graph traversal 

I still intend on having us evaluate wether reverse expand and/or some other algorithm is more appropriate, but for now I'm mostly concerned on getting an initial feature-complete version implemented.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
 	pool.WithCancelOnError()
 	pool.WithMaxGoroutines(int(l.resolveNodeBreadthLimit))
 
-	intersectionFoundUsersChan := make(chan *openfgav1.User, 1)
+	wildcardKey := fmt.Sprintf("%s:*", req.GetUserFilters()[0].GetType())
 
-	children := rewrite.Intersection.GetChild()
-	for _, childRewrite := range children {
-		copyChildRewrite := childRewrite
+	clauses := rewrite.Intersection.GetChild()
```
jon-whit, line 326:
```suggestion
	childOperands := rewrite.Intersection.GetChild()
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
...
+	var wg sync.WaitGroup
+	wg.Add(len(clauses))
+
+	wildcardCount := atomic.Uint32{}
 	foundUsersCountMap := make(map[string]uint32, 0)
-	for fu := range intersectionFoundUsersChan {
-		key := tuple.UserProtoToString(fu)
-		foundUsersCountMap[key]++
+	for _, foundUsersChan := range intersectionFoundUsersChans {
+		go func(channel chan *openfgav1.User) {
```
jon-whit, line 355:
```suggestion
		go func(foundUsersChan chan *openfgav1.User) {
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
...
+			foundUsersMap := make(map[string]uint32, 0)
+			for fu := range channel {
```
jon-whit, line 357:
```suggestion
			for foundUser := range foundUsersChan {
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
 	pool.WithCancelOnError()
 	pool.WithMaxGoroutines(int(l.resolveNodeBreadthLimit))
 
-	intersectionFoundUsersChan := make(chan *openfgav1.User, 1)
+	wildcardKey := fmt.Sprintf("%s:*", req.GetUserFilters()[0].GetType())
```
jon-whit, line 324: This should be a helper in the `pkg/tuple` package.

```go
package tuple
...
func TypedPublicWildcard(objectType string) string {
    return fmt.Sprintf("%s:%s", objectType, Wildcard)
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
...
+			}
+			for userKey := range foundUsersMap {
+				mu.Lock()
+				foundUsersCountMap[userKey]++
+				if wildcardExists {
+					foundUsersCountMap[userKey]--
+				}
+				mu.Unlock()
+			}
+			wg.Done()
```
jon-whit, line 374: Please `defer wg.Done()` at the top of this function.

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: I don’t have any strong opposition to the algorithm at this time. It seems reasonable for the current API invariants. It would get much uglier if we supported more than 1 user filter in the near future. That would be my main concern with the approach, because that multiply the number of typed wildcards and user objects we'd have to track independently as well.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1695,6 +1695,243 @@ func TestListUsersEdgePruning(t *testing.T) {
...
+		      define can_view: required_1 and required_2`,
+
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "required_1", "user:maria"),
+				tuple.NewTupleKey("document:1", "required_1", "user:jon"),
+				tuple.NewTupleKey("document:1", "required_1", "user:*"),
+				tuple.NewTupleKey("document:1", "required_2", "user:will"),
+			},
+			expectedUsers: []string{"user:will"},
+		},
```
jon-whit, line 1930: Could we add another test that walks some userset expansion stages with wildcards?

```
model
  schema 1.1

type user

type group
  relations
    define member: [user:*, user]

type document
  relations
    define group: [group]
    define viewer: [group#member] and member from group
    define can_view: viewer
```
- document:1#viewer@group:y#member
- document:1#group@group:x
- group:x#member@user:*
- group:y#member@user:jon

`ListUsers(document:1#can_view, user_filters=[{"type": "user"}])` should return ["user:jon"]

pkg/server/commands/listusers/list_users_rpc_test.go
miparnisari: Ok, then can you please leave a comment next to each "invalid" tuple ?

pkg/server/commands/listusers/list_users_rpc_test.go
miparnisari: Please leave a comment next to each invalid tuple.

pkg/server/commands/listusers/list_users_rpc.go
miparnisari, line 324: nit: please move this closer to where it's used (line 360)

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
...
+		}
 		errChan <- err
 		close(errChan)
 	}()
 
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+	wg.Add(len(clauses))
+
+	wildcardCount := atomic.Uint32{}
```
miparnisari, line 352:
```suggestion
	wildcardCount := atomic.Uint32{} // guarded by mu
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
...
 	foundUsersCountMap := make(map[string]uint32, 0)
```
miparnisari, line 353:
```suggestion
	foundUsersCountMap := make(map[string]uint32, 0) // GUARDED_BY(mu).
```

similar to https://github.com/openfga/openfga/blob/1fd71eb98377280f96708d1eaf23f1e66de13a96/pkg/storage/memory/memory.go#L110

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -327,39 +321,71 @@ func (l *listUsersQuery) expandIntersection(
...
+		// all intersection clauses plus the number of wildcards.
+		// If this summed value equals the number of clauses, the user satisfies
 		// the intersection expression and can be sent on `foundUsersChan`
-		if c == uint32(len(children)) {
-			foundUsersChan <- tuple.StringToUserProto(key)
+		if (count + wildcardCount.Load()) < uint32(len(clauses)) {
+			continue
 		}
+
+		foundUsersChan <- tuple.StringToUserProto(key)
```
miparnisari, line 388:
```
                 // Compare the number of times the specific user was returned for
		// all intersection clauses plus the number of all wildcards found.
		// If this summed value >= the number of clauses, the user satisfies
		// the intersection expression and can be sent on `foundUsersChan`
		if (count + wildcardCount.Load()) >= uint32(len(clauses)) {
			foundUsersChan <- tuple.StringToUserProto(key)
		}
```

miparnisari, line 388:
```suggestion
		if (count + wildcardCount.Load()) >= uint32(len(clauses)) {
			foundUsersChan <- tuple.StringToUserProto(key)
		}		
```

pkg/server/commands/listusers/list_users_rpc.go
miparnisari: I also don't have a strong opposition if it works, so i'm not blocking the PR :) 

 I'm just concerned about the fact that we have two very different implementations of two APIs that do relatively similar things, with all the downsides that that implies (maintainability, cognitive overload etc)

pkg/server/commands/listusers/list_users_rpc.go
willvedd: Addressed in https://github.com/openfga/openfga/pull/1490. Will pull-in that work once merged to main.

jon-whit (approved):

PR #1491, opened 2024-03-27
fix: listusers skip usersets for intersection and exclusion

miparnisari:
## Description
Address two issues:
- ListUsers queries should return usersets as well as concrete users
- When the ListUsers query filter is a clause in an `and` or `but not`, it cannot be returned as result

TO DO: some Check responses are inconsistent with the ListUsers responses, e.g.

image

## References
https://github.com/openfga/openfga/pull/1433


miparnisari:
Superseded by https://github.com/openfga/openfga/pull/1500

PR #1500, opened 2024-04-01
fix:`ListUsers` include correct relation when returning usersets

willvedd:
## Description
This PR fixes an issue in `ListUsers` where usersets weren't returning the relation in the response. Ex: (`group:fga` instead of `group:fga#member`). 

This bug occurred because we weren't checking the relation in the conditional when evaluating a userset _and_ because we weren't returning the correct object type. 

## References
Parent PR: #1428 

## Review Checklist

### DIFF
--- a/assets/tests/consolidated_1_1_tests.yaml
+++ b/assets/tests/consolidated_1_1_tests.yaml
@@ -852,7 +852,33 @@ tests:
               object: document:2
               relation: viewer
             expectation:
-
+          - request:
+              filters:
+                - folder#editor
+              object: document:1
+              relation: viewer
+            expectation:
+             - folder:X#editor
+          - request:
+              filters:
+                - folder#writer
+              object: document:1
+              relation: viewer
+            expectation:
+              - folder:X#writer
+          - request:
+              filters:
+                - folder#viewer
+              object: document:1
+              relation: viewer
+            expectation:
+              - folder:X#viewer
+          - request:
+              filters:
+                - folder
+              object: document:1
+              relation: viewer
+            expectation:
   - name: tuple_to_userset_and_intersection
     stages:
       - model: |
@@ -939,6 +965,13 @@ tests:
               object: document:1
               relation: viewer
             expectation:
+          - request:
+              filters:
+                - folder#viewer
+              object: document:1
+              relation: viewer
+            expectation:
+              - folder:X#viewer
           - request:
               filters:
                 - folder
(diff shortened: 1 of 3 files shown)


assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -1032,7 +1032,7 @@ tests:
               object: document:1
```
jon-whit, line 1032: Should we also add one for `filters: folder#viewer` and `document:1#viewer`?

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -2011,7 +2011,7 @@ tests:
                 - folder#viewer
               object: document:1
               relation: viewer
-            temporarilySkipReason: "because we need to remember if previous recursive calls had intersection or exclusion"
+            temporarilySkipReason: "because reflexive relationships not supported yet"
             expectation:
```
jon-whit, line 2015: Shouldn't this be `folder:x#viewer`?

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -179,11 +179,15 @@ func (l *listUsersQuery) expand(
 	if enteredCycle(req) {
 		return nil
 	}
-	for _, f := range req.GetUserFilters() {
-		if req.GetObject().GetType() == f.GetType() {
+	for _, userFilter := range req.GetUserFilters() {
```
jon-whit, line 182: Does it make sense to define some variables above for these intermediate values so we don't have to do `req.Get...` everywhere? For example,

```
objectType := req.GetObject().GetType()
objectID := req.GetObject().GetId()
relation := req.GetRelation()
...
```

willvedd: Yes, makes sense. Competing against another `relation` var lower down so maybe not those literal names.

jon-whit (approved):

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -6481,7 +6505,6 @@ tests:
                 - group#member
               object: document:1
               relation: can_view
-            temporarilySkipReason: "because usersets not returned properly yet, returns `group:fga` instead"
             expectation:
               - group:fga#member # TODO is this OK? not all users of group:fga#member have the can_view relation
```
miparnisari, line 6509: can you please remove the TODO? :)

miparnisari (approved):

PR #1502, opened 2024-04-02
feat: List objects dispatch throttling

poovamraj:
Implement Dispatch Throttling for ListObjects

## Description
- We have abstracted dispatch throttling into it's own `DispatchThrottler`
- CheckResolver will use this DispatchThrottler with a delegate and resolver
- For testing pupose we have introduced `ReleaseDispatch`. We have to introduce a external dependency _like_ (https://github.com/benbjohnson/clock)
- ListObjects uses dispatch throttler using the `Throttle` method
- We have a single dispatch throttler which means a single queue and single configuration will be used

TODO:
Add tests for throttling with ListObjects

## Review Checklist

### DIFF
--- a/internal/graph/dispatch_throttling_check_resolver_test.go
+++ b/internal/graph/dispatch_throttling_check_resolver_test.go
@@ -2,6 +2,7 @@ package graph
 
 import (
 	"context"
+	"github.com/openfga/openfga/internal/throttler"
 	"sync"
 	"testing"
 	"time"
@@ -19,12 +20,13 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
-		dispatchThrottlingCheckResolverConfig := DispatchThrottlingCheckResolverConfig{
+		dispatchThrottlingConfig := throttler.DispatchThrottlingConfig{
 			// We set timer ticker to 1 hour to avoid it interfering with test
 			Frequency: 1 * time.Hour,
 			Threshold: 200,
 		}
-		dut := NewDispatchThrottlingCheckResolver(dispatchThrottlingCheckResolverConfig)
+		dispatchThrottler := throttler.NewDispatchThrottlingCheckResolver(dispatchThrottlingConfig)
+		dut := NewDispatchThrottlingCheckResolver(dispatchThrottler)
 		defer dut.Close()
 
 		initialMockResolver := NewMockCheckResolver(ctrl)
@@ -54,12 +56,13 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		ctrl := gomock.NewController(t)
 		defer ctrl.Finish()
 
-		dispatchThrottlingCheckResolverConfig := DispatchThrottlingCheckResolverConfig{
+		dispatchThrottlingConfig := throttler.DispatchThrottlingConfig{
 			// We set timer ticker to 1 hour to avoid it interfering with test
-			Frequency: 1 * time.Hour,
+			Frequency: 1 * time.Second,
 			Threshold: 200,
 		}
-		dut := NewDispatchThrottlingCheckResolver(dispatchThrottlingCheckResolverConfig)
+		dispatchThrottler := throttler.NewDispatchThrottlingCheckResolver(dispatchThrottlingConfig)
+		dut := NewDispatchThrottlingCheckResolver(dispatchThrottler)
 		defer dut.Close()
 
 		initialMockResolver := NewMockCheckResolver(ctrl)
@@ -99,7 +102,7 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		require.Equal(t, 0, resolveCheckDispatchedCounter)
 
 		// simulate tick happening and we release a dispatch
-		dut.nonBlockingSend(dut.throttlingQueue)
+		dispatchThrottler.ReleaseDispatch()
 		goFuncDone.Wait()
 		require.Equal(t, 1, resolveCheckDispatchedCounter)
 	})
--- a/internal/graph/dispatch_throttling_check_resolver_test.go
+++ b/internal/graph/dispatch_throttling_check_resolver_test.go
@@ -1,8 +1,7 @@
-package throttler
+package graph
 
 import (
 	"context"
-	"github.com/openfga/openfga/internal/graph"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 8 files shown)


internal/throttler/dispatch_throttler.go
```diff
@@ -0,0 +1,103 @@
...
+			r.nonBlockingSend(r.throttlingQueue)
+		}
+	}
+}
+
+func (r *DispatchThrottler) Close() {
+	r.done <- struct{}{}
+}
+
+func (r *DispatchThrottler) Throttle(ctx context.Context, currentNumDispatch uint32) {
```
jon-whit, line 90: Can we add some godoc here please?

> Throttle provides a synchronous blocking mechanism that will block if the currentNumDispatch exceeds the configured dispatch threshold. It will block until a value is produced on the underlying throttling queue channel, which is produced by periodically sending a value on the channel based on the configured ticker frequency.

internal/throttler/dispatch_throttler.go
```diff
@@ -0,0 +1,103 @@
...
+// in the throttling queue. One item form the throttling queue will be processed ticker.
+// This allows a check / list objects request to be gradually throttled.
+type DispatchThrottler struct {
+	config          DispatchThrottlingConfig
+	ticker          *time.Ticker
+	throttlingQueue chan struct{}
+	done            chan struct{}
+}
+
+func NewDispatchThrottler(
```
jon-whit, line 34:
```suggestion
// NewDispatchThrottler constructs a DispatchThrottler that can be used to control the rate of dispatch in FGA queries.
```

internal/throttler/dispatch_throttler.go
```diff
@@ -0,0 +1,103 @@
...
+// immediately. When the number of request dispatches is above the Threshold, the dispatches are placed
+// in the throttling queue. One item form the throttling queue will be processed ticker.
+// This allows a check / list objects request to be gradually throttled.
+type DispatchThrottler struct {
+	config          DispatchThrottlingConfig
+	ticker          *time.Ticker
+	throttlingQueue chan struct{}
+	done            chan struct{}
+}
+
```
jon-whit, line 33:
```suggestion
// NewDispatchThrottler constructs a DispatchThrottler which can be used to control the rate of dispatched subproblems in FGA queries.
```

Please add this Godoc, and in general let's strive for godoc on any public signature of a package (even `internal` public signatures).

pkg/server/server.go
```diff
@@ -141,6 +142,7 @@ type Server struct {
 	dispatchThrottlingCheckResolverFrequency time.Duration
 	dispatchThrottlingThreshold              uint32
 
+	dispatchThrottler               *throttler.DispatchThrottler
```
jon-whit, line 145: Had this comment 1:1, but reposting here for broader visibility.

My only reservation right now with these changes is that ListObjects and Check will share a dispatch throttler, and so that throttling queue is going to have a lot contention potentially. I’m not sure if that’s the right approach, and I’d like to understand better if they should share or have independent throttlers. My hunch is independent throttlers (at least to start).

Can we analyze and better understand this before we consolidate them into a shared throttler?

internal/throttler/dispatch_throttler.go
jon-whit, line 1: Please put dispatch related packages under a package `internal/dispatch`. So, in this case, `internal/dispatch/throttler`. 

In the future we'll have:
* `internal/dispatch/dispatch.go` (for top-level Dispatcher interfaces such as `Dispatcher`, `CheckDispatcher`, `ReverseExpandDispatcher`, etc..).
* `internal/dispatch/check` (for Check dispatcher implementation)
* `internal/dispatch/reverseexpand` (for ReverseExpand dispatcher implementation)
* `internal/dispatch/remote` (for remote dispatch implementations)
* etc..

pkg/server/server.go
```diff
@@ -393,7 +395,8 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 			zap.Uint32("Threshold", s.dispatchThrottlingThreshold),
 		)
 
-		dispatchThrottlingCheckResolver := graph.NewDispatchThrottlingCheckResolver(dispatchThrottlingConfig)
+		s.dispatchThrottler = throttler.NewDispatchThrottler(dispatchThrottlingConfig)
```
miparnisari, line 398: Nit: by saving this object as a server field, if you accidentally call `s.dispatchThrottler.Close()` (there is nothing stopping you from doing so) in `server.Close()`, you get a panic. I'm not sure why.

internal/throttler/dispatch_throttler.go
miparnisari, line 1: Why no unit tests for this file?

internal/graph/dispatch_throttling_check_resolver_test.go
miparnisari, line 1: The core of the behavior is now in `dispatch_throttler`, which means that this unit test is really testing that file, not the `DispatchThrottlingCheckResolver`. Please rename this file to `dispatch_throttler_test.go` and create a separate unit test for `DispatchThrottlingCheckResolver` that asserts that it delegates correctly.

miparnisari (changes requested):

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -129,6 +131,12 @@ func WithResolveNodeLimit(limit uint32) ReverseExpandQueryOption {
 	}
 }
 
+func WithDispatchThrottler(dispatchThrottler *throttler.DispatchThrottler) ReverseExpandQueryOption {
+	return func(d *ReverseExpandQuery) {
+		d.dispatchThrottler = dispatchThrottler
```
miparnisari, line 136: Please create a `NoOpDispatchThrottler` and set is as the default in the constructor, then you can get rid of the nil checks in lines 345 and 553

pkg/server/server.go
```diff
@@ -393,7 +395,8 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+		dispatchThrottlingCheckResolver := graph.NewDispatchThrottlingCheckResolver(s.dispatchThrottler)
```
miparnisari, line 399: Won't block the PR for this, but i'd like to point out because it concerns me. The number of fields that we've introduced in the server is getting quite large. So is the server constructor, which for me is very hard to read.

```go
dispatchThrottler               *throttler.DispatchThrottler
dispatchThrottlingCheckResolver *graph.DispatchThrottlingCheckResolver
cachedCheckResolver    *graph.CachedCheckResolver
checkResolver graph.CheckResolver
typesystemResolver     typesystem.TypesystemResolverFunc
````

In reality, the concrete objects that need to be accessible by the APIs are:
- the model cache (almost every API needs this)
- the check cache (Check and ListObjects)
- with this PR, the throttling queue(s) (Check, ListObjects, and maybe more APIs in the future)

If y'all agree that this is something that we can improve, can you please open a ticket? Then we can talk about how to make it better.

File: internal/throttler/dispatch_throttler.go
Comment (adriantam) [reply internal/throttler/dispatch_throttler.go:1]: Actually, I would propose to make this throttler more generic as it could be used for both dispatch throttling and database throttling. Perhaps name it as throttler instead of dispatch throttler.

File: internal/throttler/dispatch_throttler.go [hunk 8]
```diff
@@ -0,0 +1,103 @@
...
+	select {
+	case signalChan <- struct{}{}:
+		// message sent
+	default:
+		// message dropped
+	}
+}
+
+// ReleaseDispatch is used for releasing a dispatch from the queue.
+func (r *DispatchThrottler) ReleaseDispatch() {
```
adriantam, line 68: Any reason this is used a public function?  If we need this for running unit test, perhaps create a test throttler that inherits this object + expose ReleaseDispatch() for test purpose?

internal/throttler/dispatch_throttler.go
```diff
@@ -0,0 +1,103 @@
...
+	"time"
+)
+
+// DispatchThrottlingConfig encapsulates configuration for dispatch throttling check resolver
+type DispatchThrottlingConfig struct {
+	Frequency time.Duration
+	Threshold uint32
+}
+
+type Ticker interface {
```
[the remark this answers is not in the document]
adriantam: Perhaps you actually want to have a Throttler interface with 
type ThrottlerInterface interface {
  Close()
  Throttle(context.Context, uint32) 
}

internal/throttler/dispatch_throttler.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+}
+
+func (r *DispatchThrottler) Close() {
+	r.done <- struct{}{}
+}
+
+func (r *DispatchThrottler) Throttle(ctx context.Context, currentNumDispatch uint32) {
+	if currentNumDispatch > r.config.Threshold {
```
adriantam, line 91: IMO, it probably is better to have the decision whether to throttle moved out to the caller.  In this way, the throttler is only responsible for throttling (single point of concern).  

Furthermore, it will allow the caller to store info on whether request has been throttled.  See https://github.com/openfga/openfga/pull/1552/files

pkg/server/server.go
poovamraj: Removed this param altogether

PR #780, opened 2023-05-30
release: add homebrew release job

chenrui333:
add homebrew release job to automate the release bump task.

TODO:
- need a committer token
- need a homebrew-core fork for openfga org

cc @aaguiarz

PR #1513, opened 2024-04-04
revert #780 (homebrew bump after every release)

miparnisari:
#780 is not needed anymore (and it doesn't work 😢 )


willvedd (approved):
Love automagic🎉

PR #1516, opened 2024-04-05
fix: `ListUsers`  API validation

willvedd:
## Description
Adding API validations to the `ListUsers` endpoint. 

- Exactly one user filter provided with valid type and valid relation
- Target object must be provided with valid type
- Valid relation
- Contextual tuples have valid relation, user type and object type

Some of these validations are enforced by the protobuf itself (see: https://github.com/openfga/api/pull/144) and the rest are enforced by a new `ValidateListUsersRequest` function which takes the requests and validates against the type system.

## References
- [Related API protobuf PR](https://github.com/openfga/api/pull/144)

## Review Checklist


pkg/server/server_test.go
```diff
@@ -1625,3 +1625,187 @@ func TestWriteAuthorizationModelWithExperimentalEnableModularModels(t *testing.T
...
+		relations
+			define viewer: [user]`
+
+	tests := []struct {
+		name             string
+		req              *openfgav1.ListUsersRequest
+		model            string
+		tuples           []*openfgav1.TupleKey
+		expectedUsers    []string
+		expectedErrorMsg string
```
jon-whit, line 1649: Instead of testing the error message, can we test the grpc status code that the server returns? That's more important, because the error code is subject to change.

Or can we at least add the grpc status code check to the test assertion?

tests/functional_test.go
```diff
@@ -989,6 +989,203 @@ func GRPCListObjectsTest(t *testing.T, client openfgav1.OpenFGAServiceClient) {
...
+		{
+			name: "invalid_storeID_because_too_short",
+			input: &openfgav1.ListUsersRequest{
+				StoreId:              "1",
+				AuthorizationModelId: ulid.Make().String(),
+				Relation:             "viewer",
+				Object: &openfgav1.Object{
+					Type: "document",
+					Id:   "1",
+				},
```
jon-whit, line 1076: Should we add a valid user filter here to make sure we're testing store id validation logic specifically? 

Otherwise, depending on ordering of validation (which is subject to change), the `InvalidArgument` could be stemming from that instead.

tests/functional_test.go
```diff
@@ -989,6 +989,203 @@ func GRPCListObjectsTest(t *testing.T, client openfgav1.OpenFGAServiceClient) {
...
+			},
+			output: output{
+				errorCode: codes.InvalidArgument,
+			},
+		},
+		{
+			name: "invalid_authorization_model_ID_because_extra_chars",
```
jon-whit, line 1113: Like we had the store id test `invalid_storeID_because_too_short`, can we also add the `invalid_authorization_model_ID_too_short` test?

tests/functional_test.go
```diff
@@ -989,6 +989,203 @@ func GRPCListObjectsTest(t *testing.T, client openfgav1.OpenFGAServiceClient) {
...
+					Type: "document",
+					Id:   "1",
+				},
+			},
+			output: output{
+				errorCode: codes.InvalidArgument,
+			},
+		},
+		{
+			name: "missing_object",
```
jon-whit, line 1144: Can we also add variations with empty objects. For example `empty_object` where `Object: &openfgav1.Object{}`.

There have been some subtleties that have arisen in the past because of this.

jon-whit (changes requested):
Left a few suggestions to address. Looks great though! Comments are minor.

miparnisari (changes requested):
`make test` is throwing some failures :(

pkg/server/server.go
```diff
@@ -653,15 +653,26 @@ func (s *Server) ListUsers(
 	ctx context.Context,
 	req *openfgav1.ListUsersRequest,
 ) (*openfgav1.ListUsersResponse, error) {
+
 	typesys, err := s.typesystemResolver(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
 	if err != nil {
 		return nil, err
 	}
```
miparnisari, line 660: Slightly related to this PR because this code contains validation:

You shouldn't use `s.typesystemResolver, you should call `s.resolveTypesystem` because that is the one that returns actual server errors (with correct http status codes).

I.e. your code

```
curl -v --location 'localhost:8080/stores/01HTZAKTWJ5Y69NVC5TMRW7RGY/list-users' \
--data '{
    "authorization_model_id": "01HV4DX6NX1WNTFVDVAQW4WR58", // not found
    "object": {"type": "folder", "id": "5"},
    "relation": "can_read",
    "user_filters": [{"type": "user"}]
}'

< HTTP/1.1 500 Internal Server Error
{"code":"internal_error","message":"authorization model not found"}
```

versus ListObjects (for example)

```
curl -v --location 'localhost:8080/stores/01HTZAKTWJ5Y69NVC5TMRW7RGY/list-objects' \
--data '{
    "authorization_model_id": "01HV4DX6NX1WNTFVDVAQW4WR58", // not found
    "type": "document",
    "relation": "viewer",
    "user": "user:jon"
}'

< HTTP/1.1 400 Bad Request
{"code":"authorization_model_not_found","message":"Authorization Model '01HV4DX6NX1WNTFVDVAQW4WR58' not found"}
```

This is badly designed - at minimum, typesystemResolver should give us all the correct error codes :(

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -57,6 +57,46 @@ func NewListUsersQuery(ds storage.RelationshipTupleReader, opts ...ListUsersQuer
 	return l
 }
 
+func ValidateListUsersRequest(req *openfgav1.ListUsersRequest, typesys *typesystem.TypeSystem) error {
```
miparnisari, line 60: Suggestion: move this code that is currently in `server.go`

```
for _, ctxTuple := range req.GetContextualTuples().GetTupleKeys() {
		if err := validation.ValidateTuple(typesys, ctxTuple); err != nil {
			return nil, serverErrors.HandleTupleValidateError(err)
		}
	}
```

to this function?

pkg/server/server.go
```diff
@@ -653,15 +653,26 @@ func (s *Server) ListUsers(
...
+		return nil, err
+	}
+
 	ctx = typesystem.ContextWithTypesystem(ctx, typesys)
 
+	for _, ctxTuple := range req.GetContextualTuples().GetTupleKeys() {
+		if err := validation.ValidateTuple(typesys, ctxTuple); err != nil {
+			return nil, serverErrors.HandleTupleValidateError(err)
+		}
+	}
```
miparnisari, line 673: Suggestion (not-blocking): could we experiment with putting this code inside `internal/server/list_users.go`? Then we can move the validation unit tests to `internal/server/list_users_test.go`. I think `server_test.go` has gotten too big and it mixes tests that are very unrelated.

miparnisari (approved):

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Care should be taken not to do the validation of every iteration of the recursive call though.

pkg/server/server.go
jon-whit: Suggestion to the suggestion: Start introducing per RPC test files so that we don't have a massive `server_test.go` file.

e.g.

```
pkg/
  server/
    server.go
    server_test.go
    check_test.go
    list_objects_test.go
    ...
```

pkg/server/list_users.go
```diff
@@ -0,0 +1,38 @@
...
+// that are relate via specific relation to a specific object
+func (s *Server) ListUsers(
+	ctx context.Context,
+	req *openfgav1.ListUsersRequest,
+) (*openfgav1.ListUsersResponse, error) {
+	if !s.IsExperimentallyEnabled(ExperimentalEnableListUsers) {
+		return nil, status.Error(codes.Unimplemented, "ListUsers is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-list-users` configuration option when running OpenFGA server")
+	}
+
+	typesys, err := s.typesystemResolver(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
```
miparnisari, line 24: This needs to be `s.resolveTypesystem` 😅

pkg/server/list_users_test.go
```diff
@@ -0,0 +1,217 @@
...
+			_, err = s.ListUsers(ctx, test.req)
+			e, ok := status.FromError(err)
+			require.True(t, ok)
+
+			fmt.Println("Actual", e.Code().String(), "Expected", test.expectedErrorCode.String())
+			require.Equal(t, test.expectedErrorCode, e.Code())
+		})
+	}
+}
+func TestExperimentalListUsers(t *testing.T) {
```
miparnisari, line 184: Can we add a test where if you pass an explicit model ID that can't be found, you get a http 400?

PR #1517, opened 2024-04-05
fix: check cycle detection with exclusion

miparnisari:
## Description
Fixing a panic that occurs on some cases where the model has cycles and exclusion is involved.

## References

The panic was introduced in https://github.com/openfga/openfga/pull/1372

### DIFF
--- a/internal/graph/cycle_check_resolver_test.go
+++ b/internal/graph/cycle_check_resolver_test.go
@@ -5,9 +5,16 @@ import (
 	"testing"
 
 	"github.com/google/uuid"
+	"github.com/oklog/ulid/v2"
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	parser "github.com/openfga/language/pkg/go/transformer"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/goleak"
 	"go.uber.org/mock/gomock"
+
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/storage/memory"
+	"github.com/openfga/openfga/pkg/typesystem"
 
 	"github.com/openfga/openfga/pkg/tuple"
 )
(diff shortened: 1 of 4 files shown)


internal/graph/cycle_check_resolver_test.go
```diff
@@ -60,3 +66,56 @@ func TestCycleDetectionCheckResolver(t *testing.T) {
...
+
+	typesys, err := typesystem.NewAndValidate(
+		context.Background(),
+		model,
+	)
+	require.NoError(t, err)
+
+	ctx := storage.ContextWithRelationshipTupleReader(context.Background(), ds)
+	ctx = typesystem.ContextWithTypesystem(ctx, typesys)
+	_, err = cycleDetectionCheckResolver.ResolveCheck(ctx, &ResolveCheckRequest{
```
jon-whit, line 114:
```suggestion
	resp, err = cycleDetectionCheckResolver.ResolveCheck(ctx, &ResolveCheckRequest{
```

internal/graph/cycle_check_resolver_test.go
```diff
@@ -60,3 +66,56 @@ func TestCycleDetectionCheckResolver(t *testing.T) {
...
+		StoreID:              storeID,
+		AuthorizationModelID: model.GetId(),
+		TupleKey:             tuple.NewTupleKey("group:1", "blocked", "user:jon"),
+		RequestMetadata:      NewCheckRequestMetadata(25),
+	})
+	require.NoError(t, err)
```
jon-whit, line 120:
```suggestion
	require.NoError(t, err)
	require.False(t, resp.GetAllowed())
```

miparnisari: I get `allowed=true` 😬

jon-whit: Wait a second, this should be an error `ErrCycleDetected`. Sorry, I stand corrected.

If you start at `group#blocked` --> `group#member` and `group#member` depends on `group#blocked`, then that's a cycle and you cannot resolve. 

So this should be an `ErrCycleDetected`.

miparnisari: should we tackle this in a different PR?

jon-whit (approved):

PR #1521, opened 2024-04-09
fix(break): ensure usersets are self defining

miparnisari:
## Description
Resurrect https://github.com/openfga/openfga/pull/1066, plus:

- Changes to Write API to forbid writing implicit tuples. ⚠️ This is breaking behavior. Previously, Writes of the form `document:1 # viewer @ document:1#viewer` would work. Now, they will throw HTTP 400.
- ~TODO more tests for Expand API~ https://github.com/openfga/openfga/pull/1522

Docs are in https://github.com/openfga/api/pull/147/files

## References

Approved RFC https://github.com/openfga/rfcs/blob/main/20240328-queries-with-usersets.md


willvedd:
Consideration: can and should we change the pruned edge logic to consider reflexive usersets as a valid edge? We had to implement a workaround in ListUsers to bypass the edge validation to enable full reflexive userset support.

pkg/server/commands/write_test.go
```diff
@@ -175,6 +203,14 @@ func TestValidateWriteRequest(t *testing.T) {
 					ReadAuthorizationModel(gomock.Any(), gomock.Any(), gomock.Any()).
 					Return(&openfgav1.AuthorizationModel{
 						SchemaVersion: typesystem.SchemaVersion1_1,
+						TypeDefinitions: parser.MustTransformDSLToProto(`
```
jon-whit, line 206: Can't you just do `testutils.MustTransformDSLToProtoWithID(modelStr)` for short here?

pkg/server/commands/write_test.go
jon-whit, line 1: Should we not also have the appropriate tests in `pkg/server/test/write.go`? Nothing is guaranteeing that the OpenFGA Server delegates to the `pkg/server/commands/WriteCommand`, so having this test here is good but not as confidence generating as testing the server RPC, and ultimately that is the unit that needs to be the most tested contractually because it is the code signature that we expect external Go apps to import and use. If the WriteCommand returns a specific error and the RPC handler maps that differently, then the Server response may not match what we expect it to.

So please add the higher fidelity test at the Server layer in favor of or in addition to the tests added here.

jon-whit (changes requested):
Needing equivalent testing at the `pkg/server/test/write.go` level.

pkg/server/commands/write_test.go
miparnisari: > Nothing is guaranteeing that the OpenFGA Server delegates to the pkg/server/commands/WriteCommand

I am not changing the delegation code in this PR. But yes we have lots of tests guaranteeing that. If you comment out that delegation, you get lots of breaking tests in the `tests` folder.

jon-whit: But we're adding new functionality that's left untested in the Server. No matter how the implementation of the Server changes the behavior around not being able to Write `obj#relation@obj#relation` should always pass, so that's why we should have that test for the Server.

miparnisari: I added one just to move along this PR, but I fundamentally disagree with the idea of duplicate tests. The test that I added is now going to run for every datastore in existence, even though this new check is datastore-agnostic. Also  we've never changed the implementation of a server RPC at runtime, so the new test doesn't give me extra confidence at this moment.

jon-whit: > The test that I added is now going to run for every datastore in existence

It doesn't have to, just use a mock instead for that test.


In the server code 
```go
func (s *Server) Write(...) (..., error) {
    writeCmd := NewWriteCmd(...)
    return writeCmd.Execute(...)
}
```
all I have to do is remove the cmd and if we're testing the cmd instead of the RPC then we don't catch our regressions.
```
func (s *Server) Write(...) (..., error) {
    return ..., status.Error(codes.Internal, ...)
}
```
The tests for the write command still pass, but our server now is broken, and the whole reason we're writing tests is to avoid some goober like me breaking code that impacts the server response behavior. The Server is the unit under test here, the command is just the proxy for it.

jon-whit (approved):

pkg/server/commands/write_test.go
miparnisari: > You're always a proponent of writing tests that will cause failures if you remove the applicable code,

Exactly, and we already have tests that assert this. Again, try making the change you suggested, you get tons of breaking test. Adding one more doesn't help us in any way.

jon-whit: Just because unrelated tests cause unrelated failures doesn't mean that you have coverage with the code introduced here. That's a false sense of test coverage. If we want to maintain a positive sense of test coverage no matter how the code changes we need test the unit under test directly, which we're not doing if we test the proxy instead of the actual function under test.

The exception to this would be if you add a test for the Server Write RPC that ensures a WriteCmd is created and is delegated to via a mock of some form. That would ensure we're always proxying, and assuming our proxy is well tested then we're 👍

PR #1522, opened 2024-04-10
test: add Expand tests

miparnisari:
## Description
In preparation for https://github.com/openfga/openfga/pull/1521, add more tests (starting in line 568) asserting the current behavior of Expand API.

## References
https://github.com/openfga/rfcs/pull/23, specifically addressing https://github.com/openfga/rfcs/pull/23#discussion_r1546394083

### DIFF
--- a/pkg/server/test/expand.go
+++ b/pkg/server/test/expand.go
@@ -11,6 +11,8 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/testing/protocmp"
 
+	"github.com/openfga/openfga/pkg/testutils"
+
 	"github.com/openfga/openfga/pkg/server/commands"
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
 	"github.com/openfga/openfga/pkg/storage"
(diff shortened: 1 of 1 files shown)


pkg/server/test/expand.go
```diff
@@ -810,53 +533,244 @@ func TestExpandQuery(t *testing.T, datastore storage.OpenFGADatastore) {
...
+			},
+			expected: &openfgav1.ExpandResponse{
+				Tree: &openfgav1.UsersetTree{
+					Root: &openfgav1.UsersetTree_Node{
+						Name: "group:1#viewer",
+						Value: &openfgav1.UsersetTree_Node_Leaf{
+							Leaf: &openfgav1.UsersetTree_Leaf{
+								Value: &openfgav1.UsersetTree_Leaf_Users{
+									Users: &openfgav1.UsersetTree_Users{
+										Users: []string{"group:1#viewer"},
```
miparnisari, line 654: I think this is the case that we will want to address in https://github.com/openfga/openfga/pull/1521, right @rhamzeh ?

rhamzeh: Yes... and no :D

This should return empty, if we're assuming it is always truthy and blocking writing that tuple, so if the tuple can no longer be written, expand will not return it but it will be assumed true.

But I don't think we should make this change without blocking writing those tuples

miparnisari: Ah, yes. In https://github.com/openfga/openfga/pull/1521 I explicitly forbid writing that tuple.

rhamzeh: I think for that case, if it's an existing tuple I'm fine with returning it.

The only reason I'm saying that is that I would prefer expand has no exceptions we need to maintain. If it's in the DB, it's fair game. Does that make sense?

pkg/server/test/expand.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/testing/protocmp"
 
+	"github.com/openfga/openfga/pkg/testutils"
+
```
adriantam, line 15: nitpik: please remove empty line

jon-whit (approved):
Looks good to me.

Note to self - need to follow up on the Expand response structure. It makes no sense why a computed relationships is represented as a "leaf". Leafs in the Expansion tree should be reserved for direct relationships. Our Expand API response is misleading or at minimum it is not clear in its semantic intent.

PR #1525, opened 2024-04-10
fix: `ListUsers` experimental flag

willvedd:
## Description
Wiring-up ListUsers endpoint with the `--experimentals` flag. This allows it to be enabled by running `openfga run --experimentals enable-list-users`. This follows a similar pattern as modular models experimental enablement in #1443. 

## References
- Similar PR: #1443 

## Review Checklist

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -664,7 +664,7 @@ func (s *Server) ListUsers(
 	req *openfgav1.ListUsersRequest,
 ) (*openfgav1.ListUsersResponse, error) {
 	if !s.IsExperimentallyEnabled(ExperimentalEnableListUsers) {
-		return nil, fmt.Errorf("ListUsers is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-list-users` configuration option when running OpenFGA server")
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -653,6 +663,10 @@ func (s *Server) ListUsers(
 	ctx context.Context,
 	req *openfgav1.ListUsersRequest,
 ) (*openfgav1.ListUsersResponse, error) {
+	if !s.IsExperimentallyEnabled(ExperimentalEnableListUsers) {
+		return nil, fmt.Errorf("ListUsers is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-list-users` configuration option when running OpenFGA server")
```
miparnisari, line 667: This is throwing a 500.

```suggestion
		return nil, status.Error(codes.Unimplemented, "ListUsers is not enabled. ....
```

Would it be possible to assert on this status code in the test?

miparnisari (approved):

pkg/server/server.go
willvedd: Good call, was wondering if there was an unimplemented code. I changed that and also added a status code assertion on the test too.

PR #1531, opened 2024-04-12
feat: add grpc tag to indicate request had been throttled

adriantam:
## Description

We need to convey request has been throttled so that caller can distinguish whether timeout is due to throttling.
Note that we cannot return timeout error in the distpatch throttler itself because the [caller](https://github.com/openfga/openfga/blob/2cdbb76c654ec92232f8295428dab2012a600642/internal/graph/check.go#L242) will not wait for error response in case of context timeout.

The caller (for example), the user of the OpenFGA library, may choose to use this grpcTag ("Throttled") to modify the error message returned due to request timeout.  They may also update the returned header to indicate the request has been throttled.

## References
Continuation of https://github.com/openfga/openfga/pull/1440

## Review Checklist


miparnisari:
> We need to convey request has been throttled so that caller can distinguish whether timeout is due to throttling.

Can you ellaborate on this in the PR description? Who is the caller that will take action based on this tag? And what will that action be?

FWIW, today, any context tags are logged automatically. Also once you apply a tag to a context, any parent or children contexts also get that tag.

adriantam:
> > We need to convey request has been throttled so that caller can distinguish whether timeout is due to throttling.
> 
> Can you ellaborate on this in the PR description? Who is the caller that will take action based on this tag? And what will that action be?
> 
> FWIW, today, any context tags are logged automatically. Also once you apply a tag to a context, any parent or children contexts also get that tag.

Done.

miparnisari:
> the user of the OpenFGA library, may choose to use this grpcTag ("Throttled") to modify the error message returned due to request timeout

This won't work if a person uses OpenFGA as a library and doesn't wrap it within a server. The grpc tags middleware must be running. I.e. this doesn't print anything:

```go
    ctx := context.Background()

	grpc_ctxtags.Extract(ctx).Set("key-a", true)

	tags := grpc_ctxtags.Extract(ctx)
	for k, v := range tags.Values() {
		fmt.Println("Tag: %s, Value: %s", k, v)
	}
```

miparnisari (approved):

PR #1534, opened 2024-04-15
fix: `ListUsers` excluded users when negating wildcards

willvedd:
## Description
This PR introduces the concept of negation "differences" which represent the list of users that do not apply to a negation clause. This occurs in models that have chained negation where the intermediate negation includes a public typed wildcard. 

**Example:**
```
model
	schema 1.1
type user
type document
	relations
```
```
```

In this example, the "unblocked" status of `user:jon` was being lost when evaluating the outer negation because the exclusion to the negation was not being communicated. This is considered an "**excluded user**". The solution here is to keep track of the excluded users so that the outer negation context can un-exclude the proper subjects. We also return the excluded users along with the response for the client to handle appropriately. See: https://github.com/openfga/api/pull/144/files#diff-d3a74ea9ecaac0a46611253dc0c4bc9c402e2ad9eb255cd41d07266453e258f0R864

## References

## Review Checklist


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -431,24 +442,47 @@ func (l *listUsersQuery) expandExclusion(
...
 	}
 	subtractFoundUsersMap := make(map[string]struct{}, len(baseFoundUsersMap))
+	negatedSubtractionMap := make(map[string]struct{})
 	for fu := range subtractFoundUsersCh {
-		key := tuple.UserProtoToString(fu)
+		for _, s := range fu.subtracted {
+			negatedSubtractionMap[tuple.UserProtoToString(s)] = struct{}{}
+		}
+		key := tuple.UserProtoToString(fu.user)
 		subtractFoundUsersMap[key] = struct{}{}
```
miparnisari, line 455: If you have `user:* but not user:maria` as an intermediate result, I don't think it is safe to separate them into separate maps.

For example, I believe this is why this test fails:

```go
tests := ListUsersTests{
		{
			name: "Mixed",
			model: `
			model
				schema 1.1
			type user
			type org
				relations
					define member: [user]
			type document
				relations
					define viewer: viewer1 or viewer2
					define viewer1: [org#member] but not blocked1
					define viewer2: [user:*] but not blocked2
					define blocked1: [user]
					define blocked2: [user]
		`,
				tuple.NewTupleKey("document:1", "viewer1", "org:fga#member"),
				tuple.NewTupleKey("org:fga", "member", "user:maria"),

				tuple.NewTupleKey("document:1", "viewer2", "user:*"),
				tuple.NewTupleKey("document:1", "blocked1", "user:will"),
			},
				Object:      &openfgav1.Object{Type: "document", Id: "1"},
				Relation:    "viewer",
				UserFilters: []*openfgav1.ListUsersFilter{{Type: "user"}},
			},
			expectedUsers: []string{"user:*", "user:maria", "user:will"},
		},
	}
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2129,3 +2137,146 @@ func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
...
+				tuple.NewTupleKey("document:1", "blocked", "user:maria"),
+				tuple.NewTupleKey("document:1", "unblocked", "user:*"),
+			},
+			expectedUsers: []string{"user:*", "user:jon", "user:maria"},
+		},
+	}
+
+	for i := range tests {
+		tests[i].model = model
+		tests[i].name = fmt.Sprintf("chained_negation_%d", i+1)
```
miparnisari, line 2273: nit: it's going to be very hard to find one specific test if you disconnect it's name from the scenario. Prefer to put the name as a hardcoded field.

pkg/server/commands/listusers/list_users_rpc.go
willvedd: > If you have `user:* but not user:maria` as an intermediate result, I don't think it is safe to separate them into separate maps.

Can you elaborate on what this means?

That aside, `user:will` is not being returned because that user is not assigned to either `org:fga#member` or `document:1#viewer`. This works exactly as I would expect.

miparnisari: > Can you elaborate on what this means?

i need to give this more thought but my gut reaction is that we should probably key just one map with `user:* but not user:maria`, not 1 map with key `user:*` and 1 map with key `user:maria`

miparnisari: > That aside, user:will is not being returned because that user is not assigned to either org:fga#member or document:1#viewer. This works exactly as I would expect.

Because of the tuple `document:1 # viewer2 @ user:*`, user:will _should_ be returned. The corresponding Check request returns `allowed=true`: https://play.fga.dev/sandbox/?store=asd2

willvedd: Checking `user:foo`,`user:bar` and `user:baz` all return `allowed=true` too but I would not expect them to be returned from ListUsers.

miparnisari: Oh... my assumption was that if we have a tuple mentioning the user, then it'd be somehow included in the response. I need to double-check other tests

miparnisari: Welp, I tried to poke holes but couldn't 😄  nice work Will!

miparnisari (approved):

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -26,6 +26,11 @@ type listUsersQuery struct {
 	resolveNodeBreadthLimit uint32
 }
 
+type foundUser struct {
+	user       *openfgav1.User
+	exceptions []*openfgav1.User
```
jon-whit, line 31:
```suggestion
	excludedUsers []*openfgav1.User
```
To be more consistent with the naming of the API itself.

jon-whit (changes requested):

jon-whit:
Superseded by #1590

PR #1539, opened 2024-04-16
`ListUsers`: add tracing

miparnisari:
## Description
Add basic tracing around one ListUsers request. This will allow us to see where errors are coming from, and troubleshoot performance issues.

Note:

- This doesn't cover tracing around evaluation of conditions, we'll have to add this later.

## Testing

Happy path. See how relevant attributes are logged.

image

Mocked error coming from DB propagates error up the call stack.

image

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -10,6 +10,12 @@ import (
 	"github.com/hashicorp/go-multierror"
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/sourcegraph/conc/pool"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+
+	"github.com/openfga/openfga/pkg/telemetry"
+
+	"github.com/openfga/openfga/pkg/logger"
 
 	"github.com/openfga/openfga/pkg/storage/storagewrappers"
 
@@ -20,7 +26,10 @@ import (
 	"github.com/openfga/openfga/pkg/typesystem"
 )
 
+var tracer = otel.Tracer("openfga/pkg/server/commands/list_users")
+
 type listUsersQuery struct {
+	logger                  logger.Logger
 	ds                      storage.RelationshipTupleReader
 	typesystemResolver      typesystem.TypesystemResolverFunc
 	resolveNodeBreadthLimit uint32
@@ -37,10 +46,17 @@ type listUsersQuery struct {
 
 type ListUsersQueryOption func(l *listUsersQuery)
 
+func WithListUsersQueryLogger(l logger.Logger) ListUsersQueryOption {
+	return func(rq *listUsersQuery) {
+		rq.logger = l
+	}
+}
+
 // NewListUsersQuery is not meant to be shared.
 func NewListUsersQuery(ds storage.RelationshipTupleReader, opts ...ListUsersQueryOption) *listUsersQuery {
 	l := &listUsersQuery{
-		ds: ds,
+		logger: logger.NewNoopLogger(),
+		ds:     ds,
 		typesystemResolver: func(ctx context.Context, storeID, modelID string) (*typesystem.TypeSystem, error) {
 			typesys, exists := typesystem.TypesystemFromContext(ctx)
 			if !exists {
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -1156,6 +1156,7 @@ func (s *Server) resolveTypesystem(ctx context.Context, storeID, modelID string)
 
 	typesys, err := s.typesystemResolver(ctx, storeID, modelID)
 	if err != nil {
+		telemetry.TraceError(span, err)
```
miparnisari, line 1159: arguably this should be in the `main` branch, I can send a separate PR

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -117,6 +137,7 @@ func (l *listUsersQuery) ListUsers(
 	for foundUser := range foundUsersUnique {
 		foundUsers = append(foundUsers, tuple.StringToUserProto(foundUser))
 	}
+	span.SetAttributes(attribute.Int("result_count", len(foundUsers)))
```
willvedd, line 140: **Note for future:** We'll also need another attribute for "exceptions" in #1534

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -186,7 +207,11 @@ func (l *listUsersQuery) expand(
 	req *internalListUsersRequest,
 	foundUsersChan chan<- *openfgav1.User,
 ) error {
+	ctx, span := tracer.Start(ctx, "expand")
+	defer span.End()
+
 	if enteredCycle(req) {
+		span.SetAttributes(attribute.Bool("cycle_detected", true))
```
willvedd, line 214: **Question:** Similarly to here, do you think it's worth to add an attribute for `doesHavePossibleEdges` validations that return false?

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -258,17 +292,25 @@ func (l *listUsersQuery) expandRewrite(
...
 
 func (l *listUsersQuery) expandDirect(
 	ctx context.Context,
 	req *internalListUsersRequest,
 	foundUsersChan chan<- *openfgav1.User,
 ) error {
+	ctx, span := tracer.Start(ctx, "expandDirect")
+	defer span.End()
 	typesys, err := l.typesystemResolver(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
 	if err != nil {
```
willvedd, line 315: `telemetry.TraceError(span, err)`?

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -459,7 +511,12 @@ func (l *listUsersQuery) expandExclusion(
 		}
 	}
 
-	return errs
+	finalErr := errs
```
willvedd, line 514: **Question:** Why the distinction of `finalErr`? If possible, I'd prefer to just keep as `errs`.

pkg/server/list_users.go
```diff
@@ -17,22 +23,47 @@ func (s *Server) ListUsers(
...
+		listusers.WithListUsersQueryLogger(s.logger))
+	resp, err := listUsersQuery.ListUsers(ctx, req)
+	if err != nil {
+		telemetry.TraceError(span, err)
+		return nil, err
+	}
+	return resp, nil
+}
+
+func UserFiltersToString(filter []*openfgav1.ListUsersFilter) string {
```
willvedd, line 60:
```suggestion
func userFiltersToString(filter []*openfgav1.ListUsersFilter) string {
```

pkg/server/server.go
willvedd: Should it? The convention seems to be tracing a function's returned error but not necessarily in the function itself.

miparnisari: Not sure what you mean? This function itself is already creating a span, so if it errors out we should mark it as errored.

pkg/server/commands/listusers/list_users_rpc.go
miparnisari: Yes I can add one.

pkg/server/commands/listusers/list_users_rpc.go
willvedd: It's just not ever clear where or if the error is being traced. Nothing actionable though.

pkg/server/commands/listusers/list_users_rpc.go
miparnisari: My bad, yes. This was an unfounded fear related to https://100go.co/#returning-a-nil-receiver-45

pkg/server/commands/listusers/list_users_rpc.go
miparnisari: The error will be reported here:

```go
listUsersQuery := listusers.NewListUsersQuery(s.datastore,
		listusers.WithListUsersQueryLogger(s.logger))
	resp, err := listUsersQuery.ListUsers(ctx, req)
	if err != nil {
		return nil, err
	}
```

willvedd: My point is, its seemingly arbitrary wether it goes inside the function (which `typesystemResolver ` does) or reads the output (above example). But its really not a big deal.

willvedd (approved):

PR #1545, opened 2024-04-17
test: Adding `ListUsers` case to TestReleasesConnections

willvedd:
## Description
Adding a ListUsers-focused test case to `TestReleasesConnections` (previously named `TestListObjectsReleasesConnections`). This test ensures that ListUsers is properly releasing DB connections after the API request completes.

## References
Parent PR: #1428 
Related PR: #1061 

## Review Checklist

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -606,7 +606,6 @@ func TestReleasesConnections(t *testing.T) {
 
 	s := MustNewServerWithOpts(
 		WithDatastore(storagewrappers.NewContextWrapper(ds)),
-		WithMaxConcurrentReadsForListObjects(1),
 		WithExperimentals(ExperimentalEnableListUsers),
 	)
 	t.Cleanup(s.Close)
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -589,7 +589,7 @@ type repo
 	require.True(t, checkResponse.GetAllowed())
 }
 
-func TestListObjectsReleasesConnections(t *testing.T) {
+func TestReleasesConnections(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/server/server_test.go
```diff
@@ -607,6 +607,7 @@ func TestListObjectsReleasesConnections(t *testing.T) {
 	s := MustNewServerWithOpts(
 		WithDatastore(storagewrappers.NewContextWrapper(ds)),
 		WithMaxConcurrentReadsForListObjects(1),
```
miparnisari, line 609: @jon-whit is `WithMaxConcurrentReadsForListObjects(1),` relevant to this test? I think not 🤔 

If it is relevant, do we need to add `WithMaxConcurrentReadsForListUsers(1)` to make this new test useful?

If it is not relevant, let's remove it @willvedd.

miparnisari (approved):

pkg/server/server_test.go
jon-whit: It appears as though it is - we use it to give better confidence that we have blocked in flight Goroutines when the context is cancelled.

willvedd: @miparnisari @jon-whit The tests will pass with or without that config and if its configured with any non-zero positive number. This suggests that `WithMaxConcurrentReadsForListObjects(1),` is not essential in this test. 

I'm actually not even able to make this test fail with any change in ListUsers. Is this test even necessary?

miparnisari: Yes, i'd keep it. It may not be applicable to the current implementation, but it's useful to know that we are not hogging connections after the API call ends.

PR #1546, opened 2024-04-18
feat: allow per request specification of dispatch throttle threshold

adriantam:
## Description
Add ability to allow per request specification on dispatch throttle threshold.  This allows other software that uses OpenFGA as 3rd party library to override the dispatch threshold on a per request basis.

## Review Checklist


.config-schema.json
```diff
@@ -451,6 +451,12 @@
                     "type": "integer",
                     "default": "100",
                     "x-env-variable": "OPENFGA_DISPATCH_THROTTLING_THRESHOLD"
+                },
+                "maxThreshold": {
+                    "description": "define the maximum dispatch threshold beyond above which requests will be throttled. 0 will use the threshold value as maximum",
+                    "type": "integer",
+                    "default": "0",
```
jon-whit, line 458: It appears the default max is 200 in the code. Why is it 0 here?

adriantam: 0 means using the threshold as maximum.  The default max in code is 0

internal/graph/dispatch_throttling_check_resolver.go
```diff
@@ -14,8 +14,9 @@ import (
 
 // DispatchThrottlingCheckResolverConfig encapsulates configuration for dispatch throttling check resolver.
 type DispatchThrottlingCheckResolverConfig struct {
-	Frequency time.Duration
-	Threshold uint32
+	Frequency    time.Duration
+	Threshold    uint32
```
adriantam, line 18: As discussed offline, we should rename this as DefaultThreshold

pkg/telemetry/context.go
```diff
@@ -5,12 +5,17 @@ import (
...
 type rpcContextName string
+type dispatchThrottlingThreshold uint32
 
 const (
 	rpcInfoContextName rpcContextName = "rpcInfo"
 	Throttled          string         = "Throttled"
 )
 
+const (
+	DispatchThrottlingThreshold dispatchThrottlingThreshold = iota
```
jon-whit, line 16: Please use a context getter and setter pattern. It's more idiomatic and safer Go when it comes to ctx scoped values. For more info see: https://go.dev/blog/context

```go

type dispatchThrottlingThresholdCtxKey string

func ContextWithDispatchThrottlingThreshold(ctx context.Context, threshold uint32) context.Context {
  return context.WithValue(ctx, dispatchThrottlingThresholdCtxKey, threshold)
}

func DispatchThrottlingThresholdFromContext(ctx) (uint32, bool) {
    return ctx.Value(dispatchThrottlingThresholdCtxKey).(uint32)
}
```

pkg/server/server.go
```diff
@@ -137,6 +137,7 @@ type Server struct {
 	dispatchThrottlingCheckResolverEnabled   bool
 	dispatchThrottlingCheckResolverFrequency time.Duration
 	dispatchThrottlingThreshold              uint32
+	dispatchThrottlingMaxThreshold           uint32
```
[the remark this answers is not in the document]
adriantam: as discussed offline, we need to as the threshold is the default.

pkg/server/server.go
```diff
@@ -318,9 +319,19 @@ func WithDispatchThrottlingCheckResolverFrequency(frequency time.Duration) OpenF
 }
 
 // WithDispatchThrottlingCheckResolverThreshold define the number of dispatches to be throttled.
+// In addition, it will update dispatchThrottlingMaxThreshold if required.
 func WithDispatchThrottlingCheckResolverThreshold(threshold uint32) OpenFGAServiceV1Option {
 	return func(s *Server) {
 		s.dispatchThrottlingThreshold = threshold
+		s.dispatchThrottlingMaxThreshold = max(s.dispatchThrottlingMaxThreshold, threshold)
```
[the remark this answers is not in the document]
adriantam: changed to error out

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -54,14 +96,114 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
...
+			return response, nil
+		}).Times(1)
+
+		ctx := context.Background()
+		ctx = telemetry.ContextWithDispatchThrottlingThreshold(ctx, uint32(205))
+
+		_, err := dut.ResolveCheck(ctx, req)
+		require.NoError(t, err)
+		require.Equal(t, 1, resolveCheckDispatchedCounter)
+		require.False(t, grpc_ctxtags.Extract(ctx).Has(telemetry.Throttled))
```
jon-whit, line 136: This should be a check on `WasThrottled`, not `grpc_ctxtags`.

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -54,14 +96,114 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
...
+		dut := NewDispatchThrottlingCheckResolver(dispatchThrottlingCheckResolverConfig)
+		defer dut.Close()
+
+		initialMockResolver := NewMockCheckResolver(ctrl)
+		dut.SetDelegate(initialMockResolver)
+
+		// this is to simulate how many times request has been dispatched
+		// Since this is a small value, we will not expect throttling
+		req := &ResolveCheckRequest{RequestMetadata: NewCheckRequestMetadata(10)}
+		req.GetRequestMetadata().DispatchCounter.Store(203)
```
jon-whit, line 118: The threshold is 205 below, but this is 203. If we want this to be throttled, then this needs to be higher.

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -54,14 +96,114 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
...
+
+		var goFuncInitiated sync.WaitGroup
+		goFuncInitiated.Add(1)
+
+		ctx := context.Background()
+
+		go func() {
+			defer goFuncDone.Done()
+			goFuncInitiated.Done()
+			ctx = grpc_ctxtags.SetInContext(ctx, grpc_ctxtags.NewTags())
```
jon-whit, line 180: Why do we need `grpc_ctxtags`?

adriantam: we don't.  it is removed

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -54,14 +96,113 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
...
+
+		ctx := context.Background()
+
+		go func() {
+			defer goFuncDone.Done()
+			goFuncInitiated.Done()
+			_, err := dut.ResolveCheck(ctx, req)
+			//nolint:testifylint
+			require.NoError(t, err)
+		}()
```
jon-whit, line 183: Declare `err` externally and assign it inside, then `require.NoError(t, err)` outside. For example,

```go
var resolveCheckErr error
go func() {
  _ , resolveCheckErr := dut.ResolveCheck(ctx, req)
}()
require.NoError(t, resolveCheckErr)
```

adriantam: updated

internal/graph/dispatch_throttling_check_resolver_test.go
```diff
@@ -91,13 +173,72 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
...
+
+		go func() {
+			defer goFuncDone.Done()
+			goFuncInitiated.Done()
+			_, resolveCheckErr = dut.ResolveCheck(ctx, req)
+		}()
+
+		goFuncInitiated.Wait()
+		// Before we send the tick, we don't expect the dispatch to run
+		time.Sleep(1 * time.Millisecond)
```
jon-whit, line 186: This is superfulous, because it doesn't guarantee that `ResolveCheck` has been called before this assertion anyways (`goFuncInitiated.Done()` runs before `ResolveCheck` is invoked).

At the time you run the assertion you shouldn't expect the dispatch to have run yet, because we haven't produced a tick. So can't we just drop the `goFuncInitiated` WaitGroup and run this assertion right after the goroutine and expect the same outcome?

Only after the `nonBlockingSend` shouold we expect a mutation to the resolver counter.

adriantam: We still need to have goFuncInitiated.  Otherwise, we may have signal before the item is in queue.  Thus, the signal is ignored (remember that it is no blocking send).

adriantam: Discussed offline - will use ticker to move forward

PR #1552, opened 2024-04-18
feat: return HTTP code 422 if timeout due to throttling

adriantam:
## Description
Instead of return HTTP 500, return HTTP code 422 if check request timeouts due to throttling

## References
https://github.com/openfga/api/pull/151

## Note
DO NOT MERGE until https://github.com/openfga/api/pull/151 is merged (and go.sum/go.mod use the official reference)

## Review Checklist

### DIFF
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -49,6 +49,7 @@ import (
 	authnmw "github.com/openfga/openfga/internal/middleware/authn"
 	serverconfig "github.com/openfga/openfga/internal/server/config"
 	"github.com/openfga/openfga/pkg/logger"
+	"github.com/openfga/openfga/pkg/middleware"
 	httpmiddleware "github.com/openfga/openfga/pkg/middleware/http"
 	"github.com/openfga/openfga/pkg/middleware/logging"
 	"github.com/openfga/openfga/pkg/middleware/requestid"
@@ -69,6 +70,8 @@ import (
 const (
 	datastoreEngineFlag = "datastore-engine"
 	datastoreURIFlag    = "datastore-uri"
+
+	additionalUpstreamTimeout = 3 * time.Second
 )
 
 func NewRunCommand() *cobra.Command {
@@ -422,6 +425,13 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 		),
 	)
 
+	if config.HTTP.UpstreamTimeout > 0 {
+		timeoutMiddleware := middleware.NewTimeoutHandler(config.HTTP.UpstreamTimeout, s.Logger)
+
+		serverOpts = append(serverOpts, grpc.ChainUnaryInterceptor(timeoutMiddleware.NewUnaryTimeoutInterceptor()))
+		serverOpts = append(serverOpts, grpc.ChainStreamInterceptor(timeoutMiddleware.NewStreamTimeoutInterceptor()))
+	}
+
 	if config.GRPC.TLS.Enabled {
 		if config.GRPC.TLS.CertPath == "" || config.GRPC.TLS.KeyPath == "" {
 			return errors.New("'grpc.tls.cert' and 'grpc.tls.key' configs must be set")
@@ -528,7 +538,11 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 	var httpServer *http.Server
 	if config.HTTP.Enabled {
 		// Set a request timeout.
-		runtime.DefaultContextTimeout = config.HTTP.UpstreamTimeout
+		if config.HTTP.UpstreamTimeout > 0 {
+			// the runtime.DefaultContextTimeout is last resort. Ideally, timeout is handled via
+			// middleware timeout so that we can process the proper error code
+			runtime.DefaultContextTimeout = config.HTTP.UpstreamTimeout + additionalUpstreamTimeout
+		}
 
 		dialOpts := []grpc.DialOption{
 			grpc.WithBlock(),
--- a/internal/graph/graph.go
+++ b/internal/graph/graph.go
@@ -56,6 +56,9 @@ type ResolveCheckRequestMetadata struct {
 	// The contents of this counter will be written by concurrent goroutines.
 	// After the root problem has been solved, this value can be read.
 	DispatchCounter *atomic.Uint32
+
+	// HasThrottled indicates whether the request had been throttled
+	HasThrottled *atomic.Bool
 }
 
 func NewCheckRequestMetadata(maxDepth uint32) *ResolveCheckRequestMetadata {
@@ -63,6 +66,7 @@ func NewCheckRequestMetadata(maxDepth uint32) *ResolveCheckRequestMetadata {
 		Depth:               maxDepth,
 		DatastoreQueryCount: 0,
 		DispatchCounter:     new(atomic.Uint32),
+		HasThrottled:        new(atomic.Bool),
 	}
 }
 
--- a/pkg/middleware/timeout.go
+++ b/pkg/middleware/timeout.go
@@ -1,0 +1,65 @@
+package middleware
+
+import (
+	"context"
+	"time"
+
+	grpcvalidator "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/validator"
+	"google.golang.org/grpc"
+
+	"github.com/openfga/openfga/pkg/logger"
+)
+
+// TimeoutHandler sets the timeout in each request
+type TimeoutHandler struct {
+	timeout time.Duration
+	logger  logger.Logger
+}
+
+// NewTimeoutHandler returns new TimeoutHandler that timeouts request if it
+// exceeds the timeout value
+func NewTimeoutHandler(timeout time.Duration, logger logger.Logger) *TimeoutHandler {
+	return &TimeoutHandler{
+		timeout: timeout,
+		logger:  logger,
+	}
+}
+
+// NewUnaryTimeoutInterceptor configures the timeout expected in the http server side
+// We need to use this middleware instead of relying on runtime.DefaultContextTimeout to allow us
+// to return proper error code
+func (h *TimeoutHandler) NewUnaryTimeoutInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		ctx, cancel := context.WithTimeout(ctx, h.timeout)
+		defer cancel()
+		return handler(ctx, req)
+	}
+}
+
+// NewStreamTimeoutInterceptor configures the timeout expected in the http server side
+// We need to use this middleware instead of relying on runtime.DefaultContextTimeout to allow us
+// to return proper error code
+func (h *TimeoutHandler) NewStreamTimeoutInterceptor() grpc.StreamServerInterceptor {
+	validator := grpcvalidator.StreamServerInterceptor()
+	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		return validator(srv, stream, info, func(srv interface{}, ss grpc.ServerStream) error {
+			ctx, cancel := context.WithTimeout(stream.Context(), h.timeout)
+			defer cancel()
+
+			return handler(srv, &recvWrapper{
+				ctx:          ctx,
+				ServerStream: ss,
+			})
+		})
+	}
+}
+
+type recvWrapper struct {
+	ctx context.Context
+	grpc.ServerStream
+}
+
+// Context returns the context associated with the recvWrapper.
+func (r *recvWrapper) Context() context.Context {
+	return r.ctx
+}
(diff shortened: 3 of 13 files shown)


cmd/run/run.go
```diff
@@ -422,6 +425,13 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 		),
 	)
 
+	if config.HTTP.UpstreamTimeout > 0 {
+		timeoutMiddleware := middleware.NewTimeoutHandler(config.HTTP.UpstreamTimeout, s.Logger)
+
+		serverOpts = append(serverOpts, grpc.ChainUnaryInterceptor(timeoutMiddleware.NewUnaryTimeoutInterceptor()))
```
jon-whit, line 431: We shouldn't be appending gRPC timeout middleware if `config.HTTP.UpstreamTimeout` is configured. That config is specific to the HTTP reverse proxy. We need timeout semantics for gRPC and HTTP, and the config shouldn't be so specific to HTTP.

Please adjust this such that we have a top-level `config.RequestTimeout`  and below where `if config.HTTP.Enabled` then we set `runtime.DefaultContextTimeout` to `config.RequestTimeout + 3s`

internal/graph/graph.go
```diff
@@ -56,13 +56,17 @@ type ResolveCheckRequestMetadata struct {
 	// The contents of this counter will be written by concurrent goroutines.
 	// After the root problem has been solved, this value can be read.
 	DispatchCounter *atomic.Uint32
+
+	// HasThrottled indicates whether the request had been throttled
+	HasThrottled *atomic.Bool
```
jon-whit, line 61:
```suggestion
	WasThrottled *atomic.Bool
```

pkg/middleware/timeout.go
```diff
@@ -0,0 +1,65 @@
...
+// to return proper error code
+func (h *TimeoutHandler) NewUnaryTimeoutInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		ctx, cancel := context.WithTimeout(ctx, h.timeout)
+		defer cancel()
+		return handler(ctx, req)
+	}
+}
+
+// NewStreamTimeoutInterceptor configures the timeout expected in the http server side
```
jon-whit, line 39: This is gRPC middleware, please avoid mention of HTTP.

pkg/server/server.go
```diff
@@ -712,6 +712,21 @@ func (s *Server) Write(ctx context.Context, req *openfgav1.WriteRequest) (*openf
 	})
 }
 
+// isDeadlineExceeded returns whether err is DeadlineExceeded or it is an internal error and the internal error
+// is DeadlineExceeded
+func isDeadlineExceeded(err error) bool {
```
jon-whit, line 717: Can we colocate this alongside the errors code it is associated with in `pkg/server/errors`?

pkg/server/server_test.go
```diff
@@ -518,6 +518,77 @@ func TestThreeProngThroughVariousLayers(t *testing.T) {
 	}
 }
 
+func TestCheckObjectsThrottledTimeout(t *testing.T) {
```
jon-whit, line 521:
```suggestion
func TestCheckDispatchThrottledTimeout(t *testing.T) {
```

pkg/server/server_test.go
```diff
@@ -518,6 +518,77 @@ func TestThreeProngThroughVariousLayers(t *testing.T) {
...
+	})
+
+	_, ds, _ := util.MustBootstrapDatastore(t, "memory")
+	s := MustNewServerWithOpts(
+		WithDatastore(ds),
+		WithDispatchThrottlingCheckResolverFrequency(5*time.Millisecond),
+		WithDispatchThrottlingCheckResolverEnabled(true),
+		WithDispatchThrottlingCheckResolverThreshold(5),
+	)
+	t.Cleanup(func() {
```
jon-whit, line 533:
```suggestion
	t.Cleanup(s.Close)
```

pkg/server/server_test.go
```diff
@@ -518,6 +518,77 @@ func TestThreeProngThroughVariousLayers(t *testing.T) {
...
+		Name: "openfga-test",
+	})
+	require.NoError(t, err)
+
+	storeID := createStoreResp.GetId()
+
+	model := testutils.MustTransformDSLToProtoWithID(`model
+	schema 1.1
+
+type user
```
jon-whit, line 547: Whitespace please. This is hard to read. It should be inline with the defining body above it.

pkg/server/server_test.go
```diff
@@ -518,6 +518,77 @@ func TestThreeProngThroughVariousLayers(t *testing.T) {
...
+				tuple.NewTupleKey("group:x", "member", "group:6#member"),
+				tuple.NewTupleKey("group:x", "member", "group:7#member"),
+				tuple.NewTupleKey("group:x", "member", "group:8#member"),
+				tuple.NewTupleKey("group:x", "member", "group:9#member"),
+				tuple.NewTupleKey("group:x", "member", "group:10#member"),
+			},
+		},
+	})
+	require.NoError(t, err)
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Millisecond)
```
jon-whit, line 581: Seems this test has been written largely based on luck - the assumption is that the 10 dispatches for the 10 different groups causes > 20ms delay.

I may advise that we just mock the storage layer and eliminate the memory datastore in this case so it's more of a controlled test.

We should set dispatch throttling threshold to 1, and mock the storage read(s) with a bit of latency, and then assert the throttled error. This way it is more deterministic and not so co-dependent on time.

jon-whit (changes requested):

pkg/server/server_test.go
adriantam: This actually is not based on luck.  
We know that this will take 10 dispatches will means that there are 5 dispatch being throttled.  Each throttling will take at least 5 milliseconds.  Therefore, the whole request must take at least 25 milliseconds to complete (5 * 5ms = 25ms).  Hence, this is always less than 20ms and must have timeout.

Perhaps I will update some comments to make it very clear.

internal/server/config/config.go
```diff
@@ -35,6 +35,8 @@ const (
 	DefaultDispatchThrottlingEnabled   = false
 	DefaultDispatchThrottlingFrequency = 10 * time.Microsecond
 	DefaultDispatchThrottlingThreshold = 100
+
+	DefaultRequestTimeout = 3 * time.Second
```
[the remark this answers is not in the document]
adriantam: Actually, it should be the other way.  RequestTimeout needs to be lower than UpstreamTimeout because we want the RequestTimeout to hit first.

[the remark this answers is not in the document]
adriantam: We can change this to 4750 * time.Millisecond as DefaultRequestTimeout or bump the HTTP.UpstreamTimeout to 5.25s and let DefaultRequestTimeout to 5s.

pkg/server/server.go
adriantam: Based on comment below, we don't need separate function to handle this and can be tested inline

pkg/server/server.go
```diff
@@ -712,6 +712,21 @@ func (s *Server) Write(ctx context.Context, req *openfgav1.WriteRequest) (*openf
...
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+	var internalError serverErrors.InternalError
+
+	ok := errors.As(err, &internalError)
+	if ok {
+		return errors.Is(internalError.Internal(), context.DeadlineExceeded)
```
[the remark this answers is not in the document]
adriantam: Removed the internal error part.

cmd/run/run.go
```diff
@@ -525,11 +550,9 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 	}()
 	s.Logger.Info(fmt.Sprintf("grpc server listening on '%s'...", config.GRPC.Addr))
 
+	runtime.DefaultContextTimeout = defaultContextTimeout(config)
```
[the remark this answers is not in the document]
adriantam: move back to original spot.

pkg/server/server_test.go
jon-whit: But that's still based on the tick frequency which we know doesn't necessarily tick super accurately below 10ms range. So the assumption that 5ms overhead is plenty is a bit of an assumption. That's what I mean by "luck".

But it's ok, no need to sweat it. If we see flakiness we can adjust accordingly.

cmd/run/run.go
```diff
@@ -303,14 +304,10 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
...
+}
 
-	s.Logger.Info(fmt.Sprintf("🧪 experimental features enabled: %v", config.Experimentals))
-
-	var experimentals []server.ExperimentalFeatureFlag
-	for _, feature := range config.Experimentals {
-		experimentals = append(experimentals, server.ExperimentalFeatureFlag(feature))
-	}
-
+func (s *ServerContext) datastoreConfig(config *serverconfig.Config) (storage.OpenFGADatastore, error) {
```
adriantam, line 310: need to split them out from Run as we are now failing lint due to
```
cmd/run/run.go:280:1: cyclomatic complexity 59 of func `(*ServerContext).Run` is high (> 58) (gocyclo)
func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) error {
```

pkg/middleware/timeout_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func TestNewUnaryTimeoutInterceptor(t *testing.T) {
+	timeoutInterceptor := TimeoutInterceptor{
+		timeout: 5 * time.Millisecond,
+		logger:  logger.NewNoopLogger(),
+	}
+
+	handler := func(ctx context.Context, req any) (any, error) {
+		goFuncInit := make(chan struct{})
```
jon-whit, line 46:
```suggestion
		select {
		case <-time.After(20 * time.Millisecond):
			return nil, nil
		case <-ctx.Done():
			return nil, ctx.Err()
		}
```
Here and below in the streaming test, all of the code in this grpc handler can be simplified to the simple snippet above.

pkg/middleware/timeout_test.go
```diff
@@ -0,0 +1,94 @@
...
+		select {
+		case <-ctx.Done():
+			return nil, nil
+		case <-goFuncDone:
+			return nil, fmt.Errorf("should have timeout")
+		}
+	}
+	interceptor := timeoutInterceptor.NewUnaryTimeoutInterceptor()
+	_, err := interceptor(context.Background(), nil, nil, handler)
+	require.NoError(t, err)
```
jon-whit, line 64:
```suggestion
	require.ErrorIs(t, err, context.DeadlineExceeded)
```
with the suggestion above now we can assert that the timeout middleware returns the actual context error.

pkg/server/errors/errors.go
```diff
@@ -26,6 +26,7 @@ var (
 	MismatchObjectType                     = status.Error(codes.Code(openfgav1.ErrorCode_query_string_type_continuation_token_mismatch), "The type in the querystring and the continuation token don't match")
 	RequestCancelled                       = status.Error(codes.Code(openfgav1.InternalErrorCode_cancelled), "Request Cancelled")
 	RequestDeadlineExceeded                = status.Error(codes.Code(openfgav1.InternalErrorCode_deadline_exceeded), "Request Deadline Exceeded")
+	ThrottledTimeout                       = status.Error(codes.Code(openfgav1.UnprocessableContentErrorCode_throttled_timeout_error), "Request Deadline Exceeded while throttled")
```
jon-whit, line 29: This message is inconsistent with our docs. I suggest the following rename and update the API doc to align with it:

> request deadline exceeded due to throttling

Screenshot 2024-04-25 at 1 47 08 PM

jon-whit:
@adriantam we have a bug with the config.

* If you set `--request-timeout=20ms` and leave all other configs at their defaults, then the server starts up just fine.

* If you set `--http-upstream-timeout=20ms` and leave all other configs at their defaults, then you get an error because `http-upstream-timeout` has to be greater than `listObjects-deadline`.

If the resolved timeout value between the `--request-timeout` and `--http-upstream-timeout` is less than `listObjects-deadline` then you should get an error. The ListObjects deadline should always be shorter than the timeout of the request.

internal/server/config/config.go
```diff
@@ -333,9 +349,35 @@ func (cfg *Config) Verify() error {
 		}
 	}
 
+	if cfg.RequestTimeout < 0 {
+		return errors.New("request duration must be non-negative time duration")
+	}
+
+	if cfg.RequestTimeout == 0 && cfg.HTTP.Enabled && cfg.HTTP.UpstreamTimeout < 0 {
+		return errors.New("http upstream timeout must be non-negative time duration")
```
jon-whit, line 357: Sorry, I didn't notice this previously, but in the error messages above we use the actual config value in the error message. I suggest we do the same here and below and strive to remain consistent.

> http.upstreamTimeout must be a non-negative time duration

internal/server/config/config.go
```diff
@@ -333,9 +349,35 @@ func (cfg *Config) Verify() error {
...
+	}
+
+	if cfg.ListObjectsDeadline < 0 {
+		return errors.New("list objects deadline must be non-negative time duration")
```
jon-whit, line 361:
```suggestion
		return errors.New("listObjectsDeadline must be non-negative time duration")
```

internal/server/config/config.go
```diff
@@ -333,9 +349,35 @@ func (cfg *Config) Verify() error {
 		}
 	}
 
+	if cfg.RequestTimeout < 0 {
+		return errors.New("request duration must be non-negative time duration")
```
jon-whit, line 353:
```suggestion
		return errors.New("requestTimeout must be a non-negative time duration")
```

PR #1555, opened 2024-04-19
List users benchmark

miparnisari:
## Description
This PR is very similar to the benchmarks for ListObjects (https://github.com/openfga/openfga/blob/7bb7fe40ab28610994cfdd75744edb0ae6523d8e/pkg/server/test/list_objects.go#L620-L675) introduced in https://github.com/openfga/openfga/pull/1067/files

## References

Parent PR https://github.com/openfga/openfga/pull/1433

Also see https://github.com/openfga/openfga/pull/1569

## Results

```
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/pkg/server
BenchmarkOpenFGAServer
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/one_result_without_conditions
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/one_result_without_conditions-12         	   10000	   1458954 ns/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/all_results_without_conditions
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/all_results_without_conditions-12        	      52	 219343181 ns/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/all_results_with_conditions
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/all_results_with_conditions-12           	      27	 434163488 ns/op
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/exclusion_without_conditions
BenchmarkOpenFGAServer/BenchmarkMemoryDatastore/BenchmarkListUsers/exclusion_without_conditions-12          	   18034	   2146579 ns/op
    server_test.go:522: found unexpected goroutines:
        [Goroutine 21340 in state chan send, 1 minutes, with github.com/openfga/openfga/pkg/server/commands/listusers.(*listUsersQuery).expandDirect on top of the stack:
```


pkg/server/test/list_users.go
```diff
@@ -0,0 +1,93 @@
...
+	// so that we assert that we don't waste time computing extra results in that case
+	oneResultIterations := int(math.Inf(1))
+	var allResultsIterations int
+
+	b.Run("allResults", func(b *testing.B) {
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			ctx = typesystem.ContextWithTypesystem(ctx, typesystem.New(model))
+
+			listUsersQuery := listusers.NewListUsersQuery(ds,
```
miparnisari, line 82: testing against the server itself creates a circular import error

pkg/server/test/list_users.go
```diff
@@ -0,0 +1,93 @@
...
+				define parent: [folder]
+				define can_view: viewer or viewer from parent`)
+	err := ds.WriteAuthorizationModel(context.Background(), storeID, model)
+	require.NoError(b, err)
+
+	numberOfUsersWithAccess := 0
+	for i := 0; i < 100; i++ {
+		var tuples []*openfgav1.TupleKey
+
+		for j := 0; j < ds.MaxTuplesPerWrite(); j++ {
```
jon-whit, line 43: What is `j` needed for? 

Did you intent to do `strconv.Itoa(j)` below?

pkg/server/test/list_users.go
```diff
@@ -0,0 +1,93 @@
...
+
+	req := &openfgav1.ListUsersRequest{
+		StoreId:              store,
+		AuthorizationModelId: model.GetId(),
+		Object:               &openfgav1.Object{Type: "document", Id: "1"},
+		Relation:             "can_view",
+		UserFilters:          []*openfgav1.ListUsersFilter{{Type: "user"}},
+	}
+
+	// TODO add benchmark for when max_results = 1
```
jon-whit, line 72: This is a reminder to add the corresponding `oneResult` sub-benchmark yeah?

pkg/server/test/list_users.go
```diff
@@ -0,0 +1,93 @@
...
+	// so that we assert that we don't waste time computing extra results in that case
+	oneResultIterations := int(math.Inf(1))
```
jon-whit, line 74:
```suggestion
	var oneResultIterations, allResultsIterations int

	b.Run("oneResult", func(b *testing.T) {
		oneResultIterations = int(math.Inf(1))
	})
```
Just to wire it up with more context in the scope of this PR.

pkg/server/test/list_users.go
miparnisari: 🤦‍♀️ yes, thank you

willvedd (commented):
Matching what we have for ListObjects is a good start but it doesn't give us a good idea of the performance characteristics specific to ListUsers. Do you have a sense of other scenarios that you want to benchmark?

Some ideas I had:
- Deeply nested w/ and w/o many tuples
- Shallow w/ and w/o many tuples
- Models w/ and w/o wildcards
- Models w/ and w/o negation

Or said a different way, there are several attributes of a situation that ListUsers may or may not handle well including depth, breadth, quantity of tuples, usage of wildcards, usage of negation, etc. 

It would be good to have an understanding of that within the Benchmarks.

miparnisari:
@willvedd we can definitely add more later! For now this covers the most important case of not doing more work than we should (after we implement the `--max-results` flag), because that brings reliability concerns. Every other improvement that we can make will be a nice to have

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -52,6 +53,12 @@ func WithListUsersQueryLogger(l logger.Logger) ListUsersQueryOption {
 	}
 }
 
+func WithListUserMaxResults(max uint32) ListUsersQueryOption {
+	return func(rq *listUsersQuery) {
+		rq.maxResults = max
+	}
+}
```
willvedd, line 60: **Question:** Will the rest of the max results be implemented in a future PR?

miparnisari: Yes, I will start that today

pkg/server/test/list_users.go
```diff
@@ -0,0 +1,98 @@
...
+		Relation:             "can_view",
+		UserFilters:          []*openfgav1.ListUsersFilter{{Type: "user"}},
+	}
+
+	var oneResultIterations, allResultsIterations int
+
+	b.Run("oneResult", func(b *testing.B) {
+		oneResultIterations = int(math.Inf(1))
+		// TODO add benchmark for when max_results = 1
+		// so that we assert that we don't waste time computing extra results in that case
```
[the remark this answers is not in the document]
miparnisari: I wouldn't say it's blocked, but I need to add this at some point

pkg/server/test/list_users.go
jon-whit: Got it 👍  - that wasn't clear to me. Thank you.

jon-whit:
@miparnisari @willvedd one thing that is hard for me to follow with the benchmark changes in this PR is the actual benchmark scenario that we're testing w.r.t tuple structure and model. This is because we have a single bootstrapped model and tuples but different variations of queries against it.

I may suggest that instead of trying to write one big model and tuples to "rule them all" - maybe we could break out individual benchmark cases that have a nice unique name and a bootstrapped model and tuples and query for that specific scenario. It would be easier to get a grasp on what is being benchmarked instead of having to cross-reference back to the global model and tuples etc.. Essentially making each benchmark self-contained.

```go
func BenchmarkListUsers(b *testing.B) {
  b.Run("deeply_nested_union", func(...) {
    WriteAuthorizationModel(...)
    Write(..., tuples)
    ListUsers(...)
  })

  b.Run("widely_nested_intersection", func(...) {
    WriteAuthorizationModel(...)
    Write(..., tuples)
    ListUsers(...)
  })

  b.Run("exclusion_with_wildcards", func(...) {
    WriteAuthorizationModel(...)
    Write(..., tuples)
    ListUsers(...)
  })
}
```
So on and so forth..

For the other scenarios that @willvedd mentioned this would also cater nicely to that, because each benchmark will have different tuple structures and that will be more obvious and clear by virtue of the benchmarks being self-contained.

pkg/server/test/list_users.go
```diff
@@ -0,0 +1,199 @@
...
+					i++
+				}
+				err = ds.Write(context.Background(), storeID, nil, tuplesToWrite)
+				require.NoError(b, err)
+			}
+
+			b.ResetTimer()
+
+			for i := 0; i < b.N; i++ {
+				bm.inputRequest.StoreId = storeID
```
jon-whit, line 182: Any reason why we're including the setup code in the benchmark run?

Shouldn't we setup the `NewListUsersQuery` externally and then just benhcmark the `ListUsers` call directly?

miparnisari: 💀 🤦‍♀️ no reason.. moved it.

jon-whit (approved):

PR #1558, opened 2024-04-20
chore: update grafana dashboard metrics

miparnisari:
## Description

Update metrics shown on Grafana dashboard. 

- I removed two "process" metrics because I don't find them useful.
- I added graphs for:
   - Check query cache hit ratio 
   - SQL metrics
   - RPC response codes

Before:
before

After: 
after


jon-whit:
Nice additions!

jon-whit (approved):

PR #1563, opened 2024-04-22
fix: stricter validation during ListObjects read of tuples

miparnisari:
## Description
In ListObjects, when reading tuples, we should skip over all the invalid ones, and the definition of "invalid" can't just be "invalid condition".

## Testing

With this model

```
model
  schema 1.1

  type user

  type group
    relations
```

and these tuples

```
group:fga#member@user:maria
group:eng#member#member@group:fga#member // invalid
```

|                                                            | Before                              | After          |
|------------------------------------------------------------|-------------------------------------|----------------|
| ListObjects(type:group, relation:member, user: user:maria) | [ "group:fga", "group:eng#member" ] | ["group:fga" ] |

### DIFF
--- a/internal/validation/validation.go
+++ b/internal/validation/validation.go
@@ -62,7 +62,7 @@ func ValidateTuple(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) error
 			return &tuple.InvalidTupleError{Cause: err, TupleKey: tk}
 		}
 
-		if err := ValidateCondition(typesys, tk); err != nil {
+		if err := validateCondition(typesys, tk); err != nil {
 			return err
 		}
 	}
@@ -176,8 +176,9 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
 	return fmt.Errorf("type '%s' is not an allowed type restriction for '%s#%s'", userType, objectType, tk.GetRelation())
 }
 
-// ValidateCondition enforces conditions on a relationship tuple
-func ValidateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) error {
+// validateCondition returns an error if the condition of the tuple is required but not present,
+// or if the tuple provides a condition but it is invalid according to the model.
+func validateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) error {
 	objectType := tuple.GetType(tk.GetObject())
 	userType := tuple.GetType(tk.GetUser())
 	userRelation := tuple.GetRelation(tk.GetUser())
@@ -259,10 +260,7 @@ func ValidateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) e
 	return nil
 }
 
-// FilterInvalidTuples implements the TupleFilterFunc signature and can be used to provide
-// a generic filtering mechanism when reading tuples. It is particularly useful to filter
-// out tuples that aren't valid according to the provided model, which can help filter
-// tuples that were introduced due to another authorization model.
+// FilterInvalidTuples filters out tuples that aren't valid according to the provided model.
 func FilterInvalidTuples(typesys *typesystem.TypeSystem) storage.TupleKeyFilterFunc {
 	return func(tupleKey *openfgav1.TupleKey) bool {
 		err := ValidateTuple(typesys, tupleKey)
--- a/pkg/server/commands/list_objects_test.go
+++ b/pkg/server/commands/list_objects_test.go
@@ -140,7 +140,7 @@ func TestListObjectsDispatchCount(t *testing.T) {
 				define member: [user, group#member]
 			`,
 			tuples: []*openfgav1.TupleKey{
-				tuple.NewTupleKey("group:eng#member", "member", "group:fga#member"),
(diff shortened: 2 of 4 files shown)


internal/validation/validation.go
```diff
@@ -176,8 +176,9 @@ func validateTypeRestrictions(typesys *typesystem.TypeSystem, tk *openfgav1.Tupl
 	return fmt.Errorf("type '%s' is not an allowed type restriction for '%s#%s'", userType, objectType, tk.GetRelation())
 }
 
-// ValidateCondition enforces conditions on a relationship tuple
-func ValidateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) error {
+// validateCondition returns an error if the condition of the tuple is required but not present,
+// or if the tuple provides a condition but it is invalid according to the model.
+func validateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) error {
```
sergiught, line 181: 🤔 (future suggestion): Perhaps it's time to improve the package design for `typesystem`? It would feel more natural to have something like `type.System` instead IMO.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -474,9 +474,7 @@ func (c *ReverseExpandQuery) readTuplesAndExecute(
 	// filter out invalid tuples yielded by the database iterator
 	filteredIter := storage.NewFilteredTupleKeyIterator(
 		storage.NewTupleKeyIteratorFromTupleIterator(iter),
-		func(tupleKey *openfgav1.TupleKey) bool {
-			return validation.ValidateCondition(c.typesystem, tupleKey) == nil
-		},
+		validation.FilterInvalidTuples(c.typesystem),
```
sergiught, line 477: (suggestion): Would this make more sense to have on the tuple package? `tuple.FilterInvalid`

Validation should be a func that's provide for a given package IMO. 

No need to change this now, but something to keep in mind as we slowly improve our package design.

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -356,3 +356,90 @@ type document
 		}
 	}
 }
+
+func TestReverseExpandIgnoresInvalidTuples(t *testing.T) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+
+	store := ulid.Make().String()
```
sergiught, line 365:
```suggestion
	storeID := ulid.Make().String()
```

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -356,3 +356,90 @@ type document
...
+	model := testutils.MustTransformDSLToProtoWithID(`
+		model
+			schema 1.1
+		type user
+		type group
+			relations
+				define member: [user, group#member]`)
+
+	mockController := gomock.NewController(t)
+	defer mockController.Finish()
```
sergiught, line 376: `t.Cleanup` ?

pkg/server/commands/reverseexpand/reverse_expand_test.go
```diff
@@ -356,3 +356,90 @@ type document
...
+
+	for {
+		select {
+		case res, open := <-resultChan:
+			if open {
+				results = append(results, res.Object)
+			} else {
+				require.ElementsMatch(t, []string{"group:fga"}, results)
+				return
+			}
```
sergiught, line 437:
```suggestion
			if !open {
				require.ElementsMatch(t, []string{"group:fga"}, results)
				return
			} 
			
			results = append(results, res.Object)
```

sergiught (dismissed):
Looks good, mainly just offering some suggestions.

internal/validation/validation.go
miparnisari: 🤷‍♀️ that'd be a huge breaking change, changing the package name

pkg/server/commands/reverseexpand/reverse_expand.go
miparnisari: Yes, it would make more sense in `tuple` but i won't change it now

pkg/server/commands/list_objects_test.go
```diff
@@ -140,7 +140,7 @@ func TestListObjectsDispatchCount(t *testing.T) {
 				define member: [user, group#member]
 			`,
 			tuples: []*openfgav1.TupleKey{
-				tuple.NewTupleKey("group:eng#member", "member", "group:fga#member"),
+				tuple.NewTupleKey("group:eng", "member", "group:fga#member"),
```
jon-whit, line 143: Why is this an invalid tuple? The model allows `group#member` to be related to `group#member`.

internal/validation/validation.go
jon-whit: We've had some things on our list of things to improve in the Typesystem package, and we have yet to get around to it. But if we're going to do this let's start with what we want to improve and why and propose this as a maintainer team, and then we can evaluate how that fits into our roadmap and the scope of the effort.

jon-whit (approved):

pkg/server/commands/list_objects_test.go
miparnisari: Yes, but the `object` field cannot be `group:eng#member`
image

PR #1573, opened 2024-04-25
fix: `ListUsers` handling of cycles in exclusion handler

willvedd:
## Description
As noted in a [recent CVE advisory](https://github.com/openfga/openfga/security/advisories/GHSA-8cph-m685-6v6r), the check resolution process may have been improperly handling cycles when performing exclusion (`but not`). Previously, it was effectively treating all cycles as `allowed=false`, however this isn't correct in all cases. When the right-side of the `but not` operator yields a cycle, we should _not_ treat that as a falsey outcome because it may unintentionally authorize the check request. This PR fixes this issue by propagating the detection of cycles up the `expand` call chain and in cases of exclusion, use that detected cycle to prevent unintentional authorization.

## References
Related CVE: https://github.com/openfga/openfga/security/advisories/GHSA-8cph-m685-6v6r

## Review Checklist

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -380,7 +380,7 @@ func (l *listUsersQuery) expandDirect(
 	pool.WithMaxGoroutines(int(l.resolveNodeBreadthLimit))
 
 	var errs error
-	var hasCycle bool
+	var hasCycle atomic.Bool
 LoopOnIterator:
 	for {
 		tupleKey, err := filteredIter.Next(ctx)
@@ -433,7 +433,7 @@ LoopOnIterator:
 			rewrittenReq.Relation = userRelation
 			resp := l.expand(ctx, rewrittenReq, foundUsersChan)
 			if resp.hasCycle {
-				hasCycle = true
+				hasCycle.Store(true)
 			}
 			return resp.err
 		})
@@ -445,7 +445,7 @@ LoopOnIterator:
 	}
 	return expandResponse{
 		err:      errs,
-		hasCycle: hasCycle,
+		hasCycle: hasCycle.Load(),
 	}
 }
 
@@ -547,14 +547,15 @@ func (l *listUsersQuery) expandExclusion(
 	baseFoundUsersCh := make(chan *openfgav1.User, 1)
 	subtractFoundUsersCh := make(chan *openfgav1.User, 1)
 
-	var baseError, substractError error
-	var baseHasCycle, subtractHasCycle bool
+	var baseError error
 	go func() {
 		resp := l.expandRewrite(ctx, req, rewrite.Difference.GetBase(), baseFoundUsersCh)
 		baseError = resp.err
-		baseHasCycle = resp.hasCycle
 		close(baseFoundUsersCh)
 	}()
+
+	var substractError error
+	var subtractHasCycle bool
 	go func() {
 		resp := l.expandRewrite(ctx, req, rewrite.Difference.GetSubtract(), subtractFoundUsersCh)
 		substractError = resp.err
@@ -573,7 +574,12 @@ func (l *listUsersQuery) expandExclusion(
 		subtractFoundUsersMap[key] = struct{}{}
 	}
 
-	if baseHasCycle || subtractHasCycle {
+	if subtractHasCycle {
+		// Because exclusion contains the only bespoke treatment of
+		// cycle, everywhere else we consider it a falsey outcome.
+		// Once we make a determination within the exclusion handler, we're
+		// able to properly handle the case and do not need to propagate
+		// the existence of a cycle to an upstream handler.
 		return expandResponse{
 			err: nil,
 		}
--- a/pkg/server/commands/listusers/list_users_rpc_test.go
+++ b/pkg/server/commands/listusers/list_users_rpc_test.go
@@ -713,6 +713,56 @@ func TestListUsersCycles(t *testing.T) {
 			expectedUsers: []string{},
 		},
 		{
+			name: "cycle_and_union",
+			req: &openfgav1.ListUsersRequest{
+				Object:   &openfgav1.Object{Type: "document", Id: "1"},
+				Relation: "can_view",
+				UserFilters: []*openfgav1.UserTypeFilter{
+					{
+						Type: "user",
+					},
+				},
+			},
+			model: `model
+			schema 1.1
+			type user
+			type document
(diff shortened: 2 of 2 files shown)


pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2252,8 +2252,8 @@ func TestListUsersCycleDetection(t *testing.T) {
 				},
 				visitedUsersetsMap: visitedUsersets,
 			}, channelWithResults)
-			if err != nil {
-				channelWithError <- err
+			if resp.err != nil {
+				channelWithError <- resp.err
 				return
 			}
 			channelDone <- struct{}{}
```
jon-whit, line 2259: Can we add this test as well please? It makes sure that if we experience cycles through independent object_ids and with a non-stratifiable resolution on the `document:1#viewer` and `document:2#blocked` relationship then we still yield the residual relationships for `user:jon`.

```go
		{
			name: "exclusion_with_chained_negation",
				Object:   &openfgav1.Object{Type: "document", Id: "2"},
				Relation: "viewer",
				UserFilters: []*openfgav1.UserTypeFilter{
					{
						Type: "user",
					},
				},
			},
			model: `model
			  schema 1.1

			type user

			type document
			  relations
			    define unblocked: [user]
				define blocked: [user, document#viewer] but not unblocked
				define viewer: [user, document#blocked] but not blocked
			`,
				tuple.NewTupleKey("document:1", "viewer", "document:2#blocked"),
				tuple.NewTupleKey("document:2", "blocked", "document:1#viewer"),
				tuple.NewTupleKey("document:2", "viewer", "user:jon"),
				tuple.NewTupleKey("document:2", "unblocked", "user:jon"),
			},
			expectedUsers: []string{"user:jon"},
		},
```

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -1507,7 +1507,7 @@ func TestListUsersExclusion(t *testing.T) {
 				tuple.NewTupleKey("group:1", "blocked", "group:1#member"),
```
jon-whit, line 1507: Can we add this test as well please? It makes sure that if we experience cycles through independent object_ids and with a non-stratifiable resolution on the `document:1#viewer` and `document:2#blocked` relationship then we still yield the residual relationships for `user:jon`.

```go
		{
			name: "exclusion_with_chained_negation",
				Object:   &openfgav1.Object{Type: "document", Id: "2"},
				Relation: "viewer",
				UserFilters: []*openfgav1.UserTypeFilter{
					{
						Type: "user",
					},
				},
			},
			model: `model
			  schema 1.1

			type user

			type document
			  relations
			    define unblocked: [user]
				define blocked: [user, document#viewer] but not unblocked
				define viewer: [user, document#blocked] but not blocked
			`,
				tuple.NewTupleKey("document:1", "viewer", "document:2#blocked"),
				tuple.NewTupleKey("document:2", "blocked", "document:1#viewer"),
				tuple.NewTupleKey("document:2", "viewer", "user:jon"),
				tuple.NewTupleKey("document:2", "unblocked", "user:jon"),
			},
			expectedUsers: []string{"user:jon"},
		},
```

jon-whit: Please add this test as well, and it is currently failing. This is a tricky corner case involving non-stratifiable queries through exclusion.

```
{
	name: "non_stratifiable_exclusion_containing_cycle",
		Object:   &openfgav1.Object{Type: "document", Id: "1"},
		Relation: "viewer",
		UserFilters: []*openfgav1.UserTypeFilter{
			{
				Type:     "document",
				Relation: "blocked",
			},
		},
	},
	model: `model
		schema 1.1

	type user

	type document
		relations
		define blocked: [user, document#viewer]
		define viewer: [user, document#blocked] but not blocked
	`,
		tuple.NewTupleKey("document:1", "viewer", "document:2#blocked"),
		tuple.NewTupleKey("document:2", "blocked", "document:1#viewer"),
	},
	expectedUsers: []string{},
},
```

For the same `Check(document:1#viewer@document:2#blocked)` this should return `{allowed: false}` as it is a non-stratifable query.

willvedd: This is a good point. The missing piece was handling the base-cycle case. There are zero situations of negation where a cycle can return a result in ListUsers.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -549,6 +573,12 @@ func (l *listUsersQuery) expandExclusion(
 		subtractFoundUsersMap[key] = struct{}{}
 	}
 
+	if baseHasCycle || subtractHasCycle {
+		return expandResponse{
+			err: nil,
+		}
+	}
```
miparnisari, line 580: Can you explain in code comments why this is not

```go
if baseHasCycle || subtractHasCycle {
		return expandResponse{
		        hasCycle: true,
			err: nil,
		}
	}
```

?

willvedd: Because exclusion contains the only bespoke treatment of cycle, everywhere else we consider it a falsey outcome. Once we make a determination within the exclusion handler, we're able to properly handle the case and do not need to propagate the existence of a cycle to an upstream handler.

miparnisari: > in code comments

😄 so we don't forget

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -415,24 +431,30 @@ LoopOnIterator:
 			rewrittenReq := req.clone()
 			rewrittenReq.Object = &openfgav1.Object{Type: userObjectType, Id: userObjectID}
 			rewrittenReq.Relation = userRelation
-			return l.expand(ctx, rewrittenReq, foundUsersChan)
+			resp := l.expand(ctx, rewrittenReq, foundUsersChan)
+			if resp.hasCycle {
+				hasCycle = true
+			}
```
miparnisari, line 437: nit: can simplify `hasCycle = hasCycle || resp.hasCycle`

willvedd: Interesting enough, that causes a data race.

miparnisari: Oh! You have the data race even without my suggestion. `var hasCycle bool` should be an `atomic.Bool`

Issue #827, opened 2023-06-27
`trace_id` and `request_id` should be equal


miparnisari:
## Description 

Right now, to debug a particular request, you have to:

1. grab the request_id from the response's header
2. look into the logs and grab the trace id
3. look into the traces

I think it would be better if `request_id == (top-level) trace_id` so that we can skip step 2.  This is also how Okta FGA works.

## Tasks

1. Add this to the existing `requestid` middleware:

```go
// InitializeTraceID will add the trace ID to the context tags so that they can be picked up by other middlewares.
// If trace is not enabled, it will use generate its own ULID as request ID.
func InitializeTraceID(ctx context.Context, traceID trace.TraceID) context.Context {
	if !traceID.IsValid() {
		id := ulid.Make().String()
        grpc_ctxtags.Extract(r.ctx).Set("request-id", id)
	} else {
        grpc_ctxtags.Extract(r.ctx).Set("request-id", traceID.String())
	}
	return ctx
}
```

2. Modify the `reportable` function in the `requestid` middleware to use ☝️ 

3. Cleanup the custom code that logs `trace_id` and `request_id`:

https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/middleware/logging/logging.go#L131-L138
because, by virtue of putting the request id as a `context tag`, it should be automatically picked up here:

https://github.com/openfga/openfga/blob/164963facc61241e1378d8fefb6632c700407163/pkg/middleware/logging/logging.go#L57-L59

4. Following the test introduced in https://github.com/openfga/openfga/pull/926, modify it so that we assert that the `request_id` field is logged.

miparnisari:
This was implemented in v1.5.4.

With tracing: "trace_id": "1e20da43269fe07e3d2ac018c0aad2d1",  "request_id": "1e20da43269fe07e3d2ac018c0aad2d1"
(same values)

Without tracing: "request_id": "38fee7ac-4bfe-4cf6-baa2-8b5ec296b485"
(uuid)

PR #1576, opened 2024-04-26
chore(logging): make request_id equal to trace_id to facilitate debug

00chorch:
chore(logging): make request_id equal to trace_id to facilitate debug

## Description

Allows to remove step 2 of the current trace debugging mechanism:
1. grab the request_id from the response's header
2. look into the logs and grab the trace id
3. look into the traces

## References

- https://github.com/openfga/openfga/issues/827

## Review Checklist
- [] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]


pkg/middleware/requestid/requestid.go
```diff
@@ -13,13 +14,24 @@ import (
 
 const (
 	requestIDCtxKey   = "request-id-context-key"
+	requestIDKey 			= "request_id"
```
miparnisari, line 17: Please run `make lint` to fix linting issues

pkg/middleware/requestid/requestid.go
```diff
@@ -45,15 +57,15 @@ func NewStreamingInterceptor() grpc.StreamServerInterceptor {
 
 func reportable() interceptors.CommonReportableFunc {
 	return func(ctx context.Context, c interceptors.CallMeta) (interceptors.Reporter, context.Context) {
-		id, _ := uuid.NewRandom()
-		requestID := id.String()
+		requestID := InitID(ctx)
 
 		ctx = metadata.AppendToOutgoingContext(ctx, requestIDCtxKey, requestID)
```
miparnisari, line 62: I don't think we need this line, nor `requestIDCtxKey`. Can you remove them?

pkg/middleware/requestid/requestid.go
```diff
@@ -13,13 +14,24 @@ import (
...
 )
 
+// InitID returns the ID to be used to identify the request.
+// If trace is enabled, returns trace ID; otherwise returns a new ULID.
+func InitID(ctx context.Context) string {
+	spanCtx := trace.SpanContextFromContext(ctx)
+	if spanCtx.TraceID().IsValid() {
+		return spanCtx.TraceID().String()
+	} 
+	return ulid.Make().String()
```
miparnisari, line 32: @openfga/backend what do you think, should we use `ulid` or `UUID` for request IDs? From https://github.com/oklog/ulid

> If you don't care about time-based ordering of generated IDs, then there's no reason to use ULIDs! There are many other kinds of IDs that are easier, faster, smaller, etc. Consider UUIDs.

Changing this format to `ulid`:
- is a breaking change for consumers. (However, we already use `ulid` for store IDs)
- we can get rid of the `"github.com/google/uuid"` dependency.
- ULID contains information about the time, which for logging is a benefit.
- [Very low chance of collision](https://pkg.go.dev/github.com/oklog/ulid/v2#section-readme:~:text=1.21e%2B24%20unique%20ULIDs%20per%20millisecond%20(1%2C208%2C925%2C819%2C614%2C629%2C174%2C706%2C176%20to%20be%20exact))

pkg/middleware/requestid/requestid.go
00chorch: - Regarding keeping `requestIDCtxKey`, Can we keep it and have different keys for different contexts?
Distinct key names may avoid potential conflicts or confusion between different contexts and their purposes. 

It will be aligned to what's done for naming keys in storeid middleware:
https://github.com/openfga/openfga/blob/abf56641636fddef9340c4462364de1505c30442/pkg/middleware/storeid/storeid.go#L17C1-L19C35 

- Regarding keeping the line, note that currently `func InitID()` is not setting any context key/value, all key/value setting is currently centralized inside `func reportable()`.
Some alternatives I see include
a) [recommended] keeping it as it is - solution is simple & more readable
b) refactor `func initCtxWithId()` to also append id to context metadata and return the updated ctx
c) mirror store.id middleware. - IMHO this option seems to introduce too much change that might not be worth it at this PR but could be evaluated for future refactor

00chorch: Any thoughts on the above proposals?

miparnisari: This is how it works today on `main`:

1. On `requestid` we call `metadata.AppendToOutgoingContext(ctx, requestIDCtxKey, requestID)`
2. On `logging`, we call `if requestID, ok := requestid.FromContext(ctx); ok { fields = append(fields, zap.String(requestIDKey, requestID)) }`

Since you removed line 2, there is no reason to keep line 1 anymore, i.e. we don't need to keep the request id as a value in the context - that's what `grpc_ctxtags` is for.

00chorch: Thanks for clarifying that request-id should no longer be stored in ctx. Changes applied

pkg/middleware/requestid/requestid.go
00chorch: While waiting for comments, I keep thinking about the "breaking change for consumers" statement.
So far I can only think of pros and can't think of negative scenarios for consumers to justify keeping UUIDs.

I look forward to learning what OpenFGA team thinks about this.

Thank you

miparnisari: I asked the team, everyone agreed that having request IDs that are OTEL UUIDs when tracing is enabled and ULIDs when tracing is disabled is a no-go. For request IDs we need some way of getting a OTEL UIID even when tracing is off. However, the ID generator they use isn't exported (https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/id_generator.go#L60). Their code hasn't changed in years though, so my thought was that we could just ... copy it here. 🤔 🤷‍♀️

00chorch: If ULID vs UUID is still a hotly debated topic, as an alternative to unblock this PR, may I keep UUID to avoid rushing the decision on UUID vs ULID.

@miparnisari / team, any thoughts?

Thank you

miparnisari: Yes, let's do that :) 

So to summarize:

```
before this PR, with tracing:  "trace_id": "203d831d32d064b47d746223b92da657", "request_id": "cf5012eb-8b32-4e68-8a85-f296cc1a8a80",

before this PR, without tracing: "request_id": "38fee7ac-4bfe-4cf6-baa2-8b5ec296b485"

after this PR, with tracing: "trace_id": "1e20da43269fe07e3d2ac018c0aad2d1",  "request_id": "1e20da43269fe07e3d2ac018c0aad2d1"
(same values)

after this PR, without tracing: "request_id": "38fee7ac-4bfe-4cf6-baa2-8b5ec296b485"
(uuid)
```

pkg/middleware/requestid/requestid.go
```diff
@@ -12,23 +13,23 @@ import (
...
 
-// FromContext extracts the request-id from the context, if it exists.
-func FromContext(ctx context.Context) (string, bool) {
-	if md, ok := metadata.FromOutgoingContext(ctx); ok {
-		if vals := md.Get(requestIDCtxKey); len(vals) > 0 {
-			return vals[0], true
-		}
+// InitID returns the ID to be used to identify the request.
+// If trace is enabled, returns trace ID; otherwise returns a new ULID.
+func InitID(ctx context.Context) string {
```
miparnisari, line 26:
```suggestion
// InitRequestID returns the ID to be used to identify the request.
// If tracing is enabled, returns trace ID; otherwise returns a new UUID.
func InitID(ctx context.Context) string {
```

and let's remove the `// TODO use ulid library?` for now

00chorch: Good catch @miparnisari. Thank your for all your collaboration. 

I've just pushed a change that added the func rename too.

miparnisari (approved):

pkg/middleware/requestid/requestid.go
```diff
@@ -12,23 +13,24 @@ import (
 )
 
 const (
-	requestIDCtxKey   = "request-id-context-key"
+	requestIDKey      = "request_id"
 	requestIDTraceKey = "request_id"
 
-	// RequestIDHeader defines the HTTP header that is set in each HTTP response
+	// RequestIDHeader defines the HTTP header that is set in each HTTP response.
```
miparnisari, line 19: @00chorch remove the final dot to fix the linter complaining :)

miparnisari (approved):

PR #1577, opened 2024-04-27
fix(list-users): leaking goroutines due to blocked sending to channel

miparnisari:


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -664,3 +665,21 @@ func enteredCycle(req *internalListUsersRequest) bool {
 	req.visitedUsersetsMap[key] = struct{}{}
 	return false
 }
+
+func (l *listUsersQuery) buildResultsChannel() chan *openfgav1.User {
+	foundUsersCh := make(chan *openfgav1.User, 1)
```
willvedd, line 670: Setting the channel size to one would block the channel from receiving until the single result was processed,  creating an unnecessary performance bottleneck. We should probably set this value much higher in cases where there is no results limit.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -664,3 +665,21 @@ func enteredCycle(req *internalListUsersRequest) bool {
...
+func (l *listUsersQuery) buildResultsChannel() chan *openfgav1.User {
+	foundUsersCh := make(chan *openfgav1.User, serverconfig.DefaultListUsersMaxResults)
+	maxResults := l.maxResults
+	if maxResults > 0 {
+		foundUsersCh = make(chan *openfgav1.User, maxResults)
+	}
+	return foundUsersCh
+}
+
+func (l *listUsersQuery) trySendResult(ctx context.Context, user *openfgav1.User, foundUsersCh chan<- *openfgav1.User) {
```
willvedd, line 678: The `listUsersQuery` isn't really aiding this function. Instead, can we introduce `type foundUserChannel chan<- *openfgav1.User` and use that as the receiver type in the method? 

```suggestion
func (l *foundUsersChan) trySendResult(ctx context.Context, user *openfgav1.User) {
```

miparnisari: Adding a new type feels unnecessary. However, I refactored the function.

willvedd (approved):

PR #1578, opened 2024-04-27
chore: remove old parser

miparnisari:
# Description

- Remove "github.com/craigpastro/openfga-dsl-parser/v2"


willvedd (approved):
This is great and long needed. Re: migrating to ULID, it's a good idea but this is a solid N+1 on its own.

jon-whit (approved):

PR #1582, opened 2024-04-29
perf: in Check, if source and target types aren't connected, return `allowed=false`

miparnisari:
## Description
Adding a performance optimization for queries of the form `Check(user, relation, object)`. If `user` type and `object` types are disconnected/unreachable, we return early with `allowed=false`.

I achieve this by first building the graph of the model using most of the code from https://github.com/jon-whit/openfga-graphviz-gen.

## Testing

With this model

```
model
  schema 1.1

type user

type user2

type folder
  relations
    define viewer: [user2]

type doc
  relations
    define can_read: viewer or viewer from parent
    define parent: [folder]
    define viewer: [user2]
```

And this tuple: `doc:budget # parent @ folder:parent`

### Before

image

### After

image


internal/graph/check.go
```diff
@@ -615,6 +629,28 @@ func (c *LocalChecker) ResolveCheck(
...
+		if tuple.IsTypedWildcard(user) {
+			sourceRel = typesystem.WildcardRelationReference(userType)
+		} else {
+			sourceRel = typesystem.DirectRelationReference(userType, "")
+		}
+	default:
+		sourceRel = typesystem.DirectRelationReference(userType, userRelation)
+	}
+
+	edges, err := g.GetRelationshipEdges(typesystem.DirectRelationReference(objectType, relation), sourceRel)
```
elbuo8, line 647: Perhaps the fast follow is improving the `typesystem` package, particularly as this object lives in memory for a long period of time, calculating this multiple times would be waste of compute that could be incrementally cached lazily.

internal/graph/check.go
```diff
@@ -615,6 +629,28 @@ func (c *LocalChecker) ResolveCheck(
 	return resp, nil
 }
 
+// areTypesConnected returns true if it is possible to reach objectType#relation starting from userType[#userRelation].
+func (c *LocalChecker) areTypesConnected(typesys *typesystem.TypeSystem, user, userType, userRelation, objectType, relation string) (bool, error) {
+	g := New(typesys)
```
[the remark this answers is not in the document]
miparnisari: Why what? 😆

internal/graph/check.go
miparnisari: Yes, we should definitely calculate the edges up front once and not in every sub-problem resolution... however, that wasn't a trivial change so I didn't do it

jon-whit: This needs to happen before we merge this PR. Graph construction from the typesystem in critical Check resolution code is a "no go" right now IMO.

Let's fix up the typesystem, do better lazy compute of the type graph, and then let's merge this work in.

jon-whit: Instead of computing BFS each time, we should have an internal state keeping track of reachability between user/subject types and object types. Otherwise we're paying the penalty on every single request even if the subject and object are already known to be connected from previous requests.

Care should be taken to avoid recomputing information we already know here, including the BFS traversal, because otherwise we're paying that penalty on every request.

miparnisari: That penalty you are referring is an in-memory lookup which is a lot cheaper than a trip to the database.

PR #1583, opened 2024-04-30
fix(list-users): return partial results when deadline is met instead of internal error

jon-whit:
## Description
This PR changes the ListUsers behavior such that when the ListUsers deadline is met we return the results we accumulated up to that point and avoid surfacing an internal error.

## Review Checklist

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -42,6 +42,10 @@ type listUsersQuery struct {
 	maxConcurrentReads      uint32
 	deadline                time.Duration
 }
+type expandResponse struct {
+	hasCycle bool
+	err      error
+}
 
 type ListUsersQueryOption func(l *listUsersQuery)
 
@@ -178,11 +182,11 @@ func (l *listUsersQuery) ListUsers(
 
 	go func() {
 		internalRequest := fromListUsersRequest(req, &datastoreQueryCount)
-		err := l.expand(cancellableCtx, internalRequest, foundUsersCh)
+		resp := l.expand(cancellableCtx, internalRequest, foundUsersCh)
 		close(foundUsersCh)
 
-		if err != nil {
-			expandErrCh <- err
+		if resp.err != nil {
+			expandErrCh <- resp.err
 		}
 	}()
 
@@ -234,18 +238,22 @@ func (l *listUsersQuery) expand(
 	ctx context.Context,
 	req *internalListUsersRequest,
 	foundUsersChan chan<- *openfgav1.User,
-) error {
+) expandResponse {
 	ctx, span := tracer.Start(ctx, "expand")
 	defer span.End()
 	span.SetAttributes(attribute.Int("depth", int(req.depth)))
 	if req.depth >= l.resolveNodeLimit {
-		return graph.ErrResolutionDepthExceeded
+		return expandResponse{
+			err: graph.ErrResolutionDepthExceeded,
+		}
 	}
 	req.depth++
 
 	if enteredCycle(req) {
 		span.SetAttributes(attribute.Bool("cycle_detected", true))
-		return nil
+		return expandResponse{
+			hasCycle: true,
+		}
 	}
 
 	reqObjectType := req.GetObject().GetType()
@@ -269,7 +277,9 @@ func (l *listUsersQuery) expand(
 
 	typesys, err := l.typesystemResolver(ctx, req.GetStoreId(), req.GetAuthorizationModelId())
 	if err != nil {
-		return err
+		return expandResponse{
+			err: err,
+		}
 	}
 
 	targetObjectType := req.GetObject().GetType()
@@ -277,16 +287,17 @@ func (l *listUsersQuery) expand(
 
 	relation, err := typesys.GetRelation(targetObjectType, targetRelation)
 	if err != nil {
-		return err
+		return expandResponse{
+			err: err,
+		}
 	}
 
 	relationRewrite := relation.GetRewrite()
-	err = l.expandRewrite(ctx, req, relationRewrite, foundUsersChan)
-	if err != nil {
-		telemetry.TraceError(span, err)
-		return err
+	resp := l.expandRewrite(ctx, req, relationRewrite, foundUsersChan)
+	if resp.err != nil {
+		telemetry.TraceError(span, resp.err)
 	}
-	return nil
+	return resp
 }
 
 func (l *listUsersQuery) expandRewrite(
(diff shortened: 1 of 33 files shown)


pkg/storage/test/storage.go
```diff
@@ -53,3 +56,34 @@ func RunAllTests(t *testing.T, ds storage.OpenFGADatastore) {
 	// Stores.
 	t.Run("TestStore", func(t *testing.T) { StoreTest(t, ds) })
 }
+
+// BootstrapFGAStore is a utility which can used to quickly bootstrap an OpenFGADatastore
+// with an FGA model and relationship tuples. It returns the (store_id, model_id), respectively.
+func BootstrapFGAStore(
```
jon-whit, line 62: Bit tangential, but added this as per prior conversations and to assist with making these kinds of changes easier next time around.

pkg/tuple/tuple.go
```diff
@@ -325,3 +325,64 @@ func IsTypedWildcard(s string) bool {
...
 }
+
+// MustParseTupleString attempts to parse a relationship tuple specified
+// in string notation and return the protobuf TupleKey for it. If parsing
+// of the string fails this  function will panic. It is meant for testing
+// purposes.
+//
+// Given string 'document:1#viewer@user:jon', return the protobuf TupleKey
+// for it.
+func MustParseTupleString(s string) *openfgav1.TupleKey {
```
jon-whit, line 336: Introduced the helpers here to help us start using the relationship tuple notation `object#relation@user` in our codebase. I wanted to do so in this PR, so I just introduced these in this.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -178,22 +177,23 @@ func (l *listUsersQuery) ListUsers(
...
-	select {
-	case err := <-expandErrCh:
-		telemetry.TraceError(span, err)
-		return nil, err
-	case <-done:
-		cancelContextIfMaxResultsMet()
-		break
+	err := <-expandErrCh
+	if err != nil {
+		if !errors.Is(err, context.DeadlineExceeded) {
```
jon-whit, line 188: One thing I want/need to add a test for here as well. If we're getting `context.DeadlineExceeded` from the storage layer, then I want to treat it differently.

I will add a test for this to make sure we're treating it properly. I think the way the code is written right now it will skip the internal error.

pkg/storage/test/storage.go
```diff
@@ -53,3 +56,34 @@ func RunAllTests(t *testing.T, ds storage.OpenFGADatastore) {
...
+
+	tuples := tuple.MustParseTupleStrings(tupleStrs...)
+
+	batchSize := ds.MaxTuplesPerWrite()
+	for batch := 0; batch < len(tuples); batch += batchSize {
+		batchEnd := batch + batchSize
+
+		if batchEnd > len(tuples) {
+			batchEnd = len(tuples)
+		}
```
miparnisari, line 82: Nit: `batchEnd := math.Min(batch + batchSize, len(tuples)`

pkg/tuple/tuple_test.go
```diff
@@ -468,3 +468,75 @@ func TestTypedPublicWildcard(t *testing.T) {
...
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			tuple, err := ParseTupleString(test.str)
+			require.Equal(t, test.expectedErr, err != nil)
+
+			if !test.expectedErr {
+				require.NotNil(t, tuple)
+				require.Equal(t, test.tuple, tuple)
+			}
```
miparnisari, line 539:
```suggestion
if test.expectedErr != nil {
	require.ErrorIs(t, actualError, test.expectedErr)
} else {
	require.NotNil(t, tuple)
	require.Equal(t, test.tuple, tuple)
}
```

pkg/tuple/tuple_test.go
```diff
@@ -468,3 +468,75 @@ func TestTypedPublicWildcard(t *testing.T) {
...
 }
+
+func TestParseTupleString(t *testing.T) {
+	tests := []struct {
+		name        string
+		str         string
+		tuple       *openfgav1.TupleKey
+		expectedErr bool
+	}{
+		{
```
miparnisari, line 479: can you add one test for `user:*` please?

pkg/tuple/tuple.go
```diff
@@ -325,3 +325,64 @@ func IsTypedWildcard(s string) bool {
...
+	return tuples
+}
+
+// ParseTupleString attempts to parse a relationship tuple specified
+// in string notation and return the protobuf TupleKey for it. If parsing
+// of the string fails this  function returns an err.
+//
+// Given string 'document:1#viewer@user:jon', return the protobuf TupleKey
+// for it or an error.
+func ParseTupleString(s string) (*openfgav1.TupleKey, error) {
```
miparnisari, line 360: Thanks for adding this 🙇‍♀️  I preferred if you submitted to `main` though, so as not to make ListUsers PR even bigger 😅

pkg/storage/test/storage.go
miparnisari, line 62:
```suggestion
// BootstrapFGAStore is a utility to write an FGA model and relationship tuples to a datastore. 
// It doesn't validate the model or the tuples.
// It returns the (store_id, model_id), respectively.
func BootstrapFGAStore(
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -178,22 +177,23 @@ func (l *listUsersQuery) ListUsers(
...
+			telemetry.TraceError(span, err)
+			return nil, err
+		}
 	}
+
+	<-done
```
miparnisari, line 194: Why did you remove the `select` statement? That will neglect the `max_results` setting.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -178,22 +177,23 @@ func (l *listUsersQuery) ListUsers(
...
 	go func() {
-		defer close(foundUsersCh)
 		internalRequest := fromListUsersRequest(req, &datastoreQueryCount)
-		if err := l.expand(cancellableCtx, internalRequest, foundUsersCh); err != nil {
-			expandErrCh <- err
-			return
-		}
+		err := l.expand(cancellableCtx, internalRequest, foundUsersCh)
+		close(foundUsersCh)
+		expandErrCh <- err
```
miparnisari, line 183: I don't like the idea of sending `nil`s through a channel. That can cause a panic if the receiver forgets to check for nil

pkg/server/commands/listusers/list_users_rpc.go
willvedd: Reminder – if we revert to a swtich (as mentioned above), we still need to nil-check the err before tracing it.

pkg/tuple/tuple_test.go
jon-whit: We're not returning sentinel errors, so `ErrorIs` doesn't work here. The intent was not to test the error message as that is subject to change, but to assert that an error is raised. So that was the intent of the bool for `test.expectedErr` vs error type.

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: I thought I had the control flow covered correctly, but I was wrong. I've updated that channel's variable name so it's more indicative of it's relevance and have added it back in 👍

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Changed this so that we only send `if err != nil`

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Adjust the control flow. Now we only send on the channel if `if err != nil`

pkg/tuple/tuple.go
jon-whit: My thoughts there - the tests utilizing these changes are only applicable to the scope of this PR right now, but I'm happy to move it to main ahead of the curve if that's the preference. Sounds like that is, so I'll just bite the bullet and do so.

jon-whit: https://github.com/openfga/openfga/pull/1586 - moved over to a PR targeting `main`.

pkg/tuple/tuple_test.go
jon-whit: Added this test https://github.com/openfga/openfga/pull/1586/files#diff-57a1b9934b1de403d56fdc98ba5b12686658b098e57e021f371a5dff321f0652R387-R391 in the PR targeting `main`.

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Fixed 👍

willvedd (approved):
Confirmed that this works as designed 👍

PR #1588, opened 2024-05-01
test: always ensure health check

miparnisari:
## Description
Fix flaky test seen in https://github.com/openfga/openfga/actions/runs/8915815591/job/24486664293?pr=1576

image


jon-whit:
@miparnisari please rename this PR title to something like "test: always ensure health check". We don't always do HTTP health checks, but we want to always do a health check of some form (either gRPC or HTTP or both).

jon-whit (approved):

PR #1589, opened 2024-05-02
perf: reduce spans in check api

miparnisari:
## Description
Prefer trace attributes over spans to reduce number of spans created.

## Demo

image

### DIFF
--- a/internal/graph/cached_resolver.go
+++ b/internal/graph/cached_resolver.go
@@ -139,13 +139,8 @@ func (c *CachedCheckResolver) ResolveCheck(
 	ctx context.Context,
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
-	ctx, span := tracer.Start(ctx, "ResolveCheck", trace.WithAttributes(
-		attribute.String("store_id", req.GetStoreID()),
-		attribute.String("resolver_type", "CachedCheckResolver"),
-		attribute.String("tuple_key", req.GetTupleKey().String()),
-		attribute.Bool("is_cached", false),
-	))
-	defer span.End()
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(attribute.Bool("is_cached", false))
 
 	checkCacheTotalCounter.Inc()
 
--- a/internal/graph/cached_resolver.go
+++ b/internal/graph/cached_resolver.go
@@ -140,8 +140,6 @@ func (c *CachedCheckResolver) ResolveCheck(
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
 	span := trace.SpanFromContext(ctx)
-	span.SetAttributes(attribute.Bool("is_cached", false))
-
 	checkCacheTotalCounter.Inc()
 
 	cacheKey, err := CheckRequestCacheKey(req)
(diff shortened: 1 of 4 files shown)


internal/graph/cached_resolver.go
```diff
@@ -139,13 +139,8 @@ func (c *CachedCheckResolver) ResolveCheck(
...
 ) (*ResolveCheckResponse, error) {
-	ctx, span := tracer.Start(ctx, "ResolveCheck", trace.WithAttributes(
-		attribute.String("store_id", req.GetStoreID()),
-		attribute.String("resolver_type", "CachedCheckResolver"),
-		attribute.String("tuple_key", req.GetTupleKey().String()),
-		attribute.Bool("is_cached", false),
-	))
-	defer span.End()
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(attribute.Bool("is_cached", false))
```
adriantam, line 143: Do we only want to set this when we know is_cached != true (i.e. line 163)?  Otherwise, would it be possible that line 157 fails because the attribute is already set (but to false)

internal/graph/dispatch_throttling_check_resolver.go
```diff
@@ -98,9 +99,7 @@ func (r *DispatchThrottlingCheckResolver) runTicker() {
 func (r *DispatchThrottlingCheckResolver) ResolveCheck(ctx context.Context,
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
-	ctx, span := tracer.Start(ctx, "ResolveCheck")
-	defer span.End()
-	span.SetAttributes(attribute.String("resolver_type", "DispatchThrottlingCheckResolver"))
+	span := trace.SpanFromContext(ctx)
```
adriantam, line 102: can we set attribute on throttled or not?  see line 121

internal/graph/cycle_check_resolver.go
```diff
@@ -30,13 +30,8 @@ func (c *CycleDetectionCheckResolver) ResolveCheck(
...
 ) (*ResolveCheckResponse, error) {
-	ctx, span := tracer.Start(ctx, "ResolveCheck", trace.WithAttributes(
-		attribute.String("store_id", req.GetStoreID()),
-		attribute.String("resolver_type", "CycleDetectionCheckResolver"),
-		attribute.String("tuple_key", req.GetTupleKey().String()),
-		attribute.Bool("cycle_detected", false),
-	))
-	defer span.End()
+	span := trace.SpanFromContext(ctx)
+	span.SetAttributes(attribute.Bool("cycle_detected", false))
```
adriantam, line 34: Similar to comment above - should we only set this in the negative path (i.e line 53)

adriantam (commented):
Can we at least manually test the positive case (i.e., cycle detected = true and cache = true case) to see if it still works correctly?

internal/graph/cached_resolver.go
miparnisari: `SetAttributes` overrides values:

https://github.com/open-telemetry/opentelemetry-go/blob/9656d0afa72646101e859bf8a1c6d05b73ee094d/trace/trace.go#L382-L385

miparnisari: > Do we only want to set this when we know is_cached != true 

I have no strong feelings about this, we could do either!

adriantam (approved):

PR #1590, opened 2024-05-02
feat(list-users): ListUsers with Excluded Users

jon-whit:
## Description
This PR introduces the concept of negation "differences" which represent the list of users that do not apply to a negation clause. This occurs in models that have chained negation where the intermediate negation includes a public typed wildcard. 

**Example:**
```
model
	schema 1.1
type user
type document
	relations
```
```
```

In this example, the "unblocked" status of `user:jon` was being lost when evaluating the outer negation because the exclusion to the negation was not being communicated. This is considered an "**excluded user**". The solution here is to keep track of the excluded users so that the outer negation context can un-exclude the proper subjects. We also return the excluded users along with the response for the client to handle appropriately. See: https://github.com/openfga/api/pull/144/files#diff-d3a74ea9ecaac0a46611253dc0c4bc9c402e2ad9eb255cd41d07266453e258f0R864

## References

## Review Checklist

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc_test.go
+++ b/pkg/server/commands/listusers/list_users_rpc_test.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/openfga/openfga/internal/graph"
 	"github.com/openfga/openfga/internal/mocks"
+	storagetest "github.com/openfga/openfga/pkg/storage/test"
 
 	"github.com/openfga/openfga/pkg/storage/memory"
 	"github.com/openfga/openfga/pkg/storage/storagewrappers"
@@ -26,12 +27,13 @@ import (
 )
 
 type ListUsersTests []struct {
-	name             string
-	req              *openfgav1.ListUsersRequest
-	model            string
-	tuples           []*openfgav1.TupleKey
-	expectedUsers    []string
-	expectedErrorMsg string
+	name                  string
+	req                   *openfgav1.ListUsersRequest
+	model                 string
+	tuples                []*openfgav1.TupleKey
+	expectedUsers         []string
+	expectedExcludedUsers []string
+	expectedErrorMsg      string
 }
 
 const maximumRecursiveDepth = 25
@@ -504,12 +506,12 @@ func TestListUsersUsersets(t *testing.T) {
 			model: `model
 			schema 1.1
 		  type user
-		  
+
 		  type org
 			relations
 			  define member: [user]
 			  define admin: [org#member]
-		  
+
 		  type repo
 			relations
 			  define owner: [org]
@@ -536,7 +538,7 @@ func TestListUsersUsersets(t *testing.T) {
 			model: `model
             schema 1.1
           type user
-          
+
           type org
             relations
               define member: [user]
@@ -627,14 +629,14 @@ func TestListUsersTTU(t *testing.T) {
 			model: `model
 			schema 1.1
 		  type user
-		
+
 		  type folder
 			relations
 				define owner: [user]
 				define editor: [user] or owner
 				define viewer: [user] or owner or editor
 				define unrelated_not_computed: [user]
-		
+
 		  type document
 			relations
 			  define parent: [folder]
@@ -665,8 +667,8 @@ func TestListUsersTTU(t *testing.T) {
 			},
 			model: `model
 			schema 1.1
-			type user 
-			type folder 
+			type user
+			type folder
 				relations
 					define parent: [folder]
 					define viewer: [user] or viewer from parent`,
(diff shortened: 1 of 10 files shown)


jon-whit:
Supersedes #1534

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2378,6 +2390,153 @@ func TestListUsersCycleDetection(t *testing.T) {
...
+			name: "chained_negation_8",
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "viewer", "user:jon"),
+				tuple.NewTupleKey("document:1", "blocked", "user:*"),
+				tuple.NewTupleKey("document:1", "unblocked", "user:*"),
+				tuple.NewTupleKey("document:1", "unblocked", "user:maria"),
+			},
+			expectedUsers: []string{"user:jon"},
+		},
+		// {
```
jon-whit, line 2476: Following up with a subsequent PR to resolve the behavior changes to get these to pass.

pkg/server/list_users_test.go
```diff
@@ -231,12 +249,39 @@ func TestExperimentalListUsers(t *testing.T) {
 		require.Equal(t, codes.Unimplemented, e.Code())
 	})
 
-	t.Run("list_users_does_not_error_if_experimentally_enabled", func(t *testing.T) {
+	t.Run("list_users_returns_error_if_latest_model_not_found", func(t *testing.T) {
+		mockDatastore.EXPECT().FindLatestAuthorizationModel(gomock.Any(), gomock.Any()).Return(nil, storage.ErrNotFound) // error demonstrates that main code path is reached
+
 		server.experimentals = []ExperimentalFeatureFlag{ExperimentalEnableListUsers}
```
miparnisari, line 255: nit: the parent test is `TestExperimentalListUsers` so i feel this line could be moved up

pkg/tuple/tuple.go
```diff
@@ -198,6 +198,43 @@ func StringToUserProto(userKey UserString) *openfgav1.User {
...
+func FromObjectOrUsersetProto(obj *openfgav1.ObjectOrUserset) UserString {
+	switch obj.GetUser().(type) {
+	case *openfgav1.ObjectOrUserset_Userset:
+		us := obj.GetUser().(*openfgav1.ObjectOrUserset_Userset)
+		return fmt.Sprintf("%s:%s#%s", us.Userset.GetType(), us.Userset.GetId(), us.Userset.GetRelation())
+	case *openfgav1.ObjectOrUserset_Object:
+		us := obj.GetUser().(*openfgav1.ObjectOrUserset_Object)
+		return fmt.Sprintf("%s:%s", us.Object.GetType(), us.Object.GetId())
+	default:
+		panic("unsupported type")
```
miparnisari, line 212:
```suggestion
		return ""
```

pkg/tuple/tuple.go
```diff
@@ -198,6 +198,43 @@ func StringToUserProto(userKey UserString) *openfgav1.User {
...
+	}
+}
+
+// StringToObjectOrUserset returns a ObjectOrUserset proto from a string. Ex: 'user:maria#member'.
```
miparnisari, line 216:
```suggestion
// StringToObjectOrUserset returns a ObjectOrUserset proto from a string. Ex: 'group:fga#member'.
```

pkg/tuple/tuple_test.go
```diff
@@ -463,6 +463,86 @@ func TestToUserProto(t *testing.T) {
...
+						Type:     "user",
+						Id:       "id-123",
+						Relation: "member",
+					},
+				},
+			},
+			expected: "user:id-123#member",
+		},
+	} {
+		tc := tc
```
miparnisari, line 498: nit: this line is only needed when you use `t.Parallel`

miparnisari (approved):

willvedd (approved):
LGTM 👍

pkg/tuple/tuple.go
willvedd: Know we've been trying to be mindful about panicking but I prefer error'ing early rather than later in the process if we return empty string.

jon-whit: I think we need to revisit these in a follow up PR anyways. We're introducing yet another series of functions to go to/from strings and structs.

I think we should overload `UserString` with methods such as `IsValid` which does string validation checks and other methods like `ToUserProto` etc.. This would make better use of that type and consolidate code that is now being duplicated more and more.

jon-whit:
Superseded by #1604

PR #1604, opened 2024-05-06
feat(list-users): ListUsers with Excluded Users

jon-whit:
## Description
This PR introduces the concept of excluded users, which represent the list of users that do not apply to an exclusion clause. This occurs in models that have chained exclusion where the intermediate exclusion includes a typed public wildcard. 

**Example:**
```
model
  schema 1.1

type user

type document
  relations
```
```
```

In this example, the "unblocked" status of `user:jon` was being lost when evaluating the outer exclusion because the exception to the exclusion was not being communicated. This is considered an "**excluded user**". The solution here is to keep track of the excluded users so that the outer negation context can un-exclude the proper subjects. We also return the excluded users along with the response for the client to handle appropriately.

Consider the same model above but the following tuples:
```
"document:1", "viewer", "user:*"
```

`ListUsers(document:1#viewer, user_filters=[{"type": "user"}])` will/should propagate up the `user:jon` subject from the inner most exclusion. That is, intermediate expansions should always yield a result, but that result may indicate the subject does or does not have the relationship for that intermediate relationships. In this case, that means the following:
```
ListUsers(document:1#viewer, user_filters=[{"type": "user"}]) --> [{user: "user:jon"}]
--> expandExclusion(document:1#viewer, filter={type: "user"}) --> [{user: "user:jon"}]
-----> base --> [{user: "user:*"}]
-----> sub  --> [{user: "user:*", noRelationship: false}, {user: "user:jon", noRelationship: true}]
--------> expandExclusion(document:1#blocked, filter={type: "user"}) --> [{user: "user:*", noRelationship: false}, {user: "user:jon", noRelationship: true}]
-----------> base --> [{user: "user:*}]
-----------> sub  --> [{user: "user:jon"}]
```

## References

## Review Checklist


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -47,6 +48,12 @@ type expandResponse struct {
 	err      error
 }
 
+type foundUser struct {
+	user           *openfgav1.User
+	excludedUsers  []*openfgav1.User
+	noRelationship bool
```
miparnisari, line 54: can you add godoc explaining what this means?

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -569,15 +612,22 @@ func (l *listUsersQuery) expandExclusion(
...
 	for fu := range baseFoundUsersCh {
-		key := tuple.UserProtoToString(fu)
-		baseFoundUsersMap[key] = struct{}{}
+		key := tuple.UserProtoToString(fu.user)
+		baseFoundUsersMap[key] = fu
 	}
-	subtractFoundUsersMap := make(map[string]struct{}, len(baseFoundUsersMap))
+
+	subtractFoundUsersMap := make(map[string]foundUser, len(baseFoundUsersMap))
+	negatedSubtractionMap := make(map[string]struct{})
```
willvedd, line 622: `negatedSubtractionMap` doesn't appear to be used anymore besides being assign to below. But its never being read from.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -546,12 +589,12 @@ func (l *listUsersQuery) expandExclusion(
...
 	rewrite *openfgav1.Userset_Difference,
-	foundUsersChan chan<- *openfgav1.User,
+	foundUsersChan chan<- foundUser,
 ) expandResponse {
 	ctx, span := tracer.Start(ctx, "expandExclusion")
 	defer span.End()
-	baseFoundUsersCh := make(chan *openfgav1.User, 1)
-	subtractFoundUsersCh := make(chan *openfgav1.User, 1)
+	baseFoundUsersCh := make(chan foundUser, 1)
+	subtractFoundUsersCh := make(chan foundUser, 1)
```
willvedd, line 597: FWIW during performance testing, having this channel as size 1 caused back pressure and modest performance degradation. Increasing the size to be much larger yielded a quick win with negation performance; it's really not great to start.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -592,14 +642,116 @@ func (l *listUsersQuery) expandExclusion(
...
-			// Iterate over base users because at minimum they need to pass
-			// but then they are further compared to the subtracted users map.
-			// If users exist in both maps, they are excluded. Only users that exist
-			// solely in the base map will be returned.
-			trySendResult(ctx, tuple.StringToUserProto(key), foundUsersChan)
+
+	_, baseWildcardExists := baseFoundUsersMap[wildcardKey]
+	subtractWildcardUser, subtractWildcardExists := subtractFoundUsersMap[wildcardKey]
+
+	if baseWildcardExists {
```
willvedd, line 649: Heads up, the linter is complaining about this set of conditionals.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -592,14 +642,116 @@ func (l *listUsersQuery) expandExclusion(
...
+		if !subtractWildcardExists || subtractWildcardUser.noRelationship {
+			trySendResult(ctx, foundUser{
+				user: tuple.StringToUserProto(wildcardKey),
+			}, foundUsersChan)
+		}
+
+		residualSubtractedUserMap := maps.Clone(subtractFoundUsersMap)
```
willvedd, line 656: Is this clone necessary? Ideally the `subtractFoundUsersMap` shouldn't be mutated.

willvedd (commented):
Still need to look over more but some immediate comments.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -592,14 +642,116 @@ func (l *listUsersQuery) expandExclusion(
...
+			excludedUsers := map[string]struct{}{}
+			if tuple.IsTypedWildcard(userKey) {
+				for _, excludedUser := range fu.excludedUsers {
+					excludedUsers[tuple.UserProtoToString(excludedUser)] = struct{}{}
+
+					trySendResult(ctx, foundUser{
+						user:           excludedUser,
+						noRelationship: false,
+					}, foundUsersChan)
+				}
```
miparnisari, line 681: This block isn't covered with tests

miparnisari (changes requested):
I wrote two tests that I've corroborated with Check API manually (not sure why we're not doing that within `runListUsersTestCases`) and that are failing with this PR:

```go
{
			name: "intersection",
				Object:   &openfgav1.Object{Type: "document", Id: "1"},
				Relation: "viewer",
				UserFilters: []*openfgav1.UserTypeFilter{
					{
						Type: "user",
					},
				},
			},
			model: `model
			schema 1.1
			type user
			type document
				relations
					define b: [user:*] but not b1
					define b1: [user]
					define a: [user:*] but not a1
					define a1: [user]
					define viewer: a and b`,

				tuple.NewTupleKey("document:1", "b", "user:*"),
				tuple.NewTupleKey("document:1", "b1", "user:will"),
				// b -> user:* but not user:will
				tuple.NewTupleKey("document:1", "a", "user:*"),
				tuple.NewTupleKey("document:1", "a1", "user:maria"),
				// a -> user:* but not user:maria
				// therefore:
				// a and -b -> (user:* but not user:maria) and (user:* but not user:will)
				//          -> user:* but not (user:maria, user:will)
			},
			expectedUsers: []string{"user:*"},
			butNot:        []string{"user:maria", "user:will"},
		},
		{
			name: "union",
				Object:   &openfgav1.Object{Type: "document", Id: "1"},
				Relation: "viewer",
				UserFilters: []*openfgav1.UserTypeFilter{
					{
						Type: "user",
					},
				},
			},
			model: `model
			schema 1.1
			type user
			type document
				relations
					define b: [user:*] but not b1
					define b1: [user]
					define a: [user:*] but not a1
					define a1: [user]
					define viewer: a or b`,

				tuple.NewTupleKey("document:1", "b", "user:*"),
				tuple.NewTupleKey("document:1", "b1", "user:will"),
				// b -> user:* but not user:will
				tuple.NewTupleKey("document:1", "a", "user:*"),
				tuple.NewTupleKey("document:1", "a1", "user:maria"),
				// a -> user:* but not user:maria
				// therefore:
				// a or -b -> (user:* but not user:maria) or (user:* but not user:will)
				//          -> user:*, user:maria, user:will
			},
			expectedUsers: []string{"user:*", "user:maria", "user:will"},
		},
```

I'm not very sure on the changes you made but it seems to be that we need changes in union and intersection handlers too.

More specifically I think we need a helper method that recursively computes the reduced versions of intersection, union and exclusion of sets.

E.g. 
```
(user:* but not user:maria) or (user:* but not user:will) -> user:* but not (user:maria, user:will)

(user:* but not user:maria) and (user:* but not user:will) -> user:*, user:maria, user:will
```

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Let's address the performance concerns and findings in a follow up related specifically to those findings.

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: I've inverted it now. The control flow now works under the outer range over the `baseFoundUsersMap`. Originally I just wanted to get the logic of the `expandExclusion` correct, but now taking a second look at it the control flow was a bit repetitive. It's more succint now.

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Removed

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: It's no longer necessary with the new control flow pattern. I've removed it 👍

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: Yep

miparnisari:
Are we going to address the case of excluded usersets in a different PR?

```
{
			name: "usersets",
			model: `model
			schema 1.1

			type user
			type group
				relations
					define member: [user]

			type document
			  relations
				define blocked: [group#member]
				define viewer: [group#member] but not blocked
			`,
				tuple.NewTupleKey("document:1", "blocked", "group:fga#member"),
			},
				Object:      &openfgav1.Object{Type: "document", Id: "1"},
				Relation:    "viewer",
				UserFilters: []*openfgav1.UserTypeFilter{{Type: "group", Relation: "member"}},
			},
			expectedUsers: []string{"group:eng#member"},
			butNot:        []string{"group:fga#member"},
		},
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -658,20 +665,15 @@ func (l *listUsersQuery) expandExclusion(
 	wildcardKey := tuple.TypedPublicWildcard(req.GetUserFilters()[0].GetType())
 
 	_, baseWildcardExists := baseFoundUsersMap[wildcardKey]
-	subtractWildcardUser, subtractWildcardExists := subtractFoundUsersMap[wildcardKey]
-
-	if baseWildcardExists && (!subtractWildcardExists || subtractWildcardUser.relationshipStatus == NoRelationship) {
-		trySendResult(ctx, foundUser{
-			user: tuple.StringToUserProto(wildcardKey),
-		}, foundUsersChan)
-	}
```
willvedd: **Note:** Realizing in hindsight that this is probably an easy optimization we can apply. So we'll need to re-add this back in and remove the redundant code in the main loop below.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -53,6 +53,13 @@ type expandResponse struct {
 // they either explicitly do have a relationship or explicitly do not.
 type userRelationshipStatus int
 
+func (r *userRelationshipStatus) InvertRelationshipStatus() userRelationshipStatus {
```
jon-whit, line 56: Why the pointer receiver instead of the non pointer receiver here? Since we just directly de-ref it, does it make more sense to just use a non-pointer?

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -658,20 +665,15 @@ func (l *listUsersQuery) expandExclusion(
...
 
 	for userKey, fu := range baseFoundUsersMap {
 		subtractedUser, userIsSubtracted := subtractFoundUsersMap[userKey]
 		_, wildcardSubtracted := subtractFoundUsersMap[wildcardKey]
 
-		if baseWildcardExists {
-			if !userIsSubtracted && !wildcardSubtracted {
+		switch {
+		case baseWildcardExists:
+			if !wildcardSubtracted {
```
jon-whit, line 676: This conditional behavior is not quite correct and has deviated from the prior commit. Previously we were only producing the wildcard subject if `baseWildcardExists && !wildcardSubtracted`. See https://github.com/openfga/openfga/blob/b872f91c05f8c130e369656e4bddf8dc3ac47914/pkg/server/commands/listusers/list_users_rpc.go#L665

Consider the following scenario:

```
type user

type document
  relations
    define restricted: [user, user:*]
    define viewer: [user, user:*] but not restricted
```
with tuples:
* document:1#viewer@user:*
* document:1#viewer@user:jon
* document:1#restricted@user:jon

When ranging over the `baseFoundUsersMap` we'll find `user:jon` and yield it here, because there is not a subtracted wildcard. The only reason a test related to this scenario is passing against this commit is because of other code in the callstack above this function, which isn't obvious and is actually misleading when reading the code. See https://github.com/openfga/openfga/blob/5f20ffb21dfee1e6aef05b6360dd4e8208bf3425/pkg/server/commands/listusers/list_users_rpc.go#L251-L253

But this logic itself is not quite correct here. We're producing on the `foundUsersChan` superfluously only to be filtered out above upstream since it is not applicable (the subject `user:jon` is not just excluded but is explicitly subtracted). Ideally this rewrite handler should only produce accurate "candidates" with respect to the found users channel.

willvedd: Are you ok with proving your above theory with a unit test? Then we can address appropriately in code.

willvedd:
This is one test case I'm still pondering though, could see it going either way:

```
		{
			name: "chained_negation_14",
				tuple.NewTupleKey("document:1", "viewer", "user:*"),
				tuple.NewTupleKey("document:1", "blocked", "user:*"),
				tuple.NewTupleKey("document:1", "blocked", "user:maria"),
				tuple.NewTupleKey("document:1", "unblocked", "user:*"),
			},
			expectedUsers: []string{"user:*"}, // should `user:maria` be included?
		},
```

jon-whit:
> This is one test case I'm still pondering though, could see it going either way:
> 
> ```
> 		{
> 			name: "chained_negation_14",
> 			tuples: []*openfgav1.TupleKey{
> 				tuple.NewTupleKey("document:1", "viewer", "user:*"),
> 				tuple.NewTupleKey("document:1", "blocked", "user:*"),
> 				tuple.NewTupleKey("document:1", "blocked", "user:maria"),
> 				tuple.NewTupleKey("document:1", "unblocked", "user:*"),
> 			},
> 			expectedUsers: []string{"user:*"}, // should `user:maria` be included?
> 		},
> ```

For consistency, `user:maria` should be included. This is for the same reason that we return the following:

```
  tuple.NewTupleKey("document:1", "viewer", "user:*"),
  tuple.NewTupleKey("document:1", "blocked", "user:*"),
  tuple.NewTupleKey("document:1", "unblocked", "user:maria"),

  expectedUsers: []string{"user:maria"}
```
Maria is explicitly blocked, but then subsequently unblocked, so she is explicitly _not_ blocked. For any subject that shows up in the `base` operand, if the subtracted operand contains a wildcard subject of that same type, then every base subject of the exclusion should have a relationship status `NO_RELATIONSHIP`. If this exclusion is then included as an operand of another chained exclusion, then those subjects will propagate up appropriately.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -3281,3 +3637,91 @@ func TestListUsersConfig_MaxConcurrency(t *testing.T) {
...
+
+func TestListUsers_ExpandExclusionHandler(t *testing.T) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+
+	ds := memory.New()
+	t.Cleanup(ds.Close)
+
+	t.Run("avoid_producing_explicitly_negated_subjects", func(t *testing.T) {
```
jon-whit, line 3654: This fails right now and demonstrates we're producing more subjects than we should from the exclusion handler. For example, we should only produce `user:*` with the tuples and model provided, but we end up producing 3-4 subjects (`user:*` and multiple copies of `user:jon`).

Screenshot 2024-05-20 at 9 14 58 AM

pkg/server/commands/listusers/list_users_rpc.go
jon-whit: https://github.com/openfga/openfga/pull/1604/files#r1606950464

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -592,14 +747,68 @@ func (l *listUsersQuery) expandExclusion(
...
+							user:               tuple.StringToUserProto(subtractedUserKey),
+							relationshipStatus: NoRelationship,
+							excludedUsers: []*openfgav1.User{
+								tuple.StringToUserProto(subtractedUserKey),
+							},
+						}, foundUsersChan)
+					}
+				}
+			}
+		case subtractWildcardExists:
```
jon-whit, line 796: This case can be collapsed along with the case below

```go
case userIsSubtracted, subtractWildcardExists:
    trySendResult(ctx, foundUser{
        user:               tuple.StringToUserProto(userKey),
        relationshipStatus: subtractedUser.relationshipStatus.InvertRelationshipStatus(),
    }, foundUsersChan)
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -592,14 +747,68 @@ func (l *listUsersQuery) expandExclusion(
...
+		switch {
+		case baseWildcardExists:
+			if !userIsSubtracted && !wildcardSubtracted {
+				trySendResult(ctx, foundUser{
+					user: tuple.StringToUserProto(userKey),
+				}, foundUsersChan)
+			}
+
+			for subtractedUserKey, subtractedFu := range subtractFoundUsersMap {
+				if tuple.IsTypedWildcard(subtractedUserKey) && !userIsSubtracted {
```
jon-whit, line 767: Can we collapse this and the code block below into a common block?

```go
if tuple.IsTypedWildcard(subtractedUserKey) {
    if !userIsSubtracted {
        trySendResult(...)
    }

    continue
}
```

willvedd (approved):
🙏

Issue #946, opened 2023-08-16
Support Contextual Tuples in Assertions API


lorenlew:
Would it possible for the assertions to take into consideration the contextual tuples? (for advanced scenarios with contextual based access)
For now the assertions are pure (object, relation, subject) tuples.

rhamzeh:
Hi @lorenlew - right now it doesn't however we are planning an overhaul of that API.

The new interface will be coming to the CLI first so that we can get all your feedback on it before adding support for it to the API itself.

You can see @aaguiarz talk about it in our latest community meeting: https://youtu.be/6oQOKgJ53V0?t=1602

It is expected to look something like this, where each test can have its own contextual tuples:
```yaml
tests:
  - name: folder-access
    description: ""
    tuples:
      - user: user:anne
        relation: owner
        object: folder:product
      - user: folder:product
        relation: parent
        object: folder:product-2021
      - user: user:beth
        relation: viewer
        object: folder:product-2021
    check:
      - user: user:anne
        object: folder:product-2021
        assertions:
          can_view: true
          can_write: true
          can_share: true
      - user: user:beth
        object: folder:product-2021
        assertions:
          can_view: true
          can_write: false
          can_share: false
    list-objects:
      - user: user:anne
        type: folder
        assertions:
          can_view:
            - folder:product
            - folder:product-2021
          can_write:
            - folder:product
            - folder:product-2021
          can_share:
            - folder:product
            - folder:product-2021
      - user: user:beth
        type: folder
        assertions:
          can_view:
            - folder:product-2021
          can_write: []
          can_share: []
```

Let us know what you think!

PR #1610, opened 2024-05-09
feat: Added Contextual Tuples to Assertions API

smeadows-okta:
Added support for contextual tuple to assertions

## Description
Currently, assertions are pure (object, relation, subject) tuples. Added support for contextual tuples to assertions for advanced scenarios.

## References
[issue-946](https://github.com/openfga/openfga/issues/946)

## Review Checklist


pkg/server/test/write_assertions.go
```diff
@@ -208,3 +250,134 @@ type repo
 		})
 	}
 }
+
+func TestConceptualTuplesWriteAssertionsFailure(t *testing.T, datastore storage.OpenFGADatastore) {
```
miparnisari, line 254: In the previous test you added elements to the test case array. 👍 

But here you basically copied `TestWriteAssertionsFailure`. Why? You could have done what you did above, and include your new test cases in `TestWriteAssertionsFailure`

pkg/server/test/write_assertions.go
```diff
@@ -208,3 +250,134 @@ type repo
...
+	}
+	ctx := context.Background()
+
+	writeAuthzModelCmd := commands.NewWriteAuthorizationModelCommand(datastore)
+	modelID, err := writeAuthzModelCmd.Execute(ctx, githubModelReq)
+	require.NoError(t, err)
+
+	var tests = []writeAssertionsTestSettings{
+		{
+			_name: "write_conceptual_tuple_assertion_with_invalid_relation_fails",
```
miparnisari, line 284:
```suggestion
			_name: "write_contextual_tuple_assertion_with_invalid_relation_fails",
```

here and elsewhere

pkg/server/test/write_assertions.go
```diff
@@ -208,3 +250,134 @@ type repo
...
+					Expectation: false,
+				},
+			},
+			modelID: modelID.GetAuthorizationModelId(),
+			err: serverErrors.ValidationError(
+				fmt.Errorf("invalid 'object' field format"),
+			),
+		},
+		{
+			_name: "write_conceptual_tuple_assertion_with_invalid_object_fails",
```
miparnisari, line 326:
```suggestion
			_name: "write_conceptual_tuple_assertion_with_invalid_user_fails",
```

pkg/server/test/write_assertions.go
```diff
@@ -208,3 +250,134 @@ type repo
...
+						},
+					},
+					Expectation: false,
+				},
+			},
+			modelID: "not_valid_id",
+			err: serverErrors.AuthorizationModelNotFound(
+				"not_valid_id",
+			),
+		},
```
miparnisari, line 367: This test isn't useful - validation of contextual happens will never happen if the model ID isn't valid.

miparnisari (changes requested):

pkg/server/test/write_assertions.go
smeadows-okta: 👍

pkg/server/test/write_assertions.go
smeadows-okta: 👍

pkg/server/test/write_assertions.go
smeadows-okta: 👍

pkg/server/test/write_assertions.go
smeadows-okta: I broke them out into separate test cases tuples vs conceptual tuples. Each are handled differently in the code (Check, WriteAssertion) and returns the same error codes and need to be tested as such and allow anyone to expand on them without impact the other.

miparnisari: > tuples vs contextual tuples. 

Both are part of the WriteAssertion input. 

In general, across the codebase, we have test cases with a `request` input that tests all the various possible combinations for any API. I think this specific suite shouldn't be any different.

> Each are handled differently in the code (Check, WriteAssertion) and returns the same error codes

If we want, in the future we can come back and create a helper utility that has all the contextual tuple validation code.

miparnisari (approved):

pkg/server/test/write_assertions.go
smeadows-okta: 👍

PR #1612, opened 2024-05-10
test: fix leak in tests that spin up server with OIDC auth

miparnisari:
## Description

Follow-up of https://github.com/openfga/openfga/pull/1611. The end goal is to enable the goroutine leak detector in the tests that spin up a server, so i'm slowly chipping away at those leaks.


internal/mocks/mock_oidc_server.go
```diff
@@ -80,11 +83,24 @@ func (server mockOidcServer) start() {
 		}
 	})
 
+	server.httpServer = &http.Server{Addr: ":" + port, Handler: mockHandler}
```
adriantam, line 86: do we need to check if server.httpServer is nil?

adriantam (approved):

internal/mocks/mock_oidc_server.go
miparnisari: 🤔 no, i'm creating that field here

adriantam: My question is more if it is possible to call start more than once?  If so, you will overwrite `server.httpServer` and stop will not shut down the previous http server.

miparnisari: Gotcha. Refactored code so that if you call `start` twice by accident, the test fails:

```
=== RUN   TestBuildServerWithOIDCAuthenticationAlias
2024/05/10 10:35:35 failed to start mock OIDC server{error 26 0  listen tcp :63714: bind: address already in use}
```

PR #1615, opened 2024-05-10
feat: add SQLite support

DanCech:
This PR adds support for using SQLite as a backend database.

## Description

The SQLite storage implementation is largely based on the existing MySQL backend, with some slight differences:
- migration 005 updated to issue separate ALTER TABLE statements for each column, as SQLite does not support multi-column ALTER statements
- date-related SQL updated to use SQLite syntax
- retry mechanism added for writes to avoid SQLITE_BUSY errors (this may not actually be needed with the busy_timeout setting, but is a backup in case the database layer does return SQLITE_BUSY when heavily loaded).
- error handling for SQLite unique constraint errors

It uses the cgo-free https://modernc.org/sqlite driver.

Unless otherwise specified, it will use the [WAL journal mode](https://www.sqlite.org/wal.html) and specify a [busy timeout](https://www.sqlite.org/c3ref/busy_timeout.html) of 500ms


jon-whit:
@DanCech I changed the base branch this PR is targeting because we're devising a strategy for how to handle community storage contributions without taking on the burden of the storage implementation longer term. We want to promote development, but the maintainer team can't practically support all of them and offer quality implementations if there are a lot of them we're having to maintain.

Consequently, we're thinking of an [openfga/openfga-contrib](https://github.com/openfga/openfga-contrib) repository structure and a plugin based mechanism that will allow the OpenFGA community to run OpenFGA with externally contributed storage implementations. We likely won't merge this into `main` and include it in an official release until we have a more well-defined design for that plugin ecosystem, and in the (hopefully near) future the SQLite implementation you provided herein will be one of the first community sponsored storage implementations in the contrib repo. See https://github.com/openfga/openfga-contrib/blob/main/storage/sqlite/sqlite.go

DanCech:
Thanks for the feedback @jon-whit , it looks like the base change has created quite a few conflicts. Is there anything I can do at this stage to help move things forward?

magnusart:
Hi as a consumer of openfga, if I want to try the SQLite support, what do I need to do at this point? I saw the contrib repo but there were no further instructions as far as I could see?

miparnisari:
Hey @magnusart, we haven't discussed as a team the best path forward on how to add support for more datastores. To unblock you, my recommendation is to fork this repo, and then make sure that you run all the tests like so:

```go
// 1. these tests need an openfga client that is hooked up to a sqllite container; it tests all the important APIs
check.RunAllTests(t, openfgaclient)
listobjects.RunAllTests(t, openfgaclient)
writemodel.RunAllTests(t, openfgaclient)
listusers.RunAllTests(t, openfgaclient)

// 2. this test would need a sqllite container; it runs the integration tests for the storage adapters
openfgateststorage.RunAllTests(t, ds)

// 3. this test would need a sqllite container; it tests all the APIs (yes, again)
openfgatest.RunAllTests(t, ds)
```

This is what we do to test the MySQL and Postgres datastores.

However, it's important to note that the test APIs above are subject to change.

assets/migrations/sqlite/001_initialize_schema.sql
```diff
@@ -0,0 +1,56 @@
+-- +goose Up
+CREATE TABLE tuple (
+    store CHAR(26) NOT NULL,
+    object_type VARCHAR(128) NOT NULL,
```
bytefish, line 4: As far as I understand SQLite, when defining something like `VARCHAR(255)` everything in parentheses is ignored and it will basically be a `TEXT` column. It’s interesting here to know, that a length of 255 is to be expected, but I would simply make it a `TEXT` and eventually add a `CHECK` constraint to check the length, if one can’t guarantee it from OpenFGA.

DanCech: Interesting, I have not experimented with CHECK constraints in SQLite. In general I'd tend toward checking length within OpenFGA itself if needed since the behavior of MySQL for example can vary according to the configuration (truncate and warn for non-strict mode, error in strict mode).

DanCech:
Rebased onto current main and updated to account for the storage interface changes, as well as updating to modernc.org/sqlite v1.32.0. I also looked through the diffs between the sqlite driver and tests, and the postgres and mysql implementations, and brought the sqlite support into line as best I could without changing either the mysql or postgres implementations.

I would be happy to open a separate PR to reduce the number of diffs between the drivers so that this process is easier in the future if that would be helpful. Right now there are quite a lot of boilerplate differences and naming inconsistencies to wade through in order to find the real differences.

miparnisari (commented):
Couple of other files that need change (either a modification or an addition):
- https://github.com/openfga/openfga/blob/319b81b68c90b4b880261a92b97e37199d63857a/.github/ISSUE_TEMPLATE/report_a_bug.yml#L58
- https://github.com/openfga/openfga/blob/8543ca24d92e4f77fbceaa398a467c591777df1c/.config-schema.json#L174
- https://github.com/openfga/openfga/blob/70440501b780eeea97e9d57042172ffc2f83985e/cmd/migrate/migrate_test.go#L19
- https://github.com/openfga/openfga/blob/a21b3ecf328673ba7a0ebf5eb14b78082f22ccfc/pkg/server/server_test.go#L186
- https://github.com/openfga/openfga/blob/2b8d62a4203c2c46d1613e65c59a59295219b5a6/Makefile#L124-L131
- Tests for ListUsers API https://github.com/openfga/openfga/blob/b40a1e73fa5d845bb1f7f29dca84ac4b255e2e1c/tests/listusers/listusers_test.go#L37-L39
- Tests for Check API https://github.com/openfga/openfga/blob/b40a1e73fa5d845bb1f7f29dca84ac4b255e2e1c/tests/check/check_test.go#L67-L69
- Tests for ListObjects API https://github.com/openfga/openfga/blob/8543ca24d92e4f77fbceaa398a467c591777df1c/tests/listobjects/listobjects_test.go#L22-L24
- Server tests https://github.com/openfga/openfga/blob/a21b3ecf328673ba7a0ebf5eb14b78082f22ccfc/pkg/server/server_test.go#L312-L319
- Metrics tests https://github.com/openfga/openfga/blob/817293c8812835c402dfe1cf7007a18e1d8c04ae/cmd/run/run_test.go#L791-L793
- Create a new benchmark https://github.com/openfga/openfga/blob/a21b3ecf328673ba7a0ebf5eb14b78082f22ccfc/pkg/server/server_test.go#L700-L708

pkg/storage/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,358 @@
...
+	require.NoError(t, err)
+	pbdata := []byte{0x01, 0x02, 0x03}
+
+	_, err = ds.db.ExecContext(ctx, "INSERT INTO authorization_model (store, authorization_model_id, schema_version, type, type_definition, serialized_protobuf) VALUES (?, ?, ?, ?, ?, ?)", store, modelID, schemaVersion, "document", bytes, pbdata)
+	require.NoError(t, err)
+
+	_, err = ds.ReadAuthorizationModel(ctx, store, modelID)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "cannot parse invalid wire-format data")
+}
```
miparnisari, line 220: We don't need this test :) @jpadilla can correct me if i'm wrong, but I think this was only applicable to MySQL and Postgres whilst in the middle of the migration from "one record per type" to "one record per model"

pkg/storage/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,358 @@
...
+
+// TestAllowNullCondition tests that tuple and changelog rows existing before
+// migration 005_add_conditions_to_tuples can be successfully read.
+func TestAllowNullCondition(t *testing.T) {
```
miparnisari, line 224: I also think we don't need this test. 

Note to self: test if there is a general storage test that asserts that all the `Read*` methods can read tuples with and without conditions correctly

pkg/storage/sqlite/sqlite_test.go
```diff
@@ -0,0 +1,358 @@
...
+	require.NoError(t, err)
+	require.Len(t, changes, 2)
+	require.Equal(t, tk, changes[0].GetTupleKey())
+	require.Equal(t, tk, changes[1].GetTupleKey())
+}
+
+// TestMarshalledAssertions tests that previously persisted marshalled
+// assertions can be read back. In any case where the Assertions proto model
+// needs to change, we'll likely need to introduce a series of data migrations.
+func TestMarshalledAssertions(t *testing.T) {
```
miparnisari, line 326: We don't need this test, it was only applicable to MySQL and Postgres https://github.com/openfga/openfga/pull/1217

pkg/storage/sqlite/sqlite.go
```diff
@@ -0,0 +1,810 @@
...
+	}
+	if len(filter.AllowedUserTypeRestrictions) > 0 {
+		orConditions := sq.Or{}
+		for _, userset := range filter.AllowedUserTypeRestrictions {
+			if _, ok := userset.GetRelationOrWildcard().(*openfgav1.RelationReference_Relation); ok {
+				orConditions = append(orConditions, sq.Like{"_user": userset.GetType() + ":%#" + userset.GetRelation()})
+			}
+			if _, ok := userset.GetRelationOrWildcard().(*openfgav1.RelationReference_Wildcard); ok {
+				orConditions = append(orConditions, sq.Eq{"_user": userset.GetType() + ":*"})
+			}
```
miparnisari, line 296: There's an optimization possible in this code, that we didn't have time to bring to the other datastores, but you can bring it here if you like :) https://github.com/openfga/openfga/issues/1873

The gist would be to change:
- the `tuple` table definition
- `Write` method
- All the `Read*` methods involved in reading tuples
- This particular block would be 

```go
if _, ok := userset.GetRelationOrWildcard().(*openfgav1.RelationReference_Relation); ok {
	orConditions = append(orConditions, sq.Eq{"subject_relation": userset.GetRelation()})
	orConditions = append(orConditions, sq.Eq{"subject_type": userset.GetType()})
}
if _, ok := userset.GetRelationOrWildcard().(*openfgav1.RelationReference_Wildcard); ok {
	orConditions = append(orConditions, sq.Eq{"subject_id": "*"})
	orConditions = append(orConditions, sq.Eq{"subject_type": userset.GetType()})
}
```

miparnisari:
Please also open a draft PR to our docs. These pages need update
- https://openfga.dev/docs/getting-started/setup-openfga/configure-openfga#configuring-data-storage
- https://openfga.dev/docs/getting-started/setup-openfga/docker#using-postgres

And finally we also need updates to the Helm chart:
- https://github.com/openfga/helm-charts

assets/migrations/sqlite/001_initialize_schema.sql
```diff
@@ -0,0 +1,56 @@
...
+
+CREATE UNIQUE INDEX idx_tuple_ulid ON tuple (ulid);
+
+CREATE TABLE authorization_model (
+    store CHAR(26) NOT NULL,
+    authorization_model_id CHAR(26) NOT NULL,
+    type VARCHAR(256) NOT NULL,
+    type_definition BLOB,
+    PRIMARY KEY (store, authorization_model_id, type)
+);
```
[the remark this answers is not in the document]
DanCech: Interesting, I checked whether there were new migrations but didn't check to see whether existing migrations had changed since I assumed that would never be the case.

assets/migrations/sqlite/001_initialize_schema.sql
```diff
@@ -0,0 +1,56 @@
...
+    name VARCHAR(64) NOT NULL,
+    created_at TIMESTAMP NOT NULL,
+    updated_at TIMESTAMP,
+    deleted_at TIMESTAMP
+);
+
+CREATE TABLE assertion (
+    store CHAR(26) NOT NULL,
+    authorization_model_id CHAR(26) NOT NULL,
+    assertions BLOB,
```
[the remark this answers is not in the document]
DanCech: The default limit for SQLite BLOB is 1GB, so we should be fine there.

https://www.sqlite.org/limits.html

assets/migrations/sqlite/002_add_authorization_model_version.sql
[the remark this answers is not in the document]
DanCech: I could definitely do that, initially I just wanted to keep things as consistent with the other stores as possible since they are very similar and it makes it easy to compare and check consistency (and the corresponding migrations all have the same number).

go.mod
```diff
@@ -50,6 +50,7 @@ require (
 	golang.org/x/time v0.6.0
 	google.golang.org/grpc v1.65.0
 	google.golang.org/protobuf v1.34.2
+	modernc.org/sqlite v1.32.0
```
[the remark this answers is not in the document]
DanCech: Should be good to go I think, it's 3-clause BSD https://gitlab.com/cznic/sqlite/-/blob/master/LICENSE

pkg/storage/sqlite/sqlite.go
DanCech: Will check it out 👍

pkg/storage/sqlite/sqlite_test.go
DanCech: Cool, I'm all for keeping things clean 😄  Intent here was mostly to demonstrate that this backend has parity with mysql and postgres, and can pass all the same tests.

assets/migrations/sqlite/001_initialize_schema.sql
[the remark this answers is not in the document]
DanCech: I think I may have misinterpreted your comment earlier. I have kept this backend as close as possible to the existing implementations for MySQL and Postgres (and in fact the only change from the MySQL migrations was to separate 005 into individual ALTER TABLE statements because SQLite doesn't support altering multiple columns in a single transaction.

Now (I think) I understand that you're suggesting jumping straight to the now-current table format vs following the same set of migrations we use for MySQL and PostgreSQL, since the SQLite backend doesn't have to go through that evolution.

assets/migrations/sqlite/002_add_authorization_model_version.sql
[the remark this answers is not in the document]
DanCech: Switched to a single migration file, luckily goose is ok with us starting at 005 so that the final schema version still matches and all tests still pass,

assets/migrations/sqlite/001_initialize_schema.sql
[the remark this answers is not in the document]
miparnisari: Exactly! :)

pkg/storage/sqlite/sqlite_test.go
jpadilla: Yep can drop this for new storage implementations

assets/migrations/sqlite/005_initialize_schema.sql
```diff
@@ -0,0 +1,63 @@
...
+);
+
+CREATE UNIQUE INDEX idx_tuple_ulid ON tuple (ulid);
+CREATE INDEX idx_reverse_lookup_user ON tuple (store, object_type, relation, _user);
+
+CREATE TABLE authorization_model (
+    store CHAR(26) NOT NULL,
+    authorization_model_id CHAR(26) NOT NULL,
+    type VARCHAR(256) NOT NULL,
+    type_definition BLOB,
```
miparnisari, line 23: Remove `type_definition` - it's superseded by `serialized_protobuf`

assets/migrations/sqlite/005_initialize_schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    schema_version VARCHAR(5) NOT NULL DEFAULT '1.0',
+    serialized_protobuf LONGBLOB,
+    PRIMARY KEY (store, authorization_model_id, type)
```
miparnisari, line 26:
```suggestion
    PRIMARY KEY (store, authorization_model_id)
```

assets/migrations/sqlite/005_initialize_schema.sql
```diff
@@ -0,0 +1,63 @@
...
+
+CREATE UNIQUE INDEX idx_tuple_ulid ON tuple (ulid);
+CREATE INDEX idx_reverse_lookup_user ON tuple (store, object_type, relation, _user);
+
+CREATE TABLE authorization_model (
+    store CHAR(26) NOT NULL,
+    authorization_model_id CHAR(26) NOT NULL,
+    type VARCHAR(256) NOT NULL,
+    type_definition BLOB,
+    schema_version VARCHAR(5) NOT NULL DEFAULT '1.0',
```
miparnisari, line 24:
```suggestion
    schema_version VARCHAR(5) NOT NULL DEFAULT '1.1',
```

assets/migrations/sqlite/005_initialize_schema.sql
```diff
@@ -0,0 +1,63 @@
...
+    serialized_protobuf LONGBLOB,
+    PRIMARY KEY (store, authorization_model_id, type)
+);
```
miparnisari, line 27:
```suggestion
CREATE TABLE authorization_model (
    store CHAR(26) NOT NULL,
    authorization_model_id CHAR(26) NOT NULL,
    schema_version VARCHAR(5) NOT NULL DEFAULT '1.1',
    serialized_protobuf LONGBLOB NOT NULL,
    PRIMARY KEY (store, authorization_model_id)
);
```

go.mod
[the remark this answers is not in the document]
miparnisari: https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist

we're good 👍

assets/migrations/sqlite/005_initialize_schema.sql
DanCech: Right now both of these columns are named in the insert query [here](https://github.com/openfga/openfga/blob/23e48b76f6cf24422c6f3517f5d5f9f7fa216ab8/pkg/storage/sqlcommon/sqlcommon.go#L579) so even though they aren't used we cannot drop them from this table without making some supporting changes.

`type_definition` is nullable so we should be able to just remove it from the insert, but `type` is declared not null so the query is explicitly inserting an empty string to keep MySQL and PostgreSQL happy.

One option may be to write migrations to drop those columns from MySQL and PostgreSQL, but there is also the problem of `type` being part of the primary key. It may not be possible to drop it from the primary key if there are entries in older databases that have duplicate values for `store` and `authorization_model_id`.

miparnisari: > Right now both of these columns are named in the insert query [here](https://github.com/openfga/openfga/blob/23e48b76f6cf24422c6f3517f5d5f9f7fa216ab8/pkg/storage/sqlcommon/sqlcommon.go#L579) so even though they aren't used we cannot drop them from this table without making some supporting changes.

I think it's okay to copy-paste this code into the SQLLite implementation and then modify it according to our needs :)

DanCech: The only concern I have with doing that is maintainability. Right now the 3 sql backends are *very* similar, so it's relatively easy to roll changes out across all 3 because it's largely a matter of copy+paste+tweak. The more differences creep in, the harder that's going to get long term. I'm happy to go ahead and make a special version of WriteAuthorizationModel for SQLite without the historical baggage if you think that's the best path forward.

pkg/storage/sqlite/sqlite_test.go
DanCech: Removed :)

DanCech:
Rebased again, I think I've hit all the other files that were identified and I went ahead and removed the `type` and `type_definition` columns for the SQLite backend.

I will look at what's needed to implement #1873 , and will open PRs for docs and helm charts.

assets/migrations/sqlite/005_initialize_schema.sql
```diff
@@ -0,0 +1,61 @@
...
+
+CREATE UNIQUE INDEX idx_tuple_ulid ON tuple (ulid);
+CREATE INDEX idx_reverse_lookup_user ON tuple (store, object_type, relation, _user);
+
+CREATE TABLE authorization_model (
+    store CHAR(26) NOT NULL,
+    authorization_model_id CHAR(26) NOT NULL,
+    schema_version VARCHAR(5) NOT NULL DEFAULT '1.0',
+    serialized_protobuf LONGBLOB,
```
miparnisari, line 23:
```suggestion
    schema_version VARCHAR(5) NOT NULL DEFAULT '1.1',
    serialized_protobuf LONGBLOB NOT NULL,
```

DanCech:
Updated the docs, but I'm not sure what the right path forward is for the helm charts. Running SQLite on k8s doesn't make a huge amount of sense unless you run a single-pod deployment with a PVC to hold the SQLite database files (and ensure that the rollout settings don't allow multiple concurrent pods). It's doable but the noise it would add to the helm chart may not be worth it.

Open to suggestions on how to move forward with that.

DanCech:
@miparnisari the most recent commit includes an implementation of #1873 for SQLite. I based the column names on the code in the linked PRs but can tweak them if you think that the "subject" naming makes more sense.

pkg/tuple/tuple.go
```diff
@@ -390,3 +390,27 @@ func ParseTupleString(s string) (*openfgav1.TupleKey, error) {
...
+	return userObjectType, userObjectID, u.GetRelation()
+}
+
+func ToUserParts(user string) (string, string, string) {
+	userObject, userRelation := SplitObjectRelation(user) // e.g. (person:bob, "") or (group:abc, member) or (person:*, "")
+
+	userObjectType, userObjectID := SplitObject(userObject)
+
+	return userObjectType, userObjectID, userRelation
+}
```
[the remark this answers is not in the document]
DanCech: it's just a couple of small tweaks to the code from #702, so you deserve most of the credit

jpadilla (approved):
✨ this is great, thank you!

DanCech:
Very cool! I'll rebase again and take care of those merge conflicts

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 * Stack trace when logging panics [#1904](https://github.com/openfga/openfga/pull/1904)
 * Throttling metric `throttled_requests_count` for observing the number of throttled requests for a given throttling configuration [#1863](https://github.com/openfga/openfga/pull/1863)
 * New metric on number of allowed vs. non-allowed Check responses [#1911](https://github.com/openfga/openfga/pull/1911)
+* SQLite support [#1615](https://github.com/openfga/openfga/pull/1615)
```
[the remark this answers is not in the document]
DanCech:
```suggestion
* SQLite support (beta) [#1615](https://github.com/openfga/openfga/pull/1615)
```

assets/migrations/sqlite/005_initialize_schema.sql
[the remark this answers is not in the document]
DanCech: goose doesn't actually care, it just wants them in order. The schema version is used internally in a couple of places though so I kept it at 5 to avoid having to make too many changes

assets/migrations/sqlite/005_initialize_schema.sql
```diff
@@ -0,0 +1,65 @@
...
+    user_type VARCHAR(7) NOT NULL,
+    ulid CHAR(26) NOT NULL,
+    inserted_at TIMESTAMP NOT NULL,
+    condition_name VARCHAR(256),
+    condition_context LONGBLOB,
+    PRIMARY KEY (store, object_type, object_id, relation, user_object_type, user_object_id, user_relation)
+);
+
+CREATE UNIQUE INDEX idx_tuple_ulid ON tuple (ulid);
+CREATE INDEX idx_reverse_lookup_user ON tuple (store, object_type, relation, user_object_type, user_object_id, user_relation);
```
miparnisari, line 19: If SQLLite supports partial indexes, could we bring these over from Postgres?

```go
CREATE INDEX idx_tuple_partial_user ON tuple (store, object_type, object_id, relation, _user) WHERE user_type = 'user';
CREATE INDEX idx_tuple_partial_userset ON tuple (store, object_type, object_id, relation, _user) WHERE user_type = 'userset';
```

These will make Check API more performant

DanCech: That should be possible, since SQLite does support partial indexes, I'll check out how that would work together with splitting `_user` up.

jpadilla (approved):

miparnisari:
We have a flaky test 😢 

image

CHANGELOG.md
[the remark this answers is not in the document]
miparnisari: Could you add this back please?

README.md
```diff
@@ -17,7 +17,7 @@ A high-performance and flexible authorization/permission engine built for develo
 
 OpenFGA is designed to make it easy for developers to model their application permissions and add and integrate fine-grained authorization into their applications.
 
-It allows in-memory data storage for quick development, as well as pluggable database modules. It currently supports PostgreSQL 14 and MySQL 8.
+It allows in-memory data storage for quick development, as well as pluggable database modules. It currently supports PostgreSQL 14, MySQL 8 and SQLite.
```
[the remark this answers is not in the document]
miparnisari: same here

miparnisari (approved):

jpadilla (approved):

PR #1571, opened 2024-04-25
feat: List objects dispatch throttling with seperate queues and configuration

poovamraj:
Implement Dispatch Throttling for ListObjects

❗ We are renaming the configurations to be more specific and the old configurations are deprecated now and will be removed in a future release

- OPENFGA_DISPATCH_THROTTLING_ENABLED -> OPENFGA_CHECK_DISPATCH_THROTTLING_ENABLED
- OPENFGA_DISPATCH_THROTTLING_FREQUENCY -> OPENFGA_CHECK_DISPATCH_THROTTLING_FREQUENCY
- OPENFGA_DISPATCH_THROTTLING_THRESHOLD -> OPENFGA_CHECK_DISPATCH_THROTTLING_THRESHOLD
- OPENFGA_DISPATCH_THROTTLING_MAX_THRESHOLD -> OPENFGA_CHECK_DISPATCH_THROTTLING_MAX_THRESHOLD

## Description
- We have abstracted dispatch throttling into it's own `Throttler`
- CheckResolver will use this Throttler with a delegate and resolver
- ListObjects uses throttler using the `Throttle` method
- We have a two throttlers which means a two queues and two configuration will be used (Seperate ones for Check and LIstObjects)

Previous PR - https://github.com/openfga/openfga/pull/1502

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

Issue #1409, opened 2024-02-29
Improve `TestListObjects` by removing flag `useCheckCache`


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Tests for the ListObjects API should not care how Check API obtains results (i.e. with/without cache)

### Describe the ideal solution

In the `pkg/server/test/list_objects.go` tests, we have `test.useCheckCache` for testing list objects with and without caching. However, we should see if this is actually needed. 

Important context: https://github.com/openfga/openfga/pull/1335#discussion_r1476797712 :

> We should not have this test flag in the first place. ListObjects shouldn't care how Check gets results.

> There is only one test that sets `useCheckCache: true`, and it's asserting (or trying, i dunno if it's succeeding from the looks it) that a call to ListObjects will set the Check cache and then a call to StreamingListObjects will use that. Please create a task to get rid of this flag and to write that test independently. That new test should assert that the second call has less `datastore_query_count` than the first

PR #1626, opened 2024-05-19
chore(refactor): Refactor and unify logic to construct the CheckResolver

poovamraj:
## Description
Currently the CheckResolver is built in 2 places - `server.go` and `graph.go`.  We have unified and moved this to a new file called `builder.go`

```
	checkBuilderOpts := []graph.CheckQueryBuilderOpt{
		graph.WithLocalCheckerOpts(localCheckOpts...),
		graph.WithCachedCheckResolverOpts(cacheOpts...)}
		
	checkResolver, closer := graph.NewCheckQueryBuilder(checkBuilderOpts...).NewLayeredCheckResolver(
		test.useCheckCache,
		false)
```

## References
https://github.com/openfga/openfga/pull/1571#discussion_r1588518113

## Review Checklist


internal/graph/builder.go
```diff
@@ -0,0 +1,107 @@
+package graph
+
+import (
+	"github.com/openfga/openfga/pkg/logger"
+)
+
+type CheckQueryBuilder struct {
```
jon-whit, line 7:
```suggestion
type CheckResolverBuilder struct {
```

willvedd:
@poovamraj Can you provide an example of how to build various resolvers in the description of the PR?

internal/graph/builder.go
```diff
@@ -0,0 +1,101 @@
...
+//			DispatchThrottlingCheckResolver |
+//				LocalChecker                   |
+//					CycleDetectionCheckResolver -|
+//
+// The returned CheckResolverCloser should be used to close all resolvers involved in the
+// composition after you are done with the CheckResolver.
+func (c *CheckResolverBuilder) NewLayeredCheckResolver(
+	cacheEnabled bool,
+	dispatchThrottlingEnabled bool,
+) (CheckResolver, CheckResolverCloser) {
```
miparnisari, line 61: 1. I'd rename this to `Build()`
2. Why are we passing in even more flags to this method? `Build()` should accept zero parameters, if we follow the convention from other libraries that also use this pattern.

pkg/server/server.go
```diff
@@ -451,30 +449,21 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
+		checkBuilderOpts = append(checkBuilderOpts, graph.WithCachedCheckResolverOpts(
 			graph.WithMaxCacheSize(int64(s.checkQueryCacheLimit)),
 			graph.WithLogger(s.logger),
 			graph.WithCacheTTL(s.checkQueryCacheTTL),
-		)
-		s.cachedCheckResolver = cachedCheckResolver
-
-		cachedCheckResolver.SetDelegate(localChecker)
-		cycleDetectionCheckResolver.SetDelegate(cachedCheckResolver)
+		))
```
miparnisari, line 466: These bits

```go
if s.checkQueryCacheEnabled {
    checkBuilderOpts = append(checkBuilderOpts, graph.WithCachedCheckResolverOpts(
   ...
}
if s.checkDispatchThrottlingEnabled {
   checkBuilderOpts = append(checkBuilderOpts, graph.WithDispatchThrottlingCheckResolverOpts(
   ...
}
````

I think should be moved to the new code you wrote (you'll have to add new builder functions for the `enabled` flags)

We don't need to preserve the logs - on a previous commit we are now logging every single server setting on startup (https://github.com/openfga/openfga/commit/fb0ed0d53ad022e3534d1ca181dc714e6d922ed9)

**Review by miparnisari — CHANGES_REQUESTED** [review]

File: internal/graph/builder.go [hunk 4]
```diff
@@ -0,0 +1,101 @@
+package graph
```
jon-whit, line 1: Can we please put this under a graph specific package. The vision is that we should eventually have (I wish we had started this way) `internal/graph/check` and `internal/graph/reverseexpand` etc.. Each query pattern ontop of the graph model warrants its own package. CheckResolverBuilder belongs to the internal (Check specific) graph API.

internal/graph/builder.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// NewLayeredCheckResolver constructs a CheckResolver that is composed of various CheckResolver layers.
+// Specifically, it constructs a CheckResolver with the following composition:
+//
+//	CycleDetectionCheckResolver  <-----|
+//		CachedCheckResolver              |
+//			DispatchThrottlingCheckResolver |
+//				LocalChecker                   |
+//					CycleDetectionCheckResolver -|
```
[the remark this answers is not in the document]
jon-whit: The fact that we impost constraints on the ordering negates the value of the arbitrary circular list. It is a circular list, but it's one that is specifically constructed with a particular order. I don't see anything wrong with being explicit about the order in which the builder will construct the circular list (if certain resolvers are enabled).

internal/graph/builder.go
jon-whit: Ideal usage pattern:

```go
import (
  "github.com/openfga/openfga/internal/graph/check"
)

checkResolverBuilder := check.NewCheckResolverBuilder(
    WithCacheEnabled(),
    WithDispatchThrottlingEnabled(),
})

checkResolver := checkResolverBuilder.Build()

checkResolver.ResolveCheck(...)
```

@miparnisari does this align with your ideal outcomes here?

pkg/server/server.go
```diff
@@ -451,30 +449,21 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 
 	// below this point, don't throw errors or we may leak resources in tests
 
-	cycleDetectionCheckResolver := graph.NewCycleDetectionCheckResolver()
-	s.checkResolver = cycleDetectionCheckResolver
-
-	localChecker := graph.NewLocalChecker(
+	var checkBuilderOpts []graph.CheckQueryBuilderOpt
```
jon-whit, line 452:
```suggestion
	checkBuilderOpts := []graph.CheckQueryBuilderOpt{
	    graph.WithLocalCheckerOpts(
	      graph.WithResolveNodeBreadthLimit(s.resolveNodeBreadthLimit),
	    ),
	}
```
Just initialize the variable with the initial value instead of the immediate append afterwards.

internal/graph/builder.go
miparnisari: Yes

internal/graph/builder_test.go
```diff
@@ -0,0 +1,135 @@
...
+			} else {
+				if test.cacheEnabled {
+					_, ok = cacheCheckResolver.GetDelegate().(*LocalChecker)
+					require.True(t, ok)
+				} else {
+					_, ok = cycleDetectionCheckResolver.GetDelegate().(*LocalChecker)
+					require.True(t, ok)
+				}
+			}
+		})
```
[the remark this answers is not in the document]
poovamraj: @miparnisari Looks like the `NewCircularListOfResolvers` will be useful only for testing purpose right? In that case I think it is better to introduce only at test level to make testing easier

internal/graph/builder.go
```diff
@@ -0,0 +1,101 @@
...
+type CheckResolverBuilder struct {
+	localCheckerOpts       []LocalCheckerOption
+	cacheOpts              []CachedCheckResolverOpt
+	dispatchThrottlingOpts []DispatchThrottlingCheckResolverOpt
+
+	cycleDetectionCheckResolver     *CycleDetectionCheckResolver
+	cachedCheckResolver             *CachedCheckResolver
+	dispatchThrottlingCheckResolver *DispatchThrottlingCheckResolver
+	localCheckResolver              *LocalChecker
+}
```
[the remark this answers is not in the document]
poovamraj: In your example you still have this right

```
	first := &CachedResolver{9, nil}
	second := &DispatchThrottlerResolver{10000, nil}
	first.SetDelegate(second)
	second.SetDelegate(first)
```

This wouldn't be a runtime change right? We do the same construction right now too.  

I understand this could be easier if we do `SetDelegate` as well using the CircularList then construction might be easier

internal/graph/builder_test.go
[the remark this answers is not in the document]
poovamraj: I have now used an array to represent our order logic, it feels much more intuitive than the `Ring` structure and achieves everything we want

internal/graph/builder.go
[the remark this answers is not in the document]
poovamraj: I have used an array instead to represent our construction order. This simplifies the construction logic, but we should be aware how we build it

pkg/server/server.go
poovamraj: But wouldn't that mean we have to pass the server configs there. I feel that would break Single responsibility.

internal/graph/builder.go
poovamraj: Agreed, but that would mean the whole of the `graph` package which doesn't have anything else other than check logics should be in this folder, Let me make this change towards the end of the PR to keep it neater for diff. Ideally I would do this in another PR related to restructuring but let me do it towards the end of this PR.

poovamraj:
@jon-whit @miparnisari This refactor should address our needs but I am slightly worried that the order in which the opts are provided is how the check resolver will be constructed is not explicit enough (documented). Check it out and let me know.

Also `LocalChecker` order could change depending on how the construction happens. I did this to be consistent with other resolvers are handled. Check this out and let me know

internal/graph/builder_test.go
```diff
@@ -0,0 +1,122 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			opts := []CheckQueryBuilderOpt{}
+			if test.cacheEnabled {
+				opts = append(opts, WithCachedCheckResolver())
+			}
+			if test.dispatchThrottlingEnabled {
+				opts = append(opts, WithDispatchThrottlingCheckResolver())
+			}
+			opts = append(opts, WithLocalChecker())
```
miparnisari, line 58: The fact that you had to write this line for every test makes me think that this should be made by default in the actual code. I can't think of a scenario today where the LocalChecker would _not_ be wanted.

internal/graph/builder_test.go
```diff
@@ -0,0 +1,122 @@
...
+		dispatchThrottlingEnabled bool
+		expectedResolverOrder     []CheckResolver
+	}
+
+	tests := []Test{
+		{
+			name:                      "when_nothing_is_enabled",
+			cacheEnabled:              false,
+			dispatchThrottlingEnabled: false,
+			expectedResolverOrder:     []CheckResolver{&CycleDetectionCheckResolver{}, &LocalChecker{}},
```
miparnisari, line 27: I like this a lot more 🙇‍♀️

internal/graph/builder.go
```diff
@@ -0,0 +1,68 @@
...
+// Build constructs a CheckResolver that is composed of various CheckResolver layers.
+// Specifically, it constructs a CheckResolver with the order in which the Opts are provided.
+// CycleDetectionCheckResolver is always the first resolver in the composition.
+
+//	CycleDetectionCheckResolver  <-----|
+//		[...Other resolvers depending on the opts order]
+//			CycleDetectionCheckResolver
+//
+// The returned CheckResolverCloser should be used to close all resolvers involved in the
+// composition after you are done with the CheckResolver.
```
miparnisari, line 50:
```suggestion
// Build constructs a CheckResolver that is composed of various CheckResolvers in the manner of a circular linked list.  CycleDetectionCheckResolver is always the first resolver in the composition and the last resolver added will always point to it.

//	CycleDetectionCheckResolver  <----------------------|
//		[...Other resolvers depending on the opts order]
//			CycleDetectionCheckResolver -------^
//
// The returned CheckResolverCloser should be used to close all resolvers involved in the list.
```

internal/graph/builder_test.go
```diff
@@ -0,0 +1,122 @@
...
+	require.Equal(t, 1*time.Second, cacheCheckResolver.cacheTTL)
+	require.Equal(t, int64(2), cacheCheckResolver.maxCacheSize)
+	require.Equal(t, uint32(3), dispatchCheckResolver.config.DefaultThreshold)
+	require.Equal(t, uint32(4), dispatchCheckResolver.config.MaxThreshold)
+	require.Equal(t, mockThrottler, dispatchCheckResolver.throttler)
+	require.Equal(t, uint32(6), localCheckResolver.maxConcurrentReads)
+	require.Equal(t, uint32(7), localCheckResolver.concurrencyLimit)
+}
+
+// TODO any good way to test the close method?
```
miparnisari, line 118: Create a chain that only has one resolver in it, `mockResolver`. Then close `checkResolver.Close()` and assert that `mockResolver.EXPECT().Close().Times(1)`.

You already did this, kinda, in line 100 :)

pkg/server/test/list_objects.go
```diff
@@ -490,19 +491,19 @@ condition condition1(x: int) {
...
-				[]graph.DispatchThrottlingCheckResolverOpt{},
-			)
+			localCheckOpts := []graph.LocalCheckerOption{
+				graph.WithResolveNodeBreadthLimit(100),
+				graph.WithMaxConcurrentReads(30),
+			}
+			cacheOpts := []graph.CachedCheckResolverOpt{
+				graph.WithMaxCacheSize(100),
+				graph.WithCacheTTL(10 * time.Second),
+			}
```
miparnisari, line 501: Is the new code equivalent to the old code? You aren't using `test.useCheckCache` anymore.

The fact that no tests failed makes me think we may have other issues too. 🤔 

In any case, I don't think we should have that flag, see https://github.com/openfga/openfga/issues/1409

internal/graph/builder_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/openfga/openfga/internal/mocks"
+)
+
+func TestNewLayeredCheckResolver(t *testing.T) {
```
miparnisari, line 14:
```suggestion
func TestNewCheckResolverBuilder(t *testing.T) {
```

pkg/server/server.go
miparnisari: In my view, today we have the responsibility split - the server and the builder both kind of know how to build the chain in the right order. My idea was that only the builder knows about the right order, which would mean it would have to accept boolean flags to know which resolvers to add to the chain.

internal/graph/builder_test.go
poovamraj: It's been sometime but I remember implementing it that way and feeling it wasn't explicit enough. It was abstracted and didn't give the full picture of the builder

internal/graph/builder_test.go
poovamraj: Yea but wanted to avoid the mock and have actual resolvers call to test the usecase, guess this should do

miparnisari: It's fine, you don't need to test the individual close methods, just that they are called

internal/graph/builder_test.go
miparnisari: 🤔 to whom should it be explicit? you can say in the godoc of the builder that the last element in the chain is always LocalChecker

poovamraj: We are keeping this for the consistency of the builder methods

internal/graph/builder_test.go
poovamraj: Couldn't find a ideal way, leaving the comments for now

pkg/server/test/list_objects.go
poovamraj: Resolving this in favour of https://github.com/openfga/openfga/issues/1409. Will be checked when the above issue is closed.

miparnisari (dismissed):

miparnisari (approved):

miparnisari (approved):

miparnisari (approved):

internal/graph/graph.go
```diff
@@ -416,60 +416,3 @@ func (g *RelationshipGraph) getRelationshipEdgesWithTargetRewrite(
...
-//
-//	CycleDetectionCheckResolver  <-----|
-//		CachedCheckResolver              |
-//			DispatchThrottlingCheckResolver |
-//				LocalChecker                   |
-//					CycleDetectionCheckResolver -|
-//
-// The returned CheckResolverCloser should be used to close all resolvers involved in the
-// composition after you are done with the CheckResolver.
-func NewLayeredCheckResolver(
```
miparnisari, line 431: For other reviewers: the point of this PR was to get rid of this function, which has duplicate logic as the `NewServer` function: https://github.com/openfga/openfga/blob/518606928ded775a86a6ed69c4e6faa245e9b372/pkg/server/server.go#L496-L546

internal/graph/builder.go
```diff
@@ -0,0 +1,84 @@
...
+// Build constructs a CheckResolver that is composed of various CheckResolvers in the manner of a circular linked list.
+// The resolvers should be added from least resource intensive to most resource intensive.
+// CycleDetectionCheckResolver is always the first resolver in the composition and LocalChecker is always the last.
+//
+//	CycleDetectionCheckResolver  <----------------------|
+//		[...Other resolvers depending on the opts order]
+//			LocalChecker    ----------------------------^
+//
+// The returned CheckResolverCloser should be used to close all resolvers involved in the list.
+func (c *CheckResolverOrderedBuilder) Build() (CheckResolver, CheckResolverCloser) {
```
jpadilla, line 55: thought: what do you think about having this function just be part of the constructor logic? You'll always call the constructor followed by Build. Calling Build more than once on an instance likely would yield incorrect behavior right now.

jpadilla (approved):

internal/graph/builder.go
miparnisari: Good catch, `Build` is not idempotent.

The issue with merging them is that we won't be able to write unit tests that assert on the value of `builder.resolvers` 🤔

jpadilla (approved):

PR #1627, opened 2024-05-19
feat(list-users): Implement dispatch count for list users

poovamraj:
## Description
Implemented DispatchCount metrics for list users. This will be useful when throttling is implemented. Since we don't have the throttling PR merged, that hasn't been implemented but it should be simple since we have the counter logic now.

## Review Checklist

### DIFF
--- a/tests/listusers/listusers_test.go
+++ b/tests/listusers/listusers_test.go
@@ -97,9 +97,13 @@ func TestListUsersLogs(t *testing.T) {
 	schema 1.1
 type user
 
+type group
+  relations
+    define member: [user]
+
 type document
   relations
-	define viewer: [user]`)
+	define viewer: [group#member]`)
 
 	writeModelResp, err := client.WriteAuthorizationModel(context.Background(), &openfgav1.WriteAuthorizationModelRequest{
 		StoreId:         storeID,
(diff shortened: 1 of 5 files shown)


poovamraj:
@willvedd some tests are failing. Can we pair over this tomorrow?

poovamraj:
Fixed the failing tests. It was due to uninitialized value in reflexive userset edge case.

tests/listusers/listusers_test.go
```diff
@@ -259,7 +259,8 @@ type document
 			require.NotEmpty(t, fields["trace_id"])
 			if !test.expectedError {
 				require.NotEmpty(t, fields["datastore_query_count"])
-				require.Len(t, fields, 13)
+				require.Empty(t, fields["dispatch_count"])
```
willvedd, line 262: Shouldn't this be _not_ empty?
```suggestion
				require.NotEmpty(t, fields["dispatch_count"])
```

willvedd (changes requested):

tests/listusers/listusers_test.go
willvedd: This was incorrectly considered empty because it was zero. I fixed this test to incur a DB query and dispatch call. These assertions are now a bit stronger.

PR #1631, opened 2024-05-21
feat: configurable max condition evaluation cost

cmmoran:
Configurable max evaluation cost for conditions over hard-coded value.

## Description

The `DefaultMaxConditionEvaluationCost` is hard-coded at 100. This is the minimum allowable value in order to prevent API compatibility problems with Conditions. This PR allows the max condition evaluation cost to be set by environment variable, CLI parameter, or within the openfga configuration file (as already defined). As the run configuration is not available outside of server startup, this value is exposed via an exported function in the config package. This function ensures that the returned value is the greater of `DefaultMaxConditionEvaluationCost` and the configured max condition evaluation cost.

## Review Checklist


jpadilla:
@cmmoran thanks for the pull request. Did you run into specific issues with the default value? If so, can you share more about your use case?

cmd/run/run.go
```diff
@@ -185,6 +185,8 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Uint32("max-concurrent-reads-for-check", defaultConfig.MaxConcurrentReadsForCheck, "the maximum allowed number of concurrent datastore reads in a single Check query. A high number will consume more connections from the datastore pool and will attempt to prioritize performance for the request at the expense of other queries performance.")
 
+	flags.Uint64("max-condition-evaluation-cost", defaultConfig.MaxConditionEvaluationCost, "defines the maximum cost for CEL condition evaluation before a request will return an error to the client")
```
jon-whit, line 188: Please use the same description as the one provided in `.config-schema.json`. They need not be different.

side note - we've discussed constructing these flags off of the `.config-schema.json` in the future as well, so this is why we strive to keep them consistent relative to one another.

internal/server/config/config.go
```diff
@@ -231,6 +231,10 @@ type Config struct {
 	// Check queries
 	MaxConcurrentReadsForCheck uint32
 
+	// MaxConditionEvaluationCost defines the maximum cost for CEL condition evaluation before a request
```
jon-whit, line 234: Same godoc as the description in the `.config-schema.json` please.

cmd/run/run_test.go
jon-whit, line 1: Please also add a test assertion to https://github.com/openfga/openfga/blob/6ff5306016d2ba6010a85d60f4d817b59aa777bd/cmd/run/run_test.go#L1267 so that we can confirm/verify that the env vars map to the config field as well. This ensures we bound the env var with the config mapping appropriately.

internal/server/config/config_test.go
```diff
@@ -250,6 +250,13 @@ func TestVerifyConfig(t *testing.T) {
 func TestDefaultMaxConditionValuationCost(t *testing.T) {
 	// check to make sure DefaultMaxConditionEvaluationCost never drops below an explicit 100, because
 	// API compatibility can be impacted otherwise
+	t.Run("max_condition_evaluation_cost_too_low", func(t *testing.T) {
```
jon-whit, line 253: Please also add a test where max condition evaluation cost is above the default.

jon-whit (changes requested):
Left some comments needing to be addressed.

cmd/run/run_test.go
cmmoran: Nice catch. Thank you.

cmd/run/run_test.go
```diff
@@ -1291,6 +1296,7 @@ func TestRunCommandConfigIsMerged(t *testing.T) {
 		require.Equal(t, "1ms", viper.GetString("dispatch-throttling-frequency"))
 		require.Equal(t, "120", viper.GetString("dispatch-throttling-threshold"))
 		require.Equal(t, "130", viper.GetString("dispatch-throttling-max-threshold"))
+		require.Equal(t, "120", viper.GetString("max-condition-evaluation-cost"))
```
jon-whit, line 1299: Here we are doing `viper.GetString(flag)`, but below we are doing `viper.GetUint64(configField)`. Why the difference? Shouldn't they remain consistent? Are we sure the `viper.GetUint64(configField)` works as well? It would be good to add a test for that as well.

internal/server/config/config_test.go
```diff
@@ -250,7 +250,22 @@ func TestVerifyConfig(t *testing.T) {
...
+	})
+	t.Run("max_condition_evaluation_cost_valid", func(t *testing.T) {
+		cfg := DefaultConfig()
+		cfg.MaxConditionEvaluationCost = 120
+
+		err := cfg.Verify()
+		require.NoError(t, err)
+	})
 	require.GreaterOrEqual(t, DefaultMaxConditionEvaluationCost, 100)
+	require.GreaterOrEqual(t, MaxConditionEvaluationCost(), uint64(100))
```
jon-whit, line 268:
```suggestion
	require.GreaterOrEqual(t, MaxConditionEvaluationCost(), DefaultMaxConditionEvaluationCost)
```

jon-whit:
@cmmoran thanks for the follow up changes. Left 2 last comments - LGTM otherwise. Thanks!

cmd/run/run_test.go
cmmoran: This was a case of "doing the same thing being done the line above it". Nice catch. Added both.

internal/server/config/config_test.go
cmmoran: Merged the two so the test now makes more sense.

cm-sentinel:
All PR requests have been addressed. Is there anything else I can do to facilitate this PR?

cmmoran:
> @cmmoran thanks for the follow up changes. Left 2 last comments - LGTM otherwise. Thanks!

Let me know if there's anything else I can do to facilitate this PR.

internal/server/config/config_test.go
```diff
@@ -250,7 +250,21 @@ func TestVerifyConfig(t *testing.T) {
...
+		require.Error(t, err)
+	})
+	t.Run("max_condition_evaluation_cost_valid", func(t *testing.T) {
+		cfg := DefaultConfig()
+		cfg.MaxConditionEvaluationCost = 120
+
+		err := cfg.Verify()
+		require.NoError(t, err)
+	})
+	require.GreaterOrEqual(t, MaxConditionEvaluationCost(), uint64(DefaultMaxConditionEvaluationCost))
```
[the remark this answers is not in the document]
jon-whit: Expecting `>=` doesn't really ensure it's not still the default. If you explicitly set it to 120 you expect nothing other than 120. So let's just assert that.

internal/server/config/config_test.go
```diff
@@ -278,7 +278,21 @@ func TestVerifyConfig(t *testing.T) {
...
+		require.Error(t, err)
+	})
+	t.Run("max_condition_evaluation_cost_valid", func(t *testing.T) {
+		cfg := DefaultConfig()
+		cfg.MaxConditionEvaluationCost = 120
+
+		err := cfg.Verify()
+		require.NoError(t, err)
+	})
+	require.Equal(t, 120, MaxConditionEvaluationCost()
```
jon-whit, line 295:
```suggestion
	require.Equal(t, 120, MaxConditionEvaluationCost())
```
You're missing a closing `)`. The tests don't build at the moment.

jon-whit (dismissed):
Tests aren't building due to a missing `)` in the MaxConditionEvalCost test.

internal/server/config/config_test.go
cmmoran: Sorry. This is what I get for using the "commit suggestion" button.

miparnisari (approved):

PR #1636, opened 2024-05-22
fix: log request id in streaming apis

miparnisari:
## Description

Log `request_id` on all calls to Streaming APIs.

Before / after:

image

## Review Checklist


cmd/run/run.go
```diff
@@ -452,12 +452,13 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
...
 			[]grpc.StreamServerInterceptor{
-				grpc_recovery.StreamServerInterceptor(
+				grpc_recovery.StreamServerInterceptor( // panic middleware must be 1st in chain
 					grpc_recovery.WithRecoveryHandlerContext(
 						recovery.PanicRecoveryHandler(s.Logger),
 					),
 				),
-				requestid.NewStreamingInterceptor(),
+				grpc_ctxtags.StreamServerInterceptor(), // needed for logging
+				requestid.NewStreamingInterceptor(),    // add request_id to ctxtags
```
miparnisari, line 461: same as 449 and 450

sergiught (approved):

PR #1639, opened 2024-05-24
test: stabilize ReadPage storage tests

miparnisari:
## Description
- Added a test helper `ConvertTuplesToTupleKeys`
- Refactor test `ReadAllTuplesTest` to use the helper
- Change test assertions in `readPage_pagination_works_properly_with_filter` and `readPage_pagination_works_properly_without_filter` so that they don't expect a tuple order in the response.


pkg/storage/storage.go
```diff
@@ -103,6 +103,7 @@ type RelationshipTupleReader interface {
 	// ReadPage functions similarly to Read but includes support for pagination. It takes
 	// mandatory pagination options (pageSize can be zero :/)
 	// and returns a slice of tuples along with a continuation token. This token can be used for retrieving subsequent pages of data.
+	// There is NO guarantee on the order returned on the iterator.
```
adriantam, line 106: I am not sure if we can reword this a bit - the order will be identical in the same datastore implementation.  (Otherwise, continuation token will not work).  However, there is no guarantee what that order is based on (whether it is based on order of insert, etc.)

miparnisari: Updated comment, let me know if it's a bit more clear

adriantam (approved):

PR #1660, opened 2024-06-04
fix(test): test should not run with pageSize = 0

adriantam:
## Description
page size of 0 is not valid. Thus, tests should never run with this option.

## References
API validation - https://github.com/openfga/api/blob/394820dfc109972eb6366118718e212fc87412e3/openfga/v1/openfga_service.proto#L1381

Closes https://github.com/openfga/openfga/issues/1661

## Review Checklist

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -53,9 +53,7 @@ func RelationshipTupleReaderFromContext(ctx context.Context) (RelationshipTupleR
 	return reader, ok
 }
 
-// PaginationOptions holds the settings for pagination in data retrieval operations. It defines
-// the number of items to be included on each page (PageSize) and a marker from where to start
-// the page (From).
+// PaginationOptions should not be instantiated directly. Use NewPaginationOptions.
 type PaginationOptions struct {
 	PageSize int
 	From     string
@@ -66,7 +64,7 @@ type PaginationOptions struct {
 // it uses DefaultPageSize.
 func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
 	pageSize := DefaultPageSize
-	if ps != 0 {
+	if ps <= 0 {
 		pageSize = int(ps)
 	}
 
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -236,6 +236,8 @@ type ChangelogBackend interface {
 	// You can optionally provide a filter to filter out changes for objects of a specific type.
 	// The horizonOffset should be specified using a unit no more granular than a millisecond
 	// and should be interpreted as a millisecond duration.
+	// If no changes are found, it will return storage.ErrNotFound.  In this case,
+	// the API layer will return the original if it was provided.
 	ReadChanges(
 		ctx context.Context,
 		store,
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -236,8 +236,7 @@ type ChangelogBackend interface {
 	// You can optionally provide a filter to filter out changes for objects of a specific type.
 	// The horizonOffset should be specified using a unit no more granular than a millisecond
 	// and should be interpreted as a millisecond duration.
-	// If no changes are found, it will return storage.ErrNotFound.  In this case,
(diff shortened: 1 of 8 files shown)


pkg/storage/test/tuples.go
```diff
@@ -785,7 +785,7 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		require.NoError(t, err)
 		require.Nil(t, tp.GetKey().GetCondition())
 
-		changes, _, err := datastore.ReadChanges(ctx, storeID, "", storage.PaginationOptions{}, 0)
+		changes, _, err := datastore.ReadChanges(ctx, storeID, "", storage.NewPaginationOptions(0, ""), 0)
```
miparnisari, line 788: Can you please also:

1. Change all usages of `storage.NewPaginationOptions(0,...)`?
2. Fix `NewPaginationOptions`?

```go
// NewPaginationOptions creates a new [PaginationOptions] instance
// with a specified page size and continuation token. If the input page size is empty,
// it uses DefaultPageSize.
func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
	pageSize := DefaultPageSize
	if ps != 0 {
		pageSize = int(ps)
	}
```

it should be 


```go
// NewPaginationOptions creates a new [PaginationOptions] instance
// with a specified page size and continuation token. If the input page size is less than 1,
// it uses DefaultPageSize.
func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
	pageSize := DefaultPageSize
	if ps >= 1 0 {
		pageSize = int(ps)
	}
```

3. Finally, update the Godoc here

```
// PaginationOptions holds the settings for pagination in data retrieval operations. It defines
// the number of items to be included on each page (PageSize) and a marker from where to start
// the page (From).
```

to

```
// PaginationOptions should not be instantiated directly. Use NewPaginationOptions.
```

pkg/storage/storage.go
```diff
@@ -66,7 +64,7 @@ type PaginationOptions struct {
 // it uses DefaultPageSize.
 func NewPaginationOptions(ps int32, contToken string) PaginationOptions {
 	pageSize := DefaultPageSize
-	if ps != 0 {
+	if ps <= 0 {
```
miparnisari, line 67:
```suggestion
	if ps > 0 {
```

pagination options can not be negative.

pkg/storage/test/tuples.go
```diff
@@ -41,7 +41,7 @@ func ReadChangesTest(t *testing.T, datastore storage.OpenFGADatastore) {
 		err := datastore.Write(ctx, storeID, nil, []*openfgav1.TupleKey{tk1, tk2})
 		require.NoError(t, err)
 
-		changes, continuationToken, err := datastore.ReadChanges(ctx, storeID, "", storage.PaginationOptions{PageSize: 1}, 0)
+		changes, continuationToken, err := datastore.ReadChanges(ctx, storeID, "", storage.NewPaginationOptions(1, ""), 0)
```
miparnisari, line 44: Like we discussed offline, can you ammend the godoc here https://github.com/openfga/openfga/blob/883fb6c63bea84778035cac383d2baf132237b42/pkg/storage/storage.go#L236-L248

to say that the implementation must return `storage.ErrNotFound` and no continuation token when there are no more changes left? (And any other details that you may have noticed during development)

miparnisari (changes requested):

pkg/storage/storage.go
```diff
@@ -238,6 +236,8 @@ type ChangelogBackend interface {
 	// You can optionally provide a filter to filter out changes for objects of a specific type.
 	// The horizonOffset should be specified using a unit no more granular than a millisecond
 	// and should be interpreted as a millisecond duration.
+	// If no changes are found, it will return storage.ErrNotFound.  In this case,
+	// the API layer will return the original if it was provided.
 	ReadChanges(
```
miparnisari, line 241:
```suggestion
	// If no changes are found, it should return storage.ErrNotFound and an empty continuation token.
	ReadChanges(
```

PR #1651, opened 2024-05-30
feat: add `ReadRelationshipTuples` storage signature

jon-whit:
## Description
Add storage interface changes to the `RelationshipTupleReader` interface in preparation for changes to the dispatch behavior for performance improvements, and implement the interface change for the supported storage implementations.

The primary changes at the storage layer needed to support improvements to userset dispatch is that we need to be able to lookup relationship tuples for multiple object_ids and with multiple subject filters in the same request. Today we make 1 request per object_id and subject filter permutation.

As an example, if we need to determine if a subject, for example `user:will`, is a member of 1 or more groups or if the typed public wildcard `user:*` is assigned to any of those groups, then we should be able to do so in a single query. For example,

## Batch Direct Object Relationship Queries (needed for `checkDirect` as a result of batch dispatch)
The query pattern here is that we want to establish which object ids of a specific list are related to either the subject `user:jon` or `user:*`.

```go
iter, err := ReadRelationshipTuples(..., <storeID>, ReadRelationshipTuplesFilter{
    ObjectType: "group",
    ObjectIDs: []string{"eng", "marketing", "sales", "product"},
    Relation: "member",
    SubjectFilters: []SubjectFilter{
        {SubjectType: "user", SubjectIDs: []string{"jon"}, SubjectRelation: ""},
        {SubjectType: "user", SubjectIDs: []string{"*"}, SubjectRelation: ""},
    },
})
```

## Batch Direct Subject Relationship Queries (needed for `checkTTU` as a result of batch dispatch)
The query pattern here is that we need to find all of the `SubjectIDs` of those subjects who are related to the tuplesets (e.g. `parent`) of `document:readme` and `document:roadmap`, and we need to be able to do so in 1 query.
```
type user

type org
  relations
    define viewer: [user]

type folder
  relations
    define viewer: [user]

type document
  relations
    define parent: [folder, org]
    define viewer: viewer from parent
```

```go
iter, err := ReadRelationshipTuples(..., <storeID>, ReadRelationshipTuplesFilter{
    ObjectType: "document",
    ObjectIDs: []string{"readme", "roadmap"},
    Relation: "parent",
    SubjectFilters: []SubjectFilter{
        {SubjectType: "folder", SubjectRelation: ""},
        {SubjectType: "org", SubjectRelation: ""},
    },
})
```

These two query patterns above can get put together to reduce many lookups when evaluating dispatched subqueries. For example, let's assume we're using the model above and have the following tuples:

* document:x#parent@folder:1
[...]

PR #1665, opened 2024-06-05
perf: improve dispatch with batch dispatch composition

jon-whit:
## Description
This PR lays the foundation for improving the way we dispatch Check subproblems. Specifically, it introduces some dispatch interface structures and transitions dispatch semantics from a single lookup `(object, relation, user/subject)` to batch dispatch semantics `(objectType, []objectIDs, relation, user/subject)`. This change of algorithm behavior is hugely important for improving Check query latency for widely nested sets. For example, consider the following model and tuples:

```
type user

type group
  relations
```
* group:x#member@group:1#member
* group:x#member@group:2#member
* ...
* group:x#member@group:N#member

A query of the form `Check(group:x#member@user:jon)` in our algorithm today will dispatch 1 subproblem per nested group. If `N` in this case is large, then that's a lot of outbound queries just to lookup if `user:jon` is a direct member of any one of those `N` objects. Instead, we can batch that into a single query of the form:

```sql
SELECT ... FROM tuples WHERE object_type='group' AND object_id IN ('1', '2', ..., 'N') AND relation='member' AND _user="user:jon"`
```
and reduce N lookups to a single lookup in these types of scenarios.

The purpose of this PR is to start with the baseline implementation of direct relationships and we will layer on the other modeling rewrite rules (ComputedUserset, TTU, union, intersection, exclusion) in subsequent PRs. These subsequent PRs will build off the structures and composition introduced herein.

## References
#1651 

## Review Checklist


internal/graph/check/check_test.go
jon-whit, line 1: Note - I need to continue improving these tests as we do not accurately report the dispatchCount and datastoreQueryCount at this time.

I'll work on that in subsequent commits, but I wanted to get this up for initial 👀

pkg/types/tuple.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (t TypedWildcardSubject) isSubject() {}
+
+// Subject represents an FGA user/subject.
+//
+// A Subject can be any one of an [Object], [TypedWildcardSubject], or
+// [Userset].
+//
+// See https://openfga.dev/docs/concepts#what-is-a-user
+type Subject interface {
```
jon-whit, line 86: How do we feel about using this internal definition as a good point to start this gradual rename in internal code from `user` to `subject` so that we put an end to the stuttering of the overloaded term "user"?

Happy to rename fields like `subject_type`, `subject_id`, and `subject_relation` to `user_type`, `user_id`, `user_relation` if we feel it's not best, but given this is an internal API contract I thought it might be nice to start evolving that direction.

PR #1332, opened 2024-02-01
test: add more tests for ReadPage correctness

miparnisari:
This is a follow-up of https://github.com/openfga/openfga/pull/1325 with more tests 😆 

Two tests were missing, filter by user and object type, for both `Read` and `ReadPage`.

PR #1680, opened 2024-06-12
test: refactor Read, ReadPage tests

miparnisari:
## Description
**Goal 1**: Address the TODOs introduced  in https://github.com/openfga/openfga/pull/1332

- https://github.com/openfga/openfga/blob/3204eaca4f2cba78b7a19445866cf5eab7a10cde/pkg/storage/test/storage.go#L40-L46
- https://github.com/openfga/openfga/blob/3204eaca4f2cba78b7a19445866cf5eab7a10cde/pkg/server/test/read.go#L21-L23

**Goal 2**: increase confidence in our test coverage, specifically in the correct application of filters (see `ReadAndReadPageTest`), and in the correct length returned (see `lots_of_writes_and_read_returns_everything`)


pkg/server/commands/read_test.go
adriantam, line 1: maybe explain why moving from pkg/server/test/read.go to pkg/server/commands/read_test.go?

pkg/storage/test/tuples.go
```diff
@@ -1179,511 +1138,172 @@ func ReadStartingWithUserTest(t *testing.T, datastore storage.OpenFGADatastore)
...
+			},
+		},
+		`filter_by_objectID_and_user`: {
+			filter: tuple.NewTupleKey("document:1", "", "user:github.com|bob@test.com"),
+			expectedTuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("document:1", "reader", "user:github.com|bob@test.com"),
+				tuple.NewTupleKey("document:1", "writer", "user:github.com|bob@test.com"),
+			},
+		},
+		`filter_by_objectID_with_special_character`: {
```
adriantam, line 1269: Did we drop filter_by_object_type_and_user_with_special_character?

pkg/server/commands/read_test.go
miparnisari: Sure thing. That file contains a mixture of tests:
- Tests that check on some API filters, I moved them to `tuples.go`
- Tests that validate the input I rewrote them as unit tests because they don't touch the datastore layer at all.
- `ReadAllTuplesTest` is checking the continuation token, I also moved to `tuples.go`
- `ReadAllTuplesInvalidContinuationTokenTest` is asserting that if the `q.encoder` can't decode the token, the API throws an error. I rewrote this as a unit test here.

pkg/storage/test/tuples.go
miparnisari: No, it's the test `filter_by_objectID_with_special_character`. Notice how both user and object have `|` 😉

adriantam: I mean the filter part.  Here, we only filter by document that has special character.  Do we have any request that filter by type + user that has special character?

adriantam: never mind I see the test above.

PR #1688, opened 2024-06-14
storage/postgres: add `NewWithDB` to allow accepting an established `*sql.DB`

unknwon:
## Description

We're importing/embedding the OpenFGA server directly in our Go binary, which constructs the `*sql.DB` from the caller because our services run on GCP Cloud Run, database connection "must" be established with GPC IAM-capable, thus no connection string could be used.

Therefore, we are making a slight change in the `storage` package to allow accepting a `*sql.DB` as an argument directly.

## Review Checklist
- [ ] ~~I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]~~
- [ ] ~~I have added tests to validate that the change in functionality is working as expected~~ should be covered by the current tests


pkg/storage/postgres/postgres.go
```diff
@@ -99,8 +103,8 @@ func New(uri string, cfg *sqlcommon.Config) (*Postgres, error) {
 	policy := backoff.NewExponentialBackOff()
 	policy.MaxElapsedTime = 1 * time.Minute
 	attempt := 1
-	err = backoff.Retry(func() error {
-		err = db.PingContext(context.Background())
+	err := backoff.Retry(func() error {
+		err := db.PingContext(context.Background())
```
miparnisari, line 107: I'm curious, why this change, what does it improve?

unknwon: Hi, this change improves nothing but to make sure the code can actually compile because this is the first declaration of the variable `err` in the block (after breaking into two functions).

unknwon:
Hi,

> * How/where are you planning to use this new function? Are you going to send a follow-up PR? Or is this just a refactor?

We are already using this function in our codebase, and I anticipant no use case (for now) within the OpenFGA codebase, since all of call sites would continue to use the current `New` constructor function. Therefore this is just a necessary refactor.

> * Could we apply the same change to the Postgres adapter please?

Where exactly is the "Postgres adapter" you're referring to? Could you point me to a link/file?

pkg/storage/postgres/postgres.go
miparnisari: Oh right haha. Thanks

miparnisari:
> > * Could we apply the same change to the Postgres adapter please?
> 
> Where exactly is the "Postgres adapter" you're referring to? Could you point me to a link/file?

Here is the `New` function for Postgres: https://github.com/openfga/openfga/blob/b40a1e73fa5d845bb1f7f29dca84ac4b255e2e1c/pkg/storage/postgres/postgres.go#L48

unknwon:
> > > * Could we apply the same change to the Postgres adapter please?
> > 
> > 
> > Where exactly is the "Postgres adapter" you're referring to? Could you point me to a link/file?
> 
> Here is the `New` function for Postgres:
> 
> https://github.com/openfga/openfga/blob/b40a1e73fa5d845bb1f7f29dca84ac4b255e2e1c/pkg/storage/postgres/postgres.go#L48

Hmm, isn't this PR changing the this exact file? I'm not sure I follow what do you mean by "apply the same change" then.

miparnisari:
@unknwon ohh my bad, sorry! I meant if we could also update MySQL 😅 https://github.com/openfga/openfga/blob/main/pkg/storage/mysql/mysql.go#L47

unknwon:
Hi @miparnisari, who should I ask for reviews for this PR?

miparnisari (approved):

Issue #1679, opened 2024-06-11
Flaky ListUsers tests


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

1.

```go
--- FAIL: TestListUsersDepthExceeded (0.04s)
    --- FAIL: TestListUsersDepthExceeded/depth_should_exceed_limit (0.02s)
        list_users_rpc_test.go:2916: 
            	Error Trace:	/home/runner/work/openfga/openfga/pkg/server/commands/listusers/list_users_rpc_test.go:2916
            	Error:      	Not equal: 
            	            	expected: "resolution depth exceeded"
            	            	actual  : "resolution depth exceeded\ncontext canceled"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1,2 @@
            	            	 resolution depth exceeded
            	            	+context canceled
            	Test:       	TestListUsersDepthExceeded/depth_should_exceed_limit
```

2.
```go
--- FAIL: TestListUsersConditions (0.08s)
    --- FAIL: TestListUsersConditions/error_in_direct_eval (0.00s)
        list_users_rpc_test.go:2916: 
            	Error Trace:	/home/runner/work/openfga/openfga/pkg/server/commands/listusers/list_users_rpc_test.go:2916
            	Error:      	Not equal: 
            	            	expected: "failed to evaluate relationship condition: parameter type error on condition 'condFloat' - failed to convert context parameter 'x': number cannot be represented as a float64: 1.797693135e+309"
            	            	actual  : ""
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1 @@
            	            	-failed to evaluate relationship condition: parameter type error on condition 'condFloat' - failed to convert context parameter 'x': number cannot be represented as a float64: 1.797693135e+309
            	            	+
            	Test:       	TestListUsersConditions/error_in_direct_eval
FAIL
```

3.

```go
--- FAIL: TestListUsersMemory (6.37s)
    tests.go:62: creating connection to address 0.0.0.0:46511
    --- FAIL: TestListUsersMemory/RunAll (0.00s)
        --- FAIL: TestListUsersMemory/RunAll/ListUsers (0.60s)
[...]

PR #1714, opened 2024-06-25
fix: ListUsers flaky tests

willvedd:
## Description
There are a few ListUsers tests that are failing intermittently, causing builds to fail. Two such examples:

**Example 1:**
```
--- FAIL: TestListUsersDepthExceeded (0.05s)
    --- FAIL: TestListUsersDepthExceeded/depth_should_exceed_limit (0.03s)
        list_users_rpc_test.go:2967: 
            	Error Trace:	/home/runner/work/openfga/openfga/pkg/server/commands/listusers/list_users_rpc_test.go:2967
            	Error:      	Not equal: 
            	            	expected: "resolution depth exceeded"
            	            	actual  : "resolution depth exceeded\ncontext canceled"
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -1 +1,2 @@
            	            	 resolution depth exceeded
            	            	+context canceled
            	Test:       	TestListUsersDepthExceeded/depth_should_exceed_limit
```
**Example 2:**
```
--- FAIL: TestListUsersMemory (6.78s)
    tests.go:62: creating connection to address 0.0.0.0:39753
    --- FAIL: TestListUsersMemory/RunAll (0.00s)
        --- FAIL: TestListUsersMemory/RunAll/ListUsers (0.72s)
            --- FAIL: TestListUsersMemory/RunAll/ListUsers/resolution_too_complex_throws_error (0.10s)
                --- FAIL: TestListUsersMemory/RunAll/ListUsers/resolution_too_complex_throws_error/stage_0 (0.10s)
                    --- FAIL: TestListUsersMemory/RunAll/ListUsers/resolution_too_complex_throws_error/stage_0/assertion_0 (0.02s)
                        listusers.go:187: 
                            	Error Trace:	/home/runner/work/openfga/openfga/tests/listusers/listusers.go:187
                            	Error:      	An error is expected but got nil.
                            	Test:       	TestListUsersMemory/RunAll/ListUsers/resolution_too_complex_throws_error/stage_0/assertion_0
                            	Messages:   	ListUsers request: object=resource:abc, relation=a1, filters=user. Model: model
                            	            	  schema 1.1
                            	            	type user
```

I cannot recreate either issue locally but I believe the following to be the culprits and proposed solutions:
- **Example 1:** A race condition that is exacerbated by slow Github action compute resources. Fix here is to loosen the assertion to allow for unequal errors so long as the expected error is included in the broader error. This is a possibility due to the usage of error concatenation.
- **Example 2:** Timeouts due to slow Github action compute resources. The solution here is to increase the timeout from 3s to 6s to allow for more reliable completion.

## References
Related Issue: #1679 

## Review Checklist


jpadilla (approved):

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -2964,7 +2964,7 @@ func (testCases ListUsersTests) runListUsersTestCases(t *testing.T) {
 			if err != nil {
 				actualErrorMsg = err.Error()
 			}
-			require.Equal(t, test.expectedErrorMsg, actualErrorMsg)
+			require.Contains(t, actualErrorMsg, test.expectedErrorMsg)
```
adriantam, line 2967: have we consider using [Errors.Is](https://pkg.go.dev/errors)?

tests/listusers/listusers_test.go
```diff
@@ -46,6 +47,7 @@ func testRunAll(t *testing.T, engine string) {
 	cfg.Log.Level = "error"
 	cfg.Datastore.Engine = engine
 	cfg.Experimentals = []string{"enable-list-users"}
+	cfg.ListUsersDeadline = 6 * time.Second
```
miparnisari, line 50: > Example 2: Timeouts due to slow Github action compute resources. The solution here is to increase the timeout from 3s to 6s to allow for more reliable completion.

I don't think this is the actual root cause. The error was

> Error Trace:	/home/runner/work/openfga/openfga/tests/listusers/listusers.go:187
                            	Error:      	An error is expected but got nil.

I believe the problem is this code
https://github.com/openfga/openfga/blob/ffe0bfd2ec0c25f366e889c857240c49a11e0ee8/pkg/server/commands/listusers/list_users_rpc.go#L226-L237

If `case <-doneWithFoundUsersCh:` executes first but there was an error, you will miss it.

willvedd: Hi Maria, thanks for raising this concern.

I don't believe its possible for both the `err` and `doneWithFoundUsersCh` cases to both be satisfied simultaneously because the `foundUsersCh` gets closed after the error is checked. However, it is possible that both the `err` and context cancelled cases could be simultaneously true.

This exact test failure can be observed locally if we set the timeout arbitrarily low, which suggests that it is completely possible that this test failure is due to timeout being hit. But I agree that it doesn't appear exactly intuitive.

Speaking pragmatically, the change to extend the timeout is simple and can be easily disproven if we continue to see test failures, which _should_ be evident by now considering the frequency of failures we've seen. At which point, we can more precisely dissect the failure.

miparnisari: > I don't believe its possible for both the err and doneWithFoundUsersCh cases to both be satisfied simultaneously because the foundUsersCh gets closed after the error is checked.

I did a quick test here: https://go.dev/play/p/EixeM07Zolf

In that code, the error channel gets a message before the `done` channel gets a message. And yet I don't see the error.

PR #1721, opened 2024-06-26
feat: add telemetry data to view consistency type for request

ewanharris:
## Description

Adds the requested consistency to the `request_duration_ms` histogram and spans of check, list objects, and streamed list objects. Currently this will report 3 possible values, `UNSPECIFIED`, `MINIMIZE_LATENCY`, `HIGHER_CONSISTENCY`, the intent here is that because a consistency of `HIGHER_CONSISTENCY` implies skipping the caching methods used within these queries, then the by being able to restrict searches based on the consistency type we can greater understand the impacts of skipping the cache.

Some examples of the data being shown below

Grafana dashboard

Trace in jaeger

## Review Checklist

### DIFF
--- a/pkg/server/list_users.go
+++ b/pkg/server/list_users.go
@@ -42,6 +42,7 @@ func (s *Server) ListUsers(
 		attribute.String("object", tuple.BuildObject(req.GetObject().GetType(), req.GetObject().GetId())),
 		attribute.String("relation", req.GetRelation()),
 		attribute.String("user_filters", userFiltersToString(req.GetUserFilters())),
+		attribute.String("consistency", req.GetConsistency().String()),
 	))
 	defer span.End()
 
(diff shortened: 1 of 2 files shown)


pkg/server/list_users.go
```diff
@@ -110,6 +110,7 @@ func (s *Server) ListUsers(
 		methodName,
 		utils.Bucketize(uint(datastoreQueryCount), s.requestDurationByQueryHistogramBuckets),
 		utils.Bucketize(uint(dispatchCount), s.requestDurationByDispatchCountHistogramBuckets),
+		"N/A",
```
ewanharris, line 113: I don't like this, but we can't reuse the same histogram without doing this as if we don't pass the expected number of label values then we will panic. I did some digging and there doesn't seem to be any alternative other than creating a separate histogram

ewanharris: Scratch that, it just never got added to the ListUsersRequest. I'll add it in this PR!

jimmyjames (approved):
lgtm

PR #1723, opened 2024-06-27
feat: typesystem to evaluate TTU are exclusively directly assignable

adriantam:
## Description
If TTU and its child relations are directly assignable, then we can short circuit check evaluation.

## Review Checklist

### DIFF
--- a/pkg/typesystem/typesystem.go
+++ b/pkg/typesystem/typesystem.go
@@ -333,6 +333,7 @@ func (t *TypeSystem) DirectlyRelatedUsersets(objectType, relation string) ([]*op
 
 // resolvesTypeRelationToDirectlyAssignable returns whether the assignable types ONLY if
 // the specified type and relation is exclusively assignable. Otherwise, it will return nil as slice.
+// TODO: memorized so that we do not need to reparse the type system in subsequent calls.
 func (t *TypeSystem) resolvesTypeRelationToDirectlyAssignable(objectType, relationName string) ([]string, bool, error) {
 	relation, err := t.GetRelation(objectType, relationName)
 	if err != nil {
@@ -344,6 +345,8 @@ func (t *TypeSystem) resolvesTypeRelationToDirectlyAssignable(objectType, relati
 	}
 
 	directlyRelatedTypes := relation.GetTypeInfo().GetDirectlyRelatedUserTypes()
+
+	// assignableTypes is needed for the TTU case
 	assignableTypes := make([]string, 0, len(directlyRelatedTypes))
 	// need to check whether these are simple types as well
 	for _, ref := range directlyRelatedTypes {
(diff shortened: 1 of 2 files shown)


pkg/typesystem/typesystem.go
```diff
@@ -331,6 +331,32 @@ func (t *TypeSystem) DirectlyRelatedUsersets(objectType, relation string) ([]*op
...
+	// need to check whether these are simple types as well
+	for _, ref := range directlyRelatedTypes {
+		if ref.GetRelationOrWildcard() != nil || ref.GetCondition() != "" {
+			// For now, we don't allow if these types are another userset, publicly assignable or has condition
+			// because local check with these relations cannot be evaluated via simple datastore query.
+			return nil, false, nil
+		}
+		assignableTypes = append(assignableTypes, ref.GetType())
+	}
+	return assignableTypes, true, nil
```
elbuo8, line 357: Why the slice? To cache later?

pkg/typesystem/typesystem.go
```diff
@@ -389,6 +405,32 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
+	if err != nil {
+		return false, err
+	}
+	if !directlyAssignable {
+		return false, nil
+	}
+	for _, tuplesetRelationType := range tuplesetRelationTypes {
+		_, childDirectlyAssignable, err := t.resolvesTypeRelationToDirectlyAssignable(tuplesetRelationType, computedRelation)
+		if err != nil {
+			var relationUndefinedError *RelationUndefinedError
```
elbuo8, line 420: 🤔 why does it need to be defined each time?

pkg/typesystem/typesystem.go
elbuo8: Got it, you need it for TTU. Lets mark this part to be memoized in subsequent calls given that it will happen _a lot_.

pkg/typesystem/typesystem.go
adriantam: Will move this out of the for loop

pkg/typesystem/typesystem.go
```diff
@@ -389,6 +408,35 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
+// TTUResolvesExclusivelyToDirectlyAssignable returns whether the computedRelation of object's tupleRelation is directly assignable.
+func (t *TypeSystem) TTUResolvesExclusivelyToDirectlyAssignable(objectType, tuplesetRelation, computedRelation string) (bool, error) {
+	tuplesetRelationTypes, directlyAssignable, err := t.resolvesTypeRelationToDirectlyAssignable(objectType, tuplesetRelation)
+	if err != nil {
+		return false, err
+	}
+	if !directlyAssignable {
+		return false, nil
+	}
+	var relationUndefinedError *RelationUndefinedError
```
elbuo8, line 420: Seems this could simply be a const

elbuo8 (approved):

pkg/typesystem/typesystem.go
adriantam: Checked - this cannot be constant since errors.As will not work.

PR #1724, opened 2024-06-27
feat: throw error when higher consistency requested but flag not enabled

jimmyjames:
Raises an error when the `consistency` parameter is `HIGHER_CONSISTENCY` but the `enable-consistency-params` experimental flag is not set.

## Description

If a request is made that requests higher consistency (`consistency: HIGHER_CONSISTENCY`), but the `enable-consistency-params` is not enabled, raise an `InvalidArgument` error.

Impacts the following endpoints:
* Check
* Read
* ListObjects
* StreamedListObjects
* Expand
* ListUsers

Note that because the API will always send the default value of the consistency enum (`UNSPECIFIED`), we are only raising an error if the caller has specifically requested `HIGHER_CONSISTENCY` without the flag enabled.

## Review Checklist

### DIFF
--- a/pkg/server/list_users.go
+++ b/pkg/server/list_users.go
@@ -36,8 +36,9 @@ func (s *Server) ListUsers(
 		return nil, status.Error(codes.Unimplemented, "ListUsers is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-list-users` configuration option when running OpenFGA server")
 	}
 
-	if !s.IsExperimentallyEnabled(ExperimentalEnableConsistencyParams) && req.GetConsistency() == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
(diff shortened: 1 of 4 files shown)


pkg/server/list_users.go
```diff
@@ -36,6 +36,10 @@ func (s *Server) ListUsers(
 		return nil, status.Error(codes.Unimplemented, "ListUsers is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-list-users` configuration option when running OpenFGA server")
 	}
 
+	if !s.IsExperimentallyEnabled(ExperimentalEnableConsistencyParams) && req.GetConsistency() == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
```
rhamzeh, line 39: Maybe we can have this in a function instead of having it repeated everywhere?

rhamzeh (approved):

pkg/server/list_users.go
jimmyjames: Updated to extract a common method 👍

PR #1732, opened 2024-07-03
feat(storage)!: use method-specific options

jimmyjames:
For every storage method that takes a `PaginationOptions`, replace with method-specific options.

## Description

As discussed in #1716, as we add the ability to specify a consistency preference, we want to avoid adding another parameter to the methods that will require it. To make the storage API more resilient to additional future changes, we should instead specify an options type.

This PR adds the method-specific options, but **does not** introduce any consistency-related changes. The intent is that this PR maintain current functional behavior, while enabling us to add the consistency changes in a later PR.

The following options types have been added in this PR:
* `ReadAuthorizationModelsOptions`
* `ListStoresOptions`
* `ReadChangesOptions`
* `ReadPageOptions`

The related methods have been updated to accept these options (which currently only hold `PaginationOptions`), instead of the `PaginationOptions` directly:

* `ReadAuthorizationModels(ctx context.Context, store string, options ReadAuthorizationModelsOptions) ([]*openfgav1.AuthorizationModel, []byte, error)`
* `ListStores(ctx context.Context, options ListStoresOptions) ([]*openfgav1.Store, []byte, error)`
* `ReadChanges(ctx context.Context, store, objectType string, options ReadChangesOptions, horizonOffset time.Duration) ([]*openfgav1.TupleChange, []byte, error)`
* `ReadPage(ctx context.Context, store string, tupleKey *openfgav1.TupleKey, options ReadPageOptions ([]*openfgav1.Tuple, []byte, error)`

## References

#1716 

## Review Checklist

### DIFF
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -368,7 +368,7 @@ func (p *Postgres) ReadAuthorizationModel(ctx context.Context, store string, mod
 }
 
 // ReadAuthorizationModels see [storage.AuthorizationModelReadBackend].ReadAuthorizationModels.
-func (p *Postgres) ReadAuthorizationModels(ctx context.Context, store string, opts storage.PaginationOptions) ([]*openfgav1.AuthorizationModel, []byte, error) {
+func (p *Postgres) ReadAuthorizationModels(ctx context.Context, store string, options storage.ReadAuthorizationModelsOptions) ([]*openfgav1.AuthorizationModel, []byte, error) {
 	ctx, span := tracer.Start(ctx, "postgres.ReadAuthorizationModels")
 	defer span.End()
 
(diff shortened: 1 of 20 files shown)


pkg/storage/postgres/postgres.go
```diff
@@ -677,8 +672,8 @@ func (p *Postgres) ReadChanges(
 	if objectTypeFilter != "" {
 		sb = sb.Where(sq.Eq{"object_type": objectTypeFilter})
 	}
-	if opts.From != "" {
-		token, err := sqlcommon.UnmarshallContToken(opts.From)
+	if options.Pagination.From != "" {
```
miparnisari, line 675: can `Pagination` be nil? in that case won't this create a panic?

jimmyjames: I wondered about this as well. If the options contain no `Pagination` value, e.g.,

`readChangesOpts := storage.ReadChangesOptions{}`

Then the zero/default values are:
`opts.Pagination.From == ''`
`opts.Pagination.PageSize == 0`

Checking for nil `Pagination` won't compile given that `options` is not a pointer.

Given that, if there is a more idiomatic Go way of ensuring nil safety I'm happy to change it :)

PR #1734, opened 2024-07-03
feat: userset short path

jpadilla:
## Review Checklist


internal/graph/check.go
```diff
@@ -883,6 +1079,8 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 			attribute.String("computed_relation", computedRelation),
 		)
 
+		// check if TTU is only directly assignable
```
elbuo8, line 1082: Remove.

internal/graph/check.go
```diff
@@ -614,6 +614,229 @@ func (c *LocalChecker) ResolveCheck(
...
+	span         trace.Span
+	filteredIter storage.TupleKeyIterator
+	typesys      *typesystem.TypeSystem
+	req          *ResolveCheckRequest
+	ds           storage.RelationshipTupleReader
+}
+
+// checkUsersetPublicWildcardSlowPath will check userset or public wildcard path.
+// This is the slow path as it requires dispatch on all its children.
+func (c *LocalChecker) checkUsersetPublicWildcardSlowPath(ctx context.Context, info checkDirectUsersetPublicInternalInfo) (*ResolveCheckResponse, error) {
```
jpadilla, line 628: We should think of a better name for this.

internal/graph/check.go
```diff
@@ -614,6 +614,229 @@ func (c *LocalChecker) ResolveCheck(
...
+//	  define member: [user]
+//	type doc
+//	  define viewer: [group#member]
+//
+// We will first look up the group(s) that are assigned to doc:1
+// Next, we will look up all the group where user:bob is a member of.
+// Finally, find the intersection between the two.
+// To use the fastpath, we will need to ensure that the userset and all the children associated with the userset are
+// exclusively directly assignable. In our case, group member must be directly exclusively assignable.
+func (c *LocalChecker) checkUsersetPublicWildcardFastPath(ctx context.Context, info checkDirectUsersetPublicInternalInfo) (*ResolveCheckResponse, error) {
```
jpadilla, line 727: We should think of a better name for this.

internal/graph/check.go
```diff
@@ -614,6 +614,229 @@ func (c *LocalChecker) ResolveCheck(
...
+		},
+	}
+
+	// usersetObjectTypeRelationMap is a map of all userset objectType and its relations.  For example,
+	// [group:1#member, group:2#member, group:1#owner, group:3#owner] will be stored as
+	// [group][member][1]
+	// [group][owner][1]
+	// [group][owner][3]
+	// We will later compare usersetObjectTypeRelationMap's entries with the object user has relation with.
+	usersetObjectTypeRelationMap := make(map[string]map[string]map[string]struct{})
```
jpadilla, line 743: I switched this up to allow a faster lookup. Originally it was:

```
		// [group:1#member, group:2#member, group:1#owner, group:3#owner] will be stored as
		// [group][member]: [1, 2]
		// [group][owner]: [1, 3]
		// We will later compare usersetObjectTypeRelationMap's entries with the object user has relation with.
		usersetObjectTypeRelationMap := make(map[string]map[string][]string)
```

internal/graph/check.go
```diff
@@ -728,73 +946,48 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+			})
+		}
 
-				if usersetRelation != "" {
-					tupleKey := tuple.NewTupleKey(usersetObject, usersetRelation, reqTupleKey.GetUser())
-					handlers = append(handlers, c.dispatch(ctx, req, tupleKey))
-				}
-			}
+		// TODO(jpadilla): can we lift this function up?
+		checkDirectUsersetTuplesFast := func(ctx context.Context) (*ResolveCheckResponse, error) {
```
jpadilla, line 959: We should think of a better name for this.

internal/graph/check.go
```diff
@@ -614,6 +614,238 @@ func (c *LocalChecker) ResolveCheck(
...
+				t, err := iter.Next(ctx)
+				if err != nil {
+					if errors.Is(err, storage.ErrIteratorDone) {
+						break
+					}
+
+					return nil, err
+				}
+
+				if actualObject := t.GetKey().GetObject(); actualObject != "" {
```
miparnisari, line 828: why do we need an `if`, when could the object be empty? This is a tuple that's already been stored so it should be valid

internal/graph/check.go
```diff
@@ -614,6 +614,238 @@ func (c *LocalChecker) ResolveCheck(
...
+					_, objectID := tuple.SplitObject(actualObject)
```
miparnisari, line 829: FYI Updated godoc for this function: https://github.com/openfga/openfga/pull/1737

internal/graph/check.go
```diff
@@ -614,6 +614,238 @@ func (c *LocalChecker) ResolveCheck(
...
+	}
+
+	// usersetsMap is a map of all ObjectRelations and its IdsFor example,
+	// [group:1#member, group:2#member, group:1#owner, group:3#owner] will be stored as
+	// [group#member][1]
+	// [group#owner][1, 3]
+	usersetsMap := make(map[string]map[string]struct{})
+
+	for {
+		t, err := iter.Next(ctx)
```
miparnisari, line 744: If the underlying iterator has to do more than 1 database trip, you will have to wait for all those trips to complete before you begin processing line 797. Suggestion:  put both pieces in different goroutines. Goroutine A sends objectIDs through a map of channels (one per object#relation)

I say leave this as a follow-up because by doing this we open a small can of worms (race conditions, possibility of leaking goroutines etc)

internal/graph/check.go
```diff
@@ -614,6 +614,238 @@ func (c *LocalChecker) ResolveCheck(
...
+						break
+					}
+
+					return nil, err
+				}
+
+				if actualObject := t.GetKey().GetObject(); actualObject != "" {
+					_, objectID := tuple.SplitObject(actualObject)
+					if _, ok := objectIDs[objectID]; ok {
+						response.Allowed = true
```
miparnisari, line 831:
```suggestion
                                                  span.SetAttributes(attribute.Bool("allowed", true))
						response.Allowed = true
```

internal/graph/check.go
elbuo8: Yep, aware of that. In fact, we could start executing the subsequent ones once we observe a whole new ObjectRelation (since the results come sorted) so we could in fact process as a stream. For the time being (this first pass) I'd like to land as is and we can optimize later since I don't envision that to be trivial 😅

internal/graph/check.go
elbuo8: I had the exact same question to @adriantam when he first did the POC. Happy to remove it or leave until he clarifies when he is back.

internal/graph/check.go
elbuo8: Adding a NOTE regarding this conversation so its not overlooked in the future.

internal/graph/check.go
jpadilla: Do we need to use the filtered iterator here as well?

internal/graph/check.go
```diff
@@ -614,6 +614,241 @@ func (c *LocalChecker) ResolveCheck(
...
+	for objectRel, objectIDs := range usersetsMap {
+		handler := func(ctx context.Context) (*ResolveCheckResponse, error) {
+			response := &ResolveCheckResponse{
+				Allowed: false,
+				ResolutionMetadata: &ResolveCheckResponseMetadata{
+					DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount + 1,
+				},
+			}
+			objectType, relation := tuple.SplitObjectRelation(objectRel)
+			iter, err := ds.ReadStartingWithUser(ctx, storeID, storage.ReadStartingWithUserFilter{
```
miparnisari, line 808: We are missing 

```go
				validation.FilterInvalidTuples(typesys),
			)
			defer filteredIter.Stop()
```

internal/graph/check.go
```diff
@@ -614,6 +614,241 @@ func (c *LocalChecker) ResolveCheck(
...
+				t, err := iter.Next(ctx)
+				if err != nil {
+					if errors.Is(err, storage.ErrIteratorDone) {
+						break
+					}
+
+					return nil, err
+				}
+
+				_, objectID := tuple.SplitObject(t.GetKey().GetObject())
```
miparnisari, line 830: don't we also have to evaluateTupleCondition here? 🤔  (how is this passing tests?)

elbuo8: Why? The condition being evaluated at the source would discard it for the intersection 🤔

miparnisari: But you could have both ends of the intersection be conditioned, e.g.

```go
doc:1, viewer, group:eng#member, cond: cond_x
user:maria, member, group:eng, cond:cond_x
```

internal/graph/check.go
elbuo8: https://github.com/openfga/openfga/pull/1746

internal/graph/check.go
```diff
@@ -614,6 +614,262 @@ func (c *LocalChecker) ResolveCheck(
...
+	if !ok {
+		return nil, fmt.Errorf("typesystem missing in context")
+	}
+
+	ds, ok := storage.RelationshipTupleReaderFromContext(ctx)
+	if !ok {
+		return nil, fmt.Errorf("relationship tuple reader datastore missing in context")
+	}
+
+	// usersetsMap is a map of all ObjectRelations and its IdsFor example,
```
miparnisari, line 737:
```suggestion
	// usersetsMap is a map of all ObjectRelations and its Ids. For example,
```

miparnisari (dismissed):

miparnisari (approved):

PR #1735, opened 2024-07-03
feat: ttu quick check

jpadilla:
## Review Checklist


internal/graph/check.go
```diff
@@ -848,6 +848,233 @@ func (c *LocalChecker) checkComputedUserset(_ context.Context, req *ResolveCheck
...
+	// type group
+	//   define member: [user]
+	// type doc
+	//   define parent: [group]
+	//   define viewer: member from parent
+	// tuplesetRelationUserMap will list all group for doc:1's parent
+	// [group][1]
+	// [group][2]
+	// [group][3]
+	tuplesetRelationUserMap := make(map[string]map[string]struct{})
```
miparnisari, line 983: nit: this example doesn't make a lot of sense, maybe change the model so that `define parent: [group1, group2]` and then it's clear the map is like

````
group1 -> 1,2,3
group2 -> 4,5
```

internal/graph/check.go
```diff
@@ -848,6 +848,233 @@ func (c *LocalChecker) checkComputedUserset(_ context.Context, req *ResolveCheck
...
+				return nil, err
+			}
+
+			if actualObject := t.GetKey().GetObject(); actualObject != "" {
+				_, objectID := tuple.SplitObject(actualObject)
+				if _, ok := tuplesetObjectList[objectID]; ok {
+					response.Allowed = true
+					return response, nil
+				}
+			}
```
miparnisari, line 1071: Please see comments in https://github.com/openfga/openfga/pull/1734

internal/graph/check.go
elbuo8: The reasoning is to use a `Map[Set]` so that lookups in the ID are faster

internal/graph/cached_resolver_test.go
```diff
@@ -654,7 +654,7 @@ func TestCachedCheckDatastoreQueryCount(t *testing.T) {
 	ttuLocalChecker.Close()
 
 	require.NoError(t, err)
-	require.Equal(t, uint32(1), res.GetResolutionMetadata().DatastoreQueryCount)
+	require.Equal(t, uint32(2), res.GetResolutionMetadata().DatastoreQueryCount)
```
elbuo8, line 657: ?

jpadilla: still need to figure this one out

internal/graph/check.go
```diff
@@ -848,6 +848,268 @@ func (c *LocalChecker) checkComputedUserset(_ context.Context, req *ResolveCheck
...
+
+	var errs error
+	var handlers []CheckHandlerFunc
+
+	typesys, ok := typesystem.TypesystemFromContext(ctx)
+	if !ok {
+		return nil, fmt.Errorf("typesystem missing in context")
+	}
+
+	computedRelation := rewrite.GetTupleToUserset().GetComputedUserset().GetRelation()
```
willvedd, line 865: This could potentially be multiple relations. Consider unions and relations that have both a directly assignable and computed relation.

elbuo8: Unions are evaluated at a different level at the current time, and computed is currently not supported. So for the context of this implementation, only one needs to be handled. This is, in-fact the _existing_ implementation of in main.

internal/graph/cached_resolver_test.go
miparnisari: It's because: in the old slow way, we divide the TTU check into subproblems. 

Eg if the query is `Check(user:maria, viewer, document:1)`

1. We find `folder:x` is parent of `document:1`
2. Do`Check(user:maria, viewer, folder:x)`. This is stored in the internal cache thanks to the call to `dispatch` here: https://github.com/openfga/openfga/blob/d32341e622f63cbaf5ff0fb384f3e2af93cd7975/internal/graph/check.go#L687

In the new way we do:
1. We find `folder:x` is parent of `document:1`
2. Find all `folder`s that` user:maria` can view. This part we are not caching, and I don't see how we'd do that, because it's not a Check subproblem (ie the result is not a bool, is a list)

elbuo8: I had a discussion with Yissell on actually caching the results of the short path in order to also optimize recursion of usersets. We can discuss that as an N+1.

internal/graph/check.go
```diff
@@ -1090,75 +1293,22 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
...
-
-			if _, err := typesys.GetRelation(tuple.GetType(userObj), computedRelation); err != nil {
-				if errors.Is(err, typesystem.ErrRelationUndefined) {
-					continue // skip computed relations on tupleset relationships if they are undefined
-				}
-			}
-
-			// Note: we add TTU read below
-			handlers = append(handlers, c.dispatch(ctx, req, tupleKey))
+		canShortCircuit, err := typesys.TTUResolvesExclusivelyToDirectlyAssignable(
```
elbuo8, line 1296: If this throws error, we should still resolve via the normal path rather than erroring out.

elbuo8:
```
		resolver := c.checkTTUSlowPath

		// TODO: optimize the case where user is an userset.
		// If the user is a userset, we will not be able to use the shortcut because the algo
		// will look up the objects associated with user.
		if !tuple.IsObjectRelation(tk.GetUser()) {
			if canShortCircuit, err := typesys.TTUResolvesExclusivelyToDirectlyAssignable(
				tuple.GetType(object), tuplesetRelation, computedRelation); err == nil && canShortCircuit {
				resolver = c.checkTTUFastPath
			}
		}
```

jpadilla: done

internal/graph/cached_resolver_test.go
miparnisari: @elbuo8 would that be changing the caching pattern to, instead of caching Check queries, cache the underlying datastore requests?

yissellokta (approved):

internal/graph/cached_resolver_test.go
miparnisari: OK. We can extend https://github.com/openfga/openfga/blob/3540b66777b2983e0f37581775759a4a5989b6d5/pkg/storage/storagewrappers/caching.go#L19-L23 for that

internal/graph/check.go
miparnisari: @willvedd re: "This could potentially be multiple relations" - if the TTU is `define viewer1: viewer2 from parent`, the computed relation will be `viewer2` and it will always be 1

internal/graph/check.go
```diff
@@ -1038,127 +970,342 @@ func (c *LocalChecker) checkComputedUserset(_ context.Context, req *ResolveCheck
...
+	defer span.End()
+
+	typesys, ok := typesystem.TypesystemFromContext(ctx)
+	if !ok {
+		return nil, fmt.Errorf("typesystem missing in context")
+	}
+
+	// usersetsMap is a map of all ObjectRelations and its Ids. For example,
+	// [group:1#member, group:2#member, group:1#owner, group:3#owner] will be stored as
+	// [group#member][1]
```
adriantam, line 1162:
```suggestion
	// [group#member][1, 2]
```

adriantam (approved):

miparnisari (dismissed):

PR #1733, opened 2024-07-03
chore: TypeSystem methods to determine if Relation resolves to Directly Assignable Types

elbuo8:
## References

https://github.com/openfga/openfga/pull/1715
https://github.com/openfga/openfga/pull/1725
https://github.com/openfga/openfga/pull/1735
https://github.com/openfga/openfga/pull/1734

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

Issue #1595, opened 2024-05-03
optimization possible for intersections


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

With a model such as

```
model
  schema 1.1

type user

type account

type document
  relations
    define a: [user]
    define b: [account]
    define intersection: a and b
    define intersection_2: b and a
```

any queries of the form

```
Check(user:...., intersection_2, document:...)
Check(user:...., intersection, document:...)
```

can be optimized based on the fact that `intersection([user], [account]) == {}`

<img width="1919" alt="image" src="https://github.com/openfga/openfga/assets/5374887/c150bf64-b941-4f57-9f57-13fd398bf502">

### Describe the ideal solution

We compute the terminal types of an intersection and if `len(terminal_types)!=1` we can return `allowed=false` immediately.

PR #1736, opened 2024-07-03
Determine all terminal types and relations for typesystem

willvedd:
## Description
Determining the ultimate terminal types and terminal relations for a given object type and relation. This data is proposed to be calculated at the time of typesystem creation and validations (`t.NewAndValidate()`) and associated in a multi-key map on the typesystem struct.

This work enables two different optimizations. The first is an optimized TTU and userset lookup with much fewer DB queries. However, this requires knowledge of the terminal types and relations for a given userset to craft the appropriate query. Currently, this is limited to usersets that result in a direct relation and necessarily doesn't result in a computed relation, union, intersection, etc. This limitation can be observed in #1733. 

Secondly, this map of connected types can ultimately be used as an optimization in Check, ListObjects and ListUsers as a screen to determine if the request's relation and types are valid with respect to the model. If not, we can exit the process with a falsey outcome and prevent some DB requests. This work was previously visited in #1582 but this PR will reduce the computation to a O(1) lookup at runtime.

**DRAFT:** Work is still in progress. Needs more testing and support for intersection and difference rewrites.

## References
Supports: #1733, #1735, #1734
May help supercede: #1582 

## Review Checklist

### DIFF
--- a/pkg/typesystem/connected_types.go
+++ b/pkg/typesystem/connected_types.go
@@ -1,10 +1,6 @@
 package typesystem
 
 import (
-	"encoding/json"
-	"fmt"
-	"log"
-
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 )
 
@@ -32,21 +28,11 @@ func (t *TypeSystem) AreTypesConnectedViaRelations(objectType, relation, subject
 	return true, terminalRelations
 }
 
-func TMPPrettyPrint(data interface{}) {
-	// Marshal the map to a pretty-printed JSON string
-	jsonData, err := json.MarshalIndent(data, "", "    ")
-	if err != nil {
-		log.Fatalf("Error pretty-printing map: %v", err)
-	}
-	fmt.Println(string(jsonData))
-}
-
-// validateRelation applies all the validation rules to a relation definition in a model. A relation
-// must meet all the rewrite validation, type restriction validation, and entrypoint validation criteria
-// for it to be valid. Otherwise, an error is returned.
+// AssignTerminalTypes will populate the `connectedTypes` property on the typesystem to indicate for a given object type
+// what the terminal user types and relations.
 func (t *TypeSystem) AssignTerminalTypes(typeName, relationName string, relationMap map[string]*openfgav1.Userset) {
 	rewrite := relationMap[relationName]
-	terminalTypesAndRelations := t.getTerminalSubjectTypesAndRelation(typeName, relationName, rewrite)
+	terminalTypesAndRelations := t.getTerminalUserTypesAndRelation(typeName, relationName, rewrite)
 
 	for _, terminalTypesAndRelation := range terminalTypesAndRelations {
 		for _, terminalType := range terminalTypesAndRelation.terminalTypes {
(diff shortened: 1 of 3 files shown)


pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,142 @@
...
+	// 	return intersectionAssignableTypesAndRelations
+	// }
+
+	// case *openfgav1.Userset_Difference:
+	// 	// All the children must have an entrypoint.
+	// 	baseTypes, baseTerminalRelation := t.getTerminalUserTypesAndRelation(typeName, relationName, rw.Difference.GetBase())
+	// 	subTypes, subTerminalRelation := t.getTerminalUserTypesAndRelation(typeName, relationName, rw.Difference.GetSubtract())
+
+	// 	return
+	// }
```
miparnisari, line 139: Intersection and difference should be basically the same as union: gather all the children and then append all the types you find

pkg/typesystem/connected_types_test.go
```diff
@@ -0,0 +1,165 @@
...
+		objectType                string
+		relation                  string
+		subjectType               string
+	}{
+		// // Intersection
+		{
+			objectType:                "document",
+			relation:                  "intersection_user_and_employee",
+			subjectType:               "employee",
+			expectedAreConnected:      false,
```
miparnisari, line 45: This is WIP, right? The expectation here should be true because `employee` is reachable from `document`.

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,142 @@
+package typesystem
+
+import (
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+)
+
+// map[objectType]map[relation]map[subjectType]terminalRelation.
+type TypesystemConnectedTypes map[string]map[string]map[string][]string
```
miparnisari, line 8: I don't love this data structure to this information 😓 which is why I used a graph data structure here: https://github.com/openfga/openfga/pull/1582

A graph ds enables us to extend the functionality to find cycles and traverse the model in any way we want

pkg/typesystem/connected_types_test.go
willvedd: First, please read PR description.

Secondly, `employee` is effectively not reachable because there are no types that intersect between the `intersection_user_and_employee` relation.

pkg/typesystem/connected_types.go
willvedd: It would be incorrect to treat union and intersection the same because intersection requires an overlap of common types.

pkg/typesystem/connected_types.go
willvedd: I'm open to the idea of modifying the data structure. However, the objective of this work is erformant lookups for terminal types and terminal relations for a given userset. The graph doesn't appear to be any more performant than traversing the typesystem itself, in which case we wouldn't store this data as a cache to begin with. The PR you link eludes to this which is why memoizing is recommended as a means of preventing duplicate function calls, which suggests storing the results in memory ultimately. That's where this PR immediately starts, without the intermediate steps.

pkg/typesystem/connected_types_test.go
miparnisari: I did read the PR description, I understood that the code was WIP but i didn't get whether the test was also WIP

pkg/typesystem/connected_types.go
miparnisari: Is that required for this work? It sounds like the fix for https://github.com/openfga/openfga/issues/1595

pkg/typesystem/connected_types.go
miparnisari: Agreed offline to keep this as is

willvedd:
Superseded by #1765

PR #1742, opened 2024-07-05
chore(storage!): readstartingfromuser can now filter based on an optional list

elbuo8:
## Description
`ReadStartingWithUser` now accepts an optional filter. If used, this reduces the data that must be processed in-memory, which can improve latency.


pkg/storage/memory/memory.go
```diff
@@ -509,9 +510,17 @@ func (s *MemoryBackend) ReadStartingWithUser(
...
-				matches = append(matches, t)
+			if targetUser != t.User {
+				continue
 			}
+
+			if len(filter.ObjectIDs) > 0 {
+				if !slices.Contains(filter.ObjectIDs, t.ObjectID) {
+					continue
+				}
+			}
```
miparnisari, line 521: Should this be moved outside of the `for` loop? There is nothing about the user filter in this block

pkg/storage/storage.go
```diff
@@ -171,6 +171,7 @@ type ReadStartingWithUserFilter struct {
 	ObjectType string
 	Relation   string
 	UserFilter []*openfgav1.ObjectRelation
+	ObjectIDs  []string // Optional.
```
miparnisari, line 174:
```suggestion
	ObjectIDs  []string // Optional. If present, the datastore returns the intersection between this filter and what is in the databse
```

pkg/storage/memory/memory.go
elbuo8: Totally right, fixing.

miparnisari (approved):

PR #1751, opened 2024-07-09
feat: check short circuit for TTU and userset with condition

elbuo8:
## Description
Adding filtered iterator for conditional tuples to improve check performance improvement for userset and TTU with conditions. 

## Review Checklist


pkg/storage/iterators.go
```diff
@@ -199,3 +199,67 @@ func NewFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKeyFilterFun
 		filter,
 	}
 }
+
+// TupleKeyConditionFilterFunc is a filter function that is used to filter out
+// tuples from a [TupleKeyIterator] that don't meet the tuple the conditions provided by the request.
+// Implementations should return true if the tuple should be returned
+// and false if it should be filtered out.
+// errors will be treated as false. The iterator can return the last seen error before ErrIteratorDone.
```
miparnisari, line 207: This comment doesn't seem accurate - if you return 5 tuples, 1 of which has an EvaluationError, and then it's the end of the iterator, you will evaluate line 229, skip it, and return `ErrIteratorDone`, no?

yissellokta (approved):

internal/graph/check.go
```diff
@@ -861,23 +822,15 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
-						fmt.Errorf("context is missing parameters '%v'", condEvalResult.MissingParameters),
-					)
-					telemetry.TraceError(span, evalErr)
-					return nil, evalErr
-				}
-
-				if !condEvalResult.ConditionMet {
+				if !conditionMet {
 					return response, nil
 				}
```
miparnisari, line 835: Can we simplify this code?

```go
			tupleKey := t.GetKey()
			err = validation.ValidateTuple(typesys, tupleKey)
			if err != nil {
				return response, nil
			}

			conditionalFilter := TupleKeyConditionalFilter(ctx, req.Context, typesys)
			conditionMet, err := conditionalFilter(tupleKey)
			if err != nil {
				return nil, err
			}

			if conditionMet {
				response.Allowed = true
			}

			return response, nil
```

pkg/storage/iterators.go
```diff
@@ -199,3 +199,67 @@ func NewFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKeyFilterFun
...
+
+// Stop see [Iterator.Stop].
+func (f *conditionsFilteredTupleKeyIterator) Stop() {
+	f.iter.Stop()
+	f.isClosed = true
+}
+
+// NewConditionsFilteredTupleKeyIterator returns a [TupleKeyIterator] that filters out all
+// [*openfgav1.Tuple](s) that don't meet the conditions of the provided [TupleKeyFilterFunc].
+func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKeyConditionFilterFunc) TupleKeyIterator {
```
miparnisari, line 260: Instead of returning an interface here why don't we return the explicit type *storage.ConditionsFilteredTupleKeyIterator? https://100go.co/#returning-interfaces-7

I'd also change the signature of `checkUsersetFastPath` to accept a `ConditionsFilteredTupleKeyIterator`. In the future I don't want to make the mistake of accidentally forgetting to perform all these validations prior to working with a tuple.

pkg/storage/iterators.go
```diff
@@ -199,3 +199,67 @@ func NewFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKeyFilterFun
...
+	lastError error
+	isClosed  bool
+	onceValid bool
+}
+
+var _ TupleKeyIterator = &conditionsFilteredTupleKeyIterator{}
+
+// Next returns the next most tuple in the underlying iterator that meets
+// the filter function this iterator was constructed with.
+func (f *conditionsFilteredTupleKeyIterator) Next(ctx context.Context) (*openfgav1.TupleKey, error) {
```
miparnisari, line 222:
```suggestion
// Next returns the next most tuple in the underlying iterator that meets
// the filter function this iterator was constructed with.
// This function is not thread-safe.
func (f *conditionsFilteredTupleKeyIterator) Next(ctx context.Context) (*openfgav1.TupleKey, error) {
```

😅

pkg/storage/iterators.go
adriantam: The behaviour was incorrect and is now fixed.  At the end of the iterator, it should return error.

pkg/storage/iterators.go
adriantam: I would defer this to a separate PR as all iterators in storage/iterators.go are private type (and only the interface is exposed).  If we are moving to this route, we should do it for all iterators, not just ConditionsFilteredTupleKeyIterator.

pkg/storage/iterators.go
miparnisari: No unit tests?

pkg/storage/iterators.go
miparnisari: This PR is the right place to make that change, IMO. We are introducing a new object, we should ensure people don't misuse it right in this PR

pkg/storage/iterators.go
adriantam: Unit tests added

pkg/storage/iterators.go
adriantam: Done

internal/graph/check.go
```diff
@@ -614,12 +614,80 @@ func (c *LocalChecker) ResolveCheck(
...
+			}
+		}
+		return response, nil
+	}
+}
+
 // checkUsersetSlowPath will check userset or public wildcard path.
 // This is the slow path as it requires dispatch on all its children.
-func (c *LocalChecker) checkUsersetSlowPath(ctx context.Context, iter storage.TupleKeyIterator, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
+func (c *LocalChecker) checkUsersetSlowPath(ctx context.Context, iter *storage.ConditionsFilteredTupleKeyIterator, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
```
elbuo8, line 688: Why not stick with the interface?

adriantam: See https://github.com/openfga/openfga/pull/1751#discussion_r1671600360

internal/graph/check.go
```diff
@@ -688,22 +735,37 @@ func (c *LocalChecker) checkUsersetSlowPath(ctx context.Context, iter storage.Tu
...
-		return nil, errors.Join(errs, err)
+		return nil, err
 	}
 
 	resp.GetResolutionMetadata().DatastoreQueryCount += response.GetResolutionMetadata().DatastoreQueryCount
 
 	return resp, nil
 }
 
+func TupleKeyConditionalFilter(ctx context.Context, reqCtx *structpb.Struct, typesys *typesystem.TypeSystem) storage.TupleKeyConditionFilterFunc {
```
elbuo8, line 749: Maybe this should be private (only meant to be used in Check for now) and perhaps it should be called `buildTupleKeyConditionFilter`?

internal/graph/check.go
```diff
@@ -860,27 +890,16 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
-
-				if !condEvalResult.ConditionMet {
-					return response, nil
-				}
-
+			if err != nil {
+				return response, nil
+			}
+			conditionalFilter := TupleKeyConditionalFilter(ctx, req.Context, typesys)
+			conditionMet, err := conditionalFilter(tupleKey)
```
elbuo8, line 897: If `buildTupleKeyConditionFilter` were to be the builder, this would be `tupleKeyConditionFilter(tupleKey)`

internal/graph/check.go
elbuo8: Makes sense, explicitly saying the iterator was already filtered out with conditions. 👍🏽

pkg/storage/iterators_test.go
```diff
@@ -104,3 +104,116 @@ func ExampleNewFilteredTupleKeyIterator() {
...
+		tuples := []*openfgav1.TupleKey{
+			tuple.NewTupleKeyWithCondition("document:doc1", "viewer", "user:jon", "condition3", nil),
+			tuple.NewTupleKeyWithCondition("document:doc1", "editor", "user:elbuo", "condition3", nil),
+			tuple.NewTupleKeyWithCondition("document:doc1", "editor", "user:maria", "condition3", nil),
+		}
+		iter := NewConditionsFilteredTupleKeyIterator(NewStaticTupleKeyIterator(tuples), filter)
+		tk, err := iter.Next(context.Background())
+		require.Error(t, err)
+		require.Nil(t, tk)
+	})
```
miparnisari, line 218:
```suggestion
	t.Run("all_invalid", func(t *testing.T) {
		tuples := []*openfgav1.TupleKey{
			tuple.NewTupleKeyWithCondition("document:doc1", "viewer", "user:jon", "condition3", nil),
			tuple.NewTupleKeyWithCondition("document:doc1", "editor", "user:elbuo", "condition4", nil),
			tuple.NewTupleKeyWithCondition("document:doc1", "editor", "user:maria", "condition5", nil),
		}
		iter := NewConditionsFilteredTupleKeyIterator(NewStaticTupleKeyIterator(tuples), filter)
		tk, err := iter.Next(context.Background())
		
		// only the last error is returned
		require.Equal(t, err.Error(), "unknown condition: condition5")
		require.Nil(t, tk)
	})
```

so that it is super clear what the error looks like (ie it's not a combination of errors)

pkg/storage/iterators_test.go
```diff
@@ -104,3 +104,116 @@ func ExampleNewFilteredTupleKeyIterator() {
...
+
+			actual = append(actual, tk)
+		}
+		expected := []*openfgav1.TupleKey{tuples[0]}
+		require.Equal(t, expected, actual)
+	})
+
+	t.Run("empty_list", func(t *testing.T) {
+		var tuples []*openfgav1.TupleKey
+		iter := NewConditionsFilteredTupleKeyIterator(NewStaticTupleKeyIterator(tuples), filter)
```
miparnisari, line 191: here and elsewhere, add `t.Cleanup(iter.Stop)`

pkg/storage/iterators_test.go
```diff
@@ -104,3 +104,116 @@ func ExampleNewFilteredTupleKeyIterator() {
...
+			tuple.NewTupleKeyWithCondition("document:doc1", "viewer", "user:jon", "condition1", nil),
+			tuple.NewTupleKeyWithCondition("document:doc1", "editor", "user:elbuo", "condition2", nil),
+			tuple.NewTupleKeyWithCondition("document:doc1", "editor", "user:maria", "condition3", nil),
+		}
+		iter := NewConditionsFilteredTupleKeyIterator(NewStaticTupleKeyIterator(tuples), filter)
+		var actual []*openfgav1.TupleKey
+		for {
+			tk, err := iter.Next(context.Background())
+			if err != nil {
+				if errors.Is(err, ErrIteratorDone) {
```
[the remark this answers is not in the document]
adriantam: onceValid -> if there are any valid, we should not emit error.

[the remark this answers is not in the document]
miparnisari: but why? in the context of a Check request, if we encountered one valid tuple _that doesn't lead to a `allowed=true` outcome, and one that we couldn't evaluate, we should return the evaluation error no? because at this point in execution in the iterator, you have no idea how the consumer used that first valid tuple?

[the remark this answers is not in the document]
adriantam: updated comments to reflect when we return error.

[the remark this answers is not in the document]
miparnisari: Never mind, I just understood - this is trying to mimic the behavior of the old way of doing things where we only return the error if we couldn't find any tuples

miparnisari (approved):

Issue #1753, opened 2024-07-09
Model validation is computationally inefficient


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Model validation runs on every query request that doesn't specify an authorization model ID as parameter. See https://github.com/openfga/openfga/blob/06e3aa9b4115173903aa30a3074e1fa57b6f1d79/pkg/typesystem/resolver.go#L50-L90. This is inefficient - if the model is in the cache, presumably it's been validated already, so we should be able to skip validating it again.

But, more importantly, model validation is computationally inefficient, as demonstrated by the trace below:
<img width="1919" alt="image" src="https://github.com/openfga/openfga/assets/5374887/740f8c9c-06d8-4d57-b535-2b70c29bc28b">

We are calling `hasCycle` and `hasEntrypoints` per every type and relation in the model.

### Describe the ideal solution

Improved model validation

PR #1754, opened 2024-07-09
perf: when reading models, skip model validation if the model ID is i…

miparnisari:
## Description
This is part of https://github.com/openfga/openfga/issues/1753, but doesn't solve it completely.

The change is that if a Check request is received but without specifying a model ID, we will attempt to fetch it from the cache because in this way we save from having to do model validation which is computationally expensive (in-memory).

As part of this change, I rewrote the tests (basically copied https://github.com/openfga/openfga/pull/1258/files#diff-679325381eda2c29f2d4794339a1efa278a4f65cdebc7514b9f45341c5f59ec8) to cover more lines.


pkg/typesystem/resolver_test.go
```diff
@@ -2,116 +2,196 @@ package typesystem
...
-	require.NoError(t, err)
-	require.NotNil(t, typesys)
-	require.Equal(t, modelID2, typesys.GetAuthorizationModelID())
-
-	relation, err = typesys.GetRelation("document", "viewer")
-	require.NoError(t, err)
-	require.NotNil(t, relation)
-}
-
-func TestSingleFlightMemoizedTypesystemResolverFunc(t *testing.T) {
```
miparnisari, line 77: Note: I deleted this test because it doesn't prove that singleflight works. I removed the singleflight code and this test still passes. This is because line 93 `.MinTimes(1).MaxTimes(numGoroutines)` works with and without singleflight.

pkg/typesystem/resolver.go
```diff
@@ -59,16 +66,18 @@ func MemoizedTypesystemResolverFunc(datastore storage.AuthorizationModelReadBack
...
-			model := v.(*openfgav1.AuthorizationModel)
-			key = fmt.Sprintf("%s/%s", storeID, model.GetId())
-		} else {
-			key = fmt.Sprintf("%s/%s", storeID, modelID)
-			item := cache.Get(key)
-			if item != nil {
-				return item.Value(), nil
-			}
+			modelID = model.GetId()
+			model = v.(*openfgav1.AuthorizationModel)
```
jpadilla, line 70: aren't these two lines inverted? 😕

miparnisari: added

elbuo8 (approved):

jpadilla (approved):

PR #1719, opened 2024-06-26
test: ignore goleak for mysql driver

adriantam:
## Description
Occasionally test fails due to perceived leak in mysql driver. However, in reality it is just sometimes the driver is slow in deallocating its threads.  Thus, we can safely ignore these "leaks".

This happens when running test with OpenFGA locally.  It errors with the following traces:

```
--- FAIL: TestListUsersMySQL (11.84s)
    tests.go:62: creating connection to address 0.0.0.0:55506
    testutils.go:206: 2024-06-26 10:08:01.229569 -0400 EDT m=+15.947387210 not serving yet at address 0.0.0.0:55506 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 0.0.0.0:55506: connect: connection refused"
    testutils.go:206: 2024-06-26 10:08:01.925779 -0400 EDT m=+16.643592043 not serving yet at address 0.0.0.0:55506 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp 0.0.0.0:55506: connect: connection refused"
    tests.go:56: waiting for server to stop
    tests.go:59: server stopped with error:  <nil>
    listusers_test.go:44: found unexpected goroutines:
        [Goroutine 56354 in state select, with github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1 on top of the stack:
        github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1()
        	/Users/adrian.tam/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:628 +0xb0
        created by github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher in goroutine 56341
        	/Users/adrian.tam/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:625 +0x1b4
         Goroutine 53368 in state select, with github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1 on top of the stack:
        github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1()
        	/Users/adrian.tam/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:628 +0xb0
        created by github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher in goroutine 53381
        	/Users/adrian.tam/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:625 +0x1b4
[...]

PR #1755, opened 2024-07-10
fix: ListUsers returns only the first error seen

miparnisari:
## Description
Proposal to fix flaky ListUsers tests, e.g.

```
--- FAIL: TestListUsersMemory/RunAll (0.00s)
          --- FAIL: TestListUsersMemory/RunAll/ListUsers (0.66s)
              --- FAIL: TestListUsersMemory/RunAll/ListUsers/relation_through_ttu_with_condition (0.21s)
                  --- FAIL: TestListUsersMemory/RunAll/ListUsers/relation_through_ttu_with_condition/stage_0 (0.20s)
                      --- FAIL: TestListUsersMemory/RunAll/ListUsers/relation_through_ttu_with_condition/stage_0/assertion_3 (0.01s)
                          listusers.go:187: 
                              	Error Trace:	/home/runner/work/openfga/openfga/tests/listusers/listusers.go:187
                              	Error:      	An error is expected but got nil.
                              	Test:       	TestListUsersMemory/RunAll/ListUsers/relation_through_ttu_with_condition/stage_0/assertion_3
                              	Messages:   	ListUsers request: object=document:1, relation=viewer, filters=user. Model: model
                              	            	  schema 1.1
                              	            	type user
                              	            	
                              	            	type folder
                              	            	  relations
                              	            	    define viewer: [user]
                              	            	
                              	            	type document
                              	            	  relations
                              	            	    define parent: [folder with str_cond, folder with xcond]
                              	            	    define viewer: [user] or viewer from parent
                              	            	
                              	            	condition str_cond(s: string) {
                              	            	  s == "hello"
                              	            	}
                              	            	
                              	            	condition xcond(x: int) {
                              	            	  x == 10
                              	            	}
                              	            	. Tuples: [user:"folder:a" relation:"parent" object:"document:1" condition:{name:"str_cond"} user:"folder:b" relation:"parent" object:"document:1" condition:{name:"xcond"} user:"user:jon" relation:"viewer" object:"folder:a" user:"user:jon" relation:"viewer" object:"folder:b"]. Contextual tuples: []
```

The root cause is that we have a `select` that is listening on two different channels, AND we're sending messages to the channel at roughly the same millisecond, which means [it's non-deterministic which message the `select` will see first](https://go.dev/ref/spec#Select_statements:~:text=If%20one%20or,communications%20can%20proceed).

Demo: https://go.dev/play/p/EixeM07Zolf, if you run enough times you will see the message `got here first`.

The proposal is to wait on the "done" channel first. After that, we either will have an error in the error channel, or we won't (but we won't have to block waiting for it).

## Related
- Original discussion https://github.com/openfga/openfga/pull/1714/files#r1656450644
- Supersedes https://github.com/openfga/openfga/pull/1719


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -216,26 +217,36 @@ func (l *listUsersQuery) ListUsers(
...
+		if deadlineExceeded || errors.Is(err, context.DeadlineExceeded) {
+			// We skip the error because we want to send at least partial results to the user (but we should probably set response headers)
+			break
+		}
+		telemetry.TraceError(span, err)
+		return nil, err
+	default:
+		break
+	}
+
```
willvedd, line 249: This second `select`  statement is blocked either by a value on the `doneWithFoundUsersCh` or context cancellation. If an error gets emitted, I'm not sure that either that either will get triggered. This means that even if an error gets emitted, we're blocking until the context cancels. Or so that's my understanding currently. 

Can you demonstrate with a test that we fail as soon as the error gets emitted; that we're not always waiting for context to cancel?

miparnisari: > If an error gets emitted, I'm not sure that either that either will get triggered.

The line `case <-doneWithFoundUsersCh:` will be executed after `foundUsersCh` is closed, which is after `expandErrCh <- resp.err`

willvedd: >The line case <-doneWithFoundUsersCh: will be executed after foundUsersCh is closed, which is after expandErrCh <- resp.err.

Ok I see this now, makes sense.

I'd also like to double-check (ideally with a test) that the appropriate tests pass when setting a deadline to `0`.

pkg/server/list_users_test.go
```diff
@@ -426,7 +427,7 @@ func TestListUsers_Deadline(t *testing.T) {
 
 		mockDatastore.EXPECT().
 			Read(gomock.Any(), storeID, gomock.Any()).
-			Return(nil, context.DeadlineExceeded).
```
willvedd, line 429: Why the change here?

internal/utils/concurrency.go
```diff
@@ -0,0 +1,17 @@
+package utils
```
willvedd, line 1: **Non-blocking idea:** Would you consider having this concurrency package nested inside `internal` instead of being in the generic `utils` directory? It might make for a nicer import and implementation:

```
pool := concurrency.NewPool(ctx, int(l.resolveNodeBreadthLimit))
```

Or possibly:

```
concurrencyPool := pool.New(ctx, int(l.resolveNodeBreadthLimit))
```

willvedd (commented):
I approve of this in principle but would like to see a test that ensures that the emitting of an error isn't blocked by context cancellation or the end of results. This doesn't appear to be the case but having confidence here would be great, especially since this is a such critical junction in the code.

internal/utils/concurrency.go
miparnisari: Moved it to `internal/concurrency`

miparnisari:
> a test that ensures that the emitting of an error isn't blocked by context cancellation or the end of results

Sorry, I don't understand. I haven't changed how we emit errors, only how we consume them. When we emit an error, it will trigger the `case <-doneWithFoundUsersCh:` line

pkg/server/commands/listusers/list_users_rpc.go
miparnisari: Added `infinite_deadline_does_not_block_return_of_errors`

willvedd (approved):

PR #1758, opened 2024-07-10
feat: update cached resolver cache when not reading from the cache

ewanharris:
## Description

This PR introduces writing to the cache in `CachedCheckResolver` even when we are not reading from it due to the consistency requested being `HIGHER_CONSISTENCY`. Previously when we implemented skipping the cache we chose to not update the cache, but as it is relatively simple to do so we are now choosing to enrich the cache.

It refactors the existing test for this behaviour to now require the cache to be updated, and updates the list objects test that relied on cache interaction to reflect the new reality.

## Review Checklist

### DIFF
--- a/internal/graph/cached_resolver.go
+++ b/internal/graph/cached_resolver.go
@@ -6,13 +6,14 @@ import (
 	"strconv"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
+
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 
 	"github.com/cespare/xxhash/v2"
 	"github.com/karlseguin/ccache/v3"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
-	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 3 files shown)


rhamzeh (approved):

internal/graph/cached_resolver.go
```diff
@@ -150,37 +151,30 @@ func (c *CachedCheckResolver) ResolveCheck(
...
 		telemetry.TraceError(span, err)
 		return nil, err
 	}
 
-	cachedResp := c.cache.Get(cacheKey)
-	isCached := cachedResp != nil && !cachedResp.Expired()
-	span.SetAttributes(attribute.Bool("is_cached", isCached))
-	if isCached {
-		checkCacheHitCounter.Inc()
+	skipCache := c.enableConsistencyOptions && req.Consistency == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY
```
adriantam, line 161: As an aside, we should have measurement tracking how frequently we have skip cache

internal/graph/cached_resolver.go
```diff
@@ -6,13 +6,14 @@ import (
 	"strconv"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
```
adriantam, line 9: nitpik: why is this moved?

adriantam (approved):

internal/graph/cached_resolver.go
ewanharris: We have telemetry data that would allow us to count the total number of requests per consistency type (#1721), do you think we could use that to get the data you're thinking of?

PR #1761, opened 2024-07-11
perf(check): use a sorted set to store usersets

miparnisari:
## Description
Follow-up to https://github.com/openfga/openfga/pull/1734 and https://github.com/openfga/openfga/pull/1735.

This PR is a breaking change to the interface `ReadStartingWIthUser`. Now, the `objectIDs` parameter will be given as a **sorted** list so that datastore implementations don't have to sort it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/docker/docker v26.0.2+incompatible
 	github.com/docker/go-connections v0.5.0
+	github.com/emirpasic/gods v1.18.1
 	github.com/go-sql-driver/mysql v1.8.1
 	github.com/golang-jwt/jwt/v5 v5.2.1
 	github.com/google/cel-go v0.20.1
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -21,8 +21,6 @@ import (
 	"github.com/openfga/openfga/pkg/telemetry"
 	"github.com/openfga/openfga/pkg/tuple"
 	"github.com/openfga/openfga/pkg/typesystem"
-
-	"github.com/emirpasic/gods/sets/treeset"
 )
 
 var tracer = otel.Tracer("internal/graph/check")
@@ -620,9 +618,9 @@ func (c *LocalChecker) ResolveCheck(
 // For example, given [group:1#member, group:2#member, group:1#owner, group:3#owner] it will be stored as:
 // [group#member][1, 2]
 // [group#owner][1, 3].
-type usersetsMapType map[string]*treeset.Set
+type usersetsMapType map[string]storage.SortedSet
 
-func (c *LocalChecker) buildCheckAssociatedObjects(req *ResolveCheckRequest, objectRel string, objectIDs *treeset.Set) CheckHandlerFunc {
+func (c *LocalChecker) buildCheckAssociatedObjects(req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		ctx, span := tracer.Start(ctx, "checkAssociatedObjects")
 		defer span.End()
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/docker/docker v26.0.2+incompatible
 	github.com/docker/go-connections v0.5.0
+	github.com/emirpasic/gods v1.18.1
```
miparnisari, line 12: Approved by CNCF: https://github.com/cncf/foundation/blob/main/license-exceptions/CNCF-licensing-exceptions.csv

internal/graph/check.go
```diff
@@ -659,11 +667,15 @@ func (c *LocalChecker) buildCheckAssociatedObjects(req *ResolveCheckRequest, obj
 			})
 		}
 
+		objectIDsArrayOfStrings := make([]string, objectIDs.Size())
```
adriantam, line 670: Why not just call 
```
objectIDsArrayOfStrings = objectIDs.Values()
```
?

miparnisari: Hehe because it doesn't work:

```go
objectIDsArrayOfStrings := objectIDs.Values().([]string)
// Invalid type assertion: objectIDs.Values().([]string) (non-interface type []interface{} on the left)
```

pkg/storage/storage.go
```diff
@@ -168,10 +170,16 @@ type RelationshipTupleWriter interface {
...
+	// Mandatory.
 	ObjectType string
-	Relation   string
+	// Mandatory.
+	Relation string
+	// Mandatory.
 	UserFilter []*openfgav1.ObjectRelation
-	ObjectIDs  []string // Optional. If present, the datastore returns the intersection between this filter and what is in the database
+
+	// Optional. If present, it will be sorted in ascending order.
```
adriantam, line 180: consider whether we really need this to be ascending order OR have the individual datastore implementation re-order as needed

miparnisari: If we require all datastore implementations to sort then I might as well scrap this PR 🤔

adriantam: In this case, should we rename it to SortedObjectIDs to make it very clear that it is to be sorted.  (Though this will be a breaking change with respect to the interface)

internal/graph/check.go
```diff
@@ -659,11 +667,15 @@ func (c *LocalChecker) buildCheckAssociatedObjects(req *ResolveCheckRequest, obj
...
+		for i, objectID := range objectIDs.Values() {
+			objectIDsArrayOfStrings[i] = objectID.(string)
+		}
 		i, err := ds.ReadStartingWithUser(ctx, storeID, storage.ReadStartingWithUserFilter{
 			ObjectType: objectType,
 			Relation:   relation,
 			UserFilter: userFilter,
-			ObjectIDs:  maps.Keys(objectIDs),
+			ObjectIDs:  objectIDsArrayOfStrings,
```
elbuo8, line 678: Why convert it into a slice? We can let the storage implementation decide how they want to use/implement it.

pkg/storage/storage.go
elbuo8: To avoid all this, why not just pass the sorted set directly?

internal/graph/check.go
miparnisari: I wasn't sure whether we wanted to break the interface. I can do that

pkg/storage/memory/memory.go
```diff
@@ -504,10 +503,8 @@ func (s *MemoryBackend) ReadStartingWithUser(
 			continue
 		}
 
-		if len(filter.ObjectIDs) > 0 {
-			if !slices.Contains(filter.ObjectIDs, t.ObjectID) {
-				continue
-			}
+		if filter.ObjectIDs != nil && !filter.ObjectIDs.Exists(t.ObjectID) {
```
miparnisari, line 506: I don't love this `nil` check here :/

elbuo8 (approved):

pkg/storage/sortedset.go
```diff
@@ -0,0 +1,56 @@
...
+func (r *RedBlackTreeSet) Exists(key string) bool {
+	_, ok := r.inner.Get(key)
+	return ok
+}
+
+func (r *RedBlackTreeSet) Size() int {
+	return r.inner.Size()
+}
+
+func (r *RedBlackTreeSet) Values() []string {
```
elbuo8, line 50: Could we accept an optional size to obtain `up to that size`? Or a [begins, ends]? That would set us up for batches in the future 🤔 

Or we can leave it for the N+1 version.

miparnisari: N+1 😆

elbuo8 (approved):

PR #1764, opened 2024-07-12
feat!: introduce consistency options for query requests

ewanharris:
## Description

This PR introduces the ability to request different consistency options per request, with the currently supported values being `MINIMIZE_LATENCY` (the existing behaviour, and the default) and `HIGHER_CONSISTENCY`. This PR is made up of a number of PRs and the most interesting ones are summarised below:

* #1732 - The storage interface has been changed so that every method that previously took a `PaginationOptions` struct now takes a method specific struct. For storage implementers, please review that PR for more details
* #1750 - The relevant storage interface options structs were updated to now include a `ConsistencyOptions` struct that includes the `Preference` property. This indicates the consistency preference of the request. 
* #1721 - We've added a `consistency` attribute to the `request_duration_ms` histogram and relevant traces so that usage can be tracked and whether a cache was skipped (i.e. `consistency == "HIGHER_CONSISTENCY"`) can be considered as part of evaluating response times

When `HIGHER_CONSISTENCY` is requested, OpenFGA will skip reading from the cache in the `CachedCheckResolver`, preferring to read directly from the datastore instead. The cache is however enriched with the value from the datastore.

For datastore implementors we recommend following that pattern where, if possible, any cache used as part of the datastore is enriched with the result.

This feature is current behind the `enable-consistency-params` experimental and can be enabled by setting the `--experimentals` CLI option or `OPENFGA_EXPERIMENTALS` environment variable.

Prior to merging, I will handle merging the `openfga/api` changes and update this branch.

BREAKING CHANGE
Note: This is a breaking change to the datastore interface. If you are using OpenFGA as a library with a custom datastore, there will be some changes you need to make to conform to the new interface.

## References

- https://github.com/openfga/roadmap/issues/54
- [Correlating OpenFGA API PR](https://github.com/openfga/api/pull/180)

## Review Checklist


adriantam:
Given https://github.com/openfga/openfga/pull/1732 is a breaking change, should this PR be labelled as a breaking change?

.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
-* @openfga/backend
+* @openfga/backend @openfga/dx
```
ewanharris, line 2: I'll revert this prior to merge so that I can still easily get any changes into the `feat/stronger-consistency-options` branch if needed

internal/graph/cached_resolver.go
```diff
@@ -140,7 +150,6 @@ func (c *CachedCheckResolver) ResolveCheck(
 	req *ResolveCheckRequest,
 ) (*ResolveCheckResponse, error) {
 	span := trace.SpanFromContext(ctx)
-	checkCacheTotalCounter.Inc()
 
 	cacheKey, err := CheckRequestCacheKey(req)
```
willvedd, line 154: **Suggestion:** Move cache key generation below to immediately above where its used. Should avoid creating the key if we're not going to use it.

internal/graph/check.go
```diff
@@ -149,6 +151,13 @@ func (r *ResolveCheckRequest) GetContext() *structpb.Struct {
 	return nil
 }
 
+func (r *ResolveCheckRequest) GetConsistency() openfgav1.ConsistencyPreference {
+	if r != nil {
+		return r.Consistency
+	}
+	return openfgav1.ConsistencyPreference_MINIMIZE_LATENCY
```
willvedd, line 158: **Question:** I know this is the general default for the functionality but is `MINIMIZE_LATENCY` really the right setting to be returned here? Shouldn't this function instead be returning `UNSPECIFIED`?

willvedd: Additionally, codecov is indicating that L158 isn't covered. Is there a practical way of getting it tested?

pkg/server/commands/list_objects_test.go
```diff
@@ -225,3 +227,109 @@ func TestListObjectsDispatchCount(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Len(t, resp.Objects, 3)
+
+	// Rerun check with MINIMIZE_LATENCY to ensure the cache was updated with the tuple we retrieved during the previous call
+	resp, err = q.Execute(ctx, &openfgav1.ListObjectsRequest{
+		StoreId:     storeID,
+		Type:        "folder",
+		Relation:    "viewer",
+		User:        "user:jon",
+		Consistency: openfgav1.ConsistencyPreference_MINIMIZE_LATENCY,
```
willvedd, line 330: **Question:** Is it worth it to also test the `UNSPECIFIED` setting?

pkg/server/list_users_test.go
```diff
@@ -244,6 +244,28 @@ func TestExperimentalListUsers(t *testing.T) {
...
+			},
+			Relation: "viewer",
+			UserFilters: []*openfgav1.UserTypeFilter{
+				{Type: "user"},
+			},
+			Consistency: openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY,
+		})
+
+		require.Error(t, err)
+		require.Equal(t, "rpc error: code = InvalidArgument desc = Consistency parameters is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-consistency-params` configuration option when running OpenFGA server", err.Error())
```
willvedd, line 262:
```suggestion
		require.Equal(t, "rpc error: code = InvalidArgument desc = Consistency parameters are not enabled. They can be enabled for experimental use by passing the `--experimentals enable-consistency-params` configuration option when running OpenFGA server", err.Error())
```
?

pkg/server/server.go
```diff
@@ -1297,3 +1337,12 @@ func (s *Server) resolveTypesystem(ctx context.Context, storeID, modelID string)
 
 	return typesys, nil
 }
+
+// If the requested consistency preference is HIGHER_CONSISTENCY, but the experimental flag is not enabled,
+// returns an error.
+func (s *Server) validateConsistencyRequest(c openfgav1.ConsistencyPreference) error {
+	if !s.IsExperimentallyEnabled(ExperimentalEnableConsistencyParams) && openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY == c {
+		return status.Error(codes.InvalidArgument, "Consistency parameters is not enabled. It can be enabled for experimental use by passing the `--experimentals enable-consistency-params` configuration option when running OpenFGA server")
```
willvedd, line 1345:
```suggestion
		return status.Error(codes.InvalidArgument, "Consistency parameters are not enabled. They can be enabled for experimental use by passing the `--experimentals enable-consistency-params` configuration option when running OpenFGA server")
```
?

pkg/server/server.go
```diff
@@ -1297,3 +1337,12 @@ func (s *Server) resolveTypesystem(ctx context.Context, storeID, modelID string)
 
 	return typesys, nil
 }
+
+// If the requested consistency preference is HIGHER_CONSISTENCY, but the experimental flag is not enabled,
+// returns an error.
+func (s *Server) validateConsistencyRequest(c openfgav1.ConsistencyPreference) error {
+	if !s.IsExperimentallyEnabled(ExperimentalEnableConsistencyParams) && openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY == c {
```
willvedd, line 1344: **Question:** Should we not also error when the `MINIMIZE_LATENCY` option is passed too? If the experimental flag isn't enabled, it may give the incorrect impression that their configuration was successfully applied; it may not be apparent that it's the default behavior by happenstance.

willvedd (commented):
Looks very sensible overall, including tests. Some questions.

internal/graph/cached_resolver.go
ewanharris: I'm not sure if we have the opportunity to move this as it's always used in this function, we will either read from the cache using it or write to the cache using it

internal/graph/check.go
ewanharris: Yeah it probably should be returning `UNSPECIFIED`.

I'll try and see if we can test that line, I'm not certain if it's possible because I _think_ the `consistency` parameter is set to `UNSPECIFIED` if it's not included in the request.

pkg/server/server.go
ewanharris: Yup this should most likely be checking it it's not `UNSPECIFIED`.

internal/graph/check.go
ewanharris: Yeah taking a look through under the debugger, if `consistency` is not set on a `ResolveCheckRequest` then it will be set to `UNSPECIFIED`. I believe this is because we set `UNSPECIFIED` as the zero value for `openfgav1.ConsistencyPreference`

internal/graph/cached_resolver_test.go
```diff
@@ -874,6 +933,30 @@ func TestCheckCacheKeyWithContext(t *testing.T) {
 	require.NotEqual(t, key3, key4)
 }
 
+func TestErrorReturnedWhenHigherConsistencyRequested(t *testing.T) {
```
miparnisari, line 936: I'm not sure i understand this test - if the consistency was `minimize_latency` but the result was not in the cache, we should also return an error, no?

What are you really testing here? It seems like you are testing that `cachedCheckResolver` is delegating to the mock, but don't we already have tests for that?

internal/graph/cycle_check_resolver.go
```diff
@@ -59,6 +59,7 @@ func (c *CycleDetectionCheckResolver) ResolveCheck(
 		RequestMetadata:      req.GetRequestMetadata(),
 		VisitedPaths:         req.VisitedPaths,
 		Context:              req.GetContext(),
+		Consistency:          req.GetConsistency(),
```
miparnisari, line 62: Instead of making this change you could change to `return c.delegate.ResolveCheck(ctx, clone(req))`

internal/graph/cached_resolver.go
```diff
@@ -149,16 +158,23 @@ func (c *CachedCheckResolver) ResolveCheck(
...
 	}
 
-	cachedResp := c.cache.Get(cacheKey)
-	isCached := cachedResp != nil && !cachedResp.Expired()
-	span.SetAttributes(attribute.Bool("is_cached", isCached))
-	if isCached {
-		checkCacheHitCounter.Inc()
+	skipCache := c.enableConsistencyOptions && req.Consistency == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY
+
+	if !skipCache {
```
miparnisari, line 163:
```suggestion
	tryCache := !c.enableConsistencyOptions || req.Consistency == openfgav1.ConsistencyPreference_MINIMIZE_LATENCY

	if tryCache {
```

because double negation makes my brain hurt

pkg/server/server_test.go
```diff
@@ -1981,3 +1981,180 @@ func TestIsExperimentallyEnabled(t *testing.T) {
 		require.False(t, s.IsExperimentallyEnabled(someExperimentalFlag))
 	})
 }
+
+func TestErrorThrownIfConsistencyRequestedWithoutFlagEnabled(t *testing.T) {
+	ds := memory.New()
+	openfga := MustNewServerWithOpts(WithDatastore(ds))
+
+	t.Run("check_throws_error_if_higher_consistency_requested_without_flag", func(t *testing.T) {
```
miparnisari, line 1989:
```suggestion
func TestErrorThrownIfConsistencyRequestedWithoutFlagEnabled(t *testing.T) {
        t.Cleanup(func() {
		goleak.VerifyNone(t)
	})
	ds := memory.New()
	t.Cleanup(ds.Close)
	openfga := MustNewServerWithOpts(WithDatastore(ds))
	t.Cleanup(openfga.Stop)

	t.Run("check_throws_error_if_higher_consistency_requested_without_flag", func(t *testing.T) {
```

pkg/server/commands/list_objects.go
```diff
@@ -300,7 +301,7 @@ func (q *ListObjectsQuery) evaluate(
 				User:             sourceUserRef,
 				ContextualTuples: req.GetContextualTuples().GetTupleKeys(),
 				Context:          req.GetContext(),
-			}, reverseExpandResultsChan, reverseExpandResolutionMetadata)
+			}, reverseExpandResultsChan, reverseExpandResolutionMetadata, storage.ConsistencyOptions{Preference: req.GetConsistency()})
```
miparnisari, line 304: why not make `consistency` part of `ReverseExpandRequest`? it would be more readable

internal/graph/cached_resolver.go
ewanharris: Heh I had been wanting to change this to avoid double negation but couldn't decide on the right wording. I think we might want to do `!= ConsistencyPreference_HIGHER_CONSISTENCY` as we want `ConsistencyPreference_UNSPECIFIED` to use the cache

internal/graph/cached_resolver_test.go
ewanharris: Yeah the consistency setting doesn't have an impact here, I think realistically this is covered by the tests in the test table at the top of the file so can be removed

pkg/server/commands/list_objects.go
ewanharris: I think that's a good call as it's request data realistically

ewanharris: Done in c3c6d511e255f424d3c85e7a61f9e376cbc5f478

pkg/server/server_test.go
ewanharris: Done in c3c6d511e255f424d3c85e7a61f9e376cbc5f478, I had to add some `Close` calls in the `TestIsExperimentallyEnabled` test in order to get this goleak check happy

internal/graph/cached_resolver.go
ewanharris: Done in c3c6d511e255f424d3c85e7a61f9e376cbc5f478, also added a test to cover the usage of `UNSPECIFIED` trying the cache

internal/graph/cycle_check_resolver.go
ewanharris: Done in c3c6d511e255f424d3c85e7a61f9e376cbc5f478

internal/graph/cached_resolver_test.go
ewanharris: Removed in c3c6d511e255f424d3c85e7a61f9e376cbc5f478

miparnisari (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,32 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 
 ## [Unreleased]
 
+### Added
+
+* Support requesting a different consistency option per request in `Check`, `Expand`, `ListObjects`, `ListUsers`, and `Read` [#1764](https://github.com/openfga/openfga/pull/1764)
+  * When `HIGHER_CONSISTENCY` is requested, OpenFGA will skip the check resolver cache. For storage implementors it is recommended to skip any caching and perform a stronger read if `HIGHER_CONSISTENCY` is requested. This can be accessed in the `Consistency` options provided to the relevant methods of the storage interface.
+
+### Breaking Changes :warning:
```
jpadilla, line 16: Let's add something along the lines of `If you implement your own storage adapter, you will need to make the following changes:`

jpadilla (changes requested):

jpadilla (approved):

PR #1765, opened 2024-07-12
refactor: Using typesystem to determine when to apply TTU fast path 

willvedd:
## Description
Determining the ultimate terminal types and terminal relations for a given object type and relation. This data is calculated at the time of typesystem creation and validations (`t.NewAndValidate()`) and associated in a multi-key map on the typesystem struct.

This work enables two different optimizations. The primary is an optimized TTU and userset lookup which requires knowledge of the terminal types and relations for a given userset to craft the appropriate query. Currently, this is limited to usersets that result in a direct relation and necessarily doesn't result in a computed relation, union, intersection, etc. This limitation can be observed in #1733. 

This PR _only_ supports TTU that result in direct relations, computed relations and public wildcards. It does not yet support the following rewrites: union, intersection, difference or usersets. However, the code establishes a pattern that will easily enable these in the future.

This work also has the potential to be used as an optimization in Check, ListObjects and ListUsers as a screen to determine if the request's relation and types are valid with respect to the model. If not, we can exit the process with a falsey outcome and prevent some DB requests. This work was previously visited in #1582 but this PR will reduce the computation to a O(1) lookup at runtime.

## References
- Supports: https://github.com/openfga/openfga/pull/1733, https://github.com/openfga/openfga/pull/1735, https://github.com/openfga/openfga/pull/1734
- Supersedes #1736 
- May help supersede: https://github.com/openfga/openfga/pull/1582

## Review Checklist

### DIFF
--- a/internal/graph/cached_resolver_test.go
+++ b/internal/graph/cached_resolver_test.go
@@ -654,7 +654,7 @@ func TestCachedCheckDatastoreQueryCount(t *testing.T) {
 	ttuLocalChecker.Close()
 
 	require.NoError(t, err)
-	require.Equal(t, uint32(2), res.GetResolutionMetadata().DatastoreQueryCount)
+	require.Equal(t, uint32(1), res.GetResolutionMetadata().DatastoreQueryCount)
 }
 
 func TestCachedCheckResolver_ResolveCheck_After_Stop_DoesNotPanic(t *testing.T) {
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -1211,8 +1211,8 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		// If the user is a userset, we will not be able to use the shortcut because the algo
 		// will look up the objects associated with user.
 		if !tuple.IsObjectRelation(tk.GetUser()) {
-			if canShortCircuit := typesys.TTUCanFastPath(
-				tuple.GetType(object), req.GetTupleKey().GetRelation(), tuple.GetType(req.GetTupleKey().GetUser())); canShortCircuit {
+			if canFastPath := typesys.TTUCanFastPath(
+				tuple.GetType(object), req.GetTupleKey().GetRelation(), tuple.GetType(req.GetTupleKey().GetUser())); canFastPath {
 				resolver = c.checkTTUFastPath
 			}
 		}
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -1077,15 +1077,28 @@ func (c *LocalChecker) checkTTUSlowPath(ctx context.Context, req *ResolveCheckRe
 //
 // check(user, viewer, doc) will find the intersection of all group assigned to the doc's parent AND
 // all group where the user is a member of.
+
 func (c *LocalChecker) checkTTUFastPath(ctx context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, iter *storage.ConditionsFilteredTupleKeyIterator) (*ResolveCheckResponse, error) {
 	ctx, span := tracer.Start(ctx, "checkTTUFastPath")
 	defer span.End()
 
+	typesys, ok := typesystem.TypesystemFromContext(ctx)
+	if !ok {
+		return nil, fmt.Errorf("typesystem missing in context")
+	}
+
+	terminalRelations := typesys.GetTerminalRelationsForTTUFastPath(
+		tuple.GetType(req.GetTupleKey().GetObject()), req.GetTupleKey().GetRelation(), tuple.GetType(req.GetTupleKey().GetUser()),
+	)
+	if len(terminalRelations) != 1 {
+		return nil, fmt.Errorf("expected exactly one terminal relation for fast path, received %d", len(terminalRelations))
+	}
+
+	computedRelation := terminalRelations[0]
 	// usersetsMap is a map of all ObjectRelations and its Ids. For example,
 	// [group:1#member, group:2#member, group:1#owner, group:3#owner] will be stored as
 	// [group#member][1, 2]
 	// [group#owner][1, 3]
-	computedRelation := rewrite.GetTupleToUserset().GetComputedUserset().GetRelation()
 	usersetsMap := make(map[string]map[string]struct{})
 
 	for {
@@ -1198,8 +1211,8 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		// If the user is a userset, we will not be able to use the shortcut because the algo
 		// will look up the objects associated with user.
 		if !tuple.IsObjectRelation(tk.GetUser()) {
-			if canShortCircuit, err := typesys.TTUResolvesExclusivelyToDirectlyAssignable(
-				tuple.GetType(object), tuplesetRelation, computedRelation); err == nil && canShortCircuit {
+			if canShortCircuit := typesys.TTUCanFastPath(
+				tuple.GetType(object), req.GetTupleKey().GetRelation(), tuple.GetType(req.GetTupleKey().GetUser())); canShortCircuit {
 				resolver = c.checkTTUFastPath
 			}
 		}
--- a/pkg/typesystem/connected_types.go
+++ b/pkg/typesystem/connected_types.go
@@ -64,7 +64,6 @@ func (t *TypeSystem) getTerminalUserTypeAndRelationsForConnectedTypes(
 	switch rw := rewrite.GetUserset().(type) {
 	case *openfgav1.Userset_This:
 		assignableTypes := []string{}
-		var terminalRelation string
 
 		thisRelation, ok := t.relations[typeName][relationName]
 		if !ok {
@@ -81,14 +80,12 @@ func (t *TypeSystem) getTerminalUserTypeAndRelationsForConnectedTypes(
 				t = tuple.TypedPublicWildcard(assignableType.GetType())
 			}
 			assignableTypes = append(assignableTypes, t)
-
-			terminalRelation = relationName
 		}
 
 		return []terminalTypesAndRelation{
 			{
 				terminalTypes:    assignableTypes,
-				terminalRelation: terminalRelation,
+				terminalRelation: relationName,
 			},
 		}
 	case *openfgav1.Userset_ComputedUserset:
--- a/pkg/typesystem/connected_types.go
+++ b/pkg/typesystem/connected_types.go
@@ -78,6 +78,7 @@ func (t *TypeSystem) getTerminalUserTypeAndRelationsForConnectedTypes(
 			t := assignableType.GetType()
 			if assignableType.GetWildcard() != nil {
 				t = tuple.TypedPublicWildcard(assignableType.GetType())
+				// return []terminalTypesAndRelation{}
 			}
 			assignableTypes = append(assignableTypes, t)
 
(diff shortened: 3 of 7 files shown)


internal/graph/check.go
```diff
@@ -1077,15 +1077,28 @@ func (c *LocalChecker) checkTTUSlowPath(ctx context.Context, req *ResolveCheckRe
...
+	if !ok {
+		return nil, fmt.Errorf("typesystem missing in context")
+	}
+
+	terminalRelations := typesys.GetTerminalRelationsForTTUFastPath(
+		tuple.GetType(req.GetTupleKey().GetObject()), req.GetTupleKey().GetRelation(), tuple.GetType(req.GetTupleKey().GetUser()),
+	)
+	if len(terminalRelations) != 1 {
+		return nil, fmt.Errorf("expected exactly one terminal relation for fast path, received %d", len(terminalRelations))
+	}
```
willvedd, line 1095: For now, direct and computed relations will only result in a single terminal relation. There should never be more than one terminal relation in the code's current state; this error helps catch this during testing. However, once support for the union and intersection rewrites is added, this will change.

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+}
+
+type terminalTypesAndRelation struct {
+	terminalTypes    []string
+	terminalRelation string
+}
+
+func (t *TypeSystem) getTerminalUserTypeAndRelationsForConnectedTypes(
```
willvedd, line 39: This function is the core of the PR. During typesystem create, recursively traverses it to determine a mapping of all the terminal types and terminal relations. As more scenarios are supported in the future, this function will need to be updated.

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,123 @@
...
+				terminalTypes:    assignableTypes,
+				terminalRelation: terminalRelation,
+			},
+		}
+	case *openfgav1.Userset_ComputedUserset:
+		return t.getTerminalUserTypeAndRelationsForConnectedTypes(typeName, rw.ComputedUserset.GetRelation(), numTTU)
+	case *openfgav1.Userset_TupleToUserset:
+		if numTTU > 0 {
+			return []terminalTypesAndRelation{}
+		}
```
willvedd, line 98: This check ensures that no chained TTU rewrites get applied to fast path.

pkg/typesystem/typesystem.go
```diff
@@ -411,33 +414,19 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
-	for _, tuplesetRelationType := range tuplesetRelationTypes {
-		_, childDirectlyAssignable, err := t.resolvesTypeRelationToDirectlyAssignable(tuplesetRelationType, computedRelation)
-		if err != nil {
-			// in the case of errors due to relation undefined, we can ignore the error because it is possible
-			// that some parents do not have the relation defined.
-			if errors.As(err, &relationUndefinedError) {
-				continue
-			}
+// TTUCanFastPath returns whether object's tupleRelation's rewrite can support the fast path optimization.
+func (t *TypeSystem) TTUCanFastPath(objectType, computedRelation, userType string) bool {
```
willvedd, line 418: As we add support for more scenarios, the question shifts from "Does TTU resolve to directly assignable?" to "Can we fast path?".

pkg/typesystem/typesystem.go
```diff
@@ -850,6 +839,7 @@ func NewAndValidate(ctx context.Context, model *openfgav1.AuthorizationModel) (*
 			if err != nil {
 				return nil, err
 			}
+			t.AssignTerminalTypes(typeName, relationName)
```
willvedd, line 842: Easy to miss in this PR but the `NewAndValidate` function is where the connected types mapping occurs. The `NewAndValidate` is memoized and is only called once per store operation per node lifetime.

pkg/typesystem/typesystem_test.go
```diff
@@ -3242,11 +3241,10 @@ func TestTTUResolvesExclusivelyToDirectlyAssignable(t *testing.T) {
...
 					`,
-			objectType:               "folder",
-			tuplesetRelation:         "parent",
-			computedRelation:         "member",
-			expectDirectlyAssignable: true,
-			expectError:              false,
+			objectType:        "folder",
+			computedRelation:  "viewer",
+			userType:          "user",
+			expectCanFastPath: true,
```
willvedd, line 3247: The lookup doesn't use the tupleset relation, instead it uses the object type, the target relation and the user type.

internal/graph/check.go
adriantam: Similarly, what if we have something like 
```
type user1
type user2
type group1
  relations
    define member: [user1, user2]
type group2
type folder
  relations
    define owner: [group1, group2]
    define viewer: member from owner
```

pkg/typesystem/connected_types.go
elbuo8: Please add this as an actual comment.

internal/graph/check.go
```diff
@@ -1198,8 +1211,8 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		// If the user is a userset, we will not be able to use the shortcut because the algo
 		// will look up the objects associated with user.
 		if !tuple.IsObjectRelation(tk.GetUser()) {
-			if canShortCircuit, err := typesys.TTUResolvesExclusivelyToDirectlyAssignable(
-				tuple.GetType(object), tuplesetRelation, computedRelation); err == nil && canShortCircuit {
+			if canShortCircuit := typesys.TTUCanFastPath(
```
elbuo8, line 1214: `TTUResolvesExclusivelyToDirectlyAssignable` can be removed in this PR to ensure our replacement path.

I assume you will be building a subsequent PR that would update the userset fast path?

willvedd: > I assume you will be building a subsequent PR that would update the userset fast path?

That would be a logical next step, yes.

internal/graph/check.go
willvedd: For the first scenario, we'll return `true` because both `group1#member` and `group2#member` usersets are fasth-path eligible. This is captured in the `multiple_ttu_references_to_multiple_types` test.

For the second, we'll return `true` because `group1#member` is fast-path eligible and `group2` lacks the `member` relation so we essentially skip it. This is captured in the `multiple_ttu_references_only_one_has_tupleset_relation` test.

pkg/typesystem/connected_types_test.go
```diff
@@ -0,0 +1,237 @@
...
+					},
+				},
+				"folder": map[string]map[string][]string{
+					"owner": {
+						"user": {"owner"},
+					},
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
willvedd: Captured in `multiple_ttu_references` of `typesystem_test.go`

pkg/typesystem/connected_types_test.go
```diff
@@ -0,0 +1,237 @@
...
+						define owner: [group]
+						define viewer: member from owner
+				type document
+					relations
+						define banned: [user]
+						define owner: [folder]
+						define viewer: viewer from owner
+						define can_view: viewer but not banned
+						define can_see: can_view
+			`,
```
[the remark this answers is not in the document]
willvedd: Added.

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,123 @@
...
+
+			assignableTypeName := assignableType.GetType()
+
+			if _, ok := t.relations[assignableTypeName][computedRelationName]; ok {
+				return t.getTerminalUserTypeAndRelationsForConnectedTypes(assignableTypeName, computedRelationName, numTTU+1)
+			}
+		}
+	}
+
+	return []terminalTypesAndRelation{}
```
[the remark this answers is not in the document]
willvedd: Agreed. Added.

internal/graph/cached_resolver_test.go
```diff
@@ -654,7 +654,7 @@ func TestCachedCheckDatastoreQueryCount(t *testing.T) {
 	ttuLocalChecker.Close()
 
 	require.NoError(t, err)
-	require.Equal(t, uint32(2), res.GetResolutionMetadata().DatastoreQueryCount)
```
willvedd, line 657: I'm not sure why in this PR specifically, but I'm able to corroborate that datastore only gets read once with this test.

pkg/typesystem/typesystem.go
```diff
@@ -411,33 +414,37 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
+	tuplesetRelation := t.relations[objectType][computedRelation].GetRewrite().GetTupleToUserset().GetTupleset().GetRelation()
+
+	computedUsersetRelation := t.relations[objectType][computedRelation].GetRewrite().GetTupleToUserset().GetComputedUserset().GetRelation()
+	ttuParentTypes := t.relations[objectType][tuplesetRelation].GetTypeInfo().GetDirectlyRelatedUserTypes()
+
+	if len(ttuParentTypes) > 1 {
+		// For TTU with multiple assignable types, need to verify that each type has the computed relation and is eligible for fast-path
+		for _, parentType := range ttuParentTypes {
+			_, relationExists := t.relations[parentType.GetType()][computedUsersetRelation]
+			if !relationExists {
```
willvedd, line 428: This swath of code is necessary for detecting cases where there are multiple possible TTU type restrictions defined where one or more of the tupleset relations are ineligible for fast-path. Example:

```
model
	schema 1.1
type user
type folder
	relations
	define owner: [user]
	define viewer: [user] or owner # TTU slow path because union
type document
	relations
	define can_read: viewer from parent
	define parent: [document, folder]
	define viewer: [user] # TTU fast path
```

The TTU relation `document#parent` directs to two relations, `document#viewer` and `folder#viewer`. The former is eligible for fast-path but the latter is not because of the union. Because of this, the entire `document#parent` relation is ineligible for fast-path resolution.

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,123 @@
...
+					terminalTypes:    []string{subjectType},
+					terminalRelation: terminalRelation,
+				})
+			}
+		}
+
+		return terminalTypesAndRelations
+	}
+
+	relation, ok := t.relations[typeName][relationName]
```
adriantam, line 63: question: Do we need to check for `t.relations[typeName][relationName]` in the case rewrite.GetUserset().(type) is not `*openfgav1.Userset_This`?  I see that we only use relation in the "this" path.  (Or this is a case where we should never happen)..

willvedd: Good point. I've change this to only apply to the `Userset_This` rewrite.

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+		}
+
+		return terminalTypesAndRelations
+	}
+
+	switch rw := rewrite.GetUserset().(type) {
+	case *openfgav1.Userset_This:
+		assignableTypes := []string{}
+		var terminalRelation string
```
adriantam, line 66: Why do we need this variable?  I thought relationName should not change.  So, when we return in line 89-94, we can do something like 
```
		return []terminalTypesAndRelation{
			{
				terminalTypes:    assignableTypes,
				terminalRelation: relationName,
			},
		}
```

pkg/typesystem/connected_types.go
```diff
@@ -0,0 +1,123 @@
...
+		for _, assignableType := range thisRelation.GetTypeInfo().GetDirectlyRelatedUserTypes() {
+			if assignableType.GetRelation() != "" {
+				// Usersets not yet supported
+				return []terminalTypesAndRelation{}
+			}
+
+			t := assignableType.GetType()
+			if assignableType.GetWildcard() != nil {
+				t = tuple.TypedPublicWildcard(assignableType.GetType())
+				// return []terminalTypesAndRelation{}
```
adriantam, line 82: should we clean this up?

adriantam (dismissed):

pkg/typesystem/connected_types.go
willvedd: Good call, `terminalRelation` is not needed. Removed.

adriantam (approved):

Issue #1766, opened 2024-07-12
"Internal Server Error" on Check and ListUsers


miparnisari:
### Checklist

- [X] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [X] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

Both of the requests below throw a 500: `folder1#viewer relation is undefined`

<img width="1166" alt="image" src="https://github.com/user-attachments/assets/868bea6d-df0d-468b-ae48-4f085e9e4469">

### Expectation

Check should return allowed=false and ListUsers should return an empty array

### Reproduction

See below

### Store data

```
name: Test

model: |+
    model
      schema 1.1

    type user
    type folder1
    type folder2
      relations
        define viewer: [user]
    type doc
      relations
        define viewer: viewer from parent
        define parent: [folder1, folder2]

tuples:
  - user: folder1:x
    relation: parent
    object: doc:x
tests:
  - name: internal-server-error
    check:
      - user: user:maria
        object: doc:x
        assertions:
          viewer: false
    list_users:
      - object: doc:x
        user_filter:
          - type: user
        assertions:
          viewer:
```

### OpenFGA version

Commit 518606928ded775a86a6ed69c4e6faa245e9b372

### How are you running OpenFGA?

As a binary

### What datastore are you using?

Postgres

### OpenFGA Flags

no

PR #1767, opened 2024-07-12
fix(check,listusers): if TTU relation is undefined, don't error

miparnisari:
## Description
Close https://github.com/openfga/openfga/issues/1766

### DIFF
--- a/assets/tests/consolidated_1_1_tests.yaml
+++ b/assets/tests/consolidated_1_1_tests.yaml
@@ -4269,46 +4269,37 @@ tests:
           model
             schema 1.1
           type user
-          type folder
+          type folder1
+          type folder2
             relations
-              define owner: [user]
-              define writer: [user,user:*] or owner
+              define viewer: [user]
           type document
             relations
-              define can_read: writer from parent
-              define parent: [document,folder]
-              define viewer: [user,user:*]
+              define viewer: viewer from parent
+              define parent: [folder1,folder2]
         tuples:
-          - user: user:anne
-            relation: owner
-            object: folder:a
-          - user: folder:a
+          - user: folder1:x
             relation: parent
-            object: document:c
-          - user: document:c
-            relation: parent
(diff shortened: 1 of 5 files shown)


assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -4269,46 +4269,37 @@ tests:
...
+            object: document:y
         checkAssertions:
           - tuple:
               user: user:anne
-              relation: can_read
-              object: document:c
-            expectation: true
+              relation: viewer
+              object: document:y
+            expectation: false
```
miparnisari, line 4289: I modified this test instead of creating a new one because: the old test provides little value. If there is a truthy path available, we may not exercise the buggy path under test.

The desired behavior is _even if there are no truthy paths available_ we should _not_ error.

pkg/typesystem/typesystem_test.go
```diff
@@ -2819,8 +2840,12 @@ func TestIsPubliclyAssignable(t *testing.T) {
 			typesys := New(model)
 
 			ok, err := typesys.IsPubliclyAssignable(test.target, test.objectType)
-			require.NoError(t, err)
-			require.Equal(t, ok, test.result)
+			if test.expectedError != "" {
+				require.ErrorContains(t, err, test.expectedError)
```
adriantam, line 2844: Since we don't check for error but instead check whether return is false in `hasPubliclyAssignedType, _ := typesys.IsPubliclyAssignable(relationReference, userType)`, should we add 
```suggestion
				require.False(ok)
				require.ErrorContains(t, err, test.expectedError)
```

assets/tests/consolidated_1_1_tests.yaml
adriantam: Should we add some assertions to verify the truth case (i.e., folder2 path?)

adriantam (approved):

PR #1745, opened 2024-07-08
FGA Cache Strategy first approach

smeadows-okta:
Captures the number of hits on an edge of the authorization model graph. 
## Description

Add a new resolver in the resolvers chain. This resolver is responsible for capturing nodes of the current execution path in the list of nodes of the execution path. Once we have a response from one of the execution paths, each node of the  is stored in the execution path hit cache (new in-memory cache). The data will be stored in memory in a map of a map with a Key:  store-id plus authorizationModelId and Value: Map2 with Key: usertype#relation#objectType and Value: #Hits. The cache will be flushed every 30 seconds or 1 min
```
2024-07-12T09:08:50.876-0400	INFO	execution path hits	{"store": "01J2KGVX4PG1AA9GWHDS21RBX3", "model": "01J2KGW5ZKY4M9TY1DY33HDPY3", "path": "folder:test-folder#writer@user", "hits": 3}
2024-07-12T09:08:50.876-0400	INFO	execution path hits	{"store": "01J2KGVX4PG1AA9GWHDS21RBX3", "model": "01J2KGW5ZKY4M9TY1DY33HDPY3", "path": "folder:test-folder#owner@user", "hits": 2}
2024-07-12T09:08:50.876-0400	INFO	execution path hits	{"store": "01J2KGVX4PG1AA9GWHDS21RBX3", "model": "01J2KGW5ZKY4M9TY1DY33HDPY3", "path": "folder:test-folder#viewer@user", "hits": 3}
2024-07-12T09:09:50.876-0400	INFO	execution path hits	{"store": "01J2KGWVHTREDXQ8GZ8NN4T6CF", "model": "01J2KGX90NEX01M0Q2BXAXFQ6G", "path": "folder:test-folder#writer@user", "hits": 6}
2024-07-12T09:09:50.881-0400	INFO	execution path hits	{"store": "01J2KGWVHTREDXQ8GZ8NN4T6CF", "model": "01J2KGX90NEX01M0Q2BXAXFQ6G", "path": "folder:test-folder#owner@user", "hits": 4}
2024-07-12T09:09:50.881-0400	INFO	execution path hits	{"store": "01J2KGWVHTREDXQ8GZ8NN4T6CF", "model": "01J2KGX90NEX01M0Q2BXAXFQ6G", "path": "folder:test-folder#viewer@user", "hits": 5}
```
## References

[...]

PR #1785, opened 2024-07-23
feat: log the number of hits on an edge of the authorization model graph

jpadilla:
Captures the number of hits on an edge of the authorization model graph. 
## Description

Add a new resolver in the resolvers chain. This resolver is responsible for capturing nodes of the current execution path in the list of nodes of the execution path. Once we have a response from one of the execution paths, each node of the  is stored in the execution path hit cache (new in-memory cache). The data will be stored in memory in a map of a map with a Key:  store-id plus authorizationModelId and Value: Map2 with Key: usertype#relation#objectType and Value: #Hits. The cache will be flushed every 30 seconds or 1 min
```
2024-07-12T09:08:50.876-0400	INFO	execution path hits	{"store": "01J2KGVX4PG1AA9GWHDS21RBX3", "model": "01J2KGW5ZKY4M9TY1DY33HDPY3", "path": "folder:test-folder#writer@user", "hits": 3}
2024-07-12T09:08:50.876-0400	INFO	execution path hits	{"store": "01J2KGVX4PG1AA9GWHDS21RBX3", "model": "01J2KGW5ZKY4M9TY1DY33HDPY3", "path": "folder:test-folder#owner@user", "hits": 2}
2024-07-12T09:08:50.876-0400	INFO	execution path hits	{"store": "01J2KGVX4PG1AA9GWHDS21RBX3", "model": "01J2KGW5ZKY4M9TY1DY33HDPY3", "path": "folder:test-folder#viewer@user", "hits": 3}
2024-07-12T09:09:50.876-0400	INFO	execution path hits	{"store": "01J2KGWVHTREDXQ8GZ8NN4T6CF", "model": "01J2KGX90NEX01M0Q2BXAXFQ6G", "path": "folder:test-folder#writer@user", "hits": 6}
2024-07-12T09:09:50.881-0400	INFO	execution path hits	{"store": "01J2KGWVHTREDXQ8GZ8NN4T6CF", "model": "01J2KGX90NEX01M0Q2BXAXFQ6G", "path": "folder:test-folder#owner@user", "hits": 4}
2024-07-12T09:09:50.881-0400	INFO	execution path hits	{"store": "01J2KGWVHTREDXQ8GZ8NN4T6CF", "model": "01J2KGX90NEX01M0Q2BXAXFQ6G", "path": "folder:test-folder#viewer@user", "hits": 5}
```
## References

https://github.com/openfga/openfga/pull/1745


internal/server/config/config.go
```diff
@@ -286,6 +286,8 @@ type Config struct {
 
 	RequestDurationDatastoreQueryCountBuckets []string
 	RequestDurationDispatchCountBuckets       []string
+
+	CheckTrackerEnabled bool
```
miparnisari, line 290: Please update `DefaultConfig()` to include this explicitly as default=false and then add a unit test.

internal/graph/track_check_resolver_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+		rn := val.(*resolutionNode)
+		rn.tm = time.Now().Add(-trackerInterval)
+		paths.Store(path, rn)
+
+		oldBurst := trackChecker.limiter.Burst()
+		trackChecker.limiter.SetBurst(0)
+
+		trackChecker.logExecutionPaths(false)
+		trackChecker.limiter.SetBurst(oldBurst)
```
miparnisari, line 251: Nit: why bother setting the burst back to its old value? You can create a new trackChecker per test

internal/graph/track_check_resolver_test.go
```diff
@@ -0,0 +1,280 @@
...
+		)
+
+		trackChecker.nodes.Store(trackerKey{store: ulid.Make().String(), model: ulid.Make().String()}, sm)
+		trackChecker.logExecutionPaths(false)
+
+		_, ok = sm.Load(path)
+		require.False(t, ok)
+	})
+
+	t.Run("tracker_success_launch_flush", func(t *testing.T) {
```
miparnisari, line 171: I have too many questions about this test 😅 Starting with the name. If you called `launchFlush` and didn't get any errors back, that should be enough. But since `launchFlush` doesn't return an error, then you need to observe its effects, right? So `launchFlush` writes to a log. But that only happens at certain time intervals. So why do you create a goroutine that sleeps for 10 ms and then wait on it? This smells like a test that will be flaky. Why not make the time interval zero (maybe an optional parameter to the constructor) so that as soon as there is an event, it flushes it, and you don't have to wait on it?

internal/graph/track_check_resolver_test.go
```diff
@@ -0,0 +1,280 @@
...
+		_, ok = paths.Load(path)
+		require.True(t, ok)
+
+		paths, ok = value.(*sync.Map)
+		require.True(t, ok)
+		_, ok = paths.Load(path)
+		require.True(t, ok)
+	})
+
+	t.Run("success_limiter_disallow", func(t *testing.T) {
```
miparnisari, line 227: I'm not sure how the code for this test achieves the goal in the name of this test, should you not assert that zero logs were emmited?

miparnisari (changes requested):

internal/server/config/config.go
miparnisari: Also, this tracker won't do anything if logging if disabled - please add validation that if tracker = true but logger = false, we error.

miparnisari:
I still have some things that i'd like addressed in follow-up PRs, but i'm approving for now to unblock development.

miparnisari (approved):

PR #1786, opened 2024-07-23
fix: computed userset should not be a complete dispatch

elbuo8:
## Description

Revert computed userset to its original behavior, so it doesn't consume from the resolution depth quota, doesn't trigger additional goroutines, doesn't get cached (since the cached entry is a throw away/proxy/virtual).

## Review Checklist


assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -7707,52 +7787,130 @@ tests:
           model
             schema 1.1
           type user
-
-          type document
+          type resource
```
elbuo8, line 7790: Can we remove this test altogether? Seems like a duplicate of `resolution_too_complex_throws_error`

miparnisari: `err_and_err_return_err` and `list_objects_with_subcheck_resolution_depth_exceeded` look identical to me.

image

 Let's keep `err_and_err_return_err` 

But let's keep `resolution_too_complex_throws_error`. This one's solution is one path that leads to an error, whereas `err_and_err_return_err`  has two paths that both lead to an error

internal/graph/check.go
```diff
@@ -974,22 +974,22 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
+		req.GetTupleKey().GetObject(),
+		rewrite.ComputedUserset.GetRelation(),
+		req.GetTupleKey().GetUser(),
+	)
 
-		return c.dispatch(ctx, req, rewrittenTupleKey)(ctx)
+	childRequest.TupleKey = rewrittenTupleKey
+
+	return func(ctx context.Context) (*ResolveCheckResponse, error) {
+		return c.ResolveCheck(ctx, childRequest)
```
miparnisari, line 992: Why `c.ResolveCheck(ctx, childRequest)` and not `c.delegate.ResolveCheck(ctx, childRequest)`? If you run with `--check-cache-enabled` this means that we will not cache the result of a computed userset check. Is that what you wanted too?

assets/tests/consolidated_1_1_tests.yaml
elbuo8: To be clear, remove: `list_objects_with_subcheck_resolution_depth_exceeded` and keep: `resolution_too_complex_throws_error`, and `err_and_err_return_err`, yes?

internal/graph/check.go
miparnisari: Here, this test fails in this PR, it never exits


```go
func TestDetectsCycle(t *testing.T) {
	ds := memory.New()
	t.Cleanup(ds.Close)

	model := testutils.MustTransformDSLToProtoWithID(`
			model
				schema 1.1
			type user
			type document
				relations
					define x: x
					define y: x`)

	ctx := typesystem.ContextWithTypesystem(
		context.Background(),
		typesystem.New(model), // not NewAndValidate because this model isn't valid
	)

	ctx = storage.ContextWithRelationshipTupleReader(ctx, ds)

	checker := NewLocalChecker()
	t.Cleanup(checker.Close)

	resp, err := checker.ResolveCheck(ctx, &ResolveCheckRequest{
		StoreID:              storeID,
		AuthorizationModelID: model.GetId(),
		TupleKey:             tuple.NewTupleKey("document:1", "x", "user:maria"),
		RequestMetadata:      NewCheckRequestMetadata(20),
	})
	require.NoError(t, err)
	require.Nil(t, resp)
	require.True(t, resp.GetCycleDetected())
}
```

assets/tests/consolidated_1_1_tests.yaml
miparnisari: Yes

internal/graph/check.go
elbuo8: Seems we are missing this test then.

elbuo8: Added the test.

internal/graph/check.go
```diff
@@ -974,23 +974,68 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
...
-			rewrite.ComputedUserset.GetRelation(),
-			req.GetTupleKey().GetUser(),
+	for {
+		tk = tuple.NewTupleKey(
+			tk.GetObject(),
+			rewrite.GetComputedUserset().GetRelation(),
+			tk.GetUser(),
 		)
+		key := tuple.TupleKeyToString(tk)
+		if _, cycleDetected := seen[key]; cycleDetected {
```
miparnisari, line 997: i'm OK with duplicating cycle detection code here, but what do you think of sending a follow-up PR where we merge the cycle resolver in this local resolver and get rid of this?

miparnisari (approved):

yissellokta (approved):

internal/graph/check.go
miparnisari: Track via https://github.com/openfga/openfga/issues/1794

PR #1788, opened 2024-07-24
refactor: extract interface for in-memory cache

miparnisari:
## Description

This PR has no functional changes.

I noticed in https://github.com/openfga/openfga/pull/1764 that we use the `ccache` 3rd party library in various places, which will make it harder to replace if we ever decide to. 

- I am introducing an interface to wrap the library.
- Refactored some tests in `internal/graph/cached_resolver_test.go`.


internal/graph/cached_resolver.go
```diff
@@ -109,9 +110,10 @@ func NewCachedCheckResolver(opts ...CachedCheckResolverOpt) *CachedCheckResolver
 
 	if checker.cache == nil {
 		checker.allocatedCache = true
-		checker.cache = ccache.New(
-			ccache.Configure[*ResolveCheckResponse]().MaxSize(checker.maxCacheSize),
-		)
+		var optsCache []storage.InMemoryLRUCacheOpt[*ResolveCheckResponse]
+		optsCache = append(optsCache, storage.WithMaxCacheSize[*ResolveCheckResponse](checker.maxCacheSize))
+		optsCache = append(optsCache, storage.WithDefaultTTL[*ResolveCheckResponse](checker.cacheTTL))
+		checker.cache = storage.NewInMemoryLRUCache[*ResolveCheckResponse](optsCache...)
```
senojj, line 116: The `append` function performs various checks that are not necessary in a scenario where we know the upper-bound and items of the slice. It is generally more performant to use direct initialization, or initialize the slice with a length (using `make`) equal to the upper-bound and use direct index access to assign values.
```suggestion
		optsCache := []storage.InMemoryLRUCacheOpt[*ResolveCheckResponse]{
		    storage.WithMaxCacheSize[*ResolveCheckResponse](checker.maxCacheSize),
		    storage.WithDefaultTTL[*ResolveCheckResponse](checker.cacheTTL)
		}
		checker.cache = storage.NewInMemoryLRUCache[*ResolveCheckResponse](optsCache...)
```

pkg/storage/cache.go
```diff
@@ -0,0 +1,84 @@
...
+
+const defaultMaxCacheSize = 10000
+const defaultCacheTTL = 10 * time.Second
+
+// InMemoryCache is a general purpose cache to store things in memory.
+type InMemoryCache[T any] interface {
+
+	// Get If the key exists, returns the value and "true".
+	// If the key didn't exist, returns nil, false.
+	Get(key string) (*CachedResult[T], bool)
```
senojj, line 19: (non-blocking/ergonomic) Is it possible for this function to ever return a non-nil value and `false`? If not, simply returning `nil` could qualify as sufficient information to infer that the key did not exist in the cache.

miparnisari: I tried to replicate the syntax used by the standard map type: https://go.dev/blog/maps

> A two-value assignment tests for the existence of a key:
> 
> i, ok := m["route"]
> 
> In this statement, the first value (i) is assigned the value stored under the key "route". If that key doesn’t exist, i is the value type’s zero value (0). The second value (ok) is a bool that is true if the key exists in the map, and false if not.
> 
> To test for a key without retrieving the value, use an underscore in place of the first value:
> 
> _, ok := m["route"]
>

miparnisari: I'll apply the suggestion :)

senojj (approved):
this looks good, but there seems to be a flakey test.

miparnisari:
@senojj yes 😢 i have not figured out the root cause yet. Somehow, when calling the Streaming ListObjects API, the logs are emmitted after the call is complete. 🤔

elbuo8 (approved):

PR #1804, opened 2024-07-29
perf(Check): stream userset results

miparnisari:
## Description

With this model:

```
model schema 1.1
type user
type folder
    relations
        define viewer: [user]
type doc
	relations
        define parent: [folder]
```

And this Check: `user:maria, viewer, doc:1`

Before this PR:

1. Fetch all N `parent` folders of `doc:1`
2. Fetch all `folder`s that `user:maria` can view 
3. Compute `union[(user:maria,viewer,parent1),...,(user:maria,viewer,parentN)]`, i.e. the intersection between (1) and (2)

With this PR:

1. fetch all N `parent` folders of `doc:1` in batches of 1000
2. For every new batch: Fetch all `folder`s that `user:maria` can view and compute `Check(user:maria,viewer,parentX)`.
3. If any of those returns `allowed=true`, exit early, else continue fetching batches.


internal/graph/check.go
```diff
@@ -790,6 +792,28 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+
+// buildUsersetDetails given tuple doc:1#viewer@group:2#member will return group#member, 2, nil.
+// This util takes into account pre-computed relationships, otherwise it will resolve it from the target UserType.
+func buildUsersetDetails(typesys *typesystem.TypeSystem, computedRelation, userType string) usersetDetailsFunc {
+	return func(t *openfgav1.TupleKey) (string, string, error) {
+		cr := computedRelation
+		object, relation := tuple.SplitObjectRelation(t.GetUser())
+		objectType, objectID := tuple.SplitObject(object)
+		if cr == "" {
+			terminalRelations := typesys.GetTerminalRelations(objectType, relation, userType)
```
adriantam, line 805: Looks like we have seen the following pattern a few times.  Maybe we should create a helper method in typesys so that the error message is standardized
```
x := typesys.GetTerminalRelations(a, b, c)
if len(x) != 1 {
  return fmt.Errorf("Foo")
}
y := x[0]
```

internal/graph/check.go
```diff
@@ -805,64 +829,151 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
-		// Consider doing it when we have better concurrency tooling for "streaming" operations.
-		t, err := iter.Next(ctx)
-		if err != nil {
-			if errors.Is(err, storage.ErrIteratorDone) {
-				break
+	cancellableCtx, cancelFunc := context.WithCancel(ctx)
+	// sending to channel in batches up to a pre-configured value of usersets to subsequently checkMembership for.
+	pool := concurrency.NewPool(cancellableCtx, 1)
+	pool.Go(func(ctx context.Context) error {
+		defer close(usersetsChan)
```
adriantam, line 864: since usersetsMap is not thread safe and we only use it inside this function, will it be better to move the variable declaration here?

internal/graph/check.go
```diff
@@ -805,64 +829,151 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
-		// usersetMap by draining the full iterator, it can start processing in batches of ObjectRelation.
-		// Consider doing it when we have better concurrency tooling for "streaming" operations.
-		t, err := iter.Next(ctx)
-		if err != nil {
-			if errors.Is(err, storage.ErrIteratorDone) {
-				break
+	cancellableCtx, cancelFunc := context.WithCancel(ctx)
+	// sending to channel in batches up to a pre-configured value of usersets to subsequently checkMembership for.
+	pool := concurrency.NewPool(cancellableCtx, 1)
+	pool.Go(func(ctx context.Context) error {
```
adriantam, line 863: Will it make it easier to read / maintain the code if we have all these in its own explicit function?

internal/graph/check.go
```diff
@@ -805,64 +829,151 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+			}
+
+			objectRel, objectID, err := usersetDetails(t)
+			if err != nil {
+				usersetsChan <- usersets{err: err}
+			}
+
+			if _, ok := usersetsMap[objectRel]; !ok {
+				if len(usersetsMap) >= 1 {
+					// flush previous results
```
adriantam, line 888: Can you explain why do we need to flush previous results?

internal/graph/check.go
```diff
@@ -805,64 +829,151 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+						}:
+							delete(usersetsMap, k)
+						}
+					}
+				}
+				usersetsMap[objectRel] = storage.NewSortedSet()
+			}
+
+			usersetsMap[objectRel].Add(objectID)
+			if usersetsMap[objectRel].Size() > 1000 {
```
adriantam, line 905: 1000 is a magical number.  Please const it or make it configurable.  (Making it configurable allows better unit testing as none of the code below are covered by unit test).

internal/graph/check.go
```diff
@@ -805,64 +829,151 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+			if newBatch.err != nil {
+				finalErr = newBatch.err
+				break ConsumerLoop
+			}
+			objectRel := newBatch.objectRelation
+			objectIDs := newBatch.objectIDs
+
+			resp, err := c.buildCheckAssociatedObjects(req, objectRel, objectIDs)(ctx)
+			if err != nil {
+				finalErr = err
```
adriantam, line 958: can you explain why do we break out of loop if newBatch.err != nil above but not breaking out of the loop here?

internal/graph/check.go
miparnisari: > none of the code below are covered by unit test

good catch, yikes

internal/graph/check.go
```diff
@@ -805,64 +838,152 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+	// Caller already verified typesys
 	typesys, _ := typesystem.TypesystemFromContext(ctx)
-	// We had just checked for the existence of the typesys in the caller.
-	// So, we are guaranteed for the presence of the context.
+	usersetDetails := buildUsersetDetails(typesys, "", tuple.GetType(req.GetTupleKey().GetUser()))
+	return c.checkMembership(ctx, req, iter, usersetDetails)
+}
 
-	reqUserType := tuple.GetType(req.GetTupleKey().GetUser())
+type usersets struct {
```
adriantam, line 850: can we rename this as `usersetsChanType` (or something like that) as usersets has various meaning

internal/server/config/config.go
```diff
@@ -23,6 +23,7 @@ const (
 	DefaultChangelogHorizonOffset           = 0
 	DefaultResolveNodeLimit                 = 25
 	DefaultResolveNodeBreadthLimit          = 100
+	DefaultUsersetBatchSize                 = 1000
```
willvedd, line 26: **Question:** Does this need to be a configured value at the server level? Seems like this would be difficult to advise people on how to configure and further, a sensible default could go a long way. 

However, I could see this being configured on a per-datastore basis.

willvedd (commented):
Honestly don't really follow how this specific mechanism works. In particular what the act of "flushing" is. But I do have a question and an observation.

internal/graph/check.go
miparnisari: Added more tests around context cancellation

internal/server/config/config.go
miparnisari: Yeah i left it out on purpose from it being an actual server flag because (a) I was lazy to do it here, and (b) I don't know if we want to expose that flag to users yet... at the very least it's going to be very hard to explain what values to use because we don't even know ourselves yet.

internal/graph/check.go
miparnisari: I kept the body of this function here but I created sub-functions for some things, let me know if this works

adriantam: Conceptually, it is easier to understand if we can encapsulate what we try to do inside the Pool as its own function (which can give us the expected input/output)

internal/graph/check.go
```diff
@@ -175,6 +179,8 @@ type LocalChecker struct {
 	delegate           CheckResolver
 	concurrencyLimit   uint32
 	maxConcurrentReads uint32
+	usersetBatchSize   uint32
+	logger             logger.Logger
```
elbuo8, line 183: This is not being used 🤔

internal/graph/check.go
```diff
@@ -805,64 +848,181 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
-	resp, err := union(ctx, c.concurrencyLimit, handlers...)
-	if err != nil {
-		telemetry.TraceError(span, err)
-		return nil, err
-	}
+	trySendUsersetsAndDeleteFromMap(ctx, usersetsMap, usersetsChan)
+}
 
-	resp.GetResolutionMetadata().DatastoreQueryCount++
+func trySendError(ctx context.Context, err error, errorChan chan usersetsChannelType) {
```
elbuo8, line 1007:
```suggestion
func trySendUsersetsError(ctx context.Context, err error, errorChan chan usersetsChannelType) {
```

internal/graph/check_test.go
```diff
@@ -1940,6 +1942,85 @@ func TestUnionCheckFuncReducer(t *testing.T) {
...
+	// add some folders as parents of the document
+	maxFolderID := int(usersetBatchSize * 10)
+	for i := 0; i <= maxFolderID; i++ {
+		err := ds.Write(context.Background(), storeID, nil, []*openfgav1.TupleKey{
+			tuple.NewTupleKey("doc:1", "parent", fmt.Sprintf("folder:%d", i)),
+		})
+		require.NoError(t, err)
+	}
+
+	// the user is viewer of the last folder so we force early flushing of the batches when computing membership
```
elbuo8, line 1974: What about the case when the user is viewer of the first one? That should prevent reading the whole iterator and short circuit properly. Forcing it to go through here: https://github.com/openfga/openfga/pull/1804/files#diff-5074b7eeccca5be19ff0fa2a8f712335a6ed075628b66a46697ca9d8ca99fbe0R993

internal/graph/check.go
miparnisari: Oh right I used this add debug logs but then I removed them. I still think it's a good idea to have the logger available though..

elbuo8: If not going to be used, remove it.

internal/graph/check.go
```diff
@@ -805,64 +848,181 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+
+	if finalErr != nil {
+		telemetry.TraceError(span, finalErr)
+		return nil, finalErr
+	}
+
+	return &ResolveCheckResponse{
+		Allowed: false,
+		ResolutionMetadata: &ResolveCheckResponseMetadata{
+			DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount,
```
elbuo8, line 955:
```suggestion
			DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount++,
```

The reason being, we still did queries, just doesn't mean that we found anything.

elbuo8: If anything we should increase the counter per each `objectRel` that we observed or per each batch we sent over the channel.

internal/graph/check.go
miparnisari: But then next time i'm debugging something in Check I will need it again, i'd rather leave it here, what is the cost?

internal/graph/check_test.go
miparnisari: In my test, the producer always finishes before the consumer finds an answer 🤔

internal/graph/check.go
miparnisari: In hindsight, passing this counter around on every request and response was a very bad idea :(

internal/graph/check.go
```diff
@@ -805,64 +848,181 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMapType, usersetsChan chan usersetsChannelType) {
+	for k, v := range usersetsMap {
+		select {
+		case <-ctx.Done():
+			return
+		case usersetsChan <- usersetsChannelType{
+			objectRelation: k,
+			objectIDs:      v,
+		}:
+			delete(usersetsMap, k)
```
adriantam, line 1023: curious why do we need to delete individual item vs. using a new usersetMap

internal/graph/check.go
```diff
@@ -805,64 +848,181 @@ func buildTupleKeyConditionFilter(ctx context.Context, reqCtx *structpb.Struct,
...
+	var finalErr error
+
+ConsumerLoop:
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case newBatch, channelOpen := <-usersetsChan:
+			if !channelOpen {
+				break ConsumerLoop
```
adriantam, line 920: instead of breaking from ConsumerLoop, would it be better to return the result (conceptually easier to understand)

internal/graph/check.go
elbuo8: I cleaned it up a bit without seeing this comment. But we clear it in order to keep track of `seen` objectRelationships and flushing when we observe a new one (based on the assumption that they will come sorted).

internal/graph/check.go
elbuo8: Can't return at this point since error handling etc happens after.

adriantam (dismissed):

adriantam (approved):

Issue #1794, opened 2024-07-25
Merge CycleDetectionChecker with LocalChecker


miparnisari:
i'm OK with duplicating cycle detection code here, but what do you think of sending a follow-up PR where we merge the cycle resolver in this local resolver and get rid of this?

_Originally posted by @miparnisari in https://github.com/openfga/openfga/pull/1786#discussion_r1690182749_

PR #1805, opened 2024-07-30
refactor: remove CycleDetectionCheckResolver in favor of inside Local…

miparnisari:
…Checker

## Description

I am cleaning up tech debt introduced in https://github.com/openfga/openfga/pull/1786: getting rid of the duplicate code that detects cycles:
- https://github.com/openfga/openfga/blob/0f960edd0b9455340a835241396c504c5d96871d/internal/graph/cycle_check_resolver.go#L41
- https://github.com/openfga/openfga/blob/0f960edd0b9455340a835241396c504c5d96871d/internal/graph/check.go#L1032

## References

Closes https://github.com/openfga/openfga/issues/1794


internal/graph/check.go
```diff
@@ -196,8 +196,7 @@ func WithMaxConcurrentReads(limit uint32) LocalCheckerOption {
 // NewLocalChecker constructs a LocalChecker that can be used to evaluate a Check
 // request locally.
 //
-// The constructed LocalChecker is not wrapped with cycle detection. Developers
-// wanting a LocalChecker without other wrapped layers (e.g caching and others)
+// Developers wanting a LocalChecker with other optional layers (e.g caching and others)
```
miparnisari, line 199: The "optional" word is key here. It would have been easy to do `NewLocalChecker` elsewhere in the codebase (e.g. for ListObjects, or in a test) and forget that it doesn't include cycle detection.

internal/graph/check.go
```diff
@@ -596,6 +601,11 @@ func (c *LocalChecker) ResolveCheck(
 		}, nil
 	}
 
+	typesys, ok := typesystem.TypesystemFromContext(ctx)
+	if !ok {
+		panic("typesystem missing in context")
```
elbuo8, line 606: Any reason why we would use a panic here? And can we get a small test for this since this is the entry point?

internal/graph/check.go
```diff
@@ -1013,64 +1039,19 @@ func (c *LocalChecker) checkComputedUserset(ctx context.Context, req *ResolveChe
...
-			break
-		}
-	}
-
 	childRequest := clone(req)
-	childRequest.TupleKey = tk
+	childRequest.TupleKey = rewrittenTupleKey
 
-	return c.checkRewrite(ctx, childRequest, rewrite)
+	return func(ctx context.Context) (*ResolveCheckResponse, error) {
```
elbuo8, line 1051: 🤔 Why are we reverting this?

internal/graph/check.go
miparnisari: No reason. I dislike this panic as much as you do, but let's tackle that in https://github.com/openfga/openfga/issues/616 :)

internal/graph/check.go
miparnisari: Because `checkRewrite` doesn't check for cycles, `ResolveCheck` does, and `ResolveCheck` eventually calls `checkRewrite`

internal/graph/check.go
elbuo8: Please add the test 😅 Cuz #616 seems like a large endeavor that might take a while to land.

elbuo8 (changes requested):

internal/graph/check.go
```diff
@@ -611,6 +621,22 @@ func (c *LocalChecker) ResolveCheck(
 	return resp, nil
 }
 
+// hasCycle returns true if a cycle has been found. It modifies the request object.
+func (c *LocalChecker) hasCycle(req *ResolveCheckRequest) bool {
```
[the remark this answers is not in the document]
miparnisari: But i need to modify the `req` object to mark the tuple as visited

internal/graph/check.go
miparnisari: https://github.com/openfga/openfga/pull/1809

internal/graph/check.go
miparnisari: I like having the call to `hasCycle` in one place only :)

elbuo8 (approved):

Issue #1707, opened 2024-06-24
Allow the caller to know how long a query took to resolve


rhamzeh:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

As users are measuring the performance of their OpenFGA nodes and how best to configure it, as well as understanding when their model or calls are too complex and causing high latency, they'd like a way to measure how long it took to resolve a request from the time the server node received the request till it returns the response. It would be nice if this was exposed to the client without them having to cross check with the `took` field in the logs.

### Describe the ideal solution

Return `fga-query-duration-ms` header with how long a query took to resolve internally

In code:
* Add a `duration` middleware that adds a `fga-query-duration-ms` field to the `context` tags, 
* In the existing places where we send headers, replace it with a `context.Add(header, value)`.
* Add a `headers` middleware that sends *all* response headers.

In testing:
* Update `TestCheckLogs` and assert that logging middleware is logging `fga-query-duration-ms`, both in success and error cases
* Update `TestHTTPHeaders` and assert that we are returning header `fga-query-duration-ms`, both in success and error cases

### Alternatives and current workarounds

Users have to cross check their requests with the server logs to look at the took field.

### Additional context

Supersedes #1366 

The OpenFGA SDKs are currently looking at: `fga-query-duration-ms` to expose as a metric from their end.
https://github.com/openfga/dotnet-sdk/blob/v0.5.1/src/OpenFga.Sdk/Telemetry/Histograms.cs#L48-L56

lalalalatt:
@rhamzeh I made a PR https://github.com/openfga/openfga/pull/1807 for this issue. 
Since the duration logic is quite duplicated, Would it be better to use gRPC middleware instead?

rhamzeh:
@lalalalatt - yep that's totally fine!

PR #1807, opened 2024-07-30
feat(server): log query duration

lalalalatt:
## Description
Implement https://github.com/openfga/openfga/issues/1707

## References

- https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/server.go#L22-L25
- https://github.com/grpc-ecosystem/go-grpc-middleware/blob/main/interceptors/server.go#L35-L36

## Review Checklist


pkg/server/server.go
```diff
@@ -568,13 +570,13 @@ func (s *Server) Close() {
 }
 
 func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequest) (*openfgav1.ListObjectsResponse, error) {
+	start := time.Now()
```
miparnisari, line 573: Let's put this in a middleware instead 😄 

In https://github.com/openfga/openfga/blob/57dd34cd538b487ef162694c5d36384cf2fb6f38/pkg/middleware/logging/logging.go#L58-L59 you can add this:

```go
func (r *reporter) PostCall(err error, _ time.Duration) {
	duration := time.Since(r.start)
	r.transport.SetHeader(r.ctx, "fga-query-duration-ms", fmt.Sprintf("%d", duration.Milliseconds()))
	r.fields = append(r.fields, zap.Duration("took", duration))
```

which will both send a header and write to the log.

You'll have to edit the constructors of this middleware to pass a `transport` field that you can get from here: https://github.com/openfga/openfga/blob/0f960edd0b9455340a835241396c504c5d96871d/cmd/run/run.go#L587

Also you'll have to update these tests: `TestHTTPHeaders` and `TestServerLogs`.

miparnisari (changes requested):

pkg/middleware/logging/logging.go
```diff
@@ -143,6 +154,8 @@ func reportable(l logger.Logger) interceptors.CommonReportableFunc {
 
 		return &reporter{
 			ctx:            ctxzap.ToContext(ctx, zapLogger.Logger),
+			grpcCtx:        ctx,
```
miparnisari, line 157: Why do you need a new `ctx` object? The one above it works

pkg/middleware/logging/logging.go
```diff
@@ -49,13 +53,20 @@ func NewStreamingLoggingInterceptor(logger logger.Logger) grpc.StreamServerInter
...
+	grpcCtx        context.Context
+	start          time.Time
 	logger         logger.Logger
 	fields         []zap.Field
 	protomarshaler protojson.MarshalOptions
 }
 
 // PostCall is invoked after all PostMsgSend operations.
 func (r *reporter) PostCall(err error, _ time.Duration) {
+	duration := time.Since(r.start)
```
miparnisari, line 65: I just realized that we don't need to record `r.start` to calculate the duration - the duration is the second parameter to this function 😉 https://github.com/grpc-ecosystem/go-grpc-middleware/blob/62b7de50cda5a5d633f1013bfbe50e0f38db34ef/interceptors/server.go#L21-L23

cmd/run/run_test.go
```diff
@@ -1459,6 +1460,9 @@ func TestHTTPHeaders(t *testing.T) {
 			require.Equal(t, storeID, httpResponse.Header[storeid.StoreIDHeader][0])
 			require.Len(t, httpResponse.Header[requestid.RequestIDHeader], 1)
 			require.NotEmpty(t, httpResponse.Header[requestid.RequestIDHeader][0])
+			t.Log(httpResponse.Header)
```
miparnisari, line 1463: Pease remove this log

miparnisari (changes requested):

pkg/middleware/logging/logging.go
miparnisari: This is unresolved

lalalalatt: Thank you for your patience.  😅

miparnisari (dismissed):

lalalalatt:
@miparnisari  the duration header test on `streamed-list-objects` fail sometime, I am seeking how to fix it. Do you have any idea on it? Maybe the stream rpc service not always intercept by the middleware(interceptor)

miparnisari:
> @miparnisari the duration header test on `streamed-list-objects` fail sometime, I am seeking how to fix it. Do you have any idea on it? Maybe the stream rpc service not always intercept by the middleware(interceptor)

Oh no 😢 I was very confident that we fixed this in https://github.com/openfga/openfga/pull/1795. @senojj do you mind taking a look?

senojj:
> Oh no 😢 I was very confident that we fixed this in #1795. @senojj do you mind taking a look?

This is actually a completely separate issue. 

The new test addition is looking for the presence of a response header that is written in the "post call" phase of the streaming interceptor. At that point in the response lifecycle, the response header may have already been written to the wire; it is a race between the server writing the response header to the wire and `grpc.SetHeader` getting executed in the interceptor reporter's `PostCall` method. This race only exists because of the streaming nature of the specific response. 

The response header is always written to the wire before the first write of body data. In streaming responses, the first write of body data is typically written immediately after the first unit of data is available. In our unary responses, the entire response dataset is compiled before writing the response to the wire. Therefore, in the case of a unary response, `PostCall` will execute before the response header is actually written to the wire.

If the error returned from the attempt to set the header in `PostCall` is inspected, you will find that it is often not `nil`, and contains the message: "rpc error: code = Internal desc = transport: SendHeader called multiple times".

That message indicates that the header had already been written to the wire prior to this call to `grpc.SetHeader`.

The grpc project [docs](https://pkg.go.dev/google.golang.org/grpc@v1.65.0#SetHeader) state the following:
> SetHeader sets the header metadata to be sent from the server to the client. The context provided must be the context passed to the server's handler.
>
>Streaming RPCs should prefer the SetHeader method of the ServerStream.
>
>When called multiple times, all the provided metadata will be merged. All the metadata will be sent out when one of the following happens:
>
> - grpc.SendHeader is called, or for streaming handlers, stream.SendHeader.
> - The first response message is sent. For unary handlers, this occurs when the handler returns; for streaming handlers, this can happen when stream's SendMsg method is called.
> - An RPC status is sent out (error or success). This occurs when the handler returns.
>
> SetHeader will fail if called after any of the events above.

miparnisari:
@lalalalatt do you mind undoing the `grpc.SetHeader` line? We'll have to change the codebase a bit before we are ready for this change. Also please update the name of this PR to `feat(server): log query duration`

lalalalatt:
> @lalalalatt do you mind undoing the `grpc.SetHeader` line? We'll have to change the codebase a bit before we are ready for this change. Also please update the name of this PR to `feat(server): log query duration`

@miparnisari  Ok.
And I can help with changing the codebase.

For the `"rpc error: code = Internal desc = transport: SendHeader called multiple times".` issue:
> - grpc.SendHeader is called, or for streaming handlers, stream.SendHeader.
> - The first response message is sent. For unary handlers, this occurs when the handler returns; for streaming handlers, this can happen when stream's SendMsg method is called.
> - An RPC status is sent out (error or success). This occurs when the handler returns.

It seems there aren't any `grpc.SendHeader` call in codebase, do you have any idea on where the other two happened?

senojj:
> It seems there aren't any `grpc.SendHeader` call in codebase, do you have any idea on where the other two happened?

The relevant piece is this:
>The first response message is sent. For unary handlers, this occurs when the handler returns; for streaming handlers, this can happen when stream's SendMsg method is called.

Specifically:
>for streaming handlers, this can happen when stream's SendMsg method is called.

The stream's `SendMsg` method is called [here](https://github.com/openfga/api/blob/c96ec773cae9caf2a9536856bf1fa206a42ef7ef/proto/openfga/v1/openfga_service_grpc.pb.go#L643).
The server's `Send` method is called [here](https://github.com/openfga/openfga/blob/07794019d080bf629a6408d9917ea1492e1eaba6/pkg/server/commands/list_objects.go#L505).

The first call to that `Send` method will cause the header to be written to the wire, preventing any possibility of adding more header fields later.

So when a call to `SetHeader` is made in the `PostCall` method of the interceptor's reporter, it fails, because a call to the stream's `SendMsg` method has already happened.

miparnisari (approved):

lalalalatt:
> > It seems there aren't any `grpc.SendHeader` call in codebase, do you have any idea on where the other two happened?
> 
> The relevant piece is this:
> 
> > The first response message is sent. For unary handlers, this occurs when the handler returns; for streaming handlers, this can happen when stream's SendMsg method is called.
> 
> Specifically:
> 
> > for streaming handlers, this can happen when stream's SendMsg method is called.
> 
> The stream's `SendMsg` method is called [here](https://github.com/openfga/api/blob/c96ec773cae9caf2a9536856bf1fa206a42ef7ef/proto/openfga/v1/openfga_service_grpc.pb.go#L643). The server's `Send` method is called [here](https://github.com/openfga/openfga/blob/07794019d080bf629a6408d9917ea1492e1eaba6/pkg/server/commands/list_objects.go#L505).
> 
> The first call to that `Send` method will cause the header to be written to the wire, preventing any possibility of adding more header fields later.
> 
> So when a call to `SetHeader` is made in the `PostCall` method of the interceptor's reporter, it fails, because a call to the stream's `SendMsg` method has already happened.

I posted a [question](https://groups.google.com/g/grpc-io/c/PCGpXH38VE4) on grpc forum

PR #1814, opened 2024-08-01
fix: connected types consolidate type definitions with and without a condition

willvedd:
## Description
Currently, when creating a typesystem from a model, we reduce certain relations' indirections into their terminal relations. This can be used to understand when we can apply "fast-path" optimization for TTU and userset subproblems. Initially, these optimizations only account for rewrites other than union, intersection and exclusion. By excluding these rewrites we can assert that there is always a single terminal relation for a certain combination of user type, object type and relation. 
However, we erroneously "double-count" under during two scenarios:
- multiple type restrictions of a single type – with and without a condition (example 1 below)
- multiple type restrictions of a single type – with multiple conditions (example 2 below)

This presents as a 500 error with a message of `"expected exactly one terminal relation for fast path, received 2"` when running a `Check` where a subproblem incurs one or more described scenarios above.

**Example 1:**
```
model
   schema 1.1

type user
type group
   relations
      define member: [user, user with x_less_than] # Note: `user` with and without condition

type folder
   relations
      define parent: [group]
      define viewer: member from parent

condition x_less_than(x: int) {
   x < 100
}
```
Calling `Check( "user:1", "viewer", "folder:1")`

**Example 2:**
```
model
   schema 1.1

type user
type group
   relations
      define member: [user with x_less_than, user with  x_greater_than] # Note: multiple conditions

type folder
   relations
      define parent: [group]
      define viewer: member from parent

condition x_less_than(x: int) {
   x < 100
}

condition x_greater_than(x: int) {
   x < 100
}
```
Calling `Check( "user:1", "viewer", "folder:1")`

This PR fixes this "double-count" by preventing duplicate terminal relations from being recorded in the typesystem.

## References
Introduced by #1765 

## Review Checklist


assets/tests/abac_tests.yaml
```diff
@@ -1092,6 +1092,184 @@ tests:
...
+              user: user:with-condition
+              relation: viewer
+              object: document:1
+            errorCode: 2000
+          - tuple:
+              user: user:with-condition
+              relation: viewer
+              object: document:1
+            context:
+              "x": 10
```
adriantam, line 1134: can we also add the case where x != 10

assets/tests/abac_tests.yaml
```diff
@@ -1092,6 +1092,184 @@ tests:
...
+            expectation: true
+          - tuple:
+              user: user:without-condition
+              relation: viewer
+              object: document:1
+            expectation: true
+          - tuple:
```
adriantam, line 1141: how is this different than the tuple above?

assets/tests/abac_tests.yaml
```diff
@@ -1092,6 +1092,184 @@ tests:
...
+              object: document:1
+            context:
+              "x": 10
+            expectation: true
+          - tuple:
+              user: user:with-ycond
+              relation: viewer
+              object: document:1
+            context:
+              "y": 10
```
adriantam, line 1230: Please also add the negative case

assets/tests/abac_tests.yaml
```diff
@@ -1092,6 +1092,184 @@ tests:
...
+              user: user:without-condition
+              relation: viewer
+              object: document:1
+            expectation: true
+          - tuple:
+              user: user:without-condition
+              relation: viewer
+              object: document:1
+            expectation: true
+        listUsersAssertions:
```
adriantam, line 1146: Do we have list objects assertion?

willvedd: Added for both integration tests.

assets/tests/abac_tests.yaml
willvedd: Removed.

assets/tests/abac_tests.yaml
willvedd: Added.

adriantam (approved):

elbuo8 (approved):

PR #1816, opened 2024-08-02
chore!: add Head for iterator

adriantam:
## Description
Iterator function Head to look at the first item without advancing iterator.

Note: custom datastore implementation will need to be updated to support Head()

This is needed in preparation for check latency optimization for TTU & userset where the child is an intersection/union/exclusion. 

## Review Checklist

### DIFF
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -23,8 +23,10 @@ type Iterator[T any] interface {
 	// the underlying iterator.
 	Stop()
 
-	// Head will return the first item or ErrIteratorDone if the list
+	// Head will return the first item or ErrIteratorDone if the iterator
 	// is empty.
+	// It's possible for this method to advance the iterator internally, but a subsequent call to Next will not miss any results.
+	// Calling Head() continuously without calling Next() will yield the same result (the first one) over and over.
 	Head(ctx context.Context) (T, error)
 }
 
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -343,7 +343,8 @@ func (f *ConditionsFilteredTupleKeyIterator) Stop() {
 
 // Head returns the next most tuple in the underlying iterator that meets
 // the filter function this iterator was constructed with.
-// The underlying iterator will not advance.
+// The underlying iterator may advance but calling consecutive Head will yield consistent result.
+// Further, calling Head following by Next will also yield consistent result.
 // This function is not thread-safe.
 func (f *ConditionsFilteredTupleKeyIterator) Head(ctx context.Context) (*openfgav1.TupleKey, error) {
 	for {
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -357,19 +357,18 @@ func (f *ConditionsFilteredTupleKeyIterator) Head(ctx context.Context) (*openfga
 		}
 
 		valid, err := f.filter(tuple)
-		if err != nil {
-			f.lastError = err
-			_, err = f.iter.Next(ctx)
+		if err != nil || !valid {
 			if err != nil {
-				// should never happen except if the underlying ds has error
-				return nil, err
+				f.lastError = err
 			}
-			continue
-		}
-		if !valid {
+			// Note that we don't care about the item returned by Next() as this is already via Head(). We call Next() solely
+			// for the purpose of getting rid of the first item.
 			_, err = f.iter.Next(ctx)
 			if err != nil {
-				// should never happen except if the underlying ds has error
+				// This should never happen except if the underlying ds has error. This is because f.iter.Head() had already
+				// checked whether we are at the end of list. For example, in a list of [1] (all invalid),
+				// Head() will return 1. If it is invalid, Next() will return 1 and move the pointer to end of list.
+				// Thus, Head() will return ErrIteratorDone next time being called.
 				return nil, err
 			}
 			continue
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -256,7 +256,7 @@ func (f *filteredTupleKeyIterator) Stop() {
 
 // Head returns the next most tuple in the underlying iterator that meets
 // the filter function this iterator was constructed with.
-// Note: the underlying iterator for unmatched filter may advance until head is satisfied.
+// Note: the underlying iterator for unmatched filter may advance until filter is satisfied.
 func (f *filteredTupleKeyIterator) Head(ctx context.Context) (*openfgav1.TupleKey, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -154,7 +154,10 @@ var _ TupleKeyIterator = (*tupleKeyIterator)(nil)
 // Next see [Iterator.Next].
 func (t *tupleKeyIterator) Next(ctx context.Context) (*openfgav1.TupleKey, error) {
 	tuple, err := t.iter.Next(ctx)
-	return tuple.GetKey(), err
+	if err != nil {
+		return nil, err
+	}
+	return tuple.GetKey(), nil
 }
 
 // Stop see [Iterator.Stop].
@@ -165,7 +168,10 @@ func (t *tupleKeyIterator) Stop() {
 // Head see [Iterator.Head].
 func (t *tupleKeyIterator) Head(ctx context.Context) (*openfgav1.TupleKey, error) {
 	tuple, err := t.iter.Head(ctx)
-	return tuple.GetKey(), err
+	if err != nil {
+		return nil, err
+	}
+	return tuple.GetKey(), nil
 }
 
 // NewTupleKeyIteratorFromTupleIterator takes a [TupleIterator] and yields
@@ -355,7 +361,7 @@ func (f *ConditionsFilteredTupleKeyIterator) Head(ctx context.Context) (*openfga
 			f.lastError = err
 			_, err = f.iter.Next(ctx)
 			if err != nil {
-				// should never happen
+				// should never happen except if the underlying ds has error
 				return nil, err
 			}
 			continue
@@ -363,7 +369,7 @@ func (f *ConditionsFilteredTupleKeyIterator) Head(ctx context.Context) (*openfga
 		if !valid {
 			_, err = f.iter.Next(ctx)
 			if err != nil {
-				// should never happen
+				// should never happen except if the underlying ds has error
 				return nil, err
 			}
 			continue
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"context"
 	"errors"
+	"sync"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 )
@@ -21,6 +22,10 @@ type Iterator[T any] interface {
 	// Stop terminates iteration over
 	// the underlying iterator.
 	Stop()
+
+	// Head will return the first item or ErrIteratorDone if the list
+	// is empty.
+	Head(ctx context.Context) (T, error)
 }
 
 // TupleIterator is an iterator for [*openfgav1.Tuple](s).
@@ -48,6 +53,15 @@ func (e *emptyTupleIterator) Next(ctx context.Context) (*openfgav1.Tuple, error)
 
 // Stop see [Iterator.Stop].
 func (e *emptyTupleIterator) Stop() {}
+
+// Head see [Iterator.Head].
+func (e *emptyTupleIterator) Head(ctx context.Context) (*openfgav1.Tuple, error) {
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
+	return nil, ErrIteratorDone
+}
 
 type combinedIterator[T any] struct {
 	iters []Iterator[T]
@@ -83,6 +97,27 @@ func (c *combinedIterator[T]) Stop() {
 			iter.Stop()
 		}
 	}
+}
+
+// Head see [Iterator.Head].
+func (c *combinedIterator[T]) Head(ctx context.Context) (T, error) {
+	for _, iter := range c.iters {
+		if iter == nil {
+			continue
+		}
+		val, err := iter.Head(ctx)
+		if err != nil {
+			if !errors.Is(err, ErrIteratorDone) {
+				return val, err
+			}
+			continue
+		}
+		return val, nil
+	}
+
+	// All iterators ended.
+	var val T
+	return val, ErrIteratorDone
 }
 
 // NewCombinedIterator takes generic iterators of a given type T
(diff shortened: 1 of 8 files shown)


pkg/storage/iterators.go
```diff
@@ -253,6 +333,46 @@ func (f *ConditionsFilteredTupleKeyIterator) Stop() {
...
+			}
+			return nil, err
+		}
+
+		valid, err := f.filter(tuple)
+		if err != nil {
+			f.lastError = err
+			_, err = f.iter.Next(ctx)
+			if err != nil {
+				// should never happen
```
elbuo8: ?

pkg/storage/iterators.go
```diff
@@ -253,6 +339,46 @@ func (f *ConditionsFilteredTupleKeyIterator) Stop() {
...
+			}
+			return nil, err
+		}
+
+		valid, err := f.filter(tuple)
+		if err != nil {
+			f.lastError = err
+			_, err = f.iter.Next(ctx)
+			if err != nil {
+				// should never happen except if the underlying ds has error
```
elbuo8, line 364: It can easily be `ErrIteratorDone` since you just called `Next`

pkg/storage/iterators.go
```diff
@@ -253,6 +339,46 @@ func (f *ConditionsFilteredTupleKeyIterator) Stop() {
...
+				return nil, err
+			}
+			continue
+		}
+		if !valid {
+			_, err = f.iter.Next(ctx)
+			if err != nil {
+				// should never happen except if the underlying ds has error
```
elbuo8, line 372: Ditto, since you called `Next`

pkg/storage/iterators.go
adriantam: In this case, the "Head" should already be ErrIteratorDone

pkg/storage/iterators.go
```diff
@@ -21,6 +22,10 @@ type Iterator[T any] interface {
 	// Stop terminates iteration over
 	// the underlying iterator.
 	Stop()
+
+	// Head will return the first item or ErrIteratorDone if the list
+	// is empty.
```
miparnisari, line 27:
```suggestion
	// Head will return the next item without advancing the iterator, or ErrIteratorDone if the list
	// is empty.
```

miparnisari (commented):
> Iterator function Head to look at the first item without advancing iterator.

could you please amend the PR description to include _why_ we need this?

pkg/storage/iterators.go
```diff
@@ -191,12 +254,35 @@ func (f *filteredTupleKeyIterator) Stop() {
 	f.iter.Stop()
 }
 
+// Head returns the next most tuple in the underlying iterator that meets
+// the filter function this iterator was constructed with.
+// Note: the underlying iterator for unmatched filter may advance until head is satisfied.
```
miparnisari, line 259:
```suggestion
// Note: the underlying iterator for unmatched filter may advance until filter is satisfied.
```

also... oof. This kind of breaks my mental model that Head was supposed to not move the iterator 😓

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -187,10 +190,74 @@ func NewSQLTupleIterator(rows *sql.Rows) *SQLTupleIterator {
 		rows:     rows,
 		resultCh: make(chan *storage.TupleRecord, 1),
 		errCh:    make(chan error, 1),
+		firstRow: nil,
```
miparnisari, line 193:
```suggestion
		firstRow: nil, // GUARDED_BY(mu)
```

It doesn't do anything, but i read that it's used in other languages https://clang.llvm.org/docs/ThreadSafetyAnalysis.html. It only helps us developers see where the mutex is being used.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -187,10 +190,74 @@ func NewSQLTupleIterator(rows *sql.Rows) *SQLTupleIterator {
...
+		mu:       sync.Mutex{},
 	}
 }
 
 func (t *SQLTupleIterator) next() (*storage.TupleRecord, error) {
+	t.mu.Lock()
+
+	if t.firstRow != nil {
```
miparnisari, line 201: I don't understand why this check is here, if you call `head()` it will read one item and return it. If then you call `next()` it will return that same item, but i'd expect it to perform a read no matter what.

pkg/storage/iterators.go
adriantam: As discussed offline, what we are doing is ok because In this case, the "Head" should already be ErrIteratorDone.

pkg/storage/iterators.go
adriantam: See above, the head should have produced the Done already.

pkg/storage/sqlcommon/sqlcommon.go
adriantam: No, calling `next()` does not depend on `head()`.  However, if we had called `head()`, we had already moved the internal iterator (because we actually called `t.rows.Next()`).  Thus, we need to consume this "firstRow" instead of calling `f.rows.Next()`.  Otherwise, that first item is lost.  I will update in the code comment to explain in more details.

elbuo8 (dismissed):

pkg/storage/sqlcommon/sqlcommon.go
senojj: It seems like this logic could be simplified, and the interface more aligned with Go's `sql.Rows` interface, by making `Head` dependent on an initial call to `Next`. In Go's `sql.Rows` interface, a call to `Scan` must be preceded by a call to `Next`. From my perspective, it is least surprising that `Head` would function much the same.

adriantam: I did not choose to use the sql format because IMO it may introduce some weirdness. The current semantic Next() itself will return the first item and increment to the iterator.  The first item is lost and it is the caller's responsibility to store it. If we choose to follow the Scan/Next semantic, this may change how caller needs to manage their "Next" usage.

miparnisari:
For my own understanding, please correct me if i'm wrong. The end goal is to be able to do this:

```
model
  schema 1.1
type user
type folder
      relations
          define viewer: a and b
          define a: [user]
          define b: [user]
          define c: [user] 
type doc
  relations
    define parent: [folder]
```

if the query is `check(user:maria, viewer, doc:1)`:
1. find all parents of doc:1
2. get an iterator for tuples of the form (folderX, a, user:maria)
3. get an iterator for tuples of the form (folderY, b, user:maria)
4. compute the intersection of (2) and (3) using a classic merge algorithm (https://github.com/Seanforfun/Algorithm-and-Leetcode/blob/master/leetcode/21.%20Merge%20Two%20Sorted%20Lists.md)

yissellokta (approved):

adriantam:
> For my own understanding, please correct me if i'm wrong. The end goal is to be able to do this:
> 
> ```
> model
>   schema 1.1
> type user
> type folder
>       relations
>           define viewer: a and b
>           define a: [user]
>           define b: [user]
>           define c: [user] 
> type doc
>   relations
>     define viewer: viewer from parent
>     define parent: [folder]
> ```
> 
> if the query is `check(user:maria, viewer, doc:1)`:
> 
> 1. find all parents of doc:1
> 2. get an iterator for tuples of the form (folderX, a, user:maria)
> 3. get an iterator for tuples of the form (folderY, b, user:maria)
> 4. compute the intersection of (2) and (3) using a classic merge algorithm (https://github.com/Seanforfun/Algorithm-and-Leetcode/blob/master/leetcode/21.%20Merge%20Two%20Sorted%20Lists.md)

Yes except the intersection will not be via merge algo - more like [finding the intersection](https://github.com/Seanforfun/Algorithm-and-Leetcode/blob/master/leetcode/349.%20Intersection%20of%20Two%20Arrays.md)

pkg/storage/iterators.go
```diff
@@ -253,6 +339,45 @@ func (f *ConditionsFilteredTupleKeyIterator) Stop() {
 	f.iter.Stop()
 }
 
+// Head returns the next most tuple in the underlying iterator that meets
+// the filter function this iterator was constructed with.
+// The underlying iterator will not advance.
```
miparnisari, line 344: it does advance

pkg/storage/memory/memory_test.go
```diff
@@ -22,98 +23,231 @@ func TestMemdbStorage(t *testing.T) {
...
-func TestStaticTupleIteratorNoRace(t *testing.T) {
-	now := timestamppb.Now()
-
-	iter := &staticIterator{
-		records: []*storage.TupleRecord{
+func TestStaticTupleIterator(t *testing.T) {
+	t.Run("empty_iterator", func(t *testing.T) {
+		tests := []struct {
+			name  string
+			mixed bool
```
miparnisari, line 30: Super minor nit:

```suggestion
			interleaveNextAndHeadCalls bool
```

or something like that

pkg/storage/mysql/mysql_test.go
```diff
@@ -43,51 +43,183 @@ func TestMySQLDatastoreAfterCloseIsNotReady(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			testDatastore := storagefixtures.RunDatastoreTestContainer(t, "mysql")
+
+			uri := testDatastore.GetConnectionURI(true)
+			ds, err := New(uri, sqlcommon.NewConfig())
+			require.NoError(t, err)
+			defer ds.Close()
+
+			ctx, cancel := context.WithCancel(context.Background())
+
```
miparnisari, line 170:
```suggestion
			ctx, cancel := context.WithCancel(context.Background())
			defer cancel()
```

I know you have `cancel` somewhere below but this ensures that in execution doesn't reach that line, the cancel() still runs

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -187,11 +190,35 @@ func NewSQLTupleIterator(rows *sql.Rows) *SQLTupleIterator {
 		rows:     rows,
 		resultCh: make(chan *storage.TupleRecord, 1),
 		errCh:    make(chan error, 1),
+		firstRow: nil, // GUARDED_BY(mu). The firstRow is used as a temporary storage place if head is called.
```
miparnisari, line 193: Nit: `// GUARDED_BY(mu)` should be on line 180

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -187,11 +190,35 @@ func NewSQLTupleIterator(rows *sql.Rows) *SQLTupleIterator {
...
+		// If firstRow is nil and Head is called, rows.Next() will return the first item and advance
+		// the iterator. Thus, we will need to store this first item so that future Head() and Next()
+		// will use this item instead. Otherwise, the first item will be lost.
+		mu: sync.Mutex{},
 	}
 }
 
 func (t *SQLTupleIterator) next() (*storage.TupleRecord, error) {
+	t.mu.Lock()
```
miparnisari, line 202: I'm not seeing a good reason to have `t.mu.Unlock` in 3 lines, instead of having `defer t.mu.Unlock` right below this line. It prevents bugs where you forget to unlock the mutex.

pkg/storage/iterators.go
```diff
@@ -21,6 +22,10 @@ type Iterator[T any] interface {
 	// Stop terminates iteration over
 	// the underlying iterator.
 	Stop()
+
+	// Head will return the first item or ErrIteratorDone if the list
+	// is empty.
+	Head(ctx context.Context) (T, error)
```
miparnisari, line 28: Nit:

```suggestion
	// Head will return the first item or ErrIteratorDone if the iterator
	// is empty.
	// It's possible for this method to advance the iterator internally, but a subsequent call to Next will not miss any results.
	Head(ctx context.Context) (T, error)
```

pkg/storage/iterators_test.go
```diff
@@ -187,14 +236,180 @@ func TestConditionsFilteredTupleKeyIterator(t *testing.T) {
...
+
+		iter1 := NewStaticTupleKeyIterator([]*openfgav1.TupleKey{tks[0]})
+		iter2 := NewStaticTupleKeyIterator([]*openfgav1.TupleKey{tks[1], tks[2]})
+		iter := NewCombinedIterator(iter1, iter2)
+		defer iter.Stop()
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		_, err := iter.Head(ctx)
+		require.Error(t, err)
+		require.NotErrorIs(t, err, ErrIteratorDone)
```
miparnisari, line 322: Um, why not `require.ErrorIs(t, err, context.Canceled)`

pkg/storage/iterators_test.go
```diff
@@ -207,21 +422,488 @@ func TestConditionsFilteredTupleKeyIterator(t *testing.T) {
...
+				iter := NewConditionsFilteredTupleKeyIterator(NewStaticTupleKeyIterator(tuples), filter)
+				t.Cleanup(iter.Stop)
+
+				ctx, cancel := context.WithCancel(context.Background())
+				cancel()
+
+				if tt.mixed {
+					_, err := iter.Head(ctx)
+					require.Error(t, err)
+					require.NotEqual(t, ErrIteratorDone, err)
```
miparnisari, line 900: Why not `require.ErrorIs(t, err, context.Canceled)` ?

pkg/storage/sqlcommon/sqlcommon.go
adriantam: want to minimize the time mu lock is being hold (especially for proto.Unmarshal in the condition as I am not sure how long that will take).

pkg/storage/mysql/mysql_test.go
adriantam: For this one, I actually want to invoke cancel at specific place (i.e., before calling Head and Next).  We are testing how context cancelled is being handled by the iterator.

pkg/storage/sqlcommon/sqlcommon.go
miparnisari: I see. What about something like this?

```go
func (t *SQLTupleIterator) next() (*storage.TupleRecord, error) {
	t.mu.RLock() // lock for reads
	firstRowAvaible := t.firstRow != nil
	t.mu.RUnlock()

	if firstRowAvaible {
		firstRow := t.firstRow
		t.mu.Lock() // lock for writes
		t.firstRow = nil
		t.mu.Unlock()
		return firstRow, nil
	}

```

pkg/storage/mysql/mysql_test.go
miparnisari: Yes, that is fine. I wasn't suggesting removing the `cancel()`, just adding the `defer cancel` as a catch-all

pkg/storage/sqlcommon/sqlcommon.go
adriantam: t.firstRow may potentially be updated in between.

elbuo8 (approved):

PR #1817, opened 2024-08-02
fix: do not error if multiple terminal relations exist during fast-path

willvedd:
## Description
As noted in #1814, when using the map of connected types for the TTU/userset "fast path" optimization, under erroneous conditions, we surfaced a `"expected exactly one terminal relation for fast path, received 2"` error. This is unexpected and indicates a flaw with the map of connected types. 

However, it is too late to detect this issue when performing a fast-path optimization. Instead of erroring and returning 500's, we should skip the fast-path for these instances and defer to the normal method of resolving TTU/userset subproblems.

The one downside to note of this approach is that flaws with connected types will not be surfaced and harder to detect.

## References
Introduced by and related to #1814 

## Review Checklist


elbuo8 (approved):

elbuo8 (changes requested):
Doesn't the same have to be applied to userset?

yissellokta (approved):

pkg/typesystem/typesystem.go
```diff
@@ -398,7 +398,7 @@ func (t *TypeSystem) TTUCanFastPath(objectType, computedRelation, userType strin
 
 	terminalRelations := t.GetTerminalRelations(objectType, computedRelation, userType)
 
-	return len(terminalRelations) > 0
```
elbuo8, line 401: Doesn't the check of line `393` have to be checking against 1 also?

willvedd: Yes, good catch. I'm going to change that.

elbuo8 (approved):

PR #1825, opened 2024-08-09
perf: tuple validation

miparnisari:
## Description

Improve speed of tuple validation by reducing redudant regex matching.

 It should be a benefit when we're reading lots of tuples from the DB.

## Testing

Before

```
BenchmarkValidateTuple
BenchmarkValidateTuple-12    	   58860	     20361 ns/op
PASS
```

image

After

```
BenchmarkValidateTuple
BenchmarkValidateTuple-12    	   65977	     18291 ns/op
PASS
```

image


pkg/tuple/tuple.go
```diff
@@ -33,7 +33,7 @@ const (
 var (
 	userIDRegex   = regexp.MustCompile(`^[^:#\s]+$`)
 	objectRegex   = regexp.MustCompile(`^[^:#\s]+:[^#:\s]+$`)
-	userSetRegex  = regexp.MustCompile(`^[^:#\s]+:[^#\s]+#[^:#\s]+$`)
+	userSetRegex  = regexp.MustCompile(`^[^:#\s]+:[^#:\s]+#[^:#\s]+$`)
```
miparnisari, line 36: Same as https://github.com/openfga/language/blob/main/pkg/go/validation/validation-rules.go#L32

```go
var RuleType       Rule = "[^:#@\\*\\s]{1,254}"
var RuleRelation  Rule = "[^:#@\\*\\s]{1,50}"
var RuleID            Rule = "[^#:\\*\\s]+"

func ValidateUserSet(userSet string) bool {
	match, _ := regexp.MatchString(fmt.Sprintf("^%s:%s#%s$", RuleType, RuleID, RuleRelation), userSet)

	return match
}
```

pkg/tuple/tuple.go
```diff
@@ -33,7 +33,7 @@ const (
 var (
 	userIDRegex   = regexp.MustCompile(`^[^:#\s]+$`)
 	objectRegex   = regexp.MustCompile(`^[^:#\s]+:[^#:\s]+$`)
-	userSetRegex  = regexp.MustCompile(`^[^:#\s]+:[^#\s]+#[^:#\s]+$`)
+	userSetRegex  = regexp.MustCompile(`^[^:#\s]+:[^#:*\s]+#[^:#*\s]+$`)
```
miparnisari, line 36: Similar to https://github.com/openfga/language/blob/main/pkg/go/validation/validation-rules.go#L32

```go
"^[^:#@\\*\\s]{1,254}:[^#:\\*\\s]+#[^:#@\\*\\s]{1,50}$",
```

internal/validation/validation_test.go
```diff
@@ -679,10 +687,47 @@ func TestValidateTuple(t *testing.T) {
...
+										typesystem.DirectRelationReference("user", ""),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedError: &tuple.InvalidTupleError{
+				Cause:    fmt.Errorf("the 'user' field is malformed"),
```
miparnisari, line 727: On `main` the error is `relation 'document#*' not found"`

internal/validation/validation.go
```diff
@@ -330,29 +330,26 @@ func ValidateUser(typesys *typesystem.TypeSystem, user string) error {
...
+		if !validObject && !validUserset {
 			return fmt.Errorf("the 'user' field must be an object (e.g. document:1) or an 'object#relation' or a typed wildcard (e.g. group:*)")
 		}
 
-		if tuple.IsObjectRelation(user) {
-			userObj, _ := tuple.SplitObjectRelation(user)
-
-			if tuple.IsTypedWildcard(userObj) {
-				return fmt.Errorf("the 'user' field cannot reference a typed wildcard in a userset value")
-			}
```
miparnisari, line 346: This was replaced by adding a `*` in the userset regex

internal/validation/validation.go
```diff
@@ -365,14 +362,5 @@ func ValidateUser(typesys *typesystem.TypeSystem, user string) error {
...
 	}
 
-	// if the model is a 1.1 model we make sure that the objectType of the 'user' field is a defined
-	// type in the model.
-	if typesystem.IsSchemaVersionSupported(schemaVersion) {
-		_, ok := typesys.GetTypeDefinition(userObjectType)
-		if !ok {
-			return &tuple.TypeNotFoundError{TypeName: userObjectType}
-		}
-	}
```
miparnisari, line 375: Moved above for easier read.

internal/validation/validation.go
```diff
@@ -330,29 +330,26 @@ func ValidateUser(typesys *typesystem.TypeSystem, user string) error {
 		return fmt.Errorf("the 'user' field is malformed")
 	}
 
+	validObject := tuple.IsValidObject(user)
+	validUserset := tuple.IsObjectRelation(user)
```
elbuo8, line 334:
```suggestion
	isUserset := tuple.IsObjectRelation(user)
```

internal/validation/validation.go
elbuo8: And remove the duplication of checking schema support.

internal/validation/validation.go
elbuo8: This case is handled much earlier with just regex, correct?

miparnisari: It is handled now by the `isValidUser` on line 329

pkg/tuple/tuple.go
senojj: do we consider `*:|#|` a valid userset? It certainly fits the pattern in this line of code.

my observation may be beyond the scope of this PR, but I understand that there are some implied constraints that aren't represented properly in our regex patterns.

are we just doing "partial" validation with this pattern? is the evaluation of this pattern dependent on other validations having been evaluated before or after this pattern is evaluated?

senojj: ah, we are just matching here to determine what it is...

miparnisari: > do we consider *:|#| a valid userset?

no, because `*` is not a valid type 😅 i'll have to amend this

> we are just matching here to determine what it is... I feel like that is different than a validation check.

this regex is validating if a string is a valid userset, it's not enough to say whether it's a valid _user_, and it's not enough to say whether it's a valid userset _for the current model_

senojj: If that is true, the regex is not checking for character length or many characters that are likely invalid such as Unicode control characters like U+0000 (null). I wouldn't trust that a value is entirely valid after being matched against this pattern.

miparnisari: Ah, this code assumes prior validation: https://github.com/openfga/api/blob/c96ec773cae9caf2a9536856bf1fa206a42ef7ef/openfga/v1/openfga_service.proto#L1134-L1141

miparnisari: @senojj opened https://github.com/openfga/openfga/issues/1876 to track that work.

elbuo8 (approved):

PR #1784, opened 2024-07-23
feat: support subject validation when using OIDC AuthN mode

Code2Life:
Support subject validation when using OIDC AuthN mode, using OPENFGA_AUTHN_OIDC_SUBJECTS env variable or authn-oidc-subjects config.

## Description

**User story:**
As a platform owner, I wanna limit which services could call OpenFGA, so that to avoid unexpected model or relations modifications by none eligible services.

For example, in a Kubernetes cluster, even if Kubernetes's OIDC configured as AuthN approach, any pod using any service account could call OpenFGA. It introduces security risks because no limits of subjects, namely service account.

With subjects limit feature added, platform owner could restrict services authenticated by Kuberenetes or other OIDC providers. For instance, in Kubernetes, the flag/env-var could be configured like this: "system:serviceaccount:{allowed-namespace}:{allowed-kubernetes-service-account}"

## References

- https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#serviceaccount-token-volume-projection

## Review Checklist
- [√] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [√] The correct base branch is being used, if not `main`
- [√] I have added tests to validate that the change in functionality is working as expected

PR #1838, opened 2024-08-13
feat: configurable request context propagation to datastore

jgraeger:
This PR adds a new option to server to configure whether the datastore should be wrapped by [storagewrappers.ContextTracerWrapper](https://github.com/openfga/openfga/blob/8d0a4f37266d82eb77f267e89f5f78598058d603/pkg/storage/storagewrappers/context.go#L16). It preserves the default behaviour.

## Description

We run a custom implementation of the `storage.OpenFGADatastore` interface for which the separation of the request context from the query context is not desirable. At the moment, the server constructor always wraps the datastore in a `ContextTracerWrapper`, doing exactly this.

This PR introduces a new configuration option to enable developers integrating with the package to control whether the request context is propagated to the datastore. It preserves the current default behaviour if the option is not set.

### Key changes

- Added a new `WithRequestContextPropagation` option function for the server constructor.
- Updated the server initialization process to incorporate this new option.

## Review Checklist


miparnisari:
Hi @jgraeger, thanks for submitting this. I'm not opposed to the change but i'd like to understand what motivated it. 

> We run a custom implementation of the storage.OpenFGADatastore interface for which the separation of the request context from the query context is not desirable. 

Do you mind giving more details on why? What side effect are you seeing?

jgraeger:
> Do you mind giving more details on why? What side effect are you seeing?

Hi @miparnisari, thank you for having a look. I'm happy to provide more details:

We operate a permission service based on OpenFGA with a custom datastore implementation that doesn't use the stdlib `database/sql` interfaces. Our setup includes a custom connection pool implementation where context cancellation doesn't necessarily close a database connection, and when it does, we handle it appropriately.

The current implementation became problematic when an issue in another internal service caused a massive spike in requests to our permission service. Many of these requests were either cancelled by clients or hit deadlines. Our implementation calls a `pool.Acquire(context.Context)` function that blocks until a connection is acquired or the context is cancelled. (The interface is similar to [non-stdlib pgxpool](https://pkg.go.dev/github.com/jackc/pgx/v5/pgxpool#Pool.Acquire))

Because request cancellations weren't propagated to our datastore, this lead to essentially having a FIFO queue of thousand of queries for which the corresponding request were long gone. This congestion in the datastore eventually lead to an unavailable service.

Other mitigations for such issues like request throttling at the API/service layer aren't applicable in our use case. 

Therefore, having the option to receive cancellations in the storage implementation would allow us to handle them appropriately and apply further optimizations to our database setup.  

Let me know if you need any further clarification or information.

pkg/server/server_test.go
```diff
@@ -505,6 +505,71 @@ func TestAvoidDeadlockWithinSingleCheckRequest(t *testing.T) {
...
+				ReadAuthorizationModel(gomock.Any(), gomock.Eq(storeID), gomock.Eq(modelID)).
+				AnyTimes().
+				DoAndReturn(func(ctx context.Context, _, _ string) (*openfgav1.AuthorizationModel, error) {
+					cancelParentCtx()
+					if tc.shouldPropagateContext {
+						require.ErrorIs(t, ctx.Err(), context.Canceled, "storage context must get cancelled if the request context is")
+					} else {
+						e := ctx.Err()
+						require.NoError(t, e, "storage context must not get canceled if request context is")
+					}
```
adriantam, line 550:
```suggestion
					e := ctx.Err() 
					if tc.shouldPropagateContext {
						require.ErrorIs(t, e, context.Canceled, "storage context must get cancelled if the request context is propagate context")
					} else {
						require.NoError(t, e, "storage context must not get canceled if request context is is disable propagate context")
					}
```
nitpik: moving the e out of individual conditional branch makes it easier to read.

pkg/server/server_test.go
```diff
@@ -505,6 +505,71 @@ func TestAvoidDeadlockWithinSingleCheckRequest(t *testing.T) {
...
+			s := MustNewServerWithOpts(
+				WithDatastore(mockDatastore),
+				WithRequestContextPropagation(tc.shouldPropagateContext),
+			)
+			t.Cleanup(func() {
+				mockDatastore.EXPECT().Close().Times(1)
+				s.Close()
+			})
+
+			_, _ = s.Check(parentCtx, &openfgav1.CheckRequest{
```
adriantam, line 564:
```suggestion
			// We do not care about the check result as we assert via mockDatastore. 
			_, _ = s.Check(parentCtx, &openfgav1.CheckRequest{
```

adriantam (approved):

jgraeger:
Thank you for the review @adriantam. I applied your suggestions.

adriantam (approved):

pkg/server/server.go
```diff
@@ -425,6 +426,17 @@ func WithDispatchThrottlingCheckResolverMaxThreshold(maxThreshold uint32) OpenFG
 	}
 }
 
+// WithRequestContextPropagation determines whether the request context is propagated to the datastore.
+// When enabled, the datastore receives cancellation signals when an API request is cancelled.
+// When disabled, datastore operations continue even if the original request context is cancelled.
+// Disabling context propagation is normally desirable to avoid unnecessary database connection churn.
+// If not specified, the default value is false (separate storage and request contexts).
+func WithRequestContextPropagation(enable bool) OpenFGAServiceV1Option {
```
miparnisari, line 434:
```suggestion
func WithDatastoreRequestContextPropagation(enable bool) OpenFGAServiceV1Option {
```

pkg/server/server_test.go
```diff
@@ -505,6 +505,72 @@ func TestAvoidDeadlockWithinSingleCheckRequest(t *testing.T) {
...
+
+			mockDatastore.EXPECT().
+				ReadAuthorizationModel(gomock.Any(), storeID, modelID).
+				AnyTimes().
+				Return(&openfgav1.AuthorizationModel{}, nil)
+
+			mockDatastore.EXPECT().
+				ReadAuthorizationModel(gomock.Any(), gomock.Eq(storeID), gomock.Eq(modelID)).
+				AnyTimes().
+				DoAndReturn(func(ctx context.Context, _, _ string) (*openfgav1.AuthorizationModel, error) {
```
miparnisari, line 543: Why are there two `EXPECT` set for the same function? 🤔

pkg/server/server.go
```diff
@@ -153,8 +153,9 @@ type Server struct {
 	listObjectsDispatchThrottler throttler.Throttler
 	listUsersDispatchThrottler   throttler.Throttler
 
-	ctx                 context.Context
-	checkTrackerEnabled bool
+	ctx                         context.Context
+	checkTrackerEnabled         bool
+	datastoreContextPropagation bool
```
miparnisari, line 158: Could you please also add 
```suggestion
	contextPropagationToDatastore bool
```

pkg/server/server.go
miparnisari, line 1: Could we please add a command-line flag for this feature? Similar to: https://github.com/openfga/openfga/pull/1784/files. This way, you can use this feature regardless of whether you use OpenFGA as a library or as a service. 😄

miparnisari:
@jgraeger could you update this PR? thanks 🙇‍♀️

matoous:
@miparnisari sorry, @jgraeger is on prolonged vacation, I will try to pick this up this week. 🙌

pkg/server/server.go
matoous: Does it make sense to use this feature if OpenFGA is used as a service? That is, is there legitimate for the flag? To my understanding, the default - using [storagewrappers.ContextTracerWrapper](https://github.com/openfga/openfga/blob/8d0a4f37266d82eb77f267e89f5f78598058d603/pkg/storage/storagewrappers/context.go#L16) - makes sense for most cases and only place where it is not desired is someone uses custom database pooling which is only possible if using OpenFGA as a library.

miparnisari: > only place where it is not desired is someone uses custom database pooling which is only possible if using OpenFGA as a library.

🤔 there is nothing preventing you from writing a custom storage adapter for your database of choice that also includes its own custom pooling

pkg/server/server.go
```diff
@@ -633,7 +645,10 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		s.listUsersDispatchThrottler = throttler.NewConstantRateThrottler(s.listUsersDispatchThrottlingFrequency, "list_users_dispatch_throttle")
 	}
 
-	s.datastore = storagewrappers.NewCachedOpenFGADatastore(storagewrappers.NewContextWrapper(s.datastore), s.maxAuthorizationModelCacheSize)
+	if !s.datastoreContextPropagation {
+		s.datastore = storagewrappers.NewContextWrapper(s.datastore)
+	}
```
miparnisari, line 650:
```suggestion
	if s.datastoreContextPropagation {
		s.datastore = storagewrappers.NewContextWrapper(s.datastore)
	}
```

if the flag is enabled we want to add the context wrapper

pkg/server/server.go
```diff
@@ -425,6 +426,17 @@ func WithDispatchThrottlingCheckResolverMaxThreshold(maxThreshold uint32) OpenFG
 	}
 }
 
+// WithRequestContextPropagation determines whether the request context is propagated to the datastore.
+// When enabled, the datastore receives cancellation signals when an API request is cancelled.
+// When disabled, datastore operations continue even if the original request context is cancelled.
+// Disabling context propagation is normally desirable to avoid unnecessary database connection churn.
+// If not specified, the default value is false (separate storage and request contexts).
```
miparnisari, line 433: the current default is true, please keep that as is

pkg/server/server_test.go
jgraeger: I updated the test

jgraeger:
@miparnisari Thank you for the review. I'm back from my longer time off.

I implemented the CLI flag and adressed the review. Let me know if this is fine, then I will make the PR to update the documentation as well.

pkg/server/server.go
jgraeger: See comment above.

adriantam:
> @miparnisari Thank you for the review. I'm back from my longer time off.
> 
> I implemented the CLI flag and adressed the review. Let me know if this is fine, then I will make the PR to update the documentation as well.

Thank you @jgraeger .  Can you update https://github.com/openfga/openfga/blob/main/CHANGELOG.md to reflect this new feature?

jgraeger:
@adriantam Sorry, I was out again. Done.

pkg/server/server.go
miparnisari: ah sorry, my bad - the default right now is that context is _not_ propagated 😆

miparnisari (dismissed):

miparnisari (approved):

jpadilla (approved):

PR #1846, opened 2024-08-14
Add memory test matrix step one

elbuo8:
## Review Checklist


tests/check/check_test.go
```diff
@@ -36,6 +36,26 @@ var tuples = []*openfgav1.TupleKey{
...
 
+func TestMatrixMemory(t *testing.T) {
+	testRunTestMatrix(t, "memory")
+}
+
+func testRunTestMatrix(t *testing.T, engine string) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+	cfg := config.MustDefaultConfig()
```
adriantam, line 47: Don't know if it will cause problem.  However, Ideally we will check for goLeak as well

tests/check/check.go
```diff
@@ -196,3 +199,430 @@ func runTest(t *testing.T, test individualTest, params testParams, contextTupleT
...
+						resp, err := client.Check(ctx, &openfgav1.CheckRequest{
+							StoreId:              storeID,
+							AuthorizationModelId: modelID,
+							TupleKey:             tupleKey,
+							ContextualTuples: &openfgav1.ContextualTupleKeys{
+								// TODO
+								TupleKeys: []*openfgav1.TupleKey{},
+							},
+							Context: assertion.Context,
+							Trace:   true,
```
adriantam, line 607: Do we want trace to be true or false?  (Most people run trace off ... do we want our test to mimic that?)

tests/check/check.go
```diff
@@ -196,3 +199,430 @@ func runTest(t *testing.T, test individualTest, params testParams, contextTupleT
...
+    define direct_wildcard_and_direct_wildcard_cond: [user:*, user:* with xcond]
+    define computed: direct
+    define computed_cond: direct_cond
+    define computed_wild: direct_wild
+    define computed_wild_cond: direct_wild_cond
+    define computed_computed: computed
+    define computed_computed_computed: computed_computed
+    define or_computed: computed or computed_cond or direct_wild
+    define and_computed: computed_cond and computed_wild
+    define butnot_computed: computed_wild_cond but not computed_computed
```
adriantam, line 234: Do we also want to have union/intersection/exclusion that does not involve cond/wildcard?  What about mixing union/intersection/exclusion?

tests/check/check.go
```diff
@@ -196,3 +199,430 @@ func runTest(t *testing.T, test individualTest, params testParams, contextTupleT
...
+						TypeDefinitions: model.GetTypeDefinitions(),
+						Conditions:      model.GetConditions(),
+					})
+					require.NoError(t, err)
+					modelID = writeModelResponse.GetAuthorizationModelId()
+				}
+
+				tuples := stage.Tuples
+				tuplesLength := len(tuples)
+				if tuplesLength > 0 {
```
senojj, line 568: nit: I don't think the `if` condition here is necessary since the loop will never enter if `tuplesLength` is `< 1`.

tests/check/check.go
```diff
@@ -196,3 +199,430 @@ func runTest(t *testing.T, test individualTest, params testParams, contextTupleT
...
+
+				if len(stage.CheckAssertions) == 0 {
+					t.Skipf("no check assertions defined")
+				}
+				for _, assertion := range stage.CheckAssertions {
+					t.Run(fmt.Sprintf("assertion_%s", assertion.Name), func(t *testing.T) {
+						detailedInfo := fmt.Sprintf("Check request: %s. Tuples: %s. Contextual tuples: %s", assertion.Tuple, stage.Tuples, assertion.ContextualTuples)
+
+						var tupleKey *openfgav1.CheckRequestTupleKey
+						if assertion.Tuple != nil {
```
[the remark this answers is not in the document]
senojj: would it make more sense to have a `nil` test case? that way we can ensure that `Check` handles the value as expected.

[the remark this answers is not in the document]
adriantam: What does that really mean though?  The GRPC / API should already have basic validation on the request.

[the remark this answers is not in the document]
senojj: Is it not possible for that to break based on a change we make to the code base?

elbuo8:
@adriantam @senojj and any other reviewer: focus on the test cases not the test runner as this was copied over the existing one. I'm less concerned about the runner and significantly more concerned about the actual cases. The runner we can improve later.

tests/check/check.go
elbuo8: Check `complexity3` and `complexity4`

tests/check/check.go
elbuo8: Turning off.

senojj (approved):

tests/check/check.go
```diff
@@ -196,3 +199,430 @@ func runTest(t *testing.T, test individualTest, params testParams, contextTupleT
...
+						}
+					})
+				}
+			})
+		}
+	})
+}
+
+func runTestMatrixSuite(t *testing.T, client ClientInterface) {
+	runTestMatrix(t, testParams{typesystem.SchemaVersion1_1, client})
```
willvedd, line 627: Understand that the focus is currently on check but where do you see LU and LO fitting into this matrix?

elbuo8: Right now the types within this package are copy pasted across the different LO/LU/Check tests. And its just unmarshalling the yaml components it needs. In the future this suite would be moved to its own package to hold the model and states.

Then we can decide if LO/LU/Checks live side by side, or they are per package and each point to the respective stage. But that is the N+1.

PR #1847, opened 2024-08-14
fix!: set a maximum limit on bytes to the WriteAssertions API

miparnisari:
## Description

This is a follow-up to https://github.com/openfga/openfga/pull/1821. 

I discovered through testing that because we declared the `assertions` table in MySQL https://github.com/openfga/openfga/blob/f33657319d7885600a19c112509868a6ee1c5b3e/assets/migrations/mysql/001_initialize_schema.sql#L32-L37 using a BLOB, we need to set a limit of 64KB on the total size of the assertions proto object.

This is a short-term solution to avoid errors. Long-term, we can look into making it similar to the limit we set on auth models through a configuration flag: https://github.com/openfga/openfga/pull/1032

⚠️ This is a breaking change on the WriteAssertions API.


elbuo8 (approved):

adriantam:
@miparnisari - is it possible to assert this on the proto file itself?  (i.e., update https://github.com/openfga/api). Or is there any limit on the number of assertion item there?

jpadilla (approved):

miparnisari:
@adriantam we have a limit of 100 here https://github.com/openfga/openfga/pull/1847#issuecomment-2291783578. Looking at https://github.com/bufbuild/protoc-gen-validate/blob/main/README.md#repeated, I cannot set a byte limit on the entire array

adriantam (approved):

PR #1848, opened 2024-08-14
fix: inefficient userfilter when there are public wildcard

adriantam:
## Description
Returned userFilter for buildCheckAssociatedObjects is inefficient if the request's tuple is public wildcard and the relation has public wildcard.

## Review Checklist


internal/graph/check.go
```diff
@@ -703,35 +740,15 @@ func (c *LocalChecker) buildCheckAssociatedObjects(req *ResolveCheckRequest, obj
...
 			},
 		}
 
-		i, err := ds.ReadStartingWithUser(ctx, storeID, storage.ReadStartingWithUserFilter{
-			ObjectType: objectType,
-			Relation:   relation,
-			UserFilter: userFilter,
-			ObjectIDs:  objectIDs,
-		}, opts)
+		// TODO: add in optimization to filter out user not matching the type
```
elbuo8, line 749: Seems to me that the TODO lets us know that the typesystem bit should not be hidden down in the `userFilter` function as it would basically determine the type of iterator it would yield.

adriantam: Update function signature to take out typesystem from user filter

elbuo8 (approved):

Issue #1567, opened 2024-04-23
Check / ListObjects API inconsistency


miparnisari:
### Checklist

- [X] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [X] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

Check and ListObjects APIs are inconsistent with respect to the treatment of condition evaluation errors.

### Expectation

Document the inconsistency, or fix the inconsistency.

### Reproduction

See https://github.com/openfga/openfga/pull/1566 for a test case

### OpenFGA version

1.5.3

### How are you running OpenFGA?

As a binary

### What datastore are you using?

In-Memory

### OpenFGA Flags

n/a

PR #1855, opened 2024-08-16
test: add tests for CheckAssociatedObjects

adriantam:
## Description
Adding unit test for buildCheckAssociatedObjects.  This will serve as building block to test consumeUsersets for graph/check.  In addition, update buildCheckAssociatedObjects so that it is not a method for the local checker 

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -740,47 +740,45 @@ func iteratorReadStartingFromUser(ctx context.Context,
 		}, opts)
 }
 
-func buildCheckAssociatedObjects(req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) CheckHandlerFunc {
-	return func(ctx context.Context) (*ResolveCheckResponse, error) {
-		ctx, span := tracer.Start(ctx, "checkAssociatedObjects")
-		defer span.End()
+func buildCheckAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "checkAssociatedObjects")
+	defer span.End()
 
-		typesys, _ := typesystem.TypesystemFromContext(ctx)
-		ds, _ := storage.RelationshipTupleReaderFromContext(ctx)
+	typesys, _ := typesystem.TypesystemFromContext(ctx)
+	ds, _ := storage.RelationshipTupleReaderFromContext(ctx)
 
-		i, err := iteratorReadStartingFromUser(ctx, typesys, ds, req, objectRel, objectIDs)
-		if err != nil {
-			telemetry.TraceError(span, err)
-			return nil, err
-		}
-
-		reqContext := req.GetContext()
-		// filter out invalid tuples yielded by the database iterator
-		filteredIter := storage.NewConditionsFilteredTupleKeyIterator(
-			storage.NewFilteredTupleKeyIterator(
-				storage.NewTupleKeyIteratorFromTupleIterator(i),
-				validation.FilterInvalidTuples(typesys),
-			),
-			buildTupleKeyConditionFilter(ctx, reqContext, typesys),
-		)
-		defer filteredIter.Stop()
-
-		allowed, err := tupleIDInSortedSet(ctx, filteredIter, objectIDs)
-		if err != nil {
-			telemetry.TraceError(span, err)
-			return nil, err
-		}
-		if allowed {
-			span.SetAttributes(attribute.Bool("allowed", true))
-		}
-		reqCount := req.GetRequestMetadata().DatastoreQueryCount + 1
-		return &ResolveCheckResponse{
-			Allowed: allowed,
-			ResolutionMetadata: &ResolveCheckResponseMetadata{
-				DatastoreQueryCount: reqCount,
-			},
-		}, nil
+	i, err := iteratorReadStartingFromUser(ctx, typesys, ds, req, objectRel, objectIDs)
+	if err != nil {
+		telemetry.TraceError(span, err)
+		return nil, err
 	}
+
+	reqContext := req.GetContext()
+	// filter out invalid tuples yielded by the database iterator
+	filteredIter := storage.NewConditionsFilteredTupleKeyIterator(
+		storage.NewFilteredTupleKeyIterator(
+			storage.NewTupleKeyIteratorFromTupleIterator(i),
+			validation.FilterInvalidTuples(typesys),
+		),
+		buildTupleKeyConditionFilter(ctx, reqContext, typesys),
+	)
+	defer filteredIter.Stop()
+
+	allowed, err := tupleIDInSortedSet(ctx, filteredIter, objectIDs)
+	if err != nil {
+		telemetry.TraceError(span, err)
+		return nil, err
+	}
+	if allowed {
+		span.SetAttributes(attribute.Bool("allowed", true))
+	}
+	reqCount := req.GetRequestMetadata().DatastoreQueryCount + 1
+	return &ResolveCheckResponse{
+		Allowed: allowed,
+		ResolutionMetadata: &ResolveCheckResponseMetadata{
+			DatastoreQueryCount: reqCount,
+		},
+	}, nil
 }
 
 // checkUsersetSlowPath will check userset or public wildcard path.
@@ -1028,7 +1026,7 @@ ConsumerLoop:
 			objectRel := newBatch.objectRelation
 			objectIDs := newBatch.objectIDs
 
-			resp, err := buildCheckAssociatedObjects(req, objectRel, objectIDs)(ctx)
+			resp, err := buildCheckAssociatedObjects(ctx, req, objectRel, objectIDs)
 			dbReads++
 			if err != nil {
 				// We don't exit because we do a best effort to find the objectId that will give `allowed=true`.
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -740,7 +740,7 @@ func iteratorReadStartingFromUser(ctx context.Context,
 		}, opts)
 }
 
-func buildCheckAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) (*ResolveCheckResponse, error) {
+func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) (*ResolveCheckResponse, error) {
 	ctx, span := tracer.Start(ctx, "checkAssociatedObjects")
 	defer span.End()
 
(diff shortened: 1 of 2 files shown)


internal/graph/check_test.go
```diff
@@ -2998,3 +3000,329 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			ts, err := typesystem.NewAndValidate(context.Background(), tt.model)
+			require.NoError(t, err)
+
+			contextStruct, err := structpb.NewStruct(tt.context)
+			require.NoError(t, err)
+
+			iterFunc := buildTupleKeyConditionFilter(context.Background(), contextStruct, ts)
```
miparnisari, line 3095: Not for this PR, but: i am starting to feel like this file will become huge with tests. That is not a bad thing, but here we are testing a function that is not strictly related to Check and could be used in other APIs.

I feel like we need a `check_utils` file or something where all the auxiliary functions could live.

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,329 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+
+				type document
+					relations
+						define can_view: [user]`),
+			context:      map[string]interface{}{},
+			conditionMet: false,
+			expectedErr:  condition.NewEvaluationError("correct_ip", fmt.Errorf("condition was not found")),
+		},
+		{
+			name:     "condition_missing_parameter",
```
miparnisari, line 3046: given this test name don't you need to define the condition in the model? otherwise this test is the same as the previous one

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,329 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+	}
+}
+
+func TestBuildCheckAssociatedObjects(t *testing.T) {
+	tests := []struct {
+		name                         string
+		model                        *openfgav1.AuthorizationModel
+		tuples                       []*openfgav1.TupleKey
+		context                      map[string]interface{}
+		dsError                      bool
```
miparnisari, line 3113: this is always false

internal/graph/check_test.go
adriantam: Absolutely, we can move the function that are not direct local checkers into this check_utils function

adriantam: I probably want to do this after the next PR (where it is nothing but move functions out into the util file)

internal/graph/check_test.go
adriantam: remove redundant call

internal/graph/check_test.go
adriantam: add test on dsError case

internal/graph/check_test.go
senojj: perhaps a new package is what we should aim for? something like `checkhelper`?

senojj: however a more meaningful package would be ideal

adriantam: We can have something like checkUtil package and put all the util function there.

miparnisari: `buildTupleKeyConditionFilter` can be re-used in Check, ListObjects and Listusers, and we can fix this: https://github.com/openfga/openfga/issues/1567

I'd move to the `storage` folder, maybe inside `iterators`

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+			for _, objectID := range tt.objectIDs {
+				objectIDs.Add(objectID)
+			}
+
+			associatedFunc := buildCheckAssociatedObjects(&ResolveCheckRequest{
+				StoreID:              ulid.Make().String(),
+				AuthorizationModelID: ulid.Make().String(),
+				TupleKey:             tuple.NewTupleKey("document:1", "viewer", "user:maria"),
+				RequestMetadata:      NewCheckRequestMetadata(20),
+			}, "group#member", objectIDs)
```
miparnisari, line 3309: Not for this PR: we can simplify the signature of this function:

```go
func buildCheckAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) (*ResolveCheckResponse, error) {
```

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKey("group:1", "member", "user:maria"),
+				tuple.NewTupleKey("group:2", "member", "user:maria"),
+				tuple.NewTupleKey("group:3", "member", "user:maria"),
+				tuple.NewTupleKey("group:4", "member", "user:maria"),
+			},
+			dsError: nil,
+			objectIDs: []string{
+				"2", "3",
+			},
```
miparnisari, line 3173: Nitpick: you don't need 4 tuples to create a truthy response, just one tuple should be enough. Or are we also testing the behavior of `tupleIDInSortedSet`, which already has its own tests? 😄

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+			t.Cleanup(ctrl.Finish)
+			ds := mocks.NewMockRelationshipTupleReader(ctrl)
+			tuples := make([]*openfgav1.Tuple, len(tt.tuples))
+			for i, tuple := range tt.tuples {
+				tuples[i] = &openfgav1.Tuple{
+					Key: tuple,
+				}
+			}
+			dsIter := storage.NewStaticTupleIterator(tuples)
+			ds.EXPECT().ReadStartingWithUser(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Return(dsIter, tt.dsError)
```
miparnisari, line 3321:
```suggestion
			ds.EXPECT().ReadStartingWithUser(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
				Times(1).
				Return(storage.NewStaticTupleIterator(tt.tuples), tt.dsError)
```

and then change the type of `tt.tuples` to `[]*openfgav1.Tuple` :)

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+			},
+			dsError: nil,
+			objectIDs: []string{
+				"2", "3",
+			},
+			context: map[string]interface{}{
+				"x": 1,
+			},
+			expectedResolveCheckResponse: &ResolveCheckResponse{
+				Allowed: false,
```
miparnisari, line 3286: Why is this false? The condition is satisfied for every tuple and there is an intersection for group:2 and group:3

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+
+				condition condX(x: int) {
+					x < 100
+				}
+`),
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKeyWithCondition("group:1", "viewer", "user:maria", "condX", nil),
+				tuple.NewTupleKeyWithCondition("group:2", "viewer", "user:maria", "condX", nil),
+				tuple.NewTupleKeyWithCondition("group:3", "viewer", "user:maria", "condX", nil),
+				tuple.NewTupleKeyWithCondition("group:4", "viewer", "user:maria", "condX", nil),
```
miparnisari, line 3237: these tuples are not valid according to the model

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+
+			objectIDs := storage.NewSortedSet()
+			for _, objectID := range tt.objectIDs {
+				objectIDs.Add(objectID)
+			}
+
+			associatedFunc := buildCheckAssociatedObjects(&ResolveCheckRequest{
+				StoreID:              ulid.Make().String(),
+				AuthorizationModelID: ulid.Make().String(),
+				TupleKey:             tuple.NewTupleKey("document:1", "viewer", "user:maria"),
```
miparnisari, line 3307: don't we need to test different inputs here too?

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,338 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+		model                        *openfgav1.AuthorizationModel
+		tuples                       []*openfgav1.TupleKey
+		context                      map[string]interface{}
+		dsError                      error
+		objectIDs                    []string
+		expectedError                bool
+		expectedResolveCheckResponse *ResolveCheckResponse
+	}{
+		{
+			name: "empty_iterator",
```
miparnisari, line 3103: should we also have `empty_object_ids_list`?

internal/graph/check_test.go
adriantam: This should be invalid test as the relation does not exist.  Let me nuke this.

internal/graph/check_test.go
adriantam: Will simplify the test - tupleIDInSortedSet already has its own tests

internal/graph/check_test.go
adriantam: Nothing in this top level function should behave differently according with various model / request.  These various combination should be in the inner function such as iteratorReadStartingFromUser / NewConditionsFilteredTupleKeyIterator / tupleIDInSortedSet.  These combo should be tested in the inner function (which is the whole point of breaking them into smaller piece so we don't have a large combo in the higher up function).  If there are specific thing we want to test, we can add them here.  I will add a TTU here for sanity for now.

miparnisari (approved):

internal/graph/check_test.go
```diff
@@ -2998,3 +3000,373 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+
+				type document
+					relations
+						define can_view: [user]`),
+			context:      map[string]interface{}{},
+			conditionMet: true,
+			expectedErr:  nil,
+		},
+		{
+			name:     "condition_not_met",
```
miparnisari, line 3030: can you add the case where the condition is met?

adriantam: Done

internal/graph/check.go
```diff
@@ -740,47 +740,45 @@ func iteratorReadStartingFromUser(ctx context.Context,
 		}, opts)
 }
 
-func (c *LocalChecker) buildCheckAssociatedObjects(req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) CheckHandlerFunc {
-	return func(ctx context.Context) (*ResolveCheckResponse, error) {
-		ctx, span := tracer.Start(ctx, "checkAssociatedObjects")
-		defer span.End()
+func buildCheckAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) (*ResolveCheckResponse, error) {
```
elbuo8, line 743:
```suggestion
func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objectRel string, objectIDs storage.SortedSet) (*ResolveCheckResponse, error) {
```

Doesn't make sense for it to be called `build` anymore since its not returning a function anymore.

miparnisari (approved):

PR #1856, opened 2024-08-16
chore: update bug report template

miparnisari:
## Description

- Mention not to use the "report a bug" flow to report security vulnerabilities
- Condense the bug report template. There are too many fields to fill out and I think a more free-form report will be less daunting on users.

Current report:
image

### DIFF
--- a/.github/ISSUE_TEMPLATE/report_a_bug.yml
+++ b/.github/ISSUE_TEMPLATE/report_a_bug.yml
@@ -25,7 +25,7 @@ body:
     id: description
     attributes:
       label: Description
-      description: Describe how to reproduce your issue, what you are seeing and what you expected to see. Please do not report security vulnerabilities here; use https://github.com/openfga/openfga/security/advisories/new or send us an email at security@openfga.dev instead
+      description: Describe how to reproduce your issue, what you are seeing and what you expected to see. If you are reporting that Check, ListObjects or ListUsers API are returning unexpected responses, please report a security vulnerability instead
     validations:
       required: true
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/report_a_bug.yml
```diff
@@ -1,18 +1,8 @@
 name: 🐛 Report a bug
-description: Have you found a bug or issue? Create a bug report for OpenFGA
+description: If you are reporting that Check, ListObjects or ListUsers API are returning unexpected responses, please report a security vulnerability instead
```
adriantam, line 2: We may need to qualify this by saying there is a Check, ListObject, ListUser behaviour delta between version. Otherwise, I am concerned that end user misunderstanding may trigger this flow.

miparnisari: I wouldn't like to put the burden on the reporter to confirm whether the issue can be reproduced in another version, we can do that for them.

.github/ISSUE_TEMPLATE/report_a_bug.yml
```diff
@@ -35,68 +25,10 @@ body:
...
-      description: Provide a clear and concise description of the issue.
-    validations:
-      required: true
-
-  - type: textarea
-    id: expectation
-    attributes:
-      label: Expectation
-      description: Tell us about the behavior you expected to see.
+      description: Describe how to reproduce your issue, what you are seeing and what you expected to see. Please do not report security vulnerabilities here; use https://github.com/openfga/openfga/security/advisories/new or send us an email at security@openfga.dev instead
```
adriantam, line 28: Do we want to re-emphasize do not send unexpected check / list objects / list users result here (in case someone skip over the description)?

adriantam (approved):

PR #1858, opened 2024-08-16
refactor: internal errors

miparnisari:
## Description 

Fixing this code 

```go
type InternalError struct {
	public   error
	internal error
}

func (e InternalError) Is(target error) bool {
	return target.Error() == e.Error()
}
```

which I believe is not a correct implementation of `Is`.

## References

- https://blog.anynines.com/posts/errors-handling-in-go-bonus-content/#A_message_to_display_to_the_user
- Some examples i've seen in the wild: https://github.com/jackc/pgx/blob/4f7e19d67df4d411ac1c19373390021a2f23aa07/rows.go#L335-L345, https://github.com/hashicorp/go-slug/blob/9b7ef214fd15260fbb4852b29d4fa83d5e9ce554/slug.go#L31-L41

## Testing

Simulating an error from the DB: `boom`, and ensuring that it appears in the log correctly:

```json
{
  "grpc_service": "openfga.v1.OpenFGAService",
  "grpc_method": "ReadChanges",
  "grpc_type": "unary",
  "trace_id": "412e7b3249aaaeef87e6864793e7d25b",
  "user_agent": "PostmanRuntime/7.41.1",
  "raw_request": {
    "store_id": "01J4W3SPVZKRYAK12S0CKDPD9R",
    "type": "",
    "page_size": null,
    "continuation_token": ""
  },
  "raw_response": {
    "code": "internal_error",
    "message": "Internal Server Error"
  },
  "query_duration_ms": "0",
  "peer.address": "127.0.0.1:41560",
  "request_id": "412e7b3249aaaeef87e6864793e7d25b",
  "store_id": "01J4W3SPVZKRYAK12S0CKDPD9R",
  "grpc_code": 4000,
  "internal_error": "boom"
}
```


pkg/server/errors/errors.go
```diff
@@ -35,23 +35,18 @@ type InternalError struct {
...
-
-func (e InternalError) Internal() error {
+// Unwrap is called by errors.Is. It returns the underlying issue.
+func (e InternalError) Unwrap() error {
 	return e.internal
 }
 
 func (e InternalError) GRPCStatus() *status.Status {
-	return status.New(codes.Code(openfgav1.InternalErrorCode_internal_error), e.public.Error())
+	st, _ := status.FromError(e.public)
```
adriantam, line 48: Do we need to do something sane if FromStatus returns not ok?

miparnisari: I don't know. I'm open to ideas. As is it today, `internalError.public` will always contain ` a `status.Error` object (line 60), so it's not possible to get a `!ok`.

adriantam: probably if !ok hard code some error?  Just a bit scary we typecast/assert without handling !ok

PR #1860, opened 2024-08-19
test: add unit test for consume usersets

adriantam:
## Description
Adding unit test for consume usersets in graph/check package

## Review Checklist


internal/graph/check_test.go
```diff
@@ -2998,3 +3005,531 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+				},
+				{
+					tuples: []*openfgav1.TupleKey{
+						tuple.NewTupleKey("group:11", "member", "user:maria"),
+						tuple.NewTupleKey("group:12", "member", "user:maria"),
+						tuple.NewTupleKey("group:13", "member", "user:maria"),
+						tuple.NewTupleKey("group:14", "member", "user:maria"),
+					},
+					err: nil,
+				},
```
miparnisari, line 3092: Is it on purpose that the three batches have the same tuples?

internal/graph/check_test.go
```diff
@@ -2998,3 +3005,531 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
 		})
 	}
 }
+
+func TestConsumeUsersets(t *testing.T) {
+	type dsResults struct {
+		tuples []*openfgav1.TupleKey
```
miparnisari, line 3011:
```suggestion
		tuples []*openfgav1.Tuple
```

makes the code easier to read down below in the test itself

internal/graph/check_test.go
```diff
@@ -2998,3 +3005,531 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+					objectRelation: "group#member",
+					objectIDs:      []string{"0", "2", "8"},
+				},
+			},
+			ctxCancelled:                 false,
+			expectedResolveCheckResponse: nil,
+			errorExpected:                fmt.Errorf("mock_error"),
+		},
+		{
+			name: "ttu_tuple_found",
```
miparnisari, line 3436: Why did we add a case specifically for TTU? There is nothing in the code under test (`consumeUsersets`) that deals with TTU, and nothing in `buildCheckAssociatedObjects` either

If you decide to remove this test please put the model in one variable so that we don't have to declare it in every test

internal/graph/check_test.go
```diff
@@ -2998,3 +3005,531 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+				},
+			},
+			usersetsChannelResult: []usersetsChannelStruct{
+				{
+					err:            nil,
+					objectRelation: "group#member",
+					objectIDs:      []string{"0", "2", "8"},
+				},
+			},
+			ctxCancelled: false,
```
miparnisari, line 3053: Nitpick: `false` is the default value, don't need to  set it

internal/graph/check_test.go
```diff
@@ -2998,3 +3005,531 @@ func TestIteratorReadStartingFromUser(t *testing.T) {
...
+					},
+					err: nil,
+				},
+			},
+			usersetsChannelResult: []usersetsChannelStruct{
+				{
+					err:            nil,
+					objectRelation: "group#member",
+					objectIDs:      []string{"0", "2", "8"},
+				},
```
miparnisari, line 3051: As I suggested in the other PR, please use the bare minimum setup needed to make this test pass (i.e. 1 tuples, 1 struct) - i don't want to have to make the intersection in my head to see if this test is supposed to pass or fail 😄

elbuo8: Agreed, in the ideal case, there wouldn't even be a typesystem involved in this test. Only that `checkAssociatedObjects` is being called with the correct parameters, nothing else.

miparnisari: We can't make that assertion because `checkAssociatedObjects` is not mock-able :/

internal/graph/check_test.go
```diff
@@ -71,6 +71,28 @@ var (
 	}
 )
 
+// helper data struct to test input for usersetsChannelType.
+type usersetsChannelStruct struct {
```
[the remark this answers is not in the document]
adriantam: The objectIDs ([SortedSet](https://github.com/openfga/openfga/blob/main/pkg/storage/sortedset.go)) does not provide an easy way to initialize with slice.

internal/graph/check_test.go
adriantam: I like things being more explicit to allow better readability :)

internal/graph/check_test.go
adriantam: nuked

internal/graph/check_test.go
adriantam: Yes, this simulates the result returned by the iterator in response to each new batch from the usersetsChannel

internal/graph/check_test.go
```diff
@@ -3412,3 +3418,365 @@ func TestCheckAssociatedObjects(t *testing.T) {
...
+				ResolutionMetadata: &ResolveCheckResponseMetadata{
+					DatastoreQueryCount: 1,
+				},
+			},
+			errorExpected: nil,
+		},
+		{
+			name: "userset_tuple_found_multi_batches",
+			tuples: []dsResults{
+				// we expect 3 ds.ReadStartingWithUser to be called in response to 3 batches from usersetsChannel
```
miparnisari, line 3474: I don't follow this 🤔 When you send the first batch

```go
{
	err:            nil,
	objectRelation: "group#member",
	objectIDs:      []string{"1"},
},
```

we'll do `resp, err := checkAssociatedObjects(ctx, req, objectRel, objectIDs)`

which does this: `i, err := iteratorReadStartingFromUser(ctx, typesys, ds, req, objectRel, objectIDs)`

if you passed `1` as the objectId, why would the datastore return `{Key: tuple.NewTupleKey("group:11", "member", "user:maria")},`?

the `objectIDs` filter passed to `storage.ReadStartingWithUser` is a strict equality, not a prefix equality

adriantam: Ah, I understand what you are getting at.  Updated to reflect this.

miparnisari (approved):

jpadilla:
Saw this possible flake recently

```
 --- FAIL: TestConsumeUsersets (0.00s)
    --- FAIL: TestConsumeUsersets/iterator_error_first_batch_but_success_second (0.00s)
        check_test.go:3106: 
            	Error Trace:	/home/runner/work/openfga/openfga/internal/graph/check_test.go:3106
            	Error:      	Not equal: 
            	            	expected: <nil>(<nil>)
            	            	actual  : *errors.errorString(&errors.errorString{s:"mock_error"})
            	Test:       	TestConsumeUsersets/iterator_error_first_batch_but_success_second
FAIL
```

PR #1862, opened 2024-08-20
test: TTU test for integration

adriantam:
## Description
First set of TTU test cases for integration

## Review Checklist

### DIFF
--- a/tests/check/check_ttu.go
+++ b/tests/check/check_ttu.go
@@ -1,0 +1,181 @@
+package check
+
+import (
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	checktest "github.com/openfga/openfga/internal/test/check"
+)
+
+var ttuCompleteTestingModelTest = []*stage{
+	{
+		Name: "direct_pa_direct_ch",
+		Tuples: []*openfgav1.TupleKey{
+			{Object: "directs-user:ttu_1", Relation: "direct", User: "user:valid"},
+			{Object: "ttus:1", Relation: "mult_parent_types", User: "directs-user:ttu_1"},
+			{Object: "directs-user:ttu_1_invalid", Relation: "direct", User: "user:ttu_invalid_group"},
+		},
+		CheckAssertions: []*checktest.Assertion{
+			{
+				Name:        "valid_user",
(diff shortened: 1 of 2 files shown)


tests/check/check_ttu.go
```diff
@@ -0,0 +1,181 @@
...
+			{
+				Name:        "invalid_user",
+				Tuple:       &openfgav1.TupleKey{Object: "ttus:1", Relation: "direct_pa_direct_ch", User: "user:invalid"},
+				Expectation: false,
+			},
+			{
+				Name:        "invalid_group",
+				Tuple:       &openfgav1.TupleKey{Object: "ttus:1", Relation: "direct_pa_direct_ch", User: "user:invalid_group"},
+				Expectation: false,
+			},
```
miparnisari, line 33: Nitpick: unsure what value this test is adding, it looks identical to the one above it

tests/check/check_ttu.go
```diff
@@ -0,0 +1,181 @@
...
+	{
+		Name: "userset_pa_userset_ch",
+		Tuples: []*openfgav1.TupleKey{
+			{Object: "ttus:5", Relation: "userset_parent", User: "usersets-user:ttu_5"},
+			{Object: "usersets-user:ttu_5", Relation: "userset", User: "directs-user:ttu_5#direct"},
+			{Object: "directs-user:ttu_5", Relation: "direct", User: "user:valid"},
+			{Object: "directs-user:ttu_5_unconnected", Relation: "direct", User: "user:ttu_5_unconnected_directuser"},
+			{Object: "directs-user:ttu_5_unconnected_userset", Relation: "direct", User: "user:ttu_5_unconnected_userset"},
+			{Object: "usersets-user:ttu_5_unconnected_userset", Relation: "userset", User: "directs-user:ttu_5_unconnected_userset#direct"},
+		},
```
miparnisari, line 157: Could we also add assertions for the case where an object has a parent (ie similar to line 151) and the parent has an associated userset, but that userset has no members so we expect `allowed=false` for any input user?

tests/check/check_ttu.go
adriantam: Will change to ttu_invalid_group.  Want to test the case where there is such user belong to `directs-user`, but that `directs-user` does not connect to the TTU itself.

tests/check/check_ttu.go
adriantam: Added both empty_userset and empty_direct

miparnisari (approved):

PR #1658, opened 2024-06-03
feat: ListUsers dispatch throttling

willvedd:
## Description
This PR adds dispatch throttling to the ListUsers API. It is very similar to the implementation of #1502 which introduced dispatch throttling to ListObjects. 

## References
- Related PR: #1502

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [x] I have added tests to validate that the change in functionality is working as expected

PR #1863, opened 2024-08-20
feat: Dispatch throttling metrics

willvedd:
## Description
Adding a new `throttled_requests_count` metric for tracking the quantity of requests that have been throttled via dispatch throttling.

This PR was tested by running `docker compose up` and observing prometheus locally:

Screenshot 2024-09-04 at 4 43 06 PM

**Note: ** This PR originally had two metrics but has been simplified. More context in my comment below.

## References
- #1440 
- #1571 
- #1658 

## Review Checklist


pkg/server/server.go
```diff
@@ -95,6 +95,18 @@ var (
...
+	throttleSucceededCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "throttled_requests_success_count",
+		Help:      "The total number of requests that have been throttled and have returned a successful response.",
+	})
+
+	throttleTimeOutCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "throttled_requests_timed_out_count",
+		Help:      "The total number of requests that have been throttled but have timed-out and returned an unsuccessful responses.",
```
adriantam, line 108: should we just call it timeout rather than unsuccessful response

adriantam (commented):
Should we have counter for how many requests was throttled + how many are throttled and timed out (vs. how many are throttled successfully + how many throttled timeout)?

From a DX stand point, an operator may be more interested more in the former than latter.

pkg/server/server.go
willvedd: We need the distinction because LU and LO will time-out and still respond with a successful status code.

willvedd:
> Should we have counter for how many requests was throttled 
@adriantam Total number of throttled requests is the sum of the two metrics.

> + how many are throttled and timed out (vs. how many are throttled successfully + how many throttled timeout)?
Possibly. As an operator I think I would be more concerned about successes/failures rather than time-outs which LU and LO are susceptible to anyway. But there could also be a world where you want to know about time-outs too. 

I think once we indicate in LU and LO wether a response is complete or not (https://github.com/orgs/openfga/discussions/110 ) then we can make better determinations around wether a throttle was "successful or not". Especially if we give the operator the ability to prevent partial responses.

adriantam (dismissed):

willvedd:
@adriantam After additional thought, I'm changing the approach to align with your suggestion. Instead of splitting the metrics on success/failure, we're now splitting on time-out/no-time-out.

pkg/server/commands/list_objects.go
```diff
@@ -306,6 +310,9 @@ func (q *ListObjectsQuery) evaluate(
 				Consistency:      req.GetConsistency(),
 			}, reverseExpandResultsChan, reverseExpandResolutionMetadata)
 			if err != nil {
+				if errors.Is(err, context.DeadlineExceeded) {
+					resolutionMetadata.DidTimeOut = true
```
adriantam, line 314: Since this is inside go func, do we need to protect DidTimeout via mutex / atomic?   (i.e., concern whether in some cases go panic due to write after write when we have detect race turned on). This was the reason why we have `WasThrottled` being `atomic.Bool`.

adriantam: Unit test to prove it is to have multi goroutine running evaluate have deadline exceeded

willvedd:
**Update:** I have decided to vastly simplify this PR by only adding a single `throttledRequestCounter` metric at this time. The reason is that detecting timeout in `Check` and `ListUsers` is straightforward but is difficult in `ListObjects` because there are several concurrency mechanisms working together in ways that aren't immediately clear or straightforward to test. The primary culprit of this complexity is needing to detect timeouts from both reverse expand and the check resolver. Both of these mechanisms feed into different concurrency controls so there is no single place to implement timeout detection. 

We will need this timeout detection in the future so this refactor of ListObjects is necessary but not in the scope of this PR. Further, its not clear how operators will use these metrics so IMO, better to be conservative and only have the single metric for now. We can always expand at a future point in time.

pkg/server/server.go
```diff
@@ -92,6 +92,12 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method", "datastore_query_count", "dispatch_count", "consistency"})
+
+	throttledRequestCounter = promauto.NewCounter(prometheus.CounterOpts{
```
adriantam, line 96: Do we want to break it down by grpc_method (so that we can distinguish between throttling in list user vs. other API in the future)?

pkg/server/list_users_test.go
```diff
@@ -400,6 +400,62 @@ func TestListUsers_Deadline(t *testing.T) {
...
+			Relation: "viewer",
+			UserFilters: []*openfgav1.UserTypeFilter{
+				{Type: "user"},
+			},
+		})
+
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+		require.Len(t, resp.GetUsers(), 1)
+	})
```
miparnisari, line 457: Just to confirm my understanding - this test is asserting on existing behavior, not on components that you added in this PR

pkg/server/server.go
```diff
@@ -92,6 +92,12 @@ var (
...
+		Namespace: build.ProjectName,
+		Name:      "throttled_requests_count",
+		Help:      "The total number of requests that have been throttled.",
+	})
```
miparnisari, line 100:
```suggestion
	throttledRequestCounter = promauto.NewCounter(prometheus.CounterOpts{
		Namespace: build.ProjectName,
		Name:      "throttled_requests_count",
		Help:      "The total number of requests that have been throttled.",
	}, []string{"grpc_service", "grpc_method"})
```

Each API can use their own throttling settings, so it's best to get data for each API individually

pkg/server/list_users_test.go
willvedd: Yes, correct. It was an attempt to improve code coverage for this PR but is also a test that should have existed in the first place. I can remove in this PR if you'd prefer.

miparnisari: No, it's ok

pkg/server/server.go
miparnisari: I thought the same but it's a counter, can you add labels on a counter? 🤔

willvedd: It's a valid question. As an operator, you're going to have different throttling configurations for the different APIs so separating into different counters may be useful. On the other hand, we're going to need to expand to also include counts for timed-out requests which would double the number of counters. Having six counters, two for each API, may become bloated for a niche metric.

willvedd: Maybe a counter isn the right type of metric?

adriantam: I thought counter metrics can have [attributes](https://opentelemetry.io/docs/specs/otel/metrics/api/#add)

miparnisari: Counter is the right type: https://prometheus.io/docs/concepts/metric_types/

I'm in favor of distinguishing by service and method, like the other metrics.

adriantam: See https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#CounterVec

adriantam: can we call it `grpc_service` so that it has similar feel as other stats that are differentiated by API?

adriantam: likely we also want to have grpc_service" as well to have the same look and feel as other API

willvedd: @adriantam you mean `grpc_method`, right?

pkg/server/server.go
```diff
@@ -92,6 +92,12 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method", "datastore_query_count", "dispatch_count", "consistency"})
+
+	throttledRequestCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "throttled_requests_count",
+		Help:      "The total number of requests that have been throttled.",
+	}, []string{"grpc_method"})
```
miparnisari, line 100:
```suggestion
	}, []string{"grpc_service", "grpc_method"})
```

willvedd: What does this achieve?

miparnisari: Make it consistent with the other metrics

jpadilla (approved):

PR #1866, opened 2024-08-22
test: add tests for Check of complexity 4

miparnisari:


tests/check/check_complexity4.go
```diff
@@ -0,0 +1,300 @@
...
+		Tuples: []*openfgav1.TupleKey{},
+		CheckAssertions: []*checktest.Assertion{
+			{
+				Name:        "path_to_user_through_ttu_and_nested_complex3",
+				Tuple:       &openfgav1.TupleKey{Object: "complexity4:pe1", Relation: "or_complex4", User: "user:valid"},
+				Expectation: true,
+			},
+			{
+				Name:        "path_to_user_through_userset_or_compute_complex3",
+				Tuple:       &openfgav1.TupleKey{Object: "complexity4:ps1", Relation: "or_complex4", User: "user:valid"},
```
adriantam, line 285: Can you verify whether we want `ps1` or `pe1`?  (also apply to the next 2 lines)

adriantam (changes requested):

tests/check/check_complexity4.go
miparnisari: Sorry I don't understand, `pe1` is in the assertion above

adriantam (approved):

PR #1868, opened 2024-08-22
fix: set consistency when calling ResolveCheck from TrackerCheckResolver

ewanharris:
## Description

The consistency parameter is getting dropped when enabling the TrackerCheckResolver, add it to the `ResolveCheckRequest` being constructed so that it's passed through and can be used.

I'm not sure whether we can potentially move to something similar to `CloneResolveCheckResponse` so that we only need to update once place to ensure that properties get carried across?

## Review Checklist

### DIFF
--- a/internal/graph/track_check_resolver.go
+++ b/internal/graph/track_check_resolver.go
@@ -204,16 +204,7 @@ func (t *TrackerCheckResolver) ResolveCheck(
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.Bool("track_execution", true))
 
-	resp, err := t.delegate.ResolveCheck(ctx, &ResolveCheckRequest{
-		StoreID:              req.GetStoreID(),
-		AuthorizationModelID: req.GetAuthorizationModelID(),
-		TupleKey:             req.GetTupleKey(),
-		ContextualTuples:     req.GetContextualTuples(),
-		RequestMetadata:      req.GetRequestMetadata(),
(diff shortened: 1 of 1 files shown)


jimmyjames (approved):
lgtm but as you said not sure if there's a better way to do this.

internal/graph/track_check_resolver.go
```diff
@@ -212,6 +212,7 @@ func (t *TrackerCheckResolver) ResolveCheck(
 		RequestMetadata:      req.GetRequestMetadata(),
 		VisitedPaths:         req.VisitedPaths,
 		Context:              req.GetContext(),
+		Consistency:          req.GetConsistency(),
```
miparnisari, line 215: Can we replace with `resp, err := t.delegate.ResolveCheck(ctx, req)` ? I don't see a reason to create a brand new object that is the same as the input

ewanharris: Makes sense to me, I wasn't sure if there was some reason for constructing a new object but looking at the associated PR it just seems like it was done that way from the start.

Done in f6ae94832a7690f07ea2d64e72f047a9b0f3c82c

miparnisari (approved):

internal/graph/track_check_resolver.go
miparnisari: Yeah it's definitely not intuitive from a developer perspective when you're supposed to use the same one or clone it 😓 we need to improve this

PR #1870, opened 2024-08-22
feat: add support for clientIDClaims in oidc

Vic-Dev:
Support for user defined `clientIDClaims`

## Description
This PR adds support for user defined client id claims, which can be set with the following configuration:
```
--authn-oidc-client-id-claims=client-id,id,<other-example>
```
Using this configuration, users can pass their clientID to the OIDC in their defined key name. The clientID will be checked in order of the defined clientIdClaims. If this configuration is not set, it will default to using:
```
`azp`,`client_id`
```

## References
https://github.com/openfga/roadmap/issues/65

## Review Checklist


internal/authn/oidc/oidc.go
```diff
@@ -21,13 +21,15 @@ import (
 	"google.golang.org/grpc/status"
 
 	"github.com/openfga/openfga/internal/authn"
+	authnPkg "github.com/openfga/openfga/pkg/authn"
```
poovamraj, line 24: Can we rename this package to something like `authclaims`, the `authnPkg` name feels a bit verbose.

pkg/authn/authn.go
```diff
@@ -0,0 +1,31 @@
+package authn
```
poovamraj, line 1: We could change this to `authclaims` to avoid clashing in other places. The content also relates more to claims than generic authn concepts

poovamraj (changes requested):
Just a small comment package naming, but it looks great otherwise :)

poovamraj (approved):

PR #1883, opened 2024-08-27
refactor: split Check into multiple files

adriantam:
## Description
No functionality change.  Only move utility functions out of graph/check.go into other files to keep graph/check and its associated unit test readable.

Added unit test for resolve_check_request_test.go

## References
Follow up on discussion from https://github.com/openfga/openfga/pull/1855#discussion_r1720289873

## Review Checklist


internal/graph/resolve_check_request_test.go
adriantam, line 1: Note that this is newly added.

internal/graph/resolve_check_response_test.go
```diff
@@ -0,0 +1,41 @@
...
+	resp2 := &ResolveCheckResponse{
+		Allowed: true,
+	}
+	clonedResp2 := CloneResolveCheckResponse(resp2)
+
+	require.NotSame(t, resp2, clonedResp2)
+	require.Equal(t, resp2.GetAllowed(), clonedResp2.GetAllowed())
+	require.NotNil(t, clonedResp2.ResolutionMetadata)
+	require.Equal(t, uint32(0), clonedResp2.GetResolutionMetadata().DatastoreQueryCount)
+	require.False(t, clonedResp2.GetResolutionMetadata().CycleDetected)
```
miparnisari, line 40: can you put this as a sub-test with an appropriate name?

internal/graph/resolve_check_response.go
```diff
@@ -0,0 +1,51 @@
+package graph
+
+// CloneResolveCheckResponse clones the provided ResolveCheckResponse.
+//
+// If 'r' defines a nil ResolutionMetadata then this function returns
+// an empty value struct for the resolution metadata instead of nil.
+func CloneResolveCheckResponse(r *ResolveCheckResponse) *ResolveCheckResponse {
```
miparnisari, line 7:
```suggestion
func (r *ResolveCheckResponse) Clone() *ResolveCheckResponse {
```

this makes the function discoverable when using auto-complete :) also applies to the request object.

pkg/checkutil/checkutil.go
miparnisari, line 1: why inside `pkg` and not `internal`?

pkg/checkutil/checkutil.go
```diff
@@ -0,0 +1,169 @@
...
+					condEvalResult.MissingParameters),
+			)
+		}
+
+		return condEvalResult.ConditionMet, nil
+	}
+}
+
+// TupleIDInSortedSet returns whether any of the iterator ID is in sorted set.
+func TupleIDInSortedSet(ctx context.Context, filteredIter *storage.ConditionsFilteredTupleKeyIterator, objectIDs storage.SortedSet) (bool, error) {
```
miparnisari, line 41:
```suggestion
func ObjectIDInSortedSet(ctx context.Context, filteredIter *storage.ConditionsFilteredTupleKeyIterator, objectIDs storage.SortedSet) (bool, error) {
```

miparnisari, line 41:
```suggestion
// ObjectIDInSortedSet returns whether any of the tuples's object ID in the iterator is in sorted set.
func ObjectIDInSortedSet(ctx context.Context, filteredIter *storage.ConditionsFilteredTupleKeyIterator, objectIDs storage.SortedSet) (bool, error) {
```

internal/graph/resolve_check_request.go
```diff
@@ -0,0 +1,64 @@
...
+			DatastoreQueryCount: r.GetRequestMetadata().DatastoreQueryCount,
+			WasThrottled:        r.GetRequestMetadata().WasThrottled,
+		},
+		VisitedPaths: maps.Clone(r.VisitedPaths),
+		Consistency:  r.Consistency,
+	}
+}
+
+func (r *ResolveCheckRequest) GetStoreID() string {
+	return r.StoreID
```
adriantam, line 39: Notice I have simplified logic from 

```
func (r *ResolveCheckRequest) GetStoreID() string {
	if r != nil {
		return r.StoreID
	}

	return ""
```

There is no way r == nil and call GetStoreID()

senojj: How can you prove that `r` is never `nil` in isolation? From the position of this function, you cannot tell how/when/where it will be used.

senojj: We absolutely _must_ employ defensive programming practices so that we properly handle cases that break our assumptions.

adriantam: Updated

miparnisari (approved):

internal/graph/resolve_check_request.go
senojj: I should rephrase my earlier statements to not be so dogmatic:

If there is a sane way to handle a `nil` point receiver value, then we should definitely do it. If the only sane option is to panic, then it may make more sense to do so.

Here, an empty string may make sense when the receiver value is `nil`.

PR #1884, opened 2024-08-28
chore: standardize text of 'missing context' error in ListUsers

miparnisari:
## Description
I am making the EvaluationError emitted by ListUsers the same as the one emitted in Check and ListObjects:

https://github.com/openfga/openfga/blob/47fb83bc711008fbd3d76cfe30cf07348ba9658f/internal/graph/check.go#L852-L857

https://github.com/openfga/openfga/blob/47fb83bc711008fbd3d76cfe30cf07348ba9658f/pkg/server/commands/reverseexpand/reverse_expand.go#L537-L542

ideally we wouldn't have to craft this error by hand in every API, but... baby steps

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -12,6 +12,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	openfgaErrors "github.com/openfga/openfga/internal/errors"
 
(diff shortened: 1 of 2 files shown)


adriantam (commented):
will it make sense to create a helper function, i.e., 
```
func errorMissingParameter(tupleKey *openfgav1.TupleKey) error {
	return fmt.Errorf("tuple '%s' is missing context parameters '%v'",
					tuple.TupleKeyToString(tupleKey),
					condEvalResult.MissingParameters)
}
```

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -962,3 +951,21 @@ func trySendResult(ctx context.Context, user foundUser, foundUsersCh chan<- foun
 		return
 	}
 }
+
+func tupleConditionMet(ctx context.Context, reqCtx *structpb.Struct, typesys *typesystem.TypeSystem, t *openfgav1.TupleKey) (bool, error) {
```
miparnisari, line 955: Exactly the same as https://github.com/openfga/openfga/blob/0a610a9607562a3fab5ffe02f0a297a5b8041c98/internal/graph/check.go#L843C1-L862C2.

We're definitely missing some abstraction here. ListObjects has similar code, too. https://github.com/openfga/openfga/blob/e20b7872fd20d7163ba9665b5896a17ee57dc904/pkg/server/commands/reverseexpand/reverse_expand.go#L529-L546

jpadilla (approved):

PR #1888, opened 2024-08-28
perf:  stream dispatches for TTU and userset slowpath

elbuo8:
## Description

Process usersets as they are discovered (in TTU and userset expansions). The existing logic would queue up N number of dispatches (where N is unbounded) and once it had all dispatches in memory it would proceed to resolve them.

With this change, as dispatches are queued up they are processed (while maintaining the existing limits). This means that any large breath model would be potentially resolved quicker in the case there is an access early on (such as a wildcard). Less memory usage as the whole list of dispatches wouldn't be needed to be held in memory (also preventing a possible OOM and less GC back pressure). 

The logic is relatively similar to `union` except that rather than a static list it uses channels on a consumer/producer pattern.

## Review Checklist


internal/graph/check.go
```diff
@@ -572,62 +572,177 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
...
 		if err != nil {
-			if errors.Is(err, storage.ErrIteratorDone) {
+			// cancelled doesn't need to flush nor send errors back to main routine
+			if storage.IterIsDoneOrCancelled(err) {
 				break
 			}
-			telemetry.TraceError(span, err)
-			return nil, err
+			trySendDispatchChan(ctx, dispatchMsg{err: err}, dispatches)
+			continue
```
miparnisari, line 600: Previously, this was a `return` but now it's a `continue`... 🤔

internal/graph/check.go
```diff
@@ -572,62 +572,177 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
...
-func (c *LocalChecker) checkUsersetSlowPath(ctx context.Context, req *ResolveCheckRequest, iter *storage.ConditionsFilteredTupleKeyIterator) (*ResolveCheckResponse, error) {
-	ctx, span := tracer.Start(ctx, "checkUsersetSlowPath")
-	defer span.End()
-	var handlers []CheckHandlerFunc
-
-	typesys, _ := typesystem.TypesystemFromContext(ctx)
+type dispatchMsg struct {
+	err          error
+	shortCircuit bool
+	dispatch     CheckHandlerFunc
```
miparnisari, line 578:
```suggestion
	handler     CheckHandlerFunc
```

So that it's more clear that doing `dispatchMsg{dispatch: c.dispatch(ctx, req, tupleKey)` is _not_ actually calling `dispatch()`, it's merely creating a function which will be executed at a later time

We could also rename `processDispatches` to `runHandlers` and `consumeDispatches` to `waitForHandlers`

internal/graph/check_test.go
```diff
@@ -3072,3 +3076,992 @@ func TestConsumeUsersets(t *testing.T) {
...
+			require.Equal(t, uint32(1), req.GetRequestMetadata().DispatchCounter.Load())
+			return nil, nil
+		})
+	dispatch := checker.dispatch(context.Background(), parentReq, tk)
+	_, _ = dispatch(context.Background())
+}
+
+// helperReceivedMessage is a helper function to listen to dispatchMsg and create
+// a list of dispatchMsg.
+func helperReceivedMessage(dispatchChan chan dispatchMsg) []dispatchMsg {
```
miparnisari, line 3113:
```suggestion
func collectMessagesFromChannel(dispatchChan chan dispatchMsg) []dispatchMsg {
```

internal/graph/check_test.go
```diff
@@ -3072,3 +3076,992 @@ func TestConsumeUsersets(t *testing.T) {
...
+					shortCircuit: false,
+					dispatch:     nil,
+				},
+			},
+		},
+		{
+			name: "good_condition_wildcard",
+			tuples: []*openfgav1.TupleKey{
+				tuple.NewTupleKeyWithCondition("group:1", "member", "user:*", "condition1", nil),
+				tuple.NewTupleKeyWithCondition("group:1", "member", "group:2#member", "condition1", nil),
```
miparnisari, line 3186: Nit: in this test, this tuple does not matter, I suggest removing it

internal/graph/check_test.go
```diff
@@ -3072,3 +3076,992 @@ func TestConsumeUsersets(t *testing.T) {
...
+			if len(expectedMsgDispatches) > 0 {
+				mockResolver.EXPECT().ResolveCheck(gomock.Any(), gomock.AssignableToTypeOf(req)).Times(len(expectedMsgDispatches)).DoAndReturn(
+					func(_ context.Context, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
+						require.NotEmpty(t, req)
+						var item *openfgav1.TupleKey
+						item, expectedMsgDispatches = expectedMsgDispatches[0], expectedMsgDispatches[1:]
+						require.Equal(t, item, req.GetTupleKey())
+						return nil, nil
+					})
+			}
```
miparnisari, line 3264: This can be simplified:

```go
for _, expectedMsg := range tt.expectedDispatches {
				if expectedMsg.dispatch != nil {
					mockResolver.EXPECT().ResolveCheck(gomock.Any(), gomock.Any()).Times(1).DoAndReturn(
						func(_ context.Context, req *ResolveCheckRequest) (*ResolveCheckResponse, error) {
							require.NotEmpty(t, req)
							require.Equal(t, expectedMsg.dispatch, req.GetTupleKey())
							return nil, nil
						})
				}
			}
```

internal/graph/check_test.go
```diff
@@ -3072,3 +3076,992 @@ func TestConsumeUsersets(t *testing.T) {
...
+func helperReceivedMessage(dispatchChan chan dispatchMsg) []dispatchMsg {
+	var receivedDispatches []dispatchMsg
+	for msg := range dispatchChan {
+		receivedDispatches = append(receivedDispatches, msg)
+	}
+	return receivedDispatches
+}
+
+// dispatchMsgAssertion is helper structure to allow easy assertion on dispatchMsg.
+type dispatchMsgAssertion struct {
```
miparnisari, line 3122: It's a code smell to me that you had to create this when you already have `dispatchMsg`.

`dispatchMsg` doesn't need to include a CheckHandlerFunc. You can replace it with what the `dispatch` function takes as parameters. And then this code:


```go
	resp, err := msg.dispatch(ctx)
	outcomes <- checkOutcome{resp: resp, err: err}
	return nil
})
```

can become

```go
	resp, err := c.dispatch(ctx, msg.parentReq, msg.tupleKey)
	outcomes <- checkOutcome{resp: resp, err: err}
	return nil
})
```

Taking it one step further, we could (not in this PR):


```go
type dispatchParams struct {
	parentReq *ResolveCheckRequest
	tk        *openfgav1.TupleKey
}

type dispatchMsg struct {
	err            error
	shortCircuit   bool
	dispatchParams *dispatchParams
}

func (c *LocalChecker) dispatch(_ context.Context, params dispatchParams) CheckHandlerFunc {
```

internal/graph/check.go
adriantam: I am in favour of just break and send the error to enforce consistency.

internal/graph/check.go
```diff
@@ -572,62 +572,177 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
...
+					Allowed: true,
+					ResolutionMetadata: &ResolveCheckResponseMetadata{
+						DatastoreQueryCount: 0,
+					},
+				}}
+				return
+			}
+			// There is a max number of goroutines based on concurrency limits
+			// a call to Go() will block until the task can be started.
+			// so if one of the inflight returns, it will be scheduled and the select will run again
```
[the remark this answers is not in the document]
adriantam: will nuke comment

internal/graph/check_test.go
adriantam: I will use the dispatchParams as it makes the unit test much easier to construct

internal/graph/check.go
adriantam: not relevant anymore as we have dispatchParams instead as per comment below.

internal/graph/check.go
```diff
@@ -572,62 +572,182 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
...
 		}
 	}
+}
 
-	resp, err := union(ctx, c.concurrencyLimit, handlers...)
+func (c *LocalChecker) processDispatches(ctx context.Context, pool *pool.ContextPool, dispatchChan chan dispatchMsg, outcomes chan checkOutcome) {
+	defer func() {
+		// We need to wait always to avoid a goroutine leak.
+		_ = pool.Wait()
+		close(outcomes)
```
senojj, line 632: Bast on personal past experience, I avoid closing a channel within a scope that did not create the channel because the (anti)pattern tends to be error prone.

In this case, it would be possible to close the channel in the calling scope, which also happens to be the scope that creates the channel.

internal/graph/check.go
```diff
@@ -572,62 +572,182 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
...
+type dispatchParams struct {
+	parentReq *ResolveCheckRequest
+	tk        *openfgav1.TupleKey
+}
 
-	typesys, _ := typesystem.TypesystemFromContext(ctx)
+type dispatchMsg struct {
+	err            error
+	shortCircuit   bool
+	dispatchParams *dispatchParams
```
senojj, line 583: Just curious, what drove the decision to use a pointer here?

senojj: I'm assuming it is because it is an optional value.

adriantam: Yes, this is optional.  If err / shortCircuit, this will be nil

senojj (approved):

internal/graph/check.go
```diff
@@ -572,62 +572,182 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
...
+		}
+	}
+}
+
+func (c *LocalChecker) consumeDispatches(ctx context.Context, req *ResolveCheckRequest, limit uint32, dispatchChan chan dispatchMsg) (*ResolveCheckResponse, error) {
+	outcomes := make(chan checkOutcome, limit)
+	cancellableCtx, cancel := context.WithCancel(ctx)
+	dispatchPool := concurrency.NewPool(cancellableCtx, int(limit))
+
+	go c.processDispatches(ctx, dispatchPool, dispatchChan, outcomes)
```
miparnisari, line 672: Notice two things:
- `dispatchPool := concurrency.NewPool(cancellableCtx, int(limit))` is passed as parameter and then never used again in this function - IMO, code smell that this really belongs inside `processDispatches`
- `processDispatches` doesn't return anything, which makes it slightly harder to unit test

We can improve `processDispatches` by having it return something: the channel. I think this is better than having to construct it outside and pass it in (because you will also have to do that in the tests).

The calling code would be

```go
func (c *LocalChecker) consumeDispatches(ctx context.Context, req *ResolveCheckRequest, limit uint32, dispatchChan chan dispatchMsg) (*ResolveCheckResponse, error) {
	cancellableCtx, cancel := context.WithCancel(ctx)
	outcomeChannel := c.processDispatches(cancellableCtx, limit, dispatchChan)
```

and then `processDispatches` would be 


```go
func (c *LocalChecker) processDispatches(ctx context.Context, limit uint32, dispatchChan chan dispatchMsg) chan checkOutcome {
	outcomes := make(chan checkOutcome, limit)
	dispatchPool := concurrency.NewPool(ctx, int(limit))

	go func() {
		// existing code (including defer)
	}()

	return outcomes
}
```

This is the proposal: [1888sug.patch](https://github.com/user-attachments/files/16912367/1888sug.patch). It passes tests locally :)

internal/graph/check_test.go
```diff
@@ -3072,3 +3076,990 @@ func TestConsumeUsersets(t *testing.T) {
...
+			}
+			outcomeChan := make(chan checkOutcome, 100)
+
+			go checker.processDispatches(ctx, dutPool, dispatchMsgChan, outcomeChan)
+
+			// now, close the channel to simulate everything is sent
+			close(dispatchMsgChan)
+			outcomes := helperReceivedOutcome(outcomeChan)
+
+			require.Equal(t, tt.expectedOutcomes, outcomes)
```
miparnisari, line 3613: Based on my other suggestion, this can be simplified:

```suggestion
			dispatchMsgChan := make(chan dispatchMsg, 100)
			for _, dispatchMsg := range tt.dispatchMsgs {
				dispatchMsgChan <- dispatchMsg
			}

			outcomeChan := checker.processDispatches(ctx, uint32(tt.poolSize), dispatchMsgChan)

			// now, close the channel to simulate everything is sent
			close(dispatchMsgChan)
			outcomes := helperReceivedOutcome(outcomeChan)

			require.Equal(t, tt.expectedOutcomes, outcomes)
```

and you can remove this line
- `dutPool := concurrency.NewPool(context.Background(), tt.poolSize)`

internal/graph/check.go
```diff
@@ -991,47 +1111,62 @@ func (c *LocalChecker) checkComputedUserset(_ context.Context, req *ResolveCheck
...
+	pool := concurrency.NewPool(cancellableCtx, 1)
+	defer func() {
+		cancelFunc()
+		// We need to wait always to avoid a goroutine leak.
+		_ = pool.Wait()
+	}()
+	pool.Go(func(ctx context.Context) error {
+		c.produceTTUDispatches(ctx, computedRelation, req, dispatchChan, iter)
+		return nil
+	})
```
miparnisari, line 1167: Oh, I think actually the same suggestion that I made before can apply here too.

miparnisari (approved):

PR #1890, opened 2024-08-28
test: use different store per test in the test matrix

miparnisari:
## Description

We want the tests not to interfere with other tests, especially for ListObjects and ListUsers assertions, which can traverse the entire universe of tuples and not just one or more paths in the graph.


jpadilla (approved):
@miparnisari I think I know the answer, but for future ref mind updating the description to know _why_ we want/need this?

miparnisari:
@jpadilla good point, added!

adriantam (approved):

PR #1891, opened 2024-08-29
 test: run ListUsers and ListObects tests as part of the matrix

miparnisari:


tests/check/check.go
```diff
@@ -1239,6 +1258,84 @@ condition xcond(x: string) {
...
+							e, ok := status.FromError(err)
+							require.True(t, ok)
+							require.Equal(t, assertion.ListObjectsErrorCode, int(e.Code()))
+							return
+						}
+						if assertion.ErrorCode == 0 {
+							require.NoError(t, err)
+							if assertion.Expectation {
+								// >=, not >, because some users have access to more than one object
+								require.GreaterOrEqual(t, len(resp.GetObjects()), 1, "at least one object should be returned")
```
adriantam, line 1287: why do we need this assertion?  Will the next line be sufficient?

tests/check/check.go
```diff
@@ -1239,6 +1258,84 @@ condition xcond(x: string) {
...
+							responseUsers := make([]string, len(resp.GetUsers()))
+							wildcardUserPresent := false
+							for i, u := range resp.GetUsers() {
+								responseUsers[i] = tuple.UserProtoToString(u)
+								if strings.HasSuffix(responseUsers[i], ":*") {
+									wildcardUserPresent = true
+								}
+							}
+							if assertion.Expectation {
+								if wildcardUserPresent {
```
adriantam, line 1328: Verify that it is the intention whenever we expect a wildcard user AND expectation == true, we will not be doing assertion for list user.

tests/check/check.go
```diff
@@ -1239,6 +1258,84 @@ condition xcond(x: string) {
...
+									return
+								}
+								// >=, not >, because some objects are related to more than one user
+								require.GreaterOrEqual(t, len(resp.GetUsers()), 1, "at least one user should be returned")
```
willvedd, line 1332: **Question:**  Wouldn't this be exactly 1 result in nearly all cases?

tests/check/check.go
```diff
@@ -1239,6 +1258,84 @@ condition xcond(x: string) {
...
+								require.Contains(t, responseUsers, assertion.Tuple.GetUser(), "user should be in the list")
```
willvedd, line 1333: **Suggestion:**

```suggestion
								require.Contains(t, responseUsers, assertion.Tuple.GetUser(), "user should be returned in response")
```

adriantam (commented):
Is it possible to break check/list objects/list users test into separate helper function?  It is a bit hard to read in such large and deeply nested function

tests/check/check.go
```diff
@@ -520,9 +529,10 @@ var matrix = individualTest{
 			},
 			CheckAssertions: []*checktest.Assertion{
```
willvedd, line 530: **Question:** `CheckAssertions` doesn't really make sense anymore, right? It may trigger a bigger change but could be appropriate.

tests/check/check.go
willvedd, line 1332:
```suggestion
								require.GreaterOrEqual(t, len(responseUsers), 1, "at least one user should be returned")
```
?

willvedd (commented):
Looks pretty good so far just a couple questions.

tests/check/check.go
senojj: To be fair, the two assertions are testing for completely different conditions.

The first condition tests that _some_ result was returned, and provides a message that indicates this specific expectation. The second condition tests the the _correct_ result was returned, and provides a message that indicates this specific expectation.

If you leave out the first condition, the a failure of the second condition does not tell you that no result was returned, only that the _wrong_ result was returned. This would make it more difficult to isolate the problem.

miparnisari: Exactly :)

tests/check/check.go
miparnisari: Yes. Here is the list of where it's 2:

image

tests/check/check.go
miparnisari: True. It's a big change though, i'd prefer to do it in a different PR

tests/check/check.go
miparnisari: So the scenario is that ListUsers response is `users:[user:*, .....]`. Then, we don't want to do the `is user:target part of the response?` assertion because it will fail.

adriantam: agree.  Can you add comment to make it easier to note that?

tests/check/check.go
```diff
@@ -1206,45 +1225,136 @@ condition xcond(x: string) {
...
+			Relation: assertion.Tuple.GetRelation(),
+			Object:   assertion.Tuple.GetObject(),
+		}
+	}
+	resp, err := client.Check(ctx, &openfgav1.CheckRequest{
+		StoreId:              storeID,
+		AuthorizationModelId: modelID,
+		TupleKey:             tupleKey,
+		ContextualTuples: &openfgav1.ContextualTupleKeys{
+			// TODO
```
adriantam, line 1259: TODO what ? :)   (I know is to have test for the context tuple keys case )

tests/check/check.go
```diff
@@ -1206,45 +1225,136 @@ condition xcond(x: string) {
...
+		User:                 assertion.Tuple.GetUser(),
+		Relation:             assertion.Tuple.GetRelation(),
+		ContextualTuples: &openfgav1.ContextualTupleKeys{
+			// TODO
+			TupleKeys: []*openfgav1.TupleKey{},
+		},
+		Context: assertion.Context,
+	})
+
+	if assertion.ListObjectsErrorCode != 0 {
```
adriantam, line 1292: Do we also want to test when assertion.ErrorCode != 0 that we will return error for ListObjects?

tests/check/check.go
```diff
@@ -1206,45 +1225,136 @@ condition xcond(x: string) {
...
+				wildcardUserPresent = true
+			}
+		}
+		if assertion.Expectation {
+			if wildcardUserPresent {
+				// if ListUsers response is [user:*, ...], we don't want to do the assertions below because they will fail.
+				return
+			}
+			// >=, not >, because some objects are related to more than one user
+			require.GreaterOrEqual(t, len(responseUsers), 1, "at least one user should be returned")
```
adriantam, line 1350:
```suggestion
			require.NotEmpty(t, responseUsers, "at least one user should be returned")
```

tests/check/check.go
```diff
@@ -1206,45 +1225,136 @@ condition xcond(x: string) {
...
+		e, ok := status.FromError(err)
+		require.True(t, ok)
+		require.Equal(t, assertion.ListObjectsErrorCode, int(e.Code()))
+		return
+	}
+	if assertion.ErrorCode == 0 {
+		require.NoError(t, err)
+		if assertion.Expectation {
+			// >=, not >, because some users have access to more than one object
+			require.GreaterOrEqual(t, len(resp.GetObjects()), 1, "at least one object should be returned")
```
adriantam, line 1303:
```suggestion
			require.NotEmpty(t, resp.GetObjects(), "at least one object should be returned")
```

adriantam (dismissed):

tests/check/check.go
adriantam: I am good either way.

tests/check/check.go
miparnisari: I experimented with adding this:

```
if assertion.ErrorCode != 0 {
		fmt.Println(resp)
		require.Error(t, err)
		require.True(t, ok)
		require.Equal(t, assertion.ErrorCode, int(e.Code()))
		return
	}
```

but there is _one_ test that fails: TestMatrixMemory/test_matrix_experimental_false/complete_testing_model/stage_usersets_userset_cond_to_computed_cond/assertion_list_objects_no_cond

It fails correctly, because ListObjects never reads the conditioned tuples, therefore doesn't throw 2000.

I _could_ bring this assertion and skip it just for this PR, but it looks uuugly.

adriantam: In this case, can we ensure all tests that has ErrorCode!=0 also has ListObjectsErrorCode != 0  (except `assertion_list_objects_no_cond`)?

tests/check/check.go
miparnisari: Yes hehe

tests/check/check.go
```diff
@@ -1206,45 +1225,152 @@ condition xcond(x: string) {
...
+	}
+	if assertion.ErrorCode != 0 {
+		require.Error(t, err)
+		e, ok := status.FromError(err)
+		require.True(t, ok)
+		require.Equal(t, assertion.ErrorCode, int(e.Code()))
+		return
+	}
+
+	if assertion.ErrorCode == 0 {
```
adriantam, line 1351: nitpike: this if statement is not necessary

tests/check/check.go
```diff
@@ -1206,45 +1225,152 @@ condition xcond(x: string) {
...
+		return
+	}
+	if assertion.ErrorCode != 0 {
+		require.Error(t, err)
+		e, ok := status.FromError(err)
+		require.True(t, ok)
+		require.Equal(t, assertion.ErrorCode, int(e.Code()))
+		return
+	}
+	if assertion.ErrorCode == 0 {
```
adriantam, line 1310: nitpik: this is not necessary as above will return

adriantam (dismissed):

adriantam (approved):

PR #1893, opened 2024-08-29
ci: add action to enforce adding changelog entries

ewanharris:
## Description

This will run on all PRs and report a failure if they do not modify the changelog, if we do not need a changelog entry for that specific PR we can add a `Skip-Changelog` label on the PR to skip the check. We can customise the label if we don't like that one.

The action used is https://github.com/dangoslen/changelog-enforcer if anyone wants to check out other config we can provide.

Some further N+1 we could make is to either also introduce things such as below, we might need to fork the action to do this:
* Parse the PR title or individual commits and if they follow conventional commits use the commit type as an additional signal for whether the changelog entry is needed (i.e. `docs`, `ci`, `test` shouldn't need an entry, `fix`, `feat`, sometimes `chore` should)
* Introduce an ignore paths field so that PRs that only touch certain folders won't need an entry
* Extend the workflow/action to also run validation if the file was touched

This might land or might not, just wanted to start a discussion about how we improve the changelog process

## Review Checklist


.github/workflows/enforce-changelog-entry.yaml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  pull_request:
+    # By default labeled/unlabeled are not included in the pull_request even so we need to list out what we want
+    types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
+
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: dangoslen/changelog-enforcer@v3
```
ewanharris, line 11: Should be a pinned sha, but I'll do that if we like it

willvedd (commented):
Looks like it is doing its job: https://github.com/openfga/openfga/actions/runs/10616883518/job/29428366571?pr=1893

Can you demonstrate applying the skip label?

ewanharris:
@willvedd, voilà!

miparnisari (approved):

.github/workflows/enforce-changelog-entry.yaml
ewanharris: Updated to use a sha so good for merge

.github/workflows/enforce-changelog-entry.yaml
```diff
@@ -0,0 +1,11 @@
+name: "Enforce changelog entry"
```
willvedd, line 1: **Suggestion:**
```suggestion
name: "Update unreleased section in CHANGELOG"
```
Just to be abundantly clear that this is about the unreleased section and not pertaining to an OpenFGA release.

.github/workflows/enforce-changelog-entry.yaml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  pull_request:
```
willvedd, line 3: **Reminder:** We need to skip this for dependabot somehow.

willvedd:
Confirmed that this GH action succeeds when updating the changelog: https://github.com/openfga/openfga/pull/1893/commits/1f6388c39c8cffd91fde3bf56d7cd3f6eacb24f7. However it doesn't validate format, which is fine for now IMO.

.github/workflows/enforce-changelog-entry.yaml
ewanharris: In 0e9dbf5398365fdb7daece91ba69501ea372375a I extended the labels to now skip on `dependencies` (that is used by dependabot by default) and `tests` (that we use)

ewanharris:
Good callout @willvedd, I added validation into the N+1 list

willvedd (approved):
Amazing!

jpadilla (approved):

PR #1896, opened 2024-08-30
test: add an intersection test for TypeSystem

senojj:
Adds a more complex test case for the `TypeSystem`.

## Description
Tests three computed usersets joined in an intersection


adriantam (approved):

adriantam (commented):
Can we rename the PR to reflect we are adding intersection not union test?  TY

PR #1902, opened 2024-09-03
feat: access control check authorization for all methods

Vic-Dev:
Add support to check authorization

## Description

For the access control experimental feature, add checks for authorization for all api methods.
- Excluding `ListStores`

To test this, create a store with the an authorization model that meets the expected spec, eg:
```
model
  schema 1.1
    
  type system
    relations
    define can_call_create_stores: [application, application:*] or admin
    define can_call_list_stores: [application, application:*] or admin
    define admin: [application]
    
  type application
    
  type module
    relations
    define can_call_write: writer or writer from store
    define store: [store]
    define writer: [application]
    
  type store
    relations
    define system: [system]
      define creator: [application]
      define can_call_delete_store: [application] or admin
      define can_call_get_store: [application] or admin
      define can_call_check: [application] or reader
      define can_call_expand: [application] or reader
      define can_call_list_objects: [application] or reader
      define can_call_list_users: [application] or reader
      define can_call_read: [application] or reader
      define can_call_read_assertions: [application] or reader or model_writer
      define can_call_read_authorization_models: [application] or reader or model_writer
      define can_call_read_changes: [application] or reader
      define can_call_write: [application] or writer
      define can_call_write_assertions: [application] or model_writer
      define can_call_write_authorization_models: [application] or model_writer
      define model_writer: [application] or admin
      define reader: [application] or admin
      define writer: [application] or admin
      define admin: [application] or creator or admin from system
```

To set up admin level access, you can set the following tuple (where `store-id` is the id for the store you created, `model-id` is the model id, and `client-id` is the client id for your application):

```
openfga.Write(context.Background(), &openfgav1.WriteRequest{
	StoreId:              <store-id>,
	AuthorizationModelId: <model-id,
	Writes: &openfgav1.WriteRequestWrites{
		TupleKeys: []*openfgav1.TupleKey{
			tuple.NewTupleKey(fmt.Sprintf("store:%s", <store-id>), "admin", fmt.Sprintf("application:%s", <client-id>)),
		},
	},
})
```

Next, run openfga with the following configuration (where `store-id` is the id for the store you created and `model-id` is the model id):

```
--experimentals=enable-access-control 
--access-control-enabled=true 
--access-control-store-id=<store-id> 
--access-control-model-id=<model-id>
```

You should now be able to run openfga requests with authorization. Initially, all requests will be unauthorized. As an example here is how you can set up `createStore` authorization:

```
ctx := authcontext.ContextWithAuthClaims(context.Background(), &authcontext.AuthClaims{ClientID:<client-id>})
openfga.Write(ctx, &openfgav1.WriteRequest{
	StoreId:              <store-id>,
	AuthorizationModelId: <model-id>,
	Writes: &openfgav1.WriteRequestWrites{
		TupleKeys: []*openfgav1.TupleKey{
			tuple.NewTupleKey("system:fga", "can_call_create_store", fmt.Sprintf("application:%s", <client-id>)),
		},
	},
})

openfga.CreateStore(ctx, &openfgav1.CreateStoreRequest{ Name: "new store" })
```

You can also see more examples on how to test authorization in the test file:
https://github.com/openfga/openfga/blob/feat/access-control-check-authz/pkg/server/server_authz_test.go#L98

## References

https://github.com/openfga/openfga/issues/1903

## Review Checklist


pkg/authz/authz.go
```diff
@@ -23,3 +82,139 @@ func NewAuthorizer(config *Config, server ServerInterface, logger logger.Logger)
...
+	allowed, err := a.individualAuthorize(ctx, clientID, relation, a.getSystem(), &openfgav1.ContextualTupleKeys{})
+	if !allowed || err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// ListAuthorizedStores returns the list of stores that the user has access to.
+func (a *Authorizer) ListAuthorizedStores(ctx context.Context, clientID string) ([]string, error) {
```
miparnisari, line 138: This is odd to me, the method above is merely checking is the clientID can call CreateStore, but here you are checking if clientID can call ListStores _and_ also executing that call, this last part seems out of scope for an authorizer

pkg/authz/authz.go
miparnisari, line 1: Could we move this code closer to the authN bit? `internal/authz/authz.go`?

pkg/server/server.go
```diff
@@ -795,6 +797,15 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
...
 }
 
+func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequest) (*openfgav1.ListObjectsResponse, error) {
+	err := s.CheckAuthz(ctx, req.GetStoreId(), authz.ListObjects)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.ListObjectsWithoutAuthz(ctx, req)
+}
```
miparnisari, line 807: Hmmm i don't know if I like how this looks architecturally speaking 🥲 Authentication is made as a middleware (https://github.com/openfga/openfga/blob/f3e9c60ee378af91c6f4e24177865961cc9b8824/internal/middleware/authn/authn.go#L12-L21) and by the time the request has made it to `server.go`, it's already been authenticated. I see no reason why authorization should be any different

pkg/server/server.go
```diff
@@ -914,9 +930,14 @@ func (s *Server) Read(ctx context.Context, req *openfgav1.ReadRequest) (*openfga
...
 	ctx = telemetry.ContextWithRPCInfo(ctx, telemetry.RPCInfo{
 		Service: s.serviceName,
-		Method:  "Read",
+		Method:  authz.Read,
 	})
 
+	err = s.CheckAuthz(ctx, req.GetStoreId(), authz.Read)
+	if err != nil {
+		return nil, err
+	}
```
miparnisari, line 939: This is a code smell having to add CheckAuthz to every endpoint here, because when we add a new endpoint we will have to remember to add this line.

Is there anything preventing you from writing this PR as an authorization middleware? Could you tell us what they are?

I prefer addressing those missing pieces first and then coming back to this PR.

cc @openfga/backend

pkg/server/server.go
```diff
@@ -1372,14 +1517,57 @@ func (s *Server) ListStores(ctx context.Context, req *openfgav1.ListStoresReques
...
+
+		if len(accessibleStores) > 0 || resp.GetContinuationToken() == "" {
+			break
+		}
+	}
+
+	return &openfgav1.ListStoresResponse{
+		Stores:            accessibleStores,
+		ContinuationToken: continuationToken,
+	}, nil
```
miparnisari, line 1570: I don't understand what or why this code is doing is doing what it's doing. You are paginating through all accessible stores but also returning a token...?

miparnisari (changes requested):

pkg/server/server.go
rhamzeh: This will need to change before the work is done - maybe we should consider taking ListStores out into it's own PR and keeping it out of this particular PR

pkg/server/server.go
willvedd: Agree with Maria here, it's not only a code smell but a security liability.

pkg/server/server_authz_test.go
```diff
@@ -0,0 +1,1570 @@
...
+
+type authzSettings struct {
+	openfga  *Server
+	clientID string
+	root     *store
+	test     *store
+}
+
+const (
+	rootStoreModel = `
```
willvedd, line 37: I see the notion of a `rootStoreModel` here but it appears to only exist in a testing context. Where is the actual model?

pkg/server/server.go
```diff
@@ -685,7 +687,7 @@ func (s *Server) Close() {
 	s.typesystemResolverStop()
 }
 
-func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequest) (*openfgav1.ListObjectsResponse, error) {
+func (s *Server) ListObjectsWithoutAuthz(ctx context.Context, req *openfgav1.ListObjectsRequest) (*openfgav1.ListObjectsResponse, error) {
```
willvedd, line 690: **Question: ** Why is this not a breaking change, or is it?

pkg/authz/authz.go
```diff
@@ -1,9 +1,68 @@
...
+	CanCallReadChanges              = "can_call_read_changes"
+
+	StoreType       = "store"
+	ApplicationType = "application"
+	SystemType      = "system"
+	RootSystemID    = "fga"
+)
+
+var (
+	ErrorResponse = &openfgav1.ForbiddenResponse{Code: 403, Message: "the principal is not authorized to perform the action"}
```
willvedd, line 63: **Question:** Is "principal" a newly introduced concept? Are there docs that use this term? If not, I'd like to converge on another existing term. Not sure what that would be though, user?

pkg/authz/authz.go
Vic-Dev: Will fix this function to just return `(bool, error)` and move the list stores logic out.

pkg/authz/authz.go
Vic-Dev: I can move context logic to a pkg called `authcontext`, and the rest to internal

pkg/server/server_authz_test.go
Vic-Dev: The idea is that users can create their own model that uses our expected relations. This model can be used as an example in documentation.

pkg/server/server.go
Vic-Dev: This is an internal function and I will update it to make it non-public

pkg/authz/authz.go
Vic-Dev: @rhamzeh any thoughts?

pkg/server/server.go
miparnisari: Discussed offline, we can't do as a middleware because people using OpenFGA as a library and not a service will miss out on this feature (since there are no middlewares when using a library)

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+		Type:                 StoreType,
+	}
+
+	// Disable authz check for the list objects request.
+	ctx = authcontext.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.ListObjects(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	authcontext.ContextWithSkipAuthzCheck(ctx, false)
```
Vic-Dev, line 195: Do I need to do this? I'm not 100% clear on how context works in Go.

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+		ContextualTuples: contextualTuples,
+	}
+
+	// Disable authz check for the check request.
+	ctx = authcontext.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.Check(ctx, req)
+	if err != nil {
+		return false, err
+	}
+	authcontext.ContextWithSkipAuthzCheck(ctx, false)
```
Vic-Dev, line 218: Same question here

pkg/server/server.go
```diff
@@ -1462,3 +1587,62 @@ func (s *Server) validateAccessControlEnabled() error {
...
+}
+
+// checkAuthz checks the authorization for calling an API method.
+func (s *Server) checkAuthz(ctx context.Context, storeID, apiMethod string) error {
+	if s.authorizer != nil && !authcontext.SkipAuthzCheckFromContext(ctx) {
+		claims, err := s.checkAuthClaims(ctx)
+		if err != nil {
+			return err
+		}
+
```
miparnisari, line 1607: can we not embed this call within `s.authorizer.Authorize(..)`?

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+	case GetStore:
+		return CanCallGetStore, nil
+	case DeleteStore:
+		return CanCallDeleteStore, nil
+	case Expand:
+		return CanCallExpand, nil
+	case ReadChanges:
+		return CanCallReadChanges, nil
+	default:
+		return "", fmt.Errorf("unknown api method: %s", apiMethod)
```
miparnisari, line 115: This error needs to be either a proper error with status code or it needs to be handled (i.e. transformed) to a proper error by the calling code, somewhere

internal/authn/authn.go
```diff
@@ -3,7 +3,7 @@ package authn
 import (
 	"context"
 
-	"github.com/openfga/openfga/pkg/authclaims"
+	"github.com/openfga/openfga/pkg/authcontext"
```
miparnisari, line 6: This PR has gotten too large, would it be possible to move the renames that are not critical to this PR to a separate PR that can go directly to `main`? 🙏

internal/authz/authz_test.go
```diff
@@ -0,0 +1,315 @@
...
+		result, err := authorizer.getRelation(ReadAuthorizationModel)
+		require.NoError(t, err)
+		require.Equal(t, CanCallReadAuthorizationModels, result)
+	})
+
+	t.Run("ReadAuthorizationModels", func(t *testing.T) {
+		result, err := authorizer.getRelation(ReadAuthorizationModels)
+		require.NoError(t, err)
+		require.Equal(t, CanCallReadAuthorizationModels, result)
+	})
```
miparnisari, line 34: These tests are better written using [table-driven tests](https://go.dev/wiki/TableDrivenTests). We use this pattern a lot in this repo :)

```go
tests := map[string]struct {
		inputAPI       string
		outputRelation string
	}{
		`read_model`: {
			inputAPI:       ReadAuthorizationModel,
			outputRelation: CanCallReadAuthorizationModels,
		},
		`read-models`: {
			inputAPI:       ReadAuthorizationModels,
			outputRelation: CanCallReadAuthorizationModels,
		},
		// FILL ME
	}

	for testname, test := range tests {
		t.Run(testname, func(t *testing.T) {
			result, err := authorizer.getRelation(test.inputAPI)
			require.NoError(t, err)
			require.Equal(t, test.outputRelation, result)
		})
	}
```

pkg/server/server.go
miparnisari: @aaguiarz what about Okta FGA?

internal/authz/authz.go
miparnisari: Every request gets its own unique context, so no, you don't need to reset it.

pkg/server/server_authz_test.go
```diff
@@ -0,0 +1,1570 @@
...
+		StoreId:         testStore.GetId(),
+		TypeDefinitions: language.MustTransformDSLToProto(testStoreModel).GetTypeDefinitions(),
+		SchemaVersion:   typesystem.SchemaVersion1_1,
+	})
+	require.NoError(t, err)
+
+	testStoreModelID := writeTestStoreAuthzModelResp.GetAuthorizationModelId()
+
+	return &authzSettings{
+		openfga:  openfga,
```
miparnisari, line 135: Nitpick: why do you store a reference to the server? The caller gave you the reference, it doesn't need it back, right?

internal/authz/server_interface.go
```diff
@@ -0,0 +1,13 @@
+package authz
+
+import (
+	"context"
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+)
+
+// ServerInterface is an interface for the server.
```
miparnisari, line 9: Please remove this comment

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+	ModelID string
+}
+
+type Authorizer struct {
+	config *Config
+	server ServerInterface
+	logger logger.Logger
+}
+
+// NewAuthorizer creates a new authorizer.
```
miparnisari, line 73: Could we make this comment more informative? Maybe use this to document decisions for future us, e.g. why this is not a middleware?

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+	ctx = authcontext.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.Check(ctx, req)
+	if err != nil {
+		return false, err
+	}
+	authcontext.ContextWithSkipAuthzCheck(ctx, false)
+
+	if !resp.GetAllowed() {
+		return false, nil
+	}
```
miparnisari, line 222:
```suggestion
	if !resp.GetAllowed() {
		return status.Error(codes.Code(ErrorResponse.GetCode()), ErrorResponse.GetMessage())
	}
```

and then the callers don't need to do

```go
authorized, err := a.individualAuthorize(ctx, clientID, relation, a.getSystem(), &openfgav1.ContextualTupleKeys{})
	if err != nil {
		return err
	}
	if !authorized {
		return status.Error(codes.Code(ErrorResponse.GetCode()), ErrorResponse.GetMessage())
	}

	return nil
```

they can just do

```go
return a.individualAuthorize(ctx, clientID, relation, a.getSystem(), &openfgav1.ContextualTupleKeys{})
```

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+	resp, err := a.server.ListObjects(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	authcontext.ContextWithSkipAuthzCheck(ctx, false)
+
+	return resp.GetObjects(), nil
+}
+
+func (a *Authorizer) individualAuthorize(ctx context.Context, clientID, relation, object string, contextualTuples *openfgav1.ContextualTupleKeys) (bool, error) {
```
miparnisari, line 200: Do we foresee needing to pass contextual tuples? if not, can we remove the last parameter as it's never being used at the moment?

pkg/server/server.go
```diff
@@ -1379,14 +1461,57 @@ func (s *Server) ListStores(ctx context.Context, req *openfgav1.ListStoresReques
...
+	for {
+		req.ContinuationToken = continuationToken
+		resp, err := q.Execute(ctx, req)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, store := range resp.GetStores() {
+			if _, ok := storesMap[store.GetId()]; ok {
```
miparnisari, line 1499: Having business logic in `server.go` breaks the existing pattern of encapsulating the APIs on `command` objects that can be re-used, it makes unit testing harder (because you need a server object), bloats `server.go`, etc.

At the very least, could we put all this logic in a new `ListObjectsWithAuthz` command?

Or, if you feel like it, could you experiment with this idea?

1. Change `NewListStoresQuery` to receive a `Authorizer` (interface, not struct) object that defaults to NoOpAuthorizer
2. Change the `Execute` method so that if `authorizer == NoOp`, then we use the existing code, else we use the code you wrote here.

cc @openfga/backend

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+			User:     a.getApplication(clientID),
+			Relation: relation,
+			Object:   object,
+		},
+		ContextualTuples: contextualTuples,
+	}
+
+	// Disable authz check for the check request.
+	ctx = authcontext.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.Check(ctx, req)
```
miparnisari, line 214: The issue with calling Check on the server and not creating a command object for this is that you won't be able to configure what runs. We will be restricted to the global Check API settings: 

- throttling and caching may be turned on
- maxConcurrentReadsForCheck may be some low number like 30

I feel like the authorizer should not use any of these settings and should use as many resources as it needs.

@Vic-Dev no action item for you, just wanted to raise this to @openfga/backend

internal/authz/authz.go
senojj: Is this function really useful? It seems only to serve to populate the internal struct properties with values that are provided directly as function input.

In this case, it likely makes more sense to export the 3 struct properties (make them public) and remove this function entirely.

Also, as opposed to removing the comment, I would prefer to see an improved comment. If the function does nothing significant to comment on, that is a good indicator that the function isn't useful.

internal/authz/authz.go
senojj: Can we instead create a custom Authorization error? Something like:
```go
var UnknownApiMethod = erros.New("unknown API method")

type AuthorizationError struct {
    Err error
}

func (e *AuthorizationError) Error() string {
    fmt.Sprintf("error authorizing request: %s", e.Err)
}

func (e *AuthorizationError) Unwrap() error {
    return e.Err
}
```
Put together:
```go
err := AuthorizationError{
    Err: UnknownApiMethod
}
```
We can then use `errors.Is` to determine what type of error we are dealing with programatically.

internal/authz/authz.go
```diff
@@ -0,0 +1,237 @@
...
+	return fmt.Sprintf(`%s:%s`, StoreType, storeID)
+}
+
+func (a *Authorizer) getApplication(clientID string) string {
+	return fmt.Sprintf(`%s:%s`, ApplicationType, clientID)
+}
+
+func (a *Authorizer) getSystem() string {
+	return fmt.Sprintf(`%s:%s`, SystemType, RootSystemID)
+}
```
senojj, line 237: These functions do not need to be methods of `Authorizer` as they do not depend on the `Authorizer` object in any way.

Also, another approach that lends itself to the type system:
```go
type StoreId string

func (s StoreId) String() string {
	return fmt.Sprintf("%s:%s", StoreType, string(s))
}

type ClientId string

func (c ClientId) String() string {
	return fmt.Sprintf("%s:%s", ApplicationType, string(c))
}

// assuming `SystemType` and `RootSystemID` are intended to be constants
var System = fmt.Sprintf("%s:%s", SystemType, RootSystemID)
```
Can be used as:
```go
storeId := StoreId("some_identifier")
output := fmt.Sprintf("%s", storeId) // output == "store:some_identifier"
```

pkg/server/server.go
Vic-Dev: Temporarily removed ListStores to be addressed in a future PR

internal/authz/authz.go
Vic-Dev: It's being used in https://github.com/openfga/openfga/pull/1914

pkg/server/server.go
Vic-Dev: Took ListStores out for now.

pkg/server/server_authz_test.go
Vic-Dev: It's so that `addAuthForRelation` can easily access the server. Maybe there is a better way to do it?

internal/authn/authn.go
Vic-Dev: Will do

Vic-Dev: Removed

internal/authz/authz.go
```diff
@@ -0,0 +1,205 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/openfga/openfga/pkg/authcontext"
+	"github.com/openfga/openfga/pkg/logger"
+)
+
+const (
+	// API methods.
+	ReadAuthorizationModel  = "ReadAuthorizationModel"
```
rhamzeh, line 18: Maybe we should prefix these and use them as enums?

e.g.
```go
type APIMethod string

const (
	APIMethodReadAuthorizationModel APIMethod = "ReadAuthorizationModels"
	APIMethodRead    APIMethod = "Read"
  ...
)
```

Same for the relations below
```go
type FgaPermission string

const (
	FgaPermissionCanCallReadAuthorizationModels FgaPermission = "can_call_read_authorization_models"
  ...
)
```

And the types
```go
type FgaTypeName string

const (
	FgaTypeNameStore FgaTypeName = "store"
	FgaTypeNameApplication FgaTypeName = "application"
  ...
)

```

internal/authz/authz.go
rhamzeh: > Caching will make the authorizer eventually consistent (is this desirable?)

Definitely yes - if anything we have higher tolerance for eventual consistency and caching here, if we can configure it separate from Check I would make it something like 1min

This is being called on every request so the more we can use the cache, the better

pkg/server/server_authz_test.go
rhamzeh: @miparnisari - do you have a suggestion for this or is this something we can merge now and fix in the final PR?

internal/authz/authz.go
rhamzeh: @senojj this seems be be the pattern we use in the server: e.g. https://github.com/openfga/openfga/blob/main/pkg/logger/logger.go#L32-L37

If we want to change that pattern can we change that in a different set of work? (and can we make sure the team is on board?)

senojj: I just want us, as a team, to be intentional with our code and not follow patterns blindly. Doing so would lead us to cargo cult thinking.

internal/authz/authz.go
senojj: Using the type name in the variable name is redundant. The Go standard library, and myself, prefer to just prefix constant names.
```go
// example: net/http/method.go
const (
	MethodGet     = "GET"
	MethodHead    = "HEAD"
	MethodPost    = "POST"
	MethodPut     = "PUT"
	MethodPatch   = "PATCH" // [RFC 5789](https://rfc-editor.org/rfc/rfc5789.html)
	MethodDelete  = "DELETE"
	MethodConnect = "CONNECT"
	MethodOptions = "OPTIONS"
	MethodTrace   = "TRACE"
)
```

internal/authz/authz.go
miparnisari: Ok yes that makes sense. What about all the other settings? E.g. do we want to apply throttling to the authorizer?

internal/authz/authz.go
miparnisari: We definitely have lots of comments in the codebase that serve no purpose but increase line count. I'd agree that just because that is the case we don't have to make the problem worse 😅

pkg/authz/authz.go
rhamzeh: I don't have anything of the top of my head - I would keep principle for now and merge it and we can discuss this again before merging the final PR

senojj: "principal" is the industry standard term for the authenticated entity that issued a request.

pkg/server/server_authz_test.go
miparnisari: Yep definitely. I don't have a suggestion atm

internal/authz/authz.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	return nil
+}
+
+// checkAuthClaims checks the auth claims in the context.
+func checkAuthClaims(ctx context.Context) (*authclaims.AuthClaims, error) {
+	claims, found := authclaims.AuthClaimsFromContext(ctx)
+	if !found || claims.ClientID == "" {
+		return nil, status.Error(codes.Internal, "client ID not found in context")
```
miparnisari, line 202: This shouldn't be `codes.Internal`. it should be `codes.InvalidArgument`.

miparnisari (approved):

PR #1906, opened 2024-09-04
fix: makes Logger.With immutable.

jcchavezs:
Currently Logger.With isn't immutable leading us to mutable loggers. For example, on every request we want to create a new logger with the request ID but with current logic it will mutate the whole logger instead of creating a new one.

Related to https://github.com/openfga/openfga-envoy/pull/17

## Review Checklist


adriantam (commented):
Please update the CHANGELOG.md unreleased section to reflect this change.

pkg/logger/logger.go
```diff
@@ -43,8 +44,8 @@ type ZapLogger struct {
 
 var _ Logger = (*ZapLogger)(nil)
 
-func (l *ZapLogger) With(fields ...zap.Field) {
-	l.Logger = l.Logger.With(fields...)
+func (l *ZapLogger) With(fields ...zap.Field) Logger {
```
miparnisari, line 47: I can't find any usages of this function that is not unit tests. What is the value of this change?

jcchavezs: The value of the change is that it fixes the behavior. The potential value of this change is that we can embed per request data into logs like I am aiming in https://github.com/openfga/openfga-envoy/pull/17

jcchavezs: BTW it is being used here https://github.com/openfga/openfga/blob/main/pkg/logger/logger_test.go#L111

adriantam: Check that it matches with the original zap logger interface - https://github.com/uber-go/zap/blob/d6b2d14fc541ed17b31b794afc139f03648842ca/logger.go#L180

adriantam (approved):

pkg/logger/logger.go
miparnisari: OK, understood now, the goal is to be able to do this:

```go
reqID := req.Attributes.GetRequest().GetHttp().GetHeaders()["x-request-id"]
logger := logger.With("request_id", reqID)
// ...
if err != nil {
		logger.Error("some error")
}
```

CHANGELOG.md
```diff
@@ -178,23 +183,22 @@ This consistency data should be used to help determine whether any form of cachi
...
 
-
 ## [1.5.3] - 2024-04-16
 
 [Full changelog](https://github.com/openfga/openfga/compare/v1.5.2...v1.5.3)
 
 ### Added
 
-* Apply tags to requests that have been intentionally throttled (https://github.com/openfga/openfga/pull/1531). This will add a new log field titled "throttled" to such requests.
+* Apply tags to requests that have been intentionally throttled (<https://github.com/openfga/openfga/pull/1531>). This will add a new log field titled "throttled" to such requests.
```
miparnisari, line 192: 🤔

miparnisari (dismissed):

CHANGELOG.md
jcchavezs: undoes all format changes.

jcchavezs:
I reverted all format changes in the CHANGELOG.md and now I need new approvals cc @miparnisari @adriantam

miparnisari (approved):

adriantam (approved):

PR #1911, opened 2024-09-05
feat: add stats on number of allowed response in check

adriantam:
## Description
Adding stats on number of allowed vs. non-allowed response in check.

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 ## Added
 
 * Stack trace when logging panics [#1904](https://github.com/openfga/openfga/pull/1904)
+* Statistics on number of allowed vs. non-allowed check [#1911](https://github.com/openfga/openfga/pull/1911)
 
 ## Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 ## Added
 
 * Stack trace when logging panics [#1904](https://github.com/openfga/openfga/pull/1904)
-* Statistics on number of allowed vs. non-allowed check [#1911](https://github.com/openfga/openfga/pull/1911)
+* New metric on number of allowed vs. non-allowed Check responses [#1911](https://github.com/openfga/openfga/pull/1911)
 
 ## Fixed
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -52,6 +52,7 @@ type ExperimentalFeatureFlag string
 const (
 	AuthorizationModelIDHeader = "Openfga-Authorization-Model-Id"
 	authorizationModelIDKey    = "authorization_model_id"
+	allowedLabel               = "allowed"
 )
 
 var tracer = otel.Tracer("openfga/pkg/server")
@@ -92,6 +93,13 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method", "datastore_query_count", "dispatch_count", "consistency"})
+
+	checkResultCounterName = "check_result_count"
+	checkResultCounter     = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      checkResultCounterName,
+		Help:      "The total number of check requests by response result",
+	}, []string{allowedLabel})
 )
 
 // A Server implements the OpenFGA service backend as both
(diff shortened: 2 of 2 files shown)


adriantam:
Do not require new stats to do this.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 ## Added
 
 * Stack trace when logging panics [#1904](https://github.com/openfga/openfga/pull/1904)
+* Statistics on number of allowed vs. non-allowed check [#1911](https://github.com/openfga/openfga/pull/1911)
```
miparnisari, line 16:
```suggestion
* New metric on number of allowed vs. non-allowed Check responses [#1911](https://github.com/openfga/openfga/pull/1911)
```

miparnisari (dismissed):

miparnisari (approved):

pkg/server/server.go
```diff
@@ -92,6 +93,13 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method", "datastore_query_count", "dispatch_count", "consistency"})
+
+	checkResultCounterName = "check_result_count"
+	checkResultCounter     = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      checkResultCounterName,
+		Help:      "The total number of check requests by response result",
+	}, []string{allowedLabel})
```
willvedd, line 102: @adriantam Should we add `grpc_service` here to be consistent with other counters?

pkg/server/server.go
```diff
@@ -92,6 +93,13 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"grpc_service", "grpc_method", "datastore_query_count", "dispatch_count", "consistency"})
+
+	checkResultCounterName = "check_result_count"
```
willvedd, line 97: What's the purpose of this variable? It's only used in the line below.

pkg/server/server.go
```diff
@@ -1024,6 +1032,8 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		Allowed: resp.Allowed,
 	}
 
+	checkResultCounter.With(prometheus.Labels{allowedLabel: strconv.FormatBool(resp.GetAllowed())}).Inc()
```
willvedd, line 1035: Simplified to:
```
checkResultCounter.WithLabelValues(strconv.FormatBool(resp.GetAllowed())).Inc()
```

pkg/server/server.go
adriantam: No, this is meant only for check.  There is no allowed in any other  API.

willvedd: It's not clear to me when why we would even add that label in the first place for any metrics at all. Ostensibly its because we want user to be able to group all openfga metrics together. I mean these metrics are namespaced too so I'm not even sure why it was made into such a big deal in https://github.com/openfga/openfga/pull/1863. Can you please explain to me what purpose the `grpc_service` label serves?

miparnisari: @adriantam @willvedd I looked deeper into this. `grpc_service` can take two values: the value for openfga, and `grpc.health.v1.health`. The latter only emits grpc metrics:
- grpc_server_handled.count
- grpc_server_handling_seconds.bucket
- grpc_server_handling_seconds.count
- grpc_server_handling_seconds.sum
- grpc_server_msg_received.count
- grpc_server_msg_sent.count
- grpc_server_started.count

As long as we have a prefix/namespace for each individual custom metric that we emit in the code, then I don't see a need to tag every metric with `grpc_service`.

willvedd: I _think_ the benefit of having the`grpc_service` metric is being able to aggregate all the metrics for OpenFGA within the context of a multi-service prometheus instance. Having the namespace helps distinguish between other services but it doesn't allow the ability to aggregate. However, that's somewhat of a guess and I'm not sure if operators actually find use in aggregating metrics like that.

miparnisari: `grpc_service ` is a tag on a metric. For aggregation of metrics, one can use the namespace, which is `openfga`. So all metrics will be named `openfga.<something>`, I think that should be enough to aggregate 🤷‍♀️

senojj: If two services were emitting metric `grpc_server_started.count` and you wanted to see their values in aggregate, you would want a tag like `grpc_service` on the metric.

Prefixing the metric name with the service name makes aggregation in a tool like DataDog rather difficult.

Issue #1903, opened 2024-09-03
Access Control Authorize API Methods


Vic-Dev:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Add support for access control experimental feature to authorize all API methods.

### Describe the ideal solution

It should check if methods are authorized before performing the requests. Additionally, there should be support for skipping check when access control is enabled.

PR #1913, opened 2024-09-06
feat: access control

Vic-Dev:
Overall Access Control feature

## Description

Access control is an experimental feature that provides fine-grained authorization on the openFGA system.

Prior to enabling the feature, it is recommended that you define an authorization model to be used to validate your API calls.

Here is an example model that can be used:
```
model
  schema 1.1

type system
  relations
    define admin: [application]
    define can_call_create_stores: admin
    define can_call_list_stores: admin

type application

type store
  relations
    define system: [system]
    define admin: [application] or admin from system
    define model_writer: [application] or admin
    define reader: [application] or admin
    define writer: [application] or admin
    define can_call_delete_store: admin
    define can_call_get_store: reader or writer or model_writer
    define can_call_check: reader
    define can_call_expand: reader
    define can_call_list_objects: reader
    define can_call_list_users: reader
    define can_call_read: reader
    define can_call_read_assertions: reader or model_writer
    define can_call_read_authorization_models: reader or model_writer
    define can_call_read_changes: reader
    define can_call_write: writer
    define can_call_write_assertions: model_writer
    define can_call_write_authorization_models: model_writer

type module
  relations
    define store: [store]
    define writer: [application]
    define can_call_write: writer or writer from store

```

Based on the above model, a store admin tuple can be added which can be used to create future authorization tuples once the feature is enabled:

```
user: "application:<clientID>", relation: "admin", object: "store:<storeID>"
```

To enable the feature, use the following settings:

```
--experimentals=enable-access-control 
--access-control-enabled=true 
--access-control-store-id=<store-id> 
--access-control-model-id=<model-id>
--authn-method=oidc
--authn-oidc-issuer=<oidc-issuer>
--authn-oidc-audience=<oidc-audience>
```

At this time, access control requires OIDC enabled with the `clientID` passed through the oidc client-id claims. It will default to using the `client_id` key on the claims, unless you override the default using the following settings:

```
--authn-oidc-client-id-claims=id,<other-example>
```

In this example, it will first look for the `clientID` in the `id` key. If it is not there it will then use the `<other-example>` key.

Now that the feature is enabled and the authorization model is set, any future API calls will be validated through the authorization model.

## References

https://github.com/openfga/openfga/issues/1903

## Review Checklist


pkg/server/config/config.go
cikasfm, line 1: @rhamzeh @Vic-Dev what was the reason for externalizing/exposing the server config?

internal/authz/authz.go
```diff
@@ -0,0 +1,434 @@
...
+	case ReadChanges:
+		return CanCallReadChanges, nil
+	default:
+		return "", AuthorizationError{Err: ErrUnknownAPIMethod}.Err
+	}
+}
+
+// Authorize checks if the user has access to the resource.
+func (a *Authorizer) Authorize(ctx context.Context, storeID, apiMethod string, modules ...string) error {
+	claims, err := checkAuthClaims(ctx)
```
miparnisari, line 204:
```suggestion
var tracer = otel.Tracer("internal/authz")

func (a *Authorizer) Authorize(ctx context.Context, storeID, apiMethod string, modules ...string) error {
       ctx, span := tracer.Start(ctx, "Authorize", trace.WithAttributes(
		attribute.String("storeID", storeID),
		attribute.String("apiMethod", apiMethod),
		attribute.String("modules", strings.Join(modules, ",")),
	))
	defer span.End()
	claims, err := checkAuthClaims(ctx)
```

This creates a new span in the trace. It's a good idea to do this in every function where something important happens. It helps me (and i'm sure others too) debug issues

internal/authz/authz.go
```diff
@@ -0,0 +1,434 @@
...
+			Object:   object,
+		},
+		ContextualTuples: contextualTuples,
+	}
+
+	// Disable authz check for the check request.
+	ctx = authclaims.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.Check(ctx, req)
+	if err != nil {
+		return err
```
miparnisari, line 369: (not blocking)

```suggestion
		return fmt.Errorf("error authorizing the store: %w", err)
```

this makes it easier for me to identify the source of the problem. Compare 

```
$ fga store --config root.config.yaml --store-id="$PROJECT_STORE_ID" get

Error: failed to get store 01J9YWE5HEDR3X0AABYFA8YAHD due to GetStore validation error for GET GetStore with body {"code":"authorization_model_not_found","message":"Authorization Model '01J9VG313KCD6573NZM80ANPET' not found"}
 with error code authorization_model_not_found error message: Authorization Model '01J9VG313KCD6573NZM80ANPET' not found
```

this gives me no indication that the error occurred during the authz call. Versus

```
Error: failed to get store 01J9YWE5HEDR3X0AABYFA8YAHD due to GetStore validation error for GET GetStore with body {"code":"authorization_model_not_found","message":"error authorizing the store: Authorization Model '01J9YWE8DR2HS2EQP5JFB05W0C' not found"}
 with error code authorization_model_not_found error message: error authorizing the store: Authorization Model '01J9YWE8DR2HS2EQP5JFB05W0C' not found
```

internal/authz/authz.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	if !resp.GetAllowed() {
+		return status.Error(codes.Code(ErrUnauthorizedResponse.GetCode()), ErrUnauthorizedResponse.GetMessage())
```
miparnisari, line 373: When testing with the CLI via ` fga store --config root.config.yaml --store-id="$PROJECT_STORE_ID"`

i don't know what HTTP status code the CLI is getting, but the message I get is

> Error: failed to get store 01J9YWE5HEDR3X0AABYFA8YAHD due to GetStore internal error for GET GetStore with body {"code":"**internal_error**","message":"the principal is not authorized to perform the action"}

which makes me think that the response was a 500, which is not what I expected (403).

I think our error handling is incorrectly transforming this to a 500 because it expects the first parameter (the code) to be >= 1000: https://github.com/openfga/openfga/blob/ce1656295b4b24181f6ac63ba8f12313eb9ac38f/pkg/server/errors/encoded_errors.go#L70-L71

pkg/server/config/config.go
Vic-Dev: The thought was that users may want to import the config, as well as we are importing it in our internal version of the codebase.

internal/authz/authz.go
Vic-Dev: I updated it to use the `openfgav1.AuthErrorCode_forbidden`, but kept the `ErrUnauthorizedResponse.GetMessage()`

miparnisari (changes requested):

Vic-Dev (commented):
Made the config internal again

.config-schema.json
```diff
@@ -627,6 +651,11 @@
                         "type": "string"
                     },
                     "x-env-variable": "OPENFGA_AUTHN_OIDC_SUBJECTS"
+                },
+                "clientIdClaims": {
+                    "description": "the OIDC client id claims that will be used to parse the clientID - configure in order of priority (first is highest). Defaults to [`azp`, `client_id`]",
+                    "type": "string",
```
jpadilla, line 657: Is this supposed to be a string or an array of strings?

internal/authz/authz.go
Vic-Dev: Putting this here is resulting in nested `error authorizing the call: error authorizing the call:...` error messages. I will refactor a bit so that we can have this context without the nesting

pkg/server/config/config.go
jpadilla: https://github.com/openfga/openfga/pull/1913#pullrequestreview-2370713805

miparnisari (approved):

jpadilla (approved):

PR #1914, opened 2024-09-06
feat: access control modules

Vic-Dev:
Adds support for modules to access control

## Description

Adds module authorization for the access control experimental feature. When write tuples request is made, it first checks if write authorization is set on the store. If there is no write authorization, it then checks if the tuple is defined in a module and if there is authorization defined for that module. If there is no authorization defined for the module or the store, it errors.

## References
https://github.com/openfga/openfga/pull/1902
https://github.com/openfga/openfga/pull/1913

## Review Checklist


internal/authz/authz.go
```diff
@@ -195,6 +276,49 @@ func (a *Authorizer) individualAuthorize(ctx context.Context, clientID, relation
 	return nil
 }
 
+// moduleAuthorize checks if the user has access to each of the modules, and exits if an error is encountered.
+func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, storeID string, modules []string) error {
```
Vic-Dev, line 280: Can I please get a review of this channel code? Basically I am concerned that there might be a case I'm not handling where there is an error in the modules, but the channel closes before the error is added to the errorChannel. @adriantam

adriantam: The problem is that there is no guarantee that line 308-312 are executed.  hence, there may be cases where the channels are closed.

internal/authz/authz.go
```diff
@@ -195,6 +276,49 @@ func (a *Authorizer) individualAuthorize(ctx context.Context, clientID, relation
...
+		close(errorChannel)
+	}()
+
+	select {
+	case err = <-errorChannel:
+		return err
+	case <-done:
+		return nil
+	}
+}
```
senojj, line 320:
```suggestion
func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, storeID string, modules []string) []error {
	var wg sync.WaitGroup
	errorChannel := make(chan error, len(modules))

	for _, module := range modules {
		wg.Add(1)
		go func(module string) {
			defer wg.Done()
			contextualTuples := openfgav1.ContextualTupleKeys{
				TupleKeys: []*openfgav1.TupleKey{
					{
						User:     StoreIDType(storeID).String(),
						Relation: StoreType,
						Object:   ModuleIDType(storeID).String(module),
					},
				},
			}

			err := a.individualAuthorize(ctx, clientID, relation, ModuleIDType(storeID).String(module), &contextualTuples)

			if err != nil {
				errorChannel <- err
			}
		}(module)
	}

	wg.Wait()
	close(errorChannel)

	errors := make([]error, len(modules))
	var i int

	for err := range errorChannel {
		if err != nil {
			errors[i] = err
			i++
		}
	}
	return errors[:i]
}
```

adriantam: @senojj : I think the original idea was to have early exit whenever there are any error.  With the suggested code, you will need to wait until all go routines are done.  in reality, I think what you propose is cleaner and at the bigger scope of things, the delta in latency should not be huge.  Just want to point it out the subtle difference here.

senojj: ah ok, understood. In that case, a separate context would need to be passed into all of the go routines that could be canceled after the first routine completes. However, we would still need to wait for all routines to synchronize before returning from the function.

Vic-Dev: For now I don't think we need to early exit since we don't really know if users will have many modules. I will apply this change, with the exception that right now we will just return the first error that is encountered, rather than all the errors as a slice

internal/authz/authz.go
Vic-Dev: Refactored as per @senojj suggestion

internal/authz/authz.go
```diff
@@ -169,6 +191,65 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
...
+	return modules, nil
+}
+
+// Define an interface that both TupleKeyWithoutCondition and TupleKey implement.
+type TupleKeyInterface interface {
+	GetObject() string
+	GetRelation() string
+}
+
+func processTupleKeys[T TupleKeyInterface](tupleKeys []T, typesys *typesystem.TypeSystem, modulesMap map[string]struct{}) (map[string]struct{}, bool, error) {
```
Vic-Dev, line 230: TODO: should return empty slice instead of false

internal/authz/authz.go
```diff
@@ -169,6 +191,65 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
...
+	}
+
+	if shouldCheckOnStore {
+		return []string{}, nil
+	}
+
+	modules := make([]string, 0, len(modulesMap))
+	for module := range modulesMap {
+		modules = append(modules, module)
+	}
```
senojj, line 219: non-blocking: pro-tip: if you know the length of the slice up-front, and you will populate each index of the slice, it is much more efficient to use direct index access to populate the array, instead of using `append()`. 
```go
modules := make([]string, len(modulesMap))
var i int
for module := range modulesMap {
    modules[i] = module
    i++
}
```

internal/authz/authz.go
```diff
@@ -151,7 +163,17 @@ func (a *Authorizer) Authorize(ctx context.Context, storeID, apiMethod string) e
 		return err
 	}
 
-	return a.individualAuthorize(ctx, claims.ClientID, relation, StoreIDType(storeID).String(), &openfgav1.ContextualTupleKeys{})
+	// Check if there is top-level authorization first, before checking modules
+	err = a.individualAuthorize(ctx, claims.ClientID, relation, StoreIDType(storeID).String(), &openfgav1.ContextualTupleKeys{})
+	if err == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
miparnisari: Excellent point, i think this should be

```
err1 := individualAuthorize()
if err1 != nil { return err1 }

err2 := moduleAuthorize()
if err2 != nil {return err2 }

return nil
```

IMO `moduleAuthorize` should be the one to check if `len(modules) !=0`

internal/authz/authz.go
```diff
@@ -169,6 +191,65 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
 	return a.individualAuthorize(ctx, claims.ClientID, relation, System, &openfgav1.ContextualTupleKeys{})
 }
 
+// GetModulesForWriteRequest returns the modules that should be checked for the write request.
+// If we encounter a type with no attached module, we should break and return no modules so that the authz check will be against the store
```
miparnisari, line 195: Please help me understand this comment. If i have

```
type user

// module A
type doc

// module B
type group
```

and I try to write three tuples
- doc:1, viewer, user:maria
- group:2, viewer, user:maria
- group:2, parent, group:3

the Write as whole will go through because `user` has no module linked to it? But i may not have permissions to write that last tuple 🤔

internal/authz/authz.go
miparnisari, line 230: Nit: could we rename `processTupleKeys` to something more meaningful such as `extractModulesFromTupleObjects`?

internal/authz/authz_test.go
```diff
@@ -179,3 +206,136 @@ func TestIndividualAuthorize(t *testing.T) {
...
+					},
+				},
+			},
+			typesys,
+		)
+		require.Error(t, err)
+		require.Empty(t, modules)
+	})
+
+	t.Run("error_when_write_tuples_errors", func(t *testing.T) {
```
miparnisari, line 276: duplicate above?

internal/authz/authz_test.go
```diff
@@ -179,3 +206,136 @@ func TestIndividualAuthorize(t *testing.T) {
...
+					},
+				},
+			},
+			typesys,
+		)
+		require.NoError(t, err)
+		require.Empty(t, modules)
+	})
+
+	t.Run("return_empty_when_a_delete_tuple_has_no_modules", func(t *testing.T) {
```
miparnisari, line 309: duplicate above?

internal/authz/authz.go
miparnisari, line 195: I don't understand this comment. If my write is 

```
folder-with-module:2, viewer, user:maria
folder:2, viewer, user:maria
```

Then because `folder` is not linked to a module, the request will be allowed? But why? I may not have permissions to write the first tuple.

pkg/server/server.go
```diff
@@ -1564,3 +1564,19 @@ func (s *Server) checkCreateStoreAuthz(ctx context.Context) error {
 	}
 	return nil
 }
+
+// checkWriteAuthz checks the authorization for modules if they exist, otherwise the store on write requests.
+func (s *Server) checkWriteAuthz(ctx context.Context, req *openfgav1.WriteRequest, typesys *typesystem.TypeSystem) error {
+	if s.authorizer != nil {
```
miparnisari, line 1570: Not for this PR: Could we make `s.authorizer` an interface so we don't have to make this nil check here? The default implementation could be a NoOpAuthorizer that allows all requests. And then if you `enable-access-controls` you instantiate your custom Authorizer

internal/authz/authz.go
Vic-Dev: Cool! Thanks

internal/authz/authz_test.go
Vic-Dev: Woops typoed the name. Good catch

internal/authz/authz_test.go
Vic-Dev: Sorry good catch again, this test isn't doing what it's supposed to

internal/authz/authz.go
Vic-Dev: If `folder` isn't in a module, then we have to check whether the user has write access on the whole store, not just the module. If a user has access on the whole store, they have access on the individual module. @rhamzeh

pkg/server/server_authz_test.go
```diff
@@ -139,17 +152,40 @@ func newSetupAuthzModelAndTuples(t *testing.T, openfga *Server, clientID string)
...
 		Writes: &openfgav1.WriteRequestWrites{
 			TupleKeys: []*openfgav1.TupleKey{
-				tuple.NewTupleKey(fmt.Sprintf("store:%s", s.testData.id), authzRelation, fmt.Sprintf("application:%s", s.clientID)),
+				tuple,
 			},
 		},
 	})
 	require.NoError(t, err)
+
+	t.Cleanup(func() {
```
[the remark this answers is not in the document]
Vic-Dev: This cleans up all the writes we do, so that the tests can be ran in any order

internal/authz/authz.go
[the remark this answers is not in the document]
Vic-Dev: @miparnisari the reason I can't check if `individualAuthorize` has an error before `moduleAuthorize` is because `individualAuthorize` checks the authorization at the store level, which we only want to accept immediately if it is authorized. If it returns an error, we then want to check if all the modules are authorized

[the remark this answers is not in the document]
Vic-Dev: @rhamzeh can you give reasoning about the above logic?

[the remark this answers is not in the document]
miparnisari: Discussed offline, the requirement is a bit counter-intuitive to me: first we want to check the coarse permissions and then the fine-grained ones. 

If that is what we want, please disregard my suggestion :)

internal/authz/authz.go
```diff
@@ -0,0 +1,327 @@
...
+	var index int
+	for _, tuple := range req.GetWrites().GetTupleKeys() {
+		tuples[index] = tuple
+		index++
+	}
+	index = 0
+	for _, tuple := range req.GetDeletes().GetTupleKeys() {
+		tuples[index] = tuple
+		index++
+	}
```
miparnisari, line 210: are you sure this is correct? by setting the index to 0 and then doing `tuples[index] =...` you are overwriting what you wrote in the previous `for` loop, no?

internal/authz/authz.go
```diff
@@ -0,0 +1,327 @@
...
+
+	modulesMap, err := extractModulesFromTupleObjects(tuples, typesys, modulesMap)
```
miparnisari, line 212: `modulesMap := make(map[string]struct{})`, why can't you crate this map within `extractModulesFromTupleObjects`?

internal/authz/authz_test.go
```diff
@@ -0,0 +1,341 @@
...
+					},
+				},
+			},
+			{
+				Type: "folder-with-module",
+				Relations: map[string]*openfgav1.Userset{
+					"viewer": typesystem.This(),
+				},
+				Metadata: &openfgav1.Metadata{
+					Module: module1,
```
miparnisari, line 246: Please add one more module to this model, and then write a test where the Write call has 1 write and 1 delete, both with a different module

internal/authz/authz.go
Vic-Dev: Thanks it was a bug!

internal/authz/authz_test.go
```diff
@@ -179,3 +206,136 @@ func TestIndividualAuthorize(t *testing.T) {
...
+					TupleKeys: []*openfgav1.TupleKeyWithoutCondition{
+						{Object: "folder-with-module:2", Relation: "viewer", User: "user:jon"},
+					},
+				},
+			},
+			typesys,
+		)
+		require.NoError(t, err)
+		require.Equal(t, []string{module1}, modules)
+	})
```
adriantam, line 340: Do we need the case where return modules for write?  We only test the case for delete

internal/authz/authz.go
```diff
@@ -196,6 +223,65 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
...
+
+// Define an interface that both TupleKeyWithoutCondition and TupleKey implement.
+type TupleKeyInterface interface {
+	GetObject() string
+	GetRelation() string
+}
+
+// extractModulesFromTuples extracts the modules from the tuples. If a type has no module, we
+// return an empty map so that the caller can handle authorization for tuples without modules.
+func extractModulesFromTuples[T TupleKeyInterface](tupleKeys []T, typesys *typesystem.TypeSystem) (map[string]struct{}, error) {
```
adriantam, line 264: Will it make sense to add unit test for this function?

internal/authz/authz.go
```diff
@@ -222,6 +308,45 @@ func (a *Authorizer) individualAuthorize(ctx context.Context, clientID, relation
 	return nil
 }
 
+// moduleAuthorize checks if the user has access to each of the modules, and exits if an error is encountered.
+func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, storeID string, modules []string) error {
```
adriantam, line 312: Since these involves go func - ideally, we have unit test on this function (with verification on no go-leak)
```
	t.Cleanup(func() {
		goleak.VerifyNone(t)
	})
```

pkg/server/server_authz_test.go
```diff
@@ -615,6 +700,91 @@ func TestCheckAuthz(t *testing.T) {
...
+			err := openfga.checkWriteAuthz(ctx, &openfgav1.WriteRequest{
+				StoreId: settings.testData.id,
+				Deletes: &openfgav1.WriteRequestDeletes{
+					TupleKeys: []*openfgav1.TupleKeyWithoutCondition{
+						{Object: "workspace:2", Relation: "guest", User: "user:jon"},
+					},
+				},
+			}, typesys)
+			require.Error(t, err)
+			require.Equal(t, "rpc error: code = Code(403) desc = the principal is not authorized to perform the action", err.Error())
```
adriantam, line 767: nitpik (non-blocking): You don't need to do two assert.  You can just call

require.Equal(t, status.Error(codes.Code..... ), err).

internal/authz/authz_test.go
```diff
@@ -179,3 +311,198 @@ func TestIndividualAuthorize(t *testing.T) {
...
+
+	t.Run("error_when_first_module_errors", func(t *testing.T) {
+		mockServer.EXPECT().Check(gomock.Any(), gomock.Any()).Return(&openfgav1.CheckResponse{Allowed: false}, nil)
+		mockServer.EXPECT().Check(gomock.Any(), gomock.Any()).Return(&openfgav1.CheckResponse{Allowed: true}, nil)
+		mockServer.EXPECT().Check(gomock.Any(), gomock.Any()).Return(&openfgav1.CheckResponse{Allowed: true}, nil)
+
+		err := authorizer.moduleAuthorize(context.Background(), "client-id", CanCallWrite, "store-id", []string{"module1", "module2", "module3"})
+		require.Error(t, err)
+		require.Equal(t, "rpc error: code = Code(403) desc = the principal is not authorized to perform the action", err.Error())
```
adriantam, line 339: nitpik (non-blocking): You don't need to do two assert. You can just call

require.Equal(t, status.Error(codes.Code..... ), err).

adriantam (approved):

Issue #1394, opened 2024-02-26
do not rate-limit in-memory reads of tuples


miparnisari:
When we call `Read` on a `boundedConcurrencyTupleReader` we are effectively doing this:

```go
       b.waitForLimiter(ctx)

	defer func() {
		<-b.limiter
	}()

	return b.RelationshipTupleReader.Read(ctx, store, tupleKey)
```

but because of the composition of functions, `RelationshipTupleReader` is `combinedTupleReader` (which reads from the database + contextual tuples). So this becomes

```go
    b.waitForLimiter(ctx)

	defer func() {
		<-b.limiter
	}()

	iter1 := storage.NewStaticTupleIterator(filterTuples(c.contextualTuples, tk.Object, tk.Relation))

	iter2, err := c.RelationshipTupleReader.Read(ctx, storeID, tk)
	if err != nil {
		return nil, err
	}

	return storage.NewCombinedIterator(iter1, iter2), nil
```

so I think in practice we are rate-limiting reading of things that are in memory.

The composition of functions should be this

```go
storagewrappers.NewCombinedTupleReader(
    storagewrappers.NewBoundedConcurrencyTupleReader(
```

_Originally posted by @miparnisari in https://github.com/openfga/openfga/pull/1375#discussion_r1493349447_

miparnisari:
Superseded by https://github.com/openfga/openfga/issues/2058

PR #1924, opened 2024-09-12
feat: cache tuples found while resolving check

jpadilla:
## Review Checklist


internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,162 @@
...
+type cachedDatastore struct {
+	storage.OpenFGADatastore
+
+	cache        storage.InMemoryCache[any]
+	maxCacheSize int64
+	ttl          time.Duration
+}
+
+// NewCachedDatastore returns a wrapper over a datastore...
+func NewCachedDatastore(inner storage.OpenFGADatastore, cache storage.InMemoryCache[any], maxSize int64, ttl time.Duration) *cachedDatastore {
```
miparnisari, line 30: Please re-use the existing https://github.com/openfga/openfga/blob/5bc7ebb087e56641fccb7027845d3141feacbcff/pkg/storage/storagewrappers/caching.go#L18-L27

jpadilla: I wanted to explicitly target storage usage from Check only. For example, Expand API uses `Read()`. Will think a bit to see if I can come up with an alternative.

miparnisari: Ohhh I see. Yeah then maybe we can keep them separate.

senojj: This is something more like `WrapDatastore()`.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,162 @@
...
+	cacheKey := fmt.Sprintf("%s/%s#%s", store, filter.Object, filter.Relation)
+	// todo(jpadilla): stable sort filter.AllowedUserTypeRestrictions
+	for _, userset := range filter.AllowedUserTypeRestrictions {
+		if _, ok := userset.GetRelationOrWildcard().(*openfgav1.RelationReference_Relation); ok {
+			cacheKey += "/" + userset.GetType() + ":%#" + userset.GetRelation()
+		}
+		if _, ok := userset.GetRelationOrWildcard().(*openfgav1.RelationReference_Wildcard); ok {
+			cacheKey += "/" + userset.GetType() + ":*"
+		}
+	}
```
senojj, line 54: for efficient string building use a [strings.Builder](https://pkg.go.dev/strings#Builder)

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,162 @@
...
+	cacheKey := cacheKeyFor(key)
+	cachedResp := c.cache.Get(cacheKey)
+	isCached := cachedResp != nil && !cachedResp.Expired && cachedResp.Value != nil
+	if isCached {
+		return storage.NewStaticTupleIterator(cachedResp.Value.([]*openfgav1.Tuple))
+	}
+
+	return &cachedIterator{
+		iter:         iter,
+		tks:          make([]*openfgav1.Tuple, 0, c.maxCacheSize),
```
senojj, line 78: I would suggest setting the length of the slice equal to `maxCacheSize` and using a variable to track the current index position. Accessing a slice by index is far more efficient than using `append()` to populate a slice of known length.

This will give your type better compute performance, and will require the same amount of memory. (setting the capacity allocates a backing array of the given size).

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,162 @@
...
+			for {
+				_, err := c.Next(context.Background())
+				if err != nil {
+					c.flush()
+					break
+				}
+			}
+
+			c.iter.Stop()
+		}()
```
senojj, line 138: what synchronizes this? wondering how we are preventing go routine leaks. while this operation may be benign when orphaned, we shouldn't allow orphaned routines in practice.

internal/graph/storagewrapper.go
miparnisari: Ufff we can apply this optimization pretty much everywhere in the codebase then 😅

internal/graph/storagewrapper.go
miparnisari, line 1: can we move this out of `internal/graph` and rename? Maybe `pkg/storagewrapper/querycache`?

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func (c *cachedDatastore) ReadUsersetTuples(ctx context.Context, store string, filter storage.ReadUsersetTuplesFilter, options storage.ReadUsersetTuplesOptions) (storage.TupleIterator, error) {
+	iter, err := c.OpenFGADatastore.ReadUsersetTuples(ctx, store, filter, options)
+	if err != nil {
+		return nil, err
+	}
+
+	cacheKey := fmt.Sprintf("%s/%s#%s", store, filter.Object, filter.Relation)
```
miparnisari, line 45: FYI, no action item: we have to be _super_ careful on what the cache keys are. If by mistake say we forgot to include `store` as cache key, we would giving the wrong tuples to some requests if there is a match on the other parts of the key.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,162 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/cespare/xxhash/v2"
```
senojj, line 11: what was the driving characteristic of xxhash that drove the decision to use it over something in the Go stdlib?

internal/graph/storagewrapper.go
miparnisari: It just occurred to me: do we have data to prove that calls to Expand are not immediately followed by Checks, or vice-versa? Because if so, it would actually be beneficial for all Check/ListUsers/ListObjects/Expand/Read calls to use the same cache.

internal/graph/storagewrapper.go
jpadilla: @senojj ah the only downside i see with that is either we possibly cache partially populated slices or _waste_ cycles filtering out all the nil values. `maxCacheSize` is just the possible ceiling, won't know the actual size of the slice at this point.

internal/graph/storagewrapper.go
jpadilla: unsure why we decided on this back when it was introduced for [`CheckRequestCacheKey`](https://github.com/openfga/openfga/blob/a2df7810d7faf1a34b21e5abecb7472da54b28d3/internal/graph/cached_resolver.go#L195)

we could benchmark an alternative non-crypto hash for sure

internal/graph/storagewrapper.go
senojj: You would know the the current position in the slice because you would keep track of it with a variable.
```go
var position int
```
You wouldn't need to filter out any `nil` values because you would just slice the slice when you flush.
```go
c.tks[:position]
```

senojj: Also note that you already enforce that you only ever cache up to `maxCacheSize` in the `Next()` method:
```go
if len(c.tks) < int(c.maxCacheSize) {
	c.tks = append(c.tks, tuple)
}
```
Therefore, at this time it is an enforced ceiling, not just a possible ceiling. However, I do also question if this could result in possible partial results being cached.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,246 @@
...
+	} else {
+		dsIter, err := dsIterFunc(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		cachedIter.iter = dsIter
+	}
+
+	span.SetAttributes(attribute.Bool("is_iter_cached", cachedIter.isCached))
```
elbuo8, line 156: What about the case for `if cacheKey == "" {`?

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,269 @@
...
+						c.flush()
+					}
+					return
+				}
+				// if the size is exceeded we don't add anymore and exit
+				if !c.addToBuffer(t) {
+					return
+				}
+			}
+		}()
```
senojj, line 242: If we are effectively closing the `cachedIterator` here, why do we need to drain the internal iterator? could we instead just dispose of any allocations in an efficient manner that could be done synchronously?

senojj: oh, is the cache reused across instances?

senojj: one thing to note is that after `Stop` completes, the caller will likely assume that the cache is in a stable state, however it is not, and may be modified at some undeterminable point in the future by a dangling go routine.

internal/graph/storagewrapper.go
jpadilla: @miparnisari it could definitely be yeah, but want to keep _this_ change strictly scoped to Check.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		// prevent goroutine if iterator was already consumed
+		ctx := context.Background()
+		if _, err := c.iter.Head(ctx); errors.Is(err, storage.ErrIteratorDone) {
+			c.flush()
+			c.iter.Stop()
+			return
+		}
+
+		c.wg.Add(1)
```
adriantam, line 232: Can you explain the purpose of the `wg`?  I don't see c.wg.Wait() anywhere?

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,281 @@
...
+func (c *cachedIterator) Head(ctx context.Context) (*openfgav1.Tuple, error) {
+	return c.iter.Head(ctx)
+}
+
+func (c *cachedIterator) flush() {
+	if c.tuples == nil {
+		return
+	}
+	tuples := make([]*openfgav1.Tuple, len(c.tuples))
+	copy(tuples, c.tuples)
```
adriantam, line 268: Why do we need to do the copy?

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,281 @@
...
+
+	c.addToBuffer(t)
+
+	return t, nil
+}
+
+// Stop see [Iterator.Stop].
+func (c *cachedIterator) Stop() {
+	c.closeOnce.Do(func() {
+		if c.tuples == nil {
```
adriantam, line 220: nitpik: probably good to call out there is a difference between c.tuples == nil and c.tuples == []*openfgav1.Tuple{} where 
- the first case indicates error and should not be cached
- the second case is not an error and we should store the empty result.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	return true
+}
+
+// Next see [Iterator.Next].
+func (c *cachedIterator) Next(ctx context.Context) (*openfgav1.Tuple, error) {
+	t, err := c.iter.Next(ctx)
+	if err != nil {
+		if !errors.Is(err, storage.ErrIteratorDone) && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
+			c.tuples = nil // don't store results that are incomplete
```
adriantam, line 207: Question: why do we want to store in the case of context.canceled or context.deadlineexceeded?  These are error cases where we don't have the complete result.  IMO, we probably want to be on the safe side and discard the cached result.

elbuo8: If the request resolves before finishing the iterator, it will receive cancelled event from the upstream HTTP context. This will happen to the majority of iterators, but we still want to cache the in the background. It won't be a partial result as on `.Close` we continue draining with a different context, not the parent associated with the Check (in this case) resolution.

internal/graph/storagewrapper.go
elbuo8: In the unit tests. If you have a solution for this please advice.

internal/graph/storagewrapper.go
elbuo8: Want to allow `cachedIterator` to be GC'ed since otherwise the cache will hold the pointer. I could be wrong in this but I think it increases memory and GC backpressure.

pkg/server/server.go
```diff
@@ -604,15 +629,26 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
 	}
 
+	var checkCacheOptions []graph.CachedCheckResolverOpt
+	if s.cacheLimit > 0 {
+		s.cache = storage.NewInMemoryLRUCache([]storage.InMemoryLRUCacheOpt[any]{
+			storage.WithMaxCacheSize[any](int64(s.cacheLimit)),
+		}...)
+	}
+
+	if s.checkQueryCacheEnabled {
```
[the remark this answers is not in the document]
elbuo8: Ah true, we should only enable these settings of `s.cache` exists.

internal/graph/storagewrapper.go
adriantam: probably good to add this as comment

internal/graph/storagewrapper.go
adriantam: In this case, we rely on an implicit behaviour that Next() MUST NOT advance / discard data upon context timeout/deadline.  Perhaps we need to document this expectation in the iterator interface?

Another possibility is to check for context deadline/cancel prior to line 204.  This will avoid the implicit behaviour dependency.

adriantam: On a side note, we should unit test this line (i.e., next yielding error) - we can use mock iterator to do that.

internal/graph/storagewrapper.go
adriantam: Maybe at least add comment in line 186?

jpadilla: Added a godoc on the struct field, it's only used for synchronization in testing.

internal/graph/storagewrapper.go
jpadilla: oh yeah, i think we're not handling empty responses.

elbuo8: We are caching empty, so it becomes a hit.

jpadilla: added some tests to just to confirm expectation of caching empty responses

internal/graph/storagewrapper.go
jpadilla: added comments

pkg/server/server.go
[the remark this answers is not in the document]
jpadilla: fixed

internal/graph/storagewrapper.go
jpadilla: > On a side note, we should unit test this line (i.e., next yielding error) - we can use mock iterator to do that.

added this test

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,340 @@
...
+			c.iter.Stop()
+			return
+		}
+
+		c.wg.Add(1)
+		go func() {
+			defer c.iter.Stop()
+			defer c.wg.Done()
+			// prevent draining on the same iterator across multiple requests
+			_, _, _ = c.sf.Do(c.cacheKey, func() (interface{}, error) {
```
miparnisari, line 276: i'm confused about why we need the `sf` object if we already had the `closeOnce` object in line 258? is that not enough to guarantee that this whole function is only called once?

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		return nil, err
+	}
+
+	c.addToBuffer(t)
+
+	return t, nil
+}
+
+// Stop see [Iterator.Stop].
```
miparnisari, line 256: could you improve this to say that this function runs the iterator til the very end in the background?

miparnisari (commented):
could we turn on this flag on all the big tests? I.e. `TestCheckMemory`, `TestCheckMySQL`, etc? since those tests are usually good at catching race issues

internal/graph/check.go
```diff
@@ -1241,7 +1241,6 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		if err != nil {
 			return nil, err
 		}
-		defer iter.Stop()
```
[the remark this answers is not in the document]
elbuo8: Because its duplicate. It's handled below properly.

internal/graph/storagewrapper_test.go
miparnisari, line 1: could we add unit tests where concurrency is exercised? e.g. what happens if two identical requests for Check come in, does the cache iterator react properly to that?

internal/graph/storagewrapper.go
jpadilla: updated comments for .Next() and .Stop()

cmd/run/flags.go
```diff
@@ -189,12 +189,19 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
 		util.MustBindPFlag("listUsersMaxResults", flags.Lookup("listUsers-max-results"))
 		util.MustBindEnv("listUsersMaxResults", "OPENFGA_LIST_USERS_MAX_RESULTS", "OPENFGA_LISTUSERSMAXRESULTS")
 
+		// TODO: make breaking change for cache limit
```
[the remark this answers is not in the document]
elbuo8: To avoid breaking existing OpenFGA deployments with this release the cache size (which is now shared with subproblems and iterators) will still use its original flag. But the flag should be renamed in subsequent releases.

internal/graph/storagewrapper.go
elbuo8: There may be 2 instances of the iterator. Say A and B, which both are working with the same `cacheKey`. `closeOnce` doesn't prevent A from closing B or vice versa (as intended). But it would be pointless for both A and B to drain the same iterator (based on the assumption of sharing the same cacheKey). So only 1 of them needs to drain and store in cache.

pkg/server/server.go
```diff
@@ -604,15 +629,26 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
...
 	}
 
+	var checkCacheOptions []graph.CachedCheckResolverOpt
+	if s.cacheLimit > 0 {
+		s.cache = storage.NewInMemoryLRUCache([]storage.InMemoryLRUCacheOpt[any]{
+			storage.WithMaxCacheSize[any](int64(s.cacheLimit)),
+		}...)
+	}
+
+	if s.cache != nil && s.checkQueryCacheEnabled {
```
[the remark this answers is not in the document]
elbuo8: Could still be used for the iterator cache. It's no longer used exclusively for sub problems.

And in the worst case, it's still nil allocation as there would be 0 entries in it.

pkg/server/server.go
```diff
@@ -641,7 +682,12 @@ func (s *Server) Close() {
 	}
 
 	s.checkResolverCloser()
+
+	if s.cache != nil {
+		s.cache.Stop()
+	}
```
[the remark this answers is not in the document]
elbuo8: No need. `checkDatastore` is a child of `datastore` so closing the parent is enough.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,346 @@
...
+
+	if isCached {
+		tuplesCacheHitCounter.Inc()
+		span.SetAttributes(attribute.Bool("cached", true))
+		return storage.NewStaticTupleIterator(cachedResp.Value.([]*openfgav1.Tuple)), nil
+	}
+
+	iter, err := dsIterFunc(ctx)
+	if err != nil {
+		return nil, err
```
elbuo8, line 198: Minor test to add?

internal/graph/storagewrapper_test.go
elbuo8: It doesn't have to react to anything. If it's in the cache it will use that, otherwise both will read from the datastore. The only thing that could be tested is for preventing both from being drained once Stop is called, which is a small window.

internal/graph/storagewrapper.go
```diff
@@ -0,0 +1,346 @@
...
+	return &CachedDatastore{
+		OpenFGADatastore: inner,
+		cache:            cache,
+		maxResultSize:    maxSize,
+		ttl:              ttl,
+	}
+}
+
+// ReadUsersetTuples see [storage.RelationshipTupleReader].ReadUsersetTuples.
+func (c *CachedDatastore) ReadUsersetTuples(
```
miparnisari, line 92: I probably missed something but why do we not also have implementation for `ReadStartingWithUser`? will it be a follow-up PR?

miparnisari:
> could we turn on this flag on all the big tests? I.e. TestCheckMemory, TestCheckMatrix, etc? since those tests are usually good at catching race issues

@jpadilla 😄

internal/graph/storagewrapper.go
jpadilla: wired this up correctly in https://github.com/openfga/openfga/pull/1924/commits/150f7da03e4b981b55efce8997ec3597930931eb

pkg/server/server.go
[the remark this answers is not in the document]
jpadilla: tweaked this a bit and added some tests for the different combinations https://github.com/openfga/openfga/pull/1924/commits/d6a0e5978d4e362f0128c2b3d6dead5cf24c9bce

internal/graph/storagewrapper.go
jpadilla: Not for now. I think the cardinality of `ReadStartingWithUser` is not as low to benefit from this iteration. In future work, it might.

pkg/server/server.go
```diff
@@ -952,7 +998,7 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 
 	const methodName = "check"
 	resp, checkRequestMetadata, err := commands.NewCheckCommand(
-		s.datastore,
+		s.checkDatastore,
```
miparnisari, line 1001: This change won't work properly.  It's related to https://github.com/openfga/openfga/issues/1394.

When you inspect the value of `datastore` inside the request, the chain of objects is 

image

the `BoundedConcurrencyTupleReader` will rate-limit reads to the cache, which will harm performance. 

The correct order should be: `cachedDs` wraps a `combinedDs` wraps a `boundedConcurrencyDs` wraps `ds`.

I think the right way to do is is to **not** have the server keep a reference to the `cachedDs` but rather the thing that we _actually_ need to share across the requests, which is the underlying cache. Then, each call to Check creates its own `datastore object` chain of objects.

jpadilla: @miparnisari I'd prefer we address that separate to these changes.

miparnisari (approved):

PR #1926, opened 2024-09-12
Extend request timeout to 10s for slow tests

DanCech:
This PR extends the RequestTimeout to 10s for some of the tests. We run tests with code coverage enabled which can make them take longer than the default 3s timeout.

This appears to impact SQLite more than it does Postgres or MySQL, which may be due to the fact that the entirety of SQLite is pure go running in-process, vs just the client for the other databases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 * When a request gets cancelled by a client, throw a 4xx, not a 5xx. [#1905](https://github.com/openfga/openfga/pull/1905)
 * Makes the `pkg.logger.Logger.With` immutable by creating a child logger instead of mutating the delegate one to prevent side effects [1906](https://github.com/openfga/openfga/pull/1906)
+* Extend request timeout to 10s for slow tests [1926](https://github.com/openfga/openfga/pull/1926)
 
 ## Performance
 
--- a/tests/listobjects/listobjects_test.go
+++ b/tests/listobjects/listobjects_test.go
@@ -35,6 +35,7 @@ func testRunAll(t *testing.T, engine string) {
 	cfg := config.MustDefaultConfig()
 	cfg.Log.Level = "error"
 	cfg.Datastore.Engine = engine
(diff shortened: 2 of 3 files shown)


tests/listobjects/listobjects_test.go
```diff
@@ -34,6 +35,9 @@ func testRunAll(t *testing.T, engine string) {
 	cfg := config.MustDefaultConfig()
 	cfg.Log.Level = "error"
 	cfg.Datastore.Engine = engine
+	cfg.ListObjectsDeadline = 0 // no deadline
```
DanCech, line 38: We could also just increase this to 10s, I followed the pattern in check_test.go of disabling the deadline altogether

jpadilla (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 
 * When a request gets cancelled by a client, throw a 4xx, not a 5xx. [#1905](https://github.com/openfga/openfga/pull/1905)
 * Makes the `pkg.logger.Logger.With` immutable by creating a child logger instead of mutating the delegate one to prevent side effects [1906](https://github.com/openfga/openfga/pull/1906)
+* Extend request timeout to 10s for slow tests [1926](https://github.com/openfga/openfga/pull/1926)
```
miparnisari, line 24: Nitpick: we don't include things related to tests in changelog 😅 (or refactors, or cleanups. if it doesn't impact customers directly we don't mention it)

tests/listobjects/listobjects_test.go
miparnisari, line 1: Please also update `listusers_test.go` :)

miparnisari: Nvm i'll do it :)

CHANGELOG.md
DanCech: Good to know, I added it because ci was complaining about there not being an entry, is there a "no-changelog" label to bypass that?

DanCech: Ah I see, Skip-Changelog 👍

PR #1927, opened 2024-09-12
chore: make changelog kacl compliant

jpadilla:
## Description

Enforce a consistent format for our changelog by leveraging https://www.npmjs.com/package/keep-a-changelog

## Review Checklist


package.json
```diff
@@ -0,0 +1,11 @@
...
+  "name": "openfga",
+  "version": "0.0.0",
+  "private": true,
+  "description": "A high performance and flexible authorization/permission engine built for developers and inspired by Google Zanzibar ",
+  "license": "Apache-2.0",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/openfga/openfga.git"
+  }
+}
```
willvedd, line 11: It's a real shame that this CLI requires a `package.json` file at all. It's required because it needs to determine the repo url. In my testing I discovered that having the below was suffificent:

```suggestion
{
  "_": "package.json file exists only for kacl changelog linting",
  "repository": {
    "url": "https://github.com/openfga/openfgxa"
  }
}
```

But would be better if we didn't have that at all. Especially because AFAICT, it isn't even being used.

willvedd (commented):
First of all, this PR makes me happy because it'll make releases easier!

I pulled this and tested a few things and first of all, its a shame that we need to have a `package.json` file but that's not the worst thing ever. If we wanted to leverage the `kacl` CLI for automating releases, we would need to start tracking the application version in that file. So I'm not sure how we feel about that. 

Further, the validations that this CLI provides actually isn't that extensive. AFAICT, it enforces the headings. Which is definitely worth something but won't save us from making some egregious mistakes.

elbuo8 (approved):

PR #1936, opened 2024-09-17
fix: handle all permutations of sqlite busy/locked errors

DanCech:
There are a number of different busy/locked errors that SQLite can return, this PR refines `busyRetry` to detect them all properly.

Since we are no longer wrapping functions that call `sqlcommon.HandleSQLError` within `busyRetry`, it also removes the hack for detecting those "handled" errors via string matching.


miparnisari (approved):

miparnisari:
@DanCech thanks for sending this. Could you please update the CHANGELOG to mention this fix?

miparnisari (approved):

miparnisari:
@DanCech please ping me on Slack when you are ready to merge `main` into this, then after I confirm you can go ahead with the update, and then I can merge asap.

PR #1938, opened 2024-09-17
Switch to local HandleSQLError function for SQLite

DanCech:
This PR allows us to remove the special handling for SQLite from sqlcommon by switching to a local implementation that just handles the SQLite errors.

This also removes `dbInfo` from SQLite since it is never used.

This approach won't work for MySQL or Postgres today since they do rely on the sqlcommon functions, but it may be possible to pass a handler in via dbInfo for those cases, or return the raw errors from sqlcommon and move the error translation out into the storage implementations.

### DIFF
--- a/pkg/storage/sqlite/sqlite.go
+++ b/pkg/storage/sqlite/sqlite.go
@@ -36,7 +36,6 @@ var tracer = otel.Tracer("openfga/pkg/storage/sqlite")
 type SQLite struct {
 	stbl                   sq.StatementBuilderType
 	db                     *sql.DB
-	dbInfo                 *sqlcommon.DBInfo
 	logger                 logger.Logger
 	dbStatsCollector       prometheus.Collector
 	maxTuplesPerWriteField int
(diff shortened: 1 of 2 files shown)


miparnisari (approved):

pkg/storage/sqlite/sqlite.go
```diff
@@ -1078,6 +1076,30 @@ func (m *SQLite) IsReady(ctx context.Context) (storage.ReadinessStatus, error) {
 	return sqlcommon.IsReady(ctx, m.db)
 }
 
+func HandleSQLError(err error, args ...interface{}) error {
+	if errors.Is(err, sql.ErrNoRows) {
+		return storage.ErrNotFound
+	}
+
+	if errors.Is(err, storage.ErrIteratorDone) {
+		return err
```
miparnisari, line 1085: looks like this is never hit and can be removed :) the iterator is the only struct that returns this err

miparnisari (approved):

PR #1943, opened 2024-09-18
chore: use backend-specific HandleSQLError implementations

### LINKED ISSUE #1709: Custom sql error handler for common database operations
### Checklist

Hey, we are experimenting with embedding open fga inside Grafana. OpenFGA already have support for Mysql and Postgres datastores but we must support sqlite on top of these. We made a [pr](https://github.com/openfga/openfga/pull/1615) to add sqlite. What I gathered from that is there will probably exist some kind of plugin system in the future where should add sqlite instead of  merging it into main.

This is not a blocker for us because we can just implement this data store directly in Grafana. This was pretty straight forward and it was nice to be able to re-use the [test suite](https://github.com/openfga/openfga/blob/main/pkg/storage/test/storage.go#L29) for stored. The one problem we are facing is `sqlcommon.HandleSQLError`. We need to special case duplication errors from sqlite similar to how it is done for both progres and mysql today. This works fine for most db operations except common operations provided by `sqlcommon` package such as [Write](https://github.com/openfga/openfga/blob/main/pkg/storage/sqlcommon/sqlcommon.go#L330). Here HandleSQLError is called directly and AFAIK there is not way to override the behavior. 

It would be good if we could pass a custom error handler `func(err error, args ...any) error` to common database operations provided by this package. For mysql and postgres provided by openFGA this would be `sqlcommon.HandleSQLError`.

The workaround we have right now is to [copy](https://github.com/grafana/grafana/pull/89486/files#diff-59de55e7bb801b29e7827713f84603133199bba04e4b853f046e34f64c90357eR19) `sqlcommon.Write` function and handle potential errors with out custom handler. 

### Additional context

_No response_

DanCech:
This PR migrates from a central SQL error handler in sqlcommon to separate handlers in the mysql and postgres storage adapters.

To support cases where they call back into sqlcommon, the error handler is provided as part of DBInfo.

Fixes #1709


pkg/storage/mysql/mysql.go
```diff
@@ -111,7 +111,7 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*MySQL, error) {
 	}
 
 	stbl := sq.StatementBuilder.RunWith(db)
-	dbInfo := sqlcommon.NewDBInfo(db, stbl, sq.Expr("NOW()"))
+	dbInfo := sqlcommon.NewDBInfo(db, stbl, sq.Expr("NOW()"), HandleSQLError)
```
miparnisari, line 114: What do you think if we rename this function `HandleMySQLError`, then for postgres `HandlePostgresError`?

This is based on this article about code searchability: https://morizbuesing.com/blog/greppability-code-metric/

DanCech: I'm not sure about other IDEs but at least in vscode it's able to distinguish between them and show only the correct one when searching for references. Even if you are grepping for strings any references outside the module would be `mysql.HandleSQLError` or `postgre.HandleSQLError` so I tend to think that keeping the same name across them makes more sense as they are implementing the same callback.

The other argument for keeping them the same is that the 2 implementations are very close to identical, If you compare mysql.go to postgres.go there are actually far more diffs due to "noise" from things like `m` vs `p` when defining methods than from actual differences in logic. I'd argue that we'll have an easier time maintaining these 2 backends if we reduce that (and would be happy to put together a PR) rather than adding to it by having a diff in every pl;ace that we call `HandleSQLError`.

miparnisari (approved):

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -594,7 +570,7 @@ func constructAuthorizationModelFromSQLRows(rows *sql.Rows) (*openfgav1.Authoriz
 		var marshalledModel []byte
 		err := rows.Scan(&modelID, &schemaVersion, &typeName, &marshalledTypeDef, &marshalledModel)
 		if err != nil {
-			return nil, HandleSQLError(err)
+			return nil, err
```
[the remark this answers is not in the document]
miparnisari: Tracking via https://github.com/openfga/openfga/issues/1952

Issue #1945, opened 2024-09-19
[Performance] improve check latency for recursive userset and tuple to userset


adriantam:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Currently Check API has high latency for userset and tuple-to-userset that reference themselves.

For example, `check(user:X, member, group:5)` for the following model

```
model
  schema 1.1
type user
type group
  relations
    define member: [user, group#member]
```

and the following tuples 
- `group:2#member@group:1#member` 
- `group:3#member@group:2#member` 
- `group:4#member@group:3#member`
- `group:5#member@group:4#member`

requires:
- 4 separate datastore query for `user:X` 
- 4 separate datastore query for `group:1`, `group2`

with a total of 8 datastore lookup.  

### Describe the ideal solution

Ideally, we only need 
- 1 datastore query for user:X for all objects with type `group`
- 4 datastore query of group.   
 
This reduces latency and datastore query cost.

PR #1946, opened 2024-09-19
chore: identify recursive userset model that can be optimized

adriantam:
## Description
 Create an internal method that identify recursive userset that can be optimized.

## References

## Review Checklist


adriantam:
Note that we will not add changelog until the optimization is done

pkg/typesystem/typesystem.go
```diff
@@ -165,13 +167,24 @@ type TypeSystem struct {
...
+	schemaVersion           string
+	authorizationModelGraph *graph.AuthorizationModelGraph
 }
 
 // New creates a *TypeSystem from an *openfgav1.AuthorizationModel.
 // It assumes that the input model is valid. If you need to run validations, use NewAndValidate.
 func New(model *openfgav1.AuthorizationModel) *TypeSystem {
+	ts, err := newTypesystem(model)
+	if err != nil {
+		panic(err)
```
elbuo8, line 180: We should change this to handle the error rather than panic.

go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/natefinch/wrap v0.2.0
 	github.com/oklog/ulid/v2 v2.1.0
 	github.com/openfga/api/proto v0.0.0-20240906203051-102620ef2a66
-	github.com/openfga/language/pkg/go v0.2.0-beta.2
+	github.com/openfga/language/pkg/go v0.2.0-beta.2.0.20240918205720-9b4b2742d552
```
adriantam, line 28: Will be updated when https://github.com/openfga/language/pull/352 is merged

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +387,59 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+			}
+		}
+	}
+	return hasNodeBackToItself && hasCorrectTerminalType
+}
+
+// RecursiveUsersetCanFastPath returns whether the specified object type and relation allows
+// for optimization.
+func (t *TypeSystem) RecursiveUsersetCanFastPath(objectTypeRelation string, userType string) bool {
+	nodes := t.authorizationModelGraph.Nodes()
```
elbuo8, line 427: Is there a way this could be faster? We are basically iterating through a list every time we ask this. We could memoize it or perhaps pre compute this. 

Doesn't need to be addressed in this PR, but this would get linearly worst for all the nodes that are added to the graph, which we add for not just types.

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +387,59 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+			// Strange, I would have expected to be able to cast to AuthorizationModelNode
+			return false
+		}
+		childNodeObjectType, childNodeRelation := tuple.SplitObjectRelation(curChildAuthorizationModelNode.Label())
+		if childNodeRelation == "" {
+			// this means the childNode is either a terminal type or one of relation keyword
+			if childNodeObjectType == userType {
+				hasCorrectTerminalType = true
+			} else if childNodeObjectType == "union" || childNodeObjectType == "intersection" || childNodeObjectType == "exclusion" {
+				// they are not terminal type and are not allowed
```
elbuo8, line 410:
```suggestion
				// childNode is a set operator
```

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +387,59 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+		if !ok {
+			// Strange, I would have expected to be able to cast to AuthorizationModelNode
+			return false
+		}
+		childNodeObjectType, childNodeRelation := tuple.SplitObjectRelation(curChildAuthorizationModelNode.Label())
+		if childNodeRelation == "" {
+			// this means the childNode is either a terminal type or one of relation keyword
+			if childNodeObjectType == userType {
+				hasCorrectTerminalType = true
+			} else if childNodeObjectType == "union" || childNodeObjectType == "intersection" || childNodeObjectType == "exclusion" {
```
elbuo8, line 409: This could perhaps be expressed as a set (`map[string]struct`) named setOperators and checking if its in the map.

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +387,59 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+	for childrenNodes.Next() {
+		curNode := childrenNodes.Node()
+		curChildAuthorizationModelNode, ok := curNode.(*graph.AuthorizationModelNode)
+		if !ok {
+			// Strange, I would have expected to be able to cast to AuthorizationModelNode
+			return false
+		}
+		childNodeObjectType, childNodeRelation := tuple.SplitObjectRelation(curChildAuthorizationModelNode.Label())
+		if childNodeRelation == "" {
+			// this means the childNode is either a terminal type or one of relation keyword
```
elbuo8, line 406: > or one of relation keyword

?

pkg/typesystem/typesystem_test.go
```diff
@@ -4411,6 +4419,164 @@ func TestHasTypeInfo(t *testing.T) {
...
+
+	tests := []struct {
+		name               string
+		model              string
+		objectTypeRelation string
+		userType           string
+		expected           bool
+	}{
+		{
+			name: "object_typerelation_not_found",
```
elbuo8, line 4433:
```suggestion
			name: "object_type_relation_not_found",
```

pkg/typesystem/typesystem.go
adriantam: Let's compute this at TypeSystem creation time.  This will allow us to avoid casting into *graph.AuthorizationModelNode a few lines later which does not help with performance.

pkg/typesystem/typesystem.go
adriantam: change to switch/case

pkg/typesystem/typesystem.go
adriantam: removed weird comment

pkg/typesystem/typesystem.go
```diff
@@ -165,13 +168,31 @@ type TypeSystem struct {
...
+	authorizationModelNodes := make(map[string]*graph.AuthorizationModelNode)
+	nodes := g.Nodes()
+	for nodes.Next() {
+		curNode := nodes.Node()
+		curAuthorizationModelNode, ok := curNode.(*graph.AuthorizationModelNode)
+		if !ok {
+			// Strange, I should have expected to be able to cast to AuthorizationModelNode.
+			return nil, fmt.Errorf("unable to cast as graph.AuthorizationModelNode. node ID: %v", curNode.ID())
+		}
+		authorizationModelNodes[curAuthorizationModelNode.Label()] = curAuthorizationModelNode
```
miparnisari, line 188: **Thought**: i'm kind of regretting calling `.Label()` like that. It's not clear enough. I think we should change it, but i'm not sure what to call it. WDYT?

pkg/typesystem/typesystem.go
```diff
@@ -207,15 +228,26 @@ func New(model *openfgav1.AuthorizationModel) *TypeSystem {
 			WithMaxEvaluationCost(config.MaxConditionEvaluationCost()).
 			WithInterruptCheckFrequency(config.DefaultInterruptCheckFrequency)
 	}
+	authorizationModelGraph, err := graph.NewAuthorizationModelGraph(model)
```
miparnisari, line 231: By default, the direction of this graph is for ListObjects (added docs for this here https://github.com/openfga/language/pull/353)

is this what we want? if not, could we add code to be 100% sure that it's like that?

```go
authorizationModelGraph, err := graph.NewAuthorizationModelGraph(model)
if err != nil {
	return err
}
if (authorizationModelGraph.drawingDirection != drawingDirectionListObjects) {
     authorizationModelGraph = authorizationModelGraph.Reversed()

}

File: pkg/typesystem/typesystem.go [hunk 10]
```diff
@@ -379,6 +411,52 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
 	return true
 }
 
+// recursiveUsersetNodeCanFastpath is a helper function to determine whether the node (object#relation) has
+// - one edge back to itself
+// - other edges lead directly to terminal types (not union/intersection/exclusion).
```
miparnisari, line 416: is `user:*` a terminal type?

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +411,52 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+func (t *TypeSystem) recursiveUsersetNodeCanFastpath(curAuthorizationModelNode *graph.AuthorizationModelNode, userType string) bool {
+	hasNodeBackToItself := false
+	hasCorrectTerminalType := false
+
+	// TODO: Optimize by memorizing the list of childrenNodes
+	childrenNodes := t.authorizationModelGraph.To(curAuthorizationModelNode.ID())
```
miparnisari, line 422: Nitpick: I prefer the naming `neighborNodes`. `childrenNodes` makes it sound like the graph is a tree, and it is not always a tree.

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +411,52 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+			case "union", "intersection", "exclusion":
+				return false
+			}
+		} else {
+			if curAuthorizationModelNode == curChildAuthorizationModelNode {
+				hasNodeBackToItself = true
+			} else {
+				return false
+			}
+		}
```
miparnisari, line 444:
```suggestion
		switch curChildAuthorizationModelNode.NodeType() {
		case graph.SpecificType:
			// TODO wildcard?? should we add a new NodeType to handle this more explicitly?
			if curChildAuthorizationModelNode.Label() == userType {
				hasCorrectTerminalType = true
			}
		case graph.OperatorNode:
			return false
		case graph.SpecificTypeAndRelation:
			if curAuthorizationModelNode == curChildAuthorizationModelNode {
				hasNodeBackToItself = true
			} else {
				return false
			}
		}
```

pkg/typesystem/typesystem.go
```diff
@@ -165,13 +168,31 @@ type TypeSystem struct {
 	// [objectType] => [relationName] => TTU relation.
 	ttuRelations map[string]map[string][]*openfgav1.TupleToUserset
 
-	modelID       string
-	schemaVersion string
+	modelID                 string
+	schemaVersion           string
+	authorizationModelGraph *graph.AuthorizationModelGraph
+	authorizationModelNodes map[string]*graph.AuthorizationModelNode
```
miparnisari, line 174: Dang it - when I wrote the code to build the graph I used a helper map for exactly this: be able to find in o(1) a node given it's label

https://github.com/openfga/language/blob/f3eb5182c6681154ebc23cc3d9a81307cb32cfcd/pkg/go/graph/graph_builder.go#L14-L18

I think we should expose this functionality as a method on the graph itself and not have two distinct fields here storing that information. I can do that for you 😄

pkg/typesystem/typesystem_test.go
```diff
@@ -4411,6 +4425,192 @@ func TestHasTypeInfo(t *testing.T) {
...
+		transformedModel := testutils.MustTransformDSLToProtoWithID(model)
+		authorizationModelGraph, err := graph.NewAuthorizationModelGraph(transformedModel)
+		require.NoError(t, err)
+		nodes, err := getAllAuthorizationModelNodes(authorizationModelGraph.DirectedGraph)
+		require.NoError(t, err)
+		keys := maps.Keys(nodes)
+		sort.Strings(keys)
+		require.Equal(t, []string{"group", "group#member", "user"}, keys)
+	})
+}
```
miparnisari, line 4454: Like I said in an earlier comment, this doesn't belong here, if you need to look up nodes by ID, then the graph should be kind enough to provide that info to you :) I'll send a PR to `language`

pkg/typesystem/typesystem_test.go
```diff
@@ -4411,6 +4425,192 @@ func TestHasTypeInfo(t *testing.T) {
...
+	}{
+		{
+			name: "object_type_relation_not_found",
+			model: `
+model
+	schema 1.1
+type user
+type group
+	relations
+		define member: [user, group#member]
```
miparnisari, line 4474: Please add these tests:

```
define member: [user with cond, group#member]
define member: [user:*, group#member]
define member: [user:* with cond, group#member]
define member: [user, user:*, group#member]
define member: [user with cond, user:*, group#member]
define member: [user, user:* with cond, group#member]
```

pkg/typesystem/typesystem.go
miparnisari: This belongs in the graph itself :D

pkg/typesystem/typesystem.go
adriantam: Good callout.  I wrote the logic with ListObjects order - we should assert that.

pkg/typesystem/typesystem.go
adriantam: user:* should be a terminal type.

pkg/typesystem/typesystem_test.go
adriantam: Once the change is made to language, I will nuke this :)

pkg/typesystem/typesystem.go
miparnisari: See https://github.com/openfga/language/pull/354

pkg/typesystem/typesystem.go
adriantam: See https://github.com/openfga/language/pull/353#discussion_r1769162266, we need the method to expose the drawing direction as this is a private member.

miparnisari: Done and merged :)

pkg/typesystem/typesystem.go
```diff
@@ -379,6 +395,59 @@ func (t *TypeSystem) UsersetCanFastPath(relationReferences []*openfgav1.Relation
...
+	for neighbourNodes.Next() {
+		curNode := neighbourNodes.Node()
+		curNeighbourAuthorizationModelNode, ok := curNode.(*graph.AuthorizationModelNode)
+		if !ok {
+			// Note: this should not happen, but adding the guard nonetheless
+			return false
+		}
+		switch curNeighbourAuthorizationModelNode.NodeType() {
+		case graph.SpecificType:
+			// TODO wildcard?? should we add a new NodeType to handle this more explicitly?'
```
miparnisari, line 417: https://github.com/openfga/language/pull/356 :)

miparnisari (approved):

pkg/typesystem/typesystem.go
adriantam: Will do it in next PR

PR #1950, opened 2024-09-19
feat: add list stores authorization

Vic-Dev:
Adds support for ListStores to access control

## Description

Adds ListStores to the access control experimental feature. When a user calls `ListStores` with access control enabled, a list of storeIDs will be created based on their access. That list will be used to filter their ListStores results.

## References
https://github.com/openfga/openfga/pull/1913

## Review Checklist


pkg/storage/postgres/postgres.go
```diff
@@ -537,10 +537,17 @@ func (p *Postgres) ListStores(ctx context.Context, options storage.ListStoresOpt
 	ctx, span := tracer.Start(ctx, "postgres.ListStores")
 	defer span.End()
 
+	var whereClause sq.Eq
+	if len(*options.IDs) > 0 {
+		whereClause = sq.Eq{"id": *options.IDs}
```
rhamzeh, line 542: You need both here, so id in ids + not deleted

```go

whereClause = sq.And{
  sq.Eq{"deleted_at": nil},
  sq.Eq{"id": []string{"id1", "id2", "id2"}}
}
```

pkg/storage/storage.go
```diff
@@ -83,6 +83,7 @@ type ReadAuthorizationModelsOptions struct {
 // ListStoresOptions represents the options that can
 // be used with the ListStores method.
 type ListStoresOptions struct {
+	IDs        *[]string
```
rhamzeh, line 86: Note that @jpadilla has mentioned that he prefers that we don't pass them in the options and instead do a breaking change
```go
type ListStoresFilter {
  IDs *[]string
}

ListStores(ctx context.Context, filter ListStoresFilter, options ListStoresOptions) ([]*openfgav1.Store, []byte, error)
```

I think we can maybe get a PR in as is and fix it once a decision is made

internal/authz/authz.go
```diff
@@ -166,7 +168,47 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
...
+	}
+
+	// Disable authz check for the list objects request.
+	ctx = authclaims.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.ListObjects(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.GetObjects(), nil
```
rhamzeh, line 211: Note that you'll want to loop over the store and remove the `store:` prefix

pkg/storage/postgres/postgres.go
```diff
@@ -537,10 +537,20 @@ func (p *Postgres) ListStores(ctx context.Context, options storage.ListStoresOpt
 	ctx, span := tracer.Start(ctx, "postgres.ListStores")
 	defer span.End()
 
+	var whereClause sq.Sqlizer
+	if len(*options.IDs) > 0 {
```
miparnisari, line 541: I'm curious, is there a specific reason why `options.IDs` is a pointer? I think `if len(options.IDs) > 0` and `sq.Eq{"id": options.IDs},` should also work, and we save ourselves from a potential nil pointer exception elsewhere in the codebase

pkg/storage/storage.go
miparnisari: I'm fine either way, but please add godoc at the top of `IDs` that explains what this filter is for :)

internal/authz/authz.go
```diff
@@ -166,7 +169,54 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
...
+	ctx = authclaims.ContextWithSkipAuthzCheck(ctx, true)
+	resp, err := a.server.ListObjects(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	storeIDs := make([]string, len(resp.GetObjects()))
+	re := regexp.MustCompile(`^.*:(.*)`)
+	for i, store := range resp.GetObjects() {
+		id := re.FindStringSubmatch(store)
```
miparnisari, line 215: Suggestion for a future PR: when you pull `main` into your branch, you can get rid of the regex matching and use the helper method `ToUserParts`: https://github.com/openfga/openfga/blob/f3be13c8969feffa443eed902b3dc377a8b9b8f4/pkg/tuple/tuple_test.go#L579-L599

internal/authz/authz.go
```diff
@@ -166,7 +169,54 @@ func (a *Authorizer) AuthorizeCreateStore(ctx context.Context) error {
...
+
+	relation, err := a.getRelation(ListStores)
+	if err != nil {
+		return err
+	}
+
+	return a.individualAuthorize(ctx, claims.ClientID, relation, SystemIDType, &openfgav1.ContextualTupleKeys{})
+}
+
+// ListAuthorizedStores returns the list of stores that the user has access to.
```
miparnisari, line 190:
```suggestion
// ListAuthorizedStores returns the list of store IDs that the user has access to.
```

pkg/server/server.go
```diff
@@ -1564,3 +1569,18 @@ func (s *Server) checkCreateStoreAuthz(ctx context.Context) error {
 	}
 	return nil
 }
+
+// checkAuthzListStores checks the authorization for listing stores.
+func (s *Server) checkAuthzListStores(ctx context.Context) ([]string, error) {
```
miparnisari, line 1574:
```suggestion
// getAccessibleStores checks whether the caller has permission to list stores and if so, returns the list of stores that the user has access to.
func (s *Server) getAccessibleStores(ctx context.Context) ([]string, error) {
```

pkg/server/server.go
```diff
@@ -1464,11 +1464,16 @@ func (s *Server) ListStores(ctx context.Context, req *openfgav1.ListStoresReques
 		Method:  method,
 	})
 
+	storeIDs, err := s.checkAuthzListStores(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	q := commands.NewListStoresQuery(s.datastore,
```
miparnisari, line 1472: Suggestion: add a note here saying that even though we have the list of store IDs, we need to call ListStoresQuery to fetch the entire metadata of the store.

This was something not immediately obvious to me  :)

(Come to think of it, `ListStoresQuery` was a poor name choice, maybe DescribeStores would have been more descriptive..)

internal/authz/authz.go
rhamzeh: I am not sure I like that we're using a regex - 

can we do something like this instead:
```go
for index, store := range resp.GetObjects() {
storeIDs[index] = strings.TrimPrefix(store, `store:`)
```

rhamzeh (commented):
Note: you will need to merge in main and do the storage changes in the SQLite adapter too

pkg/storage/memory/memory.go
```diff
@@ -777,6 +777,18 @@ func (s *MemoryBackend) ListStores(ctx context.Context, options storage.ListStor
 		stores = append(stores, t)
 	}
 
+	if len(options.IDs) > 0 {
```
adriantam, line 780: Will be good to have unit test to test this (i.e., whether the filter is working properly).  Similar for the sql storage interfaces

adriantam (approved):
Reminder that when we merge in this feature, make sure we call out the storage interface change (i.e., ListStores now allow filter on store ID).

pkg/storage/memory/memory.go
Vic-Dev: This is tested:
https://github.com/openfga/openfga/pull/1950/files#diff-a26545677701a4b3a9713c2221c6e16d199561147b48f1ae839378db43e0fcd2R87-R95

PR #1953, opened 2024-09-20
chore: reduce diffs between storage implementations

DanCech:
This PR is an exploration of how we can improve the maintainability of the storage backend implementations. Its aim is to reduce the number of differences between the different backends, making it much easier to diff the files and see the actual differences without the noise of naming and formatting differences.

The major changes are:
- change the receiver alias for all methods to `s` (for Storage, could be anything really)
- change the name of the struct to `Datastore` across the board
- add a `startTrace` helper to avoid embedding the backend name in every call to `trace.Start`
- use `sq.Expr("NOW()")` for both MySQL and Postgres, this removes the need for `DBInfo.sqlTime`
- align whitespace and formatting
- switch sqlite to use `returning` when inserting a new store (like Postgres)
- don't reuse `createdAt` as `updatedAt` in `CreateStore`

These changes stop short of any major refactor to move more common code into a shared library, but do allow us to see how (very) similar the mysql and postgres implementations actually are.

```
--- pkg/storage/mysql/mysql.go  2024-09-20 15:10:41.706817267 -0400
+++ pkg/storage/postgres/postgres.go    2024-09-20 15:10:41.706817267 -0400
@@ -1,4 +1,4 @@
-package mysql
+package postgres

 import (
        "context"
@@ -6,12 +6,13 @@
        "encoding/json"
        "errors"
        "fmt"
+       "net/url"
        "strings"
        "time"

        sq "github.com/Masterminds/squirrel"
        "github.com/cenkalti/backoff/v4"
-       "github.com/go-sql-driver/mysql"
+       _ "github.com/jackc/pgx/v5/stdlib" // PostgreSQL driver.
        "github.com/prometheus/client_golang/prometheus"
        "github.com/prometheus/client_golang/prometheus/collectors"
@@ -28,13 +29,13 @@
        tupleUtils "github.com/openfga/openfga/pkg/tuple"
 )

-var tracer = otel.Tracer("openfga/pkg/storage/mysql")
+var tracer = otel.Tracer("openfga/pkg/storage/postgres")

 func startTrace(ctx context.Context, name string) (context.Context, trace.Span) {
-       return tracer.Start(ctx, "mysql."+name)
+       return tracer.Start(ctx, "postgres."+name)
 }

-// Datastore provides a MySQL based implementation of [storage.OpenFGADatastore].
+// Datastore provides a PostgreSQL based implementation of [storage.OpenFGADatastore].
 type Datastore struct {
        stbl                   sq.StatementBuilderType
        db                     *sql.DB
@@ -51,24 +52,37 @@
 // New creates a new [Datastore] storage.
 func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
        if cfg.Username != "" || cfg.Password != "" {
-               dsnCfg, err := mysql.ParseDSN(uri)
+               parsed, err := url.Parse(uri)
                if err != nil {
-                       return nil, fmt.Errorf("parse mysql connection dsn: %w", err)
+                       return nil, fmt.Errorf("parse postgres connection uri: %w", err)
                }

+               username := ""
                if cfg.Username != "" {
-                       dsnCfg.User = cfg.Username
+                       username = cfg.Username
+               } else if parsed.User != nil {
+                       username = parsed.User.Username()
                }
-               if cfg.Password != "" {
-                       dsnCfg.Passwd = cfg.Password
+
+               switch {
+               case cfg.Password != "":
+                       parsed.User = url.UserPassword(username, cfg.Password)
+               case parsed.User != nil:
+                       if password, ok := parsed.User.Password(); ok {
+                               parsed.User = url.UserPassword(username, password)
+                       } else {
+                               parsed.User = url.User(username)
+                       }
+               default:
+                       parsed.User = url.User(username)
                }

-               uri = dsnCfg.FormatDSN()
+               uri = parsed.String()
        }

-       db, err := sql.Open("mysql", uri)
+       db, err := sql.Open("pgx", uri)
        if err != nil {
-               return nil, fmt.Errorf("initialize mysql connection: %w", err)
+               return nil, fmt.Errorf("initialize postgres connection: %w", err)
        }
        return NewWithDB(db, cfg)
 }
@@ -115,7 +129,7 @@
                }
        }

-       stbl := sq.StatementBuilder.RunWith(db)
+       stbl := sq.StatementBuilder.PlaceholderFormat(sq.Dollar).RunWith(db)
        dbInfo := sqlcommon.NewDBInfo(db, stbl, HandleSQLError)

        return &Datastore{
@@ -510,40 +524,17 @@
        var id, name string
        var createdAt, updatedAt time.Time

-       txn, err := s.db.BeginTx(ctx, &sql.TxOptions{})
-       if err != nil {
-               return nil, HandleSQLError(err)
-       }
-       defer func() {
-               _ = txn.Rollback()
-       }()
-
-       _, err = s.stbl.
+       err := s.stbl.
                Insert("store").
                Columns("id", "name", "created_at", "updated_at").
                Values(store.GetId(), store.GetName(), sq.Expr("NOW()"), sq.Expr("NOW()")).
-               RunWith(txn).
-               ExecContext(ctx)
-       if err != nil {
-               return nil, HandleSQLError(err)
-       }
-
-       err = s.stbl.
-               Select("id", "name", "created_at", "updated_at").
-               From("store").
-               Where(sq.Eq{"id": store.GetId()}).
-               RunWith(txn).
+               Suffix("returning id, name, created_at, updated_at").
                QueryRowContext(ctx).
                Scan(&id, &name, &createdAt, &updatedAt)
        if err != nil {
                return nil, HandleSQLError(err)
        }

-       err = txn.Commit()
-       if err != nil {
-               return nil, HandleSQLError(err)
-       }
-
        return &openfgav1.Store{
                Id:        id,
                Name:      name,
@@ -676,7 +667,7 @@
                Insert("assertion").
                Columns("store", "authorization_model_id", "assertions").
                Values(store, modelID, marshalledAssertions).
-               Suffix("ON DUPLICATE KEY UPDATE assertions = ?", marshalledAssertions).
+               Suffix("ON CONFLICT (store, authorization_model_id) DO UPDATE SET assertions = ?", marshalledAssertions).
                ExecContext(ctx)
        if err != nil {
                return HandleSQLError(err)
@@ -735,7 +726,7 @@
                ).
                From("changelog").
                Where(sq.Eq{"store": store}).
-               Where(fmt.Sprintf("inserted_at <= NOW() - INTERVAL %d MICROSECOND", horizonOffset.Microseconds())).
+               Where(fmt.Sprintf("inserted_at < NOW() - interval '%dms'", horizonOffset.Milliseconds())).
                OrderBy("ulid asc")

        if objectTypeFilter != "" {
@@ -834,7 +825,7 @@
                return storage.ErrNotFound
        }

-       if me, ok := err.(*mysql.MySQLError); ok && me.Number == 1062 {
+       if strings.Contains(err.Error(), "duplicate key value") {
                if len(args) > 0 {
                        if tk, ok := args[0].(*openfgav1.TupleKey); ok {
                                return storage.InvalidWriteInputError(tk, openfgav1.TupleOperation_TUPLE_OPERATION_WRITE)
```


DanCech:
Side note: I'm unclear on why the comparison operator is different here between the 2 backends:
```
-               Where(fmt.Sprintf("inserted_at <= NOW() - INTERVAL %d MICROSECOND", horizonOffset.Microseconds())).
+               Where(fmt.Sprintf("inserted_at < NOW() - interval '%dms'", horizonOffset.Milliseconds())).
```

jpadilla:
@DanCech thanks for the all the work here, I didn't particularly think this was needed but the result is pretty dang nice ✨ 

> Side note: I'm unclear on why the comparison operator is different here between the 2 backends:
> 
> ```
> -               Where(fmt.Sprintf("inserted_at <= NOW() - INTERVAL %d MICROSECOND", horizonOffset.Microseconds())).
> +               Where(fmt.Sprintf("inserted_at < NOW() - interval '%dms'", horizonOffset.Milliseconds())).
> ```

Seems like this has been the case for both adapters since their inception in this repository. @miparnisari we definitely don't need to address here, but I don't see much risk in doing so either. wdyt?

miparnisari:
That definitely looks like an oversight.

DanCech:
I'd be happy to open a separate PR to align the MySQL and SQLite backends with the Postgres logic, since it seems like that is the canonical implementation. I doubt that this specific difference has an impact in the real world, but imho it demonstrates the value of being able to audit the differences and keep the backends in sync with each other.

jpadilla (approved):

miparnisari (approved):
Thanks for doing this :)

PR #1956, opened 2024-09-23
test: update to run matrix tests for all datastore

adriantam:
## Description
Currently the matrix tests are only run for in-memory store.  We want to have it run for all stores. 

## Review Checklist

### DIFF
--- a/tests/check/check.go
+++ b/tests/check/check.go
@@ -1368,6 +1368,6 @@ func assertListUsers(ctx context.Context, t *testing.T, assertion *checktest.Ass
 	}
 }
 
-func runTestMatrixSuite(t *testing.T, client ClientInterface) {
+func RunTestMatrixSuite(t *testing.T, client ClientInterface) {
 	runTestMatrix(t, testParams{typesystem.SchemaVersion1_1, client})
 }
--- a/tests/check/check_test.go
+++ b/tests/check/check_test.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


tests/check/check.go
```diff
@@ -1368,6 +1368,6 @@ func assertListUsers(ctx context.Context, t *testing.T, assertion *checktest.Ass
 	}
 }
 
-func runTestMatrixSuite(t *testing.T, client ClientInterface) {
+func RunTestMatrixSuite(t *testing.T, client ClientInterface) {
```
adriantam, line 1371: This allows other projects that use OpenFGA as library to be able to run test matrix suite for their project's datastore.

tests/check/check_test.go
```diff
@@ -60,26 +54,42 @@ func testRunTestMatrix(t *testing.T, engine string, experimental bool) {
...
-		runTestMatrixSuite(t, openfgav1.NewOpenFGAServiceClient(conn))
+		RunTestMatrixSuite(t, openfgav1.NewOpenFGAServiceClient(conn))
 	})
 }
 
 func TestCheckMemory(t *testing.T) {
 	testRunAll(t, "memory")
 }
 
+func TestMatrixMemory(t *testing.T) {
```
elbuo8, line 65: Why not put it in `TestCheckMemory`?

adriantam: unfortunately the goleak.VerifyNone does not play nicely with the testRunAll :(

miparnisari (commented):
What is the reasoning for doing this? An author of a datastore should _not_ have to run these tests to verify whether they wrote it correctly. I realize that this is not the case today, but i would like to not aggravate the problem: https://github.com/openfga/openfga/pull/1615#issuecomment-2272577165

PR #1962, opened 2024-09-24
fix: goroutine leak in Check

miparnisari:
## Description

We have a leak in Check API. When we attempt to send a message through a channel, e.g. `chan <- msg`, if there is nobody listening on `chan`, the thread will become blocked forever, _unless_ we provide an exit path through context cancellation.


internal/graph/check.go
```diff
@@ -645,23 +638,24 @@ func (c *LocalChecker) processDispatches(ctx context.Context, limit uint32, disp
...
+					}
+					concurrency.TrySendThroughChannel(ctx, checkOutcome{resp: resp}, outcomes)
 					return
 				}
 
 				if msg.dispatchParams != nil {
 					dispatchPool.Go(func(ctx context.Context) error {
 						resp, err := c.dispatch(ctx, msg.dispatchParams.parentReq, msg.dispatchParams.tk)(ctx)
-						outcomes <- checkOutcome{resp: resp, err: err}
+						concurrency.TrySendThroughChannel(ctx, checkOutcome{resp: resp, err: err}, outcomes)
```
adriantam, line 658: This is the bug causing goroutine leak

adriantam (approved):

PR #1921, opened 2024-09-11
refactor: removed unused WithMaxConcurrentReads function code

coson-lu:
## Description

Removed unused WithMaxConcurrentReads function in openfga/internal/graph/check.go file. This is already set here: [openfga/pkg/server/server.go](https://github.com/openfga/openfga/blob/d07302bbb8cbbb095bcc3d887692d79974009e26/pkg/server/server.go#L985-L993)

## References

Issue Resolved: #1920 

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #1963, opened 2024-09-25
perf: optimize nested userset

adriantam:
## Description
Improve nested userset performance by reducing need to constantly fetch for uset -> userset.

E.g. model that will see a gain:

```
model
	schema 1.1
type user
type group
	relations
```

## References
Documentation change - https://github.com/openfga/openfga.dev/pull/861

## Review Checklist


internal/graph/check.go
```diff
@@ -952,6 +960,185 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+
+	// The following member are atomic/sync in anticipation
+	// that the algorithm will parallelize the lookup.
+	dsCount        *atomic.Uint32
+	visitedUserset *sync.Map
+}
+
+func recursiveMatchUserUserset(ctx context.Context,
+	req *ResolveCheckRequest,
+	info *recursiveMatchUserUsersetInfo) (*ResolveCheckResponse, error) {
```
elbuo8, line 979: `info` 🙈

internal/graph/check.go
```diff
@@ -952,6 +960,185 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	filteredIter := storage.NewConditionsFilteredTupleKeyIterator(
+		storage.NewFilteredTupleKeyIterator(
+			storage.NewTupleKeyIteratorFromTupleIterator(iter),
+			validation.FilterInvalidTuples(typesys),
+		),
+		checkutil.BuildTupleKeyConditionFilter(ctx, req.GetContext(), typesys),
+	)
+	defer filteredIter.Stop()
+
+	group, err := storage.GetAllObjects(ctx, filteredIter)
```
elbuo8, line 1100: Why pull in all objects when you can be scanning the results as they are yielded?

internal/graph/check.go
```diff
@@ -952,6 +960,185 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+		}, nil
+	}
+
+	reqTupleKey := req.GetTupleKey()
+	objectType := tuple.GetType(reqTupleKey.GetObject())
+	relation := reqTupleKey.GetRelation()
+
+	dsCount := &atomic.Uint32{}
+	dsCount.Store(1)
+	return recursiveMatchUserUserset(ctx, req, &recursiveMatchUserUsersetInfo{
```
elbuo8, line 1130: We could start processing this without having to wait for all the Objects to be loaded into memory?

internal/graph/check.go
```diff
@@ -1053,8 +1554,13 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 			if !tuple.IsObjectRelation(reqTupleKey.GetUser()) {
 				if typesys.UsersetCanFastPath(directlyRelatedUsersetTypes) {
 					resolver = c.checkUsersetFastPath
+				} else if c.optimizationsEnabled && typesys.RecursiveUsersetCanFastPath(
+					tuple.ToObjectRelationString(tuple.GetType(reqTupleKey.GetObject()), reqTupleKey.GetRelation()),
+					tuple.GetType(reqTupleKey.GetUser())) {
+					return nestedUsersetFastpath(ctx, typesys, ds, req, int(c.maxConcurrentReads))
```
miparnisari, line 1560: FYI -`maxConcurrentReads` isn't currently being set in `LocalChecker` and will be removed soon: https://github.com/openfga/openfga/pull/1921. We can bring it back if you need it

internal/graph/check.go
```diff
@@ -945,6 +953,499 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	req *ResolveCheckRequest,
+	userToUsersetMessageChan,
+	objectToUsersetMessageChan chan usersetMessage) (*ResolveCheckResponse, storage.SortedSet, storage.SortedSet, error) {
+	usersetFromUser := storage.NewSortedSet()
+	usersetFromObject := storage.NewSortedSet()
+
+	userToUsersetDone := false
+	objectToUsersetDone := false
+
+	for {
```
miparnisari, line 1298:
```suggestion
	for !userToUsersetDone && !objectToUsersetDone {
```

and then you don't need `if`s in line 1305 and 1335

internal/graph/check.go
```diff
@@ -945,6 +953,499 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+		select {
+		case <-ctx.Done():
+			return nil, nil, nil, ctx.Err()
+		case userToUsersetMessage, ok := <-userToUsersetMessageChan:
+			if !ok {
```
miparnisari, line 1303: when either channel is closed, `if len(<set>) == 0` you can exit early with `allowed=false` and not wait for the another channel to close too.

internal/graph/check.go
```diff
@@ -945,6 +953,499 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+			_ = pool.Wait()
+			close(checkOutcomeChan)
+			wg.Done()
+		}()
+		for _, usersetItem := range usersetItems {
+			_, visited := commonParameters.visitedUserset.LoadOrStore(usersetItem, struct{}{})
+			if !visited {
+				newReq := req.clone()
+				newTupleKey := tuple.NewTupleKey(usersetItem, req.GetTupleKey().GetRelation(), req.GetTupleKey().GetUser())
+				newReq.TupleKey = newTupleKey
```
miparnisari, line 1007: When you did `newReq.TupleKey.Object = usersetItem` you got a race condition, did you not? I got that too :( We should fix this in a separate PR

internal/graph/check.go
```diff
@@ -945,6 +953,499 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	req *ResolveCheckRequest,
+	commonParameters *recursiveMatchUserUsersetInfo,
+	level int) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "recursiveMatchUserUserset")
+	defer span.End()
+
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+	if level > int(req.GetRequestMetadata().Depth) {
```
miparnisari, line 1064: FYI This would be simpler if you didn't need a `level` variable and `depth` started at 0 and we exited when `req.GetRequestMetadata().Depth == maxAllowedDepth` 😢  https://github.com/openfga/openfga/issues/994

internal/graph/check.go
```diff
@@ -984,6 +991,491 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+				if found {
+					return &ResolveCheckResponse{
+						Allowed: true,
+						ResolutionMetadata: &ResolveCheckResponseMetadata{
+							DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount + 2,
+						},
+					}, nil, nil, nil
+				}
+				//  check to see if there is a direct assignment
+				if req.GetTupleKey().GetObject() == userToUsersetMessage.userset {
```
miparnisari, line 1387: I wonder if we should also modify `checkDirectUserTuple` so that if we are running the `nestedUsersetFastpath` approach we don't need run `checkDirectUserTuple` because we have line 1387.

adriantam: Probably check direct is better as it covers the non-nested case.

internal/graph/check.go
miparnisari: https://github.com/openfga/openfga/issues/1995

internal/graph/check.go
adriantam: Updated to common parameter

internal/graph/check.go
```diff
@@ -984,6 +991,491 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	req *ResolveCheckRequest,
+	maxConcurrentReads int) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "nestedUsersetFastpath")
+	defer span.End()
+
+	reqTupleKey := req.GetTupleKey()
+	objectType := tuple.GetType(reqTupleKey.GetObject())
+	relation := reqTupleKey.GetRelation()
+
+	userToUsersetMessageChan := streamedLookupUsersetForUser(ctx, typesys, ds, req, maxConcurrentReads)
```
miparnisari, line 1445: Within `streamedLookupUsersetForUser` and `streamedLookupUsersetForObject` the only way for you to realize when the goroutines are done is when the channels are closed. In every path where you shortcircuit, e.g. in line 1452, how are we ensuring that we are not forgetting about those two goroutines?

.config-schema.json
```diff
@@ -21,7 +21,7 @@
         "maxConcurrentReadsForCheck": {
             "description": "The maximum allowed number of concurrent reads in a single Check query (default is MaxUint32).",
             "type": "integer",
-            "default": 4294967295,
+            "default": 100,
```
elbuo8, line 24: ?

adriantam: As noted in https://github.com/openfga/openfga/pull/1963#discussion_r1788122391, we were not using this config value and config is about to be removed.  I repurpose the variable for the purpose of defining how many concurrent child branch for userset.  Since we use this variable for channel length, concurrent routine etc., this variable needs to be a sane value rather than -1.

internal/graph/check.go
adriantam: Good callout - fixed.

internal/graph/check.go
```diff
@@ -984,6 +991,494 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	defer span.End()
+
+	usersetMessageChan := make(chan usersetMessage, maxConcurrentReads)
+
+	go func() {
+		defer func() {
+			close(usersetMessageChan)
+		}()
+
+		// Note that this will take care of the publicly assignable case as well.
```
miparnisari, line 1196:
```suggestion
		// Note that if the type is publicly assignable, this will fetch that table as well.
```

internal/graph/check_test.go
```diff
@@ -4101,3 +4105,1619 @@ func TestCheckTTUSlowPath(t *testing.T) {
...
+			expected: &ResolveCheckResponse{
+				Allowed: false,
+				ResolutionMetadata: &ResolveCheckResponseMetadata{
+					DatastoreQueryCount: 3,
+					CycleDetected:       false,
+				},
+			},
+		},
+		{
+			name: "recursive_linear_found",
```
miparnisari, line 4763: Nit: this test is already up above

internal/graph/check.go
```diff
@@ -984,6 +991,494 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+		if commonParameters.userToUsersetMapping.Exists(usersetName) {
+			return &ResolveCheckResponse{
+				Allowed: true,
+				ResolutionMetadata: &ResolveCheckResponseMetadata{
+					DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount + commonParameters.dsCount.Load(),
+				},
+			}, nil
+		}
+		usersetItems = append(usersetItems, usersetName)
+	}
```
miparnisari, line 1160: Could we re-use what we already have? (Disclaimer: haven't tested)

```suggestion
objectType := tuple.GetType(req.GetTupleKey().GetObject())
relation := req.GetTupleKey().GetRelation()

cancellableCtx, cancelFunc := context.WithCancel(ctx)
defer cancelFunc()

objectToUsersetMessageChan := streamedLookupUsersetForObject(cancellableCtx, commonParameters.typesys, commonParameters.ds, req, []*openfgav1.RelationReference{
	typesystem.DirectRelationReference(objectType, relation),
}, commonParameters.maxConcurrentReads)

var usersetItems []string
for usersetMsg := range objectToUsersetMessageChan {
	if usersetMsg.err != nil {
		return nil, usersetMsg.err
	}
	usersetName := usersetMsg.userset
	if commonParameters.userToUsersetMapping.Exists(usersetName) {
		return &ResolveCheckResponse{
			Allowed: true,
			ResolutionMetadata: &ResolveCheckResponseMetadata{
				DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount + commonParameters.dsCount.Load(),
			},
		}, nil
	}
	usersetItems = append(usersetItems, usersetName)
}
```

internal/graph/check_test.go
adriantam: will rename to found_second_level.

.config-schema.json
miparnisari: Sorry I wasn't clear - the config value **is** being used, but not in LocalChecker, it's being used in `CheckQuery`. You will need to pass it down to LocalChecker.

internal/graph/check.go
adriantam: We can.  However, calling streamedLookupUsersetForObject will create un-necessary go-routine since we do not have anything to parallelize (i.e., this is a single producer single consumer).

adriantam: Will note in comment.

internal/graph/check.go
```diff
@@ -984,6 +991,494 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	primarySortedSet storage.SortedSet,
+	secondarySortedSet storage.SortedSet) (bool, error) {
+	if message.err != nil {
+		return false, message.err
+	}
+	primarySortedSet.Add(message.userset)
+	if secondarySortedSet.Exists(message.userset) {
+		return true, nil
+	}
+	return false, nil
```
miparnisari, line 1343: Nit:
```suggestion
	return secondarySortedSet.Exists(message.userset), nil
```

internal/graph/check.go
adriantam: Done.

internal/graph/check.go
miparnisari: True, but the tradeoff is that you reduce code duplication a lot, which I think makes it worth it because this code will then be more easily adaptable to handle the nested TTU scenario.

.config-schema.json
adriantam: Fixed so that we use concurrencyLimit instead.  As such, the variable value does not need to change now.

internal/graph/check.go
adriantam: Updated

miparnisari (dismissed):

internal/graph/check.go
```diff
@@ -981,6 +988,466 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	recursiveFunc recursiveMatchUserUsersetFunc,
+) (*ResolveCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, "parallelizeRecursiveMatchUserUserset")
+	defer span.End()
+	checkOutcomeChan := make(chan checkOutcome, commonParameters.concurrencyLimit)
+
+	pool := concurrency.NewPool(ctx, commonParameters.concurrencyLimit)
+	cancellableCtx, cancelFunc := context.WithCancel(ctx)
+
+	wg := &sync.WaitGroup{}
```
elbuo8, line 1024: Why the need for this?

internal/graph/check.go
```diff
@@ -981,6 +988,466 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+		case msg, ok := <-checkOutcomeChan:
+			if !ok {
+				return &ResolveCheckResponse{
+					Allowed: false,
+					ResolutionMetadata: &ResolveCheckResponseMetadata{
+						DatastoreQueryCount: req.GetRequestMetadata().DatastoreQueryCount + commonParameters.dsCount.Load(),
+					},
+				}, nil
+			}
+			if msg.err != nil && !errors.Is(msg.err, context.Canceled) && !errors.Is(msg.err, context.DeadlineExceeded) {
```
elbuo8, line 1079: Why not simplify with a `break` so its easier to understand that some cases are meant to exit explicitly?

Then the happy path of allowed can be simpler also.

internal/graph/check.go
```diff
@@ -981,6 +988,466 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+			storage.NewFilteredTupleKeyIterator(
+				storage.NewTupleKeyIteratorFromTupleIterator(iter),
+				validation.FilterInvalidTuples(typesys),
+			),
+			checkutil.BuildTupleKeyConditionFilter(ctx, req.GetContext(), typesys),
+		)
+		defer filteredIter.Stop()
+
+		for {
+			if ctx.Err() != nil {
```
elbuo8, line 1192: Why this check here?

elbuo8: In `streamedLookupUsersetForObject` you rely on `.Next` telling you its cancelled...

internal/graph/check.go
```diff
@@ -981,6 +988,466 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+			}, usersetMessageChan)
+		}
+	}()
+	return usersetMessageChan
+}
+
+// streamedLookupUsersetForObject streams the userset that are assigned to
+// the object to the usersetMessageChan channel.
+func streamedLookupUsersetForObject(ctx context.Context,
+	typesys *typesystem.TypeSystem,
```
elbuo8, line 1223: Why not pass the `common` object you have?

internal/graph/check.go
```diff
@@ -981,6 +988,466 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+			concurrency.TrySendThroughChannel(ctx, usersetMessage{
+				userset: usersetName,
+				err:     nil,
+			}, usersetMessageChan)
+		}
+	}()
+	return usersetMessageChan
+}
+
+func processUsersetMessage(message usersetMessage,
```
elbuo8, line 1305: Please document this function as it's critical and its usage shifts by the caller. This can be easily misunderstood by others.

internal/graph/check.go
adriantam: Right, we don't need this.  The wg is removed.

internal/graph/check.go
adriantam: Clean up a little bit by moving the cvheck for cacneled / deadline exceeded out to be more explicit.

internal/graph/check.go
adriantam: Right, rely on next to tell it is cancelled.

internal/graph/check.go
adriantam: Will do.

internal/graph/check.go
```diff
@@ -981,6 +988,456 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+	ctx, span := tracer.Start(ctx, "parallelizeRecursiveMatchUserUserset")
+	defer span.End()
+	checkOutcomeChan := make(chan checkOutcome, commonParameters.concurrencyLimit)
+
+	pool := concurrency.NewPool(ctx, commonParameters.concurrencyLimit)
+	cancellableCtx, cancelFunc := context.WithCancel(ctx)
+
+	defer func() {
+		cancelFunc()
+	}()
```
miparnisari, line 1026: Nit:

```suggestion
	
	cancellableCtx, cancelFunc := context.WithCancel(ctx)
	defer cancelFunc()
	pool := concurrency.NewPool(cancellableCtx, commonParameters.concurrencyLimit)
```

and then you don't have to reference `cancellableCtx` inside `pool.Go`, you can just refer to `ctcx`

miparnisari (approved):

internal/graph/check.go
miparnisari: This was so that we waited for the goroutine to exit before returning from this function. We could have flakey tests if we remove the waiting

internal/graph/check.go
adriantam: let's do that in subsequent PR

PR #1964, opened 2024-09-25
feat: minor improvements in tracing of Check API

miparnisari:
## Description
When trying to find patterns in Check requests via tracing information, i want to be able to find all top-level "Check" spans and gather all the information from there, and not have to manually dig into children spans.

I'm also improving attribute information for the model ID resolution which is a bit hard to find right now.

## Testing

`docker-compose up --build`

After this PR:

image

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -953,8 +953,6 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		return nil, err
 	}
 
-	span.SetAttributes(attribute.String("authorization_model_id", typesys.GetAuthorizationModelID()))
-
 	if err := validation.ValidateUserObjectRelation(typesys, tuple.ConvertCheckRequestTupleKeyToTupleKey(tk)); err != nil {
 		return nil, serverErrors.ValidationError(err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/typesystem/resolver.go
```diff
@@ -39,13 +39,13 @@ func MemoizedTypesystemResolverFunc(datastore storage.AuthorizationModelReadBack
 	return func(ctx context.Context, storeID, modelID string) (*TypeSystem, error) {
 		ctx, span := tracer.Start(ctx, "MemoizedTypesystemResolverFunc", trace.WithAttributes(
 			attribute.String("store_id", storeID),
-			attribute.String("authorization_model_id", modelID),
```
miparnisari, line 42: FYI this can be empty at this point, so i'm moving it down below.

pkg/server/server.go
```diff
@@ -1362,9 +1366,6 @@ func (s *Server) IsReady(ctx context.Context) (bool, error) {
 // resolveTypesystem resolves the underlying TypeSystem given the storeID and modelID and
 // it sets some response metadata based on the model resolution.
 func (s *Server) resolveTypesystem(ctx context.Context, storeID, modelID string) (*typesystem.TypeSystem, error) {
-	ctx, span := tracer.Start(ctx, "resolveTypesystem")
```
miparnisari, line 1365: this span is pretty useless. it's not logging the store id. i can't do anything with just the model_id. so i'm removing it and putting all the info within the `typesystemResolver` span.

pkg/typesystem/resolver.go
```diff
@@ -39,9 +39,11 @@ func MemoizedTypesystemResolverFunc(datastore storage.AuthorizationModelReadBack
...
```
miparnisari, line 42: FYI at this point, `model_id` can be empty

jpadilla (dismissed):

elbuo8 (approved):

jpadilla (approved):

PR #1981, opened 2024-10-01
Increase fast path concurrency

elbuo8:
## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -815,13 +815,13 @@ func (c *LocalChecker) checkMembership(ctx context.Context, req *ResolveCheckReq
 
 	cancellableCtx, cancelFunc := context.WithCancel(ctx)
 	// sending to channel in batches up to a pre-configured value to subsequently checkMembership for.
-	pool := concurrency.NewPool(cancellableCtx, 1)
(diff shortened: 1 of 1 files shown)


adriantam:
Instead of skip changelog, perhaps it will be good to mention this perf improvement

internal/graph/check.go
```diff
@@ -563,11 +563,11 @@ func checkAssociatedObjects(ctx context.Context, req *ResolveCheckRequest, objec
 	if allowed {
 		span.SetAttributes(attribute.Bool("allowed", true))
 	}
-	reqCount := req.GetRequestMetadata().DatastoreQueryCount + 1
+
 	return &ResolveCheckResponse{
 		Allowed: allowed,
 		ResolutionMetadata: &ResolveCheckResponseMetadata{
-			DatastoreQueryCount: reqCount,
+			DatastoreQueryCount: 1,
```
adriantam, line 570: Probably worth calling out this number does not matter here because the caller will set their own value.

adriantam (dismissed):

internal/graph/check.go
elbuo8: Uh? No. It actually gets accumulated.

adriantam (approved):

PR #1982, opened 2024-10-02
chore: refactor reflexive tuple/userset checker into separate functions

miparnisari:


internal/graph/check.go
```diff
@@ -473,10 +473,8 @@ func (c *LocalChecker) ResolveCheck(
 	object := tupleKey.GetObject()
 	relation := tupleKey.GetRelation()
 
-	userObject, userRelation := tuple.SplitObjectRelation(req.GetTupleKey().GetUser())
-
-	// Check(document:1#viewer@document:1#viewer) will always return true
-	if relation == userRelation && object == userObject {
+	yes := tuple.IsSelfDefining(req.GetTupleKey())
```
adriantam, line 476: nitpik: yes is not a really descriptive variable name.  Perhaps isSelfDefining?

adriantam (approved):

PR #1984, opened 2024-10-02
fix: faster tuple validation when reading from datastore

elbuo8:
Validations when reading from the datastore do not have to be the same ones as when we write to the datastore. Thus paying the penalty on each tuple read, is unnecessary. 

Additionally, some validations where being done explicitly and then implicitly under more complex types again, this PR aims to reduce the duplication also to improve overall validation.

Before:
```
BenchmarkValidateTuple
BenchmarkValidateTuple-16    	 1000000	      1167 ns/op
```
After:
```
BenchmarkValidateTupleForWrite
BenchmarkValidateTupleForWrite-16    	 1201627	       986.8 ns/op
BenchmarkValidateTupleForRead
BenchmarkValidateTupleForRead-16    	 3549020	       331.4 ns/op
```

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -999,7 +999,7 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 
 			// filter out invalid tuples yielded by the database query
 			tupleKey := t.GetKey()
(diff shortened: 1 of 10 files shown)


yissellokta (approved):

jpadilla (approved):
✨ nice!

PR #1993, opened 2024-10-04
perf: remove duplicate data from cached records

jpadilla:
## Description

Store leaner tuples in cached iterators.

Instead of storing well-formed tuples (aka `*openfgav1.Tuple`) when caching iterators we story a minimal representation. 

Before these changes the cached iterators would have results like:

```
*openfgav1.Tuple[]{
  {
    Key: { Object:"company:1", Relation:"viewer", User: "user:1", Condition: ... }
    Timestamp: ts1
  },
  {
    Key: { Object:"company:1", Relation:"viewer", User: "user:2", Condition: ... }
    Timestamp: ts2
  },
  {
    Key: { Object:"company:1", Relation:"viewer", User: "user:3", Condition: nil }
    Timestamp: ts3
  },
}
```

The proposed changes here now represent the tuples in the cached iterators with a simple struct(instead of a proto struct along with all of its baggage) containing only the properties that are unique per tuple.

```
storage.TupleRecord[]{
  {
    UserObjectID: "1"
    ConditionName: ...
    ConditionContext: ...
    Timestamp: ts1
  },
  {
    UserObjectID: "2"
    ConditionName: ...
    ConditionContext: ...
    Timestamp: ts2
  },
  {
    UserObjectID: "3"
    Timestamp: ts3
  },
}
```

The iterator returned for cache hits can fully hydrate these minimal tuples back into their well-formed version.

## References

https://github.com/openfga/openfga/pull/1924

## Review Checklist


internal/graph/storagewrapper.go
```diff
@@ -203,7 +204,9 @@ func (c *CachedDatastore) newCachedIterator(
 
 	return &cachedIterator{
 		iter:          iter,
-		tuples:        make([]*openfgav1.Tuple, 0, c.maxResultSize),
+		relation:      relation,
```
elbuo8, line 207: Can we abstract this away? Given how we can have multiple instances of an iterator? Maybe via a function?

internal/graph/iterators.go
```diff
@@ -0,0 +1,75 @@
...
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+// cachedTupled is a simple struct to hold the minimal data needed
+// to store in a cached iterator.
+type cachedTuple struct {
```
elbuo8, line 14:
```suggestion
type cachedUserTuple struct {
```

internal/graph/iterators.go
```diff
@@ -0,0 +1,75 @@
...
+	timestamp *timestamppb.Timestamp
+}
+
+// cachedUserTupleIterator is a wrapper around a cached iterator
+// for a given object/relation.
+type cachedUserTupleIterator struct {
+	object   string
+	relation string
+	iter     storage.Iterator[cachedUserTuple]
+}
```
miparnisari, line 26: I'm curious on why we are changing the code one level above of where we could be making the change (the storage interface itself). We could be reducing network transfers for everyone, regardless of whether you have caching turned on.

elbuo8: I'm not following, what are you suggesting?

miparnisari: Change `storage.ReadUsersetTuples` so that it returns `UserTupleIterator`

internal/graph/iterators.go
```diff
@@ -0,0 +1,78 @@
+package graph
```
elbuo8, line 1: Why not in storage?

internal/graph/storagewrapper.go
```diff
@@ -388,7 +432,40 @@ func (c *cachedIterator) addToBuffer(t *openfgav1.Tuple) bool {
...
+		record.InsertedAt = timestamp.AsTime()
+	}
+
+	if condition := tk.GetCondition(); condition != nil {
+		record.ConditionName = condition.GetName()
+		record.ConditionContext = condition.GetContext()
+	}
+
+	// Remove any fields that are duplicated and known by iterator
+	if c.objectID != "" && c.objectID == record.ObjectID {
```
elbuo8, line 457: Do we need the empty checks?
In which case there would be empty?

internal/graph/iterators.go
jpadilla: no super strong opinion but mostly because right atm this is only scoped/used here, and pkg vs internal.

internal/graph/storagewrapper.go
```diff
@@ -132,7 +132,7 @@ func (c *CachedDatastore) ReadStartingWithUser(
 		b.WriteString(fmt.Sprintf("/%s", strconv.FormatUint(hasher.Sum64(), 10)))
 	}
 
-	return c.newCachedIterator(ctx, store, iter, b.String(), storage.GetInvalidIteratorByUserObjectTypeCacheKeys(store, subjects, filter.ObjectType))
+	return c.newCachedIteratorByUserObjectType(ctx, store, iter, b.String(), subjects, filter.ObjectType)
```
miparnisari, line 135: unrelated to this PR: how does line 119 work? what if `userFilter` has no relation but a wildcard?

elbuo8 (dismissed):

internal/graph/iterators.go
jpadilla: > We could be reducing network transfers for everyone, regardless of whether they have caching turned on.
I'm still not following, how would "reducing network transfers" work without caching turned on in this case?

elbuo8 (approved):

internal/graph/iterators.go
miparnisari: We change the storage layer itself rather than the caching layer above it. We make the storage layer return an optimized iterator :)

internal/graph/storagewrapper.go
```diff
@@ -404,8 +508,9 @@ func (c *cachedIterator) flush() {
 	// Copy tuples buffer into new destination before storing into cache
 	// otherwise, the cache will be storing pointers. This should also help
 	// with garbage collection.
-	tuples := make([]*openfgav1.Tuple, len(c.tuples))
+	tuples := make([]*storage.TupleRecord, len(c.tuples))
```
elbuo8, line 511: We could do:
```
tuples := c.tuples
c.tuples := nil
```

Thus the pointer gets de-referenced from the struct (so it can get gc'ed) but we can still store it without having to pay the penalty of the `copy` operation.

internal/graph/storagewrapper.go
```diff
@@ -321,7 +419,7 @@ func (c *cachedIterator) Next(ctx context.Context) (*openfgav1.Tuple, error) {
 		return nil, err
 	}
 
-	c.addToBuffer(t)
+	c.tuples = append(c.tuples, t)
```
elbuo8, line 422: This shouldn't add if the limit is breached, there should be a guard here too.

internal/graph/storagewrapper.go
```diff
@@ -332,66 +430,130 @@ func (c *cachedIterator) Next(ctx context.Context) (*openfgav1.Tuple, error) {
...
+	if c.relation != "" && c.relation == record.Relation {
+		record.Relation = ""
+	}
+	if c.userType != "" && c.userType == record.UserObjectType {
+		record.UserObjectType = ""
+	}
+
+	c.records = append(c.records, record)
+
 	if len(c.tuples) >= c.maxResultSize {
```
elbuo8, line 551: This should be `c.records` as `c.tuples` is no longer growing at this point.

elbuo8 (dismissed):

elbuo8 (approved):

elbuo8 (approved):

PR #2004, opened 2024-10-08
refactor: make optimize nested userset generic

miparnisari:
## Description

Extension of https://github.com/openfga/openfga/pull/1963. I created a new `Mapper` abstraction that returns a mapping of a tuple. This will make it easier to optimize for nested userset and nested TTU scenarios.

## References


internal/graph/check.go
```diff
@@ -1438,6 +1421,27 @@ func nestedUsersetFastpath(ctx context.Context,
 	return parallelizeRecursiveMatchUserUserset(ctx, usersetFromObject.Values(), req, recursiveCommonData, recursiveMatchUserUserset)
 }
 
+func buildIterator(ctx context.Context, mapperKind tuplemapper.MapperKind, ds storage.RelationshipTupleReader, req *ResolveCheckRequest, allowedUserTypeRestrictions []*openfgav1.RelationReference) (storage.TupleIterator, error) {
```
elbuo8, line 1424:
```suggestion
func buildMappedIterator(ctx context.Context, mapperKind tuplemapper.MapperKind, ds storage.RelationshipTupleReader, req *ResolveCheckRequest, allowedUserTypeRestrictions []*openfgav1.RelationReference) (storage.TupleIterator, error) {
```

internal/graph/check.go
```diff
@@ -1438,6 +1421,27 @@ func nestedUsersetFastpath(ctx context.Context,
 	return parallelizeRecursiveMatchUserUserset(ctx, usersetFromObject.Values(), req, recursiveCommonData, recursiveMatchUserUserset)
 }
 
+func buildMappedIterator(ctx context.Context, mapperKind tuplemapper.MapperKind, ds storage.RelationshipTupleReader, req *ResolveCheckRequest, allowedUserTypeRestrictions []*openfgav1.RelationReference) (storage.TupleIterator, error) {
+	var iter storage.TupleIterator
+	var erroriter error
```
adriantam, line 1426: nitpik: erroriter suggests itself is an iterator.  Perhaps err be a better name?
Also, I am confused in to why do we need to assign to iter and erroriter first?  Why not return directly in line 1429?

internal/graph/check.go
```diff
@@ -1211,43 +1211,24 @@ func streamedLookupUsersetForUser(ctx context.Context,
...
-				userset: "",
-				err:     err,
-			}, usersetMessageChan)
-			return
-		}
-
 		filteredIter := storage.NewConditionsFilteredTupleKeyIterator(
 			storage.NewFilteredTupleKeyIterator(
-				storage.NewTupleKeyIteratorFromTupleIterator(iter),
+				storage.NewTupleKeyIteratorFromTupleIterator(castedMapper),
```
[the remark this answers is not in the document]
elbuo8: If anything, its the filtered iterator the one that gets wrapped in the mapper 🤔

internal/tuplemapper/tuplemapper.go
```diff
@@ -0,0 +1,91 @@
...
+	"context"
+	"fmt"
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/tuple"
+)
+
+type Mapper[ResultType any] interface {
```
adriantam, line 13: Wondering if allow ResultType other than string makes this too generic and hard to use.  (i.e., the discussion of `New(kind MapperKind, iter storage.TupleIterator)` below)

internal/graph/check.go
```diff
@@ -1211,43 +1211,24 @@ func streamedLookupUsersetForUser(ctx context.Context,
...
 	commonParameters *recursiveMatchUserUsersetCommonData,
+	tupleIterator storage.TupleIterator,
 	req *ResolveCheckRequest) chan usersetMessage {
 	ctx, span := tracer.Start(ctx, "streamedLookupUsersetForObject")
 	defer span.End()
 
 	usersetMessageChan := make(chan usersetMessage, commonParameters.concurrencyLimit)
 
+	mapper := tuplemapper.New(commonParameters.tupleMapperKind, tupleIterator)
+	castedMapper, _ := mapper.(tuplemapper.Mapper[string]) // hmmm..
```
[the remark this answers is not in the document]
miparnisari: Moved outside

internal/tuplemapper/tuplemapper.go
```diff
@@ -0,0 +1,91 @@
...
+	// NoOpKind is for tests only.
+	NoOpKind MapperKind = 0
+
+	// NestedUsersetKind is a mapper that returns the userset ID from the tuple's user field.
+	NestedUsersetKind MapperKind = 1
+	NestedTTUKind     MapperKind = 2
+)
+
+// New is a factory that returns the right mapper based on the request's kind.
+func New(kind MapperKind, iter storage.TupleIterator) interface{} {
```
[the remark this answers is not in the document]
miparnisari: This has been completely changed now.

internal/tuplemapper/tuplemapper.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (n NestedUsersetMapper) Head(ctx context.Context) (string, error) {
+	tupleRes, err := n.Iter.Head(ctx)
+	if err != nil {
+		return "", err
+	}
+	return n.Map(tupleRes)
+}
+
+func (n NestedUsersetMapper) Map(t *openfgav1.TupleKey) (string, error) {
```
adriantam, line 49: nitpik: 
1. Why does this need to be a function of NestedUsersetMapper instead a function that does not belong to any struct?
2. why is this public?

internal/tuplemapper/tuplemapper.go
```diff
@@ -0,0 +1,56 @@
...
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/tuple"
+)
+
+type MapperKind int64
+
+const (
+	// NestedUsersetKind is a mapper that returns the userset ID from the tuple's user field.
```
adriantam, line 16: comment (non-blocking): NestedUsersetKind / NestedTTUKind does not feel like belonging here.  It is exclusively used in graph/check.go

internal/tuplemapper/tuplemapper.go
miparnisari: 1. The idea is that in the future there will be a NestedTTUMapper that will do this: `return t.GetUser(), nil`
2. No reason, made private now :)

internal/tuplemapper/tuplemapper.go
miparnisari: moved

adriantam (approved):

PR #2006, opened 2024-10-08
feat: Add cache controller

elbuo8:
## Review Checklist


internal/cachecontroller/cache_controller.go
```diff
@@ -0,0 +1,124 @@
...
+			c.invalidateIteratorCacheByObjectRelation(storeID, changes[idx].GetTupleKey().GetObject(), changes[idx].GetTupleKey().GetRelation())
+		}
+	}
+
+	return entry.LastModified, nil
+}
+
+func (c *CacheController) invalidateIteratorCache(storeID string) {
+	// These entries do not need to expire
+	c.cache.Set(storage.GetInvalidIteratorCacheKey(storeID), &storage.InvalidEntityCacheEntry{LastModified: time.Now()}, math.MaxInt)
```
adriantam, line 119: Can you explain why we don't delete instead?

internal/cachecontroller/cache_controller.go
```diff
@@ -0,0 +1,124 @@
...
+			break
+		}
+	}
+	// all changes are new, thus we should revoke the whole query cache
+	if idx == len(changes)-1 {
+		c.invalidateIteratorCache(storeID)
+	} else {
+		// only a subset of changes are new, revoke the respective ones
+		for ; idx >= 0; idx-- {
+			c.invalidateIteratorCacheByObjectRelation(storeID, changes[idx].GetTupleKey().GetObject(), changes[idx].GetTupleKey().GetRelation())
```
adriantam, line 110: Would it be safer to simply delete all cache for the store instead?  A bit concern that data is stale when grandchildren /great-granchildren tuples are changed.  The grandparent / great-grandparent will need to be invalidated as well.

pkg/server/commands/check_command.go
```diff
@@ -57,6 +61,12 @@ func WithCheckCommandLogger(l logger.Logger) CheckQueryOption {
...
 
+func WithCacheController(ctrl *cachecontroller.CacheController) CheckQueryOption {
+	return func(c *CheckQuery) {
+		c.cacheController = ctrl
+	}
+}
+
 func NewCheckCommand(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, typesys *typesystem.TypeSystem, opts ...CheckQueryOption) *CheckQuery {
 	cmd := &CheckQuery{
 		logger:             logger.NewNoopLogger(),
```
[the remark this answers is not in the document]
elbuo8: If that is the preferred convention, happy to change it.

internal/cachecontroller/cache_controller.go
elbuo8: These are not subproblems. They are direct queries which do not have parent's, just their direct object relation.

internal/cachecontroller/cache_controller.go
elbuo8: If we were to delete _all_ entries that would be iterating through all the keys of the cache (even by using a prefix) as there is not Tree, just a hashmap. So it would grab the lock to iterate through all the keys in the cache, release the lock (thus allowing for race conditions), and grabbing another Lock to remove all the matched keys. The cache lock contention would be proportional to the number of items in the cache.

With this async pattern, we only pay the penalty of 1 read lock which is very cheap.

pkg/server/commands/check_command.go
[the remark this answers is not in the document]
adriantam: I think this is the convention we try to keep.  @miparnisari

internal/cachecontroller/cache_controller_test.go
```diff
@@ -0,0 +1,245 @@
...
+					Timestamp: timestamppb.New(time.Now()),
+					TupleKey: &openfgav1.TupleKey{
+						Object:   "test",
+						Relation: "viewer",
+						User:     "test",
+					},
+				},
+				{
+					Operation: openfgav1.TupleOperation_TUPLE_OPERATION_WRITE,
+					Timestamp: timestamppb.New(time.UnixMilli(1257894000000)),
```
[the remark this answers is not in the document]
elbuo8: Fixed time in the past. Can change to make more obvious.

internal/graph/storagewrapper.go
```diff
@@ -165,35 +158,51 @@ func (c *CachedDatastore) Read(
 
 	var b strings.Builder
 	b.WriteString(
-		fmt.Sprintf("%sr%s/%s", QueryCachePrefix, store, tuple.TupleKeyToString(tupleKey)),
+		storage.GetReadCacheKey(store, tuple.TupleKeyToString(tupleKey)),
 	)
-	return c.newCachedIterator(ctx, iter, b.String())
+	return c.newCachedIterator(ctx, store, iter, b.String(), storage.GetInvalidIteratorByObjectRelationCacheKey(store, tupleKey.GetObject(), tupleKey.GetRelation()))
```
adriantam, line 163: There is a strong assumption here that datastore.Read has object and relation defined.  This is true where we are currently using it.  However, can we document this AND add some form of check to ensure this is the case (i.e., throw out internal error)?  This allows us not to forget about it in the future and accidentally expand the use when we should not.

internal/cachecontroller/cache_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		SortDesc: true,
+		Pagination: storage.PaginationOptions{
+			PageSize: storage.DefaultPageSize,
+			From:     "",
+		}}
+	return c.ds.ReadChanges(ctx, storeID, storage.ReadChangesFilter{}, opts)
+}
+
+func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, storeID string) (time.Time, error) {
+	ctx, span := tracer.Start(ctx, "cacheController.Invalidations")
```
jpadilla, line 110: could it be useful to annotate this span with the invalidation side effects that occurred?

internal/server/config/config.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	DefaultCacheLimit = 10000
 
+	DefaultCacheControllerEnabled = false
```
jpadilla, line 39: are we wiring up config flags and such?

pkg/server/server.go
```diff
@@ -366,6 +373,20 @@ func WithCacheLimit(limit uint32) OpenFGAServiceV1Option {
 	}
 }
 
+// WithCacheControllerEnabled enables cache invalidation of different cache entities.
+func WithCacheControllerEnabled(enabled bool) OpenFGAServiceV1Option {
```
jpadilla, line 377: We could use some simple tests to validate that this wires up things correctly down the road:

```
if s.cache != nil && s.cacheControllerEnabled {
```

internal/cachecontroller/cache_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	defer span.End()
+	cacheKey := storage.GetChangelogCacheKey(storeID)
+	// this should have a deadline since it will hold everything
```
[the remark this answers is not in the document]
elbuo8: TODO. 

Will clean up.

internal/server/config/config.go
elbuo8: Subsequent PR.

pkg/server/server.go
elbuo8: Tests are actually included already.

jpadilla (approved):

PR #2013, opened 2024-10-11
fix: dispatch counter in ListObjects

poovamraj:
We are fixing miscounts in Dispatch Counts and Was Throttled metadata in List Objects for Intersection and Negation use cases

## Description
This PR fixes two issues
- The metadata from Check Resolver was not used in List User metadata count. This has been fixed by referencing the proper metadata
- The `WasThrottled` metadata was not set properly as the value changed in each loop. We have introduced a new check that will modify this value only if it was not set before and the last check dispatch in the consumer loop value will be set for the metadata.
- We haven't caught this earlier because the check resolver we used inside the LO unit test didn't have a dispatch throttler in its builder. We have introduced this in this PR
- TODO: Not sure how this PR is reducing the number of dispatches in one of the tests. Need to find and fix it. Also need to improve the intersection test case

## Review Checklist
- [] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]


miparnisari:
>  Not sure how this PR is reducing the number of dispatches in one of the tests. 

Because the Check API call made in that test with the intersection has 0 dispatches. So 1 dispatch from ReverseExpand + 0 dispatches in Check = 1 total "dispatches"

pkg/server/commands/list_objects.go
```diff
@@ -369,8 +369,10 @@ func (q *ListObjectsQuery) evaluate(
 						return
 					}
 					atomic.AddUint32(resolutionMetadata.DatastoreQueryCount, resp.GetResolutionMetadata().DatastoreQueryCount)
-					resolutionMetadata.DispatchCounter.Add(reverseExpandResolutionMetadata.DispatchCounter.Load())
-					resolutionMetadata.WasThrottled.Store(reverseExpandResolutionMetadata.WasThrottled.Load())
+					resolutionMetadata.DispatchCounter.Add(checkRequestMetadata.DispatchCounter.Load())
+					if !resolutionMetadata.WasThrottled.Load() && checkRequestMetadata.WasThrottled.Load() {
+						resolutionMetadata.WasThrottled.Store(true)
+					}
```
willvedd, line 375: We also need this in line 313 too

miparnisari (commented):
As part of this change could you move this line

```go
reverseExpandResolutionMetadata := reverseexpand.NewResolutionMetadata()
```

to _inside_ the `go func()` that uses it? 

That way, the variable goes out of scope outside of the goroutine and it will give a compile-time error if you try to use it. (This would have prevented this mistake in the first place)

miparnisari (dismissed):

miparnisari (approved):

Issue #1989, opened 2024-10-03
Replace the in-memory storage implementation with SQLite


aaguiarz:
### Checklist

- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

OpenFGA now supports a SQLite adapter.

It would be good to replace the in-memory storage implementation with SQLite, using the [SQLite in-memory database](https://www.sqlite.org/inmemorydb.html). 

It should be faster, and there would not be the need to maintain the in-memory implementation.

### Describe the ideal solution

1. Drop the `memory` implementation of the datastore and replace it with `SQLite` in-memory solution

- when `engine == "memory"` then SQL [In-Memory mode](https://www.sqlite.org/inmemorydb.html) should be used
- when `engine == "sqlite"` then SQL Filesystem mode should be used

2. Verify that `docker` and other runtimes using the default in-memory storage engine works as expected

RaghavMangla:
Hi, I would like to work on this issue

aaguiarz:
Hi @RaghavMangla 

Thanks a lot! I've assigned it to you

RaghavMangla:
Hi! Can u explain a bit more like to change the in-memory storage implementation with SQLite, based on what i can see, an implementation for using sqlite is already given in assets/migrations/sqlite,

aaguiarz:
The change would be that when OpenFGA is configured to use the in-memory database (e.g.  `--datastore-engine memory`), it should use instantiate SQLite storage adapter with an in-memory database, and run the migrations so the in-memory tables are created.

Makes sense?

RaghavMangla:
Yes

RaghavMangla:
Hi! needed a little help, I have managed to use SQLite in-memory storage for 'memory' engine as shown on console, 
![Image](https://github.com/user-attachments/assets/72d7e9ca-da7e-4c16-be39-9cab63195f7c)

I have made the following changes
1. in cmd/root.go I have set datastoreEngineFlag to sqlite and set datasetURIFlag to file::memory:?cache=shared
2. in cmd/run/run.go set engine to sqlite
3. replaced datastore engine and datastore uri to sqlite and file::memory:?cache=shared
4. in utils.go added an extra if condition in beginning of function MustBootstrapDatastore
    	if engine == "memory" {
		engine = "sqlite"
		uri := "file::memory:?cache=shared"
[...]

PR #2020, opened 2024-10-15
Add start_time param to changes API

cikasfm:
As a Developer, I want to be able to query the changelog based on a specific time period (start date), so that I can retrieve only the changes made after that specific point in time. This will improve the efficiency of tracking and processing changes for the user. 

Why are we doing this? The changelog contains all changes made to resources over time. Providing the ability to filter these changes by a specific time ensures that developers can query only the relevant data, making the API responses lighter, faster and more efficient. 

The changelog can be filtered using the start date as an additional parameter in the API, SDKs,and CLI

The start time parameter should be in the [ISO 8601 format](https://www.progress.com/blogs/understanding-iso-8601-date-and-time-format)

Example: YYYY-MM-DDTHH:MM:SSZ (ISO 8601 DateTime)

## Description

The change contains:
- API changes openfga/api#204 upgrade
- Implementation of the extra filter in all SQL providers and also in-memory DB driver

## References

- API changes openfga/api#204

## Review Checklist


pkg/storage/storage.go
```diff
@@ -300,6 +300,7 @@ type AssertionsBackend interface {
 type ReadChangesFilter struct {
 	ObjectType    string
 	HorizonOffset time.Duration
+	StartTime     *time.Time
```
jpadilla, line 303: **thought**: Will time.Time's zero value (`January 1, year 1, 00:00:00 UTC`) consider a valid value for StartTime through the API? If not, we can possibly just drop the pointer here altogether.

cikasfm: Replaced with reference

pkg/server/commands/read_changes.go
```diff
@@ -63,12 +63,17 @@ func (q *ReadChangesQuery) Execute(ctx context.Context, req *openfgav1.ReadChang
...
+	if req.GetStartTime() != nil {
+		startTime = req.GetStartTime().AsTime()
+	}
 	opts := storage.ReadChangesOptions{
 		Pagination: storage.NewPaginationOptions(req.GetPageSize().GetValue(), string(decodedContToken)),
 	}
 	filter := storage.ReadChangesFilter{
 		ObjectType:    req.GetType(),
 		HorizonOffset: q.horizonOffset,
+		StartTime:     startTime,
```
elbuo8, line 76: This should be ignored if there is `continuationToken`

pkg/storage/test/tuples.go
```diff
@@ -69,6 +69,62 @@ func ReadChangesTest(t *testing.T, datastore storage.OpenFGADatastore) {
 		})
 	})
 
+	t.Run("read_changes_with_start_time", func(t *testing.T) {
```
elbuo8, line 72: Missing a test when both `start_time` and `continuationToken` is provided, it should give precedence to the token.

poovamraj: Should we give precedence to the token or error out saying both cannot be provided? Erroring out would be more consistent but less flexible.

jpadilla: I expect it could be somewhat common to append to existing query params:

```
> ReadChanges?start_time=2021-01-01T00:00:00.000Z
< { changes:[...], continuation_token: "123" }

> ReadChanges?start_time=2021-01-01T00:00:00.000Z&continuation_token=123
< { changes:[...], continuation_token: "456" }
```

maybe we just validate if `start_time < continuation_token`

cikasfm: good point @elbuo8 I'll add one 👌

elbuo8: > maybe we just validate if start_time < continuation_token

Keep in mind that we now support sorting, so depending on the sort direction the validation would change.

cikasfm: @elbuo8 @jpadilla @poovamraj I guess this test is not super clear. It will actually do _paging_ to list all pages, meaning uses the start time first, then continuation token, so in a sense all lines are covered.

Let me think of a way to explicitly validate the results

cikasfm: Resolved by dropping the requirement for `start_time < continuation_token` and suggesting that `continuation token takes precedence over start time`

rhamzeh (approved):

internal/mocks/mock_cache.go
```diff
@@ -20,7 +20,6 @@ import (
 type MockInMemoryCache[T any] struct {
 	ctrl     *gomock.Controller
 	recorder *MockInMemoryCacheMockRecorder[T]
-	isgomock struct{}
```
elbuo8, line 23: 🤔

cikasfm: @elbuo8 this was generated after the last rebase

pkg/server/commands/read_changes.go
```diff
@@ -63,8 +65,26 @@ func (q *ReadChangesQuery) Execute(ctx context.Context, req *openfgav1.ReadChang
...
 		return nil, serverErrors.InvalidContinuationToken
 	}
+	token := string(decodedContToken)
+
+	var startTime time.Time
+	if req.GetStartTime() != nil {
+		startTime = req.GetStartTime().AsTime()
+	}
+	if token == "" && !startTime.IsZero() {
+		tokenUlid := ulid.MustNew(ulid.Timestamp(startTime), nil).String()
```
elbuo8, line 75: Are there any guarantees that this won't fail? As in this is a user provided value, and yes there is validation layer, but using `MustNew` rather than `New` seems a bit insecure. I'd argue that `MustNew` should only be used in tests.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -150,6 +150,12 @@ type ContToken struct {
 	ObjectType string `json:"ObjectType"`
 }
 
+// ContToken represents a continuation token structure used in pagination.
+type ContTokenLegacy struct {
```
elbuo8, line 154: Why is this called `Legacy`

pkg/storage/storage.go
```diff
@@ -314,6 +317,9 @@ type ChangelogBackend interface {
 	// if no changes are found, it should return storage.ErrNotFound and an empty continuation token.
 	// It the objectType and the type in the continuation token don't match, it should return ErrMismatchObjectType.
 	ReadChanges(ctx context.Context, store string, filter ReadChangesFilter, options ReadChangesOptions) ([]*openfgav1.TupleChange, []byte, error)
+
+	// CreateContinuationToken creates a continuation token for the given ULID and object type.
+	CreateContinuationToken(ulid string, objType string) ([]byte, error)
```
elbuo8, line 322: Belongs to the Changelog interface but its still in a shared package, please rename to something like `CreateChangelogContinuationToken` as other backends might start embracing this pattern and names would collide.

pkg/storage/memory/memory.go
```diff
@@ -242,14 +243,21 @@ func (s *MemoryBackend) ReadChanges(ctx context.Context, store string, filter st
 		return nil, nil, storage.ErrMismatchObjectType
 	}
 
-	var allChanges []*openfgav1.TupleChange
+	var allChanges []*TupleChangeRec
 	now := time.Now().UTC()
-	for _, change := range s.changes[store] {
-		if objectType == "" || (strings.HasPrefix(change.GetTupleKey().GetObject(), objectType+":")) {
-			if change.GetTimestamp().AsTime().After(now.Add(-horizonOffset)) {
+	for _, changeRec := range s.changes[store] {
```
elbuo8, line 248: Why the need for this change? There are operations already being done against the timestamp like horizon trimming, the same can be done by converting the ULID to time and comparing across the timestamps, no need to have additional struct or memory for something that is already encoded into `TupleChange`. Number of total operations would be the same too since you'd only have to convert to Timestamp once to calculate against horizon and `from`.

pkg/storage/sqlcommon/sqlcommon.go
elbuo8: Can't find any usage for it either 🤔

tests/functional_test.go
```diff
@@ -997,6 +998,13 @@ func GRPCListObjectsTest(t *testing.T, client openfgav1.OpenFGAServiceClient) {
 }
 
 func GRPCListUsersValidationTest(t *testing.T, client openfgav1.OpenFGAServiceClient) {
+	_101tuples := make([]*openfgav1.TupleKey, 101)
```
elbuo8, line 1001: Why is this here? Doesn't seem relevant to this PR as contextual tuples are being handled elsewhere.

internal/mocks/mock_cache.go
cikasfm: @elbuo8 here's the cause: https://github.com/uber-go/mock/pull/204

I was the first to generate mocks

pkg/storage/sqlcommon/sqlcommon.go
cikasfm: ah, crap, this wasn't supposed to get in

pkg/storage/memory/memory.go
cikasfm: @elbuo8 `timestamp` didn't work for me because 
- in memory impl will use the same timestamp for the batch writes
- ulid generated from timestamp drops nanoseconds, in a test all write happen within the same millisecond, so paging becomes impossible

cikasfm: @elbuo8 also we're planning to drop `in-mem` implementation by replacing it with SQLite in-mem #1989

tests/functional_test.go
elbuo8: Then coordinate with @rhamzeh since this won't be the only PR affected.

pkg/storage/memory/memory.go
elbuo8: > by replacing it with SQLite in-mem

Been hearing that for a while :trollface: 

> ulid generated from timestamp drops nanoseconds, in a test all write happen within the same millisecond, so paging becomes impossible

Understandable 👍🏽

tests/functional_test.go
cikasfm: @elbuo8 opened #2041 I'll rebase after it's merged

pkg/server/commands/read_changes.go
DanCech: Good call, much safer to use New and return an error if it fails.

pkg/storage/memory/memory.go
```diff
@@ -834,3 +857,7 @@ func (s *MemoryBackend) ListStores(ctx context.Context, options storage.ListStor
 func (s *MemoryBackend) IsReady(context.Context) (storage.ReadinessStatus, error) {
 	return storage.ReadinessStatus{IsReady: true}, nil
 }
+
+func (s *MemoryBackend) SerializeReadChangesContToken(next string, objType string) ([]byte, error) {
+	return []byte(fmt.Sprintf("%s|%s", next, objType)), nil
+}
```
DanCech, line 863: I see that the sql-based storages all use a standard continuation token struct and shared marshal/unmarshal functions, should we move those out of sqlcommon and into storage, then use it everywhere?

elbuo8 (dismissed):

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -159,6 +159,11 @@ func NewContToken(ulid, objectType string) *ContToken {
 	}
 }
 
+// MarshallContToken takes a ContToken struct and attempts to marshal it into a string.
+func MarshallContToken(from *ContToken) ([]byte, error) {
+	return json.Marshal(from)
+}
+
```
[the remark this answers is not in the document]
cikasfm: At this point my goal was to remove `json` dependency from the database driver, plus there was already an `UnmarshalContToken` method.

pkg/server/commands/read_changes.go
```diff
@@ -63,8 +65,29 @@ func (q *ReadChangesQuery) Execute(ctx context.Context, req *openfgav1.ReadChang
...
 		return nil, serverErrors.InvalidContinuationToken
 	}
+	token := string(decodedContToken)
+
+	var startTime time.Time
+	if req.GetStartTime() != nil {
+		startTime = req.GetStartTime().AsTime()
+	}
+	if token == "" && !startTime.IsZero() {
+		tokenUlid, ulidErr := ulid.New(ulid.Timestamp(startTime), nil)
```
jpadilla, line 75: Anything we should note/comment about the importance of `nil` entropy?

pkg/storage/memory/memory.go
```diff
@@ -327,6 +339,11 @@ func (s *MemoryBackend) read(ctx context.Context, store string, tk *openfgav1.Tu
 	return &staticIterator{records: matches}, nil
 }
 
+type TupleChangeRec struct {
```
jpadilla, line 342: Does this need to be exported?

pkg/storage/storage.go
```diff
@@ -314,6 +317,9 @@ type ChangelogBackend interface {
 	// if no changes are found, it should return storage.ErrNotFound and an empty continuation token.
 	// It the objectType and the type in the continuation token don't match, it should return ErrMismatchObjectType.
 	ReadChanges(ctx context.Context, store string, filter ReadChangesFilter, options ReadChangesOptions) ([]*openfgav1.TupleChange, []byte, error)
+
+	// SerializeReadChangesContToken serializes the continuation token into a format readable by ReadChanges
+	SerializeReadChangesContToken(ulid string, objType string) ([]byte, error)
```
jpadilla, line 322: The addition of `SerializeReadChangesContToken` into this interface seems kinda leaky imho. It's usage outside of storage adapter implementations themselves is in `pkg/server/commands/read_changes.go`.

I wonder if there is an alternative implementation where we don't need this interface change?

pkg/storage/memory/memory.go
cikasfm: No. Will update to private (still getting used to `golang` style)

pkg/server/commands/read_changes.go
cikasfm: At this point it only impacts filtering records within the same millisecond. Starting from zero makes sense for filtering by time.

cmd/run/run.go
```diff
@@ -390,36 +390,41 @@ func (s *ServerContext) datastoreConfig(config *serverconfig.Config) (storage.Op
...
+			return nil, nil, fmt.Errorf("initialize sqlite datastore: %w", err)
 		}
 	default:
-		return nil, fmt.Errorf("storage engine '%s' is unsupported", config.Datastore.Engine)
+		return nil, nil, fmt.Errorf("storage engine '%s' is unsupported", config.Datastore.Engine)
 	}
 
 	s.Logger.Info(fmt.Sprintf("using '%v' storage engine", config.Datastore.Engine))
-	return datastore, nil
+	// hack no??
```
jpadilla, line 426:
```suggestion

```

cmd/run/run.go
```diff
@@ -611,6 +616,7 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 
 	svr := server.MustNewServerWithOpts(
 		server.WithDatastore(datastore),
+		server.WithTokenSerializer(continuationTokenSerializer),
```
jpadilla, line 619: should this be a little more specific?

```suggestion
		server.WithContinuationTokenSerializer(continuationTokenSerializer),
```

pkg/storage/mysql/mysql.go
```diff
@@ -199,11 +204,11 @@ func (s *Datastore) read(ctx context.Context, store string, tupleKey *openfgav1.
 		sb = sb.Where(sq.Eq{"_user": tupleKey.GetUser()})
 	}
 	if options != nil && options.Pagination.From != "" {
-		token, err := sqlcommon.UnmarshallContToken(options.Pagination.From)
+		token, _, err := s.tokenSerializer.DeserializeContinuationToken(options.Pagination.From)
```
jpadilla, line 207: should this be a bit more specific or do we see the need for being more generic?

```suggestion
		token, _, err := s.continuationTokenSerializer.Deserialize(options.Pagination.From)
```

cikasfm: I like that. If we're not merging any interfaces, this is better.

cmd/run/run.go
```diff
@@ -394,32 +398,36 @@ func (s *ServerContext) datastoreConfig(config *serverconfig.Config) (storage.Op
 	var err error
 	switch config.Datastore.Engine {
 	case "memory":
+		// override for "memory" datastore
+		tokenSerializer = encoder.NewStringContinuationTokenSerializer()
```
jpadilla, line 402: Is there any reason we can't use the same `NewSQLContinuationTokenSerializer` here?

cikasfm: We can! I have it tested [here](https://github.com/openfga/openfga/pull/2020/files#diff-4b51ba45629b21e0486543c70a4671f3174aa81efc31290c89e174681814fe26R23-R33)

jpadilla (approved):

internal/mocks/mock_token_serializer.go
```diff
@@ -0,0 +1,70 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: token_serializer.go
+//
+// Generated by this command:
+//
+//	mockgen -source token_serializer.go -destination ../../internal/mocks/mock_token_serializer.go -package mocks OpenFGADatastore
```
miparnisari, line 6: @cikasfm did you run this command by hand?

PLease add this line to the `token_serializer.go` file to generate this on every build: `//go:generate mockgen -source token_serializer.go -destination ../../internal/mocks/mock_token_serializer.go -package mocks`

pkg/encoder/token_serializer.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+type ContinuationTokenSerializer interface {
+	// Serialize serializes the continuation token into a format readable by ReadChanges
+	Serialize(ulid string, objType string) (token []byte, err error)
+
+	// Deserialize deserializes the continuation token into a format readable by ReadChanges
+	Deserialize(token string) (ulid string, objType string, err error)
```
miparnisari, line 15: If in the future you want to introduce the same artifact for the ReadPage API or ListObjects API, what will you have to change? You will have to create a new interface or add new methods. The existing interface name makes it sound like it will work for every API (but it won't), and new methods will have to be called `DeserializeForListObjects` to distinguish them from the existing methods. 

I would have done this like:

```go
type ReadChangesContinuationTokenSerializer interface {
   Seralize(...)
   Deserialize(..)
}
```

or 

```go
type ContinuationTokenSerializer interface {
   SeralizeForReadChanges(...)
   DeserializeForReadChanges(..)
}
```

pkg/server/commands/read_changes.go
```diff
@@ -15,10 +17,11 @@ import (
 )
 
 type ReadChangesQuery struct {
-	backend       storage.ChangelogBackend
-	logger        logger.Logger
-	encoder       encoder.Encoder
-	horizonOffset time.Duration
+	backend         storage.ChangelogBackend
+	logger          logger.Logger
+	encoder         encoder.Encoder
```
miparnisari, line 22: Should we rename this to `encoder.TokenEncoder`? :)

pkg/storage/errors.go
```diff
@@ -16,6 +16,8 @@ var (
 	// ErrInvalidContinuationToken is returned when the continuation token is invalid.
 	ErrInvalidContinuationToken = errors.New("invalid continuation token")
 
+	ErrInvalidStartTime = errors.New("invalid start time")
```
miparnisari, line 19: Please add godoc for this

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -30,6 +32,8 @@ type Config struct {
 	Logger                 logger.Logger
 	MaxTuplesPerWriteField int
 	MaxTypesPerModelField  int
+	// for serializing the read changes token
+	TokenSerializer encoder.ContinuationTokenSerializer
```
miparnisari, line 36: The fact that you had to leave this comment reinforces my opinion here: https://github.com/openfga/openfga/pull/2020/files#r1824830100

pkg/storage/memory/memory_test.go
```diff
@@ -13,14 +13,23 @@ import (
...
-	test.RunAllTests(t, ds)
+	tokenSerializer := encoder.NewStringContinuationTokenSerializer()
+	ds := New(WithContinuationTokenSerializer(tokenSerializer))
+	test.RunAllTests(t, ds, tokenSerializer)
+}
+
+func TestMemdbStorageWithSqlTokenSerializer(t *testing.T) {
+	tokenSerializer := sqlcommon.NewSQLContinuationTokenSerializer()
+	ds := New(WithContinuationTokenSerializer(tokenSerializer))
+	t.Run("TestReadChanges", func(t *testing.T) { test.ReadChangesTest(t, ds, tokenSerializer) })
```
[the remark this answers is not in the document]
cikasfm: This is checking if it will work with a different serializer implementation

pkg/storage/errors.go
cikasfm: noted.

pkg/server/commands/read_changes.go
cikasfm: Right...tbh I'm thinking if we can remove the encoder and merge it with the serializer. that might be a bit more effort though

pkg/storage/sqlcommon/sqlcommon.go
cikasfm: Out of date in #2064

pkg/encoder/token_serializer.go
cikasfm: This change was made based on [Jose's comment](https://github.com/openfga/openfga/pull/2020#discussion_r1821402533)

Issue #1477, opened 2024-03-20
Use an atomic counter for datastore reads in Check


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Original discussion: https://github.com/openfga/openfga/pull/1452#discussion_r1526658647

The code that calculates how many database reads were accrued in Check is hard to follow and is error-prone.

### Describe the ideal solution

The proposed solution is to use an `atomic.uint32` counter for the `datastore_query_count`, similar to `dispatch_counter` (https://github.com/openfga/openfga/blob/b6f3d4a2461302fab10ddef2eb681c096aed7233/internal/graph/graph.go#L58)

Pros: 

- more readable code

Cons: 

- we are introducing a mutex, which introduces slight delays in the API
- goes against the future idea of introducing remote peer dispatching, because we can't share an atomic counter across nodes.

PR #2022, opened 2024-10-16
refactor: track datastore query count in datastore wrapper

miparnisari:
## Description

**Problem**: Keeping track of `datastore_query_count` value within Check, ReverseExpand, ListObjects and ListUsers is tedious, bloats the code, breaks Single Responsibility Principle, and worse of all will mean breaking tests whenever we reduce the number of DB calls made.

**Proposal of this PR**: 

- keep track of `datastore_query_count` in a wrapper over the `OpenFGAStorage` object.
- `ResolutionMetadata.DatastoreQueryCount` is only filled at the end of the request.

⚠️ important files to review:
- `pkg/server/commands/check_command.go`
- `pkg/server/commands/list_objects.go`
- `pkg/server/commands/listusers/list_users_rpc.go`

⚠️ as a side effect of this change, Check API will change as follows: before, `datastore_query_count` was increased per path in the resolution tree. Now, it's increased globally. This means that we are now counting all the paths that we explored that didn't result in `allowed=true`.

## References

Closes https://github.com/openfga/openfga/issues/1477


internal/graph/check.go
```diff
@@ -230,8 +226,7 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
 	return &ResolveCheckResponse{
 		Allowed: false,
 		ResolutionMetadata: &ResolveCheckResponseMetadata{
```
miparnisari, line 228: i want to send another PR where `ResolveCheckResponseMetadata` is not a pointer.

pkg/server/commands/list_objects.go
```diff
@@ -270,11 +269,12 @@ func (q *ListObjectsQuery) evaluate(
 		reverseExpandResultsChan := make(chan *reverseexpand.ReverseExpandResult, 1)
 		objectsFound := atomic.Uint32{}
 
+		metricsDs := storagewrappers.NewMetricsOpenFGAStorage(q.datastore)
 		ds := storagewrappers.NewCombinedTupleReader(
-			q.datastore,
+			storagewrappers.NewBoundedConcurrencyTupleReader(
+				metricsDs, q.maxConcurrentReads),
 			req.GetContextualTuples().GetTupleKeys(),
 		)
```
miparnisari, line 277: FYI i dislike how this looks. Discussed offline with @jpadilla, the current problem we have is that some datastore wrappers are server-wide, while others are request-specific.

What we could do in a future PR is to have every query object create it's own datastore wrapper _from scratch_. The server should _only_ hold any cache objects that those wrappers need.

pkg/storage/storagewrappers/metrics.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+}
+
+func (m *MetricsOpenFGAStorage) increase() {
+	m.counter.Add(1)
+}
+
+// Read see [storage.RelationshipTupleReader.ReadUserTuple].
+func (m *MetricsOpenFGAStorage) Read(ctx context.Context, store string, tupleKey *openfgav1.TupleKey, options storage.ReadOptions) (storage.TupleIterator, error) {
+	defer m.increase()
```
elbuo8, line 42: Why use `defer`? It's just an addition to the call stack with no point 🤔 More operations overall to process.

internal/mocks/mock_storage.go
```diff
@@ -22,7 +22,6 @@ import (
 type MockTupleBackend struct {
 	ctrl     *gomock.Controller
 	recorder *MockTupleBackendMockRecorder
-	isgomock struct{}
```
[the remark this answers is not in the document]
jpadilla: expected https://github.com/uber-go/mock/blob/main/CHANGELOG.md#050-15-oct-2024

jpadilla (dismissed):

pkg/server/commands/check_command.go
```diff
@@ -30,7 +30,7 @@ type CheckQuery struct {
 	logger          logger.Logger
 	checkResolver   graph.CheckResolver
 	typesys         *typesystem.TypeSystem
-	datastore       storage.RelationshipTupleReader
+	datastore       *storagewrappers.MetricsOpenFGAStorage
```
elbuo8, line 33: This shouldn't change at all.

pkg/server/commands/check_command.go
```diff
@@ -74,7 +74,7 @@ func WithCacheController(ctrl cachecontroller.CacheController) CheckQueryOption
 func NewCheckCommand(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, typesys *typesystem.TypeSystem, opts ...CheckQueryOption) *CheckQuery {
 	cmd := &CheckQuery{
 		logger:             logger.NewNoopLogger(),
-		datastore:          datastore,
+		datastore:          storagewrappers.NewMetricsOpenFGAStorage(datastore),
```
elbuo8, line 77: Store this in a var above, so the interface doesn't need to change, and subsequently have you have access to the additional methods you require.

pkg/storage/storagewrappers/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+// It is crucial that the wrapped object does NOT return results from an in-memory cache for this object to return accurate metrics.
+func NewMetricsOpenFGAStorage(wrapped storage.RelationshipTupleReader) *MetricsOpenFGAStorage {
+	return &MetricsOpenFGAStorage{
+		RelationshipTupleReader: wrapped,
+		counter:                 atomic.Uint32{},
+	}
+}
+
+type Metrics struct {
+	DatastoreQueryCount int
```
elbuo8, line 30: If you are casting it everywhere, why not use `uint32`?

pkg/server/commands/check_command.go
miparnisari: If i have the command object keep a reference to both `datastore` and `instrumentedDatastore`, then i wouldn't reference the former anywhere 🤔

elbuo8 (approved):

internal/graph/check.go
miparnisari: https://github.com/openfga/openfga/pull/2061

PR #2017, opened 2024-10-15
refactor: Speed up check benchmark tests

justincoh:
## Description

This PR is POC for moving our old check benchmark tests to use the new `check_command` format. The old benchmarks required a server to be set up and torn down for each benchmark and handled all tuple writing and checks via grpc requests.

The new benchmark invokes the `CheckQuery` directly, removing all that overhead. You'll notice the new benchmarks show significantly higher performance. My hunch is that these benchmarks are now more accurately representing the true performance of Check without server overhead, but please speak up if it seems suspicious.

This change also shaves significant time off of the benchmark run; for the two scenarios modified in this PR it saves 15-20s versus the old way. After converting all of the check benchmarks in this manner we should save several minutes per build.

Resolves #1172.

Old Benchmark run for this PR, with the server (I removed the `/healthz` log lines):
```
 🚀  Running benchmark tests
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/tests/check
cpu: Apple M3 Max
BenchmarkCheckPostgres/benchmarkCheckDirectAndUserset/race_between_direct_and_userset_check       	   16194	    363830 ns/op	  101476 B/op	    1612 allocs/op
BenchmarkCheckPostgres/benchmarkCheckDirectAndUserset/userset_check_only                          	    6828	    867534 ns/op	  124493 B/op	    2013 allocs/op
BenchmarkCheckMemory/benchmarkCheckDirectAndUserset/race_between_direct_and_userset_check         	     240	  24985205 ns/op	   66489 B/op	     735 allocs/op
BenchmarkCheckMemory/benchmarkCheckDirectAndUserset/userset_check_only                            	     240	  24992544 ns/op	   57375 B/op	     746 allocs/op
BenchmarkCheckMySQL/benchmarkCheckDirectAndUserset/race_between_direct_and_userset_check          	   12013	    488563 ns/op	   84270 B/op	    1318 allocs/op
BenchmarkCheckMySQL/benchmarkCheckDirectAndUserset/userset_check_only                             	     381	  15141481 ns/op	  120274 B/op	    1919 allocs/op
BenchmarkCheckSQLite/benchmarkCheckDirectAndUserset/race_between_direct_and_userset_check         	     240	  25008243 ns/op	  114564 B/op	    1966 allocs/op
BenchmarkCheckSQLite/benchmarkCheckDirectAndUserset/userset_check_only                            	     240	  25018211 ns/op	  126357 B/op	    2181 allocs/op
PASS
[...]

PR #2024, opened 2024-10-16
refactor: speed up check benchmark tests part 2

### LINKED ISSUE #1172: Speed up Check benchmarks

justincoh:
## Description

Follow up to #2017, Resolves #1172 

This PR is the same as #2017, it just moves all the remaining check benchmarks over. Pasting the description from #2017:
 > This PR is POC for moving our old check benchmark tests to use the new check_command format. The old benchmarks required a server to be set up and torn down for each benchmark and handled all tuple writing and checks via grpc requests.

> The new benchmark invokes the CheckQuery directly, removing all that overhead. You'll notice the new benchmarks show significantly higher performance. My hunch is that these benchmarks are now more accurately representing the true performance of Check without server overhead, but please speak up if it seems suspicious.

> This change also shaves significant time off of the benchmark run; for the two scenarios modified in this PR it saves 15-20s versus the old way. After converting all of the check benchmarks in this manner we should save several minutes per build.

Rough end result: on my local the benchmark now takes 6-7 minutes instead of ~15-17. In gh actions the benchmarks are running in about 10 min

## Review Checklist


pkg/server/test/check.go
```diff
@@ -146,15 +391,105 @@ func BenchmarkCheck(b *testing.B, ds storage.OpenFGADatastore) {
...
+	require.NoError(b, err)
+
+	// all the usersets added above are now invalid and should be skipped!
+	err = ds.WriteAuthorizationModel(context.Background(), storeID, modelTwo)
+	require.NoError(b, err)
+
+	checkQuery := commands.NewCheckCommand(
+		ds,
+		graph.NewLocalChecker(),
+		typeSystemTwo,
```
justincoh, line 479: I _think_ the second typesystem is accomplishing the same goal as the original test scenario but someone please double check me

miparnisari (approved):

pkg/server/test/check.go
```diff
@@ -146,15 +391,105 @@ func BenchmarkCheck(b *testing.B, ds storage.OpenFGADatastore) {
...
-					TupleKey:         bm.tupleKey,
-					ContextualTuples: bm.contextualTuples,
-					Context:          bm.contextStruct,
+					StoreId:  storeID,
+					TupleKey: bm.tupleKeyToCheck,
+					Context:  bm.contextGenerator(),
 				})
 
 				require.Equal(b, bm.expected, response.GetAllowed())
 				require.NoError(b, err)
```
miparnisari, line 400: these two lines should be in the opposite order

pkg/server/test/check.go
```diff
@@ -137,7 +382,7 @@ func BenchmarkCheck(b *testing.B, ds storage.OpenFGADatastore) {
 
 		checkQuery := commands.NewCheckCommand(
 			ds,
-			bm.checker,
+			graph.NewLocalChecker(),
 			typeSystem,
 			commands.WithCheckCommandMaxConcurrentReads(config.DefaultMaxConcurrentReadsForCheck),
```
miparnisari, line 387: `config.DefaultMaxConcurrentReadsForCheck` is `math.MaxUint32`. This can overwhelm the server on some queries (e.g. `with_complex_ttu`). Let's make this smaller, e.g. 50 :)

pkg/server/test/check.go
```diff
@@ -137,7 +382,7 @@ func BenchmarkCheck(b *testing.B, ds storage.OpenFGADatastore) {
 
 		checkQuery := commands.NewCheckCommand(
 			ds,
-			bm.checker,
+			graph.NewLocalChecker(),
```
miparnisari, line 385: `graph.NewLocalChecker(WithOptimizations(true))`

pkg/server/test/check.go
justincoh: indeed they should

PR #1857, opened 2024-08-16
refactor: introduce Check command

miparnisari:
## Description
Every API on our codebase has an underlying `command` struct that represents it, except for Check (we deviated [here](https://github.com/openfga/openfga/pull/476))

This PR adds a command so that:
- we follow the [command design pattern](https://refactoring.guru/design-patterns/command) more consistently across the codebase
- testing is easier as now, we can invoke the command (as opposed to doing `NewLocalChecker.ResolveCheck` which lacks validation and context building)
- more code coverage

I've attempted to keep everything as is in terms of functionality, even when I know it's incorrect.

PR #2031, opened 2024-10-18
refactor: check command execute params

justincoh:
## Description

The upcoming BatchCheck work will be calling the CheckCommand, which currently requires its `.Execute` method to receive a `CheckRequest`. Passing the full request object isn't necessary; this PR pulls the required pieces of the `CheckRequest` into a struct and passes that struct into `.Execute` instead to increase reusability.

## References

Creation of Check Command: https://github.com/openfga/openfga/pull/1857

## Review Checklist

### DIFF
--- a/pkg/server/commands/check_command.go
+++ b/pkg/server/commands/check_command.go
@@ -9,13 +9,11 @@ import (
 	"github.com/openfga/openfga/internal/cachecontroller"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/openfga/openfga/internal/condition"
 	"github.com/openfga/openfga/internal/graph"
 	"github.com/openfga/openfga/internal/validation"
-	"github.com/openfga/openfga/pkg/middleware/validator"
 	serverErrors "github.com/openfga/openfga/pkg/server/errors"
 	"github.com/openfga/openfga/pkg/storage"
 	"github.com/openfga/openfga/pkg/storage/storagewrappers"
@@ -39,6 +37,14 @@ type CheckQuery struct {
 
 	resolveNodeLimit   uint32
 	maxConcurrentReads uint32
+}
+
+type CheckRequestParams struct {
+	StoreID          string
+	TupleKey         *openfgav1.CheckRequestTupleKey
+	ContextualTuples *openfgav1.ContextualTupleKeys
+	Context          *structpb.Struct
+	Consistency      openfgav1.ConsistencyPreference
 }
 
 type CheckQueryOption func(*CheckQuery)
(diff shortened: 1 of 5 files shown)


pkg/server/commands/check_command.go
```diff
@@ -118,20 +124,14 @@ func (c *CheckQuery) Execute(ctx context.Context, req *openfgav1.CheckRequest) (
 	return resp, resolveCheckRequest.GetRequestMetadata(), nil
 }
 
-func validateCheckRequest(ctx context.Context, req *openfgav1.CheckRequest, typesys *typesystem.TypeSystem) error {
-	if !validator.RequestIsValidatedFromContext(ctx) {
-		if err := req.Validate(); err != nil {
-			return status.Error(codes.InvalidArgument, err.Error())
-		}
-	}
-
```
justincoh, line 127: this request-level validation is happening in `server.go` now

pkg/server/commands/check_command_test.go
```diff
@@ -43,33 +43,10 @@ type doc
...
-			TupleKey: &openfgav1.CheckRequestTupleKey{
-				User:     "user:1",
-				Relation: "viewer",
-				Object:   "invalid:1",
-			},
-			AuthorizationModelId: "invalid",
-		})
-		require.ErrorContains(t, err, "invalid CheckRequest.AuthorizationModelId: value does not match regex pattern \"^[ABCDEFGHJKMNPQRSTVWXYZ0-9]{26}$\"")
-	})
-
```
justincoh, line 67: these two have been moved to `server_test.go`

pkg/server/commands/check_command_test.go
```diff
@@ -81,9 +58,8 @@ type doc
 
 	t.Run("validates_input_relation", func(t *testing.T) {
 		cmd := NewCheckCommand(mockDatastore, mockCheckResolver, ts)
-		_, _, err := cmd.Execute(context.Background(), &openfgav1.CheckRequest{
-			StoreId:              ulid.Make().String(),
-			AuthorizationModelId: ulid.Make().String(),
```
justincoh, line 86: AuthorizationModelId was unused by the command, removed it

pkg/server/commands/check_command.go
```diff
@@ -41,6 +39,14 @@ type CheckQuery struct {
 	maxConcurrentReads uint32
 }
 
+type CheckRequestParams struct {
```
willvedd, line 42: **Suggestion:**
```suggestion
type CheckCommandParams struct {
```
?

pkg/server/commands/check_command.go
willvedd: **Question:** Does this validation still get applied if openfga is consumed as a library as opposed to as an API?

pkg/server/commands/check_command.go
```diff
@@ -118,20 +124,14 @@ func (c *CheckQuery) Execute(ctx context.Context, req *openfgav1.CheckRequest) (
...
 }
 
-func validateCheckRequest(ctx context.Context, req *openfgav1.CheckRequest, typesys *typesystem.TypeSystem) error {
-	if !validator.RequestIsValidatedFromContext(ctx) {
-		if err := req.Validate(); err != nil {
-			return status.Error(codes.InvalidArgument, err.Error())
-		}
-	}
-
+func validateCheckRequest(typesys *typesystem.TypeSystem, tupleKey *openfgav1.CheckRequestTupleKey, ctxTuples *openfgav1.ContextualTupleKeys) error {
```
willvedd, line 127: **Nit:**
```suggestion
func validateCheckRequest(typesys *typesystem.TypeSystem, tupleKey *openfgav1.CheckRequestTupleKey, contextualTuples *openfgav1.ContextualTupleKeys) error {
```

Let's try to separate the ideas of context and contextual tuples. I worry that `ctx` conflates them.

willvedd (commented):
Agree with the spirit of the change. This decoupling probably should've happened in the original check command refactor. A couple of small, non-blocking suggestions. Mostly concerned about the removal of the validation and if it still gets applied when used as a library.

pkg/server/commands/check_command.go
justincoh: good question, it does. Running as a library is done [like this](https://pkg.go.dev/github.com/openfga/openfga/pkg/server#example-NewServerWithOpts), which means you will get the request-level validation as part of the server handler

willvedd (approved):
Great job 👍

pkg/server/commands/check_command_test.go
miparnisari: 🤔 why is `server_test.go` a better location for these tests, and for the code itself?

justincoh: validateCheckRequest was doing two related but separate things:

1. grpc contract validation
2. actual check logic validation (object relation user)

This breaks it apart so the [grpc pattern validation](https://github.com/openfga/api/blob/main/openfga/v1/openfga_service.proto#L1107) happens in the server handler before calling the check_command, leaving the check_command responsible only for things related to the logic of check. This makes the check_command entirely separate from the inbound request, which makes sense—if someone sends us an invalid `StoreId` there's no reason to even start a check.

miparnisari: I won't block the PR for this, since this is very minor. But i think both validations you mentioned are input validations and I see no reason to put everything inside each individual command.

PR #2032, opened 2024-10-18
feat: nested fast path refactor 

elbuo8:
## Description

Refactors nested TTU and Userset implementations. The overall number of queries produced continues to be the same as the original implementation. Continues to load the first level of usersets along side the terminal type to intersect on before progressing to resolving recursion. 

The main change is preventing N*M traversal. This PR introduces a Breadth First approach bounding the concurrency of the traversal while reducing the overall cost of consumption of intermediate responses and reducing overall contention of the `sync` objects. The consumer of intermediate responses follows the exact flow we have in `consumeDispatches` and `consumeUsersets`, tho lining them up to be refactored into a single method across all flows.

It also fixes a data race introduced in the refactor of `combinedIterator` in a previous PR which started being exhibited in this PR 😅 

## Review Checklist


internal/graph/tuplemapper.go
```diff
@@ -89,12 +91,44 @@ func (n NestedTTUMapper) doMap(t *openfgav1.TupleKey) (string, error) {
 	return t.GetUser(), nil
 }
 
+type ObjectIDMapper struct {
```
adriantam, line 94: How is this different from NestedTTUMapper?  Would it make sense to just use NestedTTUMapper?  If the name makes it confusing, perhaps rename NestedTTUMapper to be more generic such as tupleKeyMapper?

elbuo8: One returns the `User`, the other returns `Object`

adriantam: Ahhh...got it.

internal/graph/check.go
```diff
@@ -979,227 +982,170 @@ func trySendUsersetsAndDeleteFromMap(ctx context.Context, usersetsMap usersetsMa
...
+// group:1#member@group:3#member
+// group:2#member@group:a#member
+// group:3#member@group:a#member
+// Note that both group:2#member and group:3#member has group:a#member. However, they are not cycles.
+
+func (c *LocalChecker) breadthFirstNestedMatch(ctx context.Context, req *ResolveCheckRequest, mapping *nestedMapping, visitedUserset *sync.Map, currentUsersetLevel []string, usersetFromUser storage.SortedSet, checkOutcomeChan chan checkOutcome) {
+	req.GetRequestMetadata().Depth--
+	if req.GetRequestMetadata().Depth == 0 {
+		concurrency.TrySendThroughChannel(ctx, checkOutcome{err: ErrResolutionDepthExceeded}, checkOutcomeChan)
+		close(checkOutcomeChan)
```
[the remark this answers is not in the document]
elbuo8: There is no concurrency in this function. It must finish all goroutines before scheduling a recursive all once again.

pkg/storage/iterators.go
```diff
@@ -53,28 +55,36 @@ func (c *combinedIterator[T]) Next(ctx context.Context) (T, error) {
...
 	}
 
+	c.mu.Lock()
 	iter := c.pending[0]
 	val, err := iter.Next(ctx)
 	if err != nil {
 		if errors.Is(err, ErrIteratorDone) {
 			c.pending = c.pending[1:]
 			c.done = append(c.done, iter)
+			c.mu.Unlock()
```
elbuo8, line 65: The introduction of this mutex is necessary as `c.pending` and `c.done` get accessed via multiple goroutines and can cause data race issues (and there are other methods which ultimately access these properties such as `.Stop`, thus a mutex is required.

adriantam (approved):

PR #2039, opened 2024-10-23
feat: batch check API

justincoh:
## Description

This PR implements the BatchCheck API. The maximum allowed size of the batch and the maximum allowed concurrent  check operations are configurable with the env vars `OPENFGA_MAX_CHECKS_PER_BATCH_CHECK` and `OPENFGA_MAX_CONCURRENT_CHECKS_PER_BATCH_CHECK` (or with flags passed into the run command, see `flags.go`).

There are currently no tests in this branch, as those are being added in [a separate PR](https://github.com/openfga/openfga/pull/2047) based off of the `feat/batch-check` branch. When all sub-branches have been merged into this one, I will update this description and request an additional round of reviews.

## References

There are no test in this PR, but there are two additional PRs based off of this branch which contain tests:

#2065 - Batch Check Unit tests
#2066 - Batch Check Integration tests

## Review Checklist


pkg/server/batch_check.go
```diff
@@ -0,0 +1,125 @@
...
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	"github.com/openfga/openfga/internal/authz"
+	"github.com/openfga/openfga/pkg/server/commands"
+	"github.com/openfga/openfga/pkg/telemetry"
+)
+
+func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckRequest) (*openfgav1.BatchCheckResponse, error) {
+	ctx, span := tracer.Start(ctx, authz.Read, trace.WithAttributes(
```
jpadilla, line 21:
```suggestion
	ctx, span := tracer.Start(ctx, authz.BatchCheck, trace.WithAttributes(
```

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,141 @@
...
+				Duration:      time.Since(start),
+				Err:           err,
+			}
+			lock.Unlock()
+
+			return nil
+		})
+	}
+
+	_ = pool.Wait()
```
jpadilla, line 138: nit: can we do this closer to NewPool?

```
	ctx, cancel := context.WithCancel(ctx)
	pool := concurrency.NewPool(ctx, int(bq.maxConcurrentChecks))
	defer func() {
		cancel()
		_ = pool.Wait()
	}()
```

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,141 @@
...
+
+			lock.Lock()
+			resultMap[check.GetCorrelationId()] = &BatchCheckOutcome{
+				CheckResponse: response,
+				Duration:      time.Since(start),
+				Err:           err,
+			}
+			lock.Unlock()
+
+			return nil
```
jpadilla, line 134: Could there be a case where we do want to return non-nil? How would something like `context.DeadlineExceeded` or `context.Canceled` behave here?

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckCommandParams) (map[string]*BatchCheckOutcome, error) {
+	if len(params.Checks) > int(bq.maxChecksAllowed) {
+		return nil, errors.ValidationError(
+			fmt.Errorf("batchCheck received %d checks, the maximum allowed is %d ", len(params.Checks), bq.maxChecksAllowed),
+		)
+	}
+
+	// This check query will be run against every check in the batch
+	checkQuery := NewCheckCommand(
```
jpadilla, line 96: This needs to happen on every request or can should it just happen once on `NewBatchCheckCommand`?

pkg/server/batch_check.go
```diff
@@ -0,0 +1,125 @@
...
+		attribute.KeyValue{Key: "store_id", Value: attribute.StringValue(req.GetStoreId())},
+		attribute.KeyValue{Key: "batch_size", Value: attribute.IntValue(len(req.GetChecks()))},
+		attribute.KeyValue{Key: "consistency", Value: attribute.StringValue(req.GetConsistency().String())},
+	))
+	defer span.End()
+
+	err := req.Validate()
```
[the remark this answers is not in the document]
justincoh: It seems like yes.

Restating to make sure I understand it from reading the code: `RequestIsValidatedFromContext` relies on our [middleware](https://github.com/openfga/openfga/blob/main/pkg/middleware/validator/validator.go#L29) having already seen this request and validating it prior. So wrapping `req.Validate` with this check is a way to make sure we're not doing extra work.

Is that right?

pkg/server/commands/batch_check_command.go
justincoh: Will and I actually talked about this the other day. How would we want something like that to behave here? Or would we want something like this at the top of each pool.Go:
```
select {
case <- ctx.Done():
  // return a timeout error from this routine
default:
  // do nothing and let the routine run its check
}
```

pkg/server/batch_check.go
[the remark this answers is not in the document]
jpadilla: yep!

pkg/server/commands/batch_check_command.go
```diff
@@ -71,14 +72,24 @@ func WithBatchCheckMaxChecksPerBatch(maxChecks uint32) BatchCheckQueryOption {
...
 
+	// This check query will be run against every check in the batch
+	cmd.checkCommand = NewCheckCommand(
+		cmd.datastore,
+		cmd.checkResolver,
+		cmd.typesys,
+		WithCheckCommandLogger(cmd.logger),
+		WithCacheController(cmd.cacheController),
+	)
+
```
justincoh: @jpadilla I tested out moving the `NewCheckCommand` bit inside of `NewBatchCheckCommand` and the interesting parts are the non-pointer attributes like `cacheController`.

Our pattern seems to be 1) initialize a noop verion of something 2) optionally override that in a `WithSomeOption()` command. The issue here is that the CheckCommand takes a value in its `cacheController` instead of a pointer. So if we build the CheckCommand at the  start of `NewBatchCheck`, we'll be initializing the CheckCommand with a noopCacheController(). Lmk if you want to chat irl for a few minutes to talk through this but I _think_ this is correct

pkg/server/batch_check.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &openfgav1.BatchCheckResponse{Result: transformCheckResultToRPC(result)}, nil
+}
+
+// transformCheckResultToRPC takes about 150ns per check, or 0.00015ms.
+// If batch sizes ever increase into the thousands, we may want to
+// refactor this to work concurrently alongside the checks themselves.
```
willvedd, line 82: **Request 1:**  I definitely see why you make this note butt he doc comment shouldn't include these performance notes. Instead, should describe the function. 

**Request 2:**  Because performance is a concern here, can you add a benchmark test for it? Also it looks like it needs unit tests too?

pkg/server/batch_check.go
```diff
@@ -0,0 +1,131 @@
...
+func transformCheckResultToRPC(checkResults map[string]*commands.BatchCheckOutcome) map[string]*openfgav1.BatchCheckSingleResult {
```
willvedd, line 83: **Request:** Add unit tests, lots of missing coverage reported.

pkg/server/batch_check.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+		}
+
+		batchResult[k] = singleResult
+	}
+
+	return batchResult
+}
+
+func transformCheckCommandErrorToBatchCheckError(cmdErr error) *openfgav1.CheckError {
```
willvedd, line 105: **Request:** Unit tests for this.

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,162 @@
...
+func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckCommandParams) (map[string]*BatchCheckOutcome, error) {
+	if len(params.Checks) > int(bq.maxChecksAllowed) {
+		return nil, errors.ValidationError(
+			fmt.Errorf("batchCheck received %d checks, the maximum allowed is %d ", len(params.Checks), bq.maxChecksAllowed),
+		)
+	}
+
+	if err := validateNoDuplicateCorrelationIDs(params.Checks); err != nil {
+		return nil, err
```
willvedd, line 97: **Question:** Additional validations need to be made on the correlation IDs, right? For instance, making sure they're not empty strings?

pkg/server/commands/batch_check_command.go
willvedd: My expectation is that the underlying check resolver handles `context.DeadlineExceeded` and `context.Canceled` errors. Each affected check would surface its own error. This allows partial results to be fulfilled. Besides validations, the batch check command is just an orchestrator that doesn't conduct any other operations that could result in an error.

willvedd (changes requested):
Good work so far. Two big things from me – unit tests and the max concurrency configuration.

pkg/server/commands/batch_check_command.go
justincoh: Correct, but those should be handled by [this rule in the API layer](https://github.com/openfga/api/pull/203/files#diff-d3a74ea9ecaac0a46611253dc0c4bc9c402e2ad9eb255cd41d07266453e258f0R1278)

pkg/server/commands/batch_check_command.go
justincoh: That makes sense. In that case, it should work as described above with no additional changes.

internal/server/config/config.go
```diff
@@ -258,6 +262,14 @@ type Config struct {
 	// MaxTuplesPerWrite defines the maximum number of tuples per Write endpoint.
 	MaxTuplesPerWrite int
 
+	// MaxChecksPerBatchCheck defines the maximum number of tuples
+	// that can be passed in each BatchCheck request.
+	MaxChecksPerBatchCheck uint32
+
+	// MaxConcurrentChecksPerBatchCheck defines the maximum number of checks
+	// that can be run in simultaneously
+	MaxConcurrentChecksPerBatchCheck uint32
```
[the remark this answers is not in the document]
justincoh: I tend to agree with you about the `maxConcurrentChecks` setting, it feels like a superfluous constraint. I'd be curious to hear what @elbuo8 thinks here. `number of goroutines` isn't a great measure of load, it depends on what those routines are actually doing.

For context: the existing SDK batchCheck typically sees batch sizes in the double digits, [with some in the low hundreds](https://layer0-prod.datadoghq.com/notebook/10271148/sdk-wrapper-method-usage?range=604800000&view=view-mode&start=1729717849082&live=true)

pkg/server/commands/batch_check_command.go
willvedd: Keep in mind that OpenFGA can be interfaced with as a standalone package as opposed to via an API. In these cases, the API-layer validations won't be enforced. This may require some redundant checks and unit tests to ensure proper validation for each respective scenario.

justincoh: ah that is a great point, will add the needed validation

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,162 @@
...
+	checkQuery := NewCheckCommand(
+		bq.datastore,
+		bq.checkResolver,
+		bq.typesys,
+		WithCheckCommandLogger(bq.logger),
+		WithCacheController(bq.cacheController),
+	)
+
+	// the keys to this map are the correlation_id associated with each check
+	var resultMap = map[string]*BatchCheckOutcome{}
```
miparnisari, line 110: Suggestion: `type correlationID string` and then `map[correlationId]` and you can get rid of the comment

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,162 @@
...
+	lock := sync.Mutex{}
+
+	ctx, cancel := context.WithCancel(ctx)
+	pool := concurrency.NewPool(ctx, int(bq.maxConcurrentChecks))
+	defer func() {
+		_ = pool.Wait()
```
miparnisari, line 116: Is the order of events right? You are returning `resultMap` and then waiting for the pool to complete? It should be

```go
pool.wait()
return resultMap
```

no? I'm actually not sure...

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,162 @@
...
+			fmt.Errorf("batchCheck received %d checks, the maximum allowed is %d ", len(params.Checks), bq.maxChecksAllowed),
+		)
+	}
+
+	if err := validateNoDuplicateCorrelationIDs(params.Checks); err != nil {
+		return nil, err
+	}
+
+	// This check query will be run against every check in the batch
+	checkQuery := NewCheckCommand(
```
miparnisari, line 101: I am so happy that I sent https://github.com/openfga/openfga/pull/1857 that enabled this PR to be much simpler than it would have been without it 😆

pkg/server/batch_check.go
```diff
@@ -0,0 +1,130 @@
...
+		s.checkResolver,
+		typesys,
+		commands.WithBatchCheckCommandCacheController(s.cacheController),
+		commands.WithBatchCheckCommandLogger(s.logger),
+		commands.WithBatchCheckMaxChecksPerBatch(s.maxChecksPerBatchCheck),
+		commands.WithBatchCheckMaxConcurrentChecks(s.maxConcurrentChecksPerBatch),
+	)
+
+	result, err := cmd.Execute(ctx, &commands.BatchCheckCommandParams{
+		AuthorizationModelID: req.GetAuthorizationModelId(),
```
miparnisari, line 67:
```suggestion
		AuthorizationModelID: typesys.GetModelID()
```

please add a test for this... i think that if the user calls the API without a model ID we are finding the latest model (in line 51) but then the way you wrote it will pass an empty string to `cmd.Execute` and i'm not sure what that will do

pkg/server/commands/batch_check_command.go
justincoh: You're right. I actually got burned by this `Wait()` earlier today while trying to add metadata to the batch

pkg/server/batch_check.go
```diff
@@ -0,0 +1,130 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	"github.com/openfga/openfga/internal/authz"
+	"github.com/openfga/openfga/pkg/server/commands"
+	"github.com/openfga/openfga/pkg/telemetry"
+)
+
+func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckRequest) (*openfgav1.BatchCheckResponse, error) {
```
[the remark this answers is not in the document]
justincoh: Yeah I separated it mostly for ease of use. Imo every endpoint should probably live in its own file inside `/server`, it's kind of the opposite of cramming them all into one. ListUsers probably got missed because it was the one exception to the pattern, if **every** endpoint was its own file they would probably all get covered in a case like that.

All that said, that rearranging is definitely not part of this PR. I wouldn't mind doing that refactor work as soon as batch check lands tho 🤔

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,162 @@
...
+
+	// the keys to this map are the correlation_id associated with each check
+	var resultMap = map[string]*BatchCheckOutcome{}
+	lock := sync.Mutex{}
+
+	ctx, cancel := context.WithCancel(ctx)
+	pool := concurrency.NewPool(ctx, int(bq.maxConcurrentChecks))
+	defer func() {
+		_ = pool.Wait()
+		cancel()
```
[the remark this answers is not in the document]
justincoh: This sounds right to me. The cancellable context is likely a leftover from my initial setup, it has no utility here. Will remove

pkg/server/batch_check.go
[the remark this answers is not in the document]
miparnisari: Sounds good. Let's discuss as a team what we prefer and then we can make changes, if any.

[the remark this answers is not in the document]
willvedd: > Can we consider putting this on server.go like the rest? (cc @willvedd WDYT? i think we should eventually move ListUsers to server.go as well)

I'm not convinced that placing all code in the same file is a legitimate way of combatting bugs 🫠 . That single file is so long that I'm actually of the opinion that it creates more bugs.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
   Learn more about this feature and how to enable it [here](https://openfga.dev/docs/getting-started/setup-openfga/access-control)
   If you do try it out, please provide feedback in the [GitHub Discussion](https://github.com/orgs/openfga/discussions)
 * Document OpenFGA release process [#1923](https://github.com/openfga/openfga/pull/1923)
+* Add `BatchCheck` api endpoint [#2039](https://github.com/openfga/openfga/pull/2039)
```
willvedd, line 26: **Reminder:** This needs to be brought back to the top before merging.

pkg/server/commands/batch_check_command.go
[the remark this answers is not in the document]
willvedd: In cases where OpenFGA is used as a package, a developer is going to provide their own context to BatchCheck. So we at the very least need to respect any cancelations or deadlines that occur.

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,162 @@
...
+		bq.datastore,
+		bq.checkResolver,
+		bq.typesys,
+		WithCheckCommandLogger(bq.logger),
+		WithCacheController(bq.cacheController),
+	)
+
+	// the keys to this map are the correlation_id associated with each check
+	var resultMap = map[string]*BatchCheckOutcome{}
+	lock := sync.Mutex{}
```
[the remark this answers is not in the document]
justincoh: sync.Map is nicer but it loses type safety, afaict there's no good way to specify types when using a sync map. I *could* use the sync.Map within the concurrency pool and then convert it to a concrete type before I return from `Exec()`, wdyt?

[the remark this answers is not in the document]
justincoh: update: i did it [here](https://github.com/openfga/openfga/pull/2039/commits/7e92dab61cd96a07694c4edfdd62dcc5c5ad7f7c)

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,182 @@
...
+			fmt.Errorf("batchCheck received %d checks, the maximum allowed is %d ", len(params.Checks), bq.maxChecksAllowed),
+		)
+	}
+
+	if err := validateCorrelationIDs(params.Checks); err != nil {
+		return nil, nil, err
+	}
+
+	var resultMap = new(sync.Map)
+	totalQueryCount := atomic.Uint32{}
```
elbuo8, line 106: This doesn't need to be atomic at all because _right before you flush_ you iterate through all the results once, you can count there.

internal/server/config/config.go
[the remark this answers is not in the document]
elbuo8: I believe the current implementation is correct and what had been previously agreed on.

pkg/server/batch_check.go
```diff
@@ -0,0 +1,130 @@
...
+			singleResult.CheckResult = &openfgav1.BatchCheckSingleResult_Error{
+				Error: transformCheckCommandErrorToBatchCheckError(v.Err),
+			}
+		} else {
+			singleResult.CheckResult = &openfgav1.BatchCheckSingleResult_Allowed{
+				Allowed: v.CheckResponse.Allowed,
+			}
+		}
+
+		batchResult[string(k)] = singleResult
```
elbuo8, line 98: How is this not covered by tests?

justincoh: There are no tests in this branch anywhere to try and make review easier. There's a separate [PR here which is all tests](https://github.com/openfga/openfga/pull/2047/files). The plan is to eventually merge that testing PR into this branch, at which point these codecov warnings should all disappear

pkg/server/commands/batch_check_command.go
justincoh: good call

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,182 @@
...
+	StoreID              string
+}
+
+type BatchCheckOutcome struct {
+	CheckResponse *graph.ResolveCheckResponse
+	Err           error
+}
+
+type BatchCheckMetadata struct {
+	TotalQueries uint32
```
willvedd, line 45:
```suggestion
	DatastoreQueryCount uint32
```
?

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,182 @@
...
+
+	var resultMap = new(sync.Map)
+
+	pool := concurrency.NewPool(ctx, int(bq.maxConcurrentChecks))
+	for _, check := range params.Checks {
+		pool.Go(func(ctx context.Context) error {
+			select {
+			case <-ctx.Done():
+				resultMap.Store(check.GetCorrelationId(), &BatchCheckOutcome{
+					Err: context.Canceled,
```
willvedd, line 112:
```suggestion
					Err: ctx.Err(),
```

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,182 @@
...
+
+		totalQueryCount += outcome.CheckResponse.GetResolutionMetadata().DatastoreQueryCount
+		return true
+	})
+
+	return results, &BatchCheckMetadata{TotalQueries: totalQueryCount}, nil
+}
+
+func validateCorrelationIDs(checks []*openfgav1.BatchCheckItem) error {
+	seen := map[string]bool{}
```
willvedd, line 163: Slight memory optimization:
```suggestion
	seen := map[string]struct{}{}
```

pkg/server/batch_check.go
justincoh: Update: #2065 (unit tests) and #2066 (integration tests) have been merged into this branch.

justincoh:
example trace since some people have asked. This was a batch of 4 where one check was invalid:
image

BatchCheck specific info:
image

pkg/server/batch_check.go
```diff
@@ -0,0 +1,148 @@
...
+		attribute.KeyValue{Key: "store_id", Value: attribute.StringValue(req.GetStoreId())},
+		attribute.KeyValue{Key: "batch_size", Value: attribute.IntValue(len(req.GetChecks()))},
+		attribute.KeyValue{Key: "consistency", Value: attribute.StringValue(req.GetConsistency().String())},
+	))
+	defer span.End()
+
+	if !validator.RequestIsValidatedFromContext(ctx) {
+		if err := req.Validate(); err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
```
willvedd, line 39: This is a block that we need coverage on. But otherwise, code coverage starting to look pretty good 👍

pkg/server/batch_check.go
```diff
@@ -0,0 +1,148 @@
...
+		AuthorizationModelID: typesys.GetAuthorizationModelID(),
+		Checks:               req.GetChecks(),
+		Consistency:          req.GetConsistency(),
+		StoreID:              storeID,
+	})
+
+	if err != nil {
+		telemetry.TraceError(span, err)
+		var batchValidationError *commands.BatchCheckValidationError
+		if errors.As(err, &batchValidationError) {
```
willvedd, line 79: Should also get this line covered.

justincoh: [covered just now](https://app.codecov.io/github/openfga/openfga/commit/d2fdf108d400336fd18d7f3a792cc7caa9c55981/indirect-changes#36a0baed52e93860bd7168f6656dfaaf-L36)

pkg/server/batch_check.go
justincoh: [covered just now](https://app.codecov.io/github/openfga/openfga/commit/d2fdf108d400336fd18d7f3a792cc7caa9c55981/indirect-changes#36a0baed52e93860bd7168f6656dfaaf-L36)

tests/batchcheck/batchcheck_test.go
```diff
@@ -0,0 +1,47 @@
...
+	testRunAll(t, "postgres")
+}
+
+func TestCheckMySQL(t *testing.T) {
+	testRunAll(t, "mysql")
+}
+
+func TestCheckSQLite(t *testing.T) {
+	testRunAll(t, "sqlite")
+}
```
miparnisari, line 29: The reason why we originally wrote these tests for Check, ListObjects, ListUsers is because the APIs are complex and there are thousands of combinations of inputs that we wanted to account for.

- That's not the case for BatchCheck, which doesn't have complex business logic.
- If Check breaks while developing something, the test will also break for BatchCheck too, which is annoying/undesirable.
- Consider how much time this adds to the test suite:

```
 ok  	github.com/openfga/openfga/tests/batchcheck	66.842s
```

Given the above, I was wondering whether what we need are unit tests and not integration tests.

pkg/server/batch_check_test.go
```diff
@@ -0,0 +1,343 @@
...
+			},
+		},
+		"def456": {
+			Err: graph.ErrResolutionDepthExceeded,
+		},
+	}
+
+	b.Run("benchmark_transform_check_result_to_proto", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			transformCheckResultToProto(outcomes)
```
miparnisari, line 340: IMO, `transformCheckResultToProto` is not complex enough to warrant a benchmark, and it's not in the critical path.

A more useful benchmark could be testing what happens when the batch check contains `MaxChecksPerBatchCheck` to ensure that we don't introduce regressions when it comes to parallelizing the work.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+* Add `BatchCheck` api endpoint [#2039](https://github.com/openfga/openfga/pull/2039)
```
willvedd, line 10: **Request:** Can you write a brief description of batch check and how to use it? Ex:

```
* `BatchCheck` API which allows multiple checks to be performed simultaneously. It can be used by caling the `/BatchCheck` endpoint and requires each check to assign a correlation ID. See [batch check docs](#) for more details. [#2039](https://github.com/openfga/openfga/pull/2039)
```

tests/batchcheck/batchcheck_test.go
justincoh: I agree. These integration tests are mostly just testing that the check responses come back grouped and formatted properly, which is also covered in the unit tests. At the very least, we don't need them for all 4 datastores, but I actually don't think we need them at all; we're getting the coverage elsewhere

tests/check/check.go
```diff
@@ -29,13 +29,13 @@ import (
 
 var writeMaxChunkSize = 40 // chunk write requests into a chunks of this max size
 
-type individualTest struct {
+type IndividualTest struct {
```
miparnisari, line 32: Nit: can we undo this change? Do we still need it?

pkg/server/batch_check_test.go
willvedd: AFAICT, it _is_ in the critical path which is why I asked for this benchmark test to begin with.

miparnisari: Ah my bad, "critical path" isn't the right term...

justincoh: I ended up leaving this benchmark in and adding an explainer comment to the test. It actually is part of the critical path, transforming all final check results into the BatchCheck proto format after they've been gathered.

pkg/server/commands/batch_check_command.go
```diff
@@ -0,0 +1,189 @@
...
+		results[CorrelationID(k.(string))] = outcome
+
+		totalQueryCount += outcome.CheckResponse.GetResolutionMetadata().DatastoreQueryCount
+		return true
+	})
+
+	return results, &BatchCheckMetadata{DatastoreQueryCount: totalQueryCount}, nil
+}
+
+func validateCorrelationIDs(checks []*openfgav1.BatchCheckItem) error {
```
willvedd, line 168: Something I'm realizing is missing is validating against an empty batch. You have that in the API validations, but that's not enforced by the server AFAICT.

justincoh: you are correct, just fixed and added an assertion in the command tests

miparnisari (approved):

internal/authz/authz.go
```diff
@@ -51,6 +52,7 @@ const (
 	CanCallWrite                    = "can_call_write"
 	CanCallListObjects              = "can_call_list_objects"
 	CanCallCheck                    = "can_call_check"
+	CanCallBatchCheck               = "can_call_batch_check"
```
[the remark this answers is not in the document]
justincoh: I'm not sure tbh. We need the constant `authz.BatchCheck` defined for telemetry purposes, but we can just have it direct to the check permission for the actual auth decision.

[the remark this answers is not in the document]
rhamzeh: I agree with having it call `can_call_check` 👍🏽

Similar to how `ReadAuthorizationModel` and `ReadAuthorizationModels` call the same `can_call_read_authorization_models`

willvedd (approved):
Great work 👍

PR #2044, opened 2024-10-25
refactor: consolidate check_command error handling

justincoh:
## Description

The CheckCommand was built with the assumption that we would be running only single checks in a request. As a result, the `check_command` itself is translating internal errors into external facing GRPC error types. For BatchCheck, we need error handling on a per-check basis (in addition to per-request).

This PR refactors the error handling in the `check_command` to make it completely agnostic to the transport layer. It just returns specific error types which the server (or BatchCheck) can transform as they need.

No logic was changed as part of this PR, it's a pure refactor.

## References

Closely related to #2031 

## Review Checklist

### DIFF
--- a/pkg/server/commands/check_command.go
+++ b/pkg/server/commands/check_command.go
@@ -28,6 +28,30 @@ const (
 	defaultMaxConcurrentReadsForCheck = math.MaxUint32
 )
 
+type InvalidTupleError struct {
+	Message string
+}
+
+func (e *InvalidTupleError) Error() string {
+	return e.Message
+}
+
+type InvalidRelationError struct {
+	Message string
+}
+
+func (e *InvalidRelationError) Error() string {
+	return e.Message
+}
+
+type ThrottledError struct {
+	Message string
+}
+
+func (e *ThrottledError) Error() string {
+	return e.Message
+}
+
 type CheckQuery struct {
 	logger          logger.Logger
 	checkResolver   graph.CheckResolver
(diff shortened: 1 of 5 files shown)


pkg/server/commands/check_command.go
```diff
@@ -28,6 +26,30 @@ const (
...
+}
+
+type ThrottledError struct {
+	Message string
+}
+
+func (e *ThrottledError) Error() string {
+	return e.Message
+}
+
```
willvedd, line 52: **Question:** I wish I had an answer for this but how does this compare to `/pkg/server/errors`? It seems like there is at least some overlap that can be consolidated. 

Errors have been a sore spot in the past so not asking for a complete overhaul but can you see if there can be some consolidation or reorganization?

willvedd (commented):
I'm in favor of the spirit of this PR but have one question that I'm wondering.

pkg/server/commands/check_command.go
justincoh: I was also unsatisfied with this setup, so I changed it significantly in the most recent commit. Probably easier to talk over zoom but here were my thoughts:

The server errors are specifically design to be [external facing](https://github.com/openfga/openfga/blob/main/pkg/server/errors/errors.go#L1) grpc errors. They build errors based on status codes and messages and then convert them to a pure error with a message like "rpc error: code = Code(2000) desc = oh no", and there's no way to backport that without matching on the string itself.

Many of our packages have their own separate error definitions, conditions, typesystem, storage. `Check` feels like a first-class citizen in our code, so it'd make sense to have its own specific error definitions, but our `package check` [is a test package](https://github.com/openfga/openfga/blob/main/tests/check/check.go#L1) for some reason.

[This commit](https://github.com/openfga/openfga/pull/2044/commits/df0101676da6ff10799b7a4c3f93101d3b46974b) creates an `errors.go` within the /commands package which absorbs all of the new error types and translation to GRPC that was scattered around in the PR from last week. I'm not 100% wed to this setup but it does feel better than stuffing more cruft into the server file.

willvedd (approved):
We talked through this a couple times. Considered some alternatives to creating bespoke types but this ended up being the safest option to implement.

PR #2063, opened 2024-10-31
perf: enable optimization for nested userset with public wildcard

adriantam:
## Description
Enabling performance optimization for nested userset that has public wildcard.  This requires moving the handling of public wildcard to separate checker.

Special note: This has slight degrade in benchmark performance for `benchmark_with_bypass_userset_read` as it now requires two calls to ReadUsersetTuples (1 for public wildcard and 1 for userset) instead of a single ReadUsersetTuples call (1 call combining public wildcard and userset).

## Review Checklist

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -1259,17 +1259,16 @@ func (c *LocalChecker) buildNestedMapper(ctx context.Context, req *ResolveCheckR
 }
 
 func (c *LocalChecker) checkPublicAssignable(ctx context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
-	ctx, span := tracer.Start(ctx, "checkPublicAssignable")
-	defer span.End()
-
 	typesys, _ := typesystem.TypesystemFromContext(ctx)
-
 	ds, _ := storage.RelationshipTupleReaderFromContext(ctx)
 	storeID := req.GetStoreID()
 	reqTupleKey := req.GetTupleKey()
 	userType := tuple.GetType(reqTupleKey.GetUser())
 	wildcardRelationReference := typesystem.WildcardRelationReference(userType)
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
+		ctx, span := tracer.Start(ctx, "checkPublicAssignable")
+		defer span.End()
+
 		response := &ResolveCheckResponse{
 			Allowed: false,
 		}
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -1271,8 +1271,7 @@ func (c *LocalChecker) checkPublicAssignable(ctx context.Context, req *ResolveCh
 	wildcardRelationReference := typesystem.WildcardRelationReference(userType)
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		response := &ResolveCheckResponse{
-			Allowed:            false,
-			ResolutionMetadata: &ResolveCheckResponseMetadata{},
+			Allowed: false,
 		}
 
 		opts := storage.ReadUsersetTuplesOptions{
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -1323,6 +1323,25 @@ func (c *LocalChecker) checkDirectUserTuple(ctx context.Context, req *ResolveChe
 	}
 }
 
+// helper function to return whether checkDirectUserTuple should run.
+func shouldCheckDirectTuple(ctx context.Context, reqTupleKey *openfgav1.TupleKey) bool {
+	typesys, _ := typesystem.TypesystemFromContext(ctx)
+
+	objectType := tuple.GetType(reqTupleKey.GetObject())
+	relation := reqTupleKey.GetRelation()
+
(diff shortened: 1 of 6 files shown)


elbuo8 (dismissed):

internal/graph/check.go
```diff
@@ -675,7 +675,7 @@ ConsumerLoop:
 	return finalResult, nil
 }
 
-// checkUsersetSlowPath will check userset or public wildcard path.
+// checkUsersetSlowPath will check userset path.
```
miparnisari, line 678: I don't understand why this comment was updated ? If I look at the code of `produceUsersetDispatches`, we still have the wildcard short-circuiting.

internal/graph/check.go
```diff
@@ -1258,28 +1258,81 @@ func (c *LocalChecker) buildNestedMapper(ctx context.Context, req *ResolveCheckR
...
+		}
+
+		filteredIter := storage.NewConditionsFilteredTupleKeyIterator(
+			storage.NewFilteredTupleKeyIterator(
+				storage.NewTupleKeyIteratorFromTupleIterator(iter),
+				validation.FilterInvalidTuples(typesys),
+			),
+			checkutil.BuildTupleKeyConditionFilter(ctx, req.GetContext(), typesys),
+		)
+		defer filteredIter.Stop()
```
miparnisari, line 1298: Suggestion: don't use `ReadUsersetTuples`. You are expecting only one result, so use `ReadUserTuple` and specify the full tuple that you are looking for

internal/graph/check.go
```diff
@@ -1258,28 +1258,81 @@ func (c *LocalChecker) buildNestedMapper(ctx context.Context, req *ResolveCheckR
...
+}
+
+func (c *LocalChecker) checkDirectUserTuple(ctx context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
+	typesys, _ := typesystem.TypesystemFromContext(ctx)
+
 	reqTupleKey := req.GetTupleKey()
 
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
-		ctx, span := tracer.Start(ctx, "checkDirectUserTuple",
+		ctx, span := tracer.Start(ctx, "checkDirectUserTupleWrapper",
```
miparnisari, line 1320: Why the name change?

internal/graph/check.go
```diff
@@ -1309,6 +1362,35 @@ func (c *LocalChecker) checkDirectUserTuple(ctx context.Context, req *ResolveChe
...
+	isPubliclyAssignable, _ := typesys.IsPubliclyAssignable(
+		typesystem.DirectRelationReference(objectType, relation), // target
+		tuple.GetType(reqTupleKey.GetUser()),
+	)
+	return isPubliclyAssignable
+}
+
 // checkDirect composes two CheckHandlerFunc which evaluate direct relationships with the provided
 // 'object#relation'. The first handler looks up direct matches on the provided 'object#relation@user',
 // while the second handler looks up relationships between the target 'object#relation' and any usersets
```
miparnisari, line 1396: Nit: this godoc is inaccurate now. Could you please update it?

internal/graph/check.go
```diff
@@ -1334,8 +1416,6 @@ func (c *LocalChecker) checkDirect(parentctx context.Context, req *ResolveCheckR
 		// directlyRelatedUsersetTypes could be "user:*" or "group#member"
```
miparnisari, line 1416: is this line accurate? it seems like no, because `user:*` can never be result

internal/graph/check_test.go
```diff
@@ -4654,3 +4654,277 @@ func TestCheckDirectUserTuple(t *testing.T) {
...
+				{},
+			},
+			readUsersetTuplesError: nil,
+			expected: &ResolveCheckResponse{
+				Allowed: false,
+			},
+			expectedError: nil,
+		},
+		{
+			name: "error",
```
miparnisari, line 4885: please add a test case where the model has conditions to ensure that we are properly evaluating the conditions

internal/graph/check.go
adriantam: The point of this PR is to have checkPublicAssignable instead of checkUsersetSlowPath checking against public wildcard.  In this way, [nestedUsersetFastPath](https://github.com/openfga/openfga/blob/af054d9693bd7ebd0420456b144c2fb6888aaf87/internal/graph/check.go#L1214) will still work when wildcard can be directly assigned to the relation.

internal/graph/check.go
adriantam: Will add comment that we want to use ReadUsersetTuples because we want to take advantage of the cache.

internal/graph/check.go
adriantam: will change it back.

internal/graph/check_test.go
adriantam: updated

miparnisari (approved):

internal/graph/check.go
adriantam: we can do it in separate PR later.

PR #2074, opened 2024-11-06
chore: add GCI linter to group import statements

miparnisari:
Because I want to introduce consistency in the grouping of imports, e.g. this feels slightly messy:

image

versus with this PR:
image

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -65,6 +65,7 @@ linters-settings:
       - standard # Standard section: captures all standard packages.
       - default # Default section: contains all imports that could not be matched to another section type.
       - prefix(github.com/openfga)  # Custom section: groups all imports with the specified Prefix.
+      - localmodule
 
 issues:
   exclude-use-default: false
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -60,13 +60,13 @@ linters-settings:
     alias:
       - pkg: github.com/openfga/api/proto/openfga/v1
         alias: openfgav1
+      - pkg: github.com/openfga/language/pkg/go/transformer
+        alias: parser
   gci:
     sections:
       - standard # Standard section: captures all standard packages.
       - default # Default section: contains all imports that could not be matched to another section type.
-      - prefix(github.com/openfga/api)  # Custom section: groups all imports with the specified Prefix.
-      - prefix(github.com/openfga/openfga)
-      - prefix(github.com/openfga/language)
+      - prefix(github.com/openfga)  # Custom section: groups all imports with the specified Prefix.
 
 issues:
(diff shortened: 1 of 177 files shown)


.golangci.yaml
```diff
@@ -52,13 +53,20 @@ linters-settings:
...
     alias:
       - pkg: github.com/openfga/api/proto/openfga/v1
         alias: openfgav1
+  gci:
+    sections:
+      - standard # Standard section: captures all standard packages.
+      - default # Default section: contains all imports that could not be matched to another section type.
+      - prefix(github.com/openfga/api)  # Custom section: groups all imports with the specified Prefix.
+      - prefix(github.com/openfga/openfga)
+      - prefix(github.com/openfga/language)
```
DanCech, line 69: interesting, shouldn't these 2 be swapped based on order of the actual imports in the files?

miparnisari: Yeah, i noticed after sending this PR that either the tool doesn't work as I expect, or I don't understand how to configure it properly.

miparnisari: Removing `local-prefixes: "github.com/openfga/openfga"` seems to have fixed this.

jpadilla (dismissed):

.golangci.yaml
DanCech: Not sure that we want to remove local-prefixes, I don't know what if anything goimports uses that for

Right now we're mixing local imports with other repos in the openfga org, which is weird imho. The standard approach would be that those other repos are considered part of the external group.

I'd suggest either going back to the standard 3-group layout; `standard, default, localmodule` or a 4-group layout; `standard, default, prefix(github.com/openfga), localmodule`

miparnisari: goimports' local-prefixes seems to affect the order of the imports: https://golangci-lint.run/usage/linters/#goimports

> Reformat imports in autofix mode.

image

> Right now we're mixing local imports with other repos in the openfga org, which is weird imho.

do you prefer to have

`openfga/api`, then
`openfga/language`, then
`openfga/openfga`?

DanCech: this config seems to do that:
```
  goimports:
    local-prefixes: "github.com/openfga/openfga"
  gci:
    sections:
      - standard # Standard section: captures all standard packages.
      - default # Default section: contains all imports that could not be matched to another section type.
      - prefix(github.com/openfga)  # Custom section: groups all imports with the specified Prefix.
      - localmodule  # Local section: contains all imports that are part of the same module.
```

miparnisari: Added `localmodule`.

DanCech (approved):
Looks great!

jpadilla (approved):

Issue #1995, opened 2024-10-05
`ResolveCheckRequest.clone()` is not thread safe


miparnisari:
## The problem

[`ResolveCheckRequest` struct](https://github.com/openfga/openfga/blob/2499898e616afa1c0094df1eeba33acdeca7a74c/internal/graph/resolve_check_request.go#L9) contains a field `TupleKey` that is a pointer, and a function [`clone()`](https://github.com/openfga/openfga/blob/main/internal/graph/resolve_check_request.go#L20-L42) wherein this field is not being cloned, and instead the pointer is copied. This can create race issues if two goroutines attempt to mutate a clone of a request.

## The task

1. Make `ResolveCheckRequest.clone()` thread safe. Some options:
    - creating a new TupleKey through `tuple.NewTupleKey()`, or 
    - call `proto.Clone(r.TupleKey)`, or
    - by having TupleKey NOT be a pointer
3. Update the existing unit tests: https://github.com/openfga/openfga/blob/main/internal/graph/resolve_check_request_test.go
4. Write a new unit test for `clone()` where we create a `ResolveCheckRequest` struct, and then two different goroutines call `clone()` on it.

ovindu-a:
@miparnisari I think I can work on this if I get some more info on the issue.

miparnisari:
@ovindu-a updated the issue with more details, let me know if it helps!

ovindu-a:
Much more clearer to me now @miparnisari. I will work on this.

siiddhantt:
hi @miparnisari , I opened a PR for the issue. Was working on it already

PR #2075, opened 2024-11-07
feat: fast path support on set operations within TTUs

elbuo8:
## Description

Introduce a new fast path strategy that allows set/algebraic operations while resolving TTUs. This is a different strategy than our existing TTU fast path implementation has the original one is more efficient as it's less memory heavy but it's cached results aren't as efficient. We will determine if it will be completely replaced by this new implementation at a later time.

The determination of usage of this fast path strategy relies on the weighted graph. So including it on the typesystem package will increase its cost until we normalize its usage.

There are additional fixes/improvements found while developing this strategy that were not caught before. Namely we are cleaning up "drained" iterators rather than keeping them in memory within the `combinedIterator`, this frees up resources as soon as possible. There was a data race in the iterator cache that was also addressed.

## Review Checklist


internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+		var err error
+		iterProducers, err = pollIteratorProducers(ctx, iterProducers)
+		if err != nil {
+			concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+			return
+		}
+		allIters := true
+		minKey := ""
+		indexes := make([]int, 0)
+		for idx, producer := range iterProducers {
```
adriantam, line 148: Line 148-175 looks very similar to the intersection case.  Any chance we want to combine them?

elbuo8: Similar but subtly different. Couldn't find a way to combine them.

pkg/typesystem/typesystem.go
```diff
@@ -560,6 +567,67 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
+	}
+
+	edges, ok := t.authzWeightedGraph.GetEdgesByNode(node)
+	if !ok {
+		return false
+	}
+
+	ttuEdges := make([]*graph.WeightedAuthorizationModelEdge, 0)
+
+	// find all TTU edges  with valid weight
```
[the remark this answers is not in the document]
elbuo8: I'm going to defer to @yissellokta to give a more detailed explanation.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+			if msg.iter != nil {
+				msg.iter.Stop()
+			}
+		}
+	}
+}
+
+func pollIteratorProducers(ctx context.Context, iterProducers []*iteratorProducer) ([]*iteratorProducer, error) {
+	for _, producer := range iterProducers {
+		if producer.iter != nil || (producer.done && producer.iter == nil) {
```
adriantam, line 51: Can you explain why we need producer.iter == nil when producer.done?  Wouldn't if producer.done == true, line 60 will always be true?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+		}
+		for msg := range iterProducer.producer {
+			if msg.iter != nil {
+				msg.iter.Stop()
+			}
+		}
+	}
+}
+
+func pollIteratorProducers(ctx context.Context, iterProducers []*iteratorProducer) ([]*iteratorProducer, error) {
```
adriantam, line 49: can you make note in the comment that this pollIteratorProducers will have the subtle functionality of trimming the producer list if the producer is done and drained?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+				batch = make([]*openfgav1.TupleKey, 0)
+			}
+			continue
+		}
+
+		// move all entries to less than the MAX to be >= than MAX
+		for _, producer := range iterProducers {
+			t, err := producer.iter.Head(ctx)
+			if err != nil {
+				if storage.IterIsDoneOrCancelled(err) {
```
adriantam, line 304: Probably worth calling in comment that it is not possible for the iter to be done because  we have checked against Head earlier.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+				}
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+				return
+			}
+			tmpKey := t.GetObject()
+			for tmpKey < maxKey {
+				t, err := producer.iter.Next(ctx)
+				if err != nil {
+					if storage.IterIsDoneOrCancelled(err) {
+						producer.iter.Stop()
```
adriantam, line 317: Do we have a bug here?  Assume you have the following
A=[5]
B=[6 7]
C=[6 7]

max key = 6.  When we process A, we would have call next and the iterator will be done.  We should have return at this point.  Otherwise, if we continue, the pollIteratorProducers would have trimmed slice A and we think there is intersection for 7.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+			}
+			batch = append(batch, t)
+			if len(batch) > IteratorBatchThreshold {
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{iter: storage.NewStaticTupleKeyIterator(batch)}, iterChan)
+				batch = make([]*openfgav1.TupleKey, 0)
+			}
+			continue
+		}
+
+		// base > diff, then move the diff to catch up with base
```
adriantam, line 422:
```suggestion
		// diff < base, then move the diff to catch up with base
```
using the same term for comment and code makes it easier to read

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+				return
+			}
+			diff = t.GetObject()
+		}
+	}
+
+	// drain the base
+	if len(iterProducers) == 1 && iterProducers[0].idx == 0 {
+		for len(iterProducers) == 1 && iterProducers[0].iter != nil {
```
adriantam, line 441: why do we still need to check for len(iterProducers) == 1

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+					return
+				}
+				batch = append(batch, t)
+			}
+			var err error
+			iterProducers, err = pollIteratorProducers(ctx, iterProducers)
+			if err != nil {
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+				return
+			}
```
adriantam, line 460: Can you add comment why we need to do this?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+				if err != nil {
+					if storage.IterIsDoneOrCancelled(err) {
+						iterProducers[0].iter.Stop()
+						iterProducers[0].iter = nil
+						break
+					}
+					concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+					return
+				}
+				batch = append(batch, t)
```
adriantam, line 453: Do we need to check for `if len(batch) > IteratorBatchThreshold` here?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+		return c.fastPathOperationSetup(ctx, req, intersectionSetOperator, rw.Intersection.GetChild()...)
+	case *openfgav1.Userset_Difference:
+		return c.fastPathOperationSetup(ctx, req, exclusionSetOperator, rw.Difference.GetBase(), rw.Difference.GetSubtract())
+	default:
+		return nil, ErrUnknownSetOperator
+	}
+}
+
+func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChan <-chan *iteratorMsg, iter TupleMapper) (*ResolveCheckResponse, error) {
+	rightOpen := true
```
adriantam, line 512: can you add some comment to explain why is left vs. right?  Hard to picture in my head what does that really mean.

pkg/typesystem/typesystem.go
```diff
@@ -560,6 +567,67 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
+	if !ok {
+		return false
+	}
+
+	w, ok := node.GetWeight(userType)
+	if !ok {
+		return false
+	}
+
+	if w == level {
```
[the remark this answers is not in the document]
yissellokta: why less? you want to apply a particular path based on the weight, not on the weight equal or less, weight 3 should not have the same algorithm that weight 2, or weight 2 should not have the same algorithm that weight 1

pkg/typesystem/typesystem.go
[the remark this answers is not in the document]
yissellokta: yes, it is the maximum weight, so you can have a node with weight 3 but that does not mean that the TTU edge you are evaluating is weight 3, that TTU edge can have weight 2 but there is another TTU edge or userset edge, etc that can have weight 3. When the weighted graph gets passed and used instead of the type system, all this evaluation will be simpler because you will be passing directly the TTU edge while traversing the graph, instead to what we have right now

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+
+type iteratorMsg struct {
+	iter storage.TupleKeyIterator
+	err  error
+}
+
+type iteratorProducer struct {
+	idx         int
+	iter        storage.TupleKeyIterator
+	initialized bool
```
elbuo8, line 30: Remove.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/internal/checkutil"
+	"github.com/openfga/openfga/internal/concurrency"
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/tuple"
+	"github.com/openfga/openfga/pkg/typesystem"
+)
+
+const IteratorBatchThreshold = 100
```
elbuo8, line 18: Increase to 1k.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,668 @@
...
+					continue
+				}
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+				return
+			}
+			if idx == 0 {
+				batch = append(batch, t)
+			}
+		}
+		if len(batch) > IteratorBatchThreshold {
```
adriantam, line 197: Please add comment to note that it is ok (and better) to exceed the batch size

internal/graph/check_fast_path.go
adriantam: Discussed offline - there is no bug because we will exit due to short circuit on len(iterProducers) != childrenTotal

internal/graph/check_fast_path.go
adriantam: never mind, we call pollIteratorProducers which can trim iterProducers

internal/graph/check_fast_path.go
elbuo8: Renamed the variable to `minBatch` to signal that implicitly.

internal/graph/check_fast_path.go
elbuo8: What do we do here?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,679 @@
...
+
+	r, ok := <-rightChan
+	if !ok {
+		return res, nil
+	}
+	if r.err != nil {
+		return nil, r.err
+	}
+	rightSet.Add(r.userset)
+	for leftOpen || rightOpen {
```
willvedd, line 560: Could this be consolidated to a single value?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,679 @@
...
+		indexes := make([]int, 0)
+		for idx, producer := range iterProducers {
+			v, err := producer.iter.Head(ctx)
+			if err != nil {
+				if storage.IterIsDoneOrCancelled(err) {
+					producer.iter.Stop()
+					producer.iter = nil
+					allIters = false
+					// we need to ensure we have all iterators at all times
+					break
```
willvedd, line 157: **Suggestion:** Could you just continue the outer loop here instead of tracking the `allIters` var?

willvedd (commented):
Still need to look through everything again but some initial questions and suggestions.

internal/concurrency/concurrency.go
```diff
@@ -18,9 +18,11 @@ func NewPool(ctx context.Context, maxGoroutines int) *pool.ContextPool {
 
 // TrySendThroughChannel attempts to send an object through a channel.
 // If the context is canceled, it will not send the object.
-func TrySendThroughChannel[T any](ctx context.Context, msg T, channel chan<- T) {
+func TrySendThroughChannel[T any](ctx context.Context, msg T, channel chan<- T) bool {
```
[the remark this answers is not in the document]
elbuo8: `defer` is executed at the function which yields the iterator. So if I were to put it on the polling function that would create no `noop` iterators. 

If I were to put it on each loop of the client calling the polling function that would create a stack of defer functions to be called once its called which is a waste of resources.

And this indicator is for when we are actually flushing. If the responsibility to close the iterator falls in the consumer, but there is no consumer to receive anymore, then it should be the producer who does it.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,679 @@
...
+	return c.fastPathRewrite(ctx, childRequest, rel.GetRewrite())
+}
+
+func fastPathUnion(ctx context.Context, iterProducers []*iteratorProducer, iterChan chan *iteratorMsg) {
+	batch := make([]*openfgav1.TupleKey, 0)
+
+	defer func() {
+		// flush
+		if len(batch) > 0 {
+			concurrency.TrySendThroughChannel(ctx, &iteratorMsg{iter: storage.NewStaticTupleKeyIterator(batch)}, iterChan)
```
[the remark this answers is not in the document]
elbuo8: Notice I don't flush at the end of each function, this does it. There are several exit points, thus this basically is a catch all handler for all of them (and preventing issues in the future).

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,679 @@
...
+			}
+		}
+	}
+}
+
+// pollIteratorProducers will return a list of the remaining active producers.
+// To be considered active your producer channel must still be open.
+func pollIteratorProducers(ctx context.Context, iterProducers []*iteratorProducer) ([]*iteratorProducer, error) {
+	for _, producer := range iterProducers {
+		if producer.iter != nil || producer.done {
```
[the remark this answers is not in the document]
elbuo8: >  Is there a case where a nil iter isn't done?

At the very first loop, initialization.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,679 @@
...
+type iteratorMsg struct {
+	iter storage.TupleKeyIterator
+	err  error
+}
+
+type iteratorProducer struct {
+	idx      int
+	iter     storage.TupleKeyIterator
+	done     bool
+	producer chan *iteratorMsg
```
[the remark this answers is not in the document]
elbuo8: Provide a name suggestion please.

internal/graph/check_fast_path.go
elbuo8: Elaborate. Keep in mind they are used different contexts.

internal/graph/check_fast_path.go
elbuo8: Potentially? But I'm def not the biggest fan of that pattern. If there is consensus from ya'll I could change it.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 ### Added
 - New flag `OPENFGA_CHECK_ITERATOR_TTL`. Please see the flag description (`./openfga run --help`) for more details. [#2082](https://github.com/openfga/openfga/pull/2082)
 - New flag `OPENFGA_CHECK_CACHE_LIMIT`. Please see the flag description (`./openfga run --help`) for more details. [#2082](https://github.com/openfga/openfga/pull/2082)
+- Improve `Check` performance for TTU relationships that include set operations. Enable via experimental flag `enable-check-optimizations`. [#2075](https://github.com/openfga/openfga/pull/2075)
```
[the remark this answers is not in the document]
elbuo8: There is no performance section. Every performance change has landed in `Added` section.

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,679 @@
...
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case i, ok := <-producer.producer:
+			if !ok {
+				producer.done = true
+				break
+			}
+			if i.err != nil {
+				// NOTE: how do we want to handle missing iterators? full error out?
```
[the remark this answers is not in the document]
elbuo8: Removed.

pkg/typesystem/typesystem.go
```diff
@@ -561,6 +568,70 @@ func (t *TypeSystem) IsDirectlyRelated(target *openfgav1.RelationReference, sour
...
+	}
+
+	w, ok := node.GetWeight(userType)
+	if !ok {
+		return false
+	}
+
+	if w == level && len(node.GetWildcards()) == 0 {
+		return true
+	}
```
miparnisari, line 590: This block puzzles me. If we don't enter it, it could mean that the weight of the node is >2, in that case, we can still return true? But i thought we _only_ wanted nodes of weight 2? 

Also, in my mind, I would have written


```
len(node.GetWildcards()) > 0 {
		return false
	}
```

regardless of the weight?

We can jump on a Zoom if it's too complex to explain here

pkg/typesystem/typesystem.go
```diff
@@ -174,6 +175,7 @@ type TypeSystem struct {
 	modelID                 string
 	schemaVersion           string
 	authorizationModelGraph *graph.AuthorizationModelGraph
+	authzWeightedGraph      *graph.WeightedAuthorizationModelGraph
```
miparnisari, line 178: Noting this here just for others, feel free to ignore. With the current implementation of `WeightedAuthorizationModelGraph` you can't remove the `authModelGraph` and just keep weighted one, because the methods in authModelGraph are not exposed through the weighted one.

pkg/storage/iterators.go
```diff
@@ -61,7 +60,7 @@ func (c *combinedIterator[T]) Next(ctx context.Context) (T, error) {
 	if err != nil {
 		if errors.Is(err, ErrIteratorDone) {
 			c.pending = c.pending[1:]
-			c.done = append(c.done, iter)
+			iter.Stop() // clean up before dropping the reference
```
miparnisari, line 63: Could this be moved to a different PR? :)

miparnisari (commented):
I've not finished the review, just leaving early feedback. I started reviewing bottom to top :)

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+			Relation: computedRelation,
+			User:     r.GetTupleKey().GetUser(),
+		}
+		rel, err := typesys.GetRelation(parentType.GetType(), computedRelation)
+		if err != nil {
+			// NOTE: is there a better way to check and filter rather than skipping?
+			// other paths can be reachable
+			continue
+		}
+		leftChan, err := c.fastPathRewrite(cancellableCtx, r, rel.GetRewrite())
```
miparnisari, line 632: I don't understand why we need a `cancellableCtx` here. You are cancelling after everything has run. But by that time, this function has already run synchronously. So the cancellation will never occur

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+	tk := req.GetTupleKey()
+	computedRelation := rewrite.GetComputedUserset().GetRelation()
+	rewrittenTupleKey := tuple.NewTupleKey(
+		tk.GetObject(),
+		computedRelation,
+		tk.GetUser(),
+	)
+
+	childRequest := req.clone()
+	childRequest.TupleKey = rewrittenTupleKey
```
miparnisari, line 107: Nit: we have fixed (https://github.com/openfga/openfga/issues/1995) the data race that appeared on cloning a `req` and then trying to modify the tuple. You can just do:

```
childRequest := req.clone()
childRequest.TupleKey.Relation = computedRelation
```

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case i, ok := <-producer.producer:
+			if !ok {
+				producer.done = true
+				break
+			}
+			if i.err != nil {
+				return nil, i.err
+			}
```
miparnisari, line 67: could you add a godoc here explaining why we return immediately? what will happen to the other possibly active iterators? they will not be returned

internal/graph/check_fast_path.go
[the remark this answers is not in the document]
miparnisari: i'm confused by the `producer.done` check. I can't see any other place where this is written other than further than below in this function, so how can it ever be true?

internal/graph/check_fast_path.go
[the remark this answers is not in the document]
miparnisari: I had the same reaction when I read `producer.producer`. Maybe `iteratorFanOut`?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+		if !leftOpen {
+			return
+		}
+		go func() {
+			for msg := range leftChan {
+				if msg.iter != nil {
+					msg.iter.Stop()
+				}
+			}
+		}()
```
miparnisari, line 537: 1. Nitpick: you don't need the `if !leftOpen return` because the ranging over a channel is a no-op if the channel is closed. https://go.dev/play/p/A9w8ThwGg54
2. Why drain the channel over a goroutine? This doesn't look like an expensive operation, what am i missing?

internal/graph/check_fast_path.go
miparnisari: probably can replace with a counter of open channels and this would be `for countopen > 0`, but i like it the way it is 🤷‍♀️

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+				}
+			}
+			return nil
+		})
+	}
+
+	go func() {
+		_ = pool.Wait()
+		close(out)
+	}()
```
miparnisari, line 671: I'm confused why you wrapped this within a gorotuine. In other places in the code, we put the `pool.Wait`s inside `defer`, why is this different?

Plus, nobody is waiting on this goroutine?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	return c.resolveFastPath(ctx, leftChans, wrapIterator(TTUKind, iter))
+}
+
+// NOTE: Can we make this generic and move it to concurrency pkg?
+func fanInIteratorChannels(ctx context.Context, chans []chan *iteratorMsg) chan *iteratorMsg {
+	limit := len(chans)
+	pool := concurrency.NewPool(ctx, limit)
+	out := make(chan *iteratorMsg, limit)
```
miparnisari, line 653: You put a limit of `len(chan)`, however, there is a double `for` loop down below, which means that potentially we could be trying to send _more_ than `len(chan)` messages into it. Should be make the channel size bigger to accomodate for this?

internal/graph/check_fast_path.go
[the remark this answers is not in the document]
miparnisari: @elbuo8 you're reading `producer.done` before writing to it on line 62. How will the `producer.done` check ever be true?

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+				if storage.IterIsDoneOrCancelled(err) {
+					iterProducers[iterIdx].iter.Stop()
+					iterProducers[iterIdx].iter = nil
+					continue
+				}
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+				return
+			}
+			if idx == 0 {
+				batch = append(batch, t)
```
miparnisari, line 192: Question: why does this happen here and not in line 163? Then you don't need this `if` anymore

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+				indexes = []int{idx}
+			}
+		}
+
+		if !allIters {
+			// we need to ensure we have all iterators at all times
+			continue
+		}
+
+		// all entries with the same value move forward, but only have to send the value once
```
miparnisari, line 179:
```suggestion
		// all iterators with the same value move forward
```

also I suggest renaming 
- `indexes` to `indexesOfITeratorsWithTheSameObject`
- `minKey` to `minObject`

internal/graph/check_fast_path.go
```diff
@@ -0,0 +1,674 @@
...
+					producer.iter.Stop()
+					producer.iter = nil
+					allIters = false
+					// we need to ensure we have all iterators at all times
+					break
+				}
+				concurrency.TrySendThroughChannel(ctx, &iteratorMsg{err: err}, iterChan)
+				return
+			}
+			if idx == 0 {
```
miparnisari, line 371: Nit / suggestion:

```go
const base = 0
const diff = 1
```

and then use these indexes throughout the function. It could make it more readable.

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	go.uber.org/zap v1.27.0
 	golang.org/x/exp v0.0.0-20240904232852-e7e105dedf7e
 	golang.org/x/sync v0.8.0
+	gonum.org/v1/gonum v0.15.1
```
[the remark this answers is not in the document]
elbuo8: I don't disagree but wasn't directly the changes in this PR. Seems in the resolution of language it came in. Additionally, we are actually using it for traversals, so perhaps when that code was brought in, `go mod tidy` wasn't executed so it got brought in in this PR.

internal/graph/check_fast_path.go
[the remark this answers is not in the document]
elbuo8: Because `iter` will become `nil` when consumers mark it so.

internal/graph/check_fast_path.go
elbuo8: 1. Not needed if I weren't launching a goroutine. But I want to avoid launching it at all costs.
2. May not look as an expensive operation if you think everything is on memory. But the reality is that you don't know if the producer is one that is yielding over the network and has yet to properly cancel. Also think of the case where you had 100s iterators queued, you have to close them all. In light of not knowing how many there are and prevent increasing the latency to reply to the customer, we simply launch the goroutine.

internal/graph/check_fast_path.go
elbuo8: Exactly for the reason of your edit.

internal/graph/check_fast_path.go
elbuo8: > Should be make the channel size bigger to accomodate for this?

Nop, having them block for a while will allow us to have backpressure control. If the consumer is overwhelmed it should signal the producer to backoff. Trying to prevent the N*M problem we have in the normal path for Check.

internal/graph/check_fast_path.go
elbuo8: > Plus, nobody is waiting on this goroutine?

The listener of the close event is.

> pool.Waits inside defer, why is this different?

The defer here wouldn't make any sense as it would `block`. So it would have to be somewhere else. But I don't see the point of leaking that if we have the `close` signal to subscribe to.

pkg/storage/cache.go
```diff
@@ -134,5 +134,5 @@ func GetReadStartingWithUserCacheKeyPrefix(store, objectType, relation string) s
 }
 
 func GetReadCacheKey(store, tuple string) string {
-	return fmt.Sprintf("%sr%s/%s", iteratorCachePrefix, store, tuple)
+	return fmt.Sprintf("%sr/%s/%s", iteratorCachePrefix, store, tuple)
```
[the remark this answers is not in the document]
elbuo8: Cosmetic ✨

pkg/storage/iterators.go
elbuo8: Sigh, yes.

pkg/typesystem/typesystem_test.go
```diff
@@ -4361,6 +4361,590 @@ func TestTTUCanUseFastTrack(t *testing.T) {
...
+							relations
+								define parent: [group]
+								define member: [user, group#admin]
+								define admin: [user]
+						type folder
+							relations
+								define parent: [group]
+								define viewer: member from parent
+					`,
+			objectType:        "folder",
```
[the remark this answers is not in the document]
elbuo8: ?

pkg/typesystem/typesystem_test.go
```diff
@@ -4361,6 +4361,590 @@ func TestTTUCanUseFastTrack(t *testing.T) {
...
+						type group
+							relations
+								define member: [user]
+						type folder
+							relations
+								define parent: [group]
+								define viewer: member from parent
+					`,
+			objectType:        "group",
+			relation:          "viewer",
```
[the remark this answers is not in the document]
elbuo8: `bad_tupleset_relation`, hence doesn't exist. Can rename the test but it's literally a copy paste from the older function.

pkg/typesystem/typesystem_test.go
```diff
@@ -4361,6 +4361,590 @@ func TestTTUCanUseFastTrack(t *testing.T) {
...
```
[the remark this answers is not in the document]
elbuo8: `bad_computed_relation`

Ditto above.

pkg/typesystem/typesystem_test.go
```diff
@@ -4361,6 +4361,590 @@ func TestTTUCanUseFastTrack(t *testing.T) {
...
+							define member: [user, user with x_greater_than]
+						type license
+						  relations
+							define holder_member: member from owner
+							define owner: [company, group, group with x_condition]
+						condition x_greater_than(x: int) {x > 1}
+						condition x_condition(x: int) {x > 1}
+					`,
+			objectType:        "license",
+			relation:          "viewer",
```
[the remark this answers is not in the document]
elbuo8: Ditto.

pkg/typesystem/typesystem.go
```diff
@@ -227,6 +229,10 @@ func New(model *openfgav1.AuthorizationModel) (*TypeSystem, error) {
 		}
 	}
 
+	wgb := graph.NewWeightedAuthorizationModelGraphBuilder()
+	// TODO: this will require a deprecation not ignore the error and remove nil checks
+	weightedGraph, _ := wgb.Build(model)
```
miparnisari, line 234: should we store the error (not return it) so that if there was an error we do _not_ use fast path v2?

miparnisari: ah never mind, i guess this is why you added

```go
if t.authzWeightedGraph == nil {
		return false
	}
```

down below.

internal/graph/check_fast_path.go
miparnisari: Ah makes sense, could we add a note in the code for this? 

```
go func() {
		_ = pool.Wait()
		// NOTE: the consumer of this channel will block waiting for it to close
		close(out)
	}()
```

pkg/typesystem/typesystem.go
miparnisari: Track via https://github.com/openfga/openfga/issues/2119

go.mod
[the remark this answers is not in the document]
miparnisari: Track via https://github.com/openfga/openfga/issues/2119

pkg/typesystem/typesystem_test.go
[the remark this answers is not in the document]
miparnisari: `relation` field is missing in the testcase

pkg/typesystem/typesystem_test.go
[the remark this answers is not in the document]
miparnisari: the test case `multiple_parent_types_with_conditions_multiple_child_types_with_conditions_or_wildcard` doesn't suggest an invalid input 🤔

pkg/typesystem/typesystem_test.go
```diff
@@ -4361,6 +4361,590 @@ func TestTTUCanUseFastTrack(t *testing.T) {
...
+			// notice that group_without_member does not have member.  However, we should
+			// still allow because group_with_member has member
+			objectType:        "folder",
+			relation:          "viewer",
+			tuplesetRelation:  "parent",
+			computedRelation:  "member",
+			expectCanFastPath: true,
+		},
+		{
+			name: "ttu_child_not_directly_assignable_union",
```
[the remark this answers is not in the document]
miparnisari: also the `relation` field is missing from the test case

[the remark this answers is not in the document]
miparnisari: ah, you removed this test.

pkg/typesystem/typesystem.go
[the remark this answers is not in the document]
miparnisari: I think it's not possible to have a TTU relation with weight less than 2. E.g. for

```
model
    schema 1.1
type user
type folder1
    relations
        define viewer:[user]
type doc
    relations
        define parent: [folder1]
```

`weight(doc#viewer)[user]` is 2.

pkg/typesystem/typesystem.go
miparnisari: it took me a while to digest this bit. @elbuo8 can we leave a comment in the code? something like:

> `node.weight(userType)` is only the maximum weight to get to "userType". if the weight is exactly 2, we exit early. if the weight is >2, it is possible that the node has multiple paths to arrive to "userType", but the path that we are exploring right now actually has weight <= 2, in which case we want to return `true`.

miparnisari (approved):

jpadilla (approved):

PR #2082, opened 2024-11-08
chore: modification of flags for Check

miparnisari:
## Description
- Address these TODOs: https://github.com/openfga/openfga/blob/0e865fd67cceb4454b9ce57476af897f078c45d8/cmd/run/flags.go#L212, https://github.com/openfga/openfga/blob/0e865fd67cceb4454b9ce57476af897f078c45d8/pkg/server/server.go#L754

## Reference

Corresponding docs update https://github.com/openfga/openfga.dev/pull/886


.config-schema.json
```diff
@@ -490,14 +501,8 @@
...
                 },
-                "limit": {
-                    "description": "if caching of Check and ListObjects calls is enabled, this is the size limit (in items) of the cache",
-                    "type": "integer",
-                    "default": "10000",
-                    "x-env-variable": "OPENFGA_CHECK_QUERY_CACHE_LIMIT"
-                },
                 "ttl": {
-                    "description": "if caching of Check and ListObjects is enabled, this is the TTL of each value",
+                    "description": "if check-query-cache-enabled or check-iterator-cache-enabled, this is the TTL of each value",
```
elbuo8, line 505: This should only apply to itself.
`checkIteratorCache` should have its own TTL.

cmd/run/run.go
```diff
@@ -226,11 +226,13 @@ func NewRunCommand() *cobra.Command {
...
 
-	flags.Uint32("check-query-cache-limit", defaultConfig.Cache.Limit, "if caching of Check and ListObjects calls is enabled, this is the size limit of the cache")
+	flags.Bool("check-query-cache-enabled", defaultConfig.CheckQueryCache.Enabled, "enable caching of Check requests. For example, if you have a relation define viewer: owner or editor, and the query is Check(user:anne, viewer, doc:1), we'll evaluate the owner relation and the editor relation and cache both results: (user:anne, viewer, doc:1) -> allowed=true and (user:anne, owner, doc:1) -> allowed=true. The cache is stored in-memory; the cached values are overwritten on every change in the result, and cleared after the configured TTL. This flag improves latency, but turns Check and ListObjects into eventually consistent APIs.")
 
-	flags.Duration("check-query-cache-ttl", defaultConfig.CheckQueryCache.TTL, "if caching of Check and ListObjects is enabled, this is the TTL of each value")
+	flags.Uint32("check-cache-limit", defaultConfig.CheckCache.Limit, "if check-query-cache-enabled or check-iterator-cache-enabled, this is the size limit of the cache")
```
elbuo8, line 233: Move this to the top so its not confused.

elbuo8 (dismissed):

elbuo8 (dismissed):

cmd/run/run.go
```diff
@@ -222,15 +222,17 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Uint32("listUsers-max-results", defaultConfig.ListUsersMaxResults, "the maximum results to return in ListUsers API responses. If 0, all results can be returned")
 
+	flags.Uint32("check-cache-limit", defaultConfig.CheckCache.Limit, "if check-query-cache-enabled or check-iterator-cache-enabled, this is the size limit of the cache")
+
 	flags.Bool("check-iterator-cache-enabled", defaultConfig.CheckIteratorCache.Enabled, "enable caching of datastore iterators of Check requests.")
```
aaguiarz, line 227: Are we explaining somewhere in the docs how "caching of datastore iterators" works, what's the benefit, and when to enable them? 

Why don't we have as single flag for all caches? How does the user decide which one to enable?

miparnisari: >Are we explaining somewhere in the docs how "caching of datastore iterators" works, what's the benefit, and when to enable them?

Yes, please review https://github.com/openfga/openfga.dev/pull/886.

>Why don't we have as single flag for all caches? How does the user decide which one to enable?

I don't have a good answer. While writing docs I realized that this is too cumbersome to explain, which probably means it's also cumbersome to use. We _should_ have just one caching flag.

.config-schema.json
```diff
@@ -464,38 +464,50 @@
...
+                    "x-env-variable": "OPENFGA_CHECK_CACHE_LIMIT"
+                }
+            }
+        },
         "checkIteratorCache": {
             "type": "object",
             "properties": {
                 "enabled": {
-                    "description": "enable caching of datastore iterators of Check requests",
+                    "description": "enable caching of datastore iterators. Each iterator is the result of a database query, for example usersets related to a specific object, or objects related to a specific user, up to a certain number of tuples per iterator. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.",
```
miparnisari, line 482: @jpadilla could you tell me what to write here in terms of cache invalidation? I cannot tell from looking at the code: https://github.com/openfga/openfga/blob/f1f1b54b2163b4918a9509dbb5894df39293bd27/internal/graph/storagewrapper.go#L410-L412

.config-schema.json
```diff
@@ -464,38 +464,50 @@
                 }
             }
         },
+        "checkCache": {
+            "type": "object",
+            "properties": {
+                "limit": {
+                    "description": "the size limit (in items) of the cache for Check (queries and iterators)",
+                    "type": "integer",
+                    "default": "10000",
```
miparnisari, line 473: @elbuo8 the default here is 10k, but the default for `checkIteratorCache.maxResults` is also 10k - this seems to imply that one DB iterator will use up the entire cache?

jpadilla: A cache entry value for an iterator is a slice. `checkIteratorCache.maxResults` limits the size/length of those slices. If a datastore query produces >= `checkIteratorCache.maxResults` we won't cache it. This `limit` is the maximum number of entries in the in memory cache instance.

.config-schema.json
```diff
@@ -464,38 +464,50 @@
...
+                    "description": "enable caching of Check requests. For example, if you have a relation `define viewer: owner or editor`, and the query is Check(user:anne, viewer, doc:1), we'll evaluate the `owner` relation and the `editor` relation and cache both results: (user:anne, viewer, doc:1) -> allowed=true and (user:anne, owner, doc:1) -> allowed=true. The cache is stored in-memory; the cached values are overwritten on every change in the result, and cleared after the configured TTL. This flag improves latency, but turns Check and ListObjects into eventually consistent APIs. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.",
                     "type": "bool",
                     "default": "false",
                     "x-env-variable": "OPENFGA_CHECK_QUERY_CACHE_ENABLED"
                 },
-                "limit": {
```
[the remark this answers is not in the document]
miparnisari: Done :)

elbuo8 (approved):

PR #2087, opened 2024-11-08
refactor: rename interface methods

miparnisari:
## Description
Discussed with @jpadilla and @cikasfm that the current methods name are misleading because they only apply to the ReadChanges API.


DanCech (approved):

miparnisari:
Closing because the methods are actually _also_ used in the Read API.

PR #2088, opened 2024-11-10
test: move tests for ReadAuthModel

miparnisari:
## Description

Why?

## References

Prior work 


pkg/storage/mysql/mysql.go
```diff
@@ -475,12 +475,6 @@ func (s *Datastore) WriteAuthorizationModel(ctx context.Context, store string, m
 	ctx, span := startTrace(ctx, "WriteAuthorizationModel")
 	defer span.End()
 
-	typeDefinitions := model.GetTypeDefinitions()
-
-	if len(typeDefinitions) > s.MaxTypesPerAuthorizationModel() {
-		return storage.ExceededMaxTypeDefinitionsLimitError(s.maxTypesPerModelField)
-	}
-
```
miparnisari, line 483: FYI deleting because this code exists in the API layer: https://github.com/openfga/openfga/blob/main/pkg/server/commands/write_authzmodel.go#L58-L60

justincoh (approved):

PR #2089, opened 2024-11-10
test: move tests for Assertions

miparnisari:
## Description

Why?

## References

Prior work 


pkg/server/commands/write_assertions_test.go
```diff
@@ -161,4 +165,350 @@ func TestWriteAssertions(t *testing.T) {
...
+					},
+				},
+			},
+			setMock: func(mockDatastore *mockstorage.MockOpenFGADatastore) {
+				mockDatastore.EXPECT().ReadAuthorizationModel(gomock.Any(), storeID, modelID).Times(1).Return(model, nil)
+				mockDatastore.EXPECT().WriteAssertions(gomock.Any(), storeID, modelID, gomock.Any()).Times(1).Return(nil)
+			},
+		},
+		{
+			_name: "succeeds_when_empty_assertions",
```
justincoh, line 226: Any reason for the `_` preceding `_name`?

justincoh (approved):

pkg/server/commands/write_assertions_test.go
miparnisari: No reason, updated :)

justincoh (approved):

PR #2091, opened 2024-11-11
test: move tests for Write

miparnisari:
## Description

Why?

## References

Prior work 


pkg/storage/mysql/mysql.go
```diff
@@ -219,10 +219,6 @@ func (s *Datastore) Write(
 	ctx, span := startTrace(ctx, "Write")
 	defer span.End()
 
-	if len(deletes)+len(writes) > s.MaxTuplesPerWrite() {
-		return storage.ErrExceededWriteBatchLimit
-	}
-
```
miparnisari, line 225: FYI this is already in the command: https://github.com/openfga/openfga/blob/move-tests-part-four/pkg/server/commands/write.go#L177-L179

pkg/server/commands/write.go
```diff
@@ -125,6 +133,7 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgav1.
 	}
 
 	for _, tk := range deletes {
+		// TODO validate relation format and object format
```
miparnisari, line 136: We're actually going to the database in these cases 🤦‍♀️

pkg/server/commands/write.go
```diff
@@ -91,7 +99,7 @@ func (c *WriteCommand) validateWriteRequest(ctx context.Context, req *openfgav1.
 			if errors.Is(err, storage.ErrNotFound) {
 				return serverErrors.AuthorizationModelNotFound(modelID)
 			}
-			return err
+			return serverErrors.HandleError("", err)
```
justincoh, line 102: I greatly dislike having `command` files return grpc errors but that's a separate PR

pkg/server/errors/errors.go
```diff
@@ -121,10 +119,6 @@ func InvalidAuthorizationModelInput(err error) error {
 // Use `public` if you want to return a useful error message to the user.
 func HandleError(public string, err error) error {
 	switch {
-	case errors.Is(err, storage.ErrTransactionalWriteFailed):
-		return status.Error(codes.Aborted, err.Error())
-	case errors.Is(err, storage.ErrInvalidWriteInput):
-		return WriteFailedDueToInvalidInput(err)
```
justincoh, line 127: no longer relevant?

justincoh (approved):

pkg/server/commands/write.go
miparnisari: Yeah this will have to change everywhere if we go that route of having commands return sentinel errors.

pkg/server/errors/errors.go
miparnisari: No, since the only API that can return these errors is Write API, and I moved this code to the Write command.

justincoh (approved):

PR #2092, opened 2024-11-11
test: move tests for Stores (part 2)

miparnisari:
## Description

Why?

## References

Prior work 


DanCech (approved):

DanCech (approved):

pkg/storage/test/stores.go
```diff
@@ -84,6 +106,11 @@ func StoreTest(t *testing.T, datastore storage.OpenFGADatastore) {
 		require.ErrorIs(t, err, storage.ErrNotFound)
 	})
 
+	t.Run("delete_store_if_not_found_succeeds", func(t *testing.T) {
+		err := datastore.DeleteStore(ctx, "unknown")
+		require.NoError(t, err)
```
justincoh, line 111: Is this the expected behavior? I see the comment about delete stores was also removed, but I just want to confirm that deleting a non-existent store should be `ok`

justincoh (approved):

PR #2101, opened 2024-11-13
Replace `ccache` with `theine-go` for in-memory caching

### LINKED ISSUE #2097: Replace existing caching library
### Checklist

Our current in-memory caching library needs replacement.

We want to replace the existing caching library, https://github.com/openfga/openfga/blob/17bd507f8ccc1e3ae84fd1eba5fc4406b1a87b51/go.mod#L24

with https://github.com/Yiling-J/theine-go.

The code that needs to be updated is in https://github.com/openfga/openfga/blob/17bd507f8ccc1e3ae84fd1eba5fc4406b1a87b51/pkg/storage/cache.go#L39-L95

_No response_

### Additional context

_No response_

Siddhant-K-code:
## Description

This PR replaces the existing ccache library with theine-go for in-memory caching functionality.

Changes:
- Switched from `github.com/karlseguin/ccache/v3` to `github.com/Yiling-J/theine-go`
- Updated InMemoryLRUCache implementation to use theine-go's Cache
- Maintained existing behavior for negative TTL handling
- Preserved all existing cache interfaces

The migration provides a more efficient caching implementation while keeping the same API surface and behavior that existing code depends on.

## References

fixes #2097

## Review Checklist


Siddhant-K-code:
A test seems to be failing 🫠 

will check it tomorrow morning

 🛌🏻

pkg/storage/cache.go
```diff
@@ -62,35 +61,43 @@ func NewInMemoryLRUCache[T any](opts ...InMemoryLRUCacheOpt[T]) *InMemoryLRUCach
...
+	cache, _ := theine.NewBuilder[string, T](t.maxElements).Build()
+	t.cache = cache
 	return t
 }
 
 func (i InMemoryLRUCache[T]) Get(key string) T {
 	var zero T
-	item := i.ccache.Get(key)
-	if item == nil {
+	if i.cache == nil {
```
elbuo8, line 71: Why are we checking for the existence of the cache in every method?

pkg/storage/cache.go
```diff
@@ -62,35 +61,43 @@ func NewInMemoryLRUCache[T any](opts ...InMemoryLRUCacheOpt[T]) *InMemoryLRUCach
...
+		return zero
 	}
-
-	return zero
+	return value
 }
 
 func (i InMemoryLRUCache[T]) Set(key string, value T, ttl time.Duration) {
-	i.ccache.Set(key, value, ttl)
+	if ttl <= 0 {
```
elbuo8, line 83: In what case do we pass in negative TTL?

Siddhant-K-code: Negative TTL handling is tested in cache_test.go in the "set_and_get_expired" test case. When a negative TTL is passed, we treat it as a delete operation - this maintains consistent behavior with the previous implementation and provides a clean way to handle invalid TTL values.


https://github.com/Siddhant-K-code/openfga/blob/8c030dc5fff1610c86c47a1de2991559b36c94e0/pkg/storage/cache_test.go#L25-L34

pkg/storage/cache.go
Siddhant-K-code: The cache existence checks are needed because after Stop() is called, the cache becomes nil. The tests verify this behavior in [TestCachedCheckResolver_ResolveCheck_After_Stop_DoesNotPanic](https://github.com/Siddhant-K-code/openfga/blob/8c030dc5fff1610c86c47a1de2991559b36c94e0/internal/graph/cached_resolver_test.go#L646-L668), ensuring graceful handling when cache operations are attempted after stopping.

elbuo8: The cache doesn't become `nil` externally. If anything you'd have to do `i.cache = nil` inside `Stop` but that would then require us to add an additional mutex.

That test you reference seems wrong since we wouldn't call `WithExistingCache` with `nil`. If that function accepts nil, it should actually panic.

There shouldn't be any `nil` checks for the methods. That test that we need to ensure behavior is, that the cache is never called after calling Stop. Which should only happen on cleanup/wrap up of the service.

miparnisari:
FYI there is a similar PR: https://github.com/openfga/openfga/pull/2098

Siddhant-K-code:
> FYI there is a similar PR: #2098

Ohh I just saw that, they haven't ref. to issue so, I missed that PR! Closing this in favour of #2098

PR #2102, opened 2024-11-13
feat: handle duplicate checks in BatchCheck

justincoh:
## Description

This PR ensures that BatchCheck does not try and resolve the same check multiple times in a request. We're unable to rely on our existing check cache for batch since we fan out the checks into multiple routines immediately, meaning we _might_ concurrently attempt to resolve the same check multiple times before the cache is filled. The changes in this PR sidestep this problem by deduplicating the checks up front:

1. Upon receipt of a batch, iterate over its checks and generate a unique cache key for each distinct check.
	* If we have 5 checks in a batch that are identical, they will receive only 1 cache key
2. Resolve each unique check (with uniqueness determined by cache key)
3. After resolution, map each check result to all of its associated `correlation_ids` and return results

## Review Checklist


internal/graph/cached_resolver_test.go
```diff
@@ -666,265 +664,3 @@ func TestCachedCheckResolver_ResolveCheck_After_Stop_DoesNotPanic(t *testing.T)
 	_, err := cachedCheckResolver.ResolveCheck(context.Background(), &ResolveCheckRequest{})
 	require.NoError(t, err)
 }
-
-func TestCheckCacheKeyDoNotOverlap(t *testing.T) {
```
justincoh, line 670: these were relocated to `storage/cache_test.go` lower in this PR

pkg/server/commands/batch_check_command.go
```diff
@@ -144,30 +158,38 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
...
 	_ = pool.Wait()
 
 	results := map[CorrelationID]*BatchCheckOutcome{}
-	var totalQueryCount uint32
 
-	resultMap.Range(func(k, v interface{}) bool {
-		// Convert types since sync.Map is `any`
-		outcome := v.(*BatchCheckOutcome)
-		results[CorrelationID(k.(string))] = outcome
+	duplicateCount := 0
```
willvedd, line 176: Can't this be derived from number of input checks minus number of cache keys?

justincoh: that is a much more clear way to do it, instead of the math currently down in 179. Will change

pkg/storage/test/cache.go
```diff
@@ -0,0 +1 @@
+package test
```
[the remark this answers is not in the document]
justincoh: ah this file should have been removed entirely, not sure how i missed it

pkg/server/batch_check.go
```diff
@@ -88,6 +89,14 @@ func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckReques
 	).Observe(queryCount)
 	span.SetAttributes(attribute.Int("duplicate_checks", metadata.DuplicateCheckCount))
 
+	if metadata.DuplicateCheckCount != 0 {
+		s.logger.InfoWithContext(ctx, "batch check received duplicate checks",
+			zap.String("store_id", storeID),
+			zap.Int("batch size", len(req.GetChecks())),
+			zap.Int("duplicate count", metadata.DuplicateCheckCount),
+		)
+	}
```
willvedd: While this works, its a divergence from how we normally use logs. Lets also consider that this could become noisy for an operator.

pkg/server/commands/batch_check_command.go
```diff
@@ -112,14 +115,25 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
 		return nil, nil, err
 	}
 
+	// Before processing the batch, deduplicate the checks based on their unique cache key
+	// After all routines have finished, we will map each individual check response to all associated CorrelationIDs
+	cacheKeyToCorrelationIDs := make(map[CacheKey][]CorrelationID)
+	cacheKeyToCheck := make(map[CacheKey]*openfgav1.BatchCheckItem)
```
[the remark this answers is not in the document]
justincoh: Yes, I think merging is a better option here. It feels weird to have two maps with identical keys which is probably a sign there should only be 1

pkg/server/commands/batch_check_command_test.go
```diff
@@ -254,6 +263,58 @@ func TestBatchCheckCommand(t *testing.T) {
 		// response should have 1 key per check
 		require.Equal(t, len(response), numChecks)
 	})
+
+	t.Run("uses_command_cache_to_resolve_dupe_checks", func(t *testing.T) {
+		mockCheckResolver := graph.NewMockCheckResolver(mockController)
+		cmd := NewBatchCheckCommand(ds, mockCheckResolver, ts)
+		numChecks := 5
+		totalChecks := numChecks * 2
```
willvedd, line 271: **Suggestion:**
This is confusing IMO. `numChecks` is actually the number of duplicate batches.
```suggestion
		numChecks := 5
		totalChecks := numChecks * 2
```
Does this test need a loop at all? Would probably be easier to understand if you just hardcoded the payload to include a couple duplicates and one non-duplicate.

pkg/server/commands/batch_check_command_test.go
```diff
@@ -254,6 +263,58 @@ func TestBatchCheckCommand(t *testing.T) {
...
+			AuthorizationModelID: ts.GetAuthorizationModelID(),
+			Checks:               checks,
+			StoreID:              ulid.Make().String(),
+		}
+
+		result, meta, err := cmd.Execute(context.Background(), params)
+
+		require.NoError(t, err)
+		require.Len(t, result, totalChecks)
+		require.Equal(t, meta.DuplicateCheckCount, totalChecks-2)
```
willvedd, line 316: **Request:** Can you add assertions for `DuplicateCheckCount` for all other tests?

pkg/server/commands/batch_check_command.go
```diff
@@ -144,30 +169,36 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
 
 			response, _, err := checkQuery.Execute(ctx, checkParams)
 
-			resultMap.Store(check.GetCorrelationId(), &BatchCheckOutcome{
+			resultMap.Store(key, &BatchCheckOutcome{
 				CheckResponse: response,
 				Err:           err,
 			})
 
+			totalQueryCount.Add(response.GetResolutionMetadata().DatastoreQueryCount)
```
miparnisari, line 177: Please check this, but i think that if `err != nil`, `response.GetResolutionMetadata()` will be nil too

pkg/server/commands/batch_check_command.go
```diff
@@ -40,6 +41,7 @@ type BatchCheckOutcome struct {
 
 type BatchCheckMetadata struct {
 	DatastoreQueryCount uint32
+	DuplicateCheckCount int
```
miparnisari, line 44: i'm curious what is the goal behind keeping tabs on this number? what are we going to do with this information?

willvedd (dismissed):
Great work 👍

pkg/storage/cache.go
```diff
@@ -136,3 +143,51 @@ func GetReadStartingWithUserCacheKeyPrefix(store, objectType, relation string) s
...
+	Context              *structpb.Struct
+}
+
+// GetCheckCacheKey converts the elements of a Check into a canonical cache key that can be
+// used for Check resolution cache key lookups in a stable way.
+//
+// For one store and model ID, the same tuple provided with the same contextual tuples and context
+// should produce the same cache key. Contextual tuple order and context parameter order is ignored,
+// only the contents are compared.
+func GetCheckCacheKey(params *CheckCacheKeyParams) (string, error) {
```
miparnisari, line 162: You improved the benchmark!  🥳 

image

miparnisari (approved):

jpadilla (approved):

jpadilla (approved):

PR #2104, opened 2024-11-14
perf: shortcut on unconnected graph

adriantam:
## Description
Performance optimization by short-cutting and return false if target type cannot be reached

## Review Checklist

### DIFF
--- a/pkg/typesystem/typesystem.go
+++ b/pkg/typesystem/typesystem.go
@@ -586,7 +586,7 @@ func (t *TypeSystem) TTUCanFastPath(objectType, tuplesetRelation, computedRelati
 	return true
 }
 
-func (t *TypeSystem) PathExists(user, relation, object string) (bool, error) {
+func (t *TypeSystem) PathExists(user, relation, objectType string) (bool, error) {
 	userType, _, userRelation := tuple.ToUserParts(user)
 	userTypeRelation := userType
 	if userRelation != "" {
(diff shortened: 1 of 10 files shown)


elbuo8 (dismissed):

pkg/typesystem/typesystem.go
```diff
@@ -585,6 +586,34 @@ func (t *TypeSystem) TTUCanFastPath(objectType, tuplesetRelation, computedRelati
...
+func (t *TypeSystem) PathExists(user, relation, object string) (bool, error) {
+	userType, _, userRelation := tuple.ToUserParts(user)
+	userTypeRelation := userType
+	if userRelation != "" {
+		// this is an userset
+		userTypeRelation = tuple.ToObjectRelationString(userType, userRelation)
+	}
+	fromNode, err := t.authorizationModelGraph.GetNodeByLabel(userTypeRelation)
+	if err != nil {
+		return false, err
```
miparnisari, line 598: If the userType doesn't exist in the model, where will the Check request fail, here or in some other method? I'm hoping it won't reach here and it will fail somewhere else

adriantam: Yes, we should never hit this.  However, from a function perspective, we still need to do validation.

miparnisari: agreed. Can you verify whether the request fails earlier if the userType doesn't exist?

miparnisari (approved):

PR #2106, opened 2024-11-17
test: move tests for WriteModel

miparnisari:
## Description

Why?

## References

Prior work 

- https://github.com/openfga/openfga/pull/2076
- https://github.com/openfga/openfga/pull/2092
- https://github.com/openfga/openfga/pull/2088
- https://github.com/openfga/openfga/pull/2089
- https://github.com/openfga/openfga/pull/2091


tests/functional_test.go
```diff
@@ -343,7 +343,6 @@ func TestFunctionalGRPC(t *testing.T) {
 	t.Run("TestReadAuthorizationModels", func(t *testing.T) { GRPCReadAuthorizationModelsTest(t, client) })
 	t.Run("TestWriteAssertions", func(t *testing.T) { GRPCWriteAssertionsTest(t, client) })
 
-	t.Run("TestWriteAuthorizationModel", func(t *testing.T) { GRPCWriteAuthorizationModelTest(t, client) })
```
miparnisari, line 346: duplicate of line 341

pkg/server/commands/write_authzmodel.go
```diff
@@ -74,10 +71,7 @@ func (w *WriteAuthorizationModelCommand) Execute(ctx context.Context, req *openf
 	// Validate the size in bytes of the wire-format encoding of the authorization model.
 	modelSize := proto.Size(model)
 	if modelSize > w.maxAuthorizationModelSizeInBytes {
-		return nil, status.Error(
-			codes.Code(openfgav1.ErrorCode_exceeded_entity_limit),
-			fmt.Sprintf("model exceeds size limit: %d bytes vs %d bytes", modelSize, w.maxAuthorizationModelSizeInBytes),
-		)
+		return nil, serverErrors.ExceededEntityLimit("bytes in an authorization model", w.maxAuthorizationModelSizeInBytes)
```
justincoh, line 74: Should we keep the error message the same, in light of this week's issue?

miparnisari: It's a fair point, however I feel like the Write API error included key information: the tuple that failed. Which customers wanted to parse because potentially they want to retry or log somewhere. People don't write models nearly as often as they write tuples, so this error is unlikely to be encountered.

justincoh: Yeah tbh I don't like being wed to our existing error messages, and I consider this week's "incident" to be more of the client's fault than ours. Users should be relying on error codes, not on message text

justincoh (approved):

PR #2117, opened 2024-11-25
switch ccache to theine cache

jpadilla:
## Description

Switch our in memory cache implementation from `github.com/karlseguin/ccache/v3` to `github.com/Yiling-J/theine-go`.

## References

https://github.com/openfga/openfga/issues/2097

## Review Checklist


pkg/storage/cache.go
```diff
@@ -53,42 +51,43 @@ var _ InMemoryCache[any] = (*InMemoryLRUCache[any])(nil)
...
 
 	for _, opt := range opts {
 		opt(t)
 	}
 
-	t.ccache = ccache.New(ccache.Configure[T]().MaxSize(t.maxElements))
+	var err error
+	t.client, err = theine.NewBuilder[string, T](t.maxElements).Build()
+	if err != nil {
+		panic(err)
```
jpadilla, line 64: open to suggestions on how to improve this bit.

- rename to `MustNewInMemoryLRUCache` to signal that it can panic?
- change func signature to also return an error.
- keep as is.

DanCech: would advise 2 if the callers can handle the error, otherwise 1

elbuo8: Return the error.

jpadilla: [`6601cac` (#2117)](https://github.com/openfga/openfga/pull/2117/commits/6601caca349a5711bdfd67d38de7e6b228996243)

DanCech (approved):
Nice!

pkg/storage/cache_test.go
```diff
@@ -22,17 +21,9 @@ func TestInMemoryCache(t *testing.T) {
...
-		t.Cleanup(func() {
-			goleak.VerifyNone(t)
-		})
-		defer cache.Stop()
-		cache.Set("key", "value", -1*time.Nanosecond)
-		result := cache.Get("key")
-		require.Equal(t, "", result)
-	})
-
 	t.Run("stop_multiple_times", func(t *testing.T) {
```
miparnisari, line 24: There seems to be a new implicit contract for `Stop` wherein the caller is allowed to call Stop multiple times (e.g. from diff goroutines). Should we add a test for this?

```
t.Run("stop_from_goroutines", func(t *testing.T) {
		cache, err := NewInMemoryLRUCache[string]()
		require.NoError(t, err)
		t.Cleanup(func() {
			goleak.VerifyNone(t)
		})

		pool := concurrency.NewPool(context.Background(), 2)
			cache.Stop()
			return nil
		})
			cache.Stop()
			return nil
		})
		err = pool.Wait()
		require.NoError(t, err)
	})
```

pkg/storage/cache.go
```diff
@@ -50,48 +52,60 @@ func WithMaxCacheSize[T any](maxElements int64) InMemoryLRUCacheOpt[T] {
...
+	}
+
+	i.stopOnce.Do(func() {
+		i.client.Close()
 	})
 }
 
+func (i InMemoryLRUCache[T]) IsReady() bool {
+	return !i.stopping.Load()
+}
```
miparnisari, line 107: This is anti-intuitive. When is someone expected to call `isReady` and why is it not embedded within `Set`, `Get`, etc?

pkg/storage/cache.go
```diff
@@ -50,48 +52,60 @@ func WithMaxCacheSize[T any](maxElements int64) InMemoryLRUCacheOpt[T] {
...
+	i.client.Delete(key)
 }
 
 func (i InMemoryLRUCache[T]) Stop() {
-	i.closeOnce.Do(func() {
-		i.ccache.Stop()
+	swapped := i.stopping.CompareAndSwap(false, true)
+	if !swapped {
+		return
+	}
```
miparnisari, line 98: Why do you need this? The tests pass without it

pkg/storage/cache.go
```diff
@@ -30,14 +29,17 @@ type InMemoryCache[T any] interface {
 
 	// Stop cleans resources.
 	Stop()
+
+	IsReady() bool
```
miparnisari, line 33: We should document _when_ the caller is expected to call this method.

internal/graph/storagewrapper.go
miparnisari, line 1: I'm still not convinced that creating a new decorator of OpenFGADatastore was a good idea. Can we revisit the plan of having this incorporated in `caching.go`? Since they look very similar and have the same goal..

jpadilla: > I'm still not convinced that creating a new decorator of OpenFGADatastore was a good idea.

Apart from looking very similar and having the same goal, any other concerns with having this as a separate decorator of OpenFGADatastore?

I don't think it's necessarily worth doing right now as some of these changes around caching are still in flux.

elbuo8 (approved):

internal/graph/storagewrapper.go
miparnisari: Well, what is the benefit of having them separate?

jpadilla: > Well, what is the benefit of having them separate?

It allows us to be tactical with how/when caching applies. We can't cache it all, and there's likely less value in caching response from datastore queries used in other APIs like ListUser or Expand for example.

Maybe once we have cost functions implemented, we can look into expanding further.

> It looks unorganized.
> Another problem is that because the cache for iterators lives on different package, i can't access its internal fields in unit tests that I am writing: [#2126 (files)](https://github.com/openfga/openfga/pull/2126/files#diff-35eb8305e35761ec5b13e99e7e6066d2f677fbd7298405bd23e32c50f0f92fcdR40-R45)

These two things we can solve _without_ having to merge them into a single instance too, right?

Issue #2120, opened 2024-11-26
Support for disabling otel GRPC exporter


kullanici0606:
### Checklist

- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

We want to enable opentelemetry tracing in OpenFGA to be able to pass and see the otel trace ID in logs but when we enable in with `--trace-enabled=true`, it automatically tries to send traces to the default address `0.0.0.0:4317`, however our infrastructure does not have otel collector. So if it is possible to support enabling tracing with a disabled exporter, it would be great.

### Describe the ideal solution

Instead of directly registering an exporter, it can be registered if an endpoint if provided similar to this:
https://github.com/letsencrypt/boulder/blob/ded2e5e610306700194333d7cc92e29418888975/cmd/shell.go#L331-L345

Or it can be explicitly disabled when the given endpoint is empty or something like `none`
From: https://github.com/openfga/openfga/blob/3391a5cf1e5a9df42fd51964219babe14323d05e/pkg/telemetry/tracing.go#L88-L99
To:
```go
tracerOpts := []sdktrace.TracerProviderOption{
		sdktrace.WithSampler(sdktrace.TraceIDRatioBased(tracer.samplingRatio)),
		sdktrace.WithResource(res),
}
if tracer.endpoint == "none" {
		var exp sdktrace.SpanExporter
		exp, err = otlptracegrpc.New(ctx, options...)
		if err != nil {
			panic(fmt.Sprintf("failed to establish a connection with the otlp exporter: %v", err))
		}
		tracerOpts = append(tracerOpts, sdktrace.WithSpanProcessor(sdktrace.NewBatchSpanProcessor(exp)))
}
	
tp := sdktrace.NewTracerProvider(tracerOpts...)
```

PR #2122, opened 2024-11-26
add support for disabling telemetry exporter

kullanici0606:
## Description
This PR adds support for disabling OpenTelemetry GRPC exporter when tracing is enabled.
This is useful for logging trace IDs but not wanting them to export to OTEL collector.

PR to documentation: https://github.com/openfga/openfga.dev/pull/897

## References
[](https://github.com/openfga/openfga/issues/2120)

## Review Checklist


adriantam:
Thanks @kullanici0606 .  Can you update [CHANGELOG.md](https://github.com/openfga/openfga/blob/main/CHANGELOG.md) to reflect the change?

adriantam (commented):
Please update https://github.com/openfga/openfga/blob/f2cc953d20c4e49252b80cfa994534e4b025319e/.config-schema.json#L412 as well

kullanici0606:
Thank you @adriantam for reviewing, I did the updates you suggested, are they good?

adriantam (dismissed):

kullanici0606:
Thank you @adriantam for the approval. Do I need to do anything else to make this PR be merged?

adriantam (dismissed):

jpadilla:
@kullanici0606 are you propagating trace contexts into openfga and would like to have `trace_id` in logs to correlate?

kullanici0606:
> @kullanici0606 are you propagating trace contexts into openfga and would like to have `trace_id` in logs to correlate?

@jpadilla yes, that’s exactly the our use case, we propagate the trace id to openfga and want to correlate it as you described.

.config-schema.json
```diff
@@ -409,7 +409,7 @@
                     "type": "object",
                     "properties": {
                         "endpoint": {
-                            "description": "The grpc endpoint of the trace collector",
+                            "description": "The grpc endpoint of the trace collector. 'none' for disabling the exporter",
```
jpadilla, line 412: @adriantam I wonder if our assumption of a good default for this(and other properties) should be challenged instead. As in, you explicitly enable tracing + explicitly configure `OTLPTraceConfig.Endpoint.

jpadilla: @miparnisari thoughts?

kullanici0606: I actually thought implementing it that way first (i.e enable exporter if endpoint not empty), but that would be a backward incompatible change, so thought that it would be better to add a keyword for disabling it. 

Do you want me to update it or keep the PR as it is?

kullanici0606: So, is the future plan like this?
```go
switch {
case trace-otlp-endpoint == "none":
    // enable tracing, but do not enable exporter
case trace-otlp-endpoint != "":
    // enable tracing and enable exporter with trace-otlp-endpoint endpoint
default:
   // empty endpoint, do not enable tracing
}
```

I think it would be more intuitive to keep `enabled` and only enable exporter like this:
```go
if trace-enabled {
    // enable tracing
    if trace-otlp-endpoint != "" {
       // enable exporter with trace-otlp-endpoint endpoint
    }
}
```

In order to update the `config.schema`, which path should I assume for the future plans?

kullanici0606:
Hello, what is the final decision on this? Could you please re-review?

senojj:
In order to make the exporter configuration extensible, I suggest adding a new flag: `trace-otlp-exporter`, having a default value of "otlp" and a possible value of "noop".

To support the "noop" exporter, a `noopExporter` type can be created that implements the `sdktrace.SpanExporter` interface, and pass it to the `sdktrace.NewBatchSpanProcessor` function.

```go
var expNoop sdktrace.SpanExporter = noopExporter{}

type noopExporter struct{}

func (e noopExporter) ExportSpans(_ context.Context, _ []sdktrace.ReadOnlySpan) error {
	return nil
}

func (e noopExporter) Shutdown(_ context.Context) error {
	return nil
}
```
used like:
```go
sdktrace.WithSpanProcessor(sdktrace.NewBatchSpanProcessor(expNoop))
```

This configuration may support additional exporters in the future, such as: "httpotlp". The list of available exporters should be kept minimal, as an OpenTelemetry Collector should be used to export to various 3rd party solutions.

Vic-Dev:
Thanks for your interest in OpenFGA! We haven't heard back since the last update in January so we are closing this PR due to inactivity.

PR #2124, opened 2024-11-26
perf: cache controller async invalidation

adriantam:
## Review Checklist

### DIFF
--- a/internal/cachecontroller/cache_controller.go
+++ b/internal/cachecontroller/cache_controller.go
@@ -36,6 +36,8 @@ var (
 	})
 )
 
+const futureCacheInvalidationTime = time.Hour
+
 type CacheController interface {
 	DetermineInvalidation(ctx context.Context, storeID string) time.Time
 }
@@ -88,15 +90,20 @@ func (c *InMemoryCacheController) DetermineInvalidation(
 		return entry.LastModified
 	}
 
-	lastModified, err, _ := c.sf.Do(storeID, func() (interface{}, error) {
-		return c.findChangesAndInvalidate(ctx, storeID)
-	})
+	go func() {
+		// if the cache cannot be found, we want to invalidate entries in the background
+		// so that it does not block the answer path.
+		_, _, _ = c.sf.Do(storeID, func() (interface{}, error) {
+			// Note that when a duplicate (i.e., same storeID) comes in, the duplicate caller waits
+			// for the original to complete and receives the same results.
+			_ = c.findChangesAndInvalidate(ctx, storeID)
+			return nil, nil
+		})
+	}()
 
-	if err != nil {
-		return time.Time{}
-	}
-
-	return lastModified.(time.Time)
+	// we want to return a time in the future so that all entries for the
+	// store will be considered as expired.
+	return time.Now().Add(futureCacheInvalidationTime)
 }
 
 func (c *InMemoryCacheController) findChanges(ctx context.Context, storeID string) ([]*openfgav1.TupleChange, string, error) {
@@ -109,7 +116,7 @@ func (c *InMemoryCacheController) findChanges(ctx context.Context, storeID strin
 	return c.ds.ReadChanges(ctx, storeID, storage.ReadChangesFilter{}, opts)
 }
 
-func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, storeID string) (time.Time, error) {
+func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, storeID string) error {
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.Bool("invalidations", true))
 	cacheKey := storage.GetChangelogCacheKey(storeID)
@@ -121,7 +128,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, 
 		telemetry.TraceError(span, err)
 		// do not allow any cache read until next refresh
 		c.invalidateIteratorCache(storeID)
-		return time.Time{}, err
+		return err
 	}
 
 	entry := &storage.ChangelogCacheEntry{
@@ -137,7 +144,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, 
 	if entry.LastModified.Before(lastVerified) {
 		// no new changes, no need to perform invalidations
 		span.SetAttributes(attribute.Bool("invalidations", false))
-		return entry.LastModified, nil
+		return nil
 	}
 
 	// need to consider there might just be 1 change
@@ -161,7 +168,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, 
 		}
 	}
 
-	return entry.LastModified, nil
+	return nil
 }
 
 func (c *InMemoryCacheController) invalidateIteratorCache(storeID string) {
--- a/internal/cachecontroller/cache_controller.go
+++ b/internal/cachecontroller/cache_controller.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/openfga/openfga/internal/build"
 	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/telemetry"
 	"github.com/openfga/openfga/pkg/tuple"
 )
 
@@ -117,7 +118,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, 
 	// re-evaluate at a later time.
 	changes, _, err := c.findChanges(ctx, storeID)
 	if err != nil {
-		span.RecordError(err)
+		telemetry.TraceError(span, err)
 		// do not allow any cache read until next refresh
 		c.invalidateIteratorCache(storeID)
 		return time.Time{}, err
--- a/internal/cachecontroller/cache_controller_test.go
+++ b/internal/cachecontroller/cache_controller_test.go
@@ -48,7 +48,7 @@ func TestInMemoryCacheController_DetermineInvalidation(t *testing.T) {
 
 		gomock.InOrder(
 			cache.EXPECT().Get(storage.GetChangelogCacheKey(storeID)).Return(nil),
-			ds.EXPECT().ReadChanges(gomock.Any(), storeID, gomock.Any(), gomock.Any()).Return([]*openfgav1.TupleChange{
+			ds.EXPECT().ReadChanges(gomock.Any(), storeID, gomock.Any(), gomock.Any()).AnyTimes().Return([]*openfgav1.TupleChange{
 				{
 					Operation: openfgav1.TupleOperation_TUPLE_OPERATION_WRITE,
 					Timestamp: timestamppb.New(changelogTimestamp),
@@ -58,10 +58,10 @@ func TestInMemoryCacheController_DetermineInvalidation(t *testing.T) {
 						User:     "test",
 					}},
 			}, "", nil),
-			cache.EXPECT().Set(storage.GetChangelogCacheKey(storeID), gomock.Any(), gomock.Any()),
+			cache.EXPECT().Set(storage.GetChangelogCacheKey(storeID), gomock.Any(), gomock.Any()).AnyTimes(),
 		)
 		invalidationTime := cacheController.DetermineInvalidation(ctx, storeID)
-		require.Equal(t, changelogTimestamp, invalidationTime)
+		require.Greater(t, invalidationTime, time.Now())
 	})
 }
 
@@ -254,7 +254,7 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
 				iteratorCacheTTL: 10 * time.Second,
 				sf:               &singleflight.Group{},
 			}
-			_, err := cacheController.findChangesAndInvalidate(ctx, test.storeID)
+			err := cacheController.findChangesAndInvalidate(ctx, test.storeID)
 			if test.expectedError != nil {
 				require.ErrorIs(t, err, test.expectedError)
 				return


internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +90,20 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
+			// Note that when a duplicate (i.e., same storeID) comes in, the duplicate caller waits
+			// for the original to complete and receives the same results.
+			_ = c.findChangesAndInvalidate(ctx, storeID)
+			return nil, nil
+		})
+	}()
+
+	// we want to return a time in the future so that all entries for the
+	// store will be considered as expired.
+	return time.Now().Add(futureCacheInvalidationTime)
```
elbuo8, line 106: This should just be `zero`.

internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +90,20 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
-	lastModified, err, _ := c.sf.Do(storeID, func() (interface{}, error) {
-		return c.findChangesAndInvalidate(ctx, storeID)
-	})
-
-	if err != nil {
-		return time.Time{}
-	}
-
-	return lastModified.(time.Time)
+	go func() {
```
elbuo8, line 93: This would create a goroutine every time there is an expiration. Creating a goroutine seems more expensive than checking for an entry in a `sync.Map` which could be cleaned up _after_  `findChangesAndInvalidate` returns.

internal/cachecontroller/cache_controller.go
adriantam: change to zero

internal/cachecontroller/cache_controller.go
adriantam: updated to have go only for the invalidation path

elbuo8 (approved):

internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +88,23 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
 	}
 
-	lastModified, err, _ := c.sf.Do(storeID, func() (interface{}, error) {
-		return c.findChangesAndInvalidate(ctx, storeID)
-	})
-
-	if err != nil {
-		return time.Time{}
+	_, present := c.mutexMap[storeID]
+	if !present {
```
elbuo8, line 92: Why not create the entry with the lock already acquired? So that when you prompt, if it exists you basically exit immediately. AKA, reducing the number of accesses to the map by one since the creator of the lock will also, be the owner of it.

internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +88,23 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
+		c.mutexMap[storeID] = &sync.Mutex{}
 	}
+	hasLock := c.mutexMap[storeID].TryLock()
+	if hasLock {
+		// if the cache cannot be found, we want to invalidate entries in the background
+		// so that it does not block the answer path.
+		go func() {
+			defer c.mutexMap[storeID].Unlock()
```
elbuo8, line 100: Why not prune the entry? Don't want to carry the entries of low activity stores forever.

internal/cachecontroller/cache_controller.go
```diff
@@ -108,7 +117,7 @@ func (c *InMemoryCacheController) findChanges(ctx context.Context, storeID strin
 	return c.ds.ReadChanges(ctx, storeID, storage.ReadChangesFilter{}, opts)
 }
 
-func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, storeID string) (time.Time, error) {
+func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context, storeID string) error {
 	span := trace.SpanFromContext(ctx)
 	span.SetAttributes(attribute.Bool("invalidations", true))
```
elbuo8, line 122: This is no longer going to be visible. It should generate it's own span and be linked to the parent.

internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +88,23 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
+		go func() {
+			defer c.mutexMap[storeID].Unlock()
+			_ = c.findChangesAndInvalidate(ctx, storeID)
+		}()
+	}
+	// if we cannot get lock, there is already invalidation going on.  As such,
+	// we don't want to spin a new go routine to do invalidation.
 
-	return lastModified.(time.Time)
+	return time.Time{}
```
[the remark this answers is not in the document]
adriantam: time.Time{} is complex type and cannot be a "const"

internal/cachecontroller/cache_controller.go
adriantam: found a bigger bug - the mutexMap itself entry is not protected.  Switch to have a single mutex protecting the map[string]struct{} instead

internal/cachecontroller/cache_controller.go
adriantam: updated to prune entry

internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +90,27 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
-
-	if err != nil {
-		return time.Time{}
+	c.currentInvalidationMutext.Lock()
+	_, present := c.currentInvalidationMap[storeID]
+	if !present {
+		c.currentInvalidationMap[storeID] = struct{}{}
 	}
+	c.currentInvalidationMutext.Unlock()
+	if present {
```
adriantam, line 99: if !present

internal/cachecontroller/cache_controller_test.go
```diff
@@ -248,13 +248,14 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
...
-				ds:               datastore,
-				cache:            cache,
-				ttl:              10 * time.Second,
-				iteratorCacheTTL: 10 * time.Second,
-				sf:               &singleflight.Group{},
+				ds:                        datastore,
+				cache:                     cache,
+				ttl:                       10 * time.Second,
+				iteratorCacheTTL:          10 * time.Second,
+				currentInvalidationMutext: sync.Mutex{},
```
elbuo8, line 255:
```suggestion
			mu: sync.Mutex{},
```

internal/cachecontroller/cache_controller_test.go
```diff
@@ -248,13 +248,14 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
...
+				currentInvalidationMap:    make(map[string]struct{}),
```
elbuo8, line 256:
```suggestion
				inflightInvalidations:    make(map[string]struct{}),
```

internal/cachecontroller/cache_controller.go
```diff
@@ -87,15 +90,27 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
+	if !present {
+		c.currentInvalidationMap[storeID] = struct{}{}
 	}
+	c.currentInvalidationMutext.Unlock()
+	if !present {
+		span.SetAttributes(attribute.Bool("checkInvalidation", true))
+		// if the cache cannot be found, we want to invalidate entries in the background
+		// so that it does not block the answer path.
+		go func() {
+			_ = c.findChangesAndInvalidate(ctx, storeID)
```
elbuo8, line 104: Should this be the same context? Its a background thing, and if the Check resolves first, the context will be cancelled, and we actually want to continue processing.

adriantam: updated to use context without timeout.

internal/cachecontroller/cache_controller.go
```diff
@@ -100,15 +103,29 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
+	if !present {
+		c.inflightInvalidations[storeID] = struct{}{}
 	}
+	c.mu.Unlock()
+	if !present {
+		span.SetAttributes(attribute.Bool("checkInvalidation", true))
+		// if the cache cannot be found, we want to invalidate entries in the background
+		// so that it does not block the answer path.
+		go func() {
+			// important to propagate the cancel without cancel to avoid
```
jpadilla, line 117:
```suggestion
			// important to propagate the context without cancel to avoid
```

jpadilla: I had no idea `WithoutCancel` was a thing

jpadilla (dismissed):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 ## [Unreleased]
 ### Added
 - Add metrics `cachecontroller_find_changes_and_invalidate_histogram` on latency for cache controller in finding changes and invalidating.  [#2135](https://github.com/openfga/openfga/pull/2135)
+- Improve `Check` performance when cache controller is enabled by invalidating asynchronously.  [#2124](https://github.com/openfga/openfga/pull/2124)
```
miparnisari, line 12:
```suggestion
- Improve `Check` performance when cache controller is enabled by invalidating _what?_ asynchronously.  [#2124](https://github.com/openfga/openfga/pull/2124)
```

Also, i just noticed that `cacheControllerEnabled` has no config flags, so how will users of openfga as a binary and not library use this?

internal/cachecontroller/cache_controller.go
```diff
@@ -100,15 +103,29 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
-	if err != nil {
-		return time.Time{}
+	c.mu.Lock()
+	_, present := c.inflightInvalidations[storeID]
+	if !present {
+		c.inflightInvalidations[storeID] = struct{}{}
 	}
+	c.mu.Unlock()
+	if !present {
+		span.SetAttributes(attribute.Bool("checkInvalidation", true))
```
miparnisari, line 113:
```suggestion
		span.SetAttributes(attribute.Bool("check_invalidation", true))
```

internal/cachecontroller/cache_controller.go
```diff
@@ -100,15 +103,29 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
 	}
 
-	lastModified, err, _ := c.sf.Do(storeID, func() (interface{}, error) {
-		return c.findChangesAndInvalidate(ctx, storeID)
-	})
-
-	if err != nil {
-		return time.Time{}
+	c.mu.Lock()
+	_, present := c.inflightInvalidations[storeID]
```
miparnisari, line 107: https://pkg.go.dev/sync#Map

> The [sync.]Map type is optimized for two common use cases: 
> 1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or 
> (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. 
> In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex

I feel that we fall under the (1) bucket, therefore we could replace the map+mutex with a `sync.map`, but i'll defer to you.

internal/cachecontroller/cache_controller.go
```diff
@@ -100,15 +103,29 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
 	}
+	c.mu.Unlock()
+	if !present {
+		span.SetAttributes(attribute.Bool("checkInvalidation", true))
+		// if the cache cannot be found, we want to invalidate entries in the background
+		// so that it does not block the answer path.
+		go func() {
+			// important to propagate the cancel without cancel to avoid
+			// cancelling the invalidation when parent request is complete.
+			_ = c.findChangesAndInvalidate(context.WithoutCancel(ctx), storeID)
```
miparnisari, line 119: This is the only usage of `findChangesAndInvalidate` and we are not using the `err` returned, why not update the function's signature?

Alternatively, could you leave a comment explaining why it's okay to ignore the error?

internal/cachecontroller/cache_controller.go
adriantam: Yeah...this is something I learn today - we should look into using it in some of the check / graph code because it will give us ability to link to the original span (which we will lack if we use a brand new context)

internal/cachecontroller/cache_controller.go
adriantam: I don't think it quite fits in this particular bucket because we will keep writing/deleting over the key in its lifetime.  Thus, I have opted for the plain old mutex.

CHANGELOG.md
adriantam: the cache controller enabled will be tracked via https://github.com/openfga/openfga/issues/2148

internal/cachecontroller/cache_controller.go
elbuo8: @adriantam it is achievable through `LoadOrStore` and `LoadAndDelete`. There is a _minor_ gain there. The same key is only written then deleted, never modified. If we needed to modify it would still work with the Map.

adriantam: updated

internal/cachecontroller/cache_controller.go
```diff
@@ -100,15 +100,22 @@ func (c *InMemoryCacheController) DetermineInvalidation(
...
-		return time.Time{}
+	_, present := c.inflightInvalidations.LoadOrStore(storeID, struct{}{})
+	if !present {
+		span.SetAttributes(attribute.Bool("check_invalidation", true))
+		// if the cache cannot be found, we want to invalidate entries in the background
+		// so that it does not block the answer path.
+		go func() {
+			// important to propagate the context without cancel to avoid
+			// cancelling the invalidation when parent request is complete.
+			c.findChangesAndInvalidate(context.WithoutCancel(ctx), storeID)
```
adriantam, line 111: Discussed with @elbuo8 offline, we want to have this span detached from the original trace to avoid increases trace length un-necessarily.

jpadilla (approved):

Issue #2058, opened 2024-10-29
Fix order of compositon of DB wrappers


miparnisari:
# Problem

The current problem we have is that some datastore wrappers are server-wide, while others are request-specific. This makes it cumbersome because they have to be instantiated at different moments, and this can break the right "order" of the composition, which should be:

[combinedwrapper](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/combinedtuplereader.go) -> [modelcachewrapper](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/caching.go) -> [boundedreaderwrapper](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/boundedconcurrency.go) -> [instrumentedwrapper](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/instrumented.go) -> [contextwrapper](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/context.go) -> the DB

(i.e. from least expensive to most expensive)

As an example of a time when we got the order wrong and this impacted performance, see https://github.com/openfga/openfga/issues/1394.

# Task

We should have every query/command object create it's own datastore wrapper _from scratch_. The server should _only_ hold any shared objects that those wrappers need (namely, `modelcachewrapper`)

_Originally posted by @miparnisari in https://github.com/openfga/openfga/pull/2022#discussion_r1807029200_

PR #2126, opened 2024-11-27
fix: composition of database decorators

miparnisari:
## Description

The PR fixes the order of instantiation of each layer of the `datastore` object needed for each request. It's done in a hacky way by modifying method signatures, but I hope I laid the groundwork to future standardization. I wrote a couple of TODOs of what we could do to make it better. The end goal would be to do this in every request:

```go
func NewCheckCommand(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, params *CheckCommandParams, opts ...CheckQueryOption) *CheckQuery {
	cmd := &CheckQuery{}

	for _, opt := range opts {
		opt(cmd)
	}

	cmd.datastore = NewRequestStorageWrapperForCheckAPI(datastore, params.ContextualTuples.GetTupleKeys(), opts.maxConcurrentReads, opts.cacheConfig)
	
	return cmd
}

func (c *CheckQuery) Execute(ctx context.Context) (*graph.ResolveCheckResponse, *graph.ResolveCheckRequestMetadata, error) {
}
```

(note how Execute doesn't accepts `params` anymore)

---

I expect performance gains for anyone using OpenFGA with `OPENFGA_CHECK_ITERATOR_CACHE_ENABLED=true` and for anyone using contextual tuples.

---

One thing I noticed while doing this:

1. right now we are very inconsistent with how we handle the building of the `datastore` object (some do it inside the `New`, others inside `Execute`, some leave it up to the caller). 
2. this leads to subtle bugs. For example, in ReverseExpand code, we initialize the contextual tuple combined reader [here](https://github.com/openfga/openfga/blob/main/pkg/server/commands/reverseexpand/reverse_expand.go#L486) but we also do it in the caller, [here](https://github.com/openfga/openfga/blob/8d794a97d56cc6778b6db2721a87c57703482444/pkg/server/commands/list_objects.go#L271-L275). 

## References
Fixes https://github.com/openfga/openfga/issues/2058


internal/graph/storagewrapper.go
```diff
@@ -62,7 +61,7 @@ var (
 type iterFunc func(ctx context.Context) (storage.TupleIterator, error)
 
 type CachedDatastore struct {
-	storage.OpenFGADatastore
```
miparnisari, line 65: So I can compose them

internal/graph/storagewrapper.go
```diff
@@ -75,17 +74,18 @@ type CachedDatastore struct {
...
 	maxSize int,
 	ttl time.Duration,
+	sf *singleflight.Group,
 ) *CachedDatastore {
 	return &CachedDatastore{
-		OpenFGADatastore: inner,
-		cache:            cache,
-		maxResultSize:    maxSize,
-		ttl:              ttl,
-		sf:               &singleflight.Group{},
```
miparnisari, line 88: the server will hold a reference to the `sfGroup` so it can be shared across requests, but we add the ability of calling `NewCachedDatastore` per request.

internal/graph/resolve_check_request.go
```diff
@@ -13,7 +13,7 @@ import (
 
 type ResolveCheckRequest struct {
 	StoreID                   string
-	AuthorizationModelID      string
+	AuthorizationModelID      string // TODO replace with typesystem
```
miparnisari, line 16: so that we don't have to read it from the `ctx`

internal/graph/storagewrapper.go
elbuo8: Uh? It was already shared across requests as this is not done per request, but at the boot of the service.

internal/graph/storagewrapper.go
elbuo8: Uh? It was already shared across requests as this is not done per request, but at the boot of the service.

pkg/server/commands/batch_check_command.go
```diff
@@ -129,9 +143,8 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
 			checkQuery := NewCheckCommand(
 				bq.datastore,
 				bq.checkResolver,
-				bq.typesys,
 				WithCheckCommandLogger(bq.logger),
-				WithCacheController(bq.cacheController),
+				WithCheckCommandCache(bq.cacheController, bq.shouldCache, bq.cacheSingleflightGroup, bq.checkCache, bq.maxCheckCacheSize, bq.checkCacheTTL),
```
elbuo8, line 147: @ewanharris this is the sort of thing I'm looking to avoid by having the reverse ingress.

pkg/server/check.go
```diff
@@ -75,6 +74,7 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 		ContextualTuples: req.GetContextualTuples(),
 		Context:          req.GetContext(),
 		Consistency:      req.GetConsistency(),
+		Typesys:          typesys,
```
elbuo8, line 77: Why are we making this change in this PR?

pkg/server/check.go
```diff
@@ -60,13 +60,12 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
 	checkQuery := commands.NewCheckCommand(
-		s.checkDatastore,
+		s.datastore,
 		s.checkResolver,
-		typesys,
 		commands.WithCheckCommandLogger(s.logger),
 		commands.WithCheckCommandMaxConcurrentReads(s.maxConcurrentReadsForCheck),
 		commands.WithCheckCommandResolveNodeLimit(s.resolveNodeLimit),
-		commands.WithCacheController(s.cacheController),
+		commands.WithCheckCommandCache(s.cacheController, shouldCacheIterators(s), s.singleflightGroup, s.checkCache, s.checkIteratorCacheMaxResults, s.checkQueryCacheTTL),
```
elbuo8, line 68: Why are we introducing this change?

internal/graph/storagewrapper.go
miparnisari: Yes exactly, right now `NewCachedDatastore` is instantiated once at the server level, so the `sf` object is reused across requests, but with this PR i want all datastore wrappers to be able to be instantiated per-request, so i need to pass in the shared `sf`.

pkg/server/check.go
miparnisari: Removed 👍

pkg/server/check.go
miparnisari: So that when a request comes in, I can decide what datastore decorators I need to instantiate; specifically with this line I can decided whether we should call `NewCachedDatastore` or not.

elbuo8 (approved):

jpadilla (approved):

PR #2132, opened 2024-12-02
fix: prevent race condition of context cancellation

elbuo8:
## Description

In the case the client/upstream cancels the context, it should yield an error rather than a `false` resolution.

## Review Checklist


justincoh (approved):

miparnisari (approved):

internal/graph/check_fast_path.go
```diff
@@ -561,7 +561,7 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
 		return nil, ctx.Err()
 	case r, ok := <-rightChan:
 		if !ok {
-			return res, nil
+			return res, ctx.Err()
```
miparnisari, line 564: I like this. Feels like we should be doing this change everywhere within Check API.

PR #2129, opened 2024-11-29
fix: always return ErrUnauthorizedResponse for access control failures, but log the corresponding error

Vic-Dev:
## Description
For any access control error we want to return unauthorized, but we do want to log the corresponding error for users to debug.

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

If you haven't done so yet, we would appreciate it if you could star the [OpenFGA repository](https://github.com/openfga/openfga). :)

PR #2136, opened 2024-12-03
refactor: authorization error

miparnisari:
## Description
Refactor https://github.com/openfga/openfga/pull/2129 by:
- utilizing the existing `authorizationError` more consistently. `authz.go` throws this error, and then the server logs it and throws a public 403 error _without_ the actual cause.
- removing unnecessary sentinel errors ErrBadRequestMaxModulesInRequestExceeded and ErrUnknownAPIMethod

## Testing

1. Locally, I set these in `docker-compose.yaml`

```
    environment:
      - OPENFGA_DATASTORE_ENGINE=postgres
      - OPENFGA_DATASTORE_URI=postgres://postgres:password@postgres:5432/postgres?sslmode=disable
      - OPENFGA_DATASTORE_MAX_OPEN_CONNS=100 #see postgres container
      - OPENFGA_PLAYGROUND_ENABLED=false
      - OPENFGA_AUTHN_METHOD=oidc
      - OPENFGA_AUTHN_OIDC_ISSUER=https://dev-miparnisari.ca.auth0.com/
      - OPENFGA_AUTHN_OIDC_AUDIENCE=https://openfga.local
      - OPENFGA_EXPERIMENTALS=enable-access-control
      - OPENFGA_ACCESS_CONTROL_ENABLED=true
      - OPENFGA_ACCESS_CONTROL_STORE_ID=01JDWS23RC4RGW2G3YGQ8C8JG4
      - OPENFGA_ACCESS_CONTROL_MODEL_ID=01JDWS2Z9QB20VYT9GZFSG8YMP
      - OPENFGA_AUTHN_OIDC_CLIENT_ID_CLAIMS=unknown # <------ this is important
```

2. Executed

```
AUTH_TOKEN=$(curl --location 'https://dev-miparnisari.ca.auth0.com/oauth/token' \
--data '{
    "audience": "https://openfga.local",
    "client_id": "CHANGE-ME",
    "client_secret": "CHANGE-ME",
    "grant_type": "client_credentials"
}' | jq .access_token --raw-output)
```

3. Verified correct response

```
curl --location 'localhost:8080/stores/01JDWS23RC4RGW2G3YGQ8C8JG4/authorization-models' \
--header 'Authorization: Bearer $AUTH_TOKEN' \
--data '{
  "schema_version": "1.1",
  "type_definitions": [
    {
      "type": "user",
      "relations": {},
      "metadata": null
    }
  ]
}'

{"code":"forbidden","message":"the principal is not authorized to perform the action"}

```

4. Verified the logs in the server

```
openfga         | 2024-12-03T03:32:42.888Z      INFO    authorization failed    {"error": "client ID not found in context or is empty"}

```


internal/authz/authz.go
```diff
@@ -76,12 +75,9 @@ const (
 )
 
 var (
-	ErrUnauthorizedResponse                  = status.Error(codes.Code(openfgav1.AuthErrorCode_forbidden), "the principal is not authorized to perform the action")
-	ErrBadRequestMaxModulesInRequestExceeded = status.Error(codes.Code(openfgav1.AuthErrorCode_forbidden), fmt.Sprintf("the principal cannot write tuples of more than %v module(s) in a single request", MaxModulesInRequest))
```
rhamzeh, line 80: @miparnisari - I feel this one in particular we may need to keep public, otherwise folks will not be able to understand why they're not authorized even though they have access to each module individually

rhamzeh (commented):
Love this change!

internal/authz/authz.go
miparnisari: Oh. But the way it is today it will be overriden here: https://github.com/openfga/openfga/blob/81e97836a35f0e06530161ebeb9a5894d97c688f/pkg/server/server.go#L907-L911

🤔 feels like we need to change this too somehow.

rhamzeh (approved):

PR #2139, opened 2024-12-03
feat: add `Duration` to check resolution metadata

justincoh:
## Description

This PR adds a new metric, `check_duration_ms`, to measure how long it takes to resolve a check query specifically **at the command level**. We have an existing request duration metric at the server level for all methods, but that metric doesn't capture data on checks that are triggered by other methods (ListObjects, BatchCheck).

## Review Checklist


pkg/server/commands/check_command.go
```diff
@@ -122,9 +150,18 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
...
+
+	var datastoreQueryCount uint32
 	if resp != nil {
-		resp.ResolutionMetadata.DatastoreQueryCount = c.datastore.GetMetrics().DatastoreQueryCount
+		datastoreQueryCount = c.datastore.GetMetrics().DatastoreQueryCount
+		resp.ResolutionMetadata.DatastoreQueryCount = datastoreQueryCount
 	}
+
+	checkDurationHistogram.WithLabelValues(
+		utils.Bucketize(uint(datastoreQueryCount), []uint{50, 200}),
```
justincoh, line 161: `{50, 200}` are the same values [from our default config](https://github.com/openfga/openfga/blob/main/internal/server/config/config.go#L577). I'm not sure where exactly to put the bucket definitions in this context. The default config values are attached to the `server` object and not available here, and passing a measurement concern through to the commands feels like muddying the waters.

Any ideas?

pkg/server/commands/check_command.go
```diff
@@ -34,6 +52,7 @@ type CheckQuery struct {
 
 	resolveNodeLimit   uint32
 	maxConcurrentReads uint32
+	parentMethodName   string
```
justincoh, line 55: would `caller` be a more illustrative name for this?

tylernix: callerMethodName?

miparnisari: I like either of your suggestions more than `parent` :)

pkg/server/commands/check_command.go
miparnisari: I don't like hardcoded values, so i prefer passing down `requestDurationByQueryHistogramBuckets` from the server to the command.

pkg/server/commands/check_command.go
```diff
@@ -122,9 +150,18 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
 
 		return nil, nil, err
 	}
+
+	var datastoreQueryCount uint32
 	if resp != nil {
-		resp.ResolutionMetadata.DatastoreQueryCount = c.datastore.GetMetrics().DatastoreQueryCount
+		datastoreQueryCount = c.datastore.GetMetrics().DatastoreQueryCount
+		resp.ResolutionMetadata.DatastoreQueryCount = datastoreQueryCount
 	}
```
miparnisari, line 158: I think the time has come to _always_ track `DatastoreQueryCount`, regardless of whether the request errored or not.

pkg/server/commands/check_command.go
```diff
@@ -23,6 +27,20 @@ import (
 const (
 	defaultResolveNodeLimit           = 25
 	defaultMaxConcurrentReadsForCheck = math.MaxUint32
+	defaultParentMethodName           = "check"
```
miparnisari, line 30: I think this should be `""`. Otherwise, imagine you forget to call `WithParentMethodName` from Batch check or list objects, we will always think that these requests came from Check, when it reality it didn't. An empty string will allow us to know that we forgot to call `WithParentMethodName`

pkg/server/commands/check_command.go
justincoh: Josh and I had a conversation about this work and decided to change the approach entirely. Summarizing:

Instead of having these metrics emitted by the `check_command` itself, we're going to add a `Duration` to our existing [ResolveCheckResponseMetadata](https://github.com/openfga/openfga/blob/main/internal/graph/resolve_check_response.go#L3). The calling server methods will then be responsible for using (or not using) that information to emit metrics as they see fit. This way we don't need to pass in a `caller`, we don't need to worry about passing through bucket configuration, and we have a much cleaner separation of concerns.

Am going to refactor starting right now

pkg/server/batch_check.go
```diff
@@ -87,32 +87,33 @@ func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckReques
 		methodName,
 	).Observe(queryCount)
 
+	var batchResult = map[string]*openfgav1.BatchCheckSingleResult{}
+	for correlationID, outcome := range result {
+		batchResult[string(correlationID)] = transformCheckResultToProto(outcome)
+		s.emitCheckDurationMetric(outcome.CheckResponse.GetResolutionMetadata(), "batch_check")
```
jpadilla, line 93: nit: maybe for consistency this should use `methodName` from
above

jpadilla (approved):

pkg/server/batch_check.go
justincoh: going to merge this PR as is but will work this feedback into [the other PR](https://github.com/openfga/openfga/pull/2102#discussion_r1870373306) you just reviewed since it needs updated error handling anyway

PR #2149, opened 2024-12-07
refactor: cache settings

miparnisari:
Follow-up of https://github.com/openfga/openfga/pull/2126. Addressing this TODO: https://github.com/openfga/openfga/blob/9e40246c4169c649ef73207c2d5c06a36e090183/pkg/server/commands/check_command.go#L75-L85

There should be no change in behaviors.


internal/graph/cached_resolver.go
```diff
@@ -57,14 +56,6 @@ var _ CheckResolver = (*CachedCheckResolver)(nil)
 // instance.
 type CachedCheckResolverOpt func(*CachedCheckResolver)
 
-// WithMaxCacheSize sets the maximum size of the Check resolution cache. After this
-// maximum size is met, then cache keys will start being evicted with an LRU policy.
-func WithMaxCacheSize(size int64) CachedCheckResolverOpt {
```
miparnisari, line 62: FYI this is never referenced, except in tests, which can make do with the default.

internal/server/config/cache_settings_test.go
```diff
@@ -0,0 +1,181 @@
...
+				cacheSettings: CacheSettings{
+					CheckCacheLimit: 0,
+				},
+				expectedCreateNewCache: false,
+			},
+			{
+				name: "not_when_limit_is_zero_and_query_cache_enabled",
+				cacheSettings: CacheSettings{
+					CheckCacheLimit:        0,
+					CheckQueryCacheEnabled: false,
```
justincoh, line 27: should `CheckQueryCacheEnabled` be `true`? Seems like maybe from the test `name`

internal/server/config/cache_settings_test.go
```diff
@@ -0,0 +1,181 @@
...
+				},
+				expectedCreateNewCache: false,
+			},
+			{
+				name: "not_when_limit_is_zero_and_iterator_cache_enabled",
+				cacheSettings: CacheSettings{
+					CheckCacheLimit:           0,
+					CheckIteratorCacheEnabled: false,
```
justincoh, line 35: same question about `true`

internal/shared/shared.go
```diff
@@ -0,0 +1,52 @@
...
+	"sync"
+
+	"golang.org/x/sync/singleflight"
+
+	"github.com/openfga/openfga/internal/cachecontroller"
+	serverconfig "github.com/openfga/openfga/internal/server/config"
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+type SharedResources struct {
```
justincoh, line 14: Resources is a broad term, it'd be good to have something indicating that these are for `check` specifically. Maybe `SharedCheckResources`?

pkg/storage/storagewrappers/request.go
```diff
@@ -22,29 +18,14 @@ type RequestStorageWrapper struct {
...
-			serverCtx,
-			a,
-			checkCache,
-			int(maxCheckCacheSize),
-			checkCacheTTL,
-			singleflightGroup,
-			waitGroup,
-		) // to read tuples from cache
+	if cacheSettings.ShouldCacheIterators() {
+		// TODO(miparnisari): pass cacheSettings and resources directly, i can't now because i would create a package import cycle
```
miparnisari, line 27: @jpadilla i can remove this TODO if you think we should do this

justincoh (dismissed):

justincoh (approved):

Issue #925, opened 2023-08-04
Add wrapper to openfgav1.CheckRequest object into a openfgav1.ExtendedCheckRequest that defines the Equals method.


adriantam:
See https://protobuf.dev/getting-started/cpptutorial/#parsing-serialization

<img width="1051" alt="image" src="https://github.com/openfga/openfga/assets/5374887/daa43ac1-9314-4f01-8816-953283fbd4ea">

We should (in another PR that is a pre-requisite of this one) wrap the `openfgav1.CheckRequest` object into a `openfgav1.ExtendedCheckRequest` that defines the `Equals` method.

I've been wanting to do something similar with `openfgav1.AuthorizationModel`. Since `Typesystem` is basically an extended AuthorizationModel with validation and parsing methods, I've been wanting to stop referencing `openfgav1.AuthorizationModel` in our codebase and replace it with `Typesystem`

_Originally posted by @miparnisari in https://github.com/openfga/openfga/pull/891#discussion_r1281234707_

PR #2150, opened 2024-12-07
feat: support contextual tuples in Check requests that are optimized

miparnisari:
## Description

- Added a new `ReadStartingWithUserOptions.WithResultsSortedAscending`
   - For now, this will only be read within storage wrappers, not within datastores themselves.
- In Check API fast path v2:
   - use this option to create a combination of sorted iterators.
   - enable the fast path v2 optimization when `len(req.contextual tuples) > 0`

## Dependencies

https://github.com/openfga/openfga/pull/2197


internal/graph/check_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/openfga/openfga/pkg/logger"
 	"github.com/openfga/openfga/pkg/storage"
 	"github.com/openfga/openfga/pkg/storage/memory"
+	"github.com/openfga/openfga/pkg/storage/storagewrappers"
```
miparnisari, line 31: @jpadilla going back to https://github.com/openfga/openfga/pull/2117#discussion_r1862797277

I would need to move your storage wrapper together with the other storage wrappers, otherwise the linter complains:

```
internal/graph/builder.go:1: : import cycle not allowed in test (typecheck)
package graph
```

pkg/storage/storagewrappers/combinedtuplereader.go
```diff
@@ -16,15 +17,35 @@ func NewCombinedTupleReader(
...
 		RelationshipTupleReader: ds,
-		contextualTuples:        contextualTuples,
 	}
+
+	co := make([]*openfgav1.TupleKey, len(contextualTuples))
+	cu := make([]*openfgav1.TupleKey, len(contextualTuples))
+	for i, t := range contextualTuples {
+		cu[i] = tuple.NewTupleKeyWithCondition(t.GetObject(), t.GetRelation(), t.GetUser(), t.GetCondition().GetName(), t.GetCondition().GetContext())
+		co[i] = tuple.NewTupleKeyWithCondition(t.GetObject(), t.GetRelation(), t.GetUser(), t.GetCondition().GetName(), t.GetCondition().GetContext())
+	}
```
miparnisari, line 29: FYI to avoid race conditions in tests 😭

pkg/storage/iterators.go
```diff
@@ -397,3 +399,124 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+	return func(t *openfgav1.Tuple) string {
+		return t.GetKey().GetUser()
+	}
+}
+
+func ObjectMapper() TupleMapper {
+	return func(t *openfgav1.Tuple) string {
+		return t.GetKey().GetObject()
+	}
+}
```
miparnisari, line 424: TODO: these belong, conceptually, in `internal/graph/tuplemapper.go`, but i don't want to have `storage` import `graph`.

pkg/storage/storagewrappers/combinedtuplereader.go
```diff
@@ -15,16 +16,28 @@ import (
 func NewCombinedTupleReader(
 	ds storage.RelationshipTupleReader,
 	contextualTuples []*openfgav1.TupleKey,
-) storage.RelationshipTupleReader {
-	return &CombinedTupleReader{
+) *CombinedTupleReader {
```
miparnisari, line 19: FYI so i can make assertions on the inner fields in tests. Also it's a good pattern to return structs instead of interfaces so that the caller can decide which interface it needs. https://medium.com/@cep21/preemptive-interface-anti-pattern-in-go-54c18ac0668a

internal/graph/check_fast_path.go
```diff
@@ -928,7 +928,7 @@ func (c *LocalChecker) recursiveFastPath(ctx context.Context, req *ResolveCheckR
 
 	userIter, err := checkutil.IteratorReadStartingFromUser(ctx, typesys, ds, req,
 		tuple.ToObjectRelationString(tuple.GetType(req.GetTupleKey().GetObject()), req.GetTupleKey().GetRelation()),
-		nil)
+		nil, true)
```
elbuo8, line 931: Why? This still relies in hashsets.

pkg/storage/iterators.go
```diff
@@ -411,3 +412,121 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+
+// NewOrderedCombinedIterator is a thread-safe iterator that combines a list of iterators into a single ordered iterator.
+// All the input iterators must be individually ordered already according to mapper.
+// Iterators can yield the same value (as defined by mapper) multiple times, but it will only be returned once.
+func NewOrderedCombinedIterator(mapper TupleMapper, sortedIters ...TupleIterator) *OrderedCombinedIterator {
+	pending := make([]TupleIterator, 0, len(sortedIters))
+	for _, sortedIter := range sortedIters {
+		if sortedIter != nil {
```
elbuo8, line 445: In what case we would pass `nil` iterators? I believe I've seen this elsewhere, doesn't need to be handled in this PR.

pkg/storage/storagewrappers/combinedtuplereader.go
```diff
@@ -15,16 +16,28 @@ import (
...
+
+	ctr.contextualTuplesOrderedByObjectID = cu
+
+	return ctr
 }
 
 type CombinedTupleReader struct {
 	storage.RelationshipTupleReader
-	contextualTuples []*openfgav1.TupleKey
+	contextualTuplesOrderedByObjectID []*openfgav1.TupleKey
```
elbuo8, line 40: Why does this need to change?

pkg/storage/iterators.go
miparnisari: In no case. I copied from the other iterators

pkg/storage/iterators_test.go
```diff
@@ -318,6 +318,141 @@ func TestCombinedIterator(t *testing.T) {
...
+		require.ErrorIs(t, err, ErrIteratorDone)
+	})
+
+	t.Run("Next", func(t *testing.T) {
+		var testcases = map[string]struct {
+			iter1    TupleIterator
+			iter2    TupleIterator
+			expected []*openfgav1.Tuple
+		}{
+			`removes_duplicates`: {
```
miparnisari, line 339: missing test: when there are duplicates within one iterator

pkg/storage/iterators.go
elbuo8: Can we clean that up in a subsequent PR across the others 🤔

pkg/storage/iterators_test.go
```diff
@@ -318,6 +318,204 @@ func TestCombinedIterator(t *testing.T) {
...
+				}),
+				expected: []*openfgav1.Tuple{
+					{Key: tuple.NewTupleKey("document:1", "2", "user:a")},
+					{Key: tuple.NewTupleKey("document:1", "2", "user:b")},
+					{Key: tuple.NewTupleKey("document:1", "2", "user:c")},
+					{Key: tuple.NewTupleKey("document:1", "2", "user:d")},
+					{Key: tuple.NewTupleKey("document:1", "2", "user:e")},
+					{Key: tuple.NewTupleKey("document:1", "2", "user:f")},
+				},
+			},
```
adriantam, line 384: Can we add test where the last entry is also duplicate?

internal/graph/check_fast_path.go
```diff
@@ -928,7 +928,7 @@ func (c *LocalChecker) recursiveFastPath(ctx context.Context, req *ResolveCheckR
 
 	userIter, err := checkutil.IteratorReadStartingFromUser(ctx, typesys, ds, req,
 		tuple.ToObjectRelationString(tuple.GetType(req.GetTupleKey().GetObject()), req.GetTupleKey().GetRelation()),
-		nil)
+		nil, false)
```
adriantam, line 931: A bit concern that in some cases the iterator read starting from user is not sorted - would it lead to error if the caller relies on it in some path.

miparnisari: This is funny. @elbuo8 told me that putting `true` here was unnecessary. https://github.com/openfga/openfga/pull/2150#discussion_r1892402696

You both have a point. I guess I'll add a godoc explaining why this is false

pkg/storage/iterators_test.go
miparnisari: Added removes_duplicates_across_iterators_last_entry, let me know if it's what you had in mind

pkg/storage/iterators.go
```diff
@@ -411,3 +415,146 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+	once        *sync.Once
+	mapper      TupleMapper
+	pending     []TupleIterator  // GUARDED_BY(mu)
+	lastHead    *openfgav1.Tuple // GUARDED_BY(mu)
+	lastYielded *openfgav1.Tuple // GUARDED_BY(mu)
+}
+
+var _ TupleIterator = (*OrderedCombinedIterator)(nil)
+
+// TODO(miparnisari): this interface & impls belong, conceptually, in internal/graph/tuplemapper.go, but i don't want to have storage import graph.
```
elbuo8, line 430: Where is `graph` being used here?

These already exist, why are we re-declaring them?

pkg/storage/iterators.go
```diff
@@ -411,3 +415,146 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+		head, err := iter.Head(ctx)
+		if err != nil {
+			if errors.Is(err, ErrIteratorDone) {
+				iter.Stop()
+				c.pending[pendingIdx] = nil
+				continue
+			}
+			return minIdx, err
+		}
+
```
elbuo8, line 512: Why not store `c.mapper(head)` if it will be called in all 3 code paths.

Same applies for ` c.mapper(c.lastYielded)` within the `c.lastYielded != nil` block.

pkg/storage/iterators.go
miparnisari: Definitely

pkg/storage/iterators.go
miparnisari: I moved these changes to a different PR and it got merged already. If you feel strongly about this I can definitely send a follow-up PR.

pkg/storage/storagewrappers/combinedtuplereader.go
senojj: perhaps could instead be a doc comment on the property?

pkg/storage/storagewrappers/combinedtuplereader.go
senojj: I'd like to go a step further and ask "why return a pointer to this struct?". The struct contains two pointers, for a total of 16 bytes, so copying an instance of this type is likely less expensive than the inevitable escape to the heap and cost of reading the value from the heap.

elbuo8 (approved):

internal/graph/check_test.go
miparnisari: https://github.com/openfga/openfga/pull/2219

pkg/storage/storagewrappers/combinedtuplereader.go
miparnisari: This is a good idea. I can send a refactor PR.

internal/graph/check_fast_path.go
miparnisari: Updated existing tests.

pkg/storage/iterators.go
miparnisari: https://github.com/openfga/openfga/pull/2226

elbuo8 (approved):

pkg/storage/storagewrappers/combinedtuplereader.go
miparnisari: Partly tracked in https://github.com/openfga/openfga/issues/925

PR #2160, opened 2024-12-11
fix: update check cache key logic

justincoh:
## Description

This PR updates cache key generation logic

## Review Checklist


justincoh:
converting back to draft to fix a bug

pkg/storage/cache_test.go
```diff
@@ -171,9 +167,145 @@ func TestCheckCacheKeyContextualTuplesWithConditionsOrdering(t *testing.T) {
...
+		AuthorizationModelID: modelID,
+		TupleKey:             tupleKey,
+		ContextualTuples:     []*openfgav1.TupleKey{jonContextTwo},
+	})
+	require.NoError(t, err)
+
 	require.Equal(t, key1, key2)
 }
 
+func TestCheckCacheKeyContextualTuplesConditionOrderMatters(t *testing.T) {
```
miparnisari, line 278: I don't understand this test, and this test name :/

We _are_ sorting the contextual tuples (line 40 of `hasher.go`), so the order in which customers send them should not matter.

I also don't understand why this test is passing. I expect line 306 to be `require.Equal`.

justincoh: This is about to change. You're right that this should be `require.Equal`, that was a bug in the new logic I added in this PR which added `condition` as part of the cache key _without_ factoring it in as part of the sorting logic. The contextual tuples were only sorted based on user/object/relation. I'm about to push some changes that will also factor the `condition` into the sorting, and this test will be fixed.

There's also the case where multiple contextual_tuples have the _same_ condition name but different contexts, which may also need to be added into the sort logic; I haven't fully worked that one out yet.

internal/keys/hasher.go
```diff
@@ -68,9 +79,24 @@ func (t tupleKeysHasher) Append(h hasher) error {
...
+		key := strings.Builder{}
+		key.WriteString(tupleKey.GetObject())
+		key.WriteString("#")
+		key.WriteString(tupleKey.GetRelation())
+		key.WriteString("@")
+		key.WriteString(tupleKey.GetUser())
+
+		cond := tupleKey.GetCondition()
+		if cond != nil {
+			key.WriteString(cond.GetName())
```
willvedd, line 91: **Important:** We need to have a separator between condition and the preceding string. Otherwise, cache collisions will occur. For example, if you have `user:is_` with condition `grant_ok` the cache key would be the same as `user:is_grant_ok` without a condition. 

We'll want to separate with a character that is forbidden from the user field (which we may or not have based on [api spec](https://github.com/openfga/api/blob/24dec682b31d2864ca016f85f5c1a1918e92fd87/openfga/v1/openfga.proto#L161)).

willvedd (changes requested):

pkg/storage/cache_test.go
```diff
@@ -137,20 +137,16 @@ func TestCheckCacheKeyContextualTuplesOrdering(t *testing.T) {
 	require.Equal(t, key1, key2)
 }
 
-func TestCheckCacheKeyContextualTuplesWithConditionsOrdering(t *testing.T) {
+func TestCheckCacheKeyConsidersCondition(t *testing.T) {
```
willvedd, line 140: **Request:** Can we also add a test asserting that two requests, one with contextual tuple with and without a condition have unequal cache keys?

internal/keys/hasher.go
justincoh: From the spec it looks like we don't restrict `user` at all. Object and Relation each have patterns with disallowed characters, but not user. What do you think we should do here? `#` seems like a safe bet, but that already has meaning in FGA. Some other special character?

internal/keys/hasher.go
```diff
@@ -58,6 +59,16 @@ func (t tupleKeysHasher) Append(h hasher) error {
 			return sortedTupleKeys[i].GetUser() < sortedTupleKeys[j].GetUser()
 		}
 
+		cond1 := sortedTupleKeys[i].GetCondition()
+		cond2 := sortedTupleKeys[j].GetCondition()
+		if cond1 == nil && cond2 == nil {
+			return true
+		} else if cond1.GetName() != cond2.GetName() {
```
DanCech, line 66:
```suggestion
		if cond1 != nil && cond2 != nil && cond1.GetName() != cond2.GetName() {
```

DanCech: This is a slight change from the current behavior when only one condition is nil, if we want to retain sort order there we can tweak things slightly:
```suggestion
if (cond1 != nil || cond2 != nil) && cond1.GetName() != cond2.GetName() {
```

internal/keys/hasher.go
justincoh: I'm not sure we need one there, as these cache keys are never used in any way that requires them to be readable. They just have to be unique so that their [eventual uint conversion](https://github.com/openfga/openfga/blob/main/pkg/storage/cache.go#L196) results in a unique integer.

The existing separators are here to prevent overlap cases, e.g. `document:reviewer` vs `document:re#viewer`, but the combination of condition name and context hash will always be unique (since the context hasher sorts keys before writing).

justincoh: Update: I took a look back over this today and there _is_ an unaccounted for edge case which your suggestion would address 👍 . I'll add another separator

jpadilla (approved):

internal/keys/hasher.go
justincoh: Update 2: I've set this to auto-merge as-is once CI passes without adding the additional separator. I plan on adding it in a quick follow up PR, but doing so requires a refactor of how our hasher behaves and I don't want the blast radius on this PR to be any bigger than it already is. I will link that PR to this one when it's up

PR #2173, opened 2024-12-17
feat: add additional separator in contextualTuple cache key

justincoh:
## Description

This PR is a small update to how we calculate cache keys for tuples containing `contextual_tuples` which themselves have `conditions` that contain `context`—it resolves the discussion in the referenced PR below

## References

Resolves [this conversation](https://github.com/openfga/openfga/pull/2160#discussion_r1882841848) from #2160 

## Review Checklist


internal/keys/hasher.go
```diff
@@ -99,14 +108,10 @@ func (t tupleKeysHasher) Append(h hasher) error {
 		key.WriteString("@")
 		key.WriteString(tupleKey.GetUser())
 
-		if err := h.WriteString(key.String()); err != nil {
-			return err
-		}
+		_ = h.WriteString(key.String())
```
justincoh, line 111: xxhash package `.WriteString` has two definitions ([safe](https://github.com/cespare/xxhash/blob/main/xxhash_safe.go#L13) and [unsafe](https://github.com/cespare/xxhash/blob/main/xxhash_unsafe.go#L45)) but both always return a `nil` error. We can ignore it

internal/keys/hasher.go
```diff
@@ -13,20 +13,20 @@ import (
 )
 
 type hasher interface {
-	WriteString(value string) error
+	WriteString(value string)
```
justincoh, line 16: the `error` being returned here was just forwarding along xxhash.WriteString `error`, which is always `nil`.

internal/keys/hasher.go
jpadilla: Could this maybe be error prone if the underlying lib starts returning non-nil errors in a future release? seems unlikely, but i wonder if it could mask an issue...

justincoh: I suppose it's possible, but it seems unlikely. That'd require xxhash to update its main `write` method, which also always returns nil.

willvedd:
Sorry, I'm not following how this PR resolves https://github.com/openfga/openfga/pull/2160#discussion_r1882841848. IMO, this PR is all about intent. Is there something functional besides achieving better test coverage? It's a very small risk to swallow the error but why introduce it at all?

justincoh:
@willvedd That is a fair point, I'll pare this PR back down just to what matters to resolve the separator discussion from the earlier PR

internal/keys/hasher.go
justincoh: removed all of the refactor-related changes. The risk is very small, but why introduce it at all?

jpadilla (approved):

PR #2184, opened 2024-12-27
fix: cache-controller flags not respecting the environment variables

zackliston:
## Description
fix: fix issue with cache-controller flags not respecting the environment variables namely for cache-controller-enabled and cache-controller-ttl

Unsure of tests to add for this - I saw tests in the [original PR](https://github.com/openfga/openfga/pull/2155) that test the run.go file, however nothing for flags.go

This is my first contribution to this project, and my first time working in go as well, so any pointers in getting this PR up-to-snuff are appreciated. 

## References
Follow-on for [this PR](https://github.com/openfga/openfga/pull/2155) which enabled setting these flags in general.

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
  - Not applicable, fix for existing feature that is already documented.
- [x] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #2190, opened 2025-01-03
fix: cache controller not invalidating stale entry

adriantam:
## Description
Fix for cache controller not invalidating stale iterator cache and  sub problem cache properly.  There are three issues identified:

1. When running as stand alone, environment variable values not reflected in server's cache controller setting
2. Due to the use of [thine in-memory cache](https://github.com/Yiling-J/theine-go), the TTL for invalidated entry cannot be math.MaxInt as it results in no-op.  Instead, the TTL should be set to 1 year.
3. The algorithm needs to take care of the case where the lastVerified time was not set in the cache (either expired or first time). 

## References
Meant to be build on top of https://github.com/openfga/openfga/pull/2184

## Review Checklist


internal/cachecontroller/cache_controller.go
```diff
@@ -152,13 +153,15 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 		LastModified: changes[0].GetTimestamp().AsTime(),
```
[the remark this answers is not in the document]
adriantam: I believe we call read changes in reverse order, so that the first entry is the last change.

internal/cachecontroller/cache_controller.go
```diff
@@ -152,13 +153,15 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
...
+	hasCacheEntry := c.cache.Get(cacheKey) != nil
+
 	// set changelog entry as soon as possible for subsequent cache
 	// lookups have the entry and not have to wait on the existing singleflight group
 	c.cache.Set(cacheKey, entry, c.ttl)
 
 	lastVerified := time.Now().Add(-c.ttl)
 
-	if entry.LastModified.Before(lastVerified) {
+	if hasCacheEntry && entry.LastModified.Before(lastVerified) {
```
miparnisari, line 164: > The algorithm needs to take care of the case where the lastVerified time was not set in the cache (either expired or first time).

I think line 164 relates to this item in your PR description but I don't understand it. Why do we care if the cacheKey existed before? We are overwriting it line 160 anyway. I'm confused

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 ## [Unreleased]
 ### Fixed
 - Fixed missing binding between flags and environment variables for the cache controller feature [#2184](https://github.com/openfga/openfga/pull/2184)
+- Fixed missing invalidation by cache controller on cache iterator. [#2190](https://github.com/openfga/openfga/pull/2190)
```
miparnisari, line 12:
```suggestion
- Fixed incorrect invalidation by cache controller on cache iterator. [#2190](https://github.com/openfga/openfga/pull/2190)
```

internal/cachecontroller/cache_controller.go
adriantam: Imagine that in the case of findChangesAndInvalidate is ran for the first time (which is found via entry not present), we will need to invalidate the iterator cache because we do not know whether the iterator cache is set before / after the last read changes.

internal/cachecontroller/cache_controller.go
```diff
@@ -61,6 +62,10 @@ func NewNoopCacheController() CacheController {
 	return &NoopCacheController{}
 }
 
+// InMemoryCacheController will invalidate cache iterator (InMemoryCache) and sub problem cache (CachedCheckResolver) entries
+// that are more recent than the last write for the specified store.
+// Note that the invalidation is done asynchronously. Therefore, the cache iterator and sub problem cache may be stale
```
miparnisari, line 67:
```suggestion
// Note that the invalidation is done asynchronously, and only after a Check request is received.
```

internal/cachecontroller/cache_controller.go
```diff
@@ -190,8 +197,9 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 }
 
 func (c *InMemoryCacheController) invalidateIteratorCache(storeID string) {
-	// These entries do not need to expire
-	c.cache.Set(storage.GetInvalidIteratorCacheKey(storeID), &storage.InvalidEntityCacheEntry{LastModified: time.Now()}, math.MaxInt)
+	// These entries do not need to expire. We set it as OneYear instead of math.MaxInt so that it could not be interpreted
+	// as negative number.
```
miparnisari, line 201:
```suggestion
	// These entries do not need to expire. 
```

- The godoc I removed will not make sense a year from now so it's best to remove it.

- Instead of making this change here, you could make the change within our caching library wrapper, so that if you give it `math.maxInt` it transforms it to a value that Theine will not overflow on. The benefit will be that any user of the cache will not have to remember about this "gotcha". 

- Please add a unit test for this, here or in our caching wrapper if you take up on my suggestion :)

internal/cachecontroller/cache_controller_test.go
```diff
@@ -243,6 +266,14 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
 				cache.EXPECT().Set(k, gomock.Any(), gomock.Any())
 			}
 
+			if test.emptyCacheKey {
+				cache.EXPECT().Get(storage.GetChangelogCacheKey(test.storeID)).AnyTimes().Return(nil)
+			} else {
+				cache.EXPECT().Get(storage.GetChangelogCacheKey(test.storeID)).AnyTimes().Return(&storage.ChangelogCacheEntry{
+					LastModified: time.Now().Add(-20 * time.Second),
+				})
+			}
```
miparnisari, line 275: We shouldn't have complex logic inside test code, it makes it harder to understand the tests. Why not have each test case define the mocks it needs?

```go
{
	name:    "empty_cache_key",
	storeID: "8",
	setMocks: func(cache *mocks.MockInMemoryCache[any], datastore *mocks.MockOpenFGADatastore) {
		gomock.InOrder(
			datastore.EXPECT().ReadChanges(gomock.Any(), "8", gomock.Any(), gomock.Any()).Return([]*openfgav1.TupleChange{
				{
					Operation: openfgav1.TupleOperation_TUPLE_OPERATION_WRITE,
					Timestamp: timestamppb.New(time.Now().Add(-20 * time.Second)),
					TupleKey: &openfgav1.TupleKey{
						Object:   "test",
						Relation: "viewer",
						User:     "test",
					}},
			}, "", nil),
			cache.EXPECT().Get(storage.GetChangelogCacheKey("8")).AnyTimes().Return(nil),
			cache.EXPECT().Set(storage.GetChangelogCacheKey("8"), gomock.Any(), gomock.Any()),
			cache.EXPECT().Set(storage.GetInvalidIteratorCacheKey("8"), gomock.Any(), gomock.Any()),
		)
	},
},
```

This makes it easy to understand and assert on the order in which the operations occur. The order is particularly important because `hasCacheEntry` MUST happen before the `set`

internal/cachecontroller/cache_controller.go
miparnisari: This took me a while to understand, longer than i'd like to admit. I got thrown off by `hasCacheEntry` because i thought it was related to the `entry` variable, but it's not related at all. Can we rename?

- `hasCacheEntry` -> `lastInvalidationOccurred`? 
-  `lastVerified` -> `timestampOfLastInvalidation`?

Also, i put some debug logs and i cannot see a test where we exercise the `false && false` scenario.

miparnisari (changes requested):

internal/cachecontroller/cache_controller.go
```diff
@@ -47,6 +46,8 @@ var (
 	}, []string{"invalidation_required", "changes_size"})
 )
 
+const OneYear = time.Hour * 24 * 365
```
elbuo8, line 49: Perhaps this belongs in `pkg/storage/cache.go` as MAX_TIME so we don't get bitten once again by the internals of the cache implementation.

internal/cachecontroller/cache_controller_test.go
adriantam: the harness (i.e., []struct) do not have access to cache/datastore as they are defined inside.  However, I have updated the test to ensure they are in order

miparnisari: > the harness

I'm sorry I didn't paste a more complete example in my original comment. You can define a `setMock` function per test that does this. E.g. https://github.com/openfga/openfga/blob/9ec0ab1d01dec37dd540638f9429b7e851099520/pkg/server/commands/read_assertions_test.go#L34-L46

and in the test itself you do `testCase.setMocks(mocks...)` 😄  We use this pattern in a couple test files already.

elbuo8 (approved):

internal/cachecontroller/cache_controller.go
adriantam: updated as noted above

internal/cachecontroller/cache_controller_test.go
adriantam: Yes.  However, the mockDatastore has not yet been defined here.  Let me know if what I did now is sufficient (as there are no reason why the order of mocks is different per test).

internal/cachecontroller/cache_controller.go
miparnisari: Much appreciated :)

internal/cachecontroller/cache_controller_test.go
```diff
@@ -85,149 +88,209 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
...
-			name:               "empty_changelog",
-			storeID:            "1",
-			readChangesResults: &readChangesResponse{err: storage.ErrNotFound},
-			setCacheKeys: []string{
-				storage.GetInvalidIteratorCacheKey("1")},
+			name:    "empty_changelog",
+			storeID: "1",
+			setMocks: func(cache *mocks.MockInMemoryCache[any], datastore *mocks.MockOpenFGADatastore) {
+				gomock.InOrder(
+					datastore.EXPECT().ReadChanges(gomock.Any(), "1", gomock.Any(), gomock.Any()).Times(1).Return(nil, "", storage.ErrNotFound),
```
miparnisari, line 102: here and elsewhere: 

```suggestion
					datastore.EXPECT().ReadChanges(gomock.Any(), "1", gomock.Any(), opts).Times(1).Return(nil, "", storage.ErrNotFound),
```

this will ensure that `SortDesc: true`

internal/cachecontroller/cache_controller_test.go
```diff
@@ -85,149 +88,209 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
...
+							}},
+					}, "", nil),
+					cache.EXPECT().Get(storage.GetChangelogCacheKey("8")).Return(nil),
+					cache.EXPECT().Set(storage.GetChangelogCacheKey("8"), gomock.Any(), gomock.Any()),
+					cache.EXPECT().Set(storage.GetInvalidIteratorCacheKey("8"), gomock.Any(), gomock.Any()),
+				)
+			},
+		},
+		{
+			name:    "initial_check_for_invalidation_change_is_recent",
```
miparnisari, line 275: The difference between this testcase and `initial_check_for_invalidation`, is only the Timestamp of the tuples. Could you clarify how this matters?

internal/cachecontroller/cache_controller_test.go
miparnisari: Pushed suggested changes in [`9536375` (#2190)](https://github.com/openfga/openfga/pull/2190/commits/9536375cebd37a012919ae1646ca2957ca28faaa)

internal/cachecontroller/cache_controller_test.go
```diff
@@ -85,149 +88,209 @@ func TestInMemoryCacheController_findChangesAndInvalidate(t *testing.T) {
...
+				)
+			},
 		},
 		{
-			name:               "hard_error",
-			storeID:            "2",
-			readChangesResults: &readChangesResponse{err: storage.ErrCollision},
-			setCacheKeys: []string{
-				storage.GetInvalidIteratorCacheKey("2"),
+			name:    "hard_error",
```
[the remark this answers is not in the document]
adriantam: Agree in principle - but another philosophical take is not to bleed choice of implementation into the test.  Let's punt this suggestion for now :)

internal/cachecontroller/cache_controller_test.go
adriantam: added note in code comment but this is more as response to your feedback in https://github.com/openfga/openfga/pull/2190#discussion_r1905963147 regarding you cannot find the double negative case in test.

miparnisari (approved):

PR #2193, opened 2025-01-06
perf: make computation of cache keys faster

justincoh:
## Description

Within a check request, there can be many sub-problems to resolve. If caching is enabled, each sub-problem will generate its own cache key. However, only the Object, Relation, and User change between a parent problem and its children. Right now, we are computing an entirely new cache key for each subproblem, including the invariant parts.

This PR changes the cached_resolver's key generation logic as follows:
1. When the check request is first received, create an `InvariantCacheKey` using the `StoreID`, `AuthorizationModelID`, `Context` (if present) and `ContextualTuples` (if present)
2. When evaluating the problem, combine the pre-computed `InvariantCacheKey` with that specific problem's tuple-specific cache key
3. If a sub-problem is necessary, only recalculate the part of the cache key that changed (user, object, relation), and re-use the `InvariantCacheKey` determined at the start

## Review Checklist


pkg/storage/cache.go
```diff
@@ -162,33 +163,59 @@ type CheckCacheKeyParams struct {
...
 	)
+
 	if err != nil {
-		return "", err
+		return err
+	}
+
+	return nil
+}
+
```
senojj, line 194: This function doesn't need all of the data in params; just the TupleKey.

Additionally, it might be more suitable to make this string generation part of a `Tuple` type:
```go
type Tuple openfgav1.TupleKey

func (t *Tuple) GetObject() string {
	return (*openfgav1.TupleKey)(t).GetObject()
}

func (t *Tuple) GetRelation() string {
	return (*openfgav1.TupleKey)(t).GetRelation()
}

func (t *Tuple) GetUser() string {
	return (*openfgav1.TupleKey)(t).GetUser()
}

func (t *Tuple) String() string {
	tk := (*openfgav1.TupleKey)(t)

	return tk.GetObject() +
		"#" +
		tk.GetRelation() +
		"@" +
		tk.GetUser()
}

func From(tk *openfgav1.TupleKey) *Tuple {
	return (*Tuple)(tk)
}
```
Doing so would eliminate the need for this function and could be used like:
```go
t := tuple.From(params.TupleKey)
_, err := hasher.WriteString(t.String())
```

internal/graph/resolve_check_request.go
```diff
@@ -39,13 +40,49 @@ type ResolveCheckRequestMetadata struct {
...
 func NewCheckRequestMetadata() *ResolveCheckRequestMetadata {
 	return &ResolveCheckRequestMetadata{
 		DispatchCounter: new(atomic.Uint32),
 		WasThrottled:    new(atomic.Bool),
 	}
 }
 
+func NewResolveCheckRequest(
+	params ResolveCheckRequestParams,
+) (*ResolveCheckRequest, error) {
```
miparnisari, line 62: Could we add unit tests for this function?
- storeId, modelID and tuplekey and InvariantCacheKey cannot be empty

pkg/storage/cache.go
miparnisari: This suggestion is great, and it's basically starting to address https://github.com/openfga/openfga/issues/925 :) 

- For performance reasons we should stop using proto objects internally, they should only be part of requests or responses.
- There are lots of functionality that we can add to a basic Tuple: ToString, Equals.  

My advice on top of @senojj is to put this code in `pkg/tuple/tuple.go`. But this is moving us in the direction we want to go :)

pkg/storage/cache_test.go
```diff
@@ -403,6 +405,103 @@ func TestCheckCacheKeyWithContext(t *testing.T) {
...
+			writer: &validWriter,
+			params: &CheckCacheKeyParams{
+				TupleKey: &openfgav1.TupleKey{
+					Object:   "document:1",
+					Relation: "can_view",
+					User:     "user:anne",
+				},
+				ContextualTuples: []*openfgav1.TupleKey{
+					tuple.NewTupleKeyWithCondition("document:1", "viewer", "user:anne", "condition_name", contextStruct),
+				},
```
miparnisari, line 429: why is this here? it doesn't affect the output

miparnisari (dismissed):
Looks good! Just a few minor comments. Also I took the liberty of updating the PR title to something more aligned with what you wrote in the changelog.

CHANGELOG.md
```diff
@@ -7,17 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-### Security
-- Address [CVE-2024-56323](https://github.com/openfga/openfga/security/advisories/GHSA-32q6-rr98-cjqv) - an issue affecting Check and ListObjects results for users using Conditions in Contextual Tuples. Please see the CVE report for more details.
-
 ### Fixed
 - Fixed missing binding between flags and environment variables for the cache controller feature [#2184](https://github.com/openfga/openfga/pull/2184)
 - Fixed Read API to validate user field and assert presence of both type and value. [#2195](https://github.com/openfga/openfga/pull/2195)
 
+### Security
+- Address [CVE-2024-56323](https://github.com/openfga/openfga/security/advisories/GHSA-32q6-rr98-cjqv) - an issue affecting Check and ListObjects results for users using Conditions in Contextual Tuples. Please see the CVE report for more details.
+
```
justincoh, line 22: this bit got auto-reordered by `npx keep-a-changelog`. Our github action has been failing because it expects `Security` to come after `Fixed` in this section

internal/graph/resolve_check_request.go
```diff
@@ -39,13 +44,55 @@ type ResolveCheckRequestMetadata struct {
...
+		ContextualTuples:     params.ContextualTuples.GetTupleKeys(),
+		Context:              params.Context,
+		VisitedPaths:         make(map[string]struct{}),
+		RequestMetadata:      NewCheckRequestMetadata(),
+		Consistency:          params.Consistency,
+		// avoid having to read from cache consistently by propagating it
+		LastCacheInvalidationTime: params.CacheInvalidationTime,
+	}
+
+	err := r.SetRequestInvariantCacheKey()
```
senojj, line 80: the fields of `ResolveCheckRequest` are all exported. if the value of one of those fields is altered, it could change the value of the "invariant cache key", but the key is never recomputed.

a solution might be to change the fields so that they are not exported, and provide accessor/mutator methods for each field that should be publicly accessible or mutable. then you can recalculate the key on each change.

ideally, I would suggest calculating the "invariant cache key" in a function that is not a method of the type:
```go
func GenerateInvariantCacheKey(r *ResolveCheckRequest) (string, error) { 
    // ... 
}
```
then it is up to the caller to either maintain a reference to the initial computed key, or recompute after mutating the data.

justincoh: I dug a bit and almost every field can be made private without issue, the getters even exist already. I have a commit making the change but it blows up the diff in this PR, mostly due to patching lots of tests. For the sake of reviewer sanity I'm going to cherry pick that work into a separate refactor PR.

---
Unrelated:

> I would suggest calculating the "invariant cache key" in a function that is not a method of the type

How come? I don't have an opinion either way, but I could see an implementation where we have methods on the type that:
1. Set a property
2. Immediately call `setInvariantCacheKey` if needed

pkg/server/commands/check_command.go
```diff
@@ -99,17 +99,20 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
...
-		// avoid having to read from cache consistently by propagating it
-		LastCacheInvalidationTime: cacheInvalidationTime,
+	resolveCheckRequest, err := graph.NewResolveCheckRequest(
+		graph.ResolveCheckRequestParams{
+			StoreID:               params.StoreID,
+			TupleKey:              tuple.ConvertCheckRequestTupleKeyToTupleKey(params.TupleKey),
+			Context:               params.Context,
+			ContextualTuples:      params.ContextualTuples,
+			Consistency:           params.Consistency,
+			CacheInvalidationTime: cacheInvalidationTime,
```
elbuo8, line 109: This rename?

justincoh: Fortunately this is only in the parameters being passed into the constructor function, but I agree this should be `LastCacheInvalidationTime` regardless—will change

internal/graph/resolve_check_request.go
senojj: > I would suggest calculating the "invariant cache key" in a function that is not a method of the type.
> > How come? I don't have an opinion either way, but I'd like to hear your thoughts

The current function is entirely a side-effect; it takes no input and produces no output. Understanding such a function requires either reading the documentation or implementation code, as the function signature does not itself provide enough information.

Let's look at an example:
```go
type Addition struct {
    a      int
    b      int
    result int
}

func NewAddition(a, b int) *Addition {
    return &Addition{
        a: a,
        b: b
    }
}

func (a *Addition) Sum() {
    a.result = a.a + a.b
}

func (a *Addition) GetResult() int {
    return a.result
}
```
There are many side-effects and a lot of indirection. Let's look at an alternative solution:
```go
func Sum(a, b int) int {
    return a + b
}
```
The function accepts input and returns output without any side-effects. I feel that this option is much clearer.

The above may seem like a contrived example, but it is very similar to what is going on in this PR.

justincoh: Those are valid points and this is a clear example, ty Josh

internal/graph/resolve_check_request.go
```diff
@@ -39,13 +44,57 @@ type ResolveCheckRequestMetadata struct {
...
+		VisitedPaths:         make(map[string]struct{}),
+		RequestMetadata:      NewCheckRequestMetadata(),
+		Consistency:          params.Consistency,
+		// avoid having to read from cache consistently by propagating it
+		LastCacheInvalidationTime: params.LastCacheInvalidationTime,
+	}
+
+	if r.GetAuthorizationModelID() == "" {
+		return nil, errors.New("missing authorization_model_id")
+	}
```
senojj, line 82: this can be checked before allocating a `ResolveCheckRequest` instance:
```go
if params.AuthorizationModelID == "" {
    // ...
}
```
the same is true of the handling of the store identifier below...

internal/graph/cached_resolver.go
```diff
@@ -139,13 +142,20 @@ func (c *CachedCheckResolver) ResolveCheck(
...
+
+	// this err is always nil, according to xxhash docs
 	if err != nil {
-		c.logger.Error("cache key computation failed with error", zap.Error(err))
-		telemetry.TraceError(span, err)
 		return nil, err
 	}
 
+	// Get unique hash int and convert to string for storage
+	cacheKey := strconv.FormatUint(hasher.Sum64(), 10)
```
justincoh, line 157: I have mixed feelings about moving the actual `xxhash` parts of cache key generation out of pkg `storage` and into the `cached_resolver` file itself. I decided to do it because:
1. Creating a unique cache string based on the `Check` feels separate enough from its eventual serialization
2. Inside `cached_resolver` is the only place we actually write to the check cache
3. It makes the cache key generation code (in pkg `storage`) more easily testable

Going further, I'm not sure we _need_ the `hasher.Sum64()` (and associated `xxhash` package) at all? Its main value add is in shortening the cache keys, which can get quite long if a request contains lots of context or contextual tuples. I'd be interested to hear other's thoughts on this point.

justincoh:
The benchmark failure here is legit on `B/op`. This branch appears to run the check benchmark suite slightly faster but at the expense of Bytes, from local testing:

This is the branch from this PR:
image


This is current main:
image

pkg/server/commands/batch_check_command.go
```diff
@@ -251,10 +253,12 @@ func generateCacheKeyFromCheck(check *openfgav1.BatchCheckItem, storeID string,
...
-	cacheKey, err := storage.GetCheckCacheKey(cacheKeyParams)
+	hasher := xxhash.New()
+	err := storage.WriteCheckCacheKey(hasher, cacheKeyParams)
 	if err != nil {
 		return "", err
 	}
 
-	return CacheKey(cacheKey), nil
+	keyStr := strconv.FormatUint(hasher.Sum64(), 10)
+	return CacheKey(keyStr), nil
```
justincoh, line 263: swapped in `xxhash.Digest` here because it is faster to write and results in smaller keys

internal/graph/resolve_check_request_test.go
```diff
@@ -180,3 +180,37 @@ func TestDefaultValueRequestMetadata(t *testing.T) {
...
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			_, err := NewResolveCheckRequest(tc.params)
+			if tc.error {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
```
miparnisari, line 212:
```suggestion
				require.NoError(t, err)
				require.NotEmpty(t, res.invariantCacheKey)
```

pkg/tuple/tuple.go
```diff
@@ -11,6 +11,34 @@ import (
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 )
 
+type Tuple openfgav1.TupleKey
+
+func (t *Tuple) GetObject() string {
+	return (*openfgav1.TupleKey)(t).GetObject()
+}
```
miparnisari, line 18: Are we not missing `GetCondition()`?

pkg/server/commands/list_objects_test.go
```diff
@@ -310,18 +313,24 @@ func TestDoesNotUseCacheWhenHigherConsistencyEnabled(t *testing.T) {
...
-	cacheKey, err := graph.CheckRequestCacheKey(req)
+	})
 	require.NoError(t, err)
 
+	// Preload the cache the way the cached_resolver does it
+	tup := tuple.From(req.GetTupleKey())
+	cacheKeyString := tup.String() + req.GetInvariantCacheKey()
+	hasher := xxhash.New()
+	_, _ = hasher.WriteString(cacheKeyString)
+	cacheKey := strconv.FormatUint(hasher.Sum64(), 10)
```
miparnisari, line 332: I'm not fond of having so much business logic in a test. 

- This logic already exists in `CachedCheckResolver.ResolveCheck`, can we build a helper that the test can re-use?
- Can we rewrite the test so that we use a mock of a cache and we hardcode the cacheKey?

justincoh: Yeah this smelled bad to me when I wrote it as well, but since it's a one-off I just left it as is. This bit of code does lend itself nicely to a small helper method though so I will go with your first suggestion and create a helper

internal/graph/resolve_check_request.go
```diff
@@ -39,13 +48,63 @@ type ResolveCheckRequestMetadata struct {
 	WasThrottled *atomic.Bool
 }
 
+type ResolveCheckRequestParams struct {
```
miparnisari, line 51: (No action item) I feel a bit iffy about having to structs (ResolveCheckRequest and ResolveCheckRequestParams) that are so similar. Maybe we can revisit this later

pkg/storage/cache_test.go
```diff
@@ -757,88 +754,188 @@ func TestCheckCacheKeyWithContext(t *testing.T) {
...
+				require.NoError(t, err)
+				require.Equal(t, test.output, test.writer.String())
+			}
+		})
+	}
+}
+
+// TupleCacheKey is calculated on every invocation of check, potentially multiple times
+// per request.
+func BenchmarkTupleCacheKey(b *testing.B) {
```
miparnisari, line 894:
```suggestion
func BenchmarkWriteCheckCacheKey(b *testing.B) {
```

pkg/tuple/tuple.go
justincoh: It's true that TupleKeys in check won't have a condition, but I'm not sure I want to make that change since it precludes this `Tuple` type from being used with Conditions in the future.

miparnisari (approved):

PR #2196, opened 2025-01-08
chore: rename cache controller to cache invalidator

miparnisari:
## Description
Rename only, no functional change should be expected. 
- The environment variables for the feature remain the same.
- So do the metric names.

## References
https://rygo6.github.io/2019/04/19/Controller-is-Probably-a-Bad-Name.html

## Testing

```
make build
./dist/openfga run --cache-controller-ttl=25s --cache-controller-enabled=true
```

See the logs:

```
 INFO    starting openfga service...  ,"CacheController":{"Enabled":true,"TTL":25000000000},
```


internal/cacheinvalidator/cache_invalidator.go
```diff
@@ -83,11 +86,11 @@ func NewCacheController(ds storage.OpenFGADatastore, cache storage.InMemoryCache
 	return c
 }
 
-func (c *InMemoryCacheController) DetermineInvalidation(
+func (c *InMemoryCacheInvalidator) GetLastWriteAndInvalidate(
 	ctx context.Context,
 	storeID string,
 ) time.Time {
-	_, span := tracer.Start(ctx, "cacheController.DetermineInvalidation", trace.WithAttributes(attribute.Bool("cached", false)))
+	_, span := tracer.Start(ctx, "cacheInvalidator.GetLastWriteAndInvalidate", trace.WithAttributes(attribute.Bool("cached", false)))
```
miparnisari, line 93: I've never received proof from the community that someone is using the traces. It seems that only us, the maintainers, find them useful. Therefore, renaming.

justincoh (dismissed):

PR #2197, opened 2025-01-08
chore: introduce new ordered combined iterator

miparnisari:
## Description
Add an iterator that is similar to [CombinedIterator](https://github.com/openfga/openfga/blob/9ec0ab1d01dec37dd540638f9429b7e851099520/pkg/storage/iterators.go#L34-L36) but with the restriction that the inner iterators must be sorted already in ascending fashion. This enables an implementation of [merging of K sorted lists](https://leetcode.com/problems/merge-k-sorted-lists/)

## References
Ripped from https://github.com/openfga/openfga/pull/2150

### DIFF
--- a/pkg/storage/iterators.go
+++ b/pkg/storage/iterators.go
@@ -3,6 +3,8 @@ package storage
 import (
 	"context"
 	"errors"
+	"fmt"
+	"slices"
 	"sync"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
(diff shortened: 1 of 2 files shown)


pkg/storage/iterators.go
```diff
@@ -407,6 +409,149 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+IterateOverPending:
+	for pendingIdx, iter := range c.pending {
+		head, err := iter.Head(ctx)
+		if err != nil {
+			if errors.Is(err, ErrIteratorDone) {
+				iter.Stop()
+				c.pending[pendingIdx] = nil
+				continue
+			}
+			return minIdx, err
```
adriantam, line 503: Can you explain why we return minIdx vs. -1?

pkg/storage/iterators.go
```diff
@@ -407,6 +409,149 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+			// discard duplicate values
+			for c.mapper(head) == c.mapper(c.lastYielded) {
+				head, err = iter.Next(ctx)
+				if err != nil {
+					if errors.Is(err, ErrIteratorDone) {
+						iter.Stop()
+						c.pending[pendingIdx] = nil
+						continue IterateOverPending
+					}
+					return minIdx, err
```
adriantam, line 519: similarly, why minIdx vs. -1?

pkg/storage/iterators.go
```diff
@@ -407,6 +409,149 @@ func NewConditionsFilteredTupleKeyIterator(iter TupleKeyIterator, filter TupleKe
...
+			}
+			return minIdx, err
+		}
+
+		if c.lastYielded != nil {
+			if c.mapper(head) < c.mapper(c.lastYielded) {
+				return minIdx, fmt.Errorf("iterator %d is not in ascending order", pendingIdx)
+			}
+			// discard duplicate values
+			for c.mapper(head) == c.mapper(c.lastYielded) {
```
adriantam, line 511: Curious why we choose to prune at the next head() vs. in the current head()?  In this way, we don't have to remember lastYielded and makes the code easier to understand.

adriantam (commented):
1.  iterators.go / iterators_test.go is getting big, perhaps it will be good to split them up into different file next time (or as follow up)
2. codecov points out some missing coverage.  Can you address them?  TY.

miparnisari:
> iterators.go / iterators_test.go is getting big, perhaps it will be good to split them up into different file next time (or as follow up)

I agree, and I actually ran into a problem with this. At the very least we need to split the Iterator interface from the implementations. Because right now, the build fails with "import cycle" if i try to use a mocked iterator from within a test inside the `storage` folder. So in its current shape, we cannot import _any_ mocks from within _any_ tests in the `storage` fodler. 😢 

image

pkg/storage/iterators.go
miparnisari: Good point, updated and added to godoc of the fn.

pkg/storage/iterators.go
miparnisari: It's been a few weeks since I wrote this code :( IIRC i did it the way you suggested originally but @elbuo8 suggested the current implementation because it's faster.

adriantam: Great.  Maybe add comments in the code to point out this is done for performance as it is not obvious at first.

adriantam (approved):

PR #2198, opened 2025-01-08
fix: iterator need to be flushed after invalidation

adriantam:
## Description
Fix case where after cache controller invalidated iterator entries, the iterator cache is not flushed after Stop().

## Review Checklist


adriantam:
Marked as skip log due to similarities with https://github.com/openfga/openfga/pull/2190

internal/graph/storagewrapper.go
```diff
@@ -449,13 +449,6 @@ func (c *cachedIterator) Stop() {
 		defer c.wg.Done()
 		defer c.iter.Stop()
 
-		// if cache is already set, we don't need to drain the iterator
```
adriantam, line 452: The cache may be present but invalid.  Thus, we still need to drain the iterator.

adriantam: we kept the above logic but added in condition on invalidation

internal/graph/storagewrapper.go
```diff
@@ -232,26 +232,30 @@ func (c *CachedDatastore) Read(
...
-func (c *CachedDatastore) findInCache(store, key string, invalidEntityKeys []string) (*storage.TupleIteratorCacheEntry, bool) {
+func findInCache(cache storage.InMemoryCache[any], store, key string, invalidEntityKeys []string) (*storage.TupleIteratorCacheEntry, bool) {
 	var tupleEntry *storage.TupleIteratorCacheEntry
+	var ok bool
 
 	// The iterator cache has a TTL and will eventually consistent.
-	if res := c.cache.Get(key); res != nil {
-		tupleEntry = res.(*storage.TupleIteratorCacheEntry)
+	if res := cache.Get(key); res != nil {
+		tupleEntry, ok = res.(*storage.TupleIteratorCacheEntry)
```
adriantam, line 241: we actually should validate the tupleEntry to ensure we are are in a sane state and not-panic.  Better to invalidate cache then panic / die

elbuo8 (approved):

PR #2200, opened 2025-01-08
perf: improve performance of a union of recursive ttu with nested algebraic operations

miparnisari:
## Description
This PR is meant to increase performance in Check API for some authorization models. E.g when querying `rel1`:

```
model
	schema 1.1
type user
type document
	relations
		define rel1: rel2 or rel1 from parent
		define parent: [document]
		define rel2: [user] and rel3
		define rel3: [user]
```

For this purpose, I introduced:
- a new method within `optimizations.go` to detect when we can apply this optimization
- an `objectProvider` interface with two implementations (`simple` is a restructuring of an existing optimization, `complex` provides the new optimization)

NOTE:  my code (v2) is a generalization of v1, so v2 can in the future replace v1 in all scenarios (unless i misunderstood the requirements doc)

## References
Depends on https://github.com/openfga/openfga/pull/2231


pkg/typesystem/optimizations.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"github.com/openfga/language/pkg/go/graph"
+
+	"github.com/openfga/openfga/pkg/tuple"
+)
+
+// This file contains methods to detect whether an authorization model exhibits certain characteristics.
+// This information can then be used to increase performance of a Check request.
```
miparnisari, line 11: This new file is a proposal - I can revert if we don't like it, or I can send a different PR with this update

internal/graph/check_fast_path.go
```diff
@@ -994,10 +985,119 @@ func (c *LocalChecker) recursiveTTUFastPath(ctx context.Context, req *ResolveChe
...
+
+var _ objectProvider = &simpleRecursiveObjectProvider{}
+
+func (s *simpleRecursiveObjectProvider) close() {
+	if s.mapper != nil {
+		s.mapper.Stop()
+	}
+}
+
+func (s *simpleRecursiveObjectProvider) build(ctx context.Context, req *ResolveCheckRequest) (chan usersetMessage, error) {
```
miparnisari, line 1014: FYI the code in this function was previously within `recursiveFastPath`

internal/graph/check.go
```diff
@@ -1411,6 +1411,9 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		if typesys.RecursiveTTUCanFastPath(objectTypeRelation, userType) {
 			resolver = c.recursiveTTUFastPath
 			span.SetAttributes(attribute.String("resolver", "recursivefastpathv1"))
+		} else if ok, _ := typesys.IsRelationWithRecursiveTTUAndAlgebraicOperations(objectType, relation, userType); ok && c.optimizationsEnabled {
```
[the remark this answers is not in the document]
elbuo8: The order of V2 and V1 should be flipped also. Since if its enabled, we should go for V2 if possible.

internal/graph/check_fast_path.go
elbuo8: It's half new half old. The majority of the implementation existed, its now encapsulated differently.

internal/graph/check_fast_path.go
```diff
@@ -994,10 +985,119 @@ func (c *LocalChecker) recursiveTTUFastPath(ctx context.Context, req *ResolveChe
...
+
+func (c *complexRecursiveObjectProvider) build(ctx context.Context, req *ResolveCheckRequest) (chan usersetMessage, error) {
+	objectType, relation := tuple.GetType(req.GetTupleKey().GetObject()), req.GetTupleKey().GetRelation()
+	userType := tuple.GetType(req.GetTupleKey().GetUser())
+	typesys, _ := typesystem.TypesystemFromContext(ctx)
+	_, operands := typesys.IsRelationWithRecursiveTTUAndAlgebraicOperations(objectType, relation, userType)
+
+	outChannel := make(chan usersetMessage, 1)
+	defer close(outChannel)
+	pool := concurrency.NewPool(ctx, len(operands))
```
elbuo8, line 1048: There should be a controlled upper bound to this.

internal/graph/check_fast_path.go
```diff
@@ -994,10 +985,119 @@ func (c *LocalChecker) recursiveTTUFastPath(ctx context.Context, req *ResolveChe
...
+			return nil
+		})
+	}
+
+	err := pool.Wait()
+	if err != nil {
+		return nil, err
+	}
+
+	return outChannel, nil
```
elbuo8, line 1084: So by the time the channel is yielded, all the goroutines have been processed? That doesn't allow you to wrap up early on a match (or a ctx cancellation).

internal/graph/check_fast_path.go
elbuo8: Do we need to have a `simple` and `complex` implementation? Seems to me the "complex" one can resolve the simple one, ultimately it's the same cost/underlying implementation.

pkg/typesystem/optimizations.go
elbuo8: I can't say I agree with the naming but I could see the value of aggregating the methods that are used to identify a resolution strategy within its own file.

miparnisari: I'm open to new names :)

internal/graph/check.go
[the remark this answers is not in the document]
miparnisari: my code is a generalization of `c.recursiveTTUFastPath`, so v2 can replace v1 in **all** scenarios (unless i misunderstood the requirements doc).

if i flip the order, `c.recursiveTTUFastPath` will never run and will become irrelevant. might as well remove it.

i don't know if i want to flip the order just now.

pkg/typesystem/optimizations.go
```diff
@@ -0,0 +1,101 @@
...
+				objType, relationName := tuple.SplitObjectRelation(nodeLabel)
+				if relationName != "" {
+					getRelation, _ := t.GetRelation(objType, relationName)
+					usersets = append(usersets, Operand{
+						Rewrite:      getRelation.GetRewrite(),
+						RelationName: relationName,
+					})
+				} else {
+					usersets = append(usersets, Operand{
+						Rewrite:      This(),
```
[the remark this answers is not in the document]
miparnisari: agree, but `This()` is used in toooons of places 😓

internal/graph/check_fast_path.go
miparnisari: > It's half new half old. The majority of the implementation existed, its now encapsulated differently.

WDYM, "half new half old"? I didn't change how it works, I just put it somewhere else.

pkg/typesystem/optimizations.go
```diff
@@ -0,0 +1,101 @@
...
+	Rewrite      *openfgav1.Userset
+	RelationName string
+}
+
+// IsRelationWithRecursiveTTUAndAlgebraicOperations returns true if all these conditions apply:
+// - node[objectType#relation].weights[userType] = infinite
+// - node[objectType#relation] has only 1 edge, and it's to an OR node
+// - The OR node has one TTU edge with weight infinite for the terminal type and the computed relation for the TTU is the same
+// - Any other edge leaving the OR node has weight 1 for the terminal type.
+// If true, it returns an array containing all the other operands leaving the OR.
```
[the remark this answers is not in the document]
miparnisari: updated, let me know if it makes more sense

pkg/typesystem/optimizations.go
adriantam: I love that we abstract the strategy here as I can envision list objects optimization will need something similar

pkg/typesystem/optimizations.go
```diff
@@ -0,0 +1,101 @@
...
+	unionNode := edge.GetTo()
+
+	edgesFromUnionNode, ok := t.authzWeightedGraph.GetEdgesByNode(unionNode)
+	if !ok {
+		return false, usersets
+	}
+
+	ttuEdgeSatisfiesCond, restOfEdgesAreWeight1, ttuEdgeCount := false, true, 0
+
+	for _, edgeFromUnionNode := range edgesFromUnionNode {
```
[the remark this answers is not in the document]
adriantam: Agree in breaking apart into sub-methods.  The code as is has high cyclomatic complexity.

internal/graph/check_fast_path.go
elbuo8: >  I wrote it to keep v1 code working.

We not keep v1 completely untouched?

miparnisari: Yes, but note that if the experimental is enabled, all code that used to go thru v1 now will go through v2 (which is a generalization of v1)

internal/graph/check.go
```diff
@@ -1408,10 +1408,23 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 			}
 		}
 
-		if typesys.RecursiveTTUCanFastPath(objectTypeRelation, userType) {
+		choice1 := typesys.RecursiveTTUCanFastPath(objectTypeRelation, userType)
```
elbuo8, line 1411: Whats going on here? Guessing this is you testing? Because we shouldn't attempt to apply these algorithms like this.

internal/graph/object_providers.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/openfga/openfga/internal/concurrency"
+	openfgaErrors "github.com/openfga/openfga/internal/errors"
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/tuple"
+	"github.com/openfga/openfga/pkg/typesystem"
+)
+
+// ObjectProvider is an interface that abstracts the building of a channel that holds object IDs or usersets.
+// It must close the channel when there are no more results.
+type ObjectProvider interface {
```
elbuo8, line 18: Why public?

internal/graph/object_providers.go
```diff
@@ -0,0 +1,142 @@
...
+	concurrencyLimit uint32
+	wg               sync.WaitGroup
+	ts               *typesystem.TypeSystem
+}
+
+func NewComplexRecursiveObjectProvider(concurrencyLimit uint32, ts *typesystem.TypeSystem) (*ComplexRecursiveObjectProvider, error) {
+	if ts == nil {
+		return nil, fmt.Errorf("%w: nil typesystem", openfgaErrors.ErrUnknown)
+	}
+	return &ComplexRecursiveObjectProvider{concurrencyLimit: concurrencyLimit, wg: sync.WaitGroup{}, ts: ts}, nil
```
elbuo8, line 76: Undefined limit would be 0, would that panic or be unbounded?

internal/graph/check.go
miparnisari: Could you be more specific? What is "like this"?

I wanted to make it so that if v2 applies, and the experimental is enabled, it will supersede v1.

internal/graph/object_providers.go
miparnisari: If the limit were 0, it would panic here: 
image

I can send a different PR to ensure that doesn't happen

internal/graph/object_providers.go
miparnisari: Made private

internal/graph/check_fast_path.go
miparnisari: Fixed

justincoh (approved):

internal/graph/object_providers.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	objectType, relation := tuple.GetType(req.GetTupleKey().GetObject()), req.GetTupleKey().GetRelation()
+	userType := tuple.GetType(req.GetTupleKey().GetUser())
+	operands, ok := c.ts.IsRelationWithRecursiveTTUAndAlgebraicOperations(objectType, relation, userType)
+	if !ok {
+		return nil, fmt.Errorf("%w: unsupported model", openfgaErrors.ErrUnknown)
+	}
+
+	outChannel := make(chan usersetMessage, len(operands)) // buffered so that the first writes don't block on the caller reading from the channel
+	pool := concurrency.NewPool(ctx, max(int(c.concurrencyLimit), len(operands)))
```
elbuo8, line 96: Could there be a case where `operands` length == 0? If so, should that be handled?

miparnisari: No, it cannot happen. `IsRelationWithRecursiveTTUAndAlgebraicOperations`  has:

```go
satisfies := ttuEdgeSatisfiesCond && len(usersets) >= 1
return usersets, satisfies
```

internal/graph/check.go
miparnisari: Amended

internal/graph/check.go
```diff
@@ -1411,10 +1411,18 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		if typesys.RecursiveTTUCanFastPath(objectTypeRelation, userType) {
 			resolver = c.recursiveTTUFastPath
 			span.SetAttributes(attribute.String("resolver", "recursivefastpathv1"))
-		} else if c.optimizationsEnabled && typesys.TTUCanFastPathWeight2(objectType, relation, userType, rewrite.GetTupleToUserset()) {
-			resolver = c.checkTTUFastPathV2
-			span.SetAttributes(attribute.String("resolver", "fastpathv2"))
+		} else if c.optimizationsEnabled {
```
elbuo8, line 1414: This seems off. In a case where `RecursiveTTUCanFastPath` is true, you wouldn't go into the new implementation. And even if you were, you are paying the cost twice. 

So if anything it should become:

if optimizationsEnabled {
  check fast paths v2
} else {
  the original ones
}

elbuo8: And within the order of the optimizations to check, its more common to hit non recursive implementations, so that should come first.

miparnisari: updated

pkg/typesystem/optimizations.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	satisfies := ttuEdgeSatisfiesCond && ttuEdgeCount >= 1
+	if !satisfies {
+		usersets = nil
+	}
+	return usersets, satisfies
+}
+
+func populateUsersetsMap(t *TypeSystem, edgeFromUnionNode *graph.WeightedAuthorizationModelEdge, usersets Operands, relation string) Operands {
```
yissellokta, line 89: we don't need to create a different way to traverse the graph when this problem is already resolved by TTU or userset weight 2 with algebraic operations (referring to the way we solve the algebraic operations for weight=2)

pkg/typesystem/optimizations.go
```diff
@@ -0,0 +1,107 @@
...
+// Operands is a map of relation names to their rewrites.
+type Operands map[string]*openfgav1.Userset
+
+// IsRelationWithRecursiveTTUAndAlgebraicOperations returns true if all these conditions apply:
+// 1. Node[objectType#relation].weights[userType] = infinite
+// 2. Node[objectType#relation] has only 1 edge, and it's to an OR node
+// 3. The OR node has one or more TTU edge with weight infinite for the terminal type and the computed relation for the TTU is the same
+// 4. Any other edge coming out of the OR node that has a weight for terminal type, it should be weight 1
+// If true, it returns a map of Operands (edges) leaving the OR.
+func (t *TypeSystem) IsRelationWithRecursiveTTUAndAlgebraicOperations(objectType, relation, userType string) (Operands, bool) {
```
yissellokta, line 22: Very similar function is already implemented for simple recursion TTU, don't need to make a completely new one, let's reuse the similar code base and refactor only the necessary to know if compute of weight=1 or direct type

yissellokta, line 22: Very similar function is already implemented for simple recursion TTU, don't need to make a completely new one, let's reuse the similar code base and refactor only the necessary to know if compute of weight=1 or direct type

pkg/typesystem/optimizations.go
miparnisari: @yissellokta the idea is not to have JUST this method here, but all the other methods that try to assess whether an optimization is applicable.

yissellokta (approved):

justincoh (approved):

PR #2216, opened 2025-01-15
fix(InMemoryCacheController): incorrect timestamp comparison for partial invalidation

adriantam:
## Description

Correct logic for InMemoryCacheController invalidation logic by compare change time against iterator cache TTL instead of cache controller TTL.

There is a subtle bug in the invalidation logic for cache controller ([line 178](https://github.com/openfga/openfga/blob/40cf00389f9498edd7edacfd2b0602888f55d1a5/internal/cachecontroller/cache_controller.go#L178)).  Right now, we determine the changelog index by comparing the change against the cache controller TTL
```
if !lastInvalidationOccurred || changes[idx].GetTimestamp().AsTime().After(timestampOfLastInvalidation) {
```

We have recently added !lastInvalidationOccured in the [PR to fix cache not invalidating](https://github.com/openfga/openfga/pull/2190).  This has "fixed" it by always forcing the whole cache to be invalidated when the cache controller entry > cache controller TTL.
However, after looking at the code more, I think we should compare it against the iterator cache TTL instead.  The reason is that our partial invalidation entry has the TTL of iterator cache TTL (not the cache controller TTL).  In this way, we can have something like

```
    cacheIteratorLastInvalidation = time.Now().Add(-c.iteratorCacheTTL)
	for ; idx >= 0; idx-- {
		if changes[idx].GetTimestamp().AsTime().After(cacheIteratorLastInvalidation) {
			break
		}
	}
```
Having this change will allow us to be more efficient by only invalidating the entire store when there are recent writes, not when the cache controller TTL expired.

## References
Continuation of https://github.com/openfga/openfga/pull/2190

## Review Checklist


elbuo8 (approved):

internal/cachecontroller/cache_controller.go
```diff
@@ -171,19 +173,28 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 		return
 	}
 
+	timestampOfLastIteratorInvalidation := time.Now().Add(-c.iteratorCacheTTL)
```
adriantam, line 176: Discussed offline - we also want to use timestampOfLastIteratorInvalidation to check for `entry.LastModified.Before` above.  This will provide higher consistency

internal/cachecontroller/cache_controller.go
```diff
@@ -182,25 +184,35 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 	// lookups have the entry and not have to wait on the existing singleflight group
 	c.cache.Set(cacheKey, entry, c.ttl)
 
-	timestampOfLastInvalidation := time.Now().Add(-c.ttl)
+	timestampOfLastIteratorInvalidation := time.Now().Add(-c.iteratorCacheTTL)
```
adriantam, line 187: actually, we should use something like

```
timestampOfLastInvalidation := time.Now().Add(-c.ttl)
if c.cache.Get(changelogCacheKey) != nil {
		timestampOfLastInvalidation = c.cache.Get(changelogCacheKey).(*storage.ChangelogCacheEntry).LastModified
	}
```

internal/cachecontroller/cache_controller.go
```diff
@@ -164,6 +164,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 	// TODO: this should have a deadline since it will hold up everything if it doesn't return
 	// could also be implemented as a fire and forget mechanism and subsequent requests can grab the result
 	// re-evaluate at a later time.
+	// Note that changes are sorted in descending order.
```
justincoh, line 167:
```suggestion
	// Note that changes are sorted most-recent first
```

internal/cachecontroller/cache_controller.go
```diff
@@ -172,19 +173,29 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
...
-	timestampOfLastInvalidation := time.Now().Add(-c.ttl)
+	timestampOfLastInvalidation := time.Time{}
+	if lastCacheRecord != nil {
+		decodedRecord, ok := lastCacheRecord.(*storage.ChangelogCacheEntry)
+		if ok {
+			// if the change log cache is available and valid, use the last modified
+			// time to have better consistency. Otherwise, the timestampOfLastInvalidation will
+			// be beginning of time which imply the need to perform read changes and invalidate record
+			timestampOfLastInvalidation = decodedRecord.LastModified
+		}
```
justincoh, line 195: is there ever a case where `lastCacheRecord != nil` and this is `!ok`?

internal/cachecontroller/cache_controller.go
```diff
@@ -172,19 +173,29 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
...
 	c.cache.Set(cacheKey, entry, c.ttl)
 
-	timestampOfLastInvalidation := time.Now().Add(-c.ttl)
+	timestampOfLastInvalidation := time.Time{}
+	if lastCacheRecord != nil {
+		decodedRecord, ok := lastCacheRecord.(*storage.ChangelogCacheEntry)
+		if ok {
+			// if the change log cache is available and valid, use the last modified
+			// time to have better consistency. Otherwise, the timestampOfLastInvalidation will
+			// be beginning of time which imply the need to perform read changes and invalidate record
```
miparnisari, line 193: `// be beginning of time which imply the need to perform read changes and invalidate record`

Nitpick: the 1st part of this comment does not make a lot of sense to me. By the time we are in this line, the ReadChanges call has already happened, irrespective of whether there was an entry in the cache

Maybe say `// be beginning of time which imply the need to invalidate one or more records`

miparnisari (dismissed):

internal/cachecontroller/cache_controller.go
miparnisari: I don't think so, but maybe there should be an `else { logger.Debug("error casting...") }`?

justincoh (approved):

PR #2220, opened 2025-01-16
test: use storage wrapper in Check unit tests

miparnisari:
## Description
This PR is meant to have unit tests of Check API mimic production code as much as possible. 

In production code, we are calling `NewRequestStorageWrapperForCheckAPI` on every Check API call: https://github.com/openfga/openfga/blob/f00569d20166aab9810ea1b6566cd0968abb4ecd/pkg/server/commands/check_command.go#L115-L118

therefore, the tests should do the same.

## References
To be able to write a complete integration test in https://github.com/openfga/openfga/pull/2150.

### DIFF
--- a/internal/graph/check_fast_path.go
+++ b/internal/graph/check_fast_path.go
@@ -93,6 +93,9 @@ func (s *iteratorStreams) getActiveStreams(ctx context.Context) ([]*iteratorStre
 	return s.streams, nil
 }
 
+// fastPathDirect assumes that req.Object + req.Relation is a directly assignable relation, e.g. define viewer: [user, user:*].
+// It returns a channel with one element, and then closes the channel.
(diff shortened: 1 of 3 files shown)


internal/graph/check_fast_path.go
```diff
@@ -93,6 +93,9 @@ func (s *iteratorStreams) getActiveStreams(ctx context.Context) ([]*iteratorStre
 	return s.streams, nil
 }
 
+// fastPathDirect assumes that req.Object + req.Relation is a directly assignable relation, e.g. define viewer: [user, user:*].
+// It returns a channel with one element, and then closes the channel.
+// The element is an iterator over all objects that are directly related to the user or the wildcard (if applicable).
```
justincoh, line 98: ty for the comments

justincoh (approved):

PR #2221, opened 2025-01-17
test: move tests for Expand

miparnisari:
## Description

Why?

## References

Prior work 

- https://github.com/openfga/openfga/pull/2076
- https://github.com/openfga/openfga/pull/2092
- https://github.com/openfga/openfga/pull/2088
- https://github.com/openfga/openfga/pull/2089
- https://github.com/openfga/openfga/pull/2091
- https://github.com/openfga/openfga/pull/2106
- https://github.com/openfga/openfga/pull/2107


pkg/server/commands/expand.go
```diff
@@ -179,6 +180,8 @@ func (q *ExpandQuery) resolveThis(ctx context.Context, store string, tk *openfga
 		users = append(users, u)
 	}
 
+	slices.Sort(users)
```
justincoh, line 183: Why this change?

justincoh (commented):
I was expecting to see lots of deletions from `/server/test/expand.go` in here, is that coming?

pkg/server/commands/expand.go
miparnisari: To make sure that the resulting array order is deterministic, otherwise I get flaky tests 😄

justincoh (approved):

PR #2222, opened 2025-01-17
chore: discarded cache hit metrics

justincoh:
## Description

This PR adds some basic counter metrics to track cache hits which are discarded as stale.

## Review Checklist


internal/cachecontroller/cache_controller.go
```diff
@@ -169,6 +182,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 	if err != nil {
 		telemetry.TraceError(span, err)
 		// do not allow any cache read until next refresh
+		// TODO: do we want to increment the counter in this scenario?
 		c.invalidateIteratorCache(storeID)
```
justincoh, line 186: Do we want to count invalidations that stem from an issue reading from the datastore?

adriantam (commented):
I would do it in https://github.com/openfga/openfga/blob/6d2f0703ae89caa4c94dd7678df73590edc9bfe8/internal/graph/cached_resolver.go#L165 (the else path there) and https://github.com/openfga/openfga/blob/6d2f0703ae89caa4c94dd7678df73590edc9bfe8/pkg/storage/storagewrappers/check_caching.go#L286 + https://github.com/openfga/openfga/blob/6d2f0703ae89caa4c94dd7678df73590edc9bfe8/pkg/storage/storagewrappers/check_caching.go#L302 instead.

The difference is that if we have the counter in cache_controller.go, the counter will increment even if the cache iterator / cached resolver has no entries (i.e., it will increment based on the cache iterator's view) while having the counter in storage wrapper / cache resolver will only be incremented if the entry is present but discarded due to invalidation.

internal/cachecontroller/cache_controller.go
miparnisari: 🤔 you _are_ counting them, by virtue of incrementing the counter inside `invalidateIteratorCache`.

internal/cachecontroller/cache_controller.go
```diff
@@ -127,6 +139,7 @@ func (c *InMemoryCacheController) DetermineInvalidation(
 	_, present := c.inflightInvalidations.LoadOrStore(storeID, struct{}{})
 	if !present {
 		span.SetAttributes(attribute.Bool("check_invalidation", true))
+		checkQueryCacheInvalidationCounter.Inc()
```
miparnisari, line 142: When placed in this line, this counter is slightly misleading, as you _may_ not end up doing any invalidations at all.

Why not move this line to

```go
for ; idx >= 0; idx-- {
       // here
	t := changes[idx].GetTupleKey()
	c.invalidateIteratorCacheByObjectRelation(storeID, t.GetObject(), t.GetRelation(), lastModified)
	c.invalidateIteratorCacheByObjectTypeRelation(storeID, t.GetUser(), tuple.GetType(t.GetObject()), lastModified)
		}
```

where you know that for sure an invalidation will take place?

miparnisari (commented):
Could you run a quick test? Edit `docker-compose.yaml` to include ` OPENFGA_CACHE_CONTROLLER_ENABLED=true` in the `environment` section, and then run Openfga via `docker-compose up --build`. Then create a store, write some tuples, then start doing 1 Check request every second, and a random write every 10 seconds or so.

You should see metrics in the Prometheus website at http://localhost:9090/.

If you could post a screenshot here on how those metrics look like, that would be 👍

internal/cachecontroller/cache_controller.go
justincoh: Right but should I be? Or should the error scenario not be considered in the metric

miparnisari: I think it's fine 👍

internal/graph/cached_resolver.go
```diff
@@ -38,6 +38,12 @@ var (
 		Name:      "check_cache_hit_count",
 		Help:      "The total number of cache hits for ResolveCheck.",
 	})
+
+	checkCacheInvalidHit = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "check_cache_invalid_hit_count",
+		Help:      "The total number of cache attempts which hit an invalid entry.",
+	})
```
miparnisari, line 46:
```suggestion
	checkCacheInvalidHit = promauto.NewCounter(prometheus.CounterOpts{
		Namespace: build.ProjectName,
		Name:      "check_cache_invalid_hit_count",
		Help:      "The total number of cache hits for ResolveCheck that were discarded because they were invalidated",
	})
```

pkg/storage/storagewrappers/check_caching.go
```diff
@@ -59,6 +59,12 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"operation"})
+
+	tuplesInvalidCacheHit = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "tuples_iterator_cache_invalid_hit_count",
+		Help:      "The total number of times we hit an invalid tuple iterator cache entry.",
+	})
```
miparnisari, line 67:
```suggestion
	tuplesInvalidCacheHit = promauto.NewCounter(prometheus.CounterOpts{
		Namespace: build.ProjectName,
		Name:      "tuples_iterator_cache_invalid_hit_count",
		Help:      "The total number of cache hits for a cached iterator that were discarded because they were invalidated",
	})
```

pkg/storage/storagewrappers/check_caching.go
```diff
@@ -287,11 +293,14 @@ func findInCache(cache storage.InMemoryCache[any], store, key string, invalidEnt
 			if ok {
 				invalidEntryLastModifiedTime = invalidEntry.LastModified
 			}
+
+			tuplesInvalidCacheHit.Inc()
```
justincoh, line 297: I had a difficult time triggering this `tuplesInvalidCacheHit.Inc()` in test (and the one below). If `OPENFGA_CHECK_QUERY_CACHE_ENABLED=true`, I was unable to reach this code; this method would always return on line 285. I haven't yet determined why, but it _seems like_ the caches are coupled in some way (I was using identical TTLs). I'll put the full information I used to test in a top-level comment on the PR description page

justincoh:
@miparnisari I ran a script to test metrics with the following config:

```
      - OPENFGA_CACHE_CONTROLLER_ENABLED=true
      - OPENFGA_CHECK_ITERATOR_CACHE_ENABLED=true
      - OPENFGA_CHECK_ITERATOR_CACHE_TTL=3s
      - OPENFGA_CACHE_CONTROLLER_TTL=3s

		# this was necessary to get tuplesInvalidCacheHit.Inc() to trigger
		# i have not yet determined why
      - OPENFGA_CHECK_QUERY_CACHE_ENABLED=false
```

The script ran the following:
- create a store and model with a userset relation
- write some tuples
- every 1s, call `check()`
- every 5s, write a new tuple

It ended up with ~2k iterator cache hits and ~60 iterator cache hits which were discarded over ~12 minutes:

Hits:
Screenshot 2025-01-22 at 10 44 08 AM

Hits which were discarded:
Screenshot 2025-01-22 at 10 44 18 AM

internal/cachecontroller/cache_controller.go
```diff
@@ -50,14 +50,14 @@ var (
 )
 
 type CacheController interface {
-	// DetermineInvalidation returns the timestamp of the last write for the specified store.
+	// DetermineInvalidationTime returns the timestamp of the last write for the specified store.
 	// It may return a cached timestamp.
-	DetermineInvalidation(ctx context.Context, storeID string) time.Time
+	DetermineInvalidationTime(ctx context.Context, storeID string) time.Time
```
justincoh, line 55: Unrelated to this PR, updated this method name for clarity

pkg/storage/storagewrappers/check_caching.go
```diff
@@ -59,6 +59,12 @@ var (
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"operation"})
+
+	tuplesInvalidCacheHit = promauto.NewCounter(prometheus.CounterOpts{
```
adriantam, line 63: Since tuplesCacheHitCounter and tuplesCacheDiscardCounter has `operation` as vector, I am wondering if we also want to do that as well.

justincoh: Yeah it seems potentially useful to tag based on `ReadStartingWithUser` vs `ReadUsersetTuples`, will add

justincoh: updated

adriantam (approved):

PR #2226, opened 2025-01-21
chore: move tuple iterators and mappers inside pkg/storage

miparnisari:
## Description
Address this https://github.com/openfga/openfga/blob/135aef4177d26d22be4cc73a7f469de27b82f0b8/pkg/storage/iterators.go#L423

by organizing the code better.

### Idea

Should we change `storage.ReadUsersetTuples` interface, that currently returns `TupleIterator`, to return a `UsersetMapper`?  

Similarly, we could change `storage.ReadStartingWithUser` interface, that currently returns `TupleIterator`, to return a `ObjectIDMapper`.

The changes would:
- make it super clear what this API must return
- reduce memory usage and data transform (assuming we update the SELECTs in each database)


pkg/storage/tuple_iterators.go
miparnisari, line 1: File renamed to `pkg/storage/tuple_iterators.go`

pkg/storage/tuple_mappers.go
miparnisari, line 1: Note that this code is now public

elbuo8 (approved):

PR #2237, opened 2025-01-23
chore: introduce new IObjectMapper interface and use it within Check API

miparnisari:
## Description

We have various places in the code, especially in Check API, where we have the need to read tuples from the DB, but we only care about the `object` component of the tuple.

Therefore, I introduce a new interface `IObjectMapper` (could also be called `IObjectIterator`..) that is an iterator that returns the `object` field of a tuple.

This change should

- not change API behavior
- reduce memory consumption, by keeping less "useless" information around
- increase code clarity: we are making more explicit what a given iterator returns, and what a consumer of the iterator needs.
- the bullet above also reduces the chance of bugs.

I also introduce `storage.NewStaticObjectMapper` that implements `IObjectMapper`.


pkg/storage/tuple_mappers.go
```diff
@@ -40,6 +42,62 @@ type TupleMapper interface {
 	Iterator[string]
 }
 
+// IObjectMapper is an iterator that, on calls to Next and Head, returns an object (type + id).
+type IObjectMapper interface {
+	TupleMapper
+}
+
+var _ IObjectMapper = (*StaticObjectMapper)(nil)
```
justincoh, line 50: What's the value add of the `IObjectMapper` as a wrapper around the existing `TupleMapper` interface?

miparnisari: It's a benefit for the calling code, as it knows exactly what it will get by calling `iter.Next()` (an object as opposed to a full tuple)

PR #2238, opened 2025-01-24
perf: remove fmt.sprintf and just add strings

justincoh:
## Description

Came across a `fmt.sprintf` in the `tuple` package which we can make faster while maintaining readability by just concatenating strings.

I did a quick benchmark comparison:
```
func BenchmarkTupleStringifiers(b *testing.B) {
	tupleKey := &openfgav1.TupleKey{
		User:     "user:justin",
		Relation: "viewer",
		Object:   "doc:1",
	}

	b.Run("new_version", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			_ = TupleKeyToString(tupleKey)
		}
	})
}
```

Old:
image

New
image

Percentage-wise it's quite a bit faster, but we're talking nanoseconds here so 🤷 

We can actually make it significantly faster still (and remove all memory usage) if we specify an explicit type for this function to accept, rather than an interface, but that requires a refactor in other places and doesn't seem worth it for such small gains.

With explicit type:
image

## Review Checklist

### DIFF
--- a/pkg/tuple/tuple.go
+++ b/pkg/tuple/tuple.go
@@ -349,7 +349,11 @@ func GetUserTypeFromUser(user string) UserType {
 // TupleKeyToString converts a tuple key into its string representation. It assumes the tupleKey is valid
 // (i.e. no forbidden characters).
 func TupleKeyToString(tk TupleWithoutCondition) string {
-	return fmt.Sprintf("%s#%s@%s", tk.GetObject(), tk.GetRelation(), tk.GetUser())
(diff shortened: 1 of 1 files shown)


miparnisari (approved):
If you want to go crazy, you can run this linter 😄 https://github.com/catenacyber/perfsprint

```
# ignore all test files
# ignore calls to `fmt.Errorf`
 perfsprint --errorf=false -test=false ./...
```

PR #2239, opened 2025-01-24
refactor: move some functionality into iterator stream

adriantam:
## Description
Clean up the code by moving some functionality from check fastpath into iterator stream.  At the same time, increase test coverage for check fastpath.

 
## Review Checklist


internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+type iteratorStream struct {
+	idx    int
+	buffer storage.TupleKeyIterator
+	done   bool
+	source chan *iteratorMsg
+}
+
+// helper function to drain the stream until the head's object is larger than target.
+// If the buffer is drained and no more items, it will set to stop and buffer will be nil.
+func (s *iteratorStream) skipToTarget(ctx context.Context, target string) error {
```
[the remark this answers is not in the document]
miparnisari: also: 

- should we add an assertion that `target` must be of the form `objecttype:id`?
- should we assert that `s.buffer != nil`?

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+			return err
+		}
+		tmpKey = t.GetObject()
+	}
+	return nil
+}
+
+// drain all item in the stream's buffer and return these items.
+func (s *iteratorStream) drain(ctx context.Context) ([]*openfgav1.TupleKey, error) {
```
miparnisari, line 49: should we assert that `s.buffer != nil`?

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+	"context"
+	"slices"
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+type iteratorStream struct {
+	idx    int
```
miparnisari, line 13: it's a code smell to me that `idx` is never referenced in this file, it's referenced outside of it 🤔  it feels like we should add a method to `iteratorStream` to get or set this `idx`.

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+			return nil, err
+		}
+		batch = append(batch, t)
+	}
+	return batch, nil
+}
+
+// for all slices with index in itersWithEqualObject, call next and return the first item
+// return error if any individual stream in the streamSlices is empty.
```
miparnisari, line 67: This godoc makes little sense to me

- `for all slices with index`? what index?
- what is `itersWithEqualObject` (looks like it's `streamToProcess`)
- `call next and return the first item` but you are overwriting the `item` variable on every loop, so presumably it actually returns the _last_ item?

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+func nextItemInSliceStreams(ctx context.Context, streamSlices []*iteratorStream, streamToProcess []int) (*openfgav1.TupleKey, error) {
+	var item *openfgav1.TupleKey
+	var err error
+	for _, iterIdx := range streamToProcess {
+		item, err = streamSlices[iterIdx].buffer.Next(ctx)
```
miparnisari, line 72: this can panic: `streamSlices[iterIdx]`, do we want to protect against this?

miparnisari (commented):
Haven't finished reviewing the whole PR but giving early feedback 😄

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+		if stream.buffer != nil {
+			stream.buffer.Stop()
+		}
+		for msg := range stream.source {
+			if msg.iter != nil {
+				msg.iter.Stop()
+			}
+		}
+	}
+}
```
miparnisari, line 102:
```
func (s *iteratorStream) Stop() {
	if s.buffer != nil {
		s.buffer.Stop()
	}
	for msg := range s.source {
		if msg.iter != nil {
			msg.iter.Stop()
		}
	}
}

func (s *iteratorStreams) Stop() {
	for _, stream := range s.streams {
		stream.Stop()
	}
}
```

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+import (
+	"context"
+	"slices"
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+type iteratorStream struct {
```
miparnisari, line 12: Could we add more methods to this struct so that callers _do not_ have to rely on internal fields? For example:
```
func (s *iteratorStream) Head(ctx context.Context) (*openfgav1.TupleKey, error) {
	return s.buffer.Head(ctx)
}
```

Then, in `check_fast_path.go`, we could replace line 199

```go
v, err := stream.buffer.Head(ctx)
```

with 

```go
v, err := stream.Head(ctx)
```

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+		for msg := range stream.source {
+			if msg.iter != nil {
+				msg.iter.Stop()
+			}
+		}
+	}
+}
+
+// getActiveStreams will return a list of the remaining active streams.
```
miparnisari, line 104:
```suggestion
// getActiveStreams will mutate the iteratorStreams object [line 129] and return a list of the remaining active streams.
```

which makes me wonder - should we change the signature to the below?


```
func (s *iteratorStreams) cleanDone(ctx context.Context) (error) {
```

internal/graph/iterator_stream.go
justincoh: or rely on the `s.Next()` method suggested above

internal/graph/iterator_stream.go
```diff
@@ -0,0 +1,133 @@
...
+			if i.err != nil {
+				return nil, i.err
+			}
+			stream.buffer = i.iter
+		}
+	}
+	// TODO: in go1.23 compare performance vs slices.Collect
+	// clean up all empty entries that are both done and drained
+	s.streams = slices.DeleteFunc(s.streams, func(entry *iteratorStream) bool {
+		return entry.done && entry.buffer == nil
```
justincoh, line 130: I'm a bit confused by the combination of `done` and `buffer == nil`. It seems like they're both means to communicate `there's nothing left to do here`.

When would a stream be `done` and have a non-nil `buffer? Or vice versa: when would we have a `nil` buffer with a stream that is `!done`?

adriantam: It is meant that when buffer is nil, we will try to supply it with source.  Looking into it more, we actually want to move some of these logics in the stream itself.

internal/graph/iterator_stream.go
adriantam: updated signature

internal/graph/iterator_stream.go
miparnisari: This code is also confusing to me. I think a better name for the `done` field is `sourceIsClosed`.

internal/graph/iterator/iterator_stream.go
```diff
@@ -0,0 +1,203 @@
...
+		return ctx.Err()
+	case i, ok := <-s.source:
+		if !ok {
+			s.done = true
+			break
+		}
+		if i.Err != nil {
+			return i.Err
+		}
+		s.buffer = i.Iter
```
miparnisari, line 142: Could it happen that before line 142, `s.buffer` was already not nil?  If not, it seems like there is an unwritten pre-condition that this function has: that the consumer of the stream must have already called Head or Next until `storage.ErrIteratorDone`.

miparnisari (approved):
Nice refactor! 🚀

internal/graph/iterator/iterator_stream_test.go
```diff
@@ -0,0 +1,528 @@
...
+			close(c)
+			streams[i] = &Stream{
+				buffer: buf,
+				source: c,
+			}
+		}
+		dut := NewStreams(streams)
+		dut.Stop()
+		for _, stream := range streams {
+			_, err := stream.buffer.Head(context.Background())
```
miparnisari, line 524:
```suggestion
			_, err := stream.Head(context.Background())
```

internal/graph/iterator/iterator_stream.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/tuple"
+)
+
+type Msg struct {
+	Iter storage.TupleKeyIterator
+	Err  error
+}
+
+type Stream struct {
```
miparnisari, line 19:
```suggestion
// Stream aggregates multiple iterators that are sent to a source channel
type Stream struct {
```

internal/graph/check_fast_path.go
```diff
@@ -24,80 +24,13 @@ const IteratorMinBatchThreshold = 1000
...
-		return entry.done && entry.buffer == nil
-	})
-	return s.streams, nil
-}
+type fastPathSetHandler func(context.Context, *iterator.Streams, chan<- *iterator.Msg)
 
 // fastPathDirect assumes that req.Object + req.Relation is a directly assignable relation, e.g. define viewer: [user, user:*].
 // It returns a channel with one element, and then closes the channel.
 // The element is an iterator over all objects that are directly related to the user or the wildcard (if applicable).
 func fastPathDirect(ctx context.Context,
```
miparnisari, line 32: Nit: could you put the signature of `fastPathDirect` in one line? Makes it more readable.

internal/graph/iterator/iterator_stream_test.go
miparnisari, line 1: I think a better location for these files is `internal/concurrency`, for 2 reasons:

1. there is nothing in these files to suggest that they cannot be used outside of Check API
2. the existing function `fanInIteratorChannels` relies on `iterator.Msg`, and we left a note saying that a better place for this would be `internal/concurrency`. This further suggests that these two pieces belong together. (You don't have to move `fanInIteratorChannels` in this PR)

internal/graph/iterator/iterator_stream.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	t, err := s.Head(ctx)
+	if err != nil {
+		if storage.IterIsDoneOrCancelled(err) {
+			return nil
+		}
+		return err
+	}
+	tmpKey := t.GetObject()
```
miparnisari, line 94: You might be interested in my proposal: https://github.com/openfga/openfga/pull/2237 😄

internal/graph/check_fast_path.go
```diff
@@ -641,10 +517,10 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
 func constructLeftChannels(ctx context.Context,
 	req *ResolveCheckRequest,
 	relationReferences []*openfgav1.RelationReference,
-	relationFunc checkutil.V2RelationFunc) ([]chan *iteratorMsg, error) {
+	relationFunc checkutil.V2RelationFunc) ([]chan *iterator.Msg, error) {
 	typesys, _ := typesystem.TypesystemFromContext(ctx)
 
-	leftChans := make([]chan *iteratorMsg, 0, len(relationReferences))
+	leftChans := make([]chan *iterator.Msg, 0, len(relationReferences))
```
miparnisari, line 523: Not for this PR: i think `leftChans` is basically `Streams` 🤔

internal/graph/check_fast_path.go
```diff
@@ -641,10 +517,10 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
 func constructLeftChannels(ctx context.Context,
 	req *ResolveCheckRequest,
 	relationReferences []*openfgav1.RelationReference,
-	relationFunc checkutil.V2RelationFunc) ([]chan *iteratorMsg, error) {
+	relationFunc checkutil.V2RelationFunc) ([]chan *iterator.Msg, error) {
```
miparnisari, line 520: **Not for this PR** i think we can utilize `Stream` and `Streams` even more. This functions is basically returning `Streams`. `fastPathDirect` and `fastPathComputed`  and `fastPathRewrite` could return `Stream`, `fastPathOperationSetup` could use `Streams` internally, etc...

justincoh (dismissed):
Code looks good, ty for the refactor.

Summarizing to ensure I fully understand:

* `Stream` type exists to aggregate multiple iterators.
* `Stream.source` is a channel containing some number of iterators
* `Stream.buffer` is the processing area for a given iterator, which originated in a message in `Stream.source`
* `Streams.CleanDone` is responsible for determining when buffers are exhausted and loading `buffer` with the next iterator waiting the `Stream.source`

adriantam:
> Code looks good, ty for the refactor.
> 
> Summarizing to ensure I fully understand:
> 
> * `Stream` type exists to aggregate multiple iterators.
> * `Stream.source` is a channel containing some number of iterators
> * `Stream.buffer` is the processing area for a given iterator, which originated in a message in `Stream.source`
> * `Streams.CleanDone` is responsible for determining when buffers are exhausted and loading `buffer` with the next iterator waiting the `Stream.source`

Correct.

internal/graph/iterator/iterator_stream_test.go
adriantam: Maybe better location is `internal/iterator` instead of `internal/concurrency` because the iterator does not have to be done in a concurrent manner.

internal/graph/check_fast_path.go
adriantam: Agree - I think having iterator as separate struct allows better abstraction and we can refactor more in separate PR

justincoh (approved):

PR #2240, opened 2025-01-24
fix: drain iterators in Check

miparnisari:
## Description

I wrote a unit test that, without the code change, randomly fails with "iterator.Stop not called". This can mean a leak in the tests.

An alternative fix is in https://github.com/openfga/openfga/pull/2247

## References

https://github.com/openfga/openfga/actions/runs/12935953707/job/36080447940

```
--- FAIL: TestCheckMySQL (25.68s)
    tests.go:63: creating connection to address localhost:45819
    testutils.go:234: 2025-01-23 18:50:21.15580025 +0000 UTC m=+68.914354291 not serving yet at address localhost:45819 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:45819: connect: connection refused"
    testutils.go:234: 2025-01-23 18:50:21.499385381 +0000 UTC m=+69.257939412 not serving yet at address localhost:45819 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:45819: connect: connection refused"
    tests.go:57: waiting for server to stop
    tests.go:60: server stopped with error:  <nil>
    mysql.go:106: stopping container mysql-01JJA804SJPYF6NMQE5FVWNF2H
    mysql.go:113: stopped container mysql-01JJA804SJPYF6NMQE5FVWNF2H
    check_test.go:329: found unexpected goroutines:
        [Goroutine 108867 in state select, with github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1 on top of the stack:
        github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1()
        	/home/runner/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:628 +0x105
        created by github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher in goroutine 108744
        	/home/runner/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:625 +0x1dd
```


internal/graph/check_fast_path.go
```diff
@@ -557,16 +557,11 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
...
 		iter.Stop()
-		if !leftOpen {
-			return
-		}
-		go func() {
-			for msg := range leftChan {
-				if msg.iter != nil {
-					msg.iter.Stop()
-				}
+		for msg := range leftChan {
```
adriantam, line 560: Is the bug due to invalid check !leftOpen or the check for leftChan is inside go func() ?

elbuo8: Im a little confused. How is this a leak. 

And the whole purpose was to not block yielding the result because this will affect your latency on short circuited resolutions.

miparnisari: It's a leak in the tests.

> And the whole purpose was to not block yielding the result because this will affect your latency on short circuited resolutions.

I understand the goal, but then we need to figure out a way of achieving it without a goroutine that nobody is waiting on.

Do we have evidence that latency is majorly affected by this? Also: doesn't the existing context cancellation code prevent the remaining iterators from doing a lot of unnecessary work once the signal has been sent?

elbuo8: So a "leak" in the tests, which is not a real leak in actual runtime because it actually gets reaped, just the tooling doing a false positive?

> that latency is majorly affected by this?

If the context cancellation happens right after the request has been fired of to the DS you have to wait for the I/O (depending on the driver, which as you know there are drivers that don't handle well the context cancellations) so that I/O could be from microseconds to 10s of milliseconds depending on the conditions.

PR #2247, opened 2025-01-28
fix: on server stop, wait for iterators to stop

miparnisari:
## Description
This is an alternative to https://github.com/openfga/openfga/pull/2240 to eliminate a goroutine leak.

Pros:
- no more leaks
- no impact to performance of Check API

Cons:
- very low chance of a data race. `c.goroutineWaiter.Wait()` is called in the main goroutine when the server stops, but `c.goroutineWaiter.Add(1)` runs inside a goroutine. It could happen that `Add` is called after `Wait`. Behavior is undefined, I think it's harmless.

## References

https://github.com/openfga/openfga/actions/runs/12935953707/job/36080447940

```
--- FAIL: TestCheckMySQL (25.68s)
    tests.go:63: creating connection to address localhost:45819
    testutils.go:234: 2025-01-23 18:50:21.15580025 +0000 UTC m=+68.914354291 not serving yet at address localhost:45819 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:45819: connect: connection refused"
    testutils.go:234: 2025-01-23 18:50:21.499385381 +0000 UTC m=+69.257939412 not serving yet at address localhost:45819 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:45819: connect: connection refused"
    tests.go:57: waiting for server to stop
    tests.go:60: server stopped with error:  <nil>
    mysql.go:106: stopping container mysql-01JJA804SJPYF6NMQE5FVWNF2H
    mysql.go:113: stopped container mysql-01JJA804SJPYF6NMQE5FVWNF2H
    check_test.go:329: found unexpected goroutines:
        [Goroutine 108867 in state select, with github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1 on top of the stack:
        github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher.func1()
        	/home/runner/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:628 +0x105
        created by github.com/go-sql-driver/mysql.(*mysqlConn).startWatcher in goroutine 108744
        	/home/runner/go/pkg/mod/github.com/go-sql-driver/mysql@v1.8.1/connection.go:625 +0x1dd
```


internal/graph/check_fast_path.go
```diff
@@ -598,6 +600,9 @@ func (c *LocalChecker) checkTTUFastPathV2(ctx context.Context, req *ResolveCheck
 	return c.resolveFastPath(ctx, leftChans, storage.WrapIterator(storage.TTUKind, iter))
 }
 
+// fanInIteratorChannels aggregates the messages coming from the input channels into a single output channel.
+// It expects the input channels to be closed.
```
miparnisari, line 604: I have not spent time investigating whether this is the case. In the test, if you don't close the input channel, the test stalls.

internal/graph/check_fast_path.go
```diff
@@ -443,7 +443,9 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
 		if !leftOpen {
 			return
 		}
+		c.goroutineWaiter.Add(1)
 		go func() {
+			defer c.goroutineWaiter.Done()
 			for msg := range leftChan {
```
miparnisari, line 449: The unit test I wrote asserts that this channel is closed. Otherwise, the `defer` will never run.

justincoh (commented):
I don't love adding a wg to this type _just_ so it fixes issues in test, but this seems fairly harmless. My only other concern would be the same one you mentioned:

> It could happen that Add is called after Wait

This means the server could just hang when we try to close it in test.

Could that also cause issues in tearing down pods during deployments?

internal/graph/check_fast_path.go
miparnisari: Sorry, i meant that i didn't investigate whether every place where this function is called is closing the input channels.

adriantam:
> Could that also cause issues in tearing down pods during deployments?

Possibly, however, since the server is shutting down / pods is tearing down, leak is not a significant issue since memory will be returned when the server is terminated.

internal/graph/check_fast_path.go
adriantam: Updated as suggested

PR #2241, opened 2025-01-27
perf: Replace `fmt.Sprintf` with efficient string operations

Siddhant-K-code:
## Description

This PR improves performance by optimizing string operations across the codebase.

Changes:
- Replace `fmt.Sprintf` with direct string concatenation using + operator
- Use `strconv.FormatBool` for boolean to string conversion
- Optimize `strings.Builder` usage in tuple operations
- Clean up imports in affected files

Performance impact:

These changes reduce allocations and CPU overhead by using more efficient string operations instead of fmt.Sprintf formatting.

## References

address [this comment](https://github.com/openfga/openfga/pull/2238#pullrequestreview-2573629219) 😄 

## Review Checklist
- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [x] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #2248, opened 2025-01-28
chore: add perfsprint linter

miparnisari:
## Description
In https://github.com/openfga/openfga/pull/2241 I forgot to comment that `perfsprint` needs to be added to `.golangci.yaml` 😅 

## References
Part of https://github.com/openfga/openfga/issues/598

### DIFF
--- a/pkg/server/test/list_objects.go
+++ b/pkg/server/test/list_objects.go
@@ -2,7 +2,6 @@ package test
 
 import (
 	"context"
-	"fmt"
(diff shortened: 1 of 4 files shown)


pkg/server/test/list_objects.go
miparnisari, line 1: Sadly, these aren't being considered tests (because they don't end with _test) so the linter complains for these

justincoh (approved):

PR #2252, opened 2025-01-30
fix: Set database dialect during initialization to prevent SQL syntax errors

### LINKED ISSUE #2251: SQL syntax error when calling IsReady() method with MySQL database
### Checklist

### Description

Hi! We use OpenFGA with MySQL database and encountered an error when trying to use `IsReady()` method (we use it for implementing health checks). When method is called, it returns error:
```
method=/grpc.health.v1.Health/Check
err="Error 1064 (42000): You have an error in your SQL syntax; 
check the manual that corresponds to your MySQL server version for the right syntax to use near 
'GENERATED BY DEFAULT AS IDENTITY,\n\t\tversion_id bigint NOT NULL,\n\t\tis_applied boo' at line 2"
```

I tried to debug and see what's happening and the problem is that `IsReady()` method implemented by using 3rd party library [goose](https://github.com/pressly/goose), which supports multiple storage types, but by default it's set to Postgres. So it tries to create a table using Postgres syntax which is incorrect for MySQL. I guess, storage type should be properly initialized before first usage, like this:
```go
goose.SetDialect(string(goose.DialectMySQL)) // or whatever DB is used
```

Error is happening [here](https://github.com/openfga/openfga/blob/94dd78f03dc99e51f46e7f71deaaea878e8340c9/pkg/storage/sqlcommon/sqlcommon.go#L701).

### OpenFGA version

1.8.4

As a library

MySQL

### OpenFGA Flags

no

### Logs

_No response_

Siddhant-K-code:
## Description

Optimizes database dialect handling by setting it once during initialization.

Changes:
- Added dialect parameter to `sqlcommon.NewDBInfo`
- Moved `goose.SetDialect` call to initialization time
- Added dbInfo field to SQLite Datastore struct
- Updated MySQL, Postgres and SQLite implementations to pass correct dialect during initialization
- Updated test files to include dialect parameter in NewDBInfo calls
- Optimized error messages using string concatenation

## References

Fixes #2251

## Review Checklist
- [x] I have updated tests to validate that the change in functionality is working as expected

### DIFF
--- a/pkg/storage/sqlcommon/sqlcommon.go
+++ b/pkg/storage/sqlcommon/sqlcommon.go
@@ -690,12 +690,16 @@ func ReadAuthorizationModel(
 
 // IsReady returns true if the connection to the datastore is successful
 // and the datastore has the latest migration applied.
-func IsReady(ctx context.Context, db *sql.DB) (storage.ReadinessStatus, error) {
+func IsReady(ctx context.Context, db *sql.DB, dialect string) (storage.ReadinessStatus, error) {
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
 	defer cancel()
 
 	if err := db.PingContext(ctx); err != nil {
 		return storage.ReadinessStatus{}, err
+	}
+
+	if err := goose.SetDialect(dialect); err != nil {
+		return storage.ReadinessStatus{}, fmt.Errorf("failed to set database dialect: %w", err)
 	}
 
 	revision, err := goose.GetDBVersion(db)
--- a/pkg/storage/sqlcommon/sqlcommon.go
+++ b/pkg/storage/sqlcommon/sqlcommon.go
@@ -716,7 +716,6 @@ func IsReady(ctx context.Context, db *sql.DB) (storage.ReadinessStatus, error) {
 				"'. Run 'openfga migrate'.",
 			IsReady: false,
 		}, nil
-
 	}
 	return storage.ReadinessStatus{
 		IsReady: true,
--- a/pkg/storage/sqlcommon/sqlcommon.go
+++ b/pkg/storage/sqlcommon/sqlcommon.go
@@ -395,7 +395,6 @@ type DBInfo struct {
 	db             *sql.DB
 	stbl           sq.StatementBuilderType
 	HandleSQLError errorHandlerFn
-	dialect        string
 }
 
 type errorHandlerFn func(error, ...interface{}) error
@@ -719,7 +718,6 @@ func IsReady(ctx context.Context, db *sql.DB) (storage.ReadinessStatus, error) {
 		}, nil
 
 	}
-
 	return storage.ReadinessStatus{
 		IsReady: true,
 	}, nil
--- a/pkg/storage/sqlcommon/sqlcommon.go
+++ b/pkg/storage/sqlcommon/sqlcommon.go
@@ -5,7 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
-	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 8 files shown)


Siddhant-K-code:
flaky tests are back 😭 

image
image

loeffel-io:
Would be great to merge and release this asap - this is currently a blocker for us ❤️

DanCech (approved):

alexanderzobnin (approved):

jpadilla:
Thanks for looking into this! Is there a test we can add to confirm how the issue manifests and proof this change fixes it?

alexanderzobnin:
If you can add integration test for MySQL which calls IsReady(), it should reveal the issue I think.

miparnisari:
We do have a test for `IsReady`, and it runs for all supported datastores: https://github.com/openfga/openfga/blob/c9412f77071bca0b4b628c79884f6bf89263e142/pkg/storage/test/storage.go#L29-L34

Why did it never fail? 🤔

aaguiarz:
And why did it work on v1.7? https://github.com/openfga/openfga/issues/2251#issuecomment-2624614328

pkg/storage/sqlite/sqlite.go
```diff
@@ -1033,7 +1033,7 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
 
 // IsReady see [sqlcommon.IsReady].
 func (s *Datastore) IsReady(ctx context.Context) (storage.ReadinessStatus, error) {
-	return sqlcommon.IsReady(ctx, s.db)
+	return sqlcommon.IsReady(ctx, s.db, "sqlite")
```
aaguiarz, line 1036: Is it "sqlite" or "sqlite3"? https://github.com/pressly/goose/blob/9c748e24f4b2cb09afa0dd252ed3548c9f17d51b/internal/dialect/dialects.go#L4

miparnisari:
The root cause of why we didn't catch this before is that our test code (https://github.com/openfga/openfga/blob/2b0f0836b14bc1117a0c2c19d6a8e988296d6daa/pkg/testfixtures/storage/mysql.go#L127-L162) is doing a lot of the legwork required (in this case, setting the dialect... but also running all the migrations) which the production code (`NewWithDB` function) isn't doing. We should probably rearchitect to either 

- have the `NewWithDB` function run the migrations (not sure of the implications of this), or
- expose a new function to do that (and then call it manually in each test), or 
- some other option?

pkg/storage/sqlite/sqlite.go
Siddhant-K-code: In our tests, we were using sqlite, that’s why i decided to keepthis way -- But, I can change that if req.

miparnisari: Both are OK. https://github.com/pressly/goose/blob/v3.24.1/dialect.go#L40

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -690,14 +690,18 @@ func ReadAuthorizationModel(
...
-func IsReady(ctx context.Context, db *sql.DB) (storage.ReadinessStatus, error) {
+func IsReady(ctx context.Context, db *sql.DB, dialect string) (storage.ReadinessStatus, error) {
 	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
 	defer cancel()
 
 	if err := db.PingContext(ctx); err != nil {
 		return storage.ReadinessStatus{}, err
 	}
 
+	if err := goose.SetDialect(dialect); err != nil {
```
miparnisari, line 701: Instead of doing it here, which will run on every call, can we do it in `sqlcommon.NewDBInfo`?

Siddhant-K-code: done in [`0af228a`](https://github.com/openfga/openfga/pull/2252/commits/0af228af1e607fa3f62b3f7acc20005e27d02d04)

Siddhant-K-code:
Locally tests are working fine 

image

miparnisari (approved):

Siddhant-K-code:
What's blocking for merge & release?

Siddhant-K-code:
I think we should merge https://github.com/openfga/openfga/pull/2259 first and update this branch with those changes.

PR #2259, opened 2025-02-03
test: fix goroutine leak in check tests

justincoh:
## Description

The `check_test` suite _sometimes_ has a goroutine leak on cleanup. The exact cause is still under investigation, but it is within [this code path](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/context.go#L25) and related to how our iterator cache interacts with the DB during shutdown; that behavior is modified by the environment variable in this PR. I have a messy branch which reproduced the leak reliably [here](https://github.com/openfga/openfga/compare/main...data-race-debugging), but after making the change in this PR I have been unable to reproduce the leak again.

## Review Checklist

### DIFF
--- a/tests/check/check_test.go
+++ b/tests/check/check_test.go
@@ -94,6 +94,7 @@ func TestServerLogs(t *testing.T) {
 	cfg.Trace.Enabled = true
 	cfg.Trace.OTLP.Endpoint = localOTLPServerURL
 	cfg.Datastore.Engine = "memory"
+	cfg.ContextPropagationToDatastore = true
 
 	observerLogger, logs := observer.New(zap.DebugLevel)
 	serverCtx := &run.ServerContext{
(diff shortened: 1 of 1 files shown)


tests/check/check_test.go
```diff
@@ -94,6 +94,7 @@ func TestServerLogs(t *testing.T) {
 	cfg.Trace.Enabled = true
 	cfg.Trace.OTLP.Endpoint = localOTLPServerURL
 	cfg.Datastore.Engine = "memory"
+	cfg.ContextPropagationToDatastore = true
```
miparnisari, line 97: - `ContextPropagationToDatastore` shouldn't have any noticeable impact on the `memory` datastore (we probably should document this) which is what this test uses (line 96). It should only affect Postgres and MySQL.
- Should we set this to `true` also in `TestCheckMySQL` and `TestCheckPostgres`?

justincoh: ah yes I pasted this into the wrong place, updating

jpadilla (approved):

PR #2263, opened 2025-02-05
perf: update order of resolvers for usersets

elbuo8:
## Description

Fix order of resolvers for usersets based on optimizations and cost. 
Fix the contract of fast path v2 optimizations to match regular resolver behavior.

## Review Checklist


internal/graph/check_fast_path.go
```diff
@@ -494,13 +502,15 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
...
-					return nil, err
+					lastErr = err
+					continue
 				}
 				if processUsersetMessage(t.GetObject(), leftSet, rightSet) {
 					msg.Iter.Stop()
 					res.Allowed = true
 					span.SetAttributes(attribute.Bool("allowed", true))
-					return res, ctx.Err()
+					lastErr = nil
```
yissellokta, line 512: why we are not returning the error, and returning nil

senojj: it appears that this is a successful path, regardless of the existence of an error on the context. I assume that we don't want to indicate an error if the error is meaningless at this point.

elbuo8: @senojj correct. To match the existing contract of swallowing errors of `allowed=true` this would be thrown away.

yissellokta (approved):

PR #2275, opened 2025-02-13
chore: handle dangling routines in check

senojj:
## Description
This PR addresses dangling go routines resulting from early terminations during a check evaluation.

### Problem
Currently, when a check evaluation terminates early due to finding a match to the request, or encountering an error, the check's context is canceled and the evaluation tree returns. When this happens, it is possible for slow running sub-tree evaluations to remain running within go routines while the main thread of execution moves on. The now orphaned go routines are never synchronized back to the main thread of execution and could, in the worst case scenario, remain running indefinitely and accumulate over time.

### Solution
I wrote this refactor in an attempt to provide code that is easier to reason about, and allow one to prove that a go routine does eventually end and synchronize. I've also pulled a generalized "work group" abstraction out for reusability and modularity.

### Testing
Many of the tests that currently exist for the check evaluation path do not properly `select` over a `context.Context` and a `time.After` channel. Therefore, the tested behavior relies on code that has the potential to continue running after the test has ended. I first added `goleak` validations in each of those tests to prove that a go routine would leak after the test was completed. I then adjusted the test callbacks to `select` over a `context.Context` that is canceled once the test completes, ensuring that the callback running in the go routine would end in a timely manner.

## Review Checklist


internal/graph/check.go
```diff
@@ -377,21 +324,21 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
...
-				return response, nil
+			switch ordinal {
+			case 0:
+				if !result.resp.GetAllowed() {
+					return response, nil
+				}
+			case 1:
+				if result.resp.GetAllowed() {
+					return response, nil
+				}
```
justincoh, line 335: Can you explain this switch further?

internal/graph/check.go
```diff
@@ -129,98 +129,88 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
 
 	wg.Add(1)
 	go func() {
-	outer:
-		for _, handler := range handlers {
-			fn := handler // capture loop var
+		defer func() {
+			pool.Wait()
+			close(ch)
+			wg.Done()
```
elbuo8, line 154: I guess we can completely remove the wait group at this point?

internal/graph/check.go
```diff
@@ -129,98 +129,88 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
 
-			select {
-			case limiter <- struct{}{}:
-				wg.Add(1)
-				go checker(fn)
-			case <-ctx.Done():
-				break outer
-			}
+		for i, handler := range handlers {
+			pool.Go(func(_ context.Context) error {
```
elbuo8, line 158: Why use the parent when you can use the contextPool you already wrapped?

internal/graph/check.go
senojj: This wait group monitors the go routine that is producing the input to the pool. we still want to synchronize it.

internal/graph/check.go
senojj: Astute question that I should have commented an answer for. I don't want the pool to close as a result of `ctx` being canceled. I specifically want to defer that to the function returned from `resolve`.

Therefore, I pass a local `poolCtx` into the pool, but capture `ctx` in the closure that is passed to `pool.Go` because `ctx` contains state necessary for `CheckHandlerFunc` closures that were passed in, such as the typesystem instance.

senojj: I actually reviewed the code a bit more, and found that I could just as easily wrap the parent context, eliminating the need for the closure to wrap the parent, and instead accept the context passed in from the pool.

senojj: Leaving a note to bring attention to a performance regression introduced by wrapping the parent context. Testing by reverting the code change.

internal/graph/check.go
senojj: added a comment to clarify this

senojj:
I tried to use the sourcegraph pool implementation for this use case, but its introduction caused a performance regression, as identified by our benchmark tests.

senojj:
setting this PR to draft to give me time to add adequate test coverage.

internal/graph/check_test.go
```diff
@@ -1578,10 +1631,13 @@ func TestUnionCheckFuncReducer(t *testing.T) {
...
 		ctx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
-		t.Cleanup(cancel)
+		t.Cleanup(func() {
+			cancel()
+			goleak.VerifyNone(t)
+		})
 
 		falseSlowHandler := func(context.Context) (*ResolveCheckResponse, error) {
-			time.Sleep(25 * time.Millisecond)
+			time.Sleep(1 * time.Second)
```
justincoh, line 1640: does this also need to select over a context to ensure no leak?

internal/graph/check.go
```diff
@@ -129,98 +130,118 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
+// resolve concurrently resolves one or more CheckHandlerFunc and yields the results on the provided resultChan.
+// Callers of the 'resolve' function should be sure to invoke the callback returned from this function to ensure
 // every concurrent check is evaluated. The concurrencyLimit can be set to provide a maximum number of concurrent
 // evaluations in flight at any point.
-func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- checkOutcome, handlers ...CheckHandlerFunc) func() {
-	limiter := make(chan struct{}, concurrencyLimit)
-
+func resolve(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (resultChan <-chan item[checkOutcome], end func()) {
+	ch := make(chan item[checkOutcome], len(handlers))
+	resultChan = ch
```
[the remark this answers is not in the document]
senojj: `resultChan` is a receiver channel type. a reference to `ch` which is a full channel is necessary for producing to it.

internal/graph/check.go
```diff
@@ -129,98 +130,118 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
+		}
 
-		select {
-		case <-ctx.Done():
-			return
-		case res := <-resolved:
-			resultChan <- res
+		ch <- item[checkOutcome]{
+			N:     fn.N, // N is the ordinal of the handler. this is important for difference evalutions.
+			Value: checkOutcome{resp, err},
```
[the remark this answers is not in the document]
senojj: Yes, good catch.

internal/graph/check_test.go
senojj: not necessarily for this test, since it is testing a path that would not terminate early, however, it doesn't hurt to add the logic :)

internal/graph/check.go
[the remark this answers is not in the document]
justincoh: ah I see, and the explicit receiver channel is to ensure callers know exactly what they're getting

internal/graph/check.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/openfga/openfga/pkg/telemetry"
 	"github.com/openfga/openfga/pkg/tuple"
 	"github.com/openfga/openfga/pkg/typesystem"
+	"github.com/openfga/openfga/pkg/workgroup"
```
elbuo8, line 26: Why not place in `concurrency` pkg?

internal/graph/check.go
```diff
@@ -128,98 +129,118 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
 
-	checker := func(fn CheckHandlerFunc) {
-		defer func() {
-			wg.Done()
-			<-limiter
-		}()
-
-		resolved := make(chan checkOutcome, 1)
-
+	p := workgroup.Bound(concurrencyLimit, func(fn item[CheckHandlerFunc]) error {
```
elbuo8, line 151: Any reason why not to use the lib we already use rather than creating our own?

internal/graph/check.go
senojj: That is a fair question. I think it would make sense to do so.

internal/graph/check.go
elbuo8: Was is consistent? And by a factor of how much? I ask because while this will be called a significant amount of times, it ultimately is a very % of the overall latency once you factor in IO.

senojj: It was consistent with respect to it always being the "with_exclusion" benchmark, and always having a ratio > 2. I observed this performance regression on 3 separate action runs, and the regression directly correlates to the introduction of the sourcegraph Pool.

[run #1](https://github.com/openfga/openfga/actions/runs/13314001482/job/37183407681)
[run #2](https://github.com/openfga/openfga/actions/runs/13314668819/job/37185647408)
[run #3](https://github.com/openfga/openfga/actions/runs/13437740750/job/37544076525)

There were no successful "with_exclusion" benchmarks while the sourcegraph Pool was in play.

justincoh (approved):

senojj:
abandoned due to performance regression

Issue #2277, opened 2025-02-16
Additional fields in authorization model response


mauriceackel:
### Checklist

- [x] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [x] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

## Summary
The OpenFGA API returns an authorization model JSON response that includes additional optional fields, which were not present in the original "write" call. This behavior is inconsistent with the expected output and causes challenges in maintaining state consistency when using tools like Terraform.

## Description
When an authorization model is submitted to the OpenFGA API, the response received upon retrieval includes additional fields such as `"relations": {}`, `"generic_types": []`, and `"module": ""` even though they have not been part of the original input JSON. The unexpected addition of these default values is due to the `grpc-gateway` framework's default marshaller setting, `EmitUnpopulated`, being set to `true` ([code reference](https://github.com/grpc-ecosystem/grpc-gateway/blob/e2eac23a7dda911a4ec04a81290f1a086dcd44f7/runtime/marshaler_registry.go#L23)). This causes the marshaller to set default values for certain data types as explained [in the docs](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#MarshalOptions). Notably, it sets `""` for undefined strings, `[]` for undefined arrays and `{}` for undefined objects.

### Steps to Reproduce
1. Submit an authorization model JSON to the OpenFGA API with specific attributes and fields. E.g., 
```json
{
  "schema_version": "1.1",
  "type_definitions": [
    {
      "type": "user"
    }
  ]
}
```
2. Retrieve the same authorization model JSON from the API. E.g., 
```json
{
  "schema_version": "1.1",
  "type_definitions": [
    {
      "type": "user",
      "relations" : {}
    }
  ]
}
```
3. Compare the retrieved JSON with the original input.

### Expected Results
[...]

PR #2278, opened 2025-02-16
fix: disable emission of unpopulated JSON fields

mauriceackel:
Prevents the emission of unpopulated JSON fields as explained in https://github.com/openfga/openfga/issues/2277.

## Description
By default, `grpc-gateway` set an option on the protojson marshaller that causes unpopulated JSON fields to be emitted with default values (e.g. `""`, `[]`, `{}`).

This MR uses the existing `WithMarshalerOption` function to provide a custom marshaller to the framework. This marshaller has the `EmitUnpopulated` flag set to `false`, hence preventing additional fields to be emitted.

## References
Fixes: https://github.com/openfga/openfga/issues/2277

## Review Checklist

### DIFF
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -41,6 +41,7 @@ import (
 	healthv1pb "google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
 	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
@@ -741,6 +742,16 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
 			}),
 			runtime.WithHealthzEndpoint(healthv1pb.NewHealthClient(conn)),
 			runtime.WithOutgoingHeaderMatcher(func(s string) (string, bool) { return s, true }),
+			runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.HTTPBodyMarshaler{
+				Marshaler: &runtime.JSONPb{
+					MarshalOptions: protojson.MarshalOptions{
+						EmitUnpopulated: false,
+					},
+					UnmarshalOptions: protojson.UnmarshalOptions{
+						DiscardUnknown: true,
+					},
+				},
+			}),
 		}
 		mux := runtime.NewServeMux(muxOpts...)
 		if err := openfgav1.RegisterOpenFGAServiceHandler(ctx, mux, conn); err != nil {


rhamzeh:
@mauriceackel instead of doing this on the API, maybe we can have a helper method in [Language](https://github.com/openfga/language/tree/main/pkg/go) - given two models, it checks to see if they are equal.

It could do so by passing both to proto, then marshal back to JSON so you get the consistent output, regardless of what you sent. Because whether an unpopulated field is emitted or not should not affect whether the models are equal.

E.g.
```json
{
  "type": "user"
}
```
and
```json
{
  "type": "user",
  "relations": {}
}
```

Should be treated as equivalent models

mauriceackel:
@rhamzeh, I think this is a step in the right direction! However, I'm still wondering if there is an underlying flaw with the marshalling.

My main topic of concern is a certain level of inconsistency between the API documentation and what is actually sent over the wire. For example, the `ClientReadAuthorizationModelsResponse` in the Go SDK defines that it can return a nullable string as the `ContinuationToken` result. However, this will never be sent over the REST API, due to the field being populated with `""` by default.

I'm not saying that we necessarily should set `EmitUnpopulated` to `false`, but maybe we should instead adapt the clients to more closely represent what is actually exchanged. To me it seems like there is a gap at the moment.

With regards to a short-term solution (and related to what you proposed above) I came up with the following approach to "compare" two model JSONs:

```golang
modelProto, err := transformer.LoadJSONStringToProto(jsonString)
if err != nil {
    return
}

marshaller := protojson.MarshalOptions{EmitDefaultValues: true}
unstableJsonBytes, err := marshaller.Marshal(modelProto)
if err != nil {
    return
}

var sanitizedAuthorizationModel AuthorizationModelWithoutId
err = json.Unmarshal(unstableJsonBytes, &sanitizedAuthorizationModel)
if err != nil {
    return
}

stableJsonBytes, err := json.Marshal(sanitizedAuthorizationModel)
if err != nil {
    return
}
```

I'm using this custom type to "remove" the `Id` field from the model. Not sure if there is a better approach to this. If there is, please let me know ☺️
```
type AuthorizationModelWithoutId struct {
	openfga.AuthorizationModel
	Id *interface{} `json:"-"`
}

func (o AuthorizationModelWithoutId) MarshalJSON() ([]byte, error) {
	toSerialize := map[string]interface{}{}
	toSerialize["schema_version"] = o.SchemaVersion
	toSerialize["type_definitions"] = o.TypeDefinitions
	if o.Conditions != nil {
		toSerialize["conditions"] = o.Conditions
	}
	var b bytes.Buffer
	enc := json.NewEncoder(&b)
	enc.SetEscapeHTML(false)
	err := enc.Encode(toSerialize)
	if err != nil {
		return nil, err
	}
	return b.Bytes(), nil
}
```

justincoh:
@mauriceackel Our apologies for letting this PR get stale, this issue still remains. We talked about it internally, and decided to move the two related issues over to the language repo: [#525](https://github.com/openfga/language/issues/525), [#526](https://github.com/openfga/language/issues/526).

Modifying the JSON responses themselves would require updates the the proto files, which we would like to avoid. The solution we're proposing is to have a method in `language` to do a logical comparison of two models, like @rhamzeh proposed above. I'm going to close this PR, but we are now tracking the issue over in language—we won't let it get this stale again.

PR #2282, opened 2025-02-18
feat: add `cachecontroller_cache_invalidation_count` metric

justincoh:
## Description

This PR adds an additional metric to track cache invalidation operations performed by the cache_controller.

## Review Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
-
 ### Added
 - Added `cachecontroller_cache_invalidation_count` metric to track invalidation operations. [#2282](https://github.com/openfga/openfga/pull/2282)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 - Improve Check API performance when experimental flag `enable-check-optimizations` is turned on and the model contains union of a TTU and algebraic operations. [#2200](https://github.com/openfga/openfga/pull/2200)
 - Implement dynamic TLS certificate reloading for HTTP and gRPC servers. [#2182](https://github.com/openfga/openfga/pull/2182)
 - Publicize `check.RunMatrixTests` method to allow testing against any `ClientInterface`. [#2267](https://github.com/openfga/openfga/pull/2267).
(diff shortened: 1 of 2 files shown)


internal/cachecontroller/cache_controller.go
```diff
@@ -263,17 +269,20 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
...
 // An alternative implementation could delete invalid keys, but this approach is faster (see storagewrappers.findInCache).
 func (c *InMemoryCacheController) invalidateIteratorCache(storeID string) {
+	cacheInvalidationCounter.WithLabelValues("full_invalidation").Inc()
 	c.cache.Set(storage.GetInvalidIteratorCacheKey(storeID), &storage.InvalidEntityCacheEntry{LastModified: time.Now()}, math.MaxInt)
 }
 
 // invalidateIteratorCacheByObjectRelation writes a new key to the cache.
 // An alternative implementation could delete invalid keys, but this approach is faster (see storagewrappers.findInCache).
 func (c *InMemoryCacheController) invalidateIteratorCacheByObjectRelation(storeID, object, relation string, ts time.Time) {
+	cacheInvalidationCounter.WithLabelValues("by_object_relation").Inc()
```
adriantam, line 279: Do we want to have increment the counter once per findChangesAndInvalidate?  (Also, will `cacheInvalidationCounter.WithLabelValues("by_user_and_object_type").Inc()` be identical since we are not unique by storeID/user/type etc...
Not that I am advocating on these as it will be expensive)

justincoh: Ah yes, the actual invalidation methods are called together inside a loop which would inflate this number–will move. I think maybe we just drop the label entirely and make this a raw count, so any type of invalidation counts as 1. The label is more of a "could be nice to have" and not something we know for sure we'll need right now.

internal/cachecontroller/cache_controller.go
```diff
@@ -238,11 +244,13 @@ func (c *InMemoryCacheController) findChangesAndInvalidate(ctx context.Context,
 
 	// all changes happened after the last invalidation, thus we should revoke all the cached iterators for the store.
 	if idx == len(changes)-1 {
+		cacheInvalidationCounter.Inc()
 		partialInvalidation = false
 		c.invalidateIteratorCache(storeID)
 	} else {
 		// only a subset of changes are new, revoke the respective ones.
 		lastModified := time.Now()
+		cacheInvalidationCounter.Inc()
```
adriantam, line 253: If it never gets to loop below, we should not increment the counter

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 - Improve Check API performance when experimental flag `enable-check-optimizations` is turned on and the model contains union of a TTU and algebraic operations. [#2200](https://github.com/openfga/openfga/pull/2200)
 - Implement dynamic TLS certificate reloading for HTTP and gRPC servers. [#2182](https://github.com/openfga/openfga/pull/2182)
 - Publicize `check.RunMatrixTests` method to allow testing against any `ClientInterface`. [#2267](https://github.com/openfga/openfga/pull/2267).
+- Added `cachecontroller_cache_invalidation_count` metric to track invalidation operations. [#2282](https://github.com/openfga/openfga/pull/2282)
```
adriantam, line 23: Please move this to unreleased section

adriantam (approved):

PR #2300, opened 2025-03-03
perf: use string iterators rather than tuple iterators

elbuo8:
## Description

Stop using Tuple Key Iterators when a simple static iterator suffices. Should hold less memory and slightly faster. 
Also simplify breadth first algorithm. Should be a similar amount of goroutines but following the newer patterns we have been using which are simpler to understand.

## Review Checklist


adriantam:
At high level, can you explain when we should use tuple iterator vs. string iterator?  Also, just want to confirm that the it will not affect how cache controller [invalidate the iterator](https://github.com/openfga/openfga/blob/7f87fbe815ce7d124cc76eec0676973fed88ebfa/pkg/storage/storagewrappers/check_caching.go#L287).

adriantam (dismissed):

elbuo8:
> At high level, can you explain when we should use tuple iterator vs. string iterator? 

Potentially anywhere you are using a tuple mapper is a good candidate.

> Also, just want to confirm that the it will not affect how cache controller.

Cache controller only interacts with the cache which its the downstream provider of said iterators here. So no, doesn't affect it.

internal/graph/check_fast_path.go
```diff
@@ -707,35 +702,57 @@ func (c *LocalChecker) breadthFirstRecursiveMatch(ctx context.Context, req *Reso
...
+		}
+	}()
+	nextUsersetLevel := hashset.New()
+
+ConsumerLoop:
+	for {
+		select {
+		case <-ctx.Done():
+			close(checkOutcomeChan)
+			return
```
adriantam, line 726: As noted by the warning below, probably a good case to unit test to ensure all the channels are drained and no orphan goroutine on context timeout.

adriantam (dismissed):

internal/graph/check_fast_path.go
elbuo8: Test added.

adriantam (approved):

PR #2361, opened 2025-04-04
feat: Add error handling for panics in check

Vic-Dev:
## Description

Proposed method for handling panics in goroutines:

- Use custom async handler for regular goroutines
  - This method accepts a function which it calls in a goroutine and recovers from any panics that occur
- Use [sourcegraph's](https://github.com/sourcegraph/conc) `Try` to catch panics in goroutines that can't be handled by the async handler
- Use [sourcegraph's](https://github.com/sourcegraph/conc) `WaitGroup` to catch panics when we aren't attempting to short-circuit and can afford to WaitAndRecover

Need team consensus on whether this approach would work throughout the library.

Caveats to this approach:
1. There are some cases where we are passing the panic error to a channel that may already be closed. If that is the cause of the panic, the panic error won't actually be handled. We intend to address this issue in a future PR.
2. There may be some cases where we don't want to handle a panic and the server should crash. These should probably be handled on a case by case basis.

## Review Checklist


internal/graph/check.go
```diff
@@ -316,19 +361,31 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 	limiter <- struct{}{}
 	wg.Add(1)
 	go func() {
-		resp, err := baseHandler(ctx)
-		baseChan <- checkOutcome{resp, err}
-		<-limiter
-		wg.Done()
+		recoveredError := panics.Try(func() {
```
justincoh, line 364: Why `panics.Try` instead of std lib [recover()](https://gobyexample.com/recover)? Simplicity?

internal/graph/check.go
```diff
@@ -216,32 +246,38 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
...
-	}, nil
+	}
+
+	return
 }
 
 // intersection implements a CheckFuncReducer that requires all of the provided CheckHandlerFunc to resolve
 // to an allowed outcome. The first falsey or erroneous outcome causes premature termination of the reducer.
-func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (*ResolveCheckResponse, error) {
+func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandlerFunc) (resp *ResolveCheckResponse, err error) {
```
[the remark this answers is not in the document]
Vic-Dev: We added named returns to be able to handle modifying the return values in the `defer func()`. @senojj can give more info

internal/graph/check.go
justincoh: Ah I see, [it uses recover() under the hood](https://github.com/sourcegraph/conc/blob/main/panics/panics.go#L26-L31) with a `compareAndSwap` operation to _only_ hold onto the first panic thrown in a thread-safe manner, just in case there were multiple.

internal/graph/check.go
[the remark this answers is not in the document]
justincoh: Ah I see, [this is why](https://go.dev/play/p/xSLQ0-BvLdH). Defer statements evaluate after the surrounding function returns, so unless you have a named return value the defer statement won't change the result. Helpful section of [the official golang spec](https://go.dev/ref/spec#Defer_statements)

internal/graph/check.go
```diff
@@ -216,32 +246,38 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
...
+			err = ctx.Err()
+			return
 		}
 	}
 
-	if err != nil {
-		return nil, err
+	if elErr != nil {
+		err = elErr
+		return
```
adriantam, line 260: We should send nil instead of resp in the error case

adriantam:
Please do not use go mod vendor as we don't store vendor files in OpenFGA

internal/graph/check.go
```diff
@@ -596,54 +655,60 @@ func (c *LocalChecker) produceUsersetDispatches(ctx context.Context, req *Resolv
 }
 
 // processDispatches returns a channel where the outcomes of the dispatched checks are sent, and begins sending messages to this channel.
-func (c *LocalChecker) processDispatches(ctx context.Context, limit uint32, dispatchChan chan dispatchMsg) chan checkOutcome {
+func (c *LocalChecker) processDispatches(ctx context.Context, limit uint32, dispatchChan chan dispatchMsg) (chan checkOutcome, chan error) {
 	outcomes := make(chan checkOutcome, limit)
 	dispatchPool := concurrency.NewPool(ctx, int(limit))
+	panicChan := make(chan error, 1)
```
adriantam, line 661: Instead of having two channels (outcomes / panicChan), should we consolidate to the same channel?

internal/graph/check.go
```diff
@@ -596,54 +655,60 @@ func (c *LocalChecker) produceUsersetDispatches(ctx context.Context, req *Resolv
...
+							resp, err := c.dispatch(ctx, msg.dispatchParams.parentReq, msg.dispatchParams.tk)(ctx)
+							concurrency.TrySendThroughChannel(ctx, checkOutcome{resp: resp, err: err}, outcomes)
+							return nil
+						})
+					}
 				}
 			}
+		})
+		if recoveredError != nil {
+			panicChan <- fmt.Errorf("panic occurred: %v", recoveredError)
```
adriantam, line 702: If we leave panicChan as separate channel, be sure to use concurrency.TrySendThroughChannel

internal/graph/check.go
```diff
@@ -165,14 +175,23 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
...
 			case limiter <- struct{}{}:
 				wg.Add(1)
-				go checker(fn)
+				go func() {
+					defer wg.Done()
+					recoveredError := panics.Try(func() {
+						checker(fn)
+					})
+					if recoveredError != nil {
+						errorChan <- fmt.Errorf("panic occurred: %v", recoveredError)
```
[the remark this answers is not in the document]
senojj: Is the recovered error typed?

internal/graph/check.go
```diff
@@ -151,8 +153,16 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
...
-			resolved <- checkOutcome{resp, err}
+			defer func() {
+				close(resolved)
+			}()
+			recoveredError := panics.Try(func() {
+				resp, err := fn(ctx)
+				resolved <- checkOutcome{resp, err}
+			})
+			if recoveredError != nil {
+				errorChan <- fmt.Errorf("panic occurred: %v", recoveredError)
```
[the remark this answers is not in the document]
Vic-Dev: ^ in all places

internal/graph/check.go
senojj: they are different types... writing to outcomes can panic, writing to panicChan is guaranteed not to.

internal/graph/check.go
```diff
@@ -181,33 +200,44 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
...
 	resultChan := make(chan checkOutcome, len(handlers))
 
 	drain := resolver(ctx, concurrencyLimit, resultChan, handlers...)
 
 	defer func() {
 		cancel()
-		drain()
+		drainErr := drain()
+		if drainErr != nil {
+			err = drainErr
```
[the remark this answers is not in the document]
senojj: @cikasfm why should join be used here?

internal/graph/check.go
```diff
@@ -216,32 +246,38 @@ func union(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHandle
...
-			return nil, ctx.Err()
+			err = ctx.Err()
+			return
 		}
 	}
 
-	if err != nil {
-		return nil, err
+	if elErr != nil {
+		err = elErr
```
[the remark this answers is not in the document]
senojj: This is setting the empty return value `err` to the value of `elErr`... I don't think the use of `errors.Join` would make sense here.

internal/graph/check.go
```diff
@@ -253,36 +289,44 @@ func intersection(ctx context.Context, concurrencyLimit uint32, handlers ...Chec
...
+			err = ctx.Err()
+			return
 		}
 	}
 
 	// all operands are either truthy or we've seen at least one error
-	if err != nil {
-		return nil, err
+	if elErr != nil {
+		err = elErr
```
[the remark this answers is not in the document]
senojj: This is setting the empty return value err to the value of elErr... I don't think the use of errors.Join would make sense here.

internal/graph/check.go
```diff
@@ -316,19 +361,31 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 	limiter <- struct{}{}
 	wg.Add(1)
 	go func() {
-		resp, err := baseHandler(ctx)
-		baseChan <- checkOutcome{resp, err}
-		<-limiter
-		wg.Done()
+		recoveredError := panics.Try(func() {
+			defer wg.Done()
+			defer func() { <-limiter }()
```
senojj, line 366: these should be combined into a single `defer` statement, so that order of execution is more explicit.

internal/graph/check.go
```diff
@@ -316,19 +361,31 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
...
 	limiter <- struct{}{}
 	wg.Add(1)
 	go func() {
-		resp, err := subHandler(ctx)
-		subChan <- checkOutcome{resp, err}
-		<-limiter
-		wg.Done()
+		recoveredError := panics.Try(func() {
+			defer wg.Done()
+			defer func() { <-limiter }()
```
senojj, line 381: these should be combined into a single defer statement, so that order of execution is more explicit.

internal/graph/check.go
```diff
@@ -132,14 +134,15 @@ type CheckFuncReducer func(ctx context.Context, concurrencyLimit uint32, handler
...
 // every concurrent check is evaluated. The concurrencyLimit can be set to provide a maximum number of concurrent
 // evaluations in flight at any point.
-func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- checkOutcome, handlers ...CheckHandlerFunc) func() {
+func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- checkOutcome, handlers ...CheckHandlerFunc) func() error {
 	limiter := make(chan struct{}, concurrencyLimit)
 
-	var wg sync.WaitGroup
+	var wg conc.WaitGroup
+
+	errorChan := make(chan error, 3)
```
adriantam, line 142: why 3?

internal/graph/check.go
```diff
@@ -300,6 +332,7 @@ func exclusion(ctx context.Context, concurrencyLimit uint32, handlers ...CheckHa
 	ctx, cancel := context.WithCancel(ctx)
 	baseChan := make(chan checkOutcome, 1)
 	subChan := make(chan checkOutcome, 1)
+	errorChan := make(chan error, 2)
```
adriantam, line 335: again, why 2?

internal/graph/check.go
```diff
@@ -163,51 +169,63 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
...
 
-		wg.Done()
-	}()
-
-	return func() {
-		wg.Wait()
+	return func() error {
+		recoveredError := wg.WaitAndRecover()
+		if recoveredError != nil {
+			errorChan <- fmt.Errorf("panic occurred: %w", recoveredError.AsError())
```
adriantam, line 191: In the past, we have goroutine leak because we attempt to send to a channel when the context is cancelled.  I would encourage us to audit whether this is possible.  We may have to consider using [TrySendThroughChannel](https://github.com/openfga/openfga/blob/73dceae92683474c2d788c9b4467abe2d2e3b3db/internal/concurrency/concurrency.go#L21)

This comment applies to other places where we attempt send to errorChan or panicChan

internal/graph/check.go
adriantam: Do we need to close errorChan?

internal/graph/check.go
adriantam: Who is responsible for closing errorChan?

senojj: > why 3?

That's a fair question. Are three errors possible to be written to this channel?

> Who is responsible for closing errorChan?

Channels are not required to be explicitly closed. They will be cleaned up by the GC once they go out of scope.

internal/graph/check.go
senojj: Why would it matter if it sends to the channel after the context is canceled in this case?

internal/graph/check.go
senojj: > again, why 2?

Looks like this logic can produce at most 2 errors. The buffer of two makes room for both of them without blocking... A comment indicating this would be good here.

> Do we need to close errorChan?

Channel is cleaned up by the GC after going out of scope.

senojj:
After seeing this implementation, I question if we are trying to capture panics using too broad of strokes here. I know we want to protect against code changes that introduce `nil` pointer dereferences, or concurrent map access, but perhaps we should be focusing on code that has the possibility of panicking as it sits today. I have ideas that could make this easier to manage.

Always employ defensive programming practices:
1. Within a new scope, always check if a pointer is `nil` before dereferencing it.
2. Always guard shared maps with a mutex or use `sync.Map`.
3. Use `TrySendThroughChannel` to send to channels (function will need to be updated to recover from attempts to write to a closed channel).

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 - Added "request.throttled" boolean to the context tags for check and batch-check
 - Added "throttled_requests_count" metric to batch-check requests.
 
+### Changed
+- Panics in check are converted to errors. [#2361](https://github.com/openfga/openfga/pull/2361).
```
willvedd, line 17:
```suggestion
- Panics in goroutines in check are converted to errors. [#2361](https://github.com/openfga/openfga/pull/2361).
```

internal/graph/check.go
```diff
@@ -163,51 +169,63 @@ func resolver(ctx context.Context, concurrencyLimit uint32, resultChan chan<- ch
...
+			errorChan <- fmt.Errorf("%w: %s", utils.ErrPanic, recoveredError.AsError())
+		}
 		close(limiter)
+
+		select {
+		case err := <-errorChan:
+			return err
+		default:
+			return nil
+		}
```
willvedd, line 200: **Question:** Do we _need_ the error channel at all? Could we simplify with something like:

```suggestion
		recoveredError := wg.WaitAndRecover()
		close(limiter)
		
		if recoveredError != nil {
			return fmt.Errorf("%w: %s", utils.ErrPanic, recoveredError.AsError())
		}
		return nil
```

internal/graph/check.go
willvedd: Do we even need this error channel? It's only being written to once, after the `wg.WaitAndRecover()`. Made a [suggestion](https://github.com/openfga/openfga/pull/2361/files#r2046915759) below.

internal/graph/check.go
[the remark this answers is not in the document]
cikasfm: @senojj There's 3 different places ( line 209, 221 and 240 ) where errors are / may be set, this just makes it:

1. hard to read
2. unclear which error will be returned

Each one is overriding the value of the previous one, so the real error might & will be hidden, unless we assume that the previous value is not important.

PR #2381, opened 2025-04-16
fix: cached_datastore.ReadStartingWithUser cache keys

justincoh:
## Description

In some cases, the cached_datastore was failing to invalidate iterator cache entries. The cache invalidation flow is complicated, here's roughly how it works for this particular case:

* When the cached_datastore calls [ReadStartingWithUser](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/cached_datastore.go#L125), it attempts to load a cached iterator and passes its constructor [a list of invalidEntityKeys](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/cached_datastore.go#L357).
* Those `invalidEntityKeys` are an indication that a particular cached iterator should not be used—[they're a faster alternative](https://github.com/openfga/openfga/blob/main/internal/cachecontroller/cache_controller.go#L287-L288) to calling `cache.Delete` on now-invalidated keys.
* If one of those `invalidEntityKeys` is found in cache **and** the cached iterator we are about to use is older than that invalidEntityKey entry, [we create a new iterator](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/cached_datastore.go#L379-L392).

In the case where ReadStartingWithUser was called with a UserFilter that had no `Relation`, the `invalidEntityKeys` contained an extra `#` which did **not** match the invalid entries added by [the cache controller here](https://github.com/openfga/openfga/blob/main/internal/cachecontroller/cache_controller.go#L289-L291), as they do not contain the `#`. So these iterators were not invalidated, and stale entries could be returned. This PR removes the extra `#` so iterators will invalidate properly.

## Review Checklist


adriantam (commented):
Do we want to have any test reproducing the scenario and checking that it has been fixed?

justincoh:
Talked it over with Adrian and we're going to go ahead without additional testing; this is a difficult area. Ideally we'd have an end to end test for this full workflow, but timing-dependent tests like this are difficult to get right. For now, the modification made to a unit test in this PR will have to suffice

pkg/storage/storagewrappers/cached_datastore.go
```diff
@@ -152,7 +152,10 @@ func (c *CachedDatastore) ReadStartingWithUser(
 	// since at most it will have 2 entries (user and wildcard if possible)
 	subjects := make([]string, 0, len(filter.UserFilter))
 	for _, objectRel := range filter.UserFilter {
-		subject := tuple.ToObjectRelationString(objectRel.GetObject(), objectRel.GetRelation())
+		subject := objectRel.GetObject()
+		if objectRel.GetRelation() != "" {
+			subject = tuple.ToObjectRelationString(objectRel.GetObject(), objectRel.GetRelation())
+		}
 		subjects = append(subjects, subject)
```
justincoh, line 159: codecov thinks this behavior isn't covered but I had to specifically adjust for this change in `cached_datstore_test` below 🤷

adriantam: Do we want to have a test for [TestReadStartingWithUser](https://github.com/openfga/openfga/blob/b71cf4d4b661e32e56ea48f32c363409b055129a/pkg/storage/storagewrappers/cached_datastore_test.go#L142) where the userfilter's relation is != nil

senojj: I also happened to notice that this function never checks for a `nil` value in `filter`.

justincoh: Where? It looks like filter is safe. The top-level `filter` argument is a [struct type](https://github.com/openfga/openfga/blob/b71cf4d4b661e32e56ea48f32c363409b055129a/pkg/storage/storage.go#L239), so `filter.UserFilter` will always be present, and within the individual `objectRel` entries inside the `range` loop we're using [nil-safe getters](https://github.com/openfga/api/blob/main/proto/openfga/v1/authzmodel.pb.go#L986-L991).

senojj: My apologies, I wrongfully assumed it was an interface... That's what I get for assuming.

yissellokta (approved):

PR #2388, opened 2025-04-21
feat: Add Iterator Cache to ListObjects

justincoh:
## Description

This PR enables ListObjects to take advantage of cached iterators. At startup, OpenFGA generates a single in-memory cache instance. This instance is passed into Check and now (with this PR) into ListObjects as well. The configuration flags for ListObjects are separate:

* `OPENFGA_LIST_OBJECTS_ITERATOR_CACHE_ENABLED`
* `OPENFGA_LIST_OBJECTS_ITERATOR_CACHE_TTL`
* `OPENFGA_LIST_OBJECTS_ITERATOR_CACHE_MAX_RESULTS`

This PR also adds a new `method` label to the CachedDatastore metrics so we can differentiate between the status of iterators in use by Check vs ListObjects.

Other changes:
* NewRequestStorageWrapper -> NewRequestStorageWrapperWithCache
* some renames of sharedCheckResources -> sharedDatastoreResources

## Review Checklist

### DIFF
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -234,7 +234,7 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Duration("check-iterator-cache-ttl", defaultConfig.CheckIteratorCache.TTL, "if caching of datastore iterators of Check requests is enabled, this is the TTL of each value")
 
-	flags.Bool("list-objects-iterator-cache-enabled", defaultConfig.ListObjectsIteratorCache.Enabled, "enable caching of datastore iterators for list objects. The key is a string representing a database query, and the value is a list of tuples. Each iterator is the result of a database query, for example usersets related to a specific object, or objects related to a specific user, up to a certain number of tuples per iterator. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.")
+	flags.Bool("list-objects-iterator-cache-enabled", defaultConfig.ListObjectsIteratorCache.Enabled, "enable caching of datastore iterators for ListObjects. The key is a string representing a database query, and the value is a list of tuples. Each iterator is the result of a database query, for example usersets related to a specific object, or objects related to a specific user, up to a certain number of tuples per iterator. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.")
 
 	flags.Uint32("list-objects-iterator-cache-max-results", defaultConfig.ListObjectsIteratorCache.MaxResults, "if caching of datastore iterators of ListObjects requests is enabled, this is the limit of tuples to cache per key.")
 
@@ -271,7 +271,7 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Uint32("listObjects-dispatch-throttling-threshold", defaultConfig.ListObjectsDispatchThrottling.Threshold, "defines the number of dispatches above which ListObjects requests will be throttled.")
 
-	flags.Uint32("listObjects-dispatch-throttling-max-threshold", defaultConfig.ListObjectsDispatchThrottling.MaxThreshold, "define the maximum dispatch threshold beyond which a list objects requests will be throttled. 0 will use the 'listObjects-dispatch-throttling-threshold' value as maximum")
+	flags.Uint32("listObjects-dispatch-throttling-max-threshold", defaultConfig.ListObjectsDispatchThrottling.MaxThreshold, "define the maximum dispatch threshold beyond which a ListObjects requests will be throttled. 0 will use the 'listObjects-dispatch-throttling-threshold' value as maximum")
 
 	flags.Bool("listUsers-dispatch-throttling-enabled", defaultConfig.ListUsersDispatchThrottling.Enabled, "enable throttling when a ListUsers request's number of dispatches is high. Enabling this feature will prioritize dispatched requests requiring less than the configured dispatch threshold over requests whose dispatch count exceeds the configured threshold.")
 
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -234,6 +234,12 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Duration("check-iterator-cache-ttl", defaultConfig.CheckIteratorCache.TTL, "if caching of datastore iterators of Check requests is enabled, this is the TTL of each value")
 
+	flags.Bool("list-objects-iterator-cache-enabled", defaultConfig.ListObjectsIteratorCache.Enabled, "enable caching of datastore iterators for list objects. The key is a string representing a database query, and the value is a list of tuples. Each iterator is the result of a database query, for example usersets related to a specific object, or objects related to a specific user, up to a certain number of tuples per iterator. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.")
+
+	flags.Uint32("list-objects-iterator-cache-max-results", defaultConfig.ListObjectsIteratorCache.MaxResults, "if caching of datastore iterators of ListObjects requests is enabled, this is the limit of tuples to cache per key.")
+
+	flags.Duration("list-objects-iterator-cache-ttl", defaultConfig.ListObjectsIteratorCache.TTL, "if caching of datastore iterators of ListObjects requests is enabled, this is the TTL of each value")
+
 	flags.Bool("check-query-cache-enabled", defaultConfig.CheckQueryCache.Enabled, "enable caching of Check requests. For example, if you have a relation define viewer: owner or editor, and the query is Check(user:anne, viewer, doc:1), we'll evaluate the owner relation and the editor relation and cache both results: (user:anne, viewer, doc:1) -> allowed=true and (user:anne, owner, doc:1) -> allowed=true. The cache is stored in-memory; the cached values are overwritten on every change in the result, and cleared after the configured TTL. This flag improves latency, but turns Check and ListObjects into eventually consistent APIs. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.")
 
 	flags.Uint32("check-query-cache-limit", defaultConfig.CheckCache.Limit, "DEPRECATED: Use check-cache-limit instead. If caching of Check and ListObjects calls is enabled, this is the size limit of the cache")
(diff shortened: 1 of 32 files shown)


pkg/server/commands/list_objects.go
```diff
@@ -410,6 +434,12 @@ func (q *ListObjectsQuery) Execute(
 
 	resolutionMetadata := NewListObjectsResolutionMetadata()
 
+	if req.GetConsistency() != openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
```
justincoh, line 437: and LO cache is enabled

internal/cachecontroller/cache_controller.go
```diff
@@ -56,9 +56,14 @@ var (
...
-	// DetermineInvalidationTime returns the timestamp of the last write for the specified store.
-	// It may return a cached timestamp.
+	// DetermineInvalidationTime returns the timestamp of the last write for the specified store if it was in cache,
+	// Else it returns Zero time and triggers InvalidateIfNeeded().
 	DetermineInvalidationTime(ctx context.Context, storeID string) time.Time
+
+	// InvalidateIfNeeded checks to see if an invalidation is currently in progress for a store,
+	// and if not it will spawn a goroutine to invalidate cached records conditionally
+	// based on timestamp. It may invalidate all cache records, some, or none.
+	InvalidateIfNeeded(storeID string, parentSpan trace.Span)
```
adriantam, line 66: why are we passing span as parameter?  Other patterns we have is to pass in context and derive the parent based on that.

internal/cachecontroller/cache_controller.go
```diff
@@ -166,9 +162,30 @@ func (c *InMemoryCacheController) findChangesDescending(ctx context.Context, sto
...
+	_, present := c.inflightInvalidations.LoadOrStore(storeID, struct{}{})
+	if present {
+		// If invalidation is already in process, abort.
+		return
+	}
+
+	span.SetAttributes(attribute.Bool("cache_controller_invalidation", true))
+
+	go func() {
+		// we do not want to propagate context to avoid early cancellation
```
adriantam, line 175: Consider using [context.WithoutCancel(parent context.Context)](https://pkg.go.dev/context#WithoutCancel) to avoid early cancellation.

internal/shared/shared.go
```diff
@@ -66,6 +69,10 @@ func NewSharedDatastoreResources(
 		s.CacheController = cachecontroller.NewCacheController(ds, s.CheckCache, settings.CacheControllerTTL, settings.CheckIteratorCacheTTL, cachecontroller.WithLogger(s.Logger))
 	}
 
+	for _, opt := range opts {
```
adriantam, line 72: Can you explain why we want to move this down?

pkg/server/config/config.go
```diff
@@ -597,6 +602,14 @@ func (cfg *Config) verifyCacheConfig() error {
 			return errors.New("'checkIteratorCache.maxResults' must be greater than zero")
 		}
 	}
+	if cfg.ListObjectsIteratorCache.Enabled {
+		if cfg.ListObjectsIteratorCache.TTL <= 0 {
+			return errors.New("'listObjectsIteratorCache.ttl' must be greater than zero")
+		}
+		if cfg.ListObjectsIteratorCache.MaxResults <= 0 {
+			return errors.New("'listObjectsIteratorCache.maxResults' must be greater than zero")
+		}
```
adriantam, line 611: Please add test on these

pkg/server/commands/list_objects.go
```diff
@@ -410,6 +434,12 @@ func (q *ListObjectsQuery) Execute(
 
 	resolutionMetadata := NewListObjectsResolutionMetadata()
 
+	if req.GetConsistency() != openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY && q.cacheSettings.ShouldCacheListObjectsIterators() {
+		_, span := tracer.Start(ctx, "commands.listObjects")
+		defer span.End()
+
+		q.sharedDatastoreResources.CacheController.InvalidateIfNeeded(req.GetStoreId(), span)
```
adriantam, line 441: curious why do we do this for list objects but not check?  I would have expected `check` call the InvalidateIfNeeded logic..

internal/cachecontroller/cache_controller.go
justincoh: This is just to keep parity with [the existing behavior here](https://github.com/openfga/openfga/blob/main/internal/cachecontroller/cache_controller.go#L169). My guess is we were passing a parentSpan because we were explicitly [*not* passing the context [through](https://github.com/openfga/openfga/blob/9b5974458b777707ed2a30ba6303699499e655ee/internal/cachecontroller/cache_controller.go#L146-L151)

internal/cachecontroller/cache_controller.go
justincoh: That is a reasonable way to do this as well, and possibly remove the need for the `span` parameter. This is our current behavior though, I didn't make any changes (just moved things around)

internal/shared/shared.go
justincoh: I also added `WithCacheController`, and this was moved down to enable me to use that hook to override the default cache_controller—it's so I can assert that methods were called in test.

pkg/server/commands/list_objects.go
justincoh: Check does call this but indirectly as part of its call to `DetermineInvalidationTime`, which ListObjects isn't using

cmd/run/run.go
```diff
@@ -234,6 +234,12 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Duration("check-iterator-cache-ttl", defaultConfig.CheckIteratorCache.TTL, "if caching of datastore iterators of Check requests is enabled, this is the TTL of each value")
 
+	flags.Bool("list-objects-iterator-cache-enabled", defaultConfig.ListObjectsIteratorCache.Enabled, "enable caching of datastore iterators for list objects. The key is a string representing a database query, and the value is a list of tuples. Each iterator is the result of a database query, for example usersets related to a specific object, or objects related to a specific user, up to a certain number of tuples per iterator. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.")
```
jpadilla, line 237:
```suggestion
	flags.Bool("list-objects-iterator-cache-enabled", defaultConfig.ListObjectsIteratorCache.Enabled, "enable caching of datastore iterators for ListObjects. The key is a string representing a database query, and the value is a list of tuples. Each iterator is the result of a database query, for example usersets related to a specific object, or objects related to a specific user, up to a certain number of tuples per iterator. If the request's consistency is HIGHER_CONSISTENCY, this cache is not used.")
```

internal/cachecontroller/cache_controller.go
```diff
@@ -287,5 +304,6 @@ func (c *InMemoryCacheController) invalidateIteratorCacheByObjectRelation(storeI
 // invalidateIteratorCacheByUserAndObjectType writes a new key to the cache.
 // An alternative implementation could delete invalid keys, but this approach is faster (see storagewrappers.findInCache).
 func (c *InMemoryCacheController) invalidateIteratorCacheByUserAndObjectType(storeID, user, objectType string, ts time.Time) {
+	c.logger.Debug("InMemoryCacheController invalidateIteratorCacheByUserAndObjectType invalidation")
```
jpadilla, line 307: can probably rm or include similar debug log in other `invalidateIterator*` functions

pkg/server/commands/list_objects.go
jpadilla: Can we encode the reasoning/purpose of what's happening here?

pkg/server/server_test.go
```diff
@@ -2123,9 +2123,9 @@ func TestServerCheckCache(t *testing.T) {
 		)
 		t.Cleanup(s.Close)
 
-		require.NotNil(t, s.sharedCheckResources.CheckCache)
```
jpadilla, line 2126: Do we want/need the ListObjects variant of this `TestServerCheckCache` test?

internal/cachecontroller/cache_controller.go
```diff
@@ -166,9 +162,30 @@ func (c *InMemoryCacheController) findChangesDescending(ctx context.Context, sto
...
+		c.findChangesAndInvalidateIfNecessary(context.Background(), storeID, span)
+		c.inflightInvalidations.Delete(storeID)
+	}()
+}
+
+// findChangesAndInvalidateIfNecessary checks the most recent entry in this store's changelog against the most
+// recent cached changelog entry. If the most recent changelog entry is older than the cached changelog timestamp,
+// no invalidation is necessary and we return. If not, we locate changelog records that have been around for longer
+// than the cache's TTL and invalidate them.
+func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(ctx context.Context, storeID string, parentSpan trace.Span) {
```
elbuo8, line 186: Why change this method? Your caller handles the "if necessary" part.

pkg/server/commands/list_objects.go
```diff
@@ -410,6 +434,13 @@ func (q *ListObjectsQuery) Execute(
 
 	resolutionMetadata := NewListObjectsResolutionMetadata()
 
+	if req.GetConsistency() != openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY && q.cacheSettings.ShouldCacheListObjectsIterators() {
+		_, span := tracer.Start(ctx, "commands.listObjects")
```
elbuo8, line 438: This span seems off to me. Is there not an existing span already that is the parent?

internal/cachecontroller/cache_controller.go
justincoh: just for clarity. `findChangesAndInvalidate` does have a path where it returns without performing any invalidations. `InvalidateIfNeeded` could be more accurately name `InvalidateIfNotInProgress` but that's a mouthful, and I can't think of a good combination of names that 1) isn't bad and 2) isn't ambiguous

pkg/server/config/cache_settings.go
```diff
@@ -38,6 +44,10 @@ func (c CacheSettings) ShouldCacheCheckQueries() bool {
 	return c.CheckCacheLimit > 0 && c.CheckQueryCacheEnabled
 }
 
-func (c CacheSettings) ShouldCacheIterators() bool {
+func (c CacheSettings) ShouldCacheCheckIterators() bool {
 	return c.CheckCacheLimit > 0 && c.CheckIteratorCacheEnabled
 }
+
+func (c CacheSettings) ShouldCacheListObjectsIterators() bool {
```
adriantam, line 51: Do we want to add test on this?

adriantam (approved):

pkg/server/commands/list_objects.go
justincoh: yes good call. server/list_objects has one, I can grab it via spanFromContext

pkg/server/config/cache_settings.go
justincoh: already [covered here](https://github.com/openfga/openfga/pull/2388/files#diff-fe7a244e18e92d8f347af17363209172e1d79224f086b0585a1bbeef3b626c4eR2187)

PR #2396, opened 2025-04-24
chore: add `check_cache_item_count` metric

justincoh:
## Description

Small metric PR for more visibility into how many items are being added to the Check query cache

## References

Screenshot 2025-04-25 at 2 12 09 PM

## Review Checklist

### DIFF
--- a/internal/graph/cached_resolver.go
+++ b/internal/graph/cached_resolver.go
@@ -45,9 +45,9 @@ var (
 		Help:      "The total number of cache hits for ResolveCheck that were discarded because they were invalidated.",
 	})
 
-	checkCacheSetCount = promauto.NewCounter(prometheus.CounterOpts{
+	CheckCacheItemCount = promauto.NewGauge(prometheus.GaugeOpts{
 		Namespace: build.ProjectName,
-		Name:      "check_cache_set_count",
+		Name:      "check_cache_item_count",
 		Help:      "The total number of times we store items in the cache",
 	})
 )
@@ -195,7 +195,7 @@ func (c *CachedCheckResolver) ResolveCheck(
 
 	clonedResp := resp.clone()
 
-	checkCacheSetCount.Inc()
+	CheckCacheItemCount.Inc()
 	c.cache.Set(cacheKey, &CheckResponseCacheEntry{LastModified: time.Now(), CheckResponse: clonedResp}, c.cacheTTL)
 	return resp, nil
 }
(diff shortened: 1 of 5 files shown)


internal/graph/cached_resolver.go
```diff
@@ -44,6 +44,12 @@ var (
 		Name:      "check_cache_invalid_hit_count",
 		Help:      "The total number of cache hits for ResolveCheck that were discarded because they were invalidated.",
 	})
+
+	checkCacheSetCount = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "check_cache_set_count",
+		Help:      "The total number of times we store items in the cache",
```
adriantam, line 51:
```suggestion
		Help:      "The total number of times items are stored in the cache",
```

adriantam (commented):
We should call this out in the changelog as external folks can start using this metric.

pkg/storage/cache.go
```diff
@@ -333,7 +344,6 @@ func WriteCheckCacheKey(w io.StringWriter, params *CheckCacheKeyParams) error {
 func WriteInvariantCheckCacheKey(w io.StringWriter, params *CheckCacheKeyParams) error {
 	_, err := w.WriteString(
 		" " + // space to separate from user in the TupleCacheKey, where spaces cannot be present
-			SubproblemCachePrefix +
```
adriantam, line 336: Why do we move SubproblemCachePrefix logic into cached_resolver.go?

justincoh: Good question. The cache prefixes were intended to be prefixes, at the beginning of the key string e.g. `{prefix}/{store}/{object}/...etc`. All of our cache keys **except** the check subquery cache keys follow some version of that `prefix + other_stuff` pattern.

[This cache key is treated differently](https://github.com/openfga/openfga/blob/main/internal/graph/cached_resolver.go#L196-L206). We create a cache key string, which is then hashed to a uint64, which is then converted back to a string for use as a key as a stringified number e.g. `"16923876468723"`. I'm not sure why exactly we did it this way, but my guess is that cache keys with context and contextual tuples could get quite long so we hashed them. This change just makes that existing cache key `"sp.16923876468723"` for consistency.

All that said, this isn't strictly necessary for this PR and can be removed if we want.

jpadilla:
closed in favor of https://github.com/openfga/openfga/pull/2412

PR #2397, opened 2025-04-25
feat: add counters on current number of iterator cache

adriantam:
## Description
Add counters on current number of iterator cache as gauge.

## Review Checklist

## Testing
Screenshot 2025-04-25 at 3 43 23 PM
Purposely not calling Stop
Screenshot 2025-04-25 at 3 15 50 PM

### DIFF
--- a/pkg/storage/storagewrappers/cached_datastore.go
+++ b/pkg/storage/storagewrappers/cached_datastore.go
@@ -490,7 +490,7 @@ func (c *cachedIterator) Stop() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	currentIteratorCacheCount.Dec()
+	defer currentIteratorCacheCount.Dec()
 
 	swapped := c.closing.CompareAndSwap(false, true)
 	if !swapped {
--- a/pkg/storage/storagewrappers/cached_datastore.go
+++ b/pkg/storage/storagewrappers/cached_datastore.go
@@ -59,9 +59,9 @@ var (
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"operation"})
 
-	currentCacheCount = promauto.NewGauge(prometheus.GaugeOpts{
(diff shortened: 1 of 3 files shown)


pkg/storage/storagewrappers/cached_datastore.go
```diff
@@ -384,6 +390,7 @@ func (c *CachedDatastore) newCachedIterator(
 		span.SetAttributes(attribute.Bool("cached", true))
 
 		staticIter := storage.NewStaticIterator[*storage.TupleRecord](cacheEntry.Tuples)
+		currentIteratorCacheCount.Inc()
```
justincoh, line 393: This looks like we're finding this iterator already in cache, should we be `Inc()` here?

adriantam: No.  From the caller perspective, they are separate "iterator".  I want to track whether there are cases where we create the iterator (either as cache or the original) but never call stop.

justincoh (approved):

pkg/storage/storagewrappers/cached_datastore.go
```diff
@@ -394,6 +400,7 @@ func (c *CachedDatastore) newCachedIterator(
...
```
jpadilla, line 403: @adriantam wouldn't this one never be decreased?

adriantam: It is decremented at https://github.com/openfga/openfga/blob/d737c55a5287cc986b31f8e9bdda7ed49e3e7f76/pkg/storage/storagewrappers/cached_datastore.go#L526

Note that both the instance generated from cache and not from cache are "separate" entity and require .Stop() to be called separately.

jpadilla: note this is an instance of `cachedTupleIterator`, not `cachedIterator`

justincoh (dismissed):

jpadilla (approved):

PR #2407, opened 2025-04-30
upgrade to Postgres v17

JeremyLoy:
Adds support for PG 17. [See Slack discussion here](https://cloud-native.slack.com/archives/C06G1NNH47N/p1746011136908079?thread_ts=1745587524.469969&cid=C06G1NNH47N)

Its best to view this PR with 'hide whitespace changes' as a significant amount of the diff is merely indentation changes from wrapping the postgres tests in a loop to run against both v14 and v17 

image

## Description
Updates the default for `docker-compose` and local dev to PG 17. Tests againstn PG 14 and PG17 for all tests

## References

https://github.com/openfga/openfga.dev/pull/1024

## Review Checklist


docker-compose.yaml
```diff
@@ -1,4 +1,20 @@
...
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=password
+    healthcheck:
+      test: [ "CMD-SHELL", "pg_isready -U postgres" ]
+      interval: 5s
+      timeout: 5s
+      retries: 5
   postgres:
     image: postgres:14
     container_name: postgres
```
JeremyLoy, line 20: I edited the `docker-compose` a bit differently than the tests, as I noticed they spun up their own containers dynamically.

If you want, this can just have a single `postgres` entry defaulted to 17. Considering there isn't a `MySQL` here, I think this docker-compose is not meant to be exhaustive

cmd/validatemodels/validate_models_test.go
```diff
@@ -15,11 +15,12 @@ import (
 
 	"github.com/openfga/openfga/cmd"
 	"github.com/openfga/openfga/cmd/util"
+	"github.com/openfga/openfga/pkg/testutils"
 	"github.com/openfga/openfga/pkg/typesystem"
 )
 
 func TestValidationResult(t *testing.T) {
-	engines := []string{"postgres", "mysql", "sqlite"}
+	engines := append(testutils.PostgresImages, "mysql", "sqlite")
```
JeremyLoy, line 23: I'm not a huge fan of this, as its conflating images with engines, but the test suite is missing that abstraction as a whole. I'm open to any architecture suggestions!

JeremyLoy:
Hey! Sorry for the rocky start here, all of the tests and lints are now passing. 😄 

Let me know if you want the scheme I made up to support multiple versions needs rearchitecting.

Also, I wanted to preserve the format of the table tests, but I also didn't want to hard code any "14" or "17". The format looks a bit weird, open to feedback on it.

Makefile
```diff
@@ -129,14 +129,22 @@ dev-run: $(GO_BIN)/CompileDaemon $(GO_BIN)/openfga ## Run the OpenFGA server wit
 			CompileDaemon -graceful-kill -build='make install' -command="openfga run --datastore-engine mysql --datastore-uri root:secret@tcp(localhost:3306)/openfga?parseTime=true"; \
 			break; \
 			;; \
-		"postgres") \
-			echo "==> Running OpenFGA with Postgres data storage"; \
+		"postgres14") \
```
adriantam, line 132: a bit concern that this is a backward breaking change.  Will it be possible to keep this as postgres? (or have both posgres and postgres14)

JeremyLoy: yeah, I'll change `postgres17` to postgres so that there is no breaking change to scripts, but the behavior is changed to default to the newest pg

jpadilla (changes requested):
👋 Hi @JeremyLoy, thanks for this.

I don't particularly have any concerns of changing the default postgres version we test against today to pg 17. Our postgres use today is pretty vanilla. If/when that ever changes we can revisit how we test against multiple versions. Basically, this can evolve in the future, we can keep it simple now.

As far as changes go, update postgres version

1. On docker compose
2. On Makefile
3. On test container

We then need to figure out how to roll out a similar bump and release on https://github.com/openfga/helm-charts/

JeremyLoy:
@jpadilla thanks for the response. 

This PR already handles the Makefile, testcontainer, and docker-compose. So I think we're good once I resolve the merge conflicts.

With regards to the helm-charts, I don't see anywhere that a specific version of `postgres` is being defined, so I think it's already using v17?

The version `12.12.10` is for the bitnami chart that is being imported, and it defaults to v17.

https://github.com/openfga/helm-charts/blob/0689e6c48576da84d077226b69e5bafdcd912fb3/charts/openfga/Chart.yaml#L19-L21

https://github.com/bitnami/charts/blob/d210adf595a8b6deb798cfdb0d8b7934a59fe6c7/bitnami/postgresql/values.yaml#L121-L124

Do you agree?

jpadilla:
> @jpadilla thanks for the response.
> 
> This PR already handles the Makefile, testcontainer, and docker-compose. So I think we're good once I resolve the merge conflicts.
> 
> With regards to the helm-charts, I don't see anywhere that a specific version of `postgres` is being defined, so I think it's already using v17?
> 
> The version `12.12.10` is for the bitnami chart that is being imported, and it defaults to v17.
> 
> [openfga/helm-charts@`0689e6c`/charts/openfga/Chart.yaml#L19-L21](https://github.com/openfga/helm-charts/blob/0689e6c48576da84d077226b69e5bafdcd912fb3/charts/openfga/Chart.yaml#L19-L21)
> 
> [bitnami/charts@`d210adf`/bitnami/postgresql/values.yaml#L121-L124](https://github.com/bitnami/charts/blob/d210adf595a8b6deb798cfdb0d8b7934a59fe6c7/bitnami/postgresql/values.yaml#L121-L124)
> 
> Do you agree?

@JeremyLoy ah thanks for the chart note.

What I think it means is that this PR would boil down to just the following:

1. Update https://github.com/openfga/openfga/blob/main/Makefile#L134
2. Update https://github.com/openfga/openfga/blob/main/docker-compose.yaml#L3
3. Update https://github.com/openfga/openfga/blob/main/pkg/testfixtures/storage/postgres.go#L25

JeremyLoy:
@jpadilla should be g2g now

jpadilla (approved):

jpadilla (approved):

PR #2408, opened 2025-04-30
fix: safeguard against nil and cancelled context in check_fast_path

Vic-Dev:
## Description

- Add safeguards around nil pointers
- Check if context Err value exists in `TrySendThroughChannel`
- Add comments in goroutines in `check_fast_path` that cannot easily include error handling for panics
  - When goroutines are spawned to handle draining channels, waiting on the channel to finish draining in order to handle potential panics may cause processes to slow down when there are no panics to handle.

## Review Checklist

### DIFF
--- a/internal/concurrency/concurrency.go
+++ b/internal/concurrency/concurrency.go
@@ -19,6 +19,9 @@ func NewPool(ctx context.Context, maxGoroutines int) *pool.ContextPool {
 // TrySendThroughChannel attempts to send an object through a channel.
 // If the context is canceled, it will not send the object.
 func TrySendThroughChannel[T any](ctx context.Context, msg T, channel chan<- T) bool {
+	if ctx.Err() != nil {
+		return false
+	}
 	select {
 	case <-ctx.Done():
 		return false
--- a/internal/graph/check_fast_path.go
+++ b/internal/graph/check_fast_path.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/emirpasic/gods/sets/hashset"
+	"github.com/sourcegraph/conc/panics"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 
@@ -392,7 +393,15 @@ func fastPathOperationSetup(ctx context.Context, req *ResolveCheckRequest, op se
 		return nil, ErrUnknownSetOperator
 	}
 	outChan := make(chan *iterator.Msg, len(children))
-	go resolver(ctx, iterator.NewStreams(iterStreams), outChan)
+	go func() {
+		recoverredError := panics.Try(func() {
+			resolver(ctx, iterator.NewStreams(iterStreams), outChan)
+		})
+
+		if recoverredError != nil {
+			concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Err: recoverredError.AsError()}, outChan)
+		}
+	}()
 	return outChan, nil
 }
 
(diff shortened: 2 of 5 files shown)


internal/graph/check_fast_path.go
```diff
@@ -629,14 +632,18 @@ func fanInIteratorChannels(ctx context.Context, chans []chan *iterator.Msg) chan
...
+				if v != nil && !concurrency.TrySendThroughChannel(ctx, v, out) {
 					if v.Iter != nil {
 						v.Iter.Stop()
 					}
 				}
 			}
+
 			return nil
 		})
 	}
```
Vic-Dev, line 649: There's currently no way I can see to handle this goroutine panicking, since the only way for the above `pool.Go` to panic would be if the `!concurrency.TrySendThroughChannel(...)` call is true which would require the ctx to be cancelled.

internal/graph/check_fast_path.go
```diff
@@ -444,6 +444,8 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
 		if !leftOpen {
 			return
 		}
+
+		// Capturing a panic here would cause us to wait for the channel to be drained, slowing the overall function down.
 		go drainIteratorChannel(leftChan)
```
Vic-Dev, line 449: Adding error handling for panics around these drain functions might cause overall performance to go down since we would have to wait for these channels to drain in order to return an error.

internal/graph/check_fast_path.go
```diff
@@ -444,7 +446,20 @@ func (c *LocalChecker) resolveFastPath(ctx context.Context, leftChans []chan *it
...
-		go drainIteratorChannel(leftChan)
+
+		go func() {
+			// Returning an error on a panic here would cause us to wait for the channel to be drained,
+			// slowing the overall function down.
+			defer func() {
+				if r := recover(); r != nil {
+					c.logger.ErrorWithContext(ctx, "panic in resolveFastPath",
+						zap.Any("error", r),
+					)
```
senojj, line 457:
```suggestion
					c.logger.ErrorWithContext(ctx, "panic recovered",
						zap.Any("error", r),
						zap.String("function", "resolveFastPath"),
					)
```

internal/concurrency/concurrency.go
```diff
@@ -19,6 +19,9 @@ func NewPool(ctx context.Context, maxGoroutines int) *pool.ContextPool {
 // TrySendThroughChannel attempts to send an object through a channel.
 // If the context is canceled, it will not send the object.
 func TrySendThroughChannel[T any](ctx context.Context, msg T, channel chan<- T) bool {
+	if ctx.Err() != nil {
```
elbuo8, line 22: 🤔 `ctx.Done` would do the same

internal/graph/check_fast_path.go
```diff
@@ -529,6 +544,7 @@ ConsumerLoop:
 }
 
 func constructLeftChannels(ctx context.Context,
+	logger logger.Logger,
```
elbuo8, line 547: Perhaps bind the function to the struct so we don't have to pass this explicitly?

internal/graph/check_fast_path.go
```diff
@@ -629,14 +657,18 @@ func fanInIteratorChannels(ctx context.Context, chans []chan *iterator.Msg) chan
 	pool := concurrency.NewPool(ctx, limit)
 
 	for _, c := range chans {
+		if c == nil {
```
elbuo8, line 660: Guessing this is just a future/potential guard?

PR #2409, opened 2025-05-01
fix: current_iterator_cache_count account for iter stop more than once

adriantam:
## Description
Update logic for `current_iterator_cache_count` so that it handles case where iterator stop more than once correctly.  Otherwise, the gauge counter can go into negative when the same iterator is stopped more than once.

## Review Checklist

### DIFF
--- a/pkg/storage/storagewrappers/cached_datastore.go
+++ b/pkg/storage/storagewrappers/cached_datastore.go
@@ -56,11 +56,11 @@ var (
 		NativeHistogramMinResetDuration: time.Hour,
 	}, []string{"operation", "method"})
 
-	currentIteratorCacheCount = promauto.NewGauge(prometheus.GaugeOpts{
+	currentIteratorCacheCount = promauto.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: build.ProjectName,
 		Name:      "current_iterator_cache_count",
 		Help:      "The current number of items of cache iterator.",
-	})
+	}, []string{"cached"})
 )
 
 // iterFunc is a function closure that returns an iterator
@@ -367,7 +367,7 @@ func (c *CachedDatastore) newCachedIterator(
 		span.SetAttributes(attribute.Bool("cached", true))
 
 		staticIter := storage.NewStaticIterator[*storage.TupleRecord](cacheEntry.Tuples)
-		currentIteratorCacheCount.Inc()
+		currentIteratorCacheCount.WithLabelValues("true").Inc()
 
 		return &cachedTupleIterator{
 			objectID:   objectID,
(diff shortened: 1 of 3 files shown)


pkg/storage/storagewrappers/cached_datastore.go
```diff
@@ -490,7 +491,13 @@ func (c *cachedIterator) Stop() {
...
 	defer c.mu.Unlock()
 
-	defer currentIteratorCacheCount.Dec()
+	newStop := !c.stopped
+	c.stopped = true
+	defer func() {
+		if newStop {
+			currentIteratorCacheCount.WithLabelValues("false").Dec()
+		}
+	}()
```
justincoh, line 500: why bother with the defer?

justincoh (commented):
Couple questions:
1. When are iterators stopped more than once?
  *  You can create a new Iterator and call Inc(). Then when that iterator calls `.Stop()` it will flush to cache and call `Dec()`.
  * When retrieving it from cache, we call `Inc()` again, and when that static iterator is stopped we `Dec()` again. 
2. If `.Stop()` is called more than once, could it be concurrent? Does `stopped` need to be atomic?

pkg/storage/storagewrappers/cached_datastore.go
adriantam: so that we don't need to add these logics to every path below.

justincoh: Oh I mean why not just do it serially? It's not an expensive operation
```
// without defer
if newStop {
  currentIteratorCacheCount.WithLabelValues("false").Dec()
}
```

adriantam: Theoretically, this should be done at the end of the function.  However, having a "defer" makes it more difficult to read and add complexity to the code.  I will remove the defer.

adriantam:
> Couple questions:
> 
> 1. When are iterators stopped more than once?
> 
> * You can create a new Iterator which calls `Inc()`. Then when that iterator calls `.Stop()` it will flush to cache and call `Dec()`.
> * When retrieving it from cache, we call `Inc()` again, and when that static iterator is stopped we `Dec()` again.
> * This seems right?
> 
> 2. If `.Stop()` is called more than once, could it be concurrent? Does `stopped` need to be atomic?

1. We probably have a few places in our code where we indiscriminately call Stop() as part of clean up.  For the scenario you described, we will not go into a negative.
2. It does not need to be concurrent/atomic.  Since stopped will not set back to "non-stopped", thus, there is no need to be atomic.

justincoh (approved):

PR #2410, opened 2025-05-01
feat: shared iterator skeleton

adriantam:
## Description
Added ability to track number of keys reused in shared iterator. This is not hooked up to be used yet - which will be covered in separate change. 

## Review Checklist


adriantam:
Note that the CHANGELOG will be updated when we have hooked up with config to be usable.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,339 @@
...
+// WithSharedIteratorDatastoreStorageLimit sets the limit on the number of items in SF iterator iters.
+func WithSharedIteratorDatastoreStorageLimit(limit int) DatastoreStorageOpt {
+	return func(b *Storage) {
+		b.limit = limit
+	}
+}
+
+func NewSharedIteratorDatastoreStorage(opts ...DatastoreStorageOpt) *Storage {
+	newStorage := &Storage{
+		mu:    sync.Mutex{},
```
senojj, line 68: fyi: the zero value of a `sync.Mutex` is useful; no need to initialize, so it can be omitted from the initialization of the `Storage` struct.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,339 @@
...
+	}, []string{"operation"})
+)
+
+const (
+	defaultSharedIteratorLimit = 1000000
+)
+
+type Storage struct {
+	mu    sync.Mutex
+	iters map[string]*internalSharedIterator // protected by mu
```
senojj, line 53: any reason that `internalSharedIterator` needs to be a pointer within the map?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,339 @@
...
+func (s *sharedIterator) Stop() {
+	// For now, we don't check whether this is the last item in the shared iterator
+	// as this is just a skeleton.  When the actual shared iterator is implemented,
+	// we will stop the inner iterator only if this is the last reference.
+	if !s.stopped {
+		// It is perfectly possible that iterator calling stop more than once.
+		// However, we only want to decrement the count on the first stop.
+		_, _ = s.manager.deref(s.key)
+	}
+	s.stopped = true
```
senojj, line 337: Consider making `s.stopped` an `atomic.Bool`:
```suggestion
	if !s.stopped.Swap(true) {
		// It is perfectly possible that iterator calling stop more than once.
		// However, we only want to decrement the count on the first stop.
		_, _ = s.manager.deref(s.key)
	}
```

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,339 @@
...
+	return false, nil
+}
+
+// sharedIterator will be shared with multiple consumers.
+// For now, this is a skeleton and no one is actually sharing the iterator.
+type sharedIterator struct {
+	manager *IteratorDatastore
+	key     string
+	inner   storage.TupleIterator
+	stopped bool
```
senojj, line 308: perhaps make this an `atomic.Bool` due to the potential for concurrent access.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,339 @@
...
+	start := time.Now()
+
+	cacheKey, err := storagewrappersutil.ReadStartingWithUserKey(store, filter)
+	if err != nil {
+		// should never happen
+		return nil, err
+	}
+	span.SetAttributes(attribute.String("cache_key", cacheKey))
+
+	sf.internalStorage.mu.Lock()
```
justincoh, line 132: Can we `defer Unlock()` immediately here, rather than multiple places below?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore_test.go
```diff
@@ -0,0 +1,648 @@
...
+		internalStorageLimit.mu.Unlock()
+
+		_, ok := iter.(*sharedIterator)
+		require.False(t, ok)
+
+		iter.Stop()
+	})
+}
+
+func TestSharedIteratorDatastore_ReadStartingWithUser(t *testing.T) {
```
justincoh, line 425: is it possible to combine some of these tests and/or their setup? There's a lot of shared boilerplate

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,339 @@
...
+var (
+	tracer = otel.Tracer("openfga/pkg/storagewrappers/sharediterator")
+
+	_ storage.RelationshipTupleReader = (*IteratorDatastore)(nil)
+	_ storage.TupleIterator           = (*sharedIterator)(nil)
+
+	sharedIteratorQueryHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:                       build.ProjectName,
+		Name:                            "shared_iterator_query_ms",
+		Help:                            "The duration (in ms) of a shared iterator query labeled by operation and shared.",
```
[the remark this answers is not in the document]
adriantam: I actually want this to be a histogram of time so that we can get both
- a count on shared vs. non-shared
- measurement on how long shared vs. non-shared take (this may/may not be similar as shared may be shorter because of cache but may also be longer due to connection contention). However, they will be different in the actual implementation because the shared entry will be taken from memory later on.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: No.  We need finer grained unlock because we want to unlock before calling `sf.RelationshipTupleReader.ReadStartingWithUser`.  This is intended so that we don't hold the map lock across datastore query to improve concurrency latency.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: more as convenient so that we call
```
item, found := storage.iters[key]
```
we can modify item directly instead of having to access the item via key again.

See https://go.dev/play/p/vYWEB2iGBww

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore_test.go
adriantam: Updated

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,337 @@
...
+		Buckets:                         []float64{1, 2, 5, 10, 25, 50, 80, 100, 150, 200, 300, 1000},
```
elbuo8, line 35: Remember, buckets == tag cardinality. 
We can survive with less granularity and still see the value.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,337 @@
...
+
+		sf.internalStorage.iters[cacheKey] = &internalSharedIterator{
+			counter: 1,
+		}
+		sf.internalStorage.mu.Unlock()
+		span.SetAttributes(attribute.Bool("found", false))
+	}
+	actual, err := sf.RelationshipTupleReader.ReadUsersetTuples(ctx, store, filter, options)
+	if err != nil {
+		go func() {
```
elbuo8, line 216: Why does this need to launch a goroutine for something that is relatively quick?

adriantam: removed go func

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -0,0 +1,331 @@
...
+	}, []string{"operation"})
+)
+
+const (
+	defaultSharedIteratorLimit = 1000000
+)
+
+type Storage struct {
+	mu    sync.Mutex
+	iters map[string]*internalSharedIterator // protected by mu
```
senojj, line 54: another thing to remember about `map` types is that they never shrink. so, if your keys are going to have high cardinality, the map will continue to grow over time, even if you delete keys.

adriantam: > if your keys are going to have high cardinality, the map will continue to grow over time, even if you delete keys.

If I understand [how map works](https://forum.golangbridge.org/t/will-golang-free-the-memory-when-i-use-delete-map-somekey/9081/6), it is true than it will not shrink.  However, the size of the map will be the high watermark on the number of keys in the map.  Therefore, I believe that the behaviour will be that the map will grow to the largest number of keys and then stop growing.

Given we have a [limit on the maximum number of keys we allowed](https://github.com/openfga/openfga/blob/722818a80e4d8c05cd5cf7aed381b10ffe34c176/pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go#L145) , I believe we are safe on this front.

senojj: You are correct in your understanding of how maps grow. The limit on keys should be sufficient to solve for my concern.

elbuo8 (approved):

PR #2415, opened 2025-05-05
Add Panic Handling to ListUsers

mkgobaco:
Add Panic Handling to ListUsers expandTTU

## Description

Use SourceGraph conc pkg/lib https://pkg.go.dev/github.com/sourcegraph/conc to catch panics and error gracefully.

## Review Checklist

### DIFF
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -153,6 +153,13 @@ func WithDispatchThrottlerConfig(config threshold.Config) ListUsersQueryOption {
 	}
 }
 
+// WithDispatchHandler allows setting a custom dispatch handler
+func WithDispatchHandler(handler dispatchHandler) ListUsersQueryOption {
+	return func(l *listUsersQuery) {
+		l.dispatchHandler = handler
+	}
+}
+
 // TODO accept ListUsersRequest instead of contextualTuples.
 func NewListUsersQuery(ds storage.RelationshipTupleReader, contextualTuples []*openfgav1.TupleKey, opts ...ListUsersQueryOption) *listUsersQuery {
 	l := &listUsersQuery{
--- a/pkg/server/commands/listusers/list_users_rpc.go
+++ b/pkg/server/commands/listusers/list_users_rpc.go
@@ -584,19 +584,19 @@ func (l *listUsersQuery) expandIntersection(
 	errChan := make(chan error, 1)
 
 	go func() {
-		var resp expandResponse
+		var err error
 		recoveredError := panics.Try(func() {
-			err := pool.Wait()
+			err = pool.Wait()
 			for i := range intersectionFoundUsersChans {
 				close(intersectionFoundUsersChans[i])
 			}
-			errChan <- err
-			close(errChan)
 		})
 		if recoveredError != nil {
-			resp = panicExpanseResponse(recoveredError)
-			errChan <- resp.err
+			resp := panicExpanseResponse(recoveredError)
+			err = resp.err
 		}
+		errChan <- err
+		close(errChan)
 	}()
 
 	var mu sync.Mutex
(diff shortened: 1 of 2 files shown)


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -902,10 +910,13 @@ LoopOnIterator:
 		userObjectType, userObjectID := tuple.SplitObject(userObject)
 
 		pool.Go(func(ctx context.Context) error {
-			rewrittenReq := req.clone()
-			rewrittenReq.Object = &openfgav1.Object{Type: userObjectType, Id: userObjectID}
-			rewrittenReq.Relation = computedRelation
-			resp := l.dispatch(ctx, rewrittenReq, foundUsersChan)
+			var resp expandResponse
+			recoveredError := panics.Try(func() {
+				resp = l.expandTTUDispatch(ctx, l, req, userObjectType, userObjectID, computedRelation, foundUsersChan)
```
justincoh, line 915: Why break this into a new method rather than just calling the existing code inside of `panics.Try`?

mkgobaco: @justincoh the reason I broke the code into a new "expandTTUDispatch " method is so I can use inject my a test version of the "expandTTUDispatch" that just panics.

Maybe there's a better way of doing it? :)

adriantam: @senojj - any thoughts here on what is the best practice?

senojj: I don't think this makes much sense as a function. First, the number of arguments creates a very large function interface, making the abstraction very weak; it doesn't really abstract what is going on. Second, why not just make the `dispatch` function injectable?

mkgobaco: @senojj I was trying to limit the scope of the changes by not changing dispatch().  But it looks like dispatch() is referred to in 3 locations only.

I will make dispatch() injectable. :)

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -506,7 +528,13 @@ LoopOnIterator:
 		pool.Go(func(ctx context.Context) error {
 			var resp expandResponse
 			recoveredError := panics.Try(func() {
-				resp = l.expandDirectDispatch(ctx, l, req, userObjectType, userObjectID, userRelation, resp, foundUsersChan, &hasCycle)
+				rewrittenReq := req.clone()
```
mkgobaco, line 531: Reverted this code back to not use dependency injection.

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -525,17 +553,6 @@ LoopOnIterator:
 	}
 }
 
-func expandDirectDispatch(ctx context.Context, l *listUsersQuery, req *internalListUsersRequest, userObjectType string, userObjectID string, userRelation string, resp expandResponse, foundUsersChan chan<- foundUser, hasCycle *atomic.Bool) expandResponse {
```
mkgobaco, line 528: Removed unneeded function used for dependency injection.

pkg/server/commands/listusers/list_users_rpc_test.go
```diff
@@ -4115,7 +4088,7 @@ func NewListUsersQueryPanicExpandDirect(ds storage.RelationshipTupleReader, cont
 		maxResults:              serverconfig.DefaultListUsersMaxResults,
 		maxConcurrentReads:      serverconfig.DefaultMaxConcurrentReadsForListUsers,
 		wasThrottled:            new(atomic.Bool),
-		expandDirectDispatch: func(ctx context.Context, listUsersQuery *listUsersQuery, req *internalListUsersRequest, userObjectType, userObjectID, userRelation string, resp expandResponse, foundUsersChan chan<- foundUser, hasCycle *atomic.Bool) expandResponse {
```
mkgobaco, line 4118: removed unneeded function field used for dependency injection

PR #2416, opened 2025-05-05
Add Panic Handling to ListUsers expandUnion

mkgobaco:
Add Panic Handling to ListUsers expandUnion

## Description

Use SourceGraph conc pkg/lib https://pkg.go.dev/github.com/sourcegraph/conc to catch panics and error gracefully.

## Review Checklist


pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -638,18 +650,21 @@ func (l *listUsersQuery) expandUnion(
 		rewrite := rewrite
 		unionFoundUsersChans[i] = make(chan foundUser, 1)
 		pool.Go(func(ctx context.Context) error {
-			resp := l.expandRewrite(ctx, req, rewrite, unionFoundUsersChans[i])
+			var resp expandResponse
+			recoveredError := panics.Try(func() {
+				resp = l.expandUnionExpandRewrite(ctx, l, req, rewrite, unionFoundUsersChans, i)
```
adriantam, line 655: why do we need expandUnionExpandRewrite vs. expandRewrite?

Couldn't we accomplish the same thing via 
```
recoveredError := panics.Try(func() {
				resp = l.expandRewrite(ctx, req, rewrite, unionFoundUsersChans, unionFoundUsersChans[i])
})
```
?

If it is because we need the functional signature for unit test?  Maybe change the function signature for expandRewrite?

pkg/server/commands/listusers/list_users_rpc.go
```diff
@@ -638,18 +650,21 @@ func (l *listUsersQuery) expandUnion(
...
 	}
 
 	errChan := make(chan error, 1)
 
 	go func() {
-		err := pool.Wait()
-		for i := range unionFoundUsersChans {
-			close(unionFoundUsersChans[i])
-		}
+		err := l.expandUnionCloseChannels(pool, unionFoundUsersChans)
```
adriantam, line 667: Just to confirm - I don't see that we are protecting from panic here

mkgobaco: @adriantam yes, there's a follow-up work for this code section.

pkg/server/commands/listusers/list_users_rpc.go
mkgobaco: @adriantam Yes, so we can inject our own function for unit tests.

I could make expandRewrite injectable but it would affect more code.

justincoh, line 655: This looks like a noop as written here. This line is calling a function `l.expandUnionExpandRewrite` but I don't ever see that method defined with a pointer receiver?

Issue #2422, opened 2025-05-05
Getting deadline_exceeded response and running out of db conns


frisket:
### Checklist

- [x] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [x] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

We've run into a problem where all fga requests fail with deadline_exceeded error response, including even the health check!

We've done some experimentation and we've found the folllowing:

- When this happens the database connections are consumed up to the max-pool size via OPENFGA_DATASTORE_MAX_OPEN_CONNS (in our case, the default 30)
- The db conns never shrink back to the min 10 (OPENFGA_DATASTORE_MAX_IDLE_CONNS) even when idle
- Versions up to 1.8.6 work correctly
- Version 1.8.7 and higher all fail basically immediately within a small number of check requests - 30-40
- On versions up to 1.8.6 the db connections are all cleared/freed up and go back to the min (10) as expected after the requests
- Once it reaches this state the service is broken and no further requests are handled.  Only a restart will fix it.

The actual response we get for every API request after this happens is:

```
{"code":"deadline_exceeded","message":"Request Deadline Exceeded"}
```

We don't have many tuples - only ~35k and the model is relatively simple.  I don't believe it is model or data related as everything works unchanged up to version 1.8.6 without issue.

Not sure where to go from here as the 1.8.7 changelogs don't really give any clues and the git diffs aren't too enlightening for me as I'm not familiar with the code base.

We've tried enabling additional logging with '--log-level=debug' but that doesn't seem to provide any data or any additional data at all.

Any suggestions or pointers to resolving this issue would be appreciated.  For now we're sticking with 1.8.6. as it works.

### OpenFGA version

1.8.7->1.8.11

### How are you running OpenFGA?

In Docker

### What datastore are you using?

Postgres

### OpenFGA Flags

[...]

PR #2425, opened 2025-05-07
fix: stream fastpath, move sql to lazy queries

elbuo8:
## Description

1. Fix cases where Iterators were not being cleaned up (thus not releasing connections or triggering the iterator cache etc) via abstraction (and missing explicit calls).
2. Move SQL that yielded iterators to async. Otherwise, if you need to create 100 iterators, but can only process 5 at the same time, you'd lost 95 connections of your pool because these wouldn't be released until Stop would be called on the iterator. Moving to async keeps the connection in the pool up until the moment it will actually be used for the first time.
3. FastPath was being calculated in batches, as in it would collect all the channels and then process them. Now it will produce and consume uniformly. 
4. FastPath was not being properly bounded because while FanInIterator would have a consumption limit, that channels were already created, this translated to queries being executed without any limit (this was specially bad in Recursion/BFS where an unbounded a list of entries would be pulled in) (this wasn't always like this, was a regression).
5. Change the default value of Concurrency (Resolve Max Breadth) from 100 to 10. 100 means that up to 100 connections would be used to evaluate edges of a node. If the default for drivers is 30 connections (and that itself is high) this could also result in a deadlock situation (the other fixes we introduced here mitigate that also). But performance wise it's significantly better to re-use existing connections that waiting to establish a new one.

## References
fix https://github.com/openfga/openfga/issues/2422 

## Review Checklist


internal/graph/object_providers_test.go
```diff
@@ -119,10 +119,6 @@ func TestRecursiveObjectProvider(t *testing.T) {
 }
 
 func TestRecursiveTTUObjectProvider(t *testing.T) {
```
adriantam, line 121: why do we remove the leak check?  If it is because of the fan_in iterator, can we add them as [IgnoreTopFunction](https://pkg.go.dev/go.uber.org/goleak#IgnoreTopFunction)?

adriantam (commented):
Do we want to split the SQL changes into separate PR?

internal/iterator/fan_in.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/openfga/openfga/internal/concurrency"
+)
+
+type FanIn struct {
+	ctx       context.Context
+	cancel    context.CancelFunc
+	count     int
+	out       chan *Msg
+	addCh     chan (chan *Msg)
+	mu        *sync.Mutex
```
senojj, line 18: One question and one suggestion...

Question: Since `FanIn` is only ever used as a pointer, any reason why `mu` needs to be a pointer? Having `mu` as a pointer guarantees that it will escape to the heap, but if it were a value type, it would follow wherever `FanIn` is allocated.

Suggestion: If `count` and `accepting` are made atomic types, or handled with atomic functions, you could remove `mu` entirely.

This comment is not blocking.

elbuo8: Removed pointer to `mu`.

Originally started with atomics but found myself moving to `mu` after all the back and forth.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -728,3 +765,31 @@ func AddFromUlid(sb sq.SelectBuilder, fromUlid string, sortDescending bool) sq.S
...
 }
+
+// HandleSQLError processes an SQL error and converts it into a more
+// specific error type based on the nature of the SQL error.
+func HandleSQLError(err error, args ...interface{}) error {
+	if errors.Is(err, sql.ErrNoRows) {
+		return storage.ErrNotFound
+	}
+
+	if me, ok := err.(*mysql.MySQLError); ok && me.Number == 1062 {
```
adriantam, line 776: should we combine the logic with line 785-792?

elbuo8: You could, when you want to handle another one it will get messy. Better for readability to keep them separated.

internal/graph/check_fast_path.go
```diff
@@ -553,15 +548,14 @@ func constructLeftChannels(
...
 		leftChan, err := fastPathRewrite(ctx, r, rel.GetRewrite())
 		if err != nil {
-			// if the resolver already started it needs to be drained
-			if len(leftChans) > 0 {
-				go drainIteratorChannel(fanInIteratorChannels(ctx, leftChans, concurrencyLimit))
-			}
-			return nil, err
+			errCh := make(chan *iterator.Msg, 1)
+			errCh <- &iterator.Msg{Err: err}
+			close(errCh)
```
adriantam, line 553: want to verify the logic here.  We have close the channel `errCh` but we still add the channel `err` to `leftChans`?  But since the channel is already closed, TrySendThroughChannel and drainOnExit for the channel will do nothing..

elbuo8: It will actually be handled by the consumer. We close it so that the consumer knows there will be no more messages in the channel and can move on to the next one.

yissellokta:
Claude code review summary: 
  
Main Changes

  1. New Iterator Mechanism
    - Created a fan-in pattern (iterator/fan_in.go) that efficiently processes multiple iterators in parallel
    - Moved and improved the stream handling from internal/graph/iterator to internal/iterator
    - Added proper resource cleanup with drainOnExit to prevent leaks
  2. Fast Path Optimization
    - Significantly refactored check_fast_path.go to improve performance
    - Implemented specialized handlers for different set operations (union, intersection, difference)
    - Added bounded concurrency controls to prevent resource exhaustion
  3. Race Condition Fixes
    - Fixed a critical race condition in breadthFirstRecursiveMatch by cloning the request before passing to goroutine
    - Added proper mutex protections around shared resources
  4. SQL Optimizations
    - Enhanced MySQL and PostgreSQL implementations for better performance
    - Improved SQL common utilities for more efficient database access

  Code Quality

  1. Test Coverage
    - Added tests for new fan-in iterator implementation
    - Tests verify no goroutine leaks using goleak
    - Tests confirm proper error handling
  2. Error Handling
    - Added panic recovery in critical paths
    - Improved error propagation through channels
    - Fixed multiple places where errors weren't properly handled
  3. Resource Management
    - Consistent use of proper cleanup patterns with defer
    - Added explicit iterator stopping to prevent resource leaks
    - Implemented context cancellation propagation

  Potential Issues

  1. The branch contains several commits with vague messages like "fix" and "fixes" without clear descriptions
  2. Some new code paths may benefit from additional test coverage, especially around complex error scenarios
  3. The complexity in the recursive matching algorithms could make future maintenance challenging

  Conclusion

  The bound-fastpath branch introduces significant performance optimizations for authorization checks with a focus on concurrency control and resource management. The core changes involve a new fan-in iterator pattern and refactored fast path implementation. The changes appear well-tested and address important race conditions.

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -728,3 +765,33 @@ func AddFromUlid(sb sq.SelectBuilder, fromUlid string, sortDescending bool) sq.S
...
+	// handle psql
+	if strings.Contains(err.Error(), "duplicate key value") {
+		if len(args) > 0 {
+			if tk, ok := args[0].(*openfgav1.TupleKey); ok {
+				return storage.InvalidWriteInputError(tk, openfgav1.TupleOperation_TUPLE_OPERATION_WRITE)
+			}
+		}
+		return storage.ErrCollision
+	}
+
```
DanCech, line 795: This was separated in #1615 so that we keep database-specific code in the individual storage plugins, is there a reason it needs to be combined together here again?

elbuo8: The query execution has been moved outside of the specific DB driver into the SQLcommon directory. Thus these error handlers need to be generic, not just pushing the errors up as they were originally.

adriantam (dismissed):

internal/iterator/fan_in_test.go
```diff
@@ -0,0 +1,71 @@
...
+		require.NoError(t, err)
+		i, err := strconv.Atoi(id)
+		require.NoError(t, err)
+		require.Positive(t, i)
+		require.LessOrEqual(t, i, 7)
+		_, err = msg.Iter.Next(ctx)
+		require.Equal(t, storage.ErrIteratorDone, err)
+		msg.Iter.Stop()
+		iterations++
+		if iterations > 2 {
```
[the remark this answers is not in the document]
elbuo8: Because I want to ensure the other mechanisms to drain also kick in. Specifically when calling `Stop`

adriantam (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ## [Unreleased]
+### Changed
+- `DefaultResolveNodeBreadthLimit` changed from 100 to 10 in order to reduce connection contention. [#2425](https://github.com/openfga/openfga/pull/2425)
+- PostgreSQL and MySQL based iterators will load tuples only when needed (lazy loading). [#2425](https://github.com/openfga/openfga/pull/2425)
+
 ### Fixed
 - Replace hardcoded Prometheus datasource UID (`PBFA97CFB590B2093`) with `${DS_PROMETHEUS}` in `telemetry/grafana/dashboards/openfga.json`. This allows the Grafana dashboard to correctly reference the dynamic Prometheus datasource, resolving issues with improper binding. [#2287](https://github.com/openfga/openfga/issues/2287)
 - Handle case where iterator is stopped more than once for `current_iterator_cache_count`. [#2409](https://github.com/openfga/openfga/pull/2409)
```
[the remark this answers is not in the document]
elbuo8: Added.

pkg/storage/sqlcommon/sqlcommon.go
DanCech: There are numerous ways of solving this that do not rely on moving database-specific code into sqlcommon and reducing the pluggability of the individual stores.

Unilaterally resolving the comment and pushing the PR through without further discussion is a surprise to me, and I would very much like to see this feedback addressed in a follow-up PR to revert these changes and maintain the separation between sqlcommon and the database-specific code.

PR #2428, opened 2025-05-08
feat: add recover log in cached iterator stop goroutine

Vic-Dev:
## Description
Add recover and log in cached iterator stop goroutine. Do not actually return an error in this case, since if a panic happens here it probably means there's a problem with state.

## Review Checklist


pkg/storage/storagewrappers/cached_datastore_test.go
```diff
@@ -1451,6 +1452,52 @@ func TestCachedIterator(t *testing.T) {
...
+		}
+
+		iter.tuples = []*openfgav1.Tuple{{}}
+		iter.iter = &mockCalledTupleIterator{
+			iter: mockIterator,
+		}
+
+		iter.Stop()
+		wg.Wait()
+		time.Sleep(1500 * time.Millisecond) // wait for the goroutine to finish
```
jpadilla, line 1499: ah this is less to do with the goroutine and more so about the defer stack

1. c.iter.Stop
2. c.wg.Done
3. recover

so after `wg.Wait()`, we still probably haven't executed the first function call that has the `recover` bits

jpadilla (approved):

PR #2432, opened 2025-05-09
fix: panic handling in shadow resolver

Vic-Dev:
## Description
Recover as the first defer in shadow resolver with nothing else, and log. Do not actually return an error in this case, since if a panic happens here it probably means there's a problem with state.

## Review Checklist


internal/graph/shadow_resolver.go
```diff
@@ -67,20 +67,22 @@ func (s ShadowResolver) ResolveCheck(ctx context.Context, req *ResolveCheckReque
 		reqClone.VisitedPaths = nil // reset completely for evaluation
 		s.wg.Add(1)
 		go func() {
-			ctx, cancel := context.WithTimeout(ctxClone, s.shadowTimeout)
 			defer func() {
```
Vic-Dev, line 70: A defer function with a recover should not include any additional code that may panic because it cannot capture it. It should also always be the first defer function, so that it is the last to run.

internal/graph/shadow_resolver.go
```diff
@@ -67,20 +67,22 @@ func (s ShadowResolver) ResolveCheck(ctx context.Context, req *ResolveCheckReque
 		reqClone.VisitedPaths = nil // reset completely for evaluation
 		s.wg.Add(1)
 		go func() {
-			ctx, cancel := context.WithTimeout(ctxClone, s.shadowTimeout)
```
justincoh, line 70: Are you sure about removing this? It's outside the defer and it looks like it's designed to kill long-running shadow resolvers to ensure they don't eat up resources

Vic-Dev: It's not removed, it's moved down to lines 84-85. Basically I restructured this function to be more in line with the functionality from before, but also to ensure that all pathways are captured by the recover. See this commit before the recover was added:
https://github.com/openfga/openfga/blame/63e4e6350ea1b4ed3f95264a12a9da6171b4a1e0/internal/graph/shadow_resolver.go

justincoh (approved):

PR #2433, opened 2025-05-12
feat: shared iterator implementation

adriantam:
## Description
Shared iterator allows an iterator to have shared consumer.  This optimizes performance by reducing concurrent query for the same key - thus reducing datastore connection contention.   

## References
Continuation of https://github.com/openfga/openfga/pull/2410

## Review Checklist


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Added
+- Performance optimization by allowing datastore query iterator to have shared. This can be enabled via `OPENFGA_SHARED_ITERATOR_ENABLED`. [#2433](https://github.com/openfga/openfga/pull/2433), [#2410](https://github.com/openfga/openfga/pull/2410) and [#2423](https://github.com/openfga/openfga/pull/2423)
```
elbuo8, line 11: `to have shared`?

You mean allow datastore query iterator to be deduplicated maybe?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -135,36 +148,55 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
-		// is fully implemented, the found path will look up the shared iterator and create
-		// a clone from it.
-	} else {
-		if len(sf.internalStorage.iters) >= sf.internalStorage.limit {
-			sf.internalStorage.mu.Unlock()
-			sharedIteratorBypassed.WithLabelValues(storagewrappersutil.OperationReadStartingWithUser).Inc()
-			// we cannot share this iterator because we have reached the size limit.
-			return sf.RelationshipTupleReader.ReadStartingWithUser(ctx, store, filter, options)
-		}
+		sharedIteratorQueryHistogram.WithLabelValues(
```
elbuo8, line 156: This should be moved right before `return` because there is a case in which you don't actually get to use it.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -135,36 +148,55 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
 
 	if found {
 		keyItem.counter++
+		keyItem.iter.mu.Lock()
```
elbuo8, line 151: Shouldn't `keyItem` hold the `mu` so you can also increase `.counter` safely?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -135,36 +148,55 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
+		defer keyItem.iter.mu.Unlock()
```
elbuo8, line 152: > sf.internalStorage.mu.Unlock()

Should be the first line in the `if found {` block. Marginal, but if the same `keyItem` is under contention, then the whole thing would be slowed down.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -135,36 +148,55 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
-		// a clone from it.
-	} else {
-		if len(sf.internalStorage.iters) >= sf.internalStorage.limit {
-			sf.internalStorage.mu.Unlock()
-			sharedIteratorBypassed.WithLabelValues(storagewrappersutil.OperationReadStartingWithUser).Inc()
-			// we cannot share this iterator because we have reached the size limit.
-			return sf.RelationshipTupleReader.ReadStartingWithUser(ctx, store, filter, options)
-		}
+		sharedIteratorQueryHistogram.WithLabelValues(
+			storagewrappersutil.OperationReadStartingWithUser, strconv.FormatBool(found),
```
elbuo8, line 157: > strconv.FormatBool(found)

Seems like wasted operations, "true" (or better yet a const) would be enough?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -135,36 +148,55 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
+		).Observe(float64(time.Since(start).Milliseconds()))
 
-		sf.internalStorage.iters[cacheKey] = &internalSharedIterator{
-			counter: 1,
+		item, err := keyItem.iter.clone()
+		if err != nil {
+			go func() {
```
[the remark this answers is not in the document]
adriantam: Will remove the goroutine

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -135,36 +148,55 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
+	sf.internalStorage.mu.Unlock()
+	span.SetAttributes(attribute.Bool("found", false))
+
 	actual, err := sf.RelationshipTupleReader.ReadStartingWithUser(ctx, store, filter, options)
 	if err != nil {
+		newIterator.queryErr = err
 		_, _ = sf.deref(cacheKey)
 		return nil, err
 	}
+	newIterator.inner = actual
```
elbuo8, line 194: So if 2 queries for the same iterator come at the same time:
1. Whichever grabs the lock first will generate the entry into `internalStorage`.
2. It will yield the lock allowing the second to go into `if found` and yield that.
3. Then the first one which grabbed the lock will attempt to build _actual_, but if it fails, there will be no `inner`.

If I'm not wrong, it will blow up in that case or am I missing something there?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: keyItem.counter is protected by sf.internalStorage.mu.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
...
+	if s.queryErr != nil {
+		return nil, s.queryErr
+	}
+	newIter := &sharedIterator{
+		manager:      s.manager,
+		key:          s.key,
+		head:         0,
+		maxAliveTime: s.maxAliveTime,
+
+		mu:        s.mu,
```
elbuo8, line 407: 🤔 why do they share `mu`?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: I will move this up to be before keyItem.iter.mu.Lock().  However, this needs to be after keyItem.counter++ so that we won't prematurely delete the key.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -43,10 +44,13 @@ var (
...
 		Help:      "Total number of iterators bypassed by the shared iterator layer because the internal map size exceed specified limit.",
 	}, []string{"operation"})
+
+	errSharedIteratorWatchdog = errors.New("shared iterator watchdog timeout")
 )
 
 const (
-	defaultSharedIteratorLimit = 1000000
+	defaultSharedIteratorLimit        = 1000000
+	defaultSharedIteratorMaxAliveTime = 4 * time.Minute
```
elbuo8, line 53: Why 4 mins?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
...
+		// If there is an error, no need to get any more items, and we just return the error
+		if s.sharedErr != nil && *s.sharedErr != nil {
+			span.SetAttributes(attribute.String("existingError", (*s.sharedErr).Error()))
+			return nil, *s.sharedErr
+		}
+
+		span.SetAttributes(attribute.Bool("newItem", true))
+		item, err := s.inner.Next(ctx)
+		if err != nil {
+			span.SetAttributes(attribute.String("newError", err.Error()))
```
elbuo8, line 459: `telemetry.TraceError`

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
...
-	// as this is just a skeleton.  When the actual shared iterator is implemented,
-	// we will stop the inner iterator only if this is the last reference.
-	if !s.stopped.Swap(true) {
+	s.mu.Lock()
+	if s.watchdogTimer != nil {
+		s.watchdogTimer.Stop()
+	}
+	s.mu.Unlock()
+
+	if s.stopped.Swap(true) {
```
elbuo8, line 526: If you already have a `mu` I don't particularly see why this has to be an atomic 🤷🏽

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: I will need to move sf.internalStorage.mu.Unlock() to be after the query

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
...
 		// It is perfectly possible that iterator calling stop more than once.
 		// However, we only want to decrement the count on the first stop.
-		_, _ = s.manager.deref(s.key)
+		return
+	}
+	lastItem, _ := s.manager.deref(s.key)
+	if lastItem {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		s.inner.Stop()
```
elbuo8, line 535: Perhaps, the manager should handle doing this. Otherwise it's possible it never happens.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
...
+		key:          s.key,
+		head:         0,
+		maxAliveTime: s.maxAliveTime,
+
+		mu:        s.mu,
+		items:     s.items,
+		inner:     s.inner,
+		sharedErr: s.sharedErr,
+	}
+	newIter.watchdogTimer = time.AfterFunc(s.maxAliveTime, newIter.watchdogTimeout)
```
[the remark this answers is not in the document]
adriantam: Yes, each individual iterator should cleanup after itself.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: yes, with mu, we can change this to non-atomic.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
elbuo8: My worry is that this will have a goroutine (per each of the entries of the map rather than 1 since my last review) for up to 4 mins. Multiplying that by the size and a potential leak and that would be quite bad, no?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
elbuo8: But their `head` is different 🤔

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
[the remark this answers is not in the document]
elbuo8: Ok. Why if they share everything, can't they share the watchdog of the very first initialization? Ultimately it will close for all of them, no?

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
 // sharedIterator will be shared with multiple consumers.
 // For now, this is a skeleton and no one is actually sharing the iterator.
```
justincoh, line 364: can kill this comment now

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -296,36 +363,175 @@ func (sf *IteratorDatastore) deref(cacheKey string) (bool, error) {
...
+}
+
+// when watchdogTimeout is invoked, it will revoke the current iterator from usage.
+// This will mark the iterator as error and stop on its behalf
+// All further usage of Head/Next() will return error.
+func (s *sharedIterator) watchdogTimeout() {
+	s.mu.Lock()
+	s.watchdogTimeoutErr = errSharedIteratorWatchdog
+	s.mu.Unlock()
+	s.Stop()
```
justincoh, line 423: Do you think it's worth tracking how often the watchdog triggers? Ideally this timeout would be as quick as possible  without interfering with the benefits of sharing, but we don't currently have a good sense for how long these might live

adriantam: added counter to track this.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: I see what you are implying - in this case, I will update it so that it is not using the shared iterator at all.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: Reduced to 1 minute + request timeout

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
+	return item, nil
 }
 
 func (s *sharedIterator) Stop() {
-	// For now, we don't check whether this is the last item in the shared iterator
-	// as this is just a skeleton.  When the actual shared iterator is implemented,
-	// we will stop the inner iterator only if this is the last reference.
-	if !s.stopped.Swap(true) {
+	s.mu.Lock()
+	if s.stopped {
```
adriantam, line 582: Need to at least reset the timer here.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
+
+// when watchdogTimeout is invoked, it will revoke the current iterator from usage.
+// This will mark the iterator as error and stop on its behalf
+// All further usage of Head/Next() will return error.
+func (s *sharedIterator) watchdogTimeout() {
+	s.mu.Lock()
+	*s.watchdogTimeoutErr = errSharedIteratorWatchdog
+	s.mu.Unlock()
+	sharedIteratorWatchDog.Inc()
+	s.manager.watchdogTimeout(s.key)
```
adriantam, line 483: Consider the following case:
```
watchdog timeout scheduled
deref last item.  Which cause the manger to remove the item
Watchdog run
```

In this case, it is possible to have two scenarios:
1. No entry is created.  In this case, it is ok because when we deref the last item, we would have all the associated iter / etc.
2. New entry is created between deref last item &&  - this is bad as we have prematurely remove the entry and associated info.  

Consider adding some form of unique identifier to avoid case #2

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
+
+func newSharedIterator(manager *IteratorDatastore, key string, maxAliveTime time.Duration) *sharedIterator {
+	newIter := &sharedIterator{
+		manager:      manager,
+		key:          key,
+		head:         0,
+		maxAliveTime: maxAliveTime,
+
+		mu:                 &sync.Mutex{},
+		items:              new([]*openfgav1.Tuple),
```
adriantam, line 435: alloc more spaces in advance

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -84,22 +92,31 @@ func WithSharedIteratorDatastoreLogger(logger logger.Logger) IteratorDatastoreOp
...
-	internalStorage *Storage
+	logger                logger.Logger
+	internalStorage       *Storage
+	watchdogTimeoutConfig time.Duration
 }
 
 type internalSharedIterator struct {
-	counter uint64
-	// in the future, there will be reference to the shared iterator
+	counter uint64 // protected by the storage mu
```
elbuo8, line 110:
```suggestion
	counter uint64 // indirectly protected by the surrounding mutex. Therefore it's usage must be made with caution.```

File: pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go [hunk 20]
```diff
@@ -136,35 +154,63 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
-		if len(sf.internalStorage.iters) >= sf.internalStorage.limit {
-			sf.internalStorage.mu.Unlock()
-			sharedIteratorBypassed.WithLabelValues(storagewrappersutil.OperationReadStartingWithUser).Inc()
-			// we cannot share this iterator because we have reached the size limit.
+		// by the time we have access to keyItem.iter.mu, we know that
+		// the inner is already set OR the queryErr is set (for which clone
+		// will fail).
+		item, err := keyItem.iter.clone()
+		if err != nil {
+			keyItem.iter.mu.Unlock()
```
elbuo8, line 166: This can be moved on stop of this `if` block and used with a `defer`.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -136,35 +154,63 @@ func (sf *IteratorDatastore) ReadStartingWithUser(
...
+		iter:    newIterator,
+	}
+	sf.internalStorage.mu.Unlock()
+	span.SetAttributes(attribute.Bool("found", false))
+
 	actual, err := sf.RelationshipTupleReader.ReadStartingWithUser(ctx, store, filter, options)
 	if err != nil {
-		_, _ = sf.deref(cacheKey)
+		newIterator.queryErr = err
+		newIterator.mu.Unlock()
```
elbuo8, line 199: You can `defer` this one.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
 		sf.logger.Error("failed to dereference cache key", zap.String("cacheKey", cacheKey))
-		return false, fmt.Errorf("failed to dereference cache key: %s", cacheKey)
+		return
 	}
 	item.counter--
 	if item.counter == 0 {
 		delete(sf.internalStorage.iters, cacheKey)
-		return true, nil
+		sf.internalStorage.mu.Unlock()
+		item.iter.cleanup()
```
elbuo8, line 390: To clean this up you can queue up `defer` statements so guaranteeing that this operation happens after the yield of `sf.internalStorage.mu`.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
+func (sf *IteratorDatastore) watchdogTimeout(cacheKey string) {
+	sf.internalStorage.mu.Lock()
+	item, ok := sf.internalStorage.iters[cacheKey]
+	if !ok {
+		sf.internalStorage.mu.Unlock()
+		sf.logger.Error("failed to watchdog key", zap.String("cacheKey", cacheKey))
+		return
 	}
-	return false, nil
+	item.counter = 0
```
elbuo8, line 404: Not needed.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
 type sharedIterator struct {
-	manager *IteratorDatastore
-	key     string
-	inner   storage.TupleIterator
-	stopped atomic.Bool
+	manager      *IteratorDatastore // non-changing
+	key          string             // non-changing
+	head         int
+	maxAliveTime time.Duration
+	queryErr     error
```
elbuo8, line 416: `initializationErr`

elbuo8: // it is shared, but only inspected on `.clone`

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -242,90 +317,276 @@ func (sf *IteratorDatastore) Read(
...
+		head:         0,
+		maxAliveTime: maxAliveTime,
+
+		mu:                 &sync.Mutex{},
+		items:              new([]*openfgav1.Tuple),
+		inner:              nil,
+		sharedErr:          new(error),
+		watchdogTimeoutErr: new(error),
+	}
+	newIter.watchdogTimer = time.AfterFunc(maxAliveTime, newIter.watchdogTimeout)
```
elbuo8, line 440: Document how the watchDog lifecycle works.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: actually, I want this to unlock before calling `return sf.RelationshipTupleReader.ReadStartingWithUser()`.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
adriantam: As noted above, the defer is first in last out.  As such, we cannot use defer because we need to unlock before cleanup.

elbuo8 (commented):
Minor comments

elbuo8 (approved):

PR #2434, opened 2025-05-12
feat: make sqlite lazy

elbuo8:
## Description

Makes SQLite iterators lazy and reverts changes to HandleSQLError in the other SQL packages.

## Review Checklist


DanCech (dismissed):
This looks great, thanks for refactoring the error handler code back out. At this point the only real difference between the sqlcommon and sqlite tuple iterators is the way the rows are scanned, except that we need to apply the t.HandleSQLError wrapper in a few places in the sqlite implementation.

It may make sense to move the sqlcommon implementation out into `sqlcommon/tuple_iterator.go` so that it's easier to diff between the 2 until we can merge them. It could also be an idea to separate the code to scan a row into a TupleRecord out into a helper which would allow us to remove much of the duplicated code.

yissellokta (dismissed):

elbuo8:
> You'll want to at least sync the handleerror wrapping between the 2 implementations then, just a few lines that don't match between sqlcommon and sqlite

Ah didn't quite see that, will loop that in 👍🏽

DanCech (approved):
LGTM

DanCech (approved):

PR #2441, opened 2025-05-13
fix: fanin draining across clients

elbuo8:
## Description
Move draining of `fanin.out` channel to consumers rather than itself to prevent race conditions.
Make draining be completely in deferred goroutines.
Change `fanin.Add` to return if new channel is accepted, and handle it's draining if rejected at the client.
Fix bug in `breadthFirstRecursiveMatch` where `fanin.Stop` was not being called as early as possible in the case of recurssion.

## Review Checklist

### DIFF
--- a/internal/iterator/fan_in.go
+++ b/internal/iterator/fan_in.go
@@ -46,7 +46,7 @@ func (f *FanIn) cleaner() {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 	for _, ch := range f.drainQueue {
-		Drain(ch)
+		Drain(ch).Wait()
 	}
 	f.wg.Done()
 }
--- a/internal/iterator/fan_in.go
+++ b/internal/iterator/fan_in.go
@@ -17,8 +17,10 @@ type FanIn struct {
 	drainQueue []chan *Msg
 	accepting  bool
 	mu         sync.Mutex
-	wg         sync.WaitGroup
 	pool       *pool.ContextPool
+
+	// for unit tests state
+	wg sync.WaitGroup
 }
 
 func NewFanIn(ctx context.Context, limit int) *FanIn {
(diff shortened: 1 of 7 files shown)


internal/iterator/fan_in.go
```diff
@@ -122,8 +140,7 @@ func (f *FanIn) Done() {
 	}
 }
 
-func (f *FanIn) Close() {
+func (f *FanIn) Stop() {
 	// Done gets called internally
 	f.cancel()
-	drainOnExit(f.out)
```
justincoh, line 128: Was this the root of the issue earlier?

elbuo8: Not completely. But this does cause one of the issues. Particularly a race.

internal/iterator/fan_in.go
```diff
@@ -10,46 +10,63 @@ import (
...
-		f.drained <- true
-		close(f.drained)
+		f.mu.Lock()
+		queueSize := len(f.drainQueue)
+		f.mu.Unlock()
+		if queueSize > 0 {
+			f.wg.Add(1)
+			go f.cleaner()
+		}
+		f.wg.Done()
```
adriantam, line 69: Who is waiting for this wg?  Is this only for unit test purpose?  If so, can you add this as comment?

internal/iterator/fan_in.go
```diff
@@ -10,46 +10,63 @@ import (
...
+
 	return f
 }
 
+func (f *FanIn) cleaner() {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	for _, ch := range f.drainQueue {
+		Drain(ch).Wait()
+	}
```
adriantam, line 50: Do we want to run all Drain(ch) first and then call Wait on them?  Right now, the second ch will not start until the first one Wait() is done.

adriantam (commented):
Mind filling in the description on what is the bug we are fixing and add in CHANGELOG?

internal/iterator/fan_in.go
elbuo8: I don't want to end up with a bunch of concurrent goroutines being spawned just for draining channels, I'm ok with them being sequential (otherwise I wouldn't `.Wait()` on them.

internal/iterator/fan_in.go
elbuo8: Added comment.

internal/iterator/fan_in.go
adriantam: Mind adding comment to call it out specifically?

adriantam (approved):

yissellokta (approved):

PR #2446, opened 2025-05-15
fix: add recover for authz goroutine

Vic-Dev:
## Description
Goroutine in authz is recovered and the error is passed through the errorChannel

## Review Checklist

### DIFF
--- a/internal/authz/authz.go
+++ b/internal/authz/authz.go
@@ -430,6 +430,11 @@ func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, st
 	for _, module := range modules {
 		wg.Add(1)
 		go func(module string) {
+			defer func() {
+				if r := recover(); r != nil {
+					errorChannel <- &authorizationError{Cause: fmt.Sprintf("panic recovered: %v", r)}
+				}
+			}()
 			defer wg.Done()
 			contextualTuples := openfgav1.ContextualTupleKeys{
 				TupleKeys: []*openfgav1.TupleKey{
@@ -443,7 +448,6 @@ func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, st
 			}
 
 			err := a.individualAuthorize(ctx, clientID, relation, ModuleIDType(storeID).String(module), &contextualTuples)
-
 			if err != nil {
 				errorChannel <- err
 			}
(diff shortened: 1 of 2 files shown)


internal/authz/authz.go
```diff
@@ -443,20 +442,24 @@ func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, st
...
 			}
-		}(module)
+		})
 	}
 
-	wg.Wait()
-	close(errorChannel)
+	recoveredError := wg.WaitAndRecover()
+	if recoveredError != nil {
+		errorChannel <- &authorizationError{Cause: fmt.Sprintf("panic recovered: %v", recoveredError.AsError())}
```
mkgobaco, line 453: do we need to close(errorChannel) so no more values can be sent to it?

internal/authz/authz.go
```diff
@@ -443,20 +442,24 @@ func (a *Authorizer) moduleAuthorize(ctx context.Context, clientID, relation, st
...
+	}
 
-	for err := range errorChannel {
+	select {
+	case err := <-errorChannel:
 		if err != nil {
 			return err
 		}
+	default:
+		return nil
```
mkgobaco, line 462: If we have a default case, the select does not wait for a value to be read from errorChannel.  Is that OK?

internal/authz/authz.go
cikasfm: @mkgobaco that's done using `defer` on line 429

```golang
	errorChannel := make(chan error, len(modules)+1)
	defer close(errorChannel)
```

internal/authz/authz.go
cikasfm: IMHO that assumes that there's no error at this point, so we can return

internal/authz/authz.go
senojj: At this point, we've already identified an error so why not return immediately with the error at this point? Don't bother with adding it to the channel.

internal/authz/authz.go
senojj: A select first evaluates all case statements for readiness before entering the default case.

Since `wg.WaitAndRecover` is called before this code, the go routine has already finished and a potential error value will be ready in `errorChannel` at this point.

mkgobaco: Ah, I see. @cikasfm @senojj you guys are right.

Vic-Dev:
Decided not to go with this approach

PR #2465, opened 2025-05-21
refactor: move ReverseExpand's LoopOnEdges into its own function

justincoh:
## Description

In the reverse_expand operation (within ListObjects), the `execute` method does several things, including looping over all necessary edges in a named loop `LoopOnEdges`. This PR extracts that LoopOnEdges logic into its own method.

There are no changes in this PR, just organizing to prepare for some coming changes in this file.

## Review Checklist

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -323,11 +323,36 @@ func (c *ReverseExpandQuery) execute(
 		return err
 	}
 
-	pool := concurrency.NewPool(ctx, int(c.resolveNodeBreadthLimit))
+	errs := c.LoopOnEdges(
+		ctx,
+		req,
+		resultChan,
+		intersectionOrExclusionInPreviousEdges,
+		resolutionMetadata,
+		edges,
+		sourceUserObj,
+	)
 
+	if errs != nil {
+		telemetry.TraceError(span, errs)
+		return errs
+	}
+
+	return nil
+}
+
+func (c *ReverseExpandQuery) LoopOnEdges(
+	ctx context.Context,
+	req *ReverseExpandRequest,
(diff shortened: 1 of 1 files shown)


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -323,11 +323,36 @@ func (c *ReverseExpandQuery) execute(
...
 
+func (c *ReverseExpandQuery) loopOnEdges(
+	ctx context.Context,
+	req *ReverseExpandRequest,
+	resultChan chan<- *ReverseExpandResult,
+	intersectionOrExclusionInPreviousEdges bool,
+	resolutionMetadata *ResolutionMetadata,
+	edges []*graph.RelationshipEdge,
+	sourceUserObj string,
+) error {
```
senojj, line 352: I tend to think of a function's signature as a type of interface. Generally, the larger an interface is, the weaker its abstraction is. Due to the number and variety of arguments in this function, I question if this function is truly abstracting the logic within, or if it is simply encapsulating.

If the function is just encapsulating, I wonder what value is being introduced through the function, as opposed to just leaving the logic where it currently resides. Is there an opportunity for a stronger abstraction?

justincoh: That is 100% correct, this is purely encapsulation. I'm working on introducing a `loopOnEdgesV2` so to speak (it will have a much more helpful name), which will create a fork in `ReverseExpand.execute` execution path. However, that fork will occur higher up in this function–before we get into this loop–so I could just leave this as-is and close this PR as `won't do`.

What do you think?

justincoh: Discussed offline and decided to close this PR pending the coming updates to reverse_expand. During/after that work we might be able to abstract some of this bettter

Issue #1645, opened 2024-05-29
Make it possible to configure 2 connection strings (one for reads and one for writes)


yann-disanto:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Hello there,

Our SRE team would love to be able to set up a read replica for our OpenFGA database. 
From our perspective, the "read" queries need to be the fastest possible and require the highest availability, more than the "write" queries (I'm talking from the APIs point of view). Basically, the usages and latency requirements are differents for reads and writes.
So we'd like to be able to configure 2 databases, one used when calling the "write" endpoints and another one used when calling the "read" endpoints (this one would be a read replica of the one used for the writes).

### Describe the ideal solution

So we'd like to be able to configure 2 connection strings, one for the writes and another one for the reads. 
Also the database settings (stuff like `OPENFGA_DATASTORE_MAX_OPEN_CONNS`) should be configurable for each database (maybe something like `OPENFGA_READ_DATASTORE_MAX_OPEN_CONNS` and `OPENFGA_WRITE_DATASTORE_MAX_OPEN_CONNS` that would fallback to `OPENFGA_DATASTORE_MAX_OPEN_CONNS` when not specified).

aaguiarz:
We'll be working on this as part of our improvements our consistency options https://github.com/openfga/roadmap/issues/54.

When you query OpenFGA with higher consistency, we'll always use the write replica, and if not, the read replica.

matthisholleville:
Hello @aaguiarz. I just submitted a [PR](https://github.com/openfga/openfga/pull/2479) that adds this feature for PostgreSQL. Looking forward to your feedback. 

It’s a feature we've been eagerly waiting for, and it will help us improve our resilience and scale.

PR #2479, opened 2025-06-01
feat: support separate read and write datastores for postgresql

matthisholleville:
## Description

support separate read and write datastores for postgresql improved performance and scalability

This change introduces the ability to configure separate read and write datastores in the application, specifically targeting PostgreSQL setups. This enhancement is aimed at improving performance and scalability by distributing read and write operations across different databases.

## References

Closes https://github.com/openfga/openfga/issues/1645

## Review Checklist


paulinejamin (approved):

pkg/storage/postgres/postgres.go
```diff
@@ -129,26 +158,63 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
...
-	dbInfo := sqlcommon.NewDBInfo(db, stbl, HandleSQLError, "postgres")
+	return collector, nil
+}
 
-	return &Datastore{
-		stbl:                   stbl,
-		db:                     db,
-		dbInfo:                 dbInfo,
+// NewWithDB creates a new [Datastore] storage with the provided database connection.
+func NewWithDB(writeDb *sql.DB, cfg *sqlcommon.Config, opts ...Option) (*Datastore, error) {
```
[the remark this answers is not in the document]
matthisholleville: Not sure it's a good idea. In fact, we're trying to create a datastore instance. I’m more in favor of keeping a generic name for this method, especially since it's used in other providers (MySQL, etc.) without the concept of options. It could become confusing.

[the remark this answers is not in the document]
DanCech: I have a more fundamental issue with the change to this function, specifically that it no longer returns a datastore that uses the provided `sql.Db` connection but rather will try to establish a second read connection and use that.

It may make more sense for `New` to handle establishing both the primary "write" connection and the secondary "read" connection (if configured) and to provide both to `NewWithDB` (passing `nil` if there is no separate "read" connection used).

aaguiarz:
All ReadOptions, ReadUserTupleOptions, ReadUsersetTuplesOptions, ReadStartingWithUserOptions have a 'Consistency' field that can be "HIGHER_CONSISTENCY".

If that's the case, we probably want to read from the primary replica and not read replicas.

matthisholleville:
That makes sense. I'm updating the PR.

matthisholleville:
@aaguiarz updated.

I’m wondering if, to avoid any confusion, we should rename read/write to primary/secondary databases. Let me know what you prefer — I’ll adapt.

pkg/storage/postgres/postgres.go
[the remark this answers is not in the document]
DanCech: I do see how the opts mechanism gives the caller the ability to provide the second database connection, though I am unclear on why we don't let New handle creating the connection based on `cfg` alongside the primary connection?

pkg/storage/postgres/postgres.go
```diff
@@ -767,7 +848,24 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
...
+	if err != nil {
+		writeStatus.Message = err.Error()
+		writeStatus.IsReady = false
+	}
+	readStatus, err = sqlcommon.IsReady(ctx, s.readDb)
+	if err != nil {
+		readStatus.Message = err.Error()
+		readStatus.IsReady = false
+	}
+
```
DanCech: if `s.readDb` is just a reference to `s.writeDb` we can likely skip this?

matthisholleville:
Thank you for your review @DanCech. Things are clearer for me now. I’ve made quite a few changes and added a test for the separation of write and read queries.

cmd/run/run.go
```diff
@@ -439,7 +447,7 @@ func (s *ServerContext) datastoreConfig(config *serverconfig.Config) (storage.Op
 			return nil, nil, fmt.Errorf("initialize mysql datastore: %w", err)
 		}
 	case "postgres":
-		datastore, err = postgres.New(config.Datastore.URI, dsCfg)
+		datastore, err = postgres.New(config.Datastore.URI, config.Datastore.SecondaryURI, dsCfg)
```
DanCech: would it make sense to just include the primary and secondary URIs in dsCfg? That would mean we can stick with the same `New(sqlcommon.Config)` signature for all the database adapters.

pkg/storage/postgres/postgres.go
```diff
@@ -118,58 +142,117 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
...
+	}
+	return s.primaryDbInfo
+}
+
+// getReadStbl returns the appropriate statement builder based on consistency options
+func (s *Datastore) getReadStbl(consistency *openfgav1.ConsistencyPreference) sq.StatementBuilderType {
+	if consistency != nil && *consistency == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
+		// If we are using higher consistency, we need to use the write database.
+		return s.primaryStbl
+	}
```
DanCech, line 218: This makes me feel a lot better about reading from a replica, read-after-write is a common pain point for these types of setups and being able to opt out of reading from the replica can definitely help with that.

We should also update the docs with an explanation of how to configure postgres replicas and sync vs async replication to avoid getting a lot of support questions.

matthisholleville: Where should this documentation be written? I can try to contribute to the documentation for the PostgreSQL part.

DanCech: This is probably the most logical place:

https://github.com/openfga/openfga.dev/blob/main/docs/content/getting-started/setup-openfga/configure-openfga.mdx#postgres

cmd/run/run.go
matthisholleville: @DanCech PR updated

pkg/storage/postgres/postgres.go
```diff
@@ -767,7 +853,35 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
...
+	secondaryStatus, err := sqlcommon.IsReady(ctx, s.secondaryDb)
+	if err != nil {
+		readyStatus.Message = fmt.Sprintf("%s, secondary: %s", readyStatus.Message, err.Error())
+		readyStatus.IsReady = false
+	}
+
+	// if secondary is configured, return primary and secondary status
+	readyStatus.IsReady = readyStatus.IsReady && secondaryStatus.IsReady
+
+	return readyStatus, nil
```
DanCech: I'm not sure that I fully understand the possible permutations of the outputs from this function, each database can have 3 possible states; ready, not ready, and error. If there is no secondary then we have the standard output.

With secondary there appear to be 9 possible states:

```
primary ready,     secondary ready:         true,  ""
primary ready,     secondary not ready:     false, "" (secondaryStatus.Message not used)
primary ready,     secondary error:         false, ", secondary: <error>"
primary not ready, secondary ready:         false, "primary: <message>"
primary not ready, secondary not ready:     false, "primary: <message>" (secondaryStatus.Message not used)
primary not ready, secondary error:         false, "primary: <message>, secondary: <error>"
primary error,     secondary ready:         false, "primary: <error>"
primary error,     secondary not ready:     false, "primary: <error>" (secondaryStatus.Message not used)
primary error,     secondary error:         false, "primary: <error>, secondary: <error>"
```

I'm thinking that the message should always be of the form `primary: <ready|error|message>, secondary: <ready|error|message>` when we have a secondary, wdyt?

This could probably use a test to verify the expected output also.

pkg/storage/postgres/postgres.go
matthisholleville: PR opened https://github.com/openfga/openfga.dev/pull/1043

pkg/storage/postgres/postgres.go
matthisholleville: I will modify the function accordingly and add a test to cover all possible cases.

pkg/storage/postgres/postgres_test.go
```diff
@@ -26,17 +26,41 @@ func TestPostgresDatastore(t *testing.T) {
...
+
+	ds, err := New(cfg)
+	require.NoError(t, err)
+	defer ds.Close()
+
+	status, err := ds.IsReady(context.Background())
+	require.NoError(t, err)
+	require.True(t, status.IsReady)
+	require.Equal(t, "primary: ready, secondary: ready", status.Message)
+}
```
matthisholleville: @DanCech PR updated with a new test. Does the structure of the `IsReady` function make more sense to you?

pkg/storage/postgres/postgres.go
```diff
@@ -767,7 +853,42 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
 
 // IsReady see [sqlcommon.IsReady].
 func (s *Datastore) IsReady(ctx context.Context) (storage.ReadinessStatus, error) {
-	return sqlcommon.IsReady(ctx, s.db)
+
```
DanCech: We can just return `sqlcommon.IsReady()` if there is no secondary, vs changing the error state behavior.
```suggestion

	// if secondary is not configured, return primary status only
	if !s.isSecondaryConfigured() {
		return sqlcommon.IsReady(ctx, s.primaryDb)
	}

```

pkg/storage/postgres/postgres.go
```diff
@@ -767,7 +853,42 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
...
+	// if secondary is not configured, return primary status only
+	if !s.isSecondaryConfigured() {
+		return primaryReadyStatus, nil
+	}
+
+	if primaryReadyStatus.Message == "" {
+		primaryReadyStatus.Message = "ready"
+	}
+
```
DanCech: We can remove the check here if it's handled above. I'd also rename `primaryReadyStatus` to just `primaryStatus` so it matches with `secondaryStatus` below.

We should only set the message to "ready" if the connection is actually ready (vs not ready but missing a message for some reason).
```suggestion

	if primaryReadyStatus.IsReady && primaryReadyStatus.Message == "" {
		primaryReadyStatus.Message = "ready"
	}

```

pkg/storage/postgres/postgres.go
```diff
@@ -767,7 +853,42 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
...
+	messageTpl := "primary: %s, secondary: %s"
+
+	// check if secondary is ready
+	secondaryStatus, err := sqlcommon.IsReady(ctx, s.secondaryDb)
+	if err != nil {
+		secondaryStatus.Message = err.Error()
+		secondaryStatus.IsReady = false
+	}
+
+	if secondaryStatus.Message == "" {
```
DanCech:
```suggestion
	if secondaryStatus.IsReady && secondaryStatus.Message == "" {
```

pkg/storage/postgres/postgres.go
```diff
@@ -767,7 +853,42 @@ func (s *Datastore) ReadChanges(ctx context.Context, store string, filter storag
...
+		return primaryReadyStatus, nil
+	}
+
+	if primaryReadyStatus.Message == "" {
+		primaryReadyStatus.Message = "ready"
+	}
+
+	multipleReadyStatus := storage.ReadinessStatus{}
+	messageTpl := "primary: %s, secondary: %s"
+
```
DanCech: This can be moved down to line 887

pkg/storage/postgres/postgres_test.go
DanCech: That definitely works to test the ready/ready state handling, to fully test we'd want to handle the other cases, to do that we'd need to mock `ds.primaryDb` and `s.secondaryDb` so we can control the result of `sqlcommon.IsReady` for each of them.

matthisholleville: That was precisely the aim of the next commits.

DanCech:
Logic looks good to me! There are some lint items that need to be tidied up, and the changelog needs to be updated.

We still don't have test coverage on the not-ready code paths, but I don't think that needs to be a blocker.

matthisholleville:
@DanCech PR updated

DanCech:
Hmm, I haven't run into the go-bench issue before, but testing locally (`go test -bench=BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheck/with_complex_ttu ./pkg/server`) I am seeing that it's ~5x slower (660ms vs 115ms) than main so we'll need to figure out what's going on there.

I was able to get some clean profiles by running
```
go test -bench=BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheck/with_complex_ttu -skip Test -cpuprofile cpu.out ./pkg/server
```

They show that we're spending all that time reconnecting to the database

matthisholleville:
I just rebase the main branch... Will take a look if I have time

matthisholleville:
@DanCech I think the definition of `maxIdleConnections` and `maxOpenConnexions` was missing after the rebase of main, which modifies this [behavior](https://github.com/openfga/openfga/commit/a25ad6296c4ed4a86ae2acd1973b8e61fd4fd9d6).

DanCech (approved):
LGTM, nice work!

matthisholleville:
@DanCech Thank you for your reviews !

Whats next now ?

DanCech:
We need a review from @openfga/backend , in the meantime if you have time to work on increasing the test coverage that would be amazing!

matthisholleville:
image

@DanCech when I see the lines that aren't covered, I wonder how reliable the coverage report is. The fact of having changed the function parameter is already covered by existing tests. Am I wrong?

matthisholleville:
@aaguiarz can you please assign @openfga/backend for review please ?

aaguiarz:
@matthisholleville they are assigned, we'll review it next week 

cc @rcocetta @senojj

pkg/storage/sqlite/sqlite.go
```diff
@@ -92,7 +92,8 @@ func PrepareDSN(uri string) (string, error) {
 }
 
 // New creates a new [Datastore] storage.
-func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
+func New(cfg *sqlcommon.Config) (*Datastore, error) {
```
adriantam: Is it possible to keep this interface the same 
```
func New(uri string, cfg *sqlcommon.Config) (*Datastore, error)
```
to avoid breaking changes?

adriantam (commented):
Thank you for the PR, the changes look good to me.  The only thing to consider is whether it is possible to keep `func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {` to avoid breaking changes.  Thanks!

pkg/storage/sqlite/sqlite.go
aaguiarz: To clarify, there are OpenFGA users that implement their own DB adapters, and this would be a breaking change for them, that seems avoidable.

matthisholleville: Thank you @aaguiarz  for clarifying. I was not aware of this use case. PR updated ✅

adriantam (commented):
One last request, can you update https://github.com/openfga/openfga/blob/main/cmd/run/flags.go and https://github.com/openfga/openfga/blob/main/.config-schema.json so that the secondary connection can be configured via environment variables as per https://openfga.dev/docs/getting-started/setup-openfga/configuration?  Thank you.

aaguiarz:
@matthisholleville Also, there's a linter issue, and the code coverage check is not passing.

matthisholleville:
@aaguiarz as I said here https://github.com/openfga/openfga/pull/2479#issuecomment-2950627172, the coverage seems to be wrong. What do you think ?

matthisholleville:
@adriantam fixed. @aaguiarz the lint issue is not related to my PR, but it has been fixed.

DanCech (approved):

adriantam (approved):

PR #2483, opened 2025-06-04
perf: sql based datastore isready consumes too many connections

adriantam:
## Description
After we have established the db version is good, we should only do ping as it is cheaper.

## Review Checklist

### DIFF
--- a/pkg/storage/mysql/mysql.go
+++ b/pkg/storage/mysql/mysql.go
@@ -76,21 +76,13 @@ func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
 
 // NewWithDB creates a new [Datastore] storage with the provided database connection.
 func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
-	if cfg.MaxOpenConns != 0 {
-		db.SetMaxOpenConns(cfg.MaxOpenConns)
-	}
-
 	if cfg.MaxIdleConns != 0 {
-		db.SetMaxIdleConns(cfg.MaxIdleConns)
+		db.SetMaxIdleConns(cfg.MaxIdleConns) // default is 2, not retaining connections(0) would be detrimental for performance
 	}
 
-	if cfg.ConnMaxIdleTime != 0 {
-		db.SetConnMaxIdleTime(cfg.ConnMaxIdleTime)
-	}
-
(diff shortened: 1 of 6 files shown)


DanCech:
Seems like a reasonable optimization, but we need to make sure that the error messages still make sense if the database connection isn't available. It wouldn't be great if we go from a clear status message that the connection has failed to a less-useful message about a query failure while trying to list migrations.

elbuo8 (dismissed):

DanCech (dismissed):

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -759,6 +760,20 @@ func IsReady(ctx context.Context, db *sql.DB) (storage.ReadinessStatus, error) {
 	}, nil
 }
 
+// IsReady returns true if the connection to the datastore is successful
+// and the datastore has the latest migration applied.
```
elbuo8:
```suggestion
```

pkg/storage/mysql/mysql.go
```diff
@@ -43,6 +43,7 @@ type Datastore struct {
 	dbStatsCollector       prometheus.Collector
 	maxTuplesPerWriteField int
 	maxTypesPerModelField  int
+	versionReady           storage.ReadinessStatus
```
elbuo8: 🤔 why store the whole struct if this is just relevant to version. 

Also, if you are going to keep this at the struct level, you can have a single method of `isReady` as not to continue adding to the interface. And just pass in if it needs to check the version.

adriantam: Changed to just keep bool

DanCech:
The only concern I have with this approach is the edge case where someone does a database schema rollback underneath a running openfga instance, in the past that would have become not ready but now it will possibly fail in some interesting way. I'm not sure whether that's a real concern though, as one would assume the operator would also be rolling back to the earlier openfga binary as well.

adriantam:
> The only concern I have with this approach is the edge case where someone does a database schema rollback underneath a running openfga instance, in the past that would have become not ready but now it will possibly fail in some interesting way. I'm not sure whether that's a real concern though, as one would assume the operator would also be rolling back to the earlier openfga binary as well.

Yes. This is the assumption.  At least if they rollback and restart OpenFGA, it will become unready.

PR #2484, opened 2025-06-04
fix sql setup

elbuo8:
## Review Checklist


adriantam:
Please update changelog

tests/check/check_test.go
```diff
@@ -340,10 +340,6 @@ func testRunAll(t *testing.T, engine string) {
 
 	cfg.CheckIteratorCache.Enabled = true
 
-	if engine == "mysql" {
```
adriantam, line 343: why do we remove this?

elbuo8: Should have never been added in the first place. We don't propagate context cancellations to the datastores by default. That is an opt-in behavior that I wouldn't recommend.

If you short circuit, it will basically prune your whole connection pool. With this, the engine becomes unstable/unusable (hence why the matrix tests were failing).

justincoh (approved):

PR #2492, opened 2025-06-09
perf: improve performance of recursive strategies

elbuo8:
## Description

Reduce the number of abstractions to move data across channels. 
Drop the usage of pool when not needed.
Don't wait for cleanup or short circuited.

It's expected that the bench for `with_nested_userset` to fail as it was short circuiting before earlier than it should have. Effectively being incorrect.

## Review Checklist


internal/graph/check_fast_path.go
```diff
@@ -616,7 +618,7 @@ func (c *LocalChecker) checkTTUFastPathV2(ctx context.Context, req *ResolveCheck
 	return c.resolveFastPath(ctx, leftChans, storage.WrapIterator(storage.TTUKind, iter))
 }
 
-func fanInIteratorChannels(ctx context.Context, chans []chan *iterator.Msg, concurrencyLimit int) chan *iterator.Msg {
+func fanInIteratorChannels(ctx context.Context, chans []chan *iterator.Msg, _ int) chan *iterator.Msg {
```
adriantam: since we don't use the concurrencyLimit anymore, should we remove it from parameter?

elbuo8: Done.

internal/graph/check_fast_path.go
```diff
@@ -701,61 +661,38 @@ func (c *LocalChecker) breadthFirstRecursiveMatch(ctx context.Context, req *Reso
...
-						break
-					}
-					concurrency.TrySendThroughChannel(ctx, checkOutcome{err: err}, checkOutcomeChan)
-					close(checkOutcomeChan)
-					return
+		// if the pool is short-circuited, the iterator should be stopped
+		defer mapper.Stop()
+		pool.Go(func(ctx context.Context) error {
+			objectToUsersetMessageChan := streamedLookupUsersetFromIterator(ctx, mapper)
+			for usersetMsg := range objectToUsersetMessageChan {
```
adriantam, line 670: Do we need to care about context timeout?

internal/graph/check_fast_path_test.go
adriantam, line 1: curious why do we remove t.Parallel()?

internal/graph/object_providers.go
```diff
@@ -79,9 +75,6 @@ func (c *recursiveTTUObjectProvider) End() {
 	if c.cancel != nil {
 		c.cancel()
 	}
-	if c.pool != nil {
```
adriantam, line 82: there is a change in behaviour in that we don't wait for the goroutine to be completed.  Is that intentional?

internal/graph/object_providers.go
```diff
@@ -124,65 +111,72 @@ func (c *recursiveUsersetObjectProvider) End() {
...
```
adriantam, line 127: Again, there is a change in behaviour in that we don't wait for goroutine to be done. Please verify this is intentional

internal/graph/object_providers_test.go
```diff
@@ -454,98 +453,111 @@ func TestRecursiveUsersetObjectProvider(t *testing.T) {
...
+		go iteratorsToUserset(ctx, chans, outChan)
+
+		count := 0
+		for msg := range outChan {
+			if msg.err != nil {
+				count++
+			}
+		}
+		require.Equal(t, 2, count)
+	})
```
adriantam, line 565: should we verify that we see the message content (i.e., there is at least 1 error)?

internal/iterator/fan_in_test.go
```diff
@@ -72,3 +72,41 @@ func TestFanIn(t *testing.T) {
...
+		i, err := strconv.Atoi(id)
+		require.NoError(t, err)
+		require.Positive(t, i)
+		require.LessOrEqual(t, i, 9)
+		_, err = msg.Iter.Next(ctx)
+		require.Equal(t, storage.ErrIteratorDone, err)
+		msg.Iter.Stop()
+		iterations++
+	}
+}
```
adriantam, line 137: Do we need to verify that at the end there are 9 iterations?

internal/graph/check_fast_path.go
elbuo8: That should be handled by `streamedLookupUsersetFromIterator`.

internal/graph/check_fast_path_test.go
elbuo8: They weren't parallel ones.

internal/graph/object_providers.go
elbuo8: Correct. This made any short circuiting have to wait for cleanup, def not something we want.

internal/graph/object_providers.go
elbuo8: Intentional.

internal/graph/check_fast_path.go
```diff
@@ -701,61 +661,38 @@ func (c *LocalChecker) breadthFirstRecursiveMatch(ctx context.Context, req *Reso
...
+				mu.Unlock()
 			}
-		}
+			return nil
+		})
 	}
 
+	// wait for all checks to wrap up
+	// if a match was found, clean up
+	if err := pool.Wait(); err != nil && errors.Is(err, ErrShortCircuit) {
```
senojj:
```suggestion
	if err := pool.Wait(); errors.Is(err, ErrShortCircuit) {
```

internal/graph/object_providers_test.go
elbuo8: We are actually counting the errors themselves 🤔

internal/iterator/fan_in.go
```diff
@@ -146,3 +144,37 @@ func (f *FanIn) Done() {
 func (f *FanIn) Stop() {
 	f.cancel()
 }
+
+func FanInIteratorChannels(ctx context.Context, chans []chan *Msg) chan *Msg {
```
senojj:
```suggestion
func FanInIteratorChannels(ctx context.Context, chans []<-chan *Msg) <-chan *Msg {
```

senojj: note that the slice being passed in would need to be of type `[]<-chan *Msg`

adriantam (dismissed):

senojj (approved):

adriantam (approved):

Issue #2493, opened 2025-06-09
regression: migration CLI is broken for Postgres


JeremyLoy:
### Checklist

- [x] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [x] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

setting the `--datastore-username` or `--datastore-password` flags (or the equivalent ENV vars like `OPENFGA_DATASTORE_USERNAME`) does not work for Postgres, and possibly the other engines. The URI is overwritten with empty strings for those values.

This used to work pre-1.8.13, and I believe was introduced in #2442 cc @eleijonmarck @jpadilla 

below I have the output using the basic docker-compose template from the tutorial. In that example, I remove the `username:password` from the URI and move them to the flags, but it is also reproduceable simply by setting the flags with the correct values, keeping the URI unaltered

### OpenFGA version

1.8.13

### How are you running OpenFGA?

As a binary

### What datastore are you using?

In-Memory

### OpenFGA Flags

no

### Logs

jeremyloy@WG4D9QKY2L hs-openfga-service % docker-compose up -d
[+] Running 3/3
 ✔ Container postgres  Healthy                                                                                                                                                                        5.7s 
 ✔ Container migrate   Exited                                                                                                                                                                         6.5s 
 ✔ Container openfga   Started                                                                                                                                                                        6.7s 

jeremyloy@WG4D9QKY2L hs-openfga-service % openfga migrate --datastore-engine=postgres --datastore-uri=postgres://postgres:password@localhost:5432/postgres
2025/06/09 16:29:25 current version 5
2025/06/09 16:29:25 running all migrations
2025/06/09 16:29:25 migration done
[...]

PR #2442, opened 2025-05-14
feat: expose `RunMigration` from `cmd`

eleijonmarck:
feat: Expose `RunMigrations()` for library migration management

**Problem:** Grafana, embedding OpenFGA as a library, faces challenges with divergent database migration approaches. OpenFGA uses `goose`, while Grafana's migrator runs once, potentially skipping future OpenFGA schema updates and preventing versioned upgrades.

**Solution:** Utilize `RunMigrations()` from cmd for reliable schema management.

**Changes:**
* Introduces `MigrationConfig` struct.
* Adds `RunMigrations(cfg MigrationConfig) error` function to the library.
* Updates CLI `runMigration` to use the new library function.

This provides any organization that uses openFGA as a library with a method to manage OpenFGA database migrations.

Example:
```go
import "github.com/openfga/openfga/pkg/storage/migrate"

// Then use it
err := migrate.RunMigrations(migrate.MigrationConfig{
    Engine:        "postgres",
    URI:           "postgres://user:password@localhost:5432/openfga",
    TargetVersion: 0, // Latest version
    Timeout:       time.Minute,
    Verbose:       true,
})
```

**NOTE:**
I could add documentation in openfga.dev for how to use this as part of a library, after we have been able to use this.

Fixes issue: https://github.com/openfga/openfga/issues/2440

Issue #870, opened 2023-07-14
Adding support for AWS IAM Authentication


KlausVii:
Hi folks,

We recently deployed openfga on our cluster and we ran into some friction with having to be forced to use user+password auth for our postgres, as all our other services use IAM auth for our db access everywhere else.

Would you be happy add this support? If so, we would be happy sort out the PR.

Many thanks :)

rhamzeh:
Hi @KlausVii - absolutely, we'd love a contribution!

KlausVii:
@rhamzeh - Lovely.

Here's the preliminary flag changes I think we'll need:
- datastore-use-aws-iam: boolean

And probably to decompose the db uri into these to make the setup a bit easier
- datastore-db-name: string
- datastore-db-port: string
- datastore-db-host: string

Are we happy to leave the AWS_ vars to be passed as is and not included the in flags? Or should there be an option to pass them explicitly?

Let me know what you think. I should be able to get started on this later this week.

rhamzeh:
cc @jon-whit : I think you have the most comments/thoughts on this

jon-whit:
@KlausVii could we open this up for a larger discussion in a more formal RFC (see https://github.com/openfga/rfcs)?

What I'd like to see, beyond just a rough cut of immediately supporting AWS IAM credentials for database credentials, is a rough plan of what this looks like as a developer (what's the setup and DX for this new credential management process) and what _may_ it look like for the other vendors out there (Azure, GCP, etc..). Some open questions:

* **Package structure** - We're introducing a new auth method for the storage layer, and this is potentially for more than one vendor. I'd like to see a recommendation/proposal for that.

* **Config structure** - We support config.yaml, env vars, and flags. Each of these need a proposed change associated with them, and the config structure will vary based on the auth provider I'm sure. We can focus on AWS, but we should at least have a rough story for one other vendor.

* **Interface composition** - At the end of the day the abstraction is a database/sql connection pool to the underlying SQL database through a driver (e.g. `pgx` for postgres or `go-sql-driver/mysql`), but how you obtain that is changing. I'd like to see an attempt for a proposal for an interface that is shared across the implementations that abstracts away the connection details and has implementations specific to the providers. The [Go Cloud CDK](https://gocloud.dev/howto/sql/) might be a good starting point for inspiration or maybe we just adopt that altogether as the defacto standard for our implementation?

    For example:
[...]

PR #2498, opened 2025-06-11
fix: regression of empty string initiation for `migrate` pkg

eleijonmarck:
We used to have https://github.com/openfga/openfga/blob/5f720784ae7c637bc9a9ae8a18adc230982aba50/cmd/migrate/migrate.go#L64-L65

Then the regression was introduced in https://github.com/openfga/openfga/pull/2442, where we set https://github.com/openfga/openfga/blob/a3cae63e511715c91198673a1a72e7b0facc8168/pkg/storage/migrate/migrate.go#L69 to empty string initiation.

We now evaluate the following priority:
- If `--datastore-username` is provided, use
- If `--datastore-password` is provided, use
- Otherwise: fall back to URI credentials

Maintains same behavior as MySQL case, and tested on mysql

Fixes:
https://github.com/openfga/openfga/issues/2493

### DIFF
--- a/pkg/storage/migrate/migrate.go
+++ b/pkg/storage/migrate/migrate.go
@@ -73,11 +73,14 @@ func RunMigrations(cfg MigrationConfig) error {
 		if err != nil {
 			return fmt.Errorf("invalid database uri: %v", err)
 		}
-		// if username not set
-		if cfg.Username == "" && dbURI.User != nil {
+		if cfg.Username != "" {
+			username = cfg.Username
+		} else if dbURI.User != nil {
 			username = dbURI.User.Username()
 		}
-		if cfg.Password == "" && dbURI.User != nil {
+		if cfg.Password != "" {
+			password = cfg.Password
+		} else if dbURI.User != nil {
 			password, _ = dbURI.User.Password()
 		}
 		dbURI.User = url.UserPassword(username, password)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 ## [Unreleased]
 ### Fixed
 - Reverts the base docker image back to `cgr.dev/chainguard/static` [#2473](https://github.com/openfga/openfga/issues/2473)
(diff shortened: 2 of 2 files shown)


pkg/storage/migrate/migrate.go
```diff
@@ -73,11 +73,14 @@ func RunMigrations(cfg MigrationConfig) error {
...
+		} else if dbURI.User != nil {
 			username = dbURI.User.Username()
 		}
-		if cfg.Password == "" && dbURI.User != nil {
+		if cfg.Password != "" {
+			password = cfg.Password
+		} else if dbURI.User != nil {
 			password, _ = dbURI.User.Password()
 		}
 		dbURI.User = url.UserPassword(username, password)
```
JeremyLoy, line 86: possibly beyond the scope of this pr, as this maintains the original v1.8.12 behavior, but this line is breaking PGPASSFILE useage for the `migrate` command. 

using `url.UserPassword` always will set the uri to `user:` or `:password` or `:` depending on which permutation of zero values exist for `username` and `password`

The main application doesn't have this problem, but it does reiterate that the migration bootstrapping and application bootstrapping are diverging a bit, which can cause problems in the future

https://github.com/openfga/openfga/issues/870#issuecomment-2768134416

JeremyLoy: otherwise, LGTM - thank you for fixing and for your efforts in improving the migration config as a whole!

adriantam: Agree on the above.  I would recommend attacking this on a separate PR

adriantam (approved):

pkg/storage/migrate/migrate.go
eleijonmarck: do we have a separate issue for what you are describing @JeremyLoy ? also; appreciate the nice comment! 👯 

if not; would you mind adding one?

JeremyLoy: > do we have a separate issue for what you are describing

not yet, but I will make one. I need to do a bit more research to see how many different things are diverging, and what the scope would be.

PR #2509, opened 2025-06-17
Feat/list objects shadow mode

cikasfm:
# Description

Enable experimental list-objects optimizations in shadow mode.

## Summary

This PR introduces **shadow mode** support for the ListObjects query in OpenFGA. Shadow mode allows both the standard and optimized ListObjects query implementations to be executed in parallel, with their results compared and logged for analysis. This feature is intended to help validate the correctness and performance of the optimized implementation before fully enabling it in production.

## Key Changes

- Added `shadowedListObjectsQuery`, which wraps both standard and optimized ListObjects queries.
- Introduced configuration options for enabling shadow mode, sampling percentage, and timeout.
- Both standard and optimized queries are executed in parallel when shadow mode is enabled; results and latencies are logged and compared.
- No changes to the public API; shadow mode is opt-in via configuration.
- Added and updated unit tests to cover shadow mode scenarios and edge cases.
- Improved documentation for new and existing code related to ListObjects queries.

## Motivation

Shadow mode provides a safe way to test and validate the optimized ListObjects query implementation in real-world scenarios, without impacting end users. It helps identify discrepancies and performance regressions before rolling out optimizations broadly.

## Testing

- Unit tests cover:
  - Shadow mode enabled/disabled
  - Error handling for both standard and optimized queries
  - Configuration edge cases

## Usage

To enable this feature, use the following server config:

```
	listObjectsOptions := []openfgaserver.OpenFGAServiceV1Option{
		openfgaserver.WithShadowListObjectsQueryEnabled(true),
		openfgaserver.WithShadowListObjectsQueryTimeout(10*time.Second),
		openfgaserver.WithShadowListObjectsQuerySamplePercentage(10),
	}
	openFGAServerOptions = append(openFGAServerOptions, listObjectsOptions...)
```

## Related Issues

- Enabling Weighted Graph use

## Checklist

- [x] Code compiles and passes all tests
- [x] Documentation updated
- [x] Unit tests added/updated

---

**Reviewer notes:**  
Please review the shadow mode logic, configuration handling, and test coverage. Feedback on logging and sampling approach is welcome.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced experimental shadow mode for list object queries, enabling concurrent comparison of standard and optimized query implementations without impacting production results.
  * Added configuration options for shadow mode, including enablement flag, sampling percentage, timeout, and maximum delta items for result comparison.
  * Enabled experimental list object optimizations via the `enable-list-objects-optimizations` flag.

* **Bug Fixes**
  * Enhanced validation of shadow mode configuration parameters to enforce valid ranges and prevent misconfiguration.

* **Documentation**
  * Updated changelog to document the addition of experimental list object optimizations in shadow mode.

* **Tests**
  * Added extensive tests covering shadow mode functionality, configuration validation, concurrency handling, and result comparison logic.

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -107,27 +107,6 @@ type ListObjectsResponse struct {
 	ResolutionMetadata ListObjectsResolutionMetadata
 }
 
-func (r *ListObjectsResponse) Clone() ListObjectsResponse {
-	// Create a new instance of ListObjectsResponse
-	clone := ListObjectsResponse{
-		Objects:            make([]string, len(r.Objects)),
-		ResolutionMetadata: NewListObjectsResolutionMetadata(),
-	}
-
-	// Copy the objects slice
-	copy(clone.Objects, r.Objects)
-
-	return clone
-}
-
-func (m ListObjectsResolutionMetadata) Clone() ListObjectsResolutionMetadata {
-	metadata := NewListObjectsResolutionMetadata()
-	metadata.DatastoreQueryCount.Store(m.DatastoreQueryCount.Load())
-	metadata.DispatchCounter.Store(m.DispatchCounter.Load())
-	metadata.WasThrottled.Store(m.WasThrottled.Load())
-	return metadata
-}
-
 type ListObjectsQueryOption func(d *ListObjectsQuery)
 
 func WithListObjectsDeadline(deadline time.Duration) ListObjectsQueryOption {
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -68,6 +68,19 @@ type ListObjectsQuery struct {
 	checkResolver            graph.CheckResolver
 	cacheSettings            serverconfig.CacheSettings
 	sharedDatastoreResources *shared.SharedDatastoreResources
+
+	listObjectsOptimizationEnabled bool // Indicates if experimental optimizations are enabled for ListObjectsResolver
+}
+
+type ListObjectsResolver interface {
+	// Execute the ListObjectsQuery, returning a list of object IDs up to a maximum of q.listObjectsMaxResults
+	// or until q.listObjectsDeadline is hit, whichever happens first.
+	Execute(ctx context.Context, req *openfgav1.ListObjectsRequest) (*ListObjectsResponse, error)
+
+	// ExecuteStreamed executes the ListObjectsQuery, returning a stream of object IDs.
+	// It ignores the value of q.listObjectsMaxResults and returns all available results
+	// until q.listObjectsDeadline is hit.
+	ExecuteStreamed(ctx context.Context, req *openfgav1.StreamedListObjectsRequest, srv openfgav1.OpenFGAService_StreamedListObjectsServer) (*ListObjectsResolutionMetadata, error)
 }
 
 type ListObjectsResolutionMetadata struct {
@@ -81,8 +94,8 @@ type ListObjectsResolutionMetadata struct {
 	WasThrottled *atomic.Bool
 }
 
-func NewListObjectsResolutionMetadata() *ListObjectsResolutionMetadata {
-	return &ListObjectsResolutionMetadata{
+func NewListObjectsResolutionMetadata() ListObjectsResolutionMetadata {
+	return ListObjectsResolutionMetadata{
 		DatastoreQueryCount: new(atomic.Uint32),
 		DispatchCounter:     new(atomic.Uint32),
 		WasThrottled:        new(atomic.Bool),
@@ -92,6 +105,27 @@ func NewListObjectsResolutionMetadata() *ListObjectsResolutionMetadata {
 type ListObjectsResponse struct {
 	Objects            []string
 	ResolutionMetadata ListObjectsResolutionMetadata
+}
+
+func (r *ListObjectsResponse) Clone() ListObjectsResponse {
+	// Create a new instance of ListObjectsResponse
+	clone := ListObjectsResponse{
+		Objects:            make([]string, len(r.Objects)),
+		ResolutionMetadata: NewListObjectsResolutionMetadata(),
+	}
+
+	// Copy the objects slice
+	copy(clone.Objects, r.Objects)
+
+	return clone
+}
+
+func (m ListObjectsResolutionMetadata) Clone() ListObjectsResolutionMetadata {
+	metadata := NewListObjectsResolutionMetadata()
+	metadata.DatastoreQueryCount.Store(m.DatastoreQueryCount.Load())
+	metadata.DispatchCounter.Store(m.DispatchCounter.Load())
+	metadata.WasThrottled.Store(m.WasThrottled.Load())
+	return metadata
 }
 
 type ListObjectsQueryOption func(d *ListObjectsQuery)
@@ -152,6 +186,12 @@ func WithListObjectsDatastoreThrottler(threshold int, duration time.Duration) Li
 	return func(d *ListObjectsQuery) {
 		d.datastoreThrottleThreshold = threshold
 		d.datastoreThrottleDuration = duration
+	}
+}
+
+func WithListObjectsOptimizationEnabled(enabled bool) ListObjectsQueryOption {
+	return func(d *ListObjectsQuery) {
+		d.listObjectsOptimizationEnabled = enabled
 	}
 }
 
@@ -456,7 +496,7 @@ func (q *ListObjectsQuery) Execute(
 		// Kick off background job to check if cache records are stale, inavlidating where needed
 		q.sharedDatastoreResources.CacheController.InvalidateIfNeeded(ctx, req.GetStoreId())
 	}
-	err := q.evaluate(timeoutCtx, req, resultsChan, maxResults, resolutionMetadata)
+	err := q.evaluate(timeoutCtx, req, resultsChan, maxResults, &resolutionMetadata)
 	if err != nil {
 		return nil, err
 	}
@@ -488,7 +528,7 @@ func (q *ListObjectsQuery) Execute(
 
 	return &ListObjectsResponse{
 		Objects:            objects,
-		ResolutionMetadata: *resolutionMetadata,
+		ResolutionMetadata: resolutionMetadata,
 	}, nil
 }
 
@@ -509,7 +549,7 @@ func (q *ListObjectsQuery) ExecuteStreamed(ctx context.Context, req *openfgav1.S
 
 	resolutionMetadata := NewListObjectsResolutionMetadata()
 
-	err := q.evaluate(timeoutCtx, req, resultsChan, maxResults, resolutionMetadata)
+	err := q.evaluate(timeoutCtx, req, resultsChan, maxResults, &resolutionMetadata)
 	if err != nil {
 		return nil, err
 	}
@@ -534,5 +574,5 @@ func (q *ListObjectsQuery) ExecuteStreamed(ctx context.Context, req *openfgav1.S
 		}
 	}
 
-	return resolutionMetadata, nil
+	return &resolutionMetadata, nil
 }
--- a/pkg/server/commands/list_objects_shadow.go
+++ b/pkg/server/commands/list_objects_shadow.go
@@ -1,0 +1,299 @@
+package commands
+
+import (
+	"context"
+	"errors"
+	"math/rand"
+	"slices"
+	"sync"
+	"time"
+
+	"go.uber.org/zap"
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/internal/graph"
+	"github.com/openfga/openfga/pkg/logger"
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+const ListObjectsShadowExecute = "ShadowedListObjectsQuery.Execute"
+
+type shadowedListObjectsQuery struct {
+	main          ListObjectsResolver
+	shadow        ListObjectsResolver
+	shadowPct     int           // An integer representing the shadowPct of list_objects requests that will also trigger the shadow query. This allows for controlled rollout and data collection without impacting all requests. Value should be between 0 and 100.
+	shadowTimeout time.Duration // A time.Duration specifying the maximum amount of time to wait for the shadow list_objects query to complete. If the shadow query exceeds this shadowTimeout, it will be cancelled, and its result will be ignored, but the shadowTimeout event will be logged.
+	maxDeltaItems int           // The maximum number of items to log in the delta between the main and shadow results. This prevents excessive logging in case of large differences.
+	logger        logger.Logger
+	random        *rand.Rand
+	// only used for testing signals
+	wg *sync.WaitGroup
+}
+
+type ShadowListObjectsQueryOption func(d *ShadowListObjectsQueryConfig)
+
+// WithShadowListObjectsQueryEnabled sets whether the shadow list_objects query should use optimizations.
+func WithShadowListObjectsQueryEnabled(enabled bool) ShadowListObjectsQueryOption {
+	return func(c *ShadowListObjectsQueryConfig) {
+		c.shadowEnabled = enabled
+	}
+}
+
+// WithShadowListObjectsQuerySamplePercentage sets the shadowPct of list_objects requests that will trigger the shadow query.
+func WithShadowListObjectsQuerySamplePercentage(samplePercentage int) ShadowListObjectsQueryOption {
+	return func(c *ShadowListObjectsQueryConfig) {
+		c.shadowPct = samplePercentage
+	}
+}
+
+// WithShadowListObjectsQueryTimeout sets the shadowTimeout for the shadow list_objects query.
+func WithShadowListObjectsQueryTimeout(timeout time.Duration) ShadowListObjectsQueryOption {
+	return func(c *ShadowListObjectsQueryConfig) {
+		c.shadowTimeout = timeout
+	}
+}
+
+func WithShadowListObjectsQueryLogger(logger logger.Logger) ShadowListObjectsQueryOption {
+	return func(c *ShadowListObjectsQueryConfig) {
+		c.logger = logger
+	}
+}
+
+func WithShadowListObjectsQueryMaxDeltaItems(maxDeltaItems int) ShadowListObjectsQueryOption {
+	return func(c *ShadowListObjectsQueryConfig) {
+		c.maxDeltaItems = maxDeltaItems
+	}
+}
+
+type ShadowListObjectsQueryConfig struct {
+	shadowEnabled bool          // A boolean flag to globally enable or disable the shadow mode for list_objects queries. When false, the shadow query will not be executed.
+	shadowPct     int           // An integer representing the shadowPct of list_objects requests that will also trigger the shadow query. This allows for controlled rollout and data collection without impacting all requests. Value should be between 0 and 100.
+	shadowTimeout time.Duration // A time.Duration specifying the maximum amount of time to wait for the shadow list_objects query to complete. If the shadow query exceeds this shadowTimeout, it will be cancelled, and its result will be ignored, but the shadowTimeout event will be logged.
+	maxDeltaItems int           // The maximum number of items to log in the delta between the main and shadow results. This prevents excessive logging in case of large differences.
+	logger        logger.Logger
+}
+
+func NewShadowListObjectsQueryConfig(opts ...ShadowListObjectsQueryOption) *ShadowListObjectsQueryConfig {
+	result := &ShadowListObjectsQueryConfig{
+		shadowEnabled: false,                  // Disabled by default
+		shadowPct:     0,                      // Default to 0% to disable shadow mode
+		shadowTimeout: 1 * time.Second,        // Default shadowTimeout for shadow queries
+		logger:        logger.NewNoopLogger(), // Default to a noop logger
+		maxDeltaItems: 100,                    // Default max delta items to log
+	}
+	for _, opt := range opts {
+		opt(result)
+	}
+	return result
+}
+
+// NewListObjectsQueryWithShadowConfig creates a new ListObjectsResolver that can run in shadow mode based on the provided ShadowListObjectsQueryConfig.
+func NewListObjectsQueryWithShadowConfig(
+	ds storage.RelationshipTupleReader,
+	checkResolver graph.CheckResolver,
+	shadowConfig *ShadowListObjectsQueryConfig,
+	opts ...ListObjectsQueryOption,
+) (ListObjectsResolver, error) {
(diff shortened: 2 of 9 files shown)


pkg/server/commands/list_objects.go
```diff
@@ -52,7 +52,7 @@ var (
 	})
 )
 
-type ListObjectsQuery struct {
+type listObjectsQuery struct {
```
senojj: Why not export this type?

pkg/server/commands/list_objects.go
```diff
@@ -95,72 +108,86 @@ type ListObjectsResponse struct {
...
+		d.listObjectsOptimizationEnabled = enabled
+	}
+}
+
 func NewListObjectsQuery(
 	ds storage.RelationshipTupleReader,
 	checkResolver graph.CheckResolver,
 	opts ...ListObjectsQueryOption,
-) (*ListObjectsQuery, error) {
+) (ListObjectsQuery, error) {
```
senojj: Why return an interface instead of the concrete type?

pkg/server/commands/list_objects.go
cikasfm: @senojj the goal is to "relieve" the client of need to know the internals of the object returned, whether it's a `ListObjectQuery` or `ShadowedListObjectsQuery` and just work with the interface.

The reason for that is that based on config, a different object type can be returned in `NewListObjectsQueryWithShadowConfig` and for consistency I'm returning the same in `NewListObjectsQuery`.

pkg/server/commands/list_objects.go
cikasfm: @senojj the goal is to "relieve" the client of need to know the internals of the object returned, whether it's a `ListObjectQuery` or `ShadowedListObjectsQuery` and just work with the interface.

The reason for that is that based on config, a different object type can be returned in `NewListObjectsQueryWithShadowConfig` and for consistency I'm returning the same in `NewListObjectsQuery`.

pkg/server/commands/list_objects.go
senojj: The `listObjectsQuery` type has unexported fields, so the user cannot know the internals of the type, even when the type itself is exported.

`NewListObjectsQuery` only ever returns a `*listObjectsQuery` so there is no benefit to returning an interface from it. By returning an interface from this function, you prematurely constrain how the value may be useful to the user.

However, if you have function that can accept as an argument or return multiple types, polymorphism makes total sense.

pkg/server/commands/list_objects.go
senojj: The key thing to note here is that this function will not ever return more than one type. Polymorphism serves no purpose here.

It makes sense for `NewListObjectsQueryWithShadowConfig`, but not here. Applying consistency only for the sake of consistency has no notable value.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,239 @@
...
+	return func(c *ShadowListObjectsQueryConfig) {
+		c.enabled = enabled
+	}
+}
+
+// WithShadowListObjectsQuerySamplePercentage sets the percentage of list_objects requests that will trigger the shadow query.
+func WithShadowListObjectsQuerySamplePercentage(samplePercentage int) ShadowListObjectsQueryOption {
+	return func(c *ShadowListObjectsQueryConfig) {
+		if samplePercentage < 0 || samplePercentage > 100 {
+			log.Fatalf("Invalid value for sample percentage value=%d, must be between 0 and 100", samplePercentage)
```
adriantam: instead of crash and burn here, should we move the validation to when config is specified and assume that it is valid by the time we hit here?

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,239 @@
...
+		defer wg.Done()
+		defer func() {
+			if r := recover(); r != nil {
+				err2 = fmt.Errorf("panic in fn2: %v", r)
+			}
+		}()
+		result2, err2 = fn2()
+		latency2 = time.Since(start)
+	}()
+	wg.Wait()
```
adriantam: Instead of waiting for both the regular and the shadowed result to complete, would it make sense to wait for the regular to complete and then run the shadow in go func?  See https://github.com/openfga/openfga/blob/main/internal/graph/shadow_resolver.go

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	if errOptimized != nil {
+		q.logger.Error("shadowed list objects error",
+			zap.String("func", fnName),
+			zap.Error(errOptimized))
+		return result, nil
+	}
+
+	q.logger.Info("shadowed list objects result",
```
adriantam: Instead of always logging the shadow result, should we only flag when there are deltas?  See https://github.com/openfga/openfga/blob/90fdaeb6b5a53c7a4089cfac4b01ccd60e738c82/internal/graph/shadow_resolver.go#L97

adriantam: With that said, we also want to highlight if the number of items hit max limit / latency  / latencyOptimized is near the list objects timeout because those case will likely produce non-identical result

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,239 @@
...
+		return q.standard.Execute(ctx, req)
+	}
+	var fnOptimized = func(ctx context.Context) (*ListObjectsResponse, error) {
+		return q.optimized.Execute(ctx, req)
+	}
+
+	return executeShadowMode(ctx, q, "Execute", fnStandard, fnOptimized)
+}
+
+func (q *shadowedListObjectsQuery) ExecuteStreamed(ctx context.Context, req *openfgav1.StreamedListObjectsRequest, srv openfgav1.OpenFGAService_StreamedListObjectsServer) (*ListObjectsResolutionMetadata, error) {
```
adriantam: ExecuteStreamed probably should not run in shadow mode because the result will inherently be different because list objects item will not return in predictable order

pkg/server/commands/list_objects_shadow.go
cikasfm: We're only waiting for the standard to complete, then cancel context for the shadow one - so it should stop running any DB queries & return immediatelly.

pkg/server/commands/list_objects_shadow.go
justincoh: +1 to this, you could add something [here](https://github.com/openfga/openfga/blob/main/pkg/server/config/config.go#L400)

pkg/server/commands/list_objects_shadow.go
justincoh: +1 to Adrian's point. If standard ListObjects errors, hits the timeout, or returns `maxResults`, then there's no value in running the shadow version for comparison, so it would be wasted effort even if we bail out early.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,239 @@
...
+		zap.Duration("latencyOptimized", latencyOptimized))
+
+	return result, nil
+}
+
+// runInParallel helper to run two functions in parallel and collect their results and latencies.
+func runInParallel[T any](
+	fn1 func() (T, error),
+	fn2 func() (T, error),
+) (latency1, latency2 time.Duration, result1, result2 T, err1, err2 error) {
```
justincoh: That's a lot of return values. Can we do the comparisons in here?

pkg/server/commands/list_objects_shadow.go
elbuo8: I'm confused, why would the successful "standard" cancel the shadow? It's not meant to be a race, and you def want to compute their result sets 🤔

pkg/server/commands/list_objects_shadow.go
elbuo8: Can't say I disagree with this 🤔

pkg/server/commands/list_objects_shadow.go
cikasfm: IMHO we're hoping for the optimized to be faster, so we're not waiting for a slower result and ignoring it.

@adriantam do you think we should run "run & forget" the shadow impl once the current is completed? My thought was to run both in parallel & then compare the results. Let's discuss

cikasfm:
@coderabbitai full review

pkg/server/commands/list_objects.go
cikasfm: https://martinfowler.com/bliki/TwoHardThings.html

pkg/server/commands/list_objects.go
cikasfm: https://martinfowler.com/bliki/TwoHardThings.html

pkg/server/commands/list_objects.go
```diff
@@ -94,6 +107,19 @@ type ListObjectsResponse struct {
 	ResolutionMetadata ListObjectsResolutionMetadata
 }
 
+func (r *ListObjectsResponse) Clone() *ListObjectsResponse {
+	// Create a new instance of ListObjectsResponse
+	clone := &ListObjectsResponse{
+		Objects:            make([]string, len(r.Objects)),
+		ResolutionMetadata: *(NewListObjectsResolutionMetadata()),
```
senojj: I know this isn't part of your code changes. But does it really make sense for `NewListObjectsResolutionMetadata` to return a pointer? `ListObjectsResolutionMetadata` is a struct containing 3 pointers, which is pretty small, and there is nothing that would prevent it from being copied properly.

Here, the `ListObjectsResolutionMetadata` value is escaping to the heap in the call to `NewListObjectsResolutionMetadata()`, and then it is being dereferenced immediately, so it is copied, and added to a `ListObjectsResponse` instance, which escapes to the heap in `Clone()`.... In reality, none of these values need to be pointers and can exist on the stack perfectly well.

cikasfm: makes sense

pkg/server/commands/list_objects.go
```diff
@@ -155,10 +181,24 @@ func WithListObjectsDatastoreThrottler(threshold int, duration time.Duration) Li
...
+		d.listObjectsOptimizationEnabled = enabled
+	}
+}
+
 func NewListObjectsQuery(
 	ds storage.RelationshipTupleReader,
 	checkResolver graph.CheckResolver,
 	opts ...ListObjectsQueryOption,
+) (*ListObjectsQuery, error) {
+	return newListObjectsQuery(ds, checkResolver, opts...)
```
adriantam: curious why do we need to have `newListObjectsQuery`?

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+	checkResolver graph.CheckResolver,
+	shadowConfig *ShadowListObjectsQueryConfig,
+	opts ...ListObjectsQueryOption,
+) (ListObjectsResolver, error) {
+	if shadowConfig == nil {
+		return nil, errors.New("shadowConfig must be set")
+	}
+	standard, err := newListObjectsQuery(ds, checkResolver,
+		// force disable optimizations
+		slices.Concat(opts, []ListObjectsQueryOption{WithListObjectsOptimizationEnabled(false)})...,
```
adriantam, line 117: I know what you are trying to do...but a bit weird to force optimization to be false.  The concern is that if someone enabled shadow mode but set the percent to 0 (thinking this is the same thing as disable the shadow mode), we will force the optimization to be off.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+	// so it is impossible to compare the results
+	if loq, ok := q.main.(*ListObjectsQuery); ok {
+		if len(res.Objects) == int(loq.listObjectsMaxResults) {
+			q.logger.DebugWithContext(parentCtx, "shadowed list objects query skipped due to max results reached",
+				// common args
+				commonFields...,
+			)
+			return
+		}
+	}
```
adriantam, line 214: we also want to skip if the initial query takes close to listObjectsDeadline as this indicates we will not get the complete result.

pkg/server/commands/list_objects.go
```diff
@@ -94,6 +107,19 @@ type ListObjectsResponse struct {
 	ResolutionMetadata ListObjectsResolutionMetadata
 }
 
+func (r *ListObjectsResponse) Clone() *ListObjectsResponse {
```
senojj: Why does the `Clone` method need to return a pointer?

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		return obj, err
+	}
+	var cloned T
+	err = json.Unmarshal(marshalled, &cloned) // Unmarshal back to the original type.
+	if err != nil {
+		return obj, err
+	}
+	return cloned, nil
+}
```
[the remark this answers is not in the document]
senojj: I agree with this comment

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+		go func() {
+			q.executeShadowModeAndCompareResults(cloneCtx, cloneReq, cloneRes)
+			q.wg.Done() // only used for testing signals
+		}()
+	}
+	return res, err
+}
+
+// cloneObj is a utility function that clones an object of type T using JSON marshalling and unmarshalling.
+func cloneObj[T any](obj T) (T, error) {
```
senojj: Why are you cloning via JSON serialization? This is extremely inefficient.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+		zap.String("store_id", req.GetStoreId()),
+		zap.String("model_id", req.GetAuthorizationModelId()),
+	}
+	defer func() {
+		if r := recover(); r != nil {
+			q.logger.ErrorWithContext(parentCtx, "panic recovered",
+				append(commonFields, zap.Any("error", r))...,
+			)
+		}
+	}()
```
senojj, line 201: 1. Is it ok for the application to continue after a panic?
2. `recover` is an expensive operation. Do you expect a panic somewhere in this function?

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+
+	if errShadow != nil {
+		q.logger.WarnWithContext(parentCtx, "shadowed list objects error",
+			append(commonFields, zap.Any("error", errShadow))...,
+		)
+		return
+	}
+
+	objects := copyAndSortArr(res.Objects)
+	shadowObj := []string{}
```
senojj: You can avoid an allocation here by initializing `shadowObj` to a `nil` value.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+	if shadowRes != nil {
+		shadowObj = copyAndSortArr(shadowRes.Objects)
+	}
+	// compare sorted string arrays - sufficient for equality check
+	if !slices.Equal(objects, shadowObj) {
```
senojj, line 247: is the point of this to test that `objects` was sorted to begin with?

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+}
+
+// calculateDelta compares two slices of strings and returns a slice of strings representing the delta.
+// Each string in the delta is prefixed with "+" if it exists in the shadow but not in the main,
+// or "-" if it exists in the main but not in the shadow.
+// Used for logging differences between the main and shadow results in the shadowed list objects query.
+func calculateDelta(inputMain []string, inputShadow []string) []string {
+	mainObjects := make(map[string]bool)
+	shadowObjects := make(map[string]bool)
```
senojj: Not critical, but you can make these maps more space efficient by making them `map[string]struct{}`

pkg/server/commands/list_objects.go
cikasfm: Mostly because the pointer is used as argument in all places

pkg/server/commands/list_objects.go
cikasfm: no need any more after 10x refactoring...

pkg/server/commands/list_objects_shadow.go
cikasfm: I agree! Maybe we don't need to clone the request at all. Otherwise, the stack of code required just to clone would be more than I think it's valuable, just for shadow mode

pkg/server/commands/list_objects_shadow.go
cikasfm: obviously it's only for unexpected panics. The goal here is not to kill the app just because something failed in shadow mode.

pkg/server/commands/list_objects_shadow.go
cikasfm: @adriantam I agree, but we need to define what is `initial query takes close to listObjectsDeadline`. Is it within 5%? Or within 5ms? Or do we want to make it configurable? Or does it even matter? At this point if the original query was very slow then we're probably timing out & failing with Context Deadline Exceeded, in which case we would not even run the shadow mode. For successful but slow - IMHO we want to see the performance gains. Do we really want to check if it "almost exceeded deadline"?

pkg/server/commands/list_objects.go
senojj: By doing that, you are forcing the cloned value to escape to the heap. If you were to return a value and reference the value when passing it into a function, you could avoid a heap allocation in scenarios that don't require the lifetime of the value to extend beyond the lifetime of the current stack frame.

I realize that pointers are currently used prolifically throughout this project, but we should be more mindful about when we force heap allocations.

pkg/server/commands/list_objects_shadow.go
cikasfm: Good point! But also...that sounds like a valid use case to me. Is there a point to have "main" run optimizations enabled AND a shadow mode, with the same optimization enabled?

pkg/server/commands/list_objects_shadow.go
adriantam: As discussed offline, we can make it within 100ms.  The reason we want to do this is that list objects will not return erorr, but instead incomplete response.  This can be done as a follow up PR

.golangci.yaml
```diff
@@ -53,10 +53,6 @@ linters:
     staticcheck:
       checks:
         - all
-    revive:
```
adriantam: why?

cikasfm: oh, wait. I don't think this was intentional

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,299 @@
...
+	shadowEnabled bool          // A boolean flag to globally enable or disable the shadow mode for list_objects queries. When false, the shadow query will not be executed.
+	shadowPct     int           // An integer representing the shadowPct of list_objects requests that will also trigger the shadow query. This allows for controlled rollout and data collection without impacting all requests. Value should be between 0 and 100.
+	shadowTimeout time.Duration // A time.Duration specifying the maximum amount of time to wait for the shadow list_objects query to complete. If the shadow query exceeds this shadowTimeout, it will be cancelled, and its result will be ignored, but the shadowTimeout event will be logged.
+	maxDeltaItems int           // The maximum number of items to log in the delta between the main and shadow results. This prevents excessive logging in case of large differences.
+	logger        logger.Logger
+}
+
+func NewShadowListObjectsQueryConfig(opts ...ShadowListObjectsQueryOption) *ShadowListObjectsQueryConfig {
+	result := &ShadowListObjectsQueryConfig{
+		shadowEnabled: false,                  // Disabled by default
```
adriantam, line 79: nitpik (non-blocking): ideally this will share the default config instead of hard code value

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,299 @@
...
+						zap.Any("request", req),
+						zap.String("store_id", req.GetStoreId()),
+						zap.String("model_id", req.GetAuthorizationModelId()),
+						zap.Any("error", r),
+					)
+				}
+			}()
+			defer q.wg.Done() // only used for testing signals
+
+			cloneRes := res.Clone()
```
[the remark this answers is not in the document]
senojj: @cikasfm did you intend to zero-out the resolution metadata on clone?

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil, err
+	}
+
+	result := &shadowedListObjectsQuery{
+		main:          standard,
+		shadow:        optimized,
+		shadowPct:     shadowConfig.shadowPct,
+		shadowTimeout: shadowConfig.shadowTimeout,
+		logger:        shadowConfig.logger,
+		random:        rand.New(rand.NewSource(time.Now().UnixNano())),
```
adriantam: why do we need to do this?  I thought that we don't need to seed anymore - https://pkg.go.dev/math/rand#Seed.
See https://go.dev/blog/randv2

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	// compare sorted string arrays - sufficient for equality check
+	if !slices.Equal(objects, shadowObj) {
+		delta := calculateDelta(objects, shadowObj)
+		if len(delta) > q.maxDeltaItems {
+			delta = delta[:q.maxDeltaItems] // limit the delta to maxDeltaItems
+		}
+		// log the differences if the shadow query failed or if the results are not equal
+		q.logger.InfoWithContext(parentCtx, "shadowed list objects result difference",
+			append(commonFields,
```
adriantam, line 245: probably good to log the number of delta items

pkg/server/commands/list_objects_shadow.go
```diff
@@ -0,0 +1,294 @@
...
+	shadowRes, errShadow := q.shadow.Execute(shadowCtx, req)
+
+	if errShadow != nil {
+		q.logger.WarnWithContext(parentCtx, "shadowed list objects error",
+			append(commonFields, zap.Any("error", errShadow))...,
+		)
+		return
+	}
+
+	objects := copyAndSortArr(mainResult)
```
adriantam: Since we are using the map to calculate the delta anyways, would it make sense to just use the map?  (I.e., no need to do duplicate work of sorting)

pkg/server/commands/list_objects_shadow.go
cikasfm: I think this was suggested by `@coderabbitai`; I guess we really don't

adriantam (approved):

PR #2518, opened 2025-06-19
feat: typesystem edges for intersection and exclusion

adriantam:
## Description
In preparation of list objects optimization via weighted graph and optimization of list objects with intersection and exclusion, add function in typesystem to return the lowest weight edge for intersection and exclusion.

## Review Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced advanced handling of weighted authorization graphs, enabling more precise management of intersection and exclusion relationships.
- **Tests**
  - Added comprehensive unit tests to validate weighted graph functionalities, ensuring correct edge and node retrieval and classification in complex scenarios.
- **Refactor**
  - Streamlined internal logic by removing redundant interfaces and helper functions related to weighted graph processing.

### DIFF
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -54,18 +54,20 @@ func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sour
 	for _, edge := range edges {
 		weight, ok := edge.GetWeight(sourceType)
 		if !ok {
+			// can ignore this edge as there is no path
 			continue
 		}
-		if edge.GetEdgeType() == graph.DirectEdge {
-			// The first step is to establish the largest weight for all the direct edges as they need to be
-			// treated as a group. This weight will serve as the baseline for other siblings to compare against.
-			directEdges = append(directEdges, edge)
-			if weight > lowestWeight {
-				directEdgesAreLowest = true
-				lowestWeight = weight
-			}
-		} else {
+		if edge.GetEdgeType() != graph.DirectEdge {
+			// the nonDirectEdges will be handled as next step
 			nonDirectEdges = append(nonDirectEdges, edge)
+			continue
+		}
+		// We want to establish the largest weight for all the direct edges as they need to be
+		// treated as a group. This weight will serve as the baseline for other siblings to compare against.
+		directEdges = append(directEdges, edge)
+		if weight > lowestWeight {
+			directEdgesAreLowest = true
+			lowestWeight = weight
 		}
 	}
 
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -63,7 +63,8 @@ func (t *TypeSystem) GetEdgesForIntersection(intersectionNode *graph.WeightedAut
 				// The first step is to establish the largest weight for all the direct edges as they need to be
 				// treated as a group. This weight will serve as the baseline for other siblings to compare against.
 				directEdges = append(directEdges, edge)
-				if weight, ok := edge.GetWeight(sourceType); ok && weight > lowestWeight {
+				// note that the weight must be defined because of hasPathTo
+				if weight, _ := edge.GetWeight(sourceType); weight > lowestWeight {
 					directEdgesAreLowest = true
 					lowestWeight = weight
 				}
@@ -76,7 +77,7 @@ func (t *TypeSystem) GetEdgesForIntersection(intersectionNode *graph.WeightedAut
 				// the parent node already.
 				return IntersectionEdges{}, nil
 			}
-			if weight, ok := edge.GetWeight(sourceType); ok && (weight < lowestWeight || edgeNum == 0) {
+			if weight, _ := edge.GetWeight(sourceType); weight < lowestWeight || edgeNum == 0 {
 				// in the case of the first edge, we need to initialize as the least weight
 				// for other edges to compare against it.
 				// For non-first edges, only replace if it has the lowest weight.
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -95,23 +95,22 @@ func (t *TypeSystem) GetEdgesForIntersection(intersectionUniqueLabel string, sou
 					directEdgesAreLowest = true
 					lowestWeight = weight
 				}
-			} else { // non-direct edges
-				if lowestWeight == 0 && edgeNum != 0 {
-					// this means that all the direct edges are not connected.
-					// In reality, should not happen because the caller should have trimmed
-					// the parent node already.
-					return IntersectionEdges{}, nil
-				}
-				if weight, ok := edge.GetWeight(sourceType); ok {
-					// in the case of the first edge, we need to initialize as the least weight
-					// for other edges to compare against it.
-					// For non-first edges, only replace if it has the lowest weight.
-					if weight < lowestWeight || edgeNum == 0 {
-						lowestEdge = edge
-						lowestWeight = weight
-						directEdgesAreLowest = false
-					}
-				}
+				continue
+			}
+			// non-direct edges
+			if lowestWeight == 0 && edgeNum != 0 {
+				// this means that all the direct edges are not connected.
+				// In reality, should not happen because the caller should have trimmed
+				// the parent node already.
+				return IntersectionEdges{}, nil
+			}
+			if weight, ok := edge.GetWeight(sourceType); ok && (weight < lowestWeight || edgeNum == 0) {
+				// in the case of the first edge, we need to initialize as the least weight
+				// for other edges to compare against it.
+				// For non-first edges, only replace if it has the lowest weight.
+				lowestEdge = edge
+				lowestWeight = weight
+				directEdgesAreLowest = false
 			}
 		}
 	}
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -56,8 +56,11 @@ func (t *TypeSystem) GetEdgesForIntersection(intersectionNode *graph.WeightedAut
 	directEdgesAreLowest := false
 	lowestWeight := 0
 
+	// Process all edges first to separate direct from non-direct
+	nonDirectEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
+
 	// It is assumed that direct edge always appear first.
-	for edgeNum, edge := range edges {
+	for _, edge := range edges {
 		if hasPathTo(edge, sourceType) {
 			if edge.GetEdgeType() == graph.DirectEdge {
 				// The first step is to establish the largest weight for all the direct edges as they need to be
@@ -68,23 +71,32 @@ func (t *TypeSystem) GetEdgesForIntersection(intersectionNode *graph.WeightedAut
 					directEdgesAreLowest = true
 					lowestWeight = weight
 				}
-				continue
+			} else {
+				nonDirectEdges = append(nonDirectEdges, edge)
 			}
-			// non-direct edges
-			if lowestWeight == 0 && edgeNum != 0 {
-				// this means that all the direct edges are not connected.
-				// In reality, should not happen because the caller should have trimmed
-				// the parent node already.
-				return IntersectionEdges{}, nil
-			}
-			if weight, _ := edge.GetWeight(sourceType); weight < lowestWeight || edgeNum == 0 {
-				// in the case of the first edge, we need to initialize as the least weight
-				// for other edges to compare against it.
-				// For non-first edges, only replace if it has the lowest weight.
-				lowestEdge = edge
-				lowestWeight = weight
-				directEdgesAreLowest = false
-			}
+		}
+	}
+
+	hasDirectEdges := true
+
+	// non-direct edges
+	if lowestWeight == 0 {
+		if edges[0].GetEdgeType() == graph.DirectEdge {
+			// this means that all the direct edges are not connected.
+			// In reality, should not happen because the caller should have trimmed
+			// the parent node already.
+			return IntersectionEdges{}, nil
+		}
+		hasDirectEdges = false
+	}
+	for edgeNum, edge := range nonDirectEdges {
+		if weight, _ := edge.GetWeight(sourceType); weight < lowestWeight || (edgeNum == 0 && !hasDirectEdges) {
+			// in the case of the first edge, we need to initialize as the least weight
+			// for other edges to compare against it.
+			// For non-first edges, only replace if it has the lowest weight.
+			lowestEdge = edge
+			lowestWeight = weight
+			directEdgesAreLowest = false
 		}
 	}
 
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -55,10 +55,14 @@ func (t *TypeSystem) getNodeAndEdgesFromWeightedGraph(
 }
 
 type IntersectionEdges struct {
-	LowestEdge                *graph.WeightedAuthorizationModelEdge   // nil if direct is lowest, otherwise lowest edge
-	Siblings                  []*graph.WeightedAuthorizationModelEdge // all non lowest and excluding direct edges siblings
-	DirectEdges               []*graph.WeightedAuthorizationModelEdge // all the direct edges
-	DirectEdgesAreLeastWeight bool                                    // whether direct edges are the lowest weight edges
+	LowestEdge *graph.WeightedAuthorizationModelEdge   // nil if direct is lowest, otherwise lowest edge
+	Siblings   []*graph.WeightedAuthorizationModelEdge // all non lowest and excluding direct edges siblings
+	// We need to separate out the direct edges from the other edges so that in the case of direct edges are not the lowest weight,
+	// the check against these direct edges do not require checking against these individual directly assigned edges.
+	// Instead, these directly assigned edges should be treated as a group such that if one of these edges satisfy,
+	// the intersection satisfy.
+	DirectEdges               []*graph.WeightedAuthorizationModelEdge
+	DirectEdgesAreLeastWeight bool // whether direct edges are the lowest weight edges
 }
 
 // GetEdgesForIntersection returns the lowest weighted edge and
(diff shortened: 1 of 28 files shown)


pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,163 @@
...
+					// this means that all the direct edges are not connected.
+					// In reality, should not happen because the caller should have trimmed
+					// the parent node already.
+					return IntersectionEdges{}, nil
+				}
+				if weight, ok := edge.GetWeight(sourceType); ok {
+					// in the case of the first edge, we need to initialize as the least weight
+					// for other edges to compare against it.
+					// For non-first edges, only replace if it has the lowest weight.
+					if weight < lowestWeight || edgeNum == 0 {
```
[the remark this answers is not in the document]
Vic-Dev: I agree with coderabbitai here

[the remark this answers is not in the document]
adriantam: Actually, we meant to have edgeNum == 0 initialized as the minimum.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,163 @@
...
+						lowestEdge = edge
+						lowestWeight = weight
+						directEdgesAreLowest = false
+					}
+				}
+			}
+		}
+	}
+
```
Vic-Dev:
```suggestion
	lowestEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
	directEdgesAreLowest := false
	lowestWeight := 0

	// It is assumed that direct edge always appear first.
	for edgeNum, edge := range edges {
		if hasPathTo(edge, sourceType) {
			if edge.GetEdgeType() == graph.DirectEdge {
				// The first step is to establish the largest weight for all the direct edges as they need to be
				// treated as a group. This weight will serve as the baseline for other siblings to compare against.
				lowestEdges = append(lowestEdges, edge)
				if weight, ok := edge.GetWeight(sourceType); ok && weight > lowestWeight {
					directEdgesAreLowest = true
					lowestWeight = weight
				}
			} else { // non-direct edges
				if lowestWeight == 0 && edgeNum != 0 {
					// this means that all the direct edges are not connected.
					// In reality, should not happen because the caller should have trimmed
					// the parent node already.
					return IntersectionEdges{}, nil
				}
				if weight, ok := edge.GetWeight(sourceType); ok {
					// in the case of the first edge, we need to initialize as the least weight
					// for other edges to compare against it.
					// For non-first edges, only replace if it has the lowest weight.
					if weight < lowestWeight || len(lowestEdges) == 0 {
						lowestEdges = []*graph.WeightedAuthorizationModelEdge{edge}
						lowestWeight = weight
						directEdgesAreLowest = false
					}
				}
			}
		}
	}

```
We could just use a slice of length 1 for lowestWeight edges that aren't direct, then have the length be longer if direct edges are lowest. This gets rid of the need for a separate `DirectEdges` in the return value and we can just use the boolean to determine if the `LowestEdges` are direct or not

adriantam: If the direct edges are not the lowest weight, we want to separate them out so that when reverse expand call for check, the userset construction will only check for This() [i.e., direct assignment] instead of checking each individual direct assignment.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,170 @@
...
+		if hasPathTo(edge, sourceType) {
+			if edge.GetEdgeType() == graph.DirectEdge {
+				// The first step is to establish the largest weight for all the direct edges as they need to be
+				// treated as a group. This weight will serve as the baseline for other siblings to compare against.
+				directEdges = append(directEdges, edge)
+				if weight, ok := edge.GetWeight(sourceType); ok && weight > lowestWeight {
+					directEdgesAreLowest = true
+					lowestWeight = weight
+				}
+			} else { // non-direct edges
```
Vic-Dev: Can just use `continue` above instead of else to flatten the if statements a bit.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,170 @@
...
+				if lowestWeight == 0 && edgeNum != 0 {
+					// this means that all the direct edges are not connected.
+					// In reality, should not happen because the caller should have trimmed
+					// the parent node already.
+					return IntersectionEdges{}, nil
+				}
+				if weight, ok := edge.GetWeight(sourceType); ok {
```
Vic-Dev:
```suggestion
				if weight, ok := edge.GetWeight(sourceType); ok && weight < lowestWeight || edgeNum == 0 {
```

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,163 @@
...
+					directEdgesAreLowest = true
+					lowestWeight = weight
+				}
+			} else { // non-direct edges
+				if lowestWeight == 0 && edgeNum != 0 {
+					// this means that all the direct edges are not connected.
+					// In reality, should not happen because the caller should have trimmed
+					// the parent node already.
+					return IntersectionEdges{}, nil
+				}
```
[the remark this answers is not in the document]
adriantam: We don't want to return error here because the list objects should simply skip over the edge and become noop

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,163 @@
...
+	// weight to the sibling edges
+	for _, edge := range edges {
+		if edge.GetEdgeType() != graph.DirectEdge && edge != lowestEdge {
+			if hasPathTo(edge, sourceType) {
+				siblings = append(siblings, edge)
+			} else {
+				// In reality, should never happen because the edge should have been trimmed
+				return IntersectionEdges{}, nil
+			}
+		}
```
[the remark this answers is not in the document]
adriantam: We don't want to return error here because the list objects should simply skip over the edge and become noop

Vic-Dev (approved):

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, nil, nil
+	}
+
+	edges, ok := wg.GetEdgesFromNode(currentNode)
+	if !ok {
+		// This should never happen because it would have been filtered because there is no path.
+		return nil, nil, fmt.Errorf("no outgoing edges from node: %s", currentNode.GetUniqueLabel())
+	}
+	return edges, currentNode, nil
+}
```
senojj: how often do we do all of this together that it deserves its own function?

senojj: I see it is used in the two functions below.... why don't we just resolve these things outside of the functions, and pass them in as these returned types? That way we can avoid the potential for wasted cycles on redundant logic.

senojj: or are we trying to fully encapsulate the weighted graph within the typesystem?

adriantam: This will also be used by GetEdgesForListObjects as well...so 3 functions share the same logic.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,169 @@
...
+		if hasPathTo(edge, sourceType) {
+			if edge.GetEdgeType() == graph.DirectEdge {
+				// The first step is to establish the largest weight for all the direct edges as they need to be
+				// treated as a group. This weight will serve as the baseline for other siblings to compare against.
+				directEdges = append(directEdges, edge)
+				if weight, ok := edge.GetWeight(sourceType); ok && weight > lowestWeight {
+					directEdgesAreLowest = true
+					lowestWeight = weight
+				}
+				continue
```
justincoh: nit: `hasPathTo` is already checking for the `ok` part of this, you can drop that piece from the condition and just check the `weight`. Same thing a few lines below.

Alternatively, you could just use a plain old `GetWeight` at the start (rather than hasPathTo), so you'll have the the `weight` available for both blocks.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,169 @@
...
+	// weight to the sibling edges. Even if the directly assigned edges are not the lowest
+	// weight, we want to treat these directly assigned edges separately. The reason is that
+	// if one of these directly assigned edge satisfy, the list objects candidate may still
+	// be valid.
+	for _, edge := range edges {
+		if edge.GetEdgeType() != graph.DirectEdge && edge != lowestEdge {
+			if hasPathTo(edge, sourceType) {
+				siblings = append(siblings, edge)
+			} else {
+				// In reality, should never happen because the edge should have been trimmed
```
justincoh, line 103: Should we filter before returning from `getNodeAndEdgesFromWeightedGraph`? That way we could remove all `hasPathTo` checks from this function

adriantam: For now, I would say it is safer to validate here instead of trimming.  Otherwise, we may have falsely return list objects that should be dropped since type is not connected,

pkg/typesystem/weighted_graph.go
adriantam: Updated to resolve the node information outside of the function

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,153 @@
...
+	DirectEdges               []*graph.WeightedAuthorizationModelEdge
+	DirectEdgesAreLeastWeight bool // whether direct edges are the lowest weight edges
+}
+
+// GetEdgesForIntersection returns the lowest weighted edge and
+// its siblings edges for intersection based via the weighted graph.
+// If the direct edges have equal weight as its sibling edges, it will choose
+// the direct edges as preference.
+// If any of the children are not connected, it will return empty IntersectionEdges.
+func (t *TypeSystem) GetEdgesForIntersection(intersectionNode *graph.WeightedAuthorizationModelNode, sourceType string) (IntersectionEdges, error) {
```
senojj: It looks like this function doesn't actually _require_ the actual `*graph.WeightedAuthorizationModelNode` at all. It _really_ only requires the _edges_ from the node.

Can we eliminate the unnecessary dependency on the node, and just accept the edges as input?

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,153 @@
...
+		LowestEdge:                lowestEdge,
+		Siblings:                  siblings,
+		DirectEdges:               directEdges,
+		DirectEdgesAreLeastWeight: directEdgesAreLowest,
+	}, nil
+}
+
+// GetEdgesForExclusion returns the base edges (i.e., edge A in "A but not B") and
+// excluded edge (edge B in "A but not B") based on weighted graph for exclusion.
+func (t *TypeSystem) GetEdgesForExclusion(
```
senojj: Again, this function doesn't actually care about the node, just the edges from the node.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,140 @@
...
+	DirectEdges               []*graph.WeightedAuthorizationModelEdge
+	DirectEdgesAreLeastWeight bool // whether direct edges are the lowest weight edges
+}
+
+// GetEdgesForIntersection returns the lowest weighted edge and
+// its siblings edges for intersection based via the weighted graph.
+// If the direct edges have equal weight as its sibling edges, it will choose
+// the direct edges as preference.
+// If any of the children are not connected, it will return empty IntersectionEdges.
+func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sourceType string) (IntersectionEdges, error) {
```
elbuo8, line 39:
```suggestion
func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sourceType string) (*IntersectionEdges, error) {
```

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,140 @@
...
+	if len(edges) < 2 {
+		// Intersection by definition must have at least 2 children
+		return IntersectionEdges{}, fmt.Errorf("invalid edges for source type %s", sourceType)
```
elbuo8, line 42:
```suggestion
		return nil, fmt.Errorf("invalid edges for source type %s", sourceType)
```

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,140 @@
...
+	var lowestEdge *graph.WeightedAuthorizationModelEdge
+	directEdgesAreLowest := false
+	lowestWeight := 0
+
+	// Process all edges first to separate direct from non-direct
+	nonDirectEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
+
+	// It is assumed that direct edge always appear first.
+	for _, edge := range edges {
+		if hasPathTo(edge, sourceType) {
```
elbuo8: Use `continue` so that we don't have these deeply nested blocks.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,140 @@
...
+
+// hasPathTo returns a boolean indicating if a path exists from a node or edge to a terminal type. E.g
+// can we reach "user" from "document".
+func hasPathTo(nodeOrEdge weightedGraphItem, destinationType string) bool {
+	_, ok := nodeOrEdge.GetWeight(destinationType)
+	return ok
+}
+
+type IntersectionEdges struct {
+	LowestEdge *graph.WeightedAuthorizationModelEdge   // nil if direct is lowest, otherwise lowest edge
```
Vic-Dev, line 24: Seems like this comment isn't correct anymore

pkg/typesystem/weighted_graph.go
adriantam: I purposely return IntersectionEdges instead of *IntersectionEdges as IntersectionEdges is rather small and I want to avoid allocating from heap un-necessary.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,140 @@
...
+	sourceType string,
+) ([]*graph.WeightedAuthorizationModelEdge, *graph.WeightedAuthorizationModelEdge, error) {
+	if len(edges) < 2 {
+		return nil, nil, fmt.Errorf("invalid exclusion edges for source type %s", sourceType)
+	}
+
+	butNotEdge := edges[len(edges)-1] // this is the edge to 'b'
+	// Filter to only return edges which have a path to the sourceType
+	relevantEdges := slices.Collect(utils.Filter(edges[:len(edges)-1], func(edge *graph.WeightedAuthorizationModelEdge) bool {
+		return hasPathTo(edge, sourceType)
```
Vic-Dev, line 131: Do you need to do this here? `hasPathTo` was already called for intersection but not for exclusion?

pkg/typesystem/weighted_graph.go
adriantam: This is still valid

pkg/typesystem/weighted_graph.go
adriantam: I am not 100% what do you mean, we still do filtering for intersection as well

Vic-Dev: https://github.com/openfga/openfga/pull/2518/files/26b6765305f94919afb2c9111eda4484eb3ef129#diff-a64572bb776907d14cbd4914b4f2c900fa15ce65d3c77bcf445d958d71cf759aR60

elbuo8: It needs to happen in both since they are completely separate operands.

elbuo8 (dismissed):

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,138 @@
...
+	siblings := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges)-len(directEdges))
+
+	// Now, assign all the non directly assigned edges that are not the lowest
+	// weight to the sibling edges. Even if the directly assigned edges are not the lowest
+	// weight, we want to treat these directly assigned edges separately. The reason is that
+	// if one of these directly assigned edge satisfy, the list objects candidate may still
+	// be valid.
+	for _, edge := range edges {
+		if edge.GetEdgeType() != graph.DirectEdge && edge != lowestEdge {
+			if hasPathTo(edge, sourceType) {
```
elbuo8: `if !hasPathTo(edge, sourceType) {return IntersectionEdges{}, nil}`

Simplifies the logic a bit more.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Process all edges first to separate direct from non-direct
+	nonDirectEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
+
+	// It is assumed that direct edge always appear first.
+	for _, edge := range edges {
+		if !hasPathTo(edge, sourceType) {
+			continue
+		}
+		if edge.GetEdgeType() == graph.DirectEdge {
```
elbuo8:
```suggestion
		if edge.GetEdgeType() != graph.DirectEdge {
		  nonDirectEdges = append(nonDirectEdges, edge)
		  continue
		}
```

Then handle the non direct case.

pkg/typesystem/weighted_graph.go
```diff
@@ -0,0 +1,137 @@
...
+			directEdges = append(directEdges, edge)
+			// hasPathTo already verified the weight exists, so we can safely ignore ok
+			if weight, _ := edge.GetWeight(sourceType); weight > lowestWeight {
+				directEdgesAreLowest = true
+				lowestWeight = weight
+			}
+		} else {
+			nonDirectEdges = append(nonDirectEdges, edge)
+		}
+	}
```
Vic-Dev:
```suggestion
	for _, edge := range edges {
		weight, ok := edge.GetWeight(sourceType)
		if !ok {
		  continue
		}
		if edge.GetEdgeType() == graph.DirectEdge {
			// The first step is to establish the largest weight for all the direct edges as they need to be
			// treated as a group. This weight will serve as the baseline for other siblings to compare against.
			directEdges = append(directEdges, edge)
			if weight > lowestWeight {
				directEdgesAreLowest = true
				lowestWeight = weight
			}
		} else {
			nonDirectEdges = append(nonDirectEdges, edge)
		}
	}
```

Vic-Dev (approved):

PR #2522, opened 2025-06-20
fix: cache controller don't invalidate all the time

elbuo8:
## Review Checklist

### DIFF
--- a/internal/cachecontroller/cache_controller.go
+++ b/internal/cachecontroller/cache_controller.go
@@ -3,6 +3,7 @@ package cachecontroller
 import (
 	"context"
 	"math"
+	"strconv"
 	"sync"
 	"time"
 
@@ -16,7 +17,6 @@ import (
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 
 	"github.com/openfga/openfga/internal/build"
-	"github.com/openfga/openfga/internal/utils"
 	"github.com/openfga/openfga/pkg/logger"
 	"github.com/openfga/openfga/pkg/storage"
 	"github.com/openfga/openfga/pkg/telemetry"
@@ -52,7 +52,7 @@ var (
 		NativeHistogramBucketFactor:     1.1,
 		NativeHistogramMaxBucketNumber:  100,
 		NativeHistogramMinResetDuration: time.Hour,
-	}, []string{"invalidation_required", "changes_size"})
+	}, []string{"invalidation_required", "invalidation_type"})
 )
 
 type CacheController interface {
@@ -227,7 +227,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(ctx contex
 		}
 	}
 
-	if entry.LastModified.Before(timestampOfLastInvalidation) {
+	if entry.LastModified.Before(timestampOfLastInvalidation) || entry.LastModified.Equal(timestampOfLastInvalidation) {
 		// no new changes, no need to perform invalidations
 		span.SetAttributes(attribute.Bool("invalidations", false))
 		c.logger.Debug("InMemoryCacheController findChangesAndInvalidateIfNecessary no invalidation as entry.LastModified before last verified",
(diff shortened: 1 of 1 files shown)


internal/cachecontroller/cache_controller.go
```diff
@@ -227,15 +227,15 @@ func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(ctx contex
 		}
 	}
 
-	if entry.LastModified.Before(timestampOfLastInvalidation) {
+	if entry.LastModified.Before(timestampOfLastInvalidation) || entry.LastModified.Equal(timestampOfLastInvalidation) {
```
senojj: I think this is equivalent
```suggestion
	if !timestampOfLastInvalidation.After(entry.LastModified) {
```

internal/cachecontroller/cache_controller.go
```diff
@@ -286,7 +289,7 @@ func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(ctx contex
 		zap.Time("timestampOfLastIteratorInvalidation", timestampOfLastIteratorInvalidation),
 		zap.Bool("partialInvalidation", partialInvalidation))
 	span.SetAttributes(attribute.Bool("invalidations", true))
-	findChangesAndInvalidateHistogram.WithLabelValues("true", utils.Bucketize(uint(len(changes)), c.changelogBuckets)).Observe(float64(time.Since(start).Milliseconds()))
+	findChangesAndInvalidateHistogram.WithLabelValues(strconv.FormatBool(invalidated), strconv.FormatBool(partialInvalidation)).Observe(float64(time.Since(start).Milliseconds()))
```
senojj: looks like "invalidation_type" will currently have values: "none", "false", and "true"... should this be changed to: "none", "partial", and "full"?

senojj (approved):

internal/cachecontroller/cache_controller.go
```diff
@@ -252,25 +249,22 @@ func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(ctx contex
...
 	if idx == len(changes)-1 {
-		cacheInvalidationCounter.Inc()
-		partialInvalidation = false
+		invalidationType = "full"
 		c.invalidateIteratorCache(storeID)
 	} else {
 		// only a subset of changes are new, revoke the respective ones.
 		lastModified := time.Now()
 
 		// only increment if we're going to enter the invalidation for loop below
```
senojj: comment now irrelevant

senojj:
Regarding the `sync.WaitGroup` used for testing purposes. Another, potentially more usable solution, could be to add a function to the `CacheController` interface, like:
```go
Done() <-chan struct{}
```
Then, instead of having the `InMemoryCacheController` type contain a `sync.WaitGroup`, it could store a `chan struct{}`.

Within the `InvalidateIfNeeded` method, when all goroutines have finished, you would simply call to close the channel.

This may be useful for clients that need to select over multiple blocking operations:
```go
select {
    case <-ctx.Done():
    case <-cacheController.Done():
    case <-time.After(2 * time.Second):
}
```

internal/cachecontroller/cache_controller.go
```diff
@@ -162,84 +169,106 @@ func (c *InMemoryCacheController) findChangesDescending(ctx context.Context, sto
...
+		}
+		changes = msg.changes
+	}
+
+	lastChangelog := changes[0]
+	entry := &storage.ChangelogCacheEntry{
+		LastModified: lastChangelog.GetTimestamp().AsTime(),
+	}
+
+	defer c.cache.Set(cacheKey, entry, utils.JitterDuration(c.ttl, time.Minute)) // add buffer between checks
```
justincoh, line 256: Is this to spread the load? This cache entry is store-specific, so theoretically any bursts in traffic would still all encounter an expired record at the same time, no?

internal/cachecontroller/cache_controller.go
```diff
@@ -141,13 +143,17 @@
...
-		span.SetAttributes(attribute.Bool("cached", true))
-		return entry.LastModified
+		if entry, ok := cacheResp.(*storage.ChangelogCacheEntry); ok {
+			// the TTL grace period hasn't been breached
+			if entry.LastModified.Add(c.ttl).Before(time.Now()) {
+				cacheHitCounter.Inc()
+				span.SetAttributes(attribute.Bool("cached", true))
+				return entry.LastModified
+			}
+		}
```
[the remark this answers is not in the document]
senojj: @elbuo8 I think this suggestion is correct.
```
LastModified = 12:00
c.ttl = 5 minutes
time.Now() = 12:10

LastModified.Add(c.ttl) = 12:05

12:05 < 12:10 == true // Before logic thinks cache is still valid but is expired.
```

internal/cachecontroller/cache_controller.go
elbuo8: You mean the jitter? Not specifically. Just needed to add a buffer of time, didn't need to be as much as a second.

PR #2525, opened 2025-06-24
feat: typesystem to return userset for listobject

adriantam:
## Description
In anticipation of list objects optimization for intersection and exclusion, allow typesystem to construct userset for calling CheckRewrite for candidates directly inside reverse expand graph.

## Review Checklist

### DIFF
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"slices"
 
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/openfga/language/pkg/go/graph"
 
 	"github.com/openfga/openfga/internal/utils"
+	"github.com/openfga/openfga/pkg/tuple"
 )
 
 type weightedGraphItem interface {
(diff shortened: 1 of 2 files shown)


justincoh (commented):
Could you write out an example of where/how this will come into play? Or we could have a quick call

adriantam:
> Could you write out an example of where/how this will come into play? Or we could have a quick call

Updated in the description with sample code of how this will come into play.

pkg/typesystem/weighted_graph.go
```diff
@@ -137,3 +139,103 @@ func GetEdgesForExclusion(
...
+						Relation: relation,
+					},
+				},
+			}, nil
+		case graph.RewriteEdge, graph.ComputedEdge:
+			_, relation := tuple.SplitObjectRelation(uniqueLabel)
+			return &openfgav1.Userset{
+				Userset: &openfgav1.Userset_ComputedUserset{
+					ComputedUserset: &openfgav1.ObjectRelation{
+						Relation: relation,
```
justincoh, line 171: How come no object here?

justincoh (approved):

pkg/typesystem/weighted_graph.go
adriantam: For computed userset, only the relation is important.  See https://github.com/openfga/openfga/blob/73e5c5d19e5e395a322093116c5123712cc60074/internal/graph/check.go#L1323

PR #2529, opened 2025-06-26
feat: ReverseExpand with weighted graph, non-infinite weights

justincoh:
## Description

This PR adds a separate ReverseExpand path utilizing the weighted graph.

### How it works
1. Graph Traversal: Starting from the target `type#relation` received in the request, `loopOverWeightedEdges` traverses the weighted graph until it locates a leaf node, e.g. `[user]`.
2. Relation Stack: As the traversal progresses, it builds a relationStack that records the path of relationships encountered on the way to a leaf.
	* **Important**: each leaf node receives its own copy of the stack specific to the path taken to reach that leaf.
	* **Also Important**: the stack contains **only** entries which can be queried in the datastore. 
		* e.g. for a computed relation like `define rel1: rel2`, the stack will disregard rel1 since no tuple can ever be written against it, and instead use `type#rel2`. There is also logic adjusting for TTUs.

3. When a DirectEdge (a direct relationship to a user or wildcard) is encountered, the accumulated relationStack is passed to `queryForTuples`, which uses the relations it contains to query the datastore.
4. queryForTuples uses the stack built in the traversal to execute a series of database queries. It will continue running queries concurrently until it either a) cannot locate tuples or b) empties the stack. If queryForTuples manages to empty the stack, that means it successfully found a tuple at every level of the traversal, and the object it currently holds will be sent to ListObjects via a channel as a candidate for return.

### Simple Example:
```
type user
type repo
  relations
    define member: [user]
    define computed_member: member
    define owner: [user]
    define admin: [user] or computed_member
    define or_admin: owner or admin
```
We can reach 3 distinct leaf nodes. For illustration purposes, let's take the longest path to a leaf (ignoring another leaf along the way).

`repo#or_admin` -> `repo#admin` -> `repo#computed_member` -> `repo#member` -> `[user]`

Since the stack built during traversal stores only queryable relationships, the final stack once we reach `[user]` looks like this: `[{typeRel: repo#member}]`, since every other relationship along this branch encountered all `RewriteEdge` in the weighted graph. You can see from the model that this a `repo#member@user` tuple one tuple is enough to satisfy `repo#or_admin` without further relations.

Query for tuples then does the following:
* .Pop() the last relation off the stack (repo#member) and query for all repos matching `repo#member@user:justin`.
* The stack is now empty. Any repos found are candidates and will be sent out to ListObjects via channel.

### Example with Tuple to Userset:

If we run `ListObjects(user:justin, type:repo, relation: #admin)` on the following model:

```
type user
type organization
	relations
		define member: [user]
		define repo_admin: [organization#member]
type repo
	relations
		define admin: repo_admin from owner
		define owner: [organization]

// Tuples
"repo:fga#owner@organization:jz",
"organization:jz#repo_admin@organization:j#member",
"organization:j#member@user:justin",
```

Relation repo#admin is a TTU pointing at a userset. Let's traverse this one and illustrate the stack:

`repo#admin` -> `organization#repo_admin` -> `[organization#member]` -> `organization#member` -> `[user]`

When we reach that `[user]` leaf node, the stack looks like this:
`[{typeRel: repo#owner},  {typeRel: organization#repo_admin, usersetRelation:member}, {typeRel: organization#member}]`

Query for tuples now does the following:
1. Pop the last relation off the stack and query for `organization#member@user:justin`
2. Based on the tuples above, we find `organization:j`.
	* the stack now looks like `[{typeRel: repo#owner},  {typeRel: organization#repo_admin, usersetRelation:member}]`
3. Pop the last relation off the stack and use `organization:j` as a @user in the next query. In this case, there is also a defined `usersetRelation` in the stack entry, so the query looks like `organization#repo_admin@organization:j#member`.
4. Based on the tuples above, we find `organization:jz`
	* the stack now looks like [{typeRel: repo#owner}]
5. Pop the last relation off the stack, and query for `repo#owner@organization:jz`.
6. Based on the tuples above, we find `repo:fga`
7. The stack is now empty, so `repo:fga` will be returned to ListObjects.

## Review Checklist


pkg/server/commands/list_objects.go
```diff
@@ -68,6 +68,8 @@ type ListObjectsQuery struct {
 	checkResolver            graph.CheckResolver
 	cacheSettings            serverconfig.CacheSettings
 	sharedDatastoreResources *shared.SharedDatastoreResources
+
+	listObjectOptimizationsEnabled bool `default:"false"`
```
adriantam: You don't need to do this.  By default bool member is false.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -311,6 +342,10 @@ func (c *ReverseExpandQuery) execute(
 		sourceUserObj = userset.ObjectRelation.GetObject()
 		sourceUserRef = typesystem.DirectRelationReference(sourceUserType, userset.ObjectRelation.GetRelation())
 
+		// Queries that come in explicitly looking for userset relations will skip weighted graph for now.
+		// e.g. ListObjects(document, viewer, team:fga#member)
+		req.skipWeightedGraph = true
```
adriantam, line 346: Do you mean to default this to false?  I don't see skipWeightedGraph set to false anywhere - render this logic rather useless...

justincoh:
@willvedd 
> Are you not supporting infinite weights or are you not supporting tuple cycles? Because I see recursive queriy test cases which are surely of infinite weight. 

It's not supporting either. There are some recursive test cases added in this PR, but those are currently falling back to existing reverse expand based on their weight. They were added during work on recursion/cycles, which I removed, but I didn't remove the test cases—I think there's still value in seeing that they still work, even if it's just to validate the fallback behavior.

> Where are how are you differentiating between tuple cycles and recursive queries?

I am not, at least not as part of this PR. I suspect the end solution for both will be similar though, since a recursive relation is basically just a cycle of length 1.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -124,7 +148,8 @@ type ReverseExpandQuery struct {
 	candidateObjectsMap *sync.Map
 
 	// localCheckResolver allows reverse expand to call check locally
-	localCheckResolver *graph.LocalChecker
+	localCheckResolver             *graph.LocalChecker
+	listObjectOptimizationsEnabled bool
```
elbuo8:
```suggestion
	optimizationsEnabled bool
```
?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -329,6 +364,53 @@ func (c *ReverseExpandQuery) execute(
...
+	// This if condition can only occur on the initial entry to execute.
+	// After that first call, one of these will be changed from its default value.
+	if !req.skipWeightedGraph && req.weightedEdgeTypeRel == "" {
+		typeRel := tuple.ToObjectRelationString(targetObjRef.GetType(), targetObjRef.GetRelation())
+		node, ok := c.typesystem.GetNode(typeRel)
+		if !ok {
+			// The weighted graph is not guaranteed to be present.
+			// If there's no weighted graph, which can happen for models with tuple cycles, we will log an error below
+			// and then fall back to the non-weighted version of reverse_expand
+			c.logger.Info("unable to find node in weighted graph", zap.String("nodeID", typeRel), zap.String("storeID", req.StoreID))
```
elbuo8:
```suggestion
			c.logger.InfoWithContext("unable to find node in weighted graph", zap.String("nodeID", typeRel), zap.String("storeID", req.StoreID))
```

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -329,6 +364,53 @@ func (c *ReverseExpandQuery) execute(
...
+				req.skipWeightedGraph = true
+			}
+		}
+	}
+
+	if c.listObjectOptimizationsEnabled && !req.skipWeightedGraph {
+		targetTypeRel := req.weightedEdgeTypeRel
+
+		if targetTypeRel == "" { // This is true on the first call of reverse expand
+			targetTypeRel = tuple.ToObjectRelationString(targetObjRef.GetType(), targetObjRef.GetRelation())
```
elbuo8: You had calculated this already.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+// relationStack represents the path of queryable relationships encountered on the way to a terminal type.
+// As reverseExpand traverses from a requested type#rel to its leaf nodes, it pushes to this stack.
+// Each entry is a `typeRelEntry` struct, which contains the `type#relation`
+// along with metadata:
+//   - `usersetRelation`: To handle transitions through usersets (e.g. `[team#member]`).
+//
+// After reaching a leaf, this stack is consumed by the `queryForTuples` function to build the precise chain of
+// database queries needed to find the resulting objects.
+// To avoid races, every leaf node receives its own copy of the stack.
+type relationStack []typeRelEntry
```
elbuo8: Is this just for convenience? https://github.com/emirpasic/gods?tab=readme-ov-file#arraystack

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+type queryJob struct {
+	foundObject string
+	req         *ReverseExpandRequest
+}
+
+// jobQueue is a thread-safe queue for managing `queryJob` instances.
+// It's used to hold jobs that need to be processed during the recursive
+// `queryForTuples` operation, allowing concurrent processing of branches
+// in the authorization graph.
+type jobQueue struct {
```
elbuo8, line 64: https://github.com/emirpasic/gods?tab=readme-ov-file#arrayqueue

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+// The behavior is determined by the edge type:
+//
+//   - DirectEdge: This represents a direct path to data. Here we initiate a call to
+//     `queryForTuples` to query the datastore for tuples that match the relationship path
+//     accumulated in the stack. This is the end of the traversal.
+//
+//   - ComputedEdge, RewriteEdge, and TTUEdge: These represent indirections in the authorization model.
+//     The function modifies the traversal 'stack' to reflect the next relationship that needs to be resolved.
+//     It then calls `dispatch` to continue traversing the graph with this new state until it reaches a DirectEdge.
+func (c *ReverseExpandQuery) loopOverWeightedEdges(
```
elbuo8:
```suggestion
func (c *ReverseExpandQuery) loopOverEdges(
```

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+
+func (r *relationStack) Push(value typeRelEntry) {
+	*r = append(*r, value)
+}
+
+func (r *relationStack) Pop() typeRelEntry {
+	element := (*r)[len(*r)-1]
+	*r = (*r)[0 : len(*r)-1]
+	return element
+}
```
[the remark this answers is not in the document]
adriantam: Agree with coderabbit, should handle len(0) case

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+				// so if we find team:fga for this user, we need to know to check for
+				// team:fga#member when we check org#teammate
+				newReq.stack[len(newReq.stack)-1].usersetRelation = tuple.GetRelation(toNode.GetUniqueLabel())
+
+				newReq.stack.Push(typeRelEntry{typeRel: toNode.GetUniqueLabel()})
+
+				// Now continue traversing
+				err := c.dispatch(ctx, newReq, resultChan, needsCheck, resolutionMetadata)
+				if err != nil {
+					errs = errors.Join(errs, err)
```
elbuo8: Plan is to swallow them?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+			tupleKey, err := filteredIter.Next(ctx)
+			if err != nil {
+				if errors.Is(err, storage.ErrIteratorDone) {
+					break
+				}
+				errs = errors.Join(errs, err)
+				break LoopOnIterator
+			}
+
+			condEvalResult, err := eval.EvaluateTupleCondition(ctx, tupleKey, c.typesystem, req.Context)
```
elbuo8: This should be built into your iterator.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+					needsCheck,
+					resultChan,
+				)
+			})
+		case weightedGraph.ComputedEdge:
+			// A computed edge is an alias (e.g., `define viewer: editor`).
+			// We replace the current relation on the stack (`viewer`) with the computed one (`editor`),
+			// as tuples are only written against `editor`.
+			if toNode.GetNodeType() != weightedGraph.OperatorNode {
+				_ = newReq.stack.Pop()
```
adriantam: Instead of pop + push, will it make sense to just have a replaceTop function? In this way, we don't need to deref item + realloc un-necessarily.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+				continue
+			}
+
+			// This will be a "type:id" e.g. "document:roadmap"
+			foundObject := tupleKey.GetObject()
+
+			// If there are no more type#rel to look for in the stack that means we have hit the base case
+			// and this object is a candidate for return to the user.
+			if len(currentReq.stack) == 0 {
+				_ = c.trySendCandidate(ctx, needsCheck, foundObject, resultChan)
```
elbuo8: I would start nomalizing into using `TrySend` and a full struct so its generic, not bounded to the method.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+	if err != nil {
+		telemetry.TraceError(span, err)
+		return err
+	}
+
+	// Populate the jobQueue with the initial jobs
+	queryJobQueue.enqueue(items...)
+
+	pool := concurrency.NewPool(ctx, int(c.resolveNodeBreadthLimit))
+	activeJobs := atomic.Int64{}
```
elbuo8: What is this for?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+		// This means this query has been run on this branch of the graph already, don't do it again.
+		return false
+	}
+
+	return true
+}
+
+// buildFilteredIterator constructs the iterator used when reverse_expand queries for tuples.
+// The returned iterator MUST have .Stop() called on it.
+func (c *ReverseExpandQuery) buildFilteredIterator(
```
elbuo8, line 492: I believe there is a util for this already. See if it can be generalized further.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -41,7 +43,29 @@ type ReverseExpandRequest struct {
 	Context          *structpb.Struct
 	Consistency      openfgav1.ConsistencyPreference
 
-	edge *graph.RelationshipEdge
+	edge              *graph.RelationshipEdge
+	skipWeightedGraph bool
+
+	weightedEdge        *weightedGraph.WeightedAuthorizationModelEdge
+	weightedEdgeTypeRel string
+	stack               relationStack
```
adriantam: To avoid constantly alloc, is it possible to pre-alloc stack in advance?  Perhaps the depth of the authorization model?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+	resultChan chan<- *ReverseExpandResult,
+) error {
+	span := trace.SpanFromContext(ctx)
+
+	// This map is used for memoization of database queries for this branch of the reverse expansion.
+	// It prevents re-running the exact same database query for a given object type, relation, and user filter.
+	jobDedupeMap := new(sync.Map)
+	queryJobQueue := newJobQueue()
+
+	var queryFunc = func(ctx context.Context, job queryJob) ([]queryJob, error) {
```
adriantam: comments (not a blocker): not a big fan of defining a func within a func - especially one that has a named loop inside.  Anyway we move this to be a separate independent function?

pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go
```diff
@@ -0,0 +1,339 @@
+package reverseexpand
```
elbuo8, line 1: Are these tests the same as the existing suite? If not, then we should have this method run the same tests too to have 1-1 comparisions.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+	pool := concurrency.NewPool(ctx, int(c.resolveNodeBreadthLimit))
+	activeJobs := atomic.Int64{}
+
+	initial := true // Needed to enter the first iteration of the for loop below
+
+	// This loop processes jobs from the queue concurrently.
+	// It continues as long as there are items in the queue OR there are active goroutines processing jobs.
+	// The `initial` flag ensures the loop runs at least once to kick off the first jobs.
+	for initial || activeJobs.Load() > 0 {
+		initial = false // set to false and rely on our activeJobs count
```
adriantam: Why do we need set this to false?  I don't see initial being used again

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+
+	// This is true on every call to queryFunc except the first, since we only trigger subsequent
+	// calls if we successfully found an object.
+	if object != "" {
+		entry := req.stack.Peek()
+		filter := &openfgav1.ObjectRelation{Object: object}
+		if entry.usersetRelation != "" {
+			filter.Relation = entry.usersetRelation
+		}
+		userFilter = append(userFilter, filter)
```
adriantam: instead of append, why not just return []*openfgav1.ObjectRelation{filter}

This will simplify the logic and render else part un-necessary

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+	}
+
+	return nil
+}
+
+func buildUserFilter(
+	req *ReverseExpandRequest,
+	object string,
+) []*openfgav1.ObjectRelation {
+	var userFilter []*openfgav1.ObjectRelation
```
adriantam: don't need to append...just return slice, i.e., 
return []*openfgav1.ObjectRelation{&openfgav1.ObjectRelation{Object: tuple.BuildObject(toNode.GetUniqueLabel(), userID)}}

this simplifies logic + do not require additional alloc of slice

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+		return current.String() + accumulator
+	})
+
+	key += relation + objectType
+	if _, loaded := dedupeMap.LoadOrStore(key, struct{}{}); loaded {
+		// This means this query has been run on this branch of the graph already, don't do it again.
+		return false
+	}
+
+	return true
```
adriantam:
```suggestion
         _, loaded := dedupeMap.LoadOrStore(key, struct{}{})
	return !loaded
```
To simplify logic

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,519 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// filter out invalid tuples yielded by the database iterator
+	filteredIter := storage.NewFilteredTupleKeyIterator(
+		storage.NewTupleKeyIteratorFromTupleIterator(iter),
+		validation.FilterInvalidTuples(c.typesystem),
+	)
+	return filteredIter, nil
```
adriantam:
```suggestion
	return storage.NewFilteredTupleKeyIterator(
		storage.NewTupleKeyIteratorFromTupleIterator(iter),
		validation.FilterInvalidTuples(c.typesystem),
	), nil
```

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: I was unsure about this approach because I agree with you. I will look into splitting it out, but iirc I did attempt to split it and the resulting function required so many parameters to work that I ended up leaving it this way. I'll give it a quick look again though

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: Hard to know what size to pre-alloc since models can vary so widely. A number of real-world models will encounter leaf nodes at `len(stack)==1`, and some may get quite deep. I think it does make sense to at least pre-alloc length 1 though, since _any_ request that comes through this path will have a stack of _at least_ length 1; we push the root relation right at the beginning.

adriantam:
> **Q:** Are you not supporting infinite weights or are you not supporting tuple cycles? Because I see recursive queriy test cases which are surely of infinite weight. Where are how are you differentiating between tuple cycles and recursive queries?

This will be addressed in a separate PR.

pkg/server/commands/reverseexpand/reverse_expand.go
senojj: `stack` would be more efficient as a linked list.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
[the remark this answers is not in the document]
justincoh: Dropped custom stack implementation for [gods/arraystack](https://github.com/emirpasic/gods#arraystack)

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: It will be defaulted to false as the zero value of the variable, so it's always false unless we change it

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: moved over to using [gods/arraystack](https://github.com/emirpasic/gods#arraystack) rather than a custom stack, per some other feedback on this PR

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: TIL `NewConditionsFilteredTupleKeyIterator`

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: Looking at this again I think the `errors.Join` calls in here are unnecessary. `errs` is defined on every call to loopOverEdges, and this is always returning immediately after calling `errors.Join` so it'll only ever be a single error anyway. Will change these to just `return err`

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: recapping offline convo: `trySendCandidate` has some baked in dedupe logic specific to ListObjects, but we can swap our util `TrySendThroughChannel` method into its internals, so that's what I'm doing.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: Update: it wasn't bad. Updating some explainer comments and pushing momentarily.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: This is to ensure the `for` loop a few lines below doesn't terminate early. As long as there are jobs still being processed, that `for` has to continue looping in case more items are added to the queue. I considered doing this with a `select` over channels, but this is simpler.

adriantam:
> This PR adds a separate ReverseExpand path utilizing the weighted graph. Note: resolution paths with possible tuple cycles (weight = ∞) are currently excluded from this new path. That work is underway as a follow up.

Do we also want to call out that we exclude the case where userset is excluded?

pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go
justincoh: These are net-new, I used them for a lot of the dev here. They're currently running specifically against the new version (or falling back to the old one for weights=∞). I could have it run 2x and compare results? Once with optimizations flag and once without

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -41,7 +44,29 @@ type ReverseExpandRequest struct {
 	Context          *structpb.Struct
 	Consistency      openfgav1.ConsistencyPreference
 
-	edge *graph.RelationshipEdge
+	edge              *graph.RelationshipEdge
```
elbuo8, line 47: Eventually, we would effectively deprecate this field in favor of the one regarding the weighted graph right?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -41,7 +44,29 @@ type ReverseExpandRequest struct {
...
+		StoreID:             r.StoreID,
+		ObjectType:          r.ObjectType,
+		Relation:            r.Relation,
+		User:                r.User,
+		ContextualTuples:    r.ContextualTuples,
+		Context:             r.Context,
+		Consistency:         r.Consistency,
+		edge:                r.edge,
+		weightedEdge:        r.weightedEdge,
+		weightedEdgeTypeRel: r.weightedEdgeTypeRel,
```
elbuo8: Are we storing this string to identify that we have yet to loop over the edge? Could we maybe use a boolean instead since it's kinda duplicative to store the string without much context.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -321,14 +357,56 @@ func (c *ReverseExpandQuery) execute(
...
+			} else {
+				weight, _ := node.GetWeight(sourceUserType)
+				if weight == weightedGraph.Infinite {
+					c.logger.InfoWithContext(ctx, "reverse_expand graph may contain cycle, skipping weighted graph", zap.String("storeID", req.StoreID))
+					req.skipWeightedGraph = true
+				}
+			}
+		}
+
+		if !req.skipWeightedGraph {
```
elbuo8, line 386: How can you get to this `if` if you can't actually pass line 366?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -321,14 +357,56 @@ func (c *ReverseExpandQuery) execute(
...
+			if !ok {
+				// The weighted graph is not guaranteed to be present.
+				// If there's no weighted graph, which can happen for models with tuple cycles, we will log an error below
+				// and then fall back to the non-weighted version of reverse_expand
+				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph", zap.String("nodeID", typeRel), zap.String("storeID", req.StoreID))
+				req.skipWeightedGraph = true
+			} else {
+				weight, _ := node.GetWeight(sourceUserType)
+				if weight == weightedGraph.Infinite {
+					c.logger.InfoWithContext(ctx, "reverse_expand graph may contain cycle, skipping weighted graph", zap.String("storeID", req.StoreID))
```
elbuo8: We don't use `storeID` as the format, but it's already done by using `InfoWithContext`. 

It would be more valuable to include which part of the relation is causing this.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -619,18 +697,14 @@ func (c *ReverseExpandQuery) trySendCandidate(ctx context.Context, intersectionO
...
-		case candidateChan <- &ReverseExpandResult{
-			Object:       candidateObject,
-			ResultStatus: resultStatus,
-		}:
+		result := &ReverseExpandResult{Object: candidateObject, ResultStatus: resultStatus}
+		ok = concurrency.TrySendThroughChannel(ctx, result, candidateChan)
+		if ok {
 			span.SetAttributes(attribute.Bool("sent", true))
+		} else {
+			c.logger.ErrorWithContext(ctx, "failed to send candidate object", zap.String("object", candidateObject))
```
elbuo8, line 705: There was no log, and honestly I wouldn't add one. If the context is cancelled, the message shouldn't be sent 🤷🏽‍♂️

elbuo8: Can't say I see any value to the `trySendCandidate` span at all but... 🤷🏽 can wait till a different PR.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,496 @@
...
+// To avoid races, every leaf node receives its own copy of the stack.
+type typeRelEntry struct {
+	typeRel string // e.g. "organization#admin"
+
+	// Only present for userset relations. Will be the userset relation string itself.
+	// For `rel admin: [team#member]`, usersetRelation is "member"
+	usersetRelation string
+}
+
+// This creates a shallow clone of the relation stack.
```
elbuo8: Shallow? This seems like a deep one.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,496 @@
...
+//     accumulated in the stack. This is the end of the traversal.
+//
+//   - ComputedEdge, RewriteEdge, and TTUEdge: These represent indirections in the authorization model.
+//     The function modifies the traversal 'stack' to reflect the next relationship that needs to be resolved.
+//     It then calls `dispatch` to continue traversing the graph with this new state until it reaches a DirectEdge.
+func (c *ReverseExpandQuery) loopOverEdges(
+	ctx context.Context,
+	edges []*weightedGraph.WeightedAuthorizationModelEdge,
+	needsCheck bool,
+	req *ReverseExpandRequest,
```
elbuo8: Should be the second parameter to match most other functions.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,496 @@
...
+				//		define member: [user]
+				//	type org:
+				//		define teammate: [team#member]
+				// A direct edge here is org#teammate --> team#member
+				// so if we find team:fga for this user, we need to know to check for
+				// team:fga#member when we check org#teammate
+				el, ok := newReq.relationStack.Pop()
+				if !ok {
+					// shouldn't be possible
+					return fmt.Errorf("unexpected empty stack while processing direct edge to userset")
```
elbuo8: Can be a const error.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,496 @@
...
+					// shouldn't be possible
+					return fmt.Errorf("unexpected type in stack: expected typeRelEntry, got %T", el)
+				}
+				entry.usersetRelation = tuple.GetRelation(toNode.GetUniqueLabel())
+				newReq.relationStack.Push(entry)
+
+				newReq.relationStack.Push(typeRelEntry{typeRel: toNode.GetUniqueLabel()})
+
+				// Now continue traversing
+				err := c.dispatch(ctx, newReq, resultChan, needsCheck, resolutionMetadata)
```
elbuo8: We want to block while this happens?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,489 @@
...
+			// as tuples are only written against `editor`.
+			if toNode.GetNodeType() != weightedGraph.OperatorNode {
+				newReq.relationStack.Pop()
+				newReq.relationStack.Push(typeRelEntry{typeRel: toNode.GetUniqueLabel()})
+			}
+
+			err := c.dispatch(ctx, newReq, resultChan, needsCheck, resolutionMetadata)
+			if err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
elbuo8: You do check elsewhere, might as well be consistent.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
elbuo8: Hm. This could become an N*M problem. Giving `resolveNodeBreadthLimit` to me seems like a mistake. I would use a finite number (say 2). And each "parent" resolves all of the children it discovers and so forth. Which itself will be bounded by the datastore layer.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,496 @@
...
+		// If there are no more type#rel to look for in the stack that means we have hit the base case
+		// and this object is a candidate for return to the user.
+		if currentReq.relationStack.Empty() {
+			c.trySendCandidate(ctx, needsCheck, foundObject, resultChan)
+			continue
+		}
+
+		// For non-recursive relations (majority of cases), if there are more items on the stack, we continue
+		// the evaluation one level higher up the tree with the `foundObject`.
+		nextJobs = append(nextJobs, queryJob{foundObject: foundObject, req: currentReq})
```
elbuo8, line 413: We treat this as a stream rather than collecting/reducing. Perhaps a subsequent PR it can tackle that.

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: Yeah I mostly added this because the previous trySendCandidate was returning an `err` if it encountered one, but with our `TrySendThroughChannel` method we lose the actual error.

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: this was a leftover from earlier dev work, we can get this information off the edge. removed it

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: We could kick it off in a pool, but it's not very expensive. Right now the only part of `loopOverEdges` which does kick off a new goroutine is DirectEdge to a leaf—when we start querying the DB. 

There are other blocking dispatches in loopOverEdges (RewriteEdges and TTU edges), we could have them all run in `pool.Go`, but traversing edges feels like it _should be_ fast enough without separate routines.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: updated to limit of 2

pkg/server/commands/reverseexpand/reverse_expand.go
elbuo8: It was only returning the context cancelation error, which likely got completely discarded.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,509 @@
...
+	entry, ok := val.(typeRelEntry)
+	if !ok {
+		// should never happen, we're only pushing typeRelEntry to this stack
+		return nil, fmt.Errorf("unexpected type encountered in query stack %T", val)
+	}
+	typeRel := entry.typeRel
+
+	// Ensure that we haven't already run this query
+	ok = checkQueryIsUnique(jobDedupeMap, userFilter, typeRel)
+	if !ok {
```
adriantam:
```suggestion
	if !checkQueryIsUnique(jobDedupeMap, userFilter, typeRel) {
```
nitpik

adriantam (dismissed):

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: implemented and also renamed this to `isDuplicateQuery` be more clear

`if isDuplicateQuery(jobDedupeMap, userFilter, typeRel){ return nil, nil }`

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,511 @@
...
+		toNode := edge.GetTo()
+
+		// Going to a userset presents risk of infinite loop. Using from + to ensures
+		// we don't traverse the exact same edge more than once.
+		goingToUserset := toNode.GetNodeType() == weightedGraph.SpecificTypeAndRelation
+		if goingToUserset {
+			key := edge.GetFrom().GetUniqueLabel() + toNode.GetUniqueLabel()
+			if _, loaded := c.visitedUsersetsMap.LoadOrStore(key, struct{}{}); loaded {
+				// we've already visited this userset through this edge, exit to avoid an infinite cycle
+				continue
```
adriantam, line 146: Not a blocker but probably good to follow up on subsequent PR is to add test for this case (i am thinking of when we add in the changes to allow recursion)

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -0,0 +1,511 @@
...
+	} else {
+		// This else block ONLY hits on the first call to queryFunc.
+		toNode := req.weightedEdge.GetTo()
+
+		switch toNode.GetNodeType() {
+		case weightedGraph.SpecificType: // Direct User Reference. To() -> "user"
+			// We will always have a UserRefObject here. Queries that come in for pure usersets do not take this code path.
+			// e.g. ListObjects(team:fga#member, document, viewer) will not make it here.
+			var userID string
+			if val, ok := req.User.(*UserRefObject); ok {
```
adriantam: is it error if !ok?

justincoh: this path shouldn't ever be able to reach the state where `!ok` but yes, that'd be an error. Updated to return error in that case to be safe

adriantam (dismissed):

adriantam (approved):

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -444,15 +444,20 @@ func buildUserFilter(
...
-			if !ok {
-				return nil, fmt.Errorf("unexpected user type when building User filter: %T", val)
+			if ok {
+				userID = val.Object.GetId()
+			} else {
+				// It might be a wildcard user, which is ok
+				_, ok = req.User.(*UserRefTypedWildcard)
+				if !ok {
+					return nil, fmt.Errorf("unexpected user type when building User filter: %T", val)
+				}
```
adriantam, line 458: In the case of wildcard user, do we still want to return the filter?  Or this should return []*openfgav1.ObjectRelation{}, nil

justincoh: We should return the empty filter as you suggested, updated

PR #2530, opened 2025-06-26
perf: further reduce lock contention in shared iterators

senojj:
## Description

This PR introduces a change to how shared iterators are created that reduces the number of mutex locks required when a shared iterator already exists for a given key.

A `sync.Once` is used instead of directly using a `sync.Mutex` because the `sync.Once` will only acquire a mutex lock the first time that it executes. Subsequent calls to `sync.Once::Do` will result in a simple atomic value evaluation. 

Additionally, this PR removes two overly communicative spans.
## Review Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of shared iterators for more efficient and reliable operation. Users may notice enhanced stability and performance when accessing shared storage iterators.

### DIFF
--- a/internal/graph/builder.go
+++ b/internal/graph/builder.go
@@ -121,6 +121,10 @@ func LocalCheckResolver(resolver CheckResolver) (*LocalChecker, bool) {
 	if ok {
 		return localChecker, true
 	}
+	shadowChecker, ok := resolver.(*ShadowResolver)
+	if ok {
+		return LocalCheckResolver(shadowChecker.main)
+	}
 	delegate := resolver.GetDelegate()
 	if delegate != nil {
 		if delegate == resolver {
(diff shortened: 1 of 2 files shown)


pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -102,27 +102,28 @@ type storageItem struct {
...
-	defer s.mu.Unlock()
+	var err error
+	var created bool
 
-	if s.iter != nil {
-		return s.iter.clone(), false, nil
-	}
+	s.once.Do(func() {
+		s.iter, err = s.producer()
+		created = true
```
justincoh: if this `err != nil` is created still true?

senojj: look at lines 113 and 114

justincoh (approved):

yissellokta (approved):

justincoh (approved):

PR #2539, opened 2025-07-03
Log main and shadow ListObjects query latencies

cikasfm:
Log main and shadow ListObjects query latencies

## Description

Feature: Log main and shadow ListObjects query latencies

As an operator,
I want the system to log the duration of both the main and shadow ListObjects queries when their results match,
so that I can monitor and compare their performance.

```gherkin
  Scenario: Both queries return matching results
    Given the main ListObjects query completes successfully
    And the shadow ListObjects query completes successfully with the same results
    When the result is logged
    Then the log entry includes the main_latency and shadow_latency fields

  Scenario: Queries return different results
    Given the main and shadow ListObjects queries return different results
    When the result is logged
    Then the log entry does not include main_latency and shadow_latency fields for matches

  Scenario: Shadow query fails
    Given the main ListObjects query completes successfully
    And the shadow ListObjects query fails
    When the result is logged
    Then the log entry does not include main_latency and shadow_latency fields for matches
```

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Enhanced logging now includes both main and shadow query execution times when comparing results, providing more detailed performance insights.

* **Tests**
  * Updated tests to verify the inclusion of main query latency in logs during shadow query comparisons.

### DIFF
--- a/pkg/server/commands/list_objects_shadow.go
+++ b/pkg/server/commands/list_objects_shadow.go
@@ -166,7 +166,7 @@ func (q *shadowedListObjectsQuery) Execute(
 			}()
 			defer q.wg.Done() // only used for testing signals
 
-			q.executeShadowModeAndCompareResults(cloneCtx, req, res.Objects)
+			q.executeShadowModeAndCompareResults(cloneCtx, req, res.Objects, latency)
 		}()
 	}
 	return res, err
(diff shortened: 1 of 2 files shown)


cikasfm:
> Why not have latency even when there's a mismatch?

@justincoh when there's a mismatch we assume that there was a problem, or the results are incomparable, thus latency comparison is also not much valuable.

We could add that, but then also add some way of separating match vs no-match.

senojj:
> when there's a mismatch we assume that there was a problem, or the results are incomparable, thus latency comparison is also not much valuable.
> 
> We could add that, but then also add some way of separating match vs no-match.

If the latency is different beyond some defined threshold, this could indicate that there is a problem.

Indicating match vs no-match seems reasonable.

senojj:
Currently this implementation runs the shadow execution only after the main execution has completed.

Is there some implicit dependency that prevents us from running the shadow execution at the same time as the main execution?

cikasfm:
@senojj technically nothing other than ( as mentioned in #2509 review ) expectation that we need the `main` query to be successful, fast ( not "almost reaching deadline" ) and not maxing out the results.

Essentially we need the results of the main query to check the preconditions for execution of the shadow query, otherwise if the main one fails, is slow or maxes out the results, then comparison would not show any value.

P.S. that was the initial implementation, running in parallel, but then decision was made to make it sequential for the reasons mentioned above.

cikasfm:
> Currently this implementation runs the shadow execution only after the main execution has completed.
> 
> Is there some implicit dependency that prevents us from running the shadow execution at the same time as the main execution?

@senojj the reason we changed this from parallel to sequential w "fire & forget" is that before we run the shadow mode we want to make sure that:
- the main query is successful
- the main query does not max out return items ( might mean truncated results, thus incomparable )
- the main query is fast enough - if it's "almost times out", meaning it was stopped and the results are truncated, thus incomparable

PR #2542, opened 2025-07-03
refactor: use a linked list in ReverseExpand instead of copying slices

justincoh:
## Description

This is a performance follow up to the reverse expand [weighted graph work](https://github.com/openfga/openfga/pull/2529) which just landed on OpenFGA. Instead of creating and copying slices to use as our stack of relations, this PR uses a linked list to track the stack of relations.

## Review Checklist

## Summary by CodeRabbit

* **Refactor**
  * Replaced the external stack implementation with a custom stack for improved stack management and immutability.
  * Updated stack-related logic and error handling to use the new stack structure.

* **Tests**
  * Refactored and expanded stack operation tests to cover custom stack behavior, immutability, and error conditions.


pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
 	// Only present for userset relations. Will be the userset relation string itself.
 	// For `rel admin: [team#member]`, usersetRelation is "member"
 	usersetRelation string
 }
 
-// When cloning requests to pass into separate goroutines, we need to create a copy of that
-// request's stack as well to avoid races.
-func cloneStack(original lls.Stack) lls.Stack {
-	newStack := lls.New()
+type Stack struct {
```
adriantam: Do we intend to have this as public?  Looks like this is not particularly generic - which suggests that this is only used for reverse expand.  If this is intended to be more generic, then put this in somewhere like util. 

(Though thinking about it..nothing stops it from being generic).

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
+	value TypeRelEntry
+	next  *Stack
+}
+
+func Push(s *Stack, entry TypeRelEntry) *Stack {
+	newStack := Stack{value: entry, next: nil}
+	newStack.next = s
+	return &newStack
```
cikasfm: Why not just create the instance in place
```golang
    return &Stack{value: entry, next: s}
```

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
 }
 
-// When cloning requests to pass into separate goroutines, we need to create a copy of that
-// request's stack as well to avoid races.
-func cloneStack(original lls.Stack) lls.Stack {
-	newStack := lls.New()
+type Stack struct {
+	value TypeRelEntry
+	next  *Stack
+}
```
cikasfm: Some `godoc` would be nice explaining how this works, does not look straight forward

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
+
+func Push(s *Stack, entry TypeRelEntry) *Stack {
+	newStack := Stack{value: entry, next: nil}
+	newStack.next = s
```
cikasfm: ~Isn't this supposed to be `s.next = newStack`?~

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
-	vals := original.Values()
-	for i := len(vals) - 1; i >= 0; i-- {
-		newStack.Push(vals[i])
+func Pop(s *Stack) (TypeRelEntry, *Stack, error) {
+	if IsEmpty(s) {
+		return TypeRelEntry{}, nil, ErrEmptyStack
 	}
-	return *newStack
+	return s.value, s.next, nil
+}
```
senojj: This is an odd implementation.. why is the error needed? `nil` is perfectly valid to return for `*Stack` when the tail of the stack is empty.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
 
-	vals := original.Values()
-	for i := len(vals) - 1; i >= 0; i-- {
-		newStack.Push(vals[i])
+func Pop(s *Stack) (TypeRelEntry, *Stack, error) {
+	if IsEmpty(s) {
+		return TypeRelEntry{}, nil, ErrEmptyStack
 	}
-	return *newStack
+	return s.value, s.next, nil
```
cikasfm: I'm starting to see how this works. `s.next` is actually the previous value in the stack; i.e. we keep track of the `tail` of the stack, and once popped, the new tail is the previous node.

~Can we rename it to make it clear? i.e. `next` is actually `previous`?~

cikasfm (dismissed):

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
adriantam: Feels like these should be property of stack.  ie.,e 

func (s *Stack) Push(entry TypeRelEntry) *Stack {
...
}
and 
func (s *Stack) Pop() (TelRelEntry, error) {
}

adriantam: Similarly Peek and IsEmpty

senojj: If the value given is `nil`, a panic due to dereferencing a `nil` value is valid.

senojj: An empty stack is denoted by a `nil` value.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -22,31 +21,44 @@ import (
...
+}
+
+func Peek(s *Stack) TypeRelEntry {
+	return s.value
+}
+
+func IsEmpty(s *Stack) bool {
+	return s.next == nil
 }
```
senojj: I don't know that this function adds any value over `if value == nil`

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -158,20 +170,15 @@ func (c *ReverseExpandQuery) loopOverEdges(
...
-					// shouldn't be possible
-					return ErrEmptyStack
-				}
-				entry, ok := el.(typeRelEntry)
-				if !ok {
-					// shouldn't be possible
-					return fmt.Errorf("unexpected type in stack: expected typeRelEntry, got %T", el)
+				entry, stack, err := Pop(newReq.relationStack)
+				if err != nil {
+					return err
```
senojj: I would rather see a `nil` check before the call to `Pop`, than an error check after.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
senojj: It's definitely odd because none of the sibling functions check for the presence of a `nil` value or return an error.

justincoh: That would provide a nice way to make this generic as well

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
cikasfm: Ignore that. Seems like this is a common pattern

pkg/server/commands/reverseexpand/type_rel_stack.go
```diff
@@ -0,0 +1,28 @@
...
+// ensure thread safety, since ReverseExpand kicks off many routines all relying on their own stacks.
+type typeRelStack struct {
+	value typeRelEntry
+	next  *typeRelStack
+}
+
+func newTypeRelStack(val typeRelEntry) *typeRelStack {
+	return &typeRelStack{value: val}
+}
+
```
senojj: making methods that return a new value of the same type is a bit odd... consider its usage:
```go
var stack typeRelStack
stack = stack.push(someValue)
```
this is a bit awkward compared to using a function with a signature similar to Go's `append` for slices:
```go
var stack typeRelStack
stack = push(&stack, someValue)
```
This is strictly speaking in terms of ergonomics.

In the function approach, it more clear that a new value is being returned and the variable `stack` is receiving a new value.

pkg/server/commands/reverseexpand/type_rel_stack.go
```diff
@@ -0,0 +1,28 @@
+package reverseexpand
+
+// typeRelStack is a stack implementation on top of a linked list, specific to reverse_expand.
+//
+// *Important*: Each push() or pop() operation creates and returns a pointer to a new stack entirely to
+// ensure thread safety, since ReverseExpand kicks off many routines all relying on their own stacks.
+type typeRelStack struct {
+	value typeRelEntry
+	next  *typeRelStack
+}
```
justincoh: @adriantam @senojj @cikasfm I moved the stack implementation here. I considered making it generic, e.g. `LinkedListStack[T]` and moving it to `internal`, but this implementation is pretty non-standard. Typically I would expect methods with pointer receivers to modify their receivers, not return entirely new pointers like `push()` and `pop()` are here. We're returning new instances for good reason (thread safety) but it's very specific to ReverseExpand's use case and would be surprising in a generic context.

senojj: I don't think it is necessarily "non-standard" to use functions over methods. It really depends upon the use case. For instance, if you observe Go's own `append` function, or `copy` function, they are functions for good reason, as opposed to being methods on the slice header object itself.

justincoh: I refactored in this fashion and I think the code is now much more clear.

internal/stack/stack.go
```diff
@@ -0,0 +1,18 @@
+package stack
+
+// Stack is an implementation of a stack based on a linked list.
+//
+// *Important*: Each push() or pop() operation creates and returns a pointer to a new stack entirely to
+// ensure thread safety.
+type Stack[T any] struct {
+	Value T
```
senojj: in order to guide the user of `Stack` to make use of the functional interface, I would suggest making `Value` an unexported property and adding a `Peek` method that returns the value at the head of the `Stack`.

internal/stack/stack_test.go
```diff
@@ -0,0 +1,43 @@
...
+		secondStack := Push(firstStack, "world")
+
+		require.NotEqual(t, firstStack.Value, secondStack.Value)
+	})
+
+	t.Run("test_pop_does_not_affect_original", func(t *testing.T) {
+		firstStack := &Stack[string]{Value: "hello"}
+
+		val, secondStack := Pop(firstStack)
+		require.Equal(t, "hello", val)
```
adriantam, line 19: Would the test makes more sense if we have world as the top item and see if popping the second stack modifies the first?

adriantam (dismissed):

internal/stack/stack.go
```diff
@@ -0,0 +1,26 @@
...
+// *Important*: Each push() or pop() operation creates and returns a pointer to a new stack entirely to
+// ensure thread safety.
+type Stack[T any] struct {
+	value T
+	next  *Stack[T]
+}
+
+func New[T any](value T) *Stack[T] {
+	return &Stack[T]{value: value}
+}
```
senojj: `New` really isn't necessary because you can do:
```go
stack := Push(nil, "something")
```

internal/stack/stack_test.go
justincoh: Added an additional case along these lines

PR #2543, opened 2025-07-04
feat: add list objects intersection exclusion

Vic-Dev:
## Description

Add intersection and exclusion handling to reverse expand weighted, where check calls are made at a per operator level rather than higher up in the model.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Enhanced reverse expansion logic to support intersection and exclusion operators, improving handling of complex authorization scenarios.
  * Expanded test coverage for reverse expansion, including unions, intersections, exclusions, recursion, cycles, and error cases.

* **Bug Fixes**
  * Corrected userset construction logic for tuple-to-userset edges to align with model definitions.

* **Refactor**
  * Centralized edge and node retrieval logic in the type system for improved maintainability.
  * Updated interfaces and mock implementations to support new reverse expansion features.

* **Documentation**
  * Updated changelog to document changes to reverse expansion functionality.

* **Tests**
  * Added comprehensive tests for new stack operations, error handling, and weighted graph traversal.

### DIFF
--- a/pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
+++ b/pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
@@ -564,8 +564,9 @@ type await struct {
 func (a *await) Do(fn func()) {
 	a.mu.Lock()
 	if a.active {
+		wg := a.wg
 		a.mu.Unlock()
-		a.wg.Wait()
+		wg.Wait()
 		return
 	}
 	a.active = true


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -67,6 +67,22 @@ func (r *ReverseExpandRequest) clone() *ReverseExpandRequest {
...
+		User:              r.User,
+		ContextualTuples:  r.ContextualTuples,
+		Context:           r.Context,
+		Consistency:       r.Consistency,
+		edge:              r.edge,
+		weightedEdge:      r.weightedEdge,
+		skipWeightedGraph: r.skipWeightedGraph,
+		relationStack:     stack,
+	}
+}
```
senojj: if the `stack` is a value type, and all of the properties are copied by value, why is this function necessary?

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -245,6 +261,24 @@ func WithLogger(logger logger.Logger) ReverseExpandQueryOption {
...
+		typesystem:              c.typesystem,
+		resolveNodeLimit:        c.resolveNodeLimit,
+		resolveNodeBreadthLimit: c.resolveNodeBreadthLimit,
+		dispatchThrottlerConfig: c.dispatchThrottlerConfig,
+		candidateObjectsMap:     new(sync.Map),
+		visitedUsersetsMap:      c.visitedUsersetsMap,
+		localCheckResolver:      c.localCheckResolver,
+		optimizationsEnabled:    c.optimizationsEnabled,
+	}
+}
```
senojj: This could be a bit less verbose:
```suggestion
func (c *ReverseExpandQuery) shallowClone() *ReverseExpandQuery {
    copy := *c
    copy.candidateObjectsMap = new(sync.Map)
    return &copy
}
```

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -49,6 +56,19 @@ func cloneStack(original lls.Stack) lls.Stack {
 	return *newStack
 }
 
+// When cloning requests for intersection/exclusion subqueries.
+func createStackCloneAndStackWithTopItem(original lls.Stack) (lls.Stack, lls.Stack, error) {
+	newStack := cloneStack(original)
```
senojj: this stack data structure is really inefficient as it requires iteration to copy.

adriantam: Keep in mind of https://github.com/openfga/openfga/pull/2542
CC @justincoh

pkg/server/commands/reverseexpand/reverse_expand.go
Vic-Dev: This function is called 4 times, it's extrapolated to reduce the lines of code each time we have to do it

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -277,14 +324,24 @@ func (c *ReverseExpandQuery) queryForTuples(
...
+
+func (c *ReverseExpandQuery) queryForTuplesWithItems(
+	ctx context.Context,
+	needsCheck bool,
+	resultChan chan<- *ReverseExpandResult,
+	items []queryJob,
+	jobDedupeMap *sync.Map,
+) error {
+	span := trace.SpanFromContext(ctx)
+	queryJobQueue := newJobQueue()
```
justincoh: How come this extra method? It looks like it's doing the same thing as the existing `queryForTuples` method

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -277,14 +324,24 @@ func (c *ReverseExpandQuery) queryForTuples(
 	// This map is used for memoization of database queries for this branch of the reverse expansion.
 	// It prevents re-running the exact same database query for a given object type, relation, and user filter.
 	jobDedupeMap := new(sync.Map)
-	queryJobQueue := newJobQueue()
 
-	// Now kick off the chain of queries
```
justincoh, line 264: deleted by mistake?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -518,3 +575,249 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+	sourceUserType string,
+	resolutionMetadata *ResolutionMetadata,
+) error {
+	tmpResultChan := make(chan *ReverseExpandResult, listObjectsResultChannelLength)
+
+	intersectionEdgeComparison, err := typesystem.GetEdgesForIntersection(edges, sourceUserType)
+	if err != nil {
+		c.logger.Error("Failed to get edges from weighted graph",
+			zap.Error(err),
+			zap.Any("edges", &edges),
```
justincoh: I don't think you want this `&` as part of this logger, it'll give you the address of the slice aka something like `0xc00001c088`

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
Vic-Dev: added back in

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
Vic-Dev: `queryForTuples` is calling this method with the items it got from calling `executeQueryJob`. We needed a way to call `queryForTuples` with different items

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -518,3 +554,254 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+			zap.Error(err),
+			zap.Any("edges", edges),
+			zap.String("sourceUserType", sourceUserType))
+		return err
+	}
+
+	if !intersectionEdgeComparison.DirectEdgesAreLeastWeight && intersectionEdgeComparison.LowestEdge == nil {
+		// no need to go further because list objects must return empty
+		return nil
+	}
```
justincoh, line 662: This scenario means that there is no relevant edge to check, right? e.g. this path is a dead end

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -518,3 +554,254 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+
+	var lowestWeightEdges []*weightedGraph.WeightedAuthorizationModelEdge
+	if intersectionEdgeComparison.DirectEdgesAreLeastWeight {
+		lowestWeightEdges = intersectionEdgeComparison.DirectEdges
+	} else {
+		lowestWeightEdges = []*weightedGraph.WeightedAuthorizationModelEdge{intersectionEdgeComparison.LowestEdge}
+	}
+
+	topItemStack := lls.New()
+	topItem, ok := req.relationStack.Pop()
```
justincoh: Could you add an explainer comment about the logic behind the `topItemStack`?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -518,3 +554,254 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+				return err
+			}
+
+			if tmpCheckResult.GetAllowed() {
+				// Check returns true. Hence, we have candidates
+				if req.relationStack.Size() == 0 {
+					// If the original stack only had 1 value, we can trySendCandidate right away (nothing more to check)
+					c.trySendCandidate(ctx, false, tmpResult.Object, resultChan)
+				} else {
+					// TODO: should we be doing these items all at once or is it okay to do it one at a time?
```
justincoh: just marking this TODO since we spoke about it yesterday. One at a time is fine, just process them as they come

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -518,3 +554,254 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+	baseEdges, excludedEdge, err := typesystem.GetEdgesForExclusion(edges, sourceUserType)
+	if err != nil {
+		c.logger.Error("Failed to get lowest weight edge",
+			zap.String("function", "exclusionHandler"),
+			zap.Error(err),
+			zap.Any("edges", edges),
+			zap.String("sourceUserType", sourceUserType))
+		return err
+	}
+	if excludedEdge == nil {
```
justincoh, line 732:
```suggestion
	// This means the exclusion edge does not have a path to the terminal type.
	// e.g. `B` in `A but not B` is not relevant to this query.
	if excludedEdge == nil {
```

pkg/typesystem/typesystem.go
```diff
@@ -1700,6 +1700,35 @@ func (t *TypeSystem) IsTuplesetRelation(objectType, relation string) (bool, erro
 	return false, nil
 }
 
+// helper function to return all edges from weighted graph.
+func (t *TypeSystem) GetEdgesFromWeightedGraph(
```
justincoh: Might be more accurately named `GetEdgesFromNodeToType`?

pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go
```diff
@@ -46,16 +53,17 @@ func TestReverseExpandWithWeightedGraph(t *testing.T) {
...
 			},
-			objectType:      "repo",
-			relation:        "or_admin",
-			user:            &UserRefObject{Object: &openfgav1.Object{Type: "user", Id: "justin"}},
-			expectedObjects: []string{"repo:fga"},
+			objectType:                 "repo",
+			relation:                   "or_admin",
+			user:                       &UserRefObject{Object: &openfgav1.Object{Type: "user", Id: "justin"}},
+			expectedOptimizedObjects:   []string{"repo:fga"},
+			expectedUnoptimizedObjects: []string{"repo:fga"},
```
justincoh, line 60: I don't think you need both of these expectations. On `main` right now this test file is running ReverseExpand 2x, once with optimizations and once without, and then comparing the results to make sure they're the same. There shouldn't ever be a case where the results differ based on the experimental flag

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
Vic-Dev: yes

pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go
justincoh: recapping offline convo: we *do* need different expectations here. Existing reverse_expand knowingly returns potentially-incorrect results to ListObjects, which will then run `Check` to rule them out. The new version in this PR runs `Check` _before_ sending the results out to ListObjects, so we do need separate expectations in some cases.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -491,3 +527,232 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+	// If the edge is a tuple to userset edge, we need to later check the candidates against the
+	// original relationStack with the top item removed.
+	var topItemStack stack.Stack[typeRelEntry]
+	if req.relationStack != nil {
+		topItem, newStack := stack.Pop(req.relationStack)
+		req.relationStack = newStack
+		topItemStack = stack.Push(nil, topItem)
+	}
+
+	pool := concurrency.NewPool(ctx, 2)
```
senojj: might make more sense for the pool to be passed into this function, instead of being created and returned by it.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -228,10 +237,36 @@ func (c *ReverseExpandQuery) loopOverEdges(
...
 				newStack = stack.Push(newStack, typeRelEntry{typeRel: toNode.GetUniqueLabel()})
 				newReq.relationStack = newStack
+
+				pool.Go(func(ctx context.Context) error {
+					return c.dispatch(ctx, newReq, resultChan, needsCheck, resolutionMetadata)
+				})
+			} else {
+				switch toNode.GetLabel() {
+				case weightedGraph.IntersectionOperator:
+					intersectionEdges, _, err := c.typesystem.GetEdgesFromNodeToType(toNode.GetUniqueLabel(), sourceUserType)
```
senojj: Here we are passing the unique label of a node into `GetEdgesFromNodeToType`, which then has to use that label to get the node again... Seems like the `GetEdgesFromNodeToType` should just accept a node instead.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -491,3 +527,232 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+	var lowestWeightEdges []*weightedGraph.WeightedAuthorizationModelEdge
+	if intersectionEdgeComparison.DirectEdgesAreLeastWeight {
+		lowestWeightEdges = intersectionEdgeComparison.DirectEdges
+	} else {
+		lowestWeightEdges = []*weightedGraph.WeightedAuthorizationModelEdge{intersectionEdgeComparison.LowestEdge}
+	}
+
+	tmpResultChan := make(chan *ReverseExpandResult, listObjectsResultChannelLength)
+	pool := c.findCandidatesForLowestWeightEdge(ctx, req, tmpResultChan, lowestWeightEdges, sourceUserType, resolutionMetadata)
+
```
justincoh: Similar to Josh's comment above: wdyt about moving this down next to the other goroutine `c.callCheckForCandidates`, and having the pool defined and passed in down there with a brief explainer? I think it would help readability here

e.g. 
```
// Concurrently find candidates and call check on them as they are found
pool := concurency.NewPool(ctx, 2)
c.findCandidatesForLowestWeightEdge(ctx, pool, otherStuff...)
c.callCheckForCandidates(ctx, pool, otherStuff...)
```

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -52,19 +52,13 @@ type ReverseExpandRequest struct {
...
-		skipWeightedGraph: r.skipWeightedGraph,
-		relationStack:     r.relationStack,
-	}
+	return r.cloneWithStack(r.relationStack)
+}
+
+func (r *ReverseExpandRequest) cloneWithStack(stack stack.Stack[typeRelEntry]) *ReverseExpandRequest {
+	copy := *r
+	copy.relationStack = stack
+	return &copy
```
justincoh: It looks like `cloneWithStack` is only ever used inside of `clone()`, where it is _always_ called. Can just eliminate `cloneWithStack` entirely and move all of its functionality inside `clone`.

Also a nit: might want to rename the var `copy` here, since that collides with the [builtin copy function](https://pkg.go.dev/builtin#copy)

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -228,10 +237,36 @@ func (c *ReverseExpandQuery) loopOverEdges(
...
+						return err
+					}
+					err = c.exclusionHandler(ctx, newReq, resultChan, exclusionEdges, sourceUserType, resolutionMetadata)
+					if err != nil {
+						return err
+					}
+				default:
+					pool.Go(func(ctx context.Context) error {
+						return c.dispatch(ctx, newReq, resultChan, needsCheck, resolutionMetadata)
+					})
```
justincoh: Should this be an error instead? I _think_ the only `OperatorNode` types are `intersection` and `exclusion`, if something else is here it is unexpected

Vic-Dev: It can be a `union`

justincoh: indeed it can, never mind this thread

Vic-Dev: updated to be more explicit about the union case and error on default

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -52,19 +52,8 @@ type ReverseExpandRequest struct {
...
-		User:              r.User,
-		ContextualTuples:  r.ContextualTuples,
-		Context:           r.Context,
-		Consistency:       r.Consistency,
-		edge:              r.edge,
-		weightedEdge:      r.weightedEdge,
-		skipWeightedGraph: r.skipWeightedGraph,
-		relationStack:     r.relationStack,
-	}
+	copyRequest := *r
```
sergiught, line 58: What happens if `r` is `nil`? Any chance we can guard against this? 

```go
if r == nil {
	return nil
}

copy := *r
return &copy

```

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -245,6 +234,15 @@ func WithLogger(logger logger.Logger) ReverseExpandQueryOption {
 	}
 }
 
+// shallowClone creates an identical copy of reverseExpandQuery except
+// candidateObjectsMap as list object candidates need to be validated
+// via check.
+func (c *ReverseExpandQuery) shallowClone() *ReverseExpandQuery {
+	copy := *c
```
sergiught, line 247: Same as above, what happens if c is nil?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -228,10 +237,38 @@ func (c *ReverseExpandQuery) loopOverEdges(
 				_, newStack := stack.Pop(newReq.relationStack)
 				newStack = stack.Push(newStack, typeRelEntry{typeRel: toNode.GetUniqueLabel()})
 				newReq.relationStack = newStack
+
+				pool.Go(func(ctx context.Context) error {
+					return c.dispatch(ctx, newReq, resultChan, needsCheck, resolutionMetadata)
+				})
+			} else {
```
sergiught: Instead of using an else here, is there an opportunity to break within the if above so we can exist the switch and reduce complexity?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -491,3 +529,232 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+			}
+
+			// Check if the check result matches the expect isAllowed value
+			// eg, for intersection we expect the check result to be true
+			// and for exclusion we expect the check result to be false.
+			if tmpCheckResult.GetAllowed() == isAllowed {
+				if stack.Len(req.relationStack) == 0 {
+					// If the original stack only had 1 value, we can trySendCandidate right away (nothing more to check)
+					c.trySendCandidate(ctx, false, tmpResult.Object, resultChan)
+				} else {
```
sergiught: Is there an opportunity to remove this else here as well and simplify the complexity by continuing within the if above?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -491,3 +529,232 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+					zap.String("object", tmpResult.Object),
+					zap.String("relation", req.Relation),
+					zap.String("user", req.User.String()))
+				return err
+			}
+
+			// Check if the check result matches the expect isAllowed value
+			// eg, for intersection we expect the check result to be true
+			// and for exclusion we expect the check result to be false.
+			if tmpCheckResult.GetAllowed() == isAllowed {
```
sergiught: Can we invert this and continue the loop?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -491,3 +529,232 @@ func (c *ReverseExpandQuery) buildFilteredIterator(
...
+	if !intersectionEdgeComparison.DirectEdgesAreLeastWeight && intersectionEdgeComparison.LowestEdge == nil {
+		// no need to go further because list objects must return empty
+		return nil
+	}
+
+	var lowestWeightEdges []*weightedGraph.WeightedAuthorizationModelEdge
+	if intersectionEdgeComparison.DirectEdgesAreLeastWeight {
+		lowestWeightEdges = intersectionEdgeComparison.DirectEdges
+	} else {
+		lowestWeightEdges = []*weightedGraph.WeightedAuthorizationModelEdge{intersectionEdgeComparison.LowestEdge}
```
sergiught: You could set this as default on line 659 and change it on 661 effectively removing the else.

pkg/typesystem/typesystem.go
```diff
@@ -1700,6 +1700,30 @@ func (t *TypeSystem) IsTuplesetRelation(objectType, relation string) (bool, erro
 	return false, nil
 }
 
+// helper function to return all edges from weighted graph.
+func (t *TypeSystem) GetEdgesFromNodeToType(
```
justincoh, line 1704: this method name implies that we're _only_ returning edges that have a potential path to the sourceType, which I don't think is true.

```
type doc
	relation viewer: [user, employee, other]
```
If I call `GetEdgesFromNodeToType(doc#viewer, user)`, this will return all 3 edges rather than just 1.

I see we're pruning edges later in the intersectionHandler and exclusionHandler, but would it be better to prune here? Or does the `but not` edge logic make that more difficult?

justincoh (approved):
Consider my comment non-blocking. If we want to address it we can always do it in a quick follow up PR so we can get this one into `main`.

pkg/typesystem/typesystem.go
Vic-Dev: To be addressed in a future PR

PR #2553, opened 2025-07-09
perf: shared iterator performance refactor

senojj:
### Problem
Shared iterators aren't quite performing at the level that is needed yet. A few opportunities exist to both simplify the code while improving performance by reducing CPU cycles and allocations.

### Solution
1. The `await` mechanism was changed to utilize a simple atomic boolean and mutex combination.
2. A `fetchMore` function is extracted from `fetchAndWait` and is passed to `await.Do`, allowing the removal of a closure.
3. Read method on the iterator reader is refactored for simplicity.

### Outcome
These changes result in a concurrent stress benchmark of ~600ns reducing to ~130ns per iteration on a hot path.

## Review Checklist

### DIFF
--- a/pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
+++ b/pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
@@ -2,7 +2,6 @@ package sharediterator
 
 import (
 	"context"
-	"fmt"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -547,35 +546,23 @@ const bufferSize = 100
 // The singleflight.Group type was used as a comparison to the await type, but was found to be ~59% slower than await
 // in concurrent stress test benchmarks.
 type await struct {
-	executing atomic.Bool
-	ch        atomic.Pointer[chan struct{}]
+	active atomic.Bool
+	mu     sync.Mutex
 }
 
 // Do executes the provided function fn if it is not already being executed.
 // The channel returned will be closed once the execution has completed.
-func (a *await) Do(fn func()) chan struct{} {
-	i := make(chan struct{})
-	a.ch.CompareAndSwap(nil, &i)
-	ch := *a.ch.Load()
+func (a *await) Do(fn func()) {
+	ex := a.active.Swap(true)
+	a.mu.Lock()
 
-	if !a.executing.Swap(true) {
-		// If we are not already executing, we start a new goroutine to execute.
-		go func() {
-			fn()
-
-			// Important! The executing state must be set to false before closing the old channel.
-			// This avoids a race condition in which a goroutine waiting on the channel close might
-			// enter this function again and try to execute before the execution state is reset.
-			a.executing.Store(false)
-
-			newCh := make(chan struct{})
-			currentCh := a.ch.Swap(&newCh)
-
-			// Close the old channel to signal waiting goroutines to wake up.
-			close(*currentCh)
-		}()
+	if !ex {
+		fn()
+		a.mu.Unlock()
+		a.active.Store(false)
+	} else {
+		a.mu.Unlock()
 	}
-	return ch
 }
 
 // iteratorReader is a wrapper around a storage.Iterator that implements the reader interface.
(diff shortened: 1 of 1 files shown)


pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -554,35 +553,23 @@ const bufferSize = 100
...
-
-			// Close the old channel to signal waiting goroutines to wake up.
-			close(*currentCh)
-		}()
+	if !ex {
+		fn()
+		a.mu.Unlock()
+		a.active.Store(false)
+	} else {
+		a.mu.Unlock()
```
adriantam: instead of locking regardless, will it be better if we check whether ex == true before obtaining lock?

For example,
```
	ex := a.active.Swap(true)
        if ex {
           return
        }
       a.mu.Lock()
       fn()
       a.mu.Unlock()
       a.mu.Store(false)
```

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -594,20 +581,14 @@ type iteratorReader[T any] struct {
...
-			break
+	for i := range buf {
+		t, err := ir.Next(ctx)
+		if err != nil {
+			return i, err
 		}
 		buf[i] = t
 	}
-	return i, e
+	return len(buf), nil
```
adriantam, line 592: We probably need unit test in case where buf < number of iter returned

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
senojj: Nope. The idea is that we want to block the goroutines if a lock is held while fetching... Once the fetching completes, the goroutines will obtain the lock and evaluate `!ex`... those who came in while fetching was in process will evaluate as false and immediately release the lock for the next waiting goroutine.

Your suggestion would allow goroutines to proceed without waiting for new items to be fetched.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
senojj: doh, stupid change by me. Thank you.

senojj: The only time the number of items returned would be less than `len(buf)` is if the iterator was done... in which case, it would return `ErrIteratorDone` and we would return the number of items up to that point on line 587.

senojj: Are you thinking of adding a unit test where the number of tuples in the iterator are greater than the size of the buffer? If so, we have a few tests that use 1K tuples.

senojj: Oh, those are benchmarks though... I likely need to create a test for same.

senojj: Added such a test.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -554,35 +553,23 @@ const bufferSize = 100
...
+	mu     sync.Mutex
 }
 
 // Do executes the provided function fn if it is not already being executed.
 // The channel returned will be closed once the execution has completed.
-func (a *await) Do(fn func()) chan struct{} {
-	i := make(chan struct{})
-	a.ch.CompareAndSwap(nil, &i)
-	ch := *a.ch.Load()
+func (a *await) Do(fn func()) {
```
adriantam, line 563: as discussed, please update comment as channel is no longer used

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -731,56 +694,46 @@ func (s *sharedIterator) clone() *sharedIterator {
...
+	newState := &iteratorState{
+		items: make([]*openfgav1.Tuple, len(state.items)+read),
+		err:   state.err,
+	}
+
+	copy(newState.items, state.items)
+	copy(newState.items[len(state.items):], buf[:read])
+
+	if e != nil {
+		newState.err = e
```
adriantam, line 717: what happens if e == storage.ErrIteratorDone (i.e., thinking of the case where bufferSize is 10 but Read only has 5 item)?  Will the newState's err be storage.ErrIteratorDone?  In this case, will the caller sees the shared iterator as done even though there are still 5 items to return?

senojj: no, the new state will have the additional item(s) _and_ the error. however, if the sharedIterator's head is not yet beyond the bounds of the items, the error is ignored until the head exceeds the bounds.

PR #2556, opened 2025-07-10
Fix: experimental reverse_expand must traverse all DirectEdges from Intersection

justincoh:
Caveat: this is fixing a bug in currently-experimental code.

When we encounter an intersection while traversing the graph in ReverseExpand, we select the lowest-weight edge and then mark that traversal path as `needsCheck`. However, if you have the following relation:

```define viewer: [user, team#member] and _user_in_context```

The current code will take the edge to `[user]` only, and ignore the direct edge to the userset `[team#member]`. This can lead to incorrect results.

#### How is it being solved?
When we encounter an intersection node in traversal:

1. If there are any DirectEdges which have a path to the desired type, we traverse them all regardless of weight
2. If there are no DirectEdges, we select the outgoing edge with the lowest weight
3. Mark all paths as `needsCheck`, regardless of whether we did 1 or 2

* Added a test case that reproduces the problem
* Modify GetEdgesForListObjects to implement the logic in the section above.

## Review Checklist

### CODE REVIEW

File: pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go [hunk 1]
```diff
@@ -216,6 +216,35 @@ func TestReverseExpandWithWeightedGraph(t *testing.T) {
...
+				  relations
+					define can_view: owner or viewer from parent
+					define owner: [user]
+					define parent: [project]
+				type team
+				  relations
+					define member: [user]
+				type project
+				  relations
+    				define viewer: [user, team#member] and _user_in_context
```
justincoh: the existing code ignores the `team#member` direct relation here, and fails this test

pkg/typesystem/typesystem.go
```diff
@@ -1762,13 +1762,22 @@ func (t *TypeSystem) GetEdgesForListObjects(
...
 
-			// return only the lowest weight edge
-			edges = []*graph.WeightedAuthorizationModelEdge{lowestWeightEdge}
+			// If there are any direct edges which reach destination, we have to take them all
+			if len(directEdges) > 0 {
+				edges = directEdges
+			} else {
+				// Otherwise take the lowest weight edge
+				lowestWeightEdge := cheapestEdgeTo(edges, sourceType)
+				edges = []*graph.WeightedAuthorizationModelEdge{lowestWeightEdge}
```
adriantam:
```suggestion
				edges = []*graph.WeightedAuthorizationModelEdge{cheapestEdgeTo(edges, sourceType)}
```

adriantam (approved):

PR #2567, opened 2025-07-16
fix: Reduce number of DB queries needed in ReverseExpand

justincoh:
In the weighted reverse_expand, each individual leaf node kicks off its own independent chain of queries. Each query chain attempts to deduplicate its own work, but for some models these query chains can start to overlap as they move up the graph because they have common ancestors. This model for example, has 3 leaf nodes which all share a common ancestor in `document#viewer`:

```
type user
type document
	relations
		define commenter: [user]
		define editor: [user]
		define writer: [user]
		define viewer: writer or editor or commenter
		define owner: [document#viewer]
```

You can clearly see the common ancestor `document#viewer` in the graph:
image

If a `user` is related to `document:a` via `#commenter`, `#editor`, and `#writer`, all 3 of those independent branches will attempt to read from the DB to find the common ancestor `document#owner@document:a#viewer`. That means this query will run 3 times instead of 1.

#### How is it being solved?
We have an existing `jobDedupeMap`, but it is scoped _within_ each query chain. So multiple query chains which lead to the same tuples will do duplicate work.

Add `queryDedupeMap` to the `ReverseExpandQuery` object itself, so all querying routines will be aware of all duplicate work.

## Review Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of duplicate queries to optimize performance and efficiency. Query deduplication is now managed centrally within the query process, reducing redundant operations during exploration. No changes to user-facing behavior.

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand_weighted.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand_weighted.go
@@ -390,7 +390,7 @@ func (c *ReverseExpandQuery) executeQueryJob(
 	currentReq.relationStack = newStack
 
 	// Ensure that we haven't already run this query
-	if c.isDuplicateQuery(userFilter, typeRel) {
+	if c.isDuplicateQuery(userFilter, typeRel, newStack) {
 		return nil, nil
 	}
 
@@ -420,6 +420,7 @@ func (c *ReverseExpandQuery) executeQueryJob(
 		// and this object is a candidate for return to the user.
 		if currentReq.relationStack == nil {
 			c.trySendCandidate(ctx, needsCheck, foundObject, resultChan)
+			fmt.Printf("SENDING CANDIDATE: %s\n", foundObject)
 			continue
 		}
 
@@ -482,6 +483,7 @@ func buildUserFilter(
 func (c *ReverseExpandQuery) isDuplicateQuery(
 	userFilter []*openfgav1.ObjectRelation,
 	typeRel string,
+	relationStack stack.Stack[typeRelEntry],
 ) bool {
 	objectType, relation := tuple.SplitObjectRelation(typeRel)
 
@@ -489,6 +491,16 @@ func (c *ReverseExpandQuery) isDuplicateQuery(
 	key := utils.Reduce(userFilter, "", func(accumulator string, current *openfgav1.ObjectRelation) string {
 		return current.String() + accumulator
 	})
+
+	// Also need to consider the query's context, as it is possible that many different branches end at the
+	// same leaf or have a common ancestor as part of a unique traversal, and we don't want to kill a unique branch.
+	stackStr := ""
+	for relationStack != nil {
+		var val typeRelEntry
+		val, relationStack = stack.Pop(relationStack)
+		stackStr += fmt.Sprintf("%v", val)
+	}
+	key += stackStr
 
 	key += relation + objectType
 	_, loaded := c.queryDedupeMap.LoadOrStore(key, struct{}{})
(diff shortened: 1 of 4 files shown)


pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -495,8 +492,18 @@ func isDuplicateQuery(
 		return current.String() + accumulator
 	})
 
+	// Also need to consider the query's context, as it is possible that many different branches end at the
+	// same leaf or have a common ancestor as part of a unique traversal, and we don't want to kill a unique branch.
```
justincoh, line 495: This was caught by a model in our tests:
```
model
				  schema 1.1

				type organization
				  relations
					define member: [user]
					define repo_admin: [team#member] or member
				type repo
				  relations
					define admin: [team#member] or repo_admin from owner
					define owner: [organization]
				type team
				  relations
				    define member: [user]
				type user
```
This model has multiple userset leaf nodes `[team#member]`. Without knowing the subsequent query path, we can't dedupe a query for `team, #member, @user:justin` as it is the beginning of two separate resolution paths. Only when the *entire* ancestry tree is the same can we stop a query early.

Image

`team#member` above belongs to two different paths, so we have to allow it to be queried by each unique path.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -424,6 +420,7 @@ func (c *ReverseExpandQuery) executeQueryJob(
 		// and this object is a candidate for return to the user.
 		if currentReq.relationStack == nil {
 			c.trySendCandidate(ctx, needsCheck, foundObject, resultChan)
+			fmt.Printf("SENDING CANDIDATE: %s\n", foundObject)
```
cikasfm: Remove from prod code; or change to `log.Debug`?

elbuo8: ?

justincoh: yeah that shouldn't have made it in here

cikasfm (commented):
You could achieve similar results by using `golang.org/x/sync/singleflight`; see https://github.com/openfga/openfga/blob/e147e4b9dc9e403f1f01de963030ab812ebb3de7/pkg/typesystem/resolver.go#L35

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: removed

justincoh: yeah missed this one during cleanup, removed

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -495,8 +491,18 @@ func isDuplicateQuery(
...
 	})
 
+	// Also need to consider the query's context, as it is possible that many different branches end at the
+	// same leaf or have a common ancestor as part of a unique traversal, and we don't want to kill a unique branch.
+	stackStr := ""
+	for relationStack != nil {
+		var val typeRelEntry
+		val, relationStack = stack.Pop(relationStack)
+		stackStr += fmt.Sprintf("%v", val)
+	}
```
adriantam, line 501: nitpik (and can be addressed in follow up PR): this can be made to be a property of the stack class.  In this way, this code can be used and simplified the reader in reading this code.

adriantam (approved):

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
justincoh: I did consider adding a `String()` method in the stack package which does exactly this. I ended up doing it here since this is so specific to reverse_expand. But then again, the entire package was created because of reverse_expand, so I could go either way.

PR #2568, opened 2025-07-17
short circuit iteratorsToUserset

elbuo8:
Upon timeout, iteratorsToUserset takes a long time to drain.

#### How is it being solved?
Forcing all goroutine to exit as soon as possible

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of cancellation and error scenarios to provide clearer feedback when operations are interrupted or terminated early.

### DIFF
--- a/internal/graph/object_providers.go
+++ b/internal/graph/object_providers.go
@@ -2,6 +2,7 @@ package graph
 
 import (
 	"context"
+	"errors"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 
@@ -148,7 +149,7 @@ func iteratorsToUserset(ctx context.Context, chans []<-chan *iterator.Msg, out c
 			for {
 				select {
 				case <-ctx.Done():
-					return nil
+					return ErrShortCircuit
 				case msg, ok := <-c:
 					if !ok {
 						open = false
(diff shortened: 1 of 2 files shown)


yissellokta (approved):

internal/graph/object_providers.go
```diff
@@ -148,22 +149,25 @@ func iteratorsToUserset(ctx context.Context, chans []<-chan *iterator.Msg, out c
...
 					for {
 						t, err := msg.Iter.Next(ctx)
 						if err != nil {
 							msg.Iter.Stop()
 							if storage.IterIsDoneOrCancelled(err) {
-								break
+								if errors.Is(err, storage.ErrIteratorDone) {
+									break
+								}
+								return ErrShortCircuit
```
adriantam, line 170: to make code easier to understand, should we do something like this
```
if err != nil {
							msg.Iter.Stop()
								break
							}
							if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
								return ErrShortCircuit
							}
```

adriantam (commented):
Do we need to drain the iterator on ErrShortCircuit?

elbuo8:
> Do we need to drain the iterator on ErrShortCircuit?

Yes. We already need to drain iterators regardless if thrown away.

internal/graph/object_providers.go
elbuo8: I started with that and had realized I missed `DeadlineExceeded`. I prefer to keep this way since we already took the time to encapsulate what `ctx.Done` means. If we already have functions for these same reasons, I'd rather re-use the accordingly to prevent these subtle mistakes.

adriantam (approved):

PR #2569, opened 2025-07-17
fix: use pool per intersection/exclusion so as not to block

Vic-Dev:
Improve list objects performance where experimental flag is enabled. Currently, intersection and exclusion experiences long latency because check for the candidates are done serially.

#### How is it being solved?
Process list objects candidate asynchronously so that multiple candidates can be processed concurrently.

1. Update callCheckForCandidates and findCandidatesForLowestWeightEdge to use the same pool as loopOverEdges.  This allows better concurrency and reduces the wait when intersection/exclusion is encountered.
2. The candidates (tmpResult) for findCandidatesForLowestWeightEdge are processed in separate go routine. This allows multiple candidates to be processed concurrently.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved performance for list objects intersection and exclusion operations when a specific optimization flag is enabled.

* **Refactor**
  * Enhanced concurrency management in list object operations to improve efficiency and reduce resource contention.

* **Tests**
  * Updated tests to use a concurrency pool for more accurate simulation of concurrent operations and improved error handling.

* **Documentation**
  * Updated changelog to reflect recent performance improvements and bug fixes.

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -96,6 +96,11 @@ type ListObjectsResolutionMetadata struct {
 
 	// WasWeightedGraphUsed indicates whether the weighted graph was used as the algorithm for the ListObjects request.
 	WasWeightedGraphUsed *atomic.Bool
+
+	// Temporary solution to indicate whether shadow list objects query should be run.
+	// For queries with Infinite weight, the weighted graph implementation falls back
+	// to the original code, making any comparison useless.
+	ShouldRunShadowQuery *atomic.Bool
 }
 
 func NewListObjectsResolutionMetadata() ListObjectsResolutionMetadata {
@@ -104,6 +109,7 @@ func NewListObjectsResolutionMetadata() ListObjectsResolutionMetadata {
 		DispatchCounter:      new(atomic.Uint32),
 		WasThrottled:         new(atomic.Bool),
 		WasWeightedGraphUsed: new(atomic.Bool),
+		ShouldRunShadowQuery: new(atomic.Bool),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -589,52 +644,17 @@ func (c *ReverseExpandQuery) callCheckForCandidates(
...
-			// If the allowed value does not match what we expect, we skip this candidate.
-			// eg, for intersection we expect the check result to be true
-			// and for exclusion we expect the check result to be false.
-			if tmpCheckResult.GetAllowed() != isAllowed {
-				continue
-			}
+		// note that we create a separate goroutine pool instead of the main pool
+		// to avoid starvation on the main pool as there could be many candidates
+		// arriving concurrently.
+		tmpResultPool := concurrency.NewPool(ctx, int(c.resolveNodeBreadthLimit))
```
justincoh, line 650: My only concern would be the limit here. We could have multiple `callCheckForCandidates` in flight simultaneously, since `findCandidatesForLowestWeightEdge` kicks off new traversals, which could hit more intersections / exclusions, and so on. That _could_ lead to `resolveNodeBreadthLimit * numberOfFunctionCalls` goroutines going live altogether.

We should get a good sense of how this behaves via the shadow checker, but something to keep in mind

adriantam: Agree on this, this can be updated later on to reduce the pool size if this is the case.  However, this needs to be a separate pool to avoid deadlock / starvation.

PR #2572, opened 2025-07-21
fix: Don't run shadow ListObjects query for request weight=∞

justincoh:
Right now the shadow list objects query is configured to run on a user-defined % of listObjects requests. However, the new ReverseExpand code, which the shadow resolver is designed to test, falls back to the current implementation (non-weighted) when it receives queries of infinite weight. That means that for some models, the shadow query is just re-running the exact same query as the main list objects query; doing duplicate work and providing no value.

#### How is it being solved?
Added some additional metadata to prevent running an additional shadow query when it isn't meaningful.

If the query is going to skip the weighted graph, we will no longer run the shadow query on it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of shadow queries by preventing reporting on requests that use the old reverse expansion path.
  * Enhanced tracking of weighted graph usage in object listing operations for more accurate metadata reporting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 - Update ReverseExpand to use a LinkedList to track its relation stack for performance. [#2542](https://github.com/openfga/openfga/pull/2542)
 - Update ReverseExpand to use a intersection and exclusion handler to fast path check calls. [#2543](https://github.com/openfga/openfga/pull/2543)
 - Deduplicate queries more effectively in ReverseExpand. [#2567](https://github.com/openfga/openfga/pull/2567)
+- Update go version to 1.24.5 [#2577](https://github.com/openfga/openfga/pull/2577)
 
 ### Fixed
 - Shared iterator race condition and deadlock. [#2544](https://github.com/openfga/openfga/pull/2544)
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.38@sha256:cacececcc3e94fde691141b80c34e288bc2743323d4f450274a0c636d07400a9 AS grpc_health_probe
-FROM cgr.dev/chainguard/go:1.24.2@sha256:f08c86a2abec9cf7518eb1b561e40627929b56d40f2196b632eecb0589dc1504 AS builder
+FROM cgr.dev/chainguard/go:1.24.5@sha256:894edfc7e7e6d2a2cd3cf53025da0c6a3a6a8769c30dfe0f2b5bcff7229f195e AS builder
 
 WORKDIR /app
 
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/openfga/openfga
 
-go 1.24.2
-
-toolchain go1.24.3
+go 1.24.5
 
 require (
 	github.com/Masterminds/squirrel v1.5.4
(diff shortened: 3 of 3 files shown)


pkg/server/commands/list_objects_shadow.go
```diff
@@ -209,6 +209,11 @@ func (q *shadowedListObjectsQuery) executeShadowModeAndCompareResults(parentCtx
 		return
 	}
 
+	// Don't report on requests that fell through to the old reverse_expand (infinite weight requests)
```
poovamraj: Should we move this code above to 194 so that we don't run anything in this method saving resource. Why do shadow when we know we are running the old algorithm.

poovamraj (changes requested):

pkg/server/commands/list_objects_shadow.go
justincoh: recapping slack conversation: this requires a good amount of copy and pasted code direct from ReverseExpand, which we'd rather not add in here. Instead, adding an additional comment string in here and leaving as is.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -209,6 +209,12 @@ func (q *shadowedListObjectsQuery) executeShadowModeAndCompareResults(parentCtx
 		return
 	}
 
+	// Don't report on requests that fell through to the old reverse_expand
+	// NOTE: remove this once weighted graph code supports infinite weight queries
+	if !shadowRes.ResolutionMetadata.WasWeightedGraphUsed.Load() {
+		return
+	}
+
```
cikasfm: This works...but I'm thinking, is there a way to avoid running the WG query if we know that its in shadow mode and we know we can't use WG and we fallback instead?

cikasfm (approved):
LGTM

But is there a way to avoid running the WG query if we know that its in shadow mode and we know we can't use WG and we fallback instead?

pkg/server/commands/list_objects_shadow.go
justincoh: Yeah Poovam and I spoke about this earlier and I think we're going to add an additional piece of metadata to ReverseExpand and ListObjects like `ShouldRunShadowQuery`, to indicate whether it would be valuable.

poovamraj (dismissed):

poovamraj (dismissed):

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -424,6 +430,22 @@ func (c *ReverseExpandQuery) execute(
...
+		if !ok {
+			resolutionMetadata.ShouldRunShadowQuery.Store(false)
+		} else {
+			weight, _ := node.GetWeight(sourceUserType)
+			if weight == weightedGraph.Infinite {
+				resolutionMetadata.ShouldRunShadowQuery.Store(false)
+			}
+		}
+	}
+
```
cikasfm: We can probably move this duplicating code combined with lines [385-404](https://github.com/openfga/openfga/pull/2572/files#diff-a38cc14cbddeddd8dbfaa62475bff6cdc19100cab41b87c150c4c431e44da843R385-R404) into a block _**before**_ the weighed graph resolution ( ie 383 ) and simplify the whole thing.

pkg/server/commands/list_objects_shadow.go
```diff
@@ -271,6 +271,11 @@ func (q *shadowedListObjectsQuery) checkShadowModePreconditions(ctx context.Cont
 			return false
 		}
 
+		if !res.ResolutionMetadata.ShouldRunShadowQuery.Load() {
+			q.logger.DebugWithContext(ctx, "shadowed list objects query skipped due to infinite weight query")
+			return false
+		}
```
cikasfm: please add common log fields

```golang
		if !res.ResolutionMetadata.ShouldRunShadowQuery.Load() {
			q.logger.DebugWithContext(ctx, "shadowed list objects query skipped due to infinite weight query"),
				loShadowLogFields(req)...,
			)
			return false
		}
```

cikasfm (approved):

PR #2573, opened 2025-07-21
perf: change shared iterator to a more traditional single-flight implementation

senojj:
The current implementation creates a `sharedIterator` instance that is shared with any callers for the same key until one of two events occurs:

- The max admission time has expired.
- The count of references to the sharedIterator instance reaches 0.

The idea is that this solution should allow callers to fetch a copy of a live iterator as long as it is currently being fetched, iterated over, or has not exceeded its TTL or iterator count threshold. By allowing a clone to be made for as long as there is ongoing iteration, the possibility of multiple requests to share the same iterator increases. By enforcing a TTL, the potential staleness of an iterator’s data is reduced. By enforcing an iterator count threshold, the potential for excessive memory consumption is mitigated.

However, the `sharedIterator` instance is often returned immediately and stopped soon after, before any other goroutines have had a chance to clone the instance and increase its internal reference counter. Therefore, the current implementation is ineffective at reducing back-end calls.

#### How is it being solved?
To solve the issue, a more "traditional" single-flight approach is used. The first goroutine to call for a given key is the one that will issue the single back-end request. All other goroutines will wait for the initial goroutine to return a result. However, instead of keeping the shared iterator around in a "cache" for further cloning, subsequent goroutines calling for the same key will be part of a new single-flight group, yielding a new shared iterator.

This implementation eliminates the need for any kind of "cache eviction", either by TTL or reference count.

`group` and `call` types are introduced for handling the single-flight implementation. It is fairly similar to the Go development team's approach, however it has a slight modification for this specific use case.

Upon receiving a newly created `sharedIterator` instance in the goroutine that performed the single-flight, a `sharedIterator` must be allocated for each blocked participating goroutine and cloned before the initial `sharedIterator` instance is returned. This prevents the calling code from stopping the `sharedIterator` instance before any other goroutines have had a chance to clone it, effectively preventing clone creation.

Additionally, now unused parameters, default values, configurations, and metrics have been removed.

## Review Checklist

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -110,7 +110,10 @@
                 "type": "integer"
             },
             "minItems": 1,
-            "default": [50, 200],
+            "default": [
+                50,
+                200
+            ],
             "x-env-variable": "OPENFGA_REQUEST_DURATION_DATASTORE_QUERY_COUNT_BUCKETS"
         },
         "requestDurationDispatchCountBuckets": {
@@ -121,7 +124,10 @@
                 "type": "integer"
             },
             "minItems": 1,
-            "default": [50, 200],
+            "default": [
+                50,
+                200
+            ],
             "x-env-variable": "OPENFGA_REQUEST_DURATION_DISPATCH_COUNT_BUCKETS"
         },
         "contextPropagationToDatastore": {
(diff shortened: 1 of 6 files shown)


senojj:
@coderabbitai review

senojj:
@coderabbitai review

senojj:
Closed in favor of #2590

PR #2574, opened 2025-07-21
feat: shadow list objects resolver should use its own cache

adriantam:
The check resolver used in shadow list objects do not provide accurate latency result because it uses cached value from the main list objects' check resolver.   

#### How is it being solved?

Shadow list object's check resolver should use its own cache.

1. Create separate cache and cache controller for shadow list objects' check resolver.
2. When shadow list objects is enabled, ensure shadow list object cache controller invalidates the shadow list object cache.  In addition, serve from shadow list object cache for the shadow list objects' check resolver.

## Review Checklist


justincoh (approved):

internal/shared/shared.go
```diff
@@ -74,6 +83,27 @@ func NewSharedDatastoreResources(
...
+		s.ShadowCheckCache, err = storage.NewInMemoryLRUCache([]storage.InMemoryLRUCacheOpt[any]{
+			storage.WithMaxCacheSize[any](int64(settings.CheckCacheLimit)),
+		}...)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if settings.ShouldCreateShadowCacheController() {
+		s.ShadowCacheController = cachecontroller.NewCacheController(ds, s.ShadowCheckCache, settings.CacheControllerTTL, settings.CheckIteratorCacheTTL, cachecontroller.WithLogger(s.Logger))
```
cikasfm, line 104: There's one interesting case to consider:
- if we don't create a shadow cache ( but reuse the main cache ) AND create a shadow cache controller: we might end up having 2 cache invalidation controllers on the same cache, which may or may not be a problem, but something to consider.

I think we should simplify to a single flag for all - shadow cache enabled, new shadow cache and new shadow cache invalidation controller. I don't see a reason why those should be separated.

justincoh (approved):

cikasfm (approved):

internal/shared/shared.go
cikasfm: Ignore the comment above, those flags are derived from the main one anyway

PR #2581, opened 2025-07-23
fix: add store_id and authorization_model_id to the LO logs

Vic-Dev:
## Description
Add `store_id` and `authorization_model_id` to the ListObjects error logs for easier debugging

## Review Checklist

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -375,7 +375,7 @@ func (c *ReverseExpandQuery) execute(
 
 		// ReverseExpand(type=document, rel=viewer, user=document:1#viewer) will return "document:1"
 		if tuple.UsersetMatchTypeAndRelation(userset.String(), req.Relation, req.ObjectType) {
-			c.trySendCandidate(ctx, intersectionOrExclusionInPreviousEdges, sourceUserObj, resultChan)
+			c.trySendCandidate(ctx, req, intersectionOrExclusionInPreviousEdges, sourceUserObj, resultChan)
 		}
 	}
 
@@ -392,12 +392,20 @@ func (c *ReverseExpandQuery) execute(
 				// The weighted graph is not guaranteed to be present.
 				// If there's no weighted graph, which can happen for models with tuple cycles, we will log an error below
 				// and then fall back to the non-weighted version of reverse_expand
-				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph", zap.String("nodeID", typeRel))
+				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph",
+					zap.String("authorization_model_id", c.typesystem.GetAuthorizationModelID()),
+					zap.String("store_id", req.StoreID),
+					zap.String("node_id", typeRel),
+				)
 				req.skipWeightedGraph = true
 			} else {
 				weight, _ := node.GetWeight(sourceUserType)
 				if weight == weightedGraph.Infinite {
-					c.logger.InfoWithContext(ctx, "reverse_expand graph may contain cycle, skipping weighted graph", zap.String("nodeID", typeRel))
+					c.logger.InfoWithContext(ctx, "reverse_expand graph may contain cycle, skipping weighted graph",
+						zap.String("authorization_model_id", c.typesystem.GetAuthorizationModelID()),
+						zap.String("store_id", req.StoreID),
+						zap.String("node_id", typeRel),
+					)
 					req.skipWeightedGraph = true
 				}
 			}
@@ -728,7 +736,13 @@ LoopOnIterator:
 	return nil
 }
 
-func (c *ReverseExpandQuery) trySendCandidate(ctx context.Context, intersectionOrExclusionInPreviousEdges bool, candidateObject string, candidateChan chan<- *ReverseExpandResult) {
+func (c *ReverseExpandQuery) trySendCandidate(
+	ctx context.Context,
+	req *ReverseExpandRequest,
+	intersectionOrExclusionInPreviousEdges bool,
+	candidateObject string,
+	candidateChan chan<- *ReverseExpandResult,
+) {
 	_, span := tracer.Start(ctx, "trySendCandidate", trace.WithAttributes(
 		attribute.String("object", candidateObject),
 		attribute.Bool("sent", false),
@@ -747,7 +761,11 @@ func (c *ReverseExpandQuery) trySendCandidate(ctx context.Context, intersectionO
 		if ok {
 			span.SetAttributes(attribute.Bool("sent", true))
 		} else {
-			c.logger.ErrorWithContext(ctx, "failed to send candidate object", zap.String("object", candidateObject))
+			c.logger.ErrorWithContext(ctx, "failed to send candidate object",
+				zap.String("authorization_model_id", c.typesystem.GetAuthorizationModelID()),
+				zap.String("store_id", req.StoreID),
+				zap.String("object", candidateObject),
+			)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -392,12 +392,20 @@ func (c *ReverseExpandQuery) execute(
 				// The weighted graph is not guaranteed to be present.
 				// If there's no weighted graph, which can happen for models with tuple cycles, we will log an error below
 				// and then fall back to the non-weighted version of reverse_expand
-				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph", zap.String("nodeID", typeRel))
+				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph",
```
senojj: is this intended to be a DEBUG log?

senojj: INFO seems too verbose.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -747,7 +761,11 @@ func (c *ReverseExpandQuery) trySendCandidate(ctx context.Context, intersectionO
 		if ok {
 			span.SetAttributes(attribute.Bool("sent", true))
 		} else {
-			c.logger.ErrorWithContext(ctx, "failed to send candidate object", zap.String("object", candidateObject))
+			c.logger.ErrorWithContext(ctx, "failed to send candidate object",
+				zap.String("authorization_model_id", c.typesystem.GetAuthorizationModelID()),
+				zap.String("store_id", req.StoreID),
+				zap.String("object", candidateObject),
+			)
```
senojj: any reason we log here instead of returning an error? could be logged higher up in the stack.

justincoh (commented):
Can we get by with just the auth model ID? That should be enough to backtrack to a store, and that way we don't have to pass `req` objects through the various functions in here

pkg/server/commands/reverseexpand/reverse_expand.go
cikasfm: @senojj we try to log all the reasons why WG could not be used, so `Info` makes sense here. And TBH this would probably be an ultra-rare case based no our knowledge, but even then it would be valuable to know, which customer/store causes this.

senojj: does it _always_ need to be in the output when it happens? or just when we suspect a problem?

Vic-Dev:
> Can we get by with just the auth model ID? That should be enough to backtrack to a store, and that way we don't have to pass `req` objects through the various functions in here

Passed through the context instead

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -644,12 +667,7 @@ func (c *ReverseExpandQuery) intersectionHandler(
 ) error {
 	intersectionEdgeComparison, err := typesystem.GetEdgesForIntersection(edges, sourceUserType)
 	if err != nil {
-		c.logger.Error("Failed to get lowest weight edge",
-			zap.String("function", "intersectionHandler"),
-			zap.Error(err),
-			zap.Any("edges", edges),
-			zap.String("sourceUserType", sourceUserType))
-		return err
+		return fmt.Errorf("%w: operation: intersection: %s", ErrLowestWeightFail, err.Error())
```
cikasfm, line 670: Why wrap custom error, and not the root cause? If you already know the error type, you can add the string here, but wrap the `original` error instead

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -677,11 +695,7 @@ func (c *ReverseExpandQuery) intersectionHandler(
 		userset, err := c.typesystem.ConstructUserset(sibling)
 		if err != nil {
 			// This should never happen.
-			c.logger.Error("Failed to construct userset",
-				zap.String("function", "intersectionHandler"),
-				zap.Any("edge", sibling),
-				zap.Error(err))
-			return err
+			return fmt.Errorf("%w: operation: intersection: %s", ErrConstructUsersetFail, err.Error())
```
cikasfm, line 698: same as above

> Why wrap custom error, and not the root cause? If you already know the error type, you can add the string here, but wrap the `original` error instead

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -715,12 +729,7 @@ func (c *ReverseExpandQuery) exclusionHandler(
 ) error {
 	baseEdges, excludedEdge, err := typesystem.GetEdgesForExclusion(edges, sourceUserType)
 	if err != nil {
-		c.logger.Error("Failed to get lowest weight edge",
-			zap.String("function", "exclusionHandler"),
-			zap.Error(err),
-			zap.Any("edges", edges),
-			zap.String("sourceUserType", sourceUserType))
-		return err
+		return fmt.Errorf("%w: operation: exclusion: %s", ErrLowestWeightFail, err.Error())
```
cikasfm, line 732: same as above

> Why wrap custom error, and not the root cause? If you already know the error type, you can add the string here, but wrap the `original` error instead

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -744,11 +753,7 @@ func (c *ReverseExpandQuery) exclusionHandler(
 	userset, err := c.typesystem.ConstructUserset(excludedEdge)
 	if err != nil {
 		// This should never happen.
-		c.logger.Error("Failed to construct userset",
-			zap.String("function", "exclusionHandler"),
-			zap.Any("edge", excludedEdge),
-			zap.Error(err))
-		return err
+		return fmt.Errorf("%w: operation: exclusion: %s", ErrConstructUsersetFail, err.Error())
```
cikasfm, line 756: same as above

> Why wrap custom error, and not the root cause? If you already know the error type, you can add the string here, but wrap the `original` error instead

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
senojj: the original error is just a string. it is not assertable with `errors.Is` or `errors.As`.

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -396,12 +396,12 @@ func (c *ReverseExpandQuery) execute(
 				// The weighted graph is not guaranteed to be present.
 				// If there's no weighted graph, which can happen for models with tuple cycles, we will log an error below
 				// and then fall back to the non-weighted version of reverse_expand
-				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph", zap.String("nodeID", typeRel))
+				c.logger.InfoWithContext(ctx, "unable to find node in weighted graph", zap.String("node_id", typeRel))
```
senojj, line 399: could we also capture the context tags in InfoWithContext like is being done in ErrorWithContext?

Vic-Dev: Will do as a fast follow up

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
senojj, line 670: Actually, since Go 1.20 `fmt.Errorf` is able to wrap multiple errors. Before that version, only one `%w` could be used per format string... Therefore, it would be a good idea to do something like this instead:
```suggestion
		return fmt.Errorf("%w: operation: intersection: %w", ErrLowestWeightFail, err)
```

senojj: While the second error is only a string as of this time, if that changes in the future, we won't need to change this code to adapt it.

PR #2584, opened 2025-07-25
chore: Benchmarks for ListObjects

justincoh:
We don't currently have great benchmark coverage of ListObjects. We have [an existing benchmark](https://github.com/openfga/openfga/blob/main/pkg/server/test/list_objects.go#L638), but it doesn't distinguish between the various weights of relationships, and in the future our different optimization strategies will be based on the weight of the query being run.

#### How is it being solved?
This benchmark suite creates a single model with relations of weights 1, 2, 3, and INF. It then writes 5k tuples for each of those relationship weights and runs a specific assertion against each relation weight. It does the whole thing twice, once `WithListObjectsOptimizationsEnabled(true)` and once `WithListObjectsOptimizationsEnabled(false)`.

These Benchmarks also add additional reported metrics for `max_us`, `min_us`, `p95_us`, and `p99_us`, exactly how we currently have them [in the shared_iterator test suite](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore_test.go#L305)

## Notes from local testing:
These benchmarks are run _without_ a `maxResults`, and every case is designed to return 5k tuples; significantly more than the default ListObjects item return limit of 1k objects.

---
### Summary
Weighted graph implementation is equal to or faster than non-weighted for all cases, and the weighted graph implementation uses significantly fewer resources in all cases.

Full results below, running with the same setup as our `make test-bench`. To reproduce run `go test -bench BenchmarkListObjects ./pkg/server/commands/. -benchtime 1s -cpu 1 -benchmem -count 1`

image

These results *do not* hold when we increase the CPU count above ~4. At that point the non-optimization code outperforms the optimization-enabled code by a decent margin (on weight_two and weight_three). We know exactly [why this is](https://github.com/openfga/openfga/compare/main...test/bump-querying-goroutine-limit), and tweaking the limiter value in that link fixes the issue. We're currently testing the CPU implications of merging the code [in that link](https://github.com/openfga/openfga/compare/main...test/bump-querying-goroutine-limit) to resolve the issue.

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive benchmark suite to evaluate the performance of listing objects with varying relationship complexities.
  * Introduced new helpers for efficiently setting up test data across different relationship types for benchmarking.


pkg/server/commands/list_objects_test.go
```diff
@@ -576,3 +579,282 @@ func TestAttemptsToInvalidateWhenIteratorCacheIsEnabled(t *testing.T) {
...
+				relations
+					define parent: [company]
+					define weight_three: org_member from parent
+		`).GetTypeDefinitions(),
+	}
+	ctx := context.Background()
+	err := datastore.WriteAuthorizationModel(ctx, storeID, model)
+	require.NoError(b, err)
+
+	n := 5000
```
justincoh, line 665: @senojj the previous `10000` was timing out the test. I added logs and it was failing with a different number of results (6-8k) each time for the weight_three test cases

cikasfm: can we just use `b.N` here?

cikasfm: Nevermind, I agree we need some static value

pkg/server/commands/list_objects_test.go
```diff
@@ -576,3 +580,257 @@ func TestAttemptsToInvalidateWhenIteratorCacheIsEnabled(t *testing.T) {
...
+	optimizationsEnabled bool,
+	query ListObjectsQuery,
+	request *openfgav1.ListObjectsRequest,
+) {
+	if optimizationsEnabled {
+		name += "_with_optimization"
+		query.optimizationsEnabled = true
+	} else {
+		query.optimizationsEnabled = false
+	}
```
Vic-Dev, line 614: If we separate this part out, we can reuse this logic for other benchmarks

justincoh: Was looking back at this just now and there's also the issue of the `require.Len(b res.Objects, 5000)` below. For now I'd rather leave this specific to this test file; if someone needs similar functionality in a future benchmark we can pick it apart then

PR #2590, opened 2025-07-29
perf: shared iterators with timed eviction

senojj:
The current implementation creates a `sharedIterator` instance that is shared with any callers for the same key until one of two events occurs:

- The max admission time has expired.
- The count of references to the sharedIterator instance reaches 0.

The idea is that this solution should allow callers to fetch a copy of a live iterator as long as it is currently being fetched, iterated over, or has not exceeded its TTL or iterator count threshold. By allowing a clone to be made for as long as there is ongoing iteration, the possibility of multiple requests to share the same iterator increases. By enforcing a TTL, the potential staleness of an iterator’s data is reduced. By enforcing an iterator count threshold, the potential for excessive memory consumption is mitigated.

However, the `sharedIterator` instance is often returned immediately and stopped soon after, before any other goroutines have had a chance to clone the instance and increase its internal reference counter. Therefore, the current implementation is ineffective at reducing back-end calls.

#### How is it being solved?
To solve the issue, shared iterators are no longer evicted from the cache when their reference counter reaches 0. Instead, they will remain in the cache for the duration of the admission time. However, to mitigate the problem of unused shared iterators hanging around for longer than necessary, a max idle time is introduced that will evict a shared iterator from the cache if it has not been cloned for the duration of the max idle time.

- A max idle time is introduced.
- sharedIterator instances no longer use a `closed` field, and instead no longer allow cloning after the current instance has been stopped.
- The sharedIterator Close and clone methods now utilize the instance local RWMutex. This is primarily what allowed the `closed` field to be removed. This also reduces the opportunity for a race condition regression.
- A sharedIteratorCloneCount metric is introduced to track how many clones of shared iterators exist at any given moment. This will allow us to gauge the efficacy of the feature.
- Tests are added to evaluate the max idle time mechanism.
- Tests are removed that are no longer applicable due to relying entirely on timed eviction.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced idle timeout management for shared iterators, automatically cleaning up unused iterators after a configurable idle period.
  * Added a Prometheus metric to track the number of active shared iterator clones.

* **Improvements**
  * Enhanced lifecycle management for shared iterators to ensure timely cleanup and resource efficiency.
  * Improved concurrency handling for shared iterators.

* **Bug Fixes**
  * Addressed potential issues with stale iterators by enforcing admission and idle timeouts.

* **Tests**
  * Added tests for iterator expiration and cleanup based on admission and idle timeouts.
  * Introduced a new benchmark to measure read latency under varying loads.
  * Removed outdated benchmarks and metric tests.


pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -271,31 +289,18 @@
...
-		return newIterator, nil
+		return si, nil
+	}
+
+	newStorageItem.cleanup = func() {
+		if sf.internalStorage.rswu.CompareAndDelete(cacheKey, newStorageItem) {
+			sf.internalStorage.ctr.Add(-1)
+			sharedIteratorCount.Dec()
+		}
 	}
```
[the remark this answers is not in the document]
senojj: if the CompareAndDelete call returns false, then the sharedIteratorCount was already decremented elsewhere.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -113,21 +122,29 @@ func (s *storageItem) unwrap() (*sharedIterator, bool, error) {
 
 	s.once.Do(func() {
 		s.iter, s.err = s.producer()
+		if s.err != nil {
+			return
+		}
+
+		admissionTimer := time.AfterFunc(s.admissionDuration, func() {
+			s.iter.Stop()
```
elbuo8, line 130: Why were `s.iter.Stop` calls moved into these timers if you originally closed it immediately (conditionally) after clone?

senojj:
Depends on #2608

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
```diff
@@ -113,21 +122,29 @@ func (s *storageItem) unwrap() (*sharedIterator, bool, error) {
...
+			s.cleanup()
+		})
+
+		s.idleTimer = time.AfterFunc(s.idleDuration, func() {
```
elbuo8, line 134: So effectively the watch dog of idleness comes back.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
elbuo8: No issue that these 2 timers fire together?

senojj: So in this scenario, we want the iterator to remain "alive" (take up a reference in the counter) for as long as one of the eviction times has not expired. This makes the shared iterator "clonable" for the duration of its lifetime in the "cache". 

If we were to stop the shared iterator immediately after the first clone, its reference counter would be released, dropping the value to 1 (held by the one clone). The clone would then be returned to the caller, and potentially iterated and stopped (dropping the reference counter to 0) before any of the other waiting goroutines have had a chance to clone the iterator. Therefore the cached shared iterator would not be "clonable" for the duration of its time in the cache.

senojj: In the current implementation, the clone is often returned, iterated, and stopped before any other waiting goroutines have had an opportunity to awaken and clone the shared iterator instance from the cache. This resulted in an extremely ineffective solution.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
senojj: Effectively, yes.

senojj: Except the previous iteration of the watchdog timer would stop each of the shared iterators by stopping the inner iterator (if my memory serves me well). This timer only stops the cached instance.

pkg/storage/storagewrappers/sharediterator/shared_iterator_datastore.go
senojj: The two timers can fire together, or independently in any order... Their cleanup is an idempotent operation that only succeeds if the shared iterator pointer for the given cache key is equal to the pointer at that key when the timer was created.

Issue #2151, opened 2024-12-09
Provide a way to update the context of a tuple with condition in a single transaction


thomas-rudlof-bl:
### Checklist

- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

If you have a tuple with condition and (partial) context and want to update the (partial) context, currently you first have to remove the tuple and then add it with the new context. Unfortunately, this cannot be done in one single write operation, since it involves the same tuple twice - once in the Writes list and once in the Deletes list - which is rejected by the API.

So it must be done in two separate requests which results in a raise condition: Should a check be done after the deletion of the old tuple but before the insertion of the new tuple, it might fail where it should not have failed. 

Use case: We deduce access privileges in our system from JWTs of another system. We enter the induced tuples in our model on reception and validation of the access tokens, but guard them with the expiration time of those token. Should the user need longer access, a new access token needs to be sent and the expiration time of the corresponding tuples need to be updated.

### Describe the ideal solution

Provide a way to update the context of one or several tuples in a single transaction.

For convenience, it would be nice, if the insertion of a new tuple or the update of an existing tuple could be done with the same kind of request. 

I would love to have an idempotent counterpart of the write operation:
* deleting non existing tuples does not abort and roll back the operation
* adding existing tuples does not abort and roll back the operation but replaces their context with the one given in the operation

hamilton-earthscope:
Now that we have `on_duplicate==ignore` during writes...

According to https://openfga.dev/docs/getting-started/update-tuples#05-ignoring-duplicate-or-missing-tuples
> Conflicting Tuples: If a tuple key (user, relation, object) matches an existing tuple, but the condition name or parameters are different, this is a conflict. The write attempt will be rejected, and the entire transaction will fail with a 409 Conflict error.


It would be great to have a `replace` or `upsert` write mode so we can easily replace the condition/context associated with an existing tuple

loeffel-io:
seeing this and can't believe that this isn't implemented. we are not able to transactional safe delete and write 

**our use case** 

we want to update the `endTime` condition of the old subscription entitlement tuple. we expected that within a single
request, openfga internally applies the delete before the write.

```bash
name: "projects/mindful/authorizationModel"
writes { <--------------------------------------------------
  user: "account:QepeMQ5CyYhMk5W1ilDTOyjdMEj2"
  relation: "account"
[...]

PR #2593, opened 2025-07-30
feat: datastore idempotent tuples write

le-yams:
Hello there,

here is a PR to add [idempotency support](https://github.com/openfga/roadmap/issues/79).

## Description

A parameter allows to make idempotent tuple writes, when activated:
* deleting non existing tuple should not fail
* adding an existing tuple should not fail
    * the condition should be updated

#### How is it being solved?

When idempotency is enabled (false by default):
 - deletes ignore the ErrInvalidWriteInput error
 - inserts are turned into upserts

 * added options on the `storage.RelationshipTupleWriter` interface `Write` method (non breaking). 
 * added a `UpsertSuffixForTuple` field in `storage.DBInfo` so each storage implementation can provide the SQL suffix needed to turn the insert into an upsert.
 * implemented the `Write` method idempotency in sqlcommon.go (mysql and postgres) and in sqlite.go (sqlite).
 * added related tests in `TupleWritingAndReadingTest`

What remains to be done:
 * add the idempotency parameter in the WriteCommand
 * add the idempotency parameter (optional and false by default) in the api write endpoint
 * update the documentation accordingly

## References
 * closes https://github.com/openfga/roadmap/issues/79

## Review Checklist


le-yams:
I opened this PR as a draft as I only implemented the storage part (command and endpoint remains to be done). 
I wanted your feedback before going further :pray:.
Regards,
Yann

cikasfm:
@le-yams this is a great initiative!

We want to review this and follow up with it. There are a few considerations we need to address, like we cannot simply `upsert` or ignore delete failures, writes & deletes must to be transactional together with `changelog` changes, i.e. the `changelog` must be correct no matter what.

We'll be working on finalizing the requirements and then get back to this.

le-yams:
Hi @cikasfm,

For your information, here are the assumptions I made starting this PR:
On the delete, if we get an error it means the tuple does not exist thus nothing is actually deleted. That's why I didn't add it to the changelog (I assumed the changelog should reflect what is actually done). I haven't check if the raised error fails the transaction or not, so it indeed is to be verified (maybe it depends on the storage implementation). I'll try to add a test that verifies it (disclaimer: I won't be able to work on it on the next couple of weeks).

On the upsert, it will not raise any error and thus will be included in the changelog like any other insert. 

Hope it helps 🙏

tylernix:
Thank you @le-yams. Your PR gave us the push to finally implement this frequently requested feature. We started with your PR, but decided to make some different design considerations for simplicity. Rather than do a traditional “idempotent” operation, we chose to simply “ignore” 100% identical, duplicate tuples. We felt this was simpler, but still achieved the desired result.

1. Therefore, we don’t include “upsert” support for tuples with conditions, only "ignore" for exact duplicates. 
2. We added two new parameters in the writes and deletes sections of the Write request. Writes: `on_duplicate`, Deletes: `on_missing`
3. We lock the existing rows for deletion & use that as check for duplicate writes
4. SQL statement execution are now batched, instead of making a DB call per row
5. We validate the number of rows affected to exactly match the requested change to ensure no race condition could have affected the write or delete operation (i.e. write a tuple we don’t think exists but it suddenly exists, probably due to a parallel request). In this case, we will abort the race condition immediately and just return a 409 Conflict error.

We want to thank you for the effort you put into this PR, and we will give you a shout-out in the Release notes.

You can review more details of the design https://github.com/openfga/roadmap/issues/79 and the new PR https://github.com/openfga/openfga/pull/2663

yann-disanto:
Hi @tylernix,

Thank you (and the whole team) very much for moving forward on this, this is a great improvement for us 🙂.

If I understand well, the main difference (regarding final state) with the idempotency option is regarding the write of an existing tuple with a difference in the condition, which will generate a conflict error in case of a different condition values?

This also is a use case we have and that's why I was proposing an idempotency option (there is an issue #2151 opened a few months ago that was asking the same thing). Is it a feature that could be considered (update the condition of an existing tuple)?

In any case, thanks again for the work you've done on this, I was happy to help 🙏.
 
Regards, 
Yann

[edit] 
I just realized I was logged in with my professional account and not my personnal one (@le-yams) which is the one I made the PR with 😄. 
[/edit]

tylernix:
Yes. It was a hard decision for us. We discussed including this functionality, and intentionally left the door open for a `on_duplicate: upsert` option to solve this use case, but to reduce scope creep based on the timeline given, we decided to not include it. We see this (true idempotency for tuples with conditions) as a worthy next iteration and something we haven't forgotten. 

The issue https://github.com/openfga/openfga/issues/2151 you referenced mentions another pain point that we want to fix as well: deleting+writing the same tuple but with a different condition in the same transactional request. This is not possible today, and something we want to fix.

PR #2594, opened 2025-07-30
fix: make weighted reverse expand less aggressive in deduping

justincoh:
In the new reverse expand implementation, we have a check to ensure that we don't enter an infinite loop when traversing the graph. The logic in that check is basically `If we're going from Node A --> Node B, and we have gone over this exact edge before, don't do it again`. However, there are models for which we **have to** traverse an edge more than once, for example:

```			
type user
type thing
	relations
		define resource: [resource]
		define can_view: can_view from parent or admin from resource
		define parent: [document]
type document
	relations
		define resource: [resource]
		define can_view: admin from resource
type resource
	relations
		define owner: [user] and also_user
		define admin: ([user] and also_user) or owner
		define also_user: [user]
```

In the model above, `thing#can_view` begins two traversal paths, both of which can eventually end up at `resource#owner`.

1. `thing#can_view` --> `document#can_view` --> `resource#admin` --> `resource#owner`
2. `thing#can_view` --> `resource#admin` --> `resource#owner`

The existing dedupe logic would see that we are going from `resource#admin --> resource#owner` two different times, and whichever happens second will be prevented. Depending on which path finished first, ReverseExpand could return different results.

We need both of these traversals to complete to ensure we fully investigate all paths.

#### How is it being solved?
We added information about previous steps in the graph traversal (the `relationStack`) to the unique key we use to dedupe.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of duplicate edge traversals in the experimental ReverseExpand feature to prevent infinite loops in complex permission graphs.

* **New Features**
  * Added a function to display the contents of a stack in string format.

* **Tests**
  * Introduced comprehensive tests for the new stack string function.
  * Added new test cases covering advanced scenarios in the ReverseExpand feature, ensuring correct behavior with multiple and overlapping relationships.

* **Documentation**
  * Updated changelog to reflect the latest bug fix.

### DIFF
--- a/internal/stack/stack_test.go
+++ b/internal/stack/stack_test.go
@@ -67,6 +67,6 @@ func TestStack(t *testing.T) {
 		require.Equal(t, "hello", String(s))
 
 		_, s = Pop(s)
-		require.Equal(t, "", String(s))
+		require.Empty(t, String(s))
 	})
 }


Vic-Dev (approved):
coderabbit is confused 🤣

PR #2605, opened 2025-08-01
fix: flaky unit test for should_recover_from_panic

adriantam:
Flaky unit test for should_recover_from_panic due to incorrect wg order (defer for wg needs to be called before recover).

Flaky unit test TestShadowResolver_ResolveCheck/should_recover_from_panic due to incorrect order (defer wg.Done() needs to be set before setting the log because defer are executed in LIFO order).  Otherwise, the unit test could be flaky in that s.wg.Done() being called before the log is written.

#### How is it being solved?
Correct order of defer so that wg.Done() will be called last (by have this defer written ahead of defer panic recover).

1. Update shadow_resolver.go's  ResolveCheck to have the right order.
2. Break up `should_recover_from_panic` into two separate tests (`should_recover_from_panic` and `output_difference`) as they are testing two separate things.

## References
Flaky test run https://github.com/openfga/openfga/actions/runs/16682737971/job/47225020920?pr=2604

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Added a new test case to verify behavior when there is a difference between main and shadow resolver outputs.
* **Refactor**
  * Updated internal logic to improve the timing of certain background operations.


internal/graph/shadow_resolver.go
```diff
@@ -80,7 +82,6 @@ func (s ShadowResolver) ResolveCheck(ctx context.Context, req *ResolveCheckReque
 				}
 			}()
 
-			defer s.wg.Done()
 			ctx, cancel := context.WithTimeout(ctxClone, s.shadowTimeout)
 			defer cancel()
```
Vic-Dev, line 86: Do we want the cancel above the recover as well?

adriantam: No, we want the cancel to be inside the recover (in case the cancel causing panic)

senojj (approved):

PR #2608, opened 2025-08-04
fix: support tuple iterator cache in integration tests

senojj:
Currently, the tuple iterator cache TTL is set to 1 nanosecond when running integration tests. This is necessary because 3 tests cannot currently work with the cache in play:
- ttu_discard_invalid
- userset_discard_invalid
- userset_discard_invalid_wildcard

However, this effectively disables the cache for testing purposes and adds the potential for these tests to be flaky. These tests need not be designed in such a way.
#### How is it being solved?
The tests are changed so that their subsequent stage assertions do not depend upon or conflict with the tuples of the previous stages.

The three tests are updated to use different user and object identifiers from previous stages.
The tuple iterator cache TTL limitation is removed, allowing the default value to be used.

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Updated test cases with new object and user identifiers, and added new tuples to reflect updated user-role relationships.
  * Adjusted test data in permission checks to use the new identifiers and relationships.
  * Removed explicit cache TTL configuration from test setup, now using default cache settings.

### DIFF
--- a/assets/tests/consolidated_1_1_tests.yaml
+++ b/assets/tests/consolidated_1_1_tests.yaml
@@ -10164,43 +10164,48 @@ tests:
               define can_read: [role#assignee]
         tuples:
           - user: user:1
-            relation: assignee
-            object: role:admin
+            relation: placeholder
+            object: role:awesome
 
           # This tuple will become obsolete after the model change in the next stage.
           - object: role:invalid
             user: role:awesome#placeholder
             relation: assignee
+
+          - object: job:1
+            user: role:invalid#assignee
+            relation: can_read
         checkAssertions:
           - tuple:
               user: user:1
               relation: can_read
               object: job:1
-            expectation: false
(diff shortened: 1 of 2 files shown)


adriantam (commented):
The original intend of the tests was to assert whether check returns valid value when the model is changed.  I agree that we should remove the small wait as it does not produce stable value. However, since these tests become invalid due to introduction of caches, we should choose either of the two routes:
- remove the cache in these tests
- remove these "discard" tests.

senojj:
I've changed the tests to make them useful. Their original state when I approached them was not entirely useful because they didn't actually assert that the obsolete tuple was ignored. This is because there were no relationships to the obsolete tuple. I've addressed this issue and have included assertions that confirm that the obsolete tuple is ignored.

adriantam (approved):

PR #2624, opened 2025-08-08
Query planner

elbuo8:
## Review Checklist


internal/planner/planner.go
```diff
@@ -0,0 +1,77 @@
...
+			minSampledTime = sampledTime
+			resolver = name
+		}
+	}
+
+	return resolver
+}
+
+// UpdateStats performs the Bayesian update for the given resolver.
+func (p *Planner) UpdateStats(key, resolver string, duration time.Duration) {
```
senojj: an alternative approach might be to expose `Planner.GetKeyPlan` and put `UpdateStats` as a method on the `KeyPlan`. This would turn the call into:
```go
p.GetKeyPlan(key).UpdateStats(resolver, duration)
```
this also enables clients of a planner to take multiple actions on a key plan without needing to fetch out of the planner map again.

elbuo8: Less map access, sold.

internal/planner/thompson.go
```diff
@@ -0,0 +1,68 @@
...
+	ts.Alpha = newAlpha
+	ts.Beta = newBeta
+	ts.Runs++
+}
+
+// NewThompsonStats creates a new stats object with a diffuse prior,
+// representing our initial uncertainty about a strategy's performance.
+func NewThompsonStats() *ThompsonStats {
+	return &ThompsonStats{
+		Mu:     7.0,  // Initial guess for mean execution time (in ms)
```
yissellokta: For clarity, can you add a comment that this is the expectation for each algorithm execution not the problem resolution

internal/planner/planner.go
```diff
@@ -0,0 +1,75 @@
+package planner
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+)
+
+type Planner struct {
+	// ttu|userset_storeID_objectType_relation_userType
```
yissellokta: in the case of ttu the key is built differently. To avoid confusion, can you add as a comment the two patterns

yissellokta (dismissed):

internal/planner/planner.go
```diff
@@ -0,0 +1,76 @@
...
+		// and the subsequent update are consistent for a given key.
+		sampledTime := ts.Sample(kp.rng)
+		if resolver == "" || sampledTime < minSampledTime {
+			minSampledTime = sampledTime
+			resolver = name
+		}
+	}
+
+	return resolver
+}
```
senojj:
```suggestion
func (kp *KeyPlan) SelectResolver(resolvers []string) string {
	kp.mu.Lock()
	defer kp.mu.Unlock()

	resolver := ""
	var minSampledTime float64 = -1

    // Sample from each resolver's distribution and pick the one with the best (lowest) sample.
	for _, name := range resolvers {
	    // Ensure stats exist for this key for all resolvers.
		if _, ok := kp.stats[name]; !ok {
			kp.stats[name] = NewThompsonStats()
		}
		ts := kp.stats[name]
		// We use the global rng, but the lock ensures that the read of the stats
		// and the subsequent update are consistent for a given key.
		sampledTime := ts.Sample(kp.rng)
		if resolver == "" || sampledTime < minSampledTime {
			minSampledTime = sampledTime
			resolver = name
		}
	}

	return resolver
}
```

internal/planner/thompson.go
```diff
@@ -0,0 +1,151 @@
...
+	newMu := (currentLambda*currentMu + x) / newLambda
+	newAlpha := currentAlpha + 0.5
+	diff := x - currentMu
+	newBeta := currentBeta + (currentLambda*diff*diff)/(2*newLambda)
+
+	// Store new values atomically
+	ts.mu.Store(newMu)
+	ts.lambda.Store(newLambda)
+	ts.alpha.Store(newAlpha)
+	ts.beta.Store(newBeta)
```
senojj: since these values are not set as an atomic unit, there a possibility for two parallel threads of execution to observe a partially updated state when invoking `Update`... What is the potential impact of that?

senojj: Note that currently, `Update` is only called from within a `KeyPlan` `UpdateStats` method, which holds an exclusive lock for the duration. Therefore, this likely isn't an issue in the current implementation, but if someone were to alter the code without taking this into account, it could be troublesome.

internal/graph/check.go
```diff
@@ -696,6 +707,18 @@ func shouldCheckPublicAssignable(ctx context.Context, reqTupleKey *openfgav1.Tup
 	return isPubliclyAssignable
 }
 
+func profiledCheckHandler(keyPlan *planner.KeyPlan, resolverName string, resolver CheckHandlerFunc) CheckHandlerFunc {
+	return func(ctx context.Context) (*ResolveCheckResponse, error) {
+		start := time.Now()
+		res, err := resolver(ctx)
+		if err != nil {
+			return nil, err
+		}
```
willvedd, line 716: Might there be certain errors (ex: timeouts) where you'd want to update stats?

internal/planner/planner.go
```diff
@@ -0,0 +1,127 @@
...
+// selectResolverLocked assumes mu is already held (either read or write lock).
+func (kp *KeyPlan) selectResolverLocked(resolvers []string) string {
+	// Get RNG from pool
+	rng := kp.planner.rngPool.Get().(*rand.Rand)
+	defer kp.planner.rngPool.Put(rng)
+
+	resolver := ""
+	var minSampledTime float64 = -1
+
+	// Sample from each resolver's distribution and pick the one with the best (lowest) sample.
```
adriantam: If I understand correctly, this will always pick the "best" sample...do we want to leak certain percentage of calls to use the non-best samples so that we can continue to build up samples over time?  Otherwise, there is a possibility that if the "best" changes over time (due to added / removed tuples), we will pick the stale "best" method.

adriantam: Or we rely on the Thomspson sample to give us a randomized strategy over time.

internal/graph/check.go
elbuo8: I had thought about this in my original pass, and implemented it for a second here. The problem is, the deadline could be at the tail end of the resolution of the end of a different resolution for the plan.
 
Since the timeout/deadline budget is being consumed by multiple resolutions I would prefer to only store concrete stats rather than partial errors that might influence negatively.

internal/planner/thompson.go
elbuo8: Added a comment.

internal/planner/planner.go
yissellokta: After converging to a strategy, when changing the data distribution, there will most probably be a change in latency, and this will trigger the sampling to explore other solutions again until it converge to a new one or keeps the existing one

elbuo8: No, it won't _always_ pick the best. It will _favor_ the best. But will continue exploring the other strategies every so often in order to identify if the "best" has actually changed over time.

internal/planner/thompson.go
elbuo8: Actually cleaned this up.

yissellokta (approved):

internal/planner/planner.go
```diff
@@ -0,0 +1,91 @@
...
+			bestResolver = name
+		}
+	}
+
+	return bestResolver
+}
+
+// UpdateStats performs the Bayesian update for the given resolver's statistics.
+func (kp *KeyPlan) UpdateStats(resolver string, duration time.Duration) {
+	ts, _ := kp.stats.LoadOrStore(resolver, NewThompsonStats(kp.initialGuess))
```
senojj: Additional thought; non-gating:

The `UpdateStats` function might be a hot one. Creating a new `*ThompsonStats` value on every invocation could be expensive (would need to benchmark). The value likely escapes to the heap, creating additional GC pressure, especially for values that are immediately discarded.

A lazy-loaded option:

```go
v, _ := kp.stats.LoadOrStore(resolver, unsafe.Pointer(nil))
p := v.(unsafe.Pointer)

if p == nil {
    atomic.CompareAndSwapPointer(&p, nil, unsafe.Pointer(NewThompsonStats(kp.initialGuess)))
}

ts := (*ThompsonStats)(atomic.LoadPointer(&p))
```
Unsure if that is "better", but might be worth testing.

internal/planner/thompson.go
senojj: Looks good.

internal/planner/planner.go
elbuo8: Why would it be immediately discarded?

senojj: When `kp.stats.LoadOrStore(resolver, NewThompsonStats(kp.initialGuess))` loads an existing item, the `*ThompsonStats` pointer that was produced by `NewThompsonStats()` will be discarded by the `LoadOrStore` function, immediately. The value on the heap will not be reachable during the next GC mark & sweep operation, and thus will be reclaimed.

elbuo8: Ah yes, not great. Will check.

senojj: Good solution.

senojj:
The cleaning up of values that haven't been used recently is a great addition.

However, deleting the unused items from the map does not reduce the map's capacity (that memory is not reclaimed AFAIK). Therefore, if the map's size grows very large during a period of high throughput, it's capacity will remain that size for the duration of the service's lifetime.

elbuo8:
I'm not too concerned about that becoming a problem. The memory will be used by new entries and while we won't reclaim the map size itself, its marginal in comparison to the values it stores. If it ever comes a problem we can create a new map and copy the none stale values, but at the current time I find that to be overkill.

senojj (approved):

cmd/run/flags.go
adriantam, line 1: Do we need to update https://github.com/openfga/openfga/blob/main/.config-schema.json?

internal/graph/recursive_resolver.go
```diff
@@ -19,13 +19,15 @@ import (
...
+const recursiveResolver = "recursive"
+
 type recursiveMapping struct {
 	kind                        storage.TupleMapperKind
 	tuplesetRelation            string
 	allowedUserTypeRestrictions []*openfgav1.RelationReference
 }
 
-func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, rightIter storage.TupleKeyIterator) CheckHandlerFunc {
+func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, _ []*openfgav1.RelationReference, rightIter storage.TupleKeyIterator) CheckHandlerFunc {
```
adriantam, line 30: I notice that for both defaultUserset, weight2Userset, and recursiveUserset - we don't use the first parameter (context.Context).  Is it used for any of the userset resolver?  If not, perhaps we can remove the parameter?

internal/planner/thompson.go
```diff
@@ -0,0 +1,133 @@
...
+	"sync/atomic"
+	"time"
+	"unsafe"
+
+	"gonum.org/v1/gonum/stat/distuv"
+)
+
+// ThompsonStats holds the parameters for the Normal-gamma distribution,
+// which models our belief about the performance (execution time) of a strategy.
+type ThompsonStats struct {
```
adriantam, line 15: is ThompsonStats exported outside of internal/planner?  If not, perhaps make this private `thompsonStats`?

internal/planner/planner.go
adriantam, line 1: will be good to call out as TODO that a future improvement is to bound the number of items being tracked

adriantam (commented):
A feature like this should be noted in changelog

internal/graph/recursive_resolver.go
elbuo8: In another PR, given this is the whole pattern of usage of several other calls. I particularly don't like it either and I actually want to decouple the strategies of the LocalChecker itself. But again, that will be another PR.

internal/planner/thompson.go
elbuo8: Not at the current time, but makes no difference, for now will keep as is.

internal/planner/planner.go
elbuo8: Added.

cmd/run/flags.go
elbuo8: 🤦🏽 why do we even have this. Any flag/config value requires a minimum of 5 files to be added, this really needs to change.

adriantam: Agree

PR #2642, opened 2025-08-15
chore: refactor userset

elbuo8:
## Description

When handling usersets, each assignable userset must be handled in its own instance of the weight2 resolver. If not all possible usersets are handled via weight2, then the remaining ones should be handled by the default resolver.

If the tuple requested is a userset, it will use the default handler.

If the object relation can be resolved by recursion, it will just use the default handler.

## Review Checklist

## Summary by CodeRabbit

- New Features
  - Optimized reading of userset relationship data with iterator-based processing and improved filtering.

- Refactor
  - Centralized and streamlined userset resolution logic with deferred execution handlers for better performance and maintainability.
  - Simplified recursive and weight-based resolvers, improving concurrency control and reducing overhead.

- Performance
  - Faster and more efficient access checks for userset-related queries, especially in weight-2 scenarios.

- Tests
  - Updated tests to reflect new resolver patterns and added coverage for multi-userset evaluation.

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -703,7 +703,6 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		ctx, span := tracer.Start(ctx, "checkDirectUsersetTuples", trace.WithAttributes(
 			attribute.String("userset", tuple.ToObjectRelationString(reqTupleKey.GetObject(), reqTupleKey.GetRelation())),
-			attribute.String("resolver", "slow"),
 		))
 		defer span.End()
 
@@ -720,7 +719,7 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
 			}
 			defer iter.Stop()
 
-			return union(ctx, c.concurrencyLimit, c.defaultUserset(ctx, req, iter))
+			return c.defaultUserset(ctx, req, iter)(ctx)
 		}
 
 		// if the type#relation is resolvable recursively, then it can only be resolved recursively
@@ -731,7 +730,7 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
 			}
 			defer iter.Stop()
 
-			return union(ctx, c.concurrencyLimit, c.recursiveUserset(ctx, req, iter))
+			return c.recursiveUserset(ctx, req, iter)(ctx)
 		}
 
 		var resolvers []CheckHandlerFunc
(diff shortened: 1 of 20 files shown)


internal/graph/check.go
```diff
@@ -696,6 +696,88 @@ func shouldCheckPublicAssignable(ctx context.Context, reqTupleKey *openfgav1.Tup
...
+				defer iter.Stop()
+				resolvers = append(resolvers, c.defaultUserset(ctx, req, iter))
+			}
+		} else {
+			iter, err := checkutil.IteratorReadUsersetTuples(ctx, req, directlyRelatedUsersetTypes)
+			if err != nil {
+				return nil, err
+			}
+			defer iter.Stop()
+			if len(directlyRelatedUsersetTypes) == 0 && typesys.UsersetUseWeight2Resolver(objectType, relation, userType, directlyRelatedUsersetTypes[0]) {
```
adriantam: See https://github.com/openfga/openfga/pull/2645 - w2 does not apply only in the case of >1 usersets  and their object types are the same.  If they are of different type, w2 optimization still applies.

adriantam: Please see https://github.com/openfga/openfga/pull/2650 for proposed changes

internal/graph/check.go
```diff
@@ -696,6 +696,88 @@ func shouldCheckPublicAssignable(ctx context.Context, reqTupleKey *openfgav1.Tup
...
+		if c.optimizationsEnabled {
+			var remainingUsersetTypes []*openfgav1.RelationReference
+			for _, userset := range directlyRelatedUsersetTypes {
+				if typesys.UsersetUseWeight2Resolver(objectType, relation, userType, userset) {
+					usersets := []*openfgav1.RelationReference{userset}
+					iter, err := checkutil.IteratorReadUsersetTuples(ctx, req, usersets)
+					if err != nil {
+						return nil, err
+					}
+					defer iter.Stop()
```
adriantam, line 748: To confirm do we really want to defer iter.Stop() until the entire function is done or within the context of the for loop?  As per https://engineering.01cloud.com/2023/04/06/using-defer-inside-a-loop/, iter.Stop() will not be called until the entire function is done...so, if this is the actual intend, let's call this out in the comment to avoid confusion.

pkg/typesystem/typesystem_test.go
```diff
@@ -3819,7 +3819,7 @@ func TestUsersetCanFastPathWeight2(t *testing.T) {
 			objectType: "wallet",
 			relation:   "can_write",
 			userType:   "driver",
-			expected:   false,
+			expected:   true,
```
adriantam: Probably be good to call out that only account#owner will be true.  resource#can_write will be false...but since our test only check the presence of a single true, this will be true

internal/graph/check.go
elbuo8: For the entire function. Will add comment.

adriantam (approved):

PR #2661, opened 2025-08-26
fix collation to always use byte-wise ordering

senojj:
The core of the issue lies in a mismatch between how the database sorts strings and how the Go application logic compares them. The OrderedCombinedIterator, implemented in Go, implicitly relies on Go's default byte-wise string comparison for its merging logic. It assumes the data it receives from the database is sorted in this same lexicographical manner.

However, when the database uses a locale-aware collation (like `en_US.UTF-8`), it sorts strings according to language-specific rules, not simple byte order. This is a common scenario, as major managed database services like AWS RDS for PostgreSQL default to `en_US.UTF-8`. For example, with this collation, the string a-b might be sorted before a_b. This conflicts with Go's byte-wise comparison, where _ (underscore) comes after - (hyphen). This discrepancy violates the iterator's core assumption, leading to incorrectly merged results and ultimately, fatal errors such as "iterator 0 is not in ascending order". To avoid such errors, we must choose a collation that uses a lexicographical, byte-wise order consistent with Go's native string comparison.

#### How is it being solved?
This PR proposes a database-specific strategy to enforce lexicographical ordering on the object_id column, which is critical for the ReadStartingWithUser query.

1. For MySQL: A database migration will alter the tuple table to change the object_id column's collation to utf8mb4_bin. The migration will use LOCK=SHARED to permit concurrent reads during the operation.

2. For PostgreSQL: To avoid a blocking table rewrite, we will not alter the column itself. Instead, a database migration will create a new, non-blocking composite index covering the columns used by the ReadStartingWithUser query, with a specific binary collation applied to the object_id column. The query will be modified to use an ORDER BY object_id COLLATE "C" clause, which will allow the query planner to use the new index for efficient, correctly-ordered reads.

3. Simplify Application Logic: With both strategies guaranteeing lexicographical ordering from the database, all in-memory comparison logic within the OrderedCombinedIterator will use Go's native, byte-wise string comparison.

This hybrid approach fixes the problem at its root—the data layer—while choosing the most efficient and least disruptive path for each supported database.

The implementation will involve the following steps:

1. Add New Database Migration Scripts:

    - New, separate migration scripts will be added to handle this schema change. These scripts will run for both new installations and existing databases, ensuring all environments are consistent.

    - MySQL Migration: The script will contain ALTER TABLE tuple LOCK=SHARED, MODIFY COLUMN object_id VARCHAR(256) COLLATE utf8mb4_bin;. The LOCK=SHARED clause allows concurrent reads while the table is being modified, but blocks writes.

    - PostgreSQL Migration: The script will contain CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_rswu_c ON tuple (_user, relation, object_type, object_id COLLATE "C");. The CONCURRENTLY keyword ensures the operation is non-blocking.

    - SQLite: No changes are needed.

2. Update PostgreSQL Queries:

    - The ReadStartingWithUser query in the PostgreSQL storage backend will be modified to include the ORDER BY object_id COLLATE "C" clause to match the index.

3. Ensure Native String Comparison in Go:

    - Confirm that all relevant in-memory sorting and comparison logic uses standard Go string comparison.

Note that the ULID implementation employed by OpenFGA only uses numeric and capital letter characters, so identifier fields are not subject to ordering variations.

## References

Discussion: https://github.com/orgs/openfga/discussions/491

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures deterministic, binary string ordering across databases for list/read/change operations, producing consistent sorting of IDs and keys.
  * Stabilizes pagination and continuation tokens to prevent item reordering across pages and feeds.
  * Uniform ordering behavior across MySQL and Postgres-backed deployments with no API changes.

* **Documentation**
  * Added changelog entry noting collation-related fixes.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -67,7 +67,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@f1f6e5f6af878fb37288ce1c627459e94dbf7d01
+      uses: github/codeql-action/init@192325c86100d080feab897ff886c34abd4c83a3
       with:
         languages: ${{ matrix.language }}
         build-mode: ${{ matrix.build-mode }}
@@ -95,6 +95,6 @@ jobs:
         exit 1
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@f1f6e5f6af878fb37288ce1c627459e94dbf7d01
+      uses: github/codeql-action/analyze@192325c86100d080feab897ff886c34abd4c83a3
       with:
         category: "/language:${{matrix.language}}"
--- a/.github/workflows/enforce-changelog-entry.yaml
+++ b/.github/workflows/enforce-changelog-entry.yaml
@@ -15,7 +15,7 @@ jobs:
         with:
           skipLabels: 'Skip-Changelog,dependencies,tests'
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v3.5.2
-      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # pin@v4
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # pin@v4
       - run: npx keep-a-changelog@2.5.3
       - run: |
           files=$(git status --porcelain)
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v3.5.2
 
-      - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+      - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
         with:
           go-version-file: './go.mod'
           cache-dependency-path: './go.sum'
@@ -23,7 +23,7 @@ jobs:
         run: make test
 
       - name: upload coverage to codecov
-        uses: codecov/codecov-action@fdcc8476540edceab3de004e990f80d881c6cc00 # v5.5.0
+        uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
         with:
           files: ./coverageunit.out
           verbose: true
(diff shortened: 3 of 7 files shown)


senojj:
@coderabbitai review

assets/migrations/mysql/007_add_collate_index.sql
```diff
@@ -0,0 +1,15 @@
...
+    object_type,
+    (object_id COLLATE utf8mb4_bin)
+);
+
+DROP INDEX idx_reverse_lookup_user ON tuple;
+
+-- +goose Down
+DROP INDEX idx_collate on tuple;
+
+CREATE INDEX idx_reverse_lookup_user ON tuple (store, object_type, relation, _user);
```
sergiught:
```suggestion
CREATE INDEX idx_reverse_lookup_user ON tuple (store, object_type, relation, _user);

```

nit: please add a trailing newline

assets/migrations/postgres/006_add_collate_index.sql
```diff
@@ -0,0 +1,16 @@
...
+    object_type,
+    object_id collate "C"
+);
+
+DROP INDEX IF EXISTS idx_reverse_lookup_user;
+
+-- +goose Down
+DROP INDEX IF EXISTS idx_collate;
+
+CREATE INDEX idx_reverse_lookup_user on tuple (store, object_type, relation, _user);
```
sergiught:
```suggestion
CREATE INDEX idx_reverse_lookup_user on tuple (store, object_type, relation, _user);

```
nit: please add a trailing newline

sergiught (dismissed):

assets/migrations/mysql/007_collate_object_id.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+ALTER TABLE tuple MODIFY COLUMN object_id VARCHAR(255) COLLATE utf8mb4_bin,
+LOCK = SHARED;
+
+CREATE INDEX idx_user_lookup ON tuple (store, _user, relation, object_type, object_id);
+
+DROP INDEX idx_reverse_lookup_user ON tuple;
+
+-- +goose Down
+DROP INDEX idx_collate on tuple;
```
aaguiarz:
```suggestion
DROP INDEX idx_user_lookup on tuple;
```

assets/migrations/postgres/006_add_collate_index.sql
```diff
@@ -0,0 +1,16 @@
...
+-- +goose Up
+CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_user_lookup on tuple (
+    store,
+    _user,
+    relation,
+    object_type,
+    object_id collate "C"
+);
+
+DROP INDEX IF EXISTS idx_reverse_lookup_user;
```
elbuo8: Ah this is safe to be deleted because its using the PK rather than this index at all. Perhaps worth adding that comment 🤔

senojj: That index is over `(store, object_type, relation, _user)`, and the QP was preferring the PK in the explain plans that I observed. However, are there any queries that might use this index instead of the new one `(store, _user, relation, object_type, object_id)`?

elbuo8: Afaik, no since it was created exclusively for RSWU (as the name kinda suggests).

adriantam:
Should we add unit test to verify all storage interface satisfy the behaviour?

PR #2662, opened 2025-08-27
feat: duplicate writes (sqlite)

cikasfm:
Implements idempotent write and delete functionality in the SQLite datastore, leveraging a read-before-write pattern to ensure atomicity and consistency.

## Description

This pull request implements the idempotent write and delete functionality for the SQLite datastore. It introduces a new `write` method that uses a robust read-before-write pattern within a single transaction to handle idempotent operations and prevent race conditions.

This PR addresses the issue where batched `Write` requests fail entirely if any single tuple already exists. By implementing the `on_duplicate: ignore` and `on_missing: ignore` behaviors, this change significantly improves the developer experience and simplifies client-side logic for data synchronization and retries.

#### How is it being solved?
The solution is implemented entirely within a single SQLite transaction. Instead of a series of individual `DELETE` and `INSERT` statements, the new logic performs a single `SELECT ... FOR UPDATE` at the beginning of the transaction to lock all relevant rows. This allows the system to determine which operations are no-ops and then execute all remaining `DELETE` and `INSERT` statements in two efficient, batched operations.

- A new `write` method is introduced to handle the core transaction logic.
- The `Write` API method is refactored to use a new `TupleWriteOptions` struct, which carries the `on_duplicate` and `on_missing` parameters.
- The new `write` method includes a `SELECT ... FOR UPDATE` statement that locks all tuples mentioned in the request.
- The logic for processing deletes and writes is updated to use the results from the `SELECT` query, allowing for no-op behavior when `ignore` flags are present.
- Batched `INSERT` and `DELETE` statements are used to replace individual statements, drastically reducing database round trips and improving performance.
- The `busyRetry` mechanism is applied to the new batched operations to handle potential contention.
- New test cases are added to validate the new idempotent behavior.

## References

* Required for https://github.com/openfga/roadmap/issues/79

## Review Checklist

### DIFF
--- a/pkg/storage/sqlite/sqlite.go
+++ b/pkg/storage/sqlite/sqlite.go
@@ -255,8 +255,8 @@ func (s *Datastore) write(
 	// select columns required for the SQLTupleIterator
 	selectBuilder := s.stbl.
 		Select(tupleColumns...).
+		Where(sq.Eq{"store": store}).
 		From("tuple")
-	// TODO: Clarify if we need to use FOR UPDATE here as SQLite does not support select * FOR UPDATE
 
 	// Assume deletes is a slice of tuple keys (e.g., storage.Deletes)
 	var orConditions []sq.Sqlizer
@@ -265,7 +265,6 @@ func (s *Datastore) write(
 		objectType, objectID := tupleUtils.SplitObject(tk.GetObject())
 		userObjectType, userObjectID, userRelation := tupleUtils.ToUserParts(tk.GetUser())
 		orConditions = append(orConditions, sq.Eq{
-			"store":            store,
 			"object_type":      objectType,
 			"object_id":        objectID,
 			"relation":         tk.GetRelation(),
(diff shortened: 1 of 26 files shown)


pkg/storage/sqlite/sqlite.go
```diff
@@ -35,6 +35,12 @@ func startTrace(ctx context.Context, name string) (context.Context, trace.Span)
 	return tracer.Start(ctx, "sqlite."+name)
 }
 
+var tupleColumns = []string{
+	"store", "object_type", "object_id", "relation",
+	"user_object_type", "user_object_id", "user_relation", /* TODO: "user_type"?, what is the difference */
```
DanCech, line 40: user_type is extracted from user_relation, and we don't select or scan it here because TupleRecord doesn't have a field for it. It is used in postgres and sqlite (but not mysql) to create partial indexes and (hopefully) speed up selecting from the tuple table.

PR #2663, opened 2025-08-27
feat: duplicate writes (SQL only)

cikasfm:
### Summary

This pull request implements the idempotent write and delete functionality for the SQLite datastore. It introduces a new `write` method that uses a robust read-before-write pattern within a single transaction to ensure atomicity and consistency for idempotent operations.

-----

### Description

This pull request implements the idempotent write and delete functionality for the SQLite datastore, as outlined in the related https://github.com/openfga/roadmap/issues/79. This feature addresses the common developer pain point where batched `Write` requests fail if any tuple already exists, thereby significantly improving the developer experience and simplifying client-side logic for data synchronization and retries.

This PR solves the problem of unreliable batched `Write` operations caused by the current behavior of failing an entire request on a single duplicate tuple.

#### How is it being solved?

The solution is implemented entirely within a **single SQLite transaction**. It uses a read-before-write pattern, performing a single `SELECT ... FOR UPDATE` at the beginning of the transaction. This allows the system to validate which operations are no-ops based on the `on_duplicate` and `on_missing` flags, and then execute the remaining `DELETE` and `INSERT` statements in two efficient, batched operations.

  - A new `write` method is introduced to handle the core transaction logic.
  - The main `Write` method is refactored to pass the new `TupleWriteOptions` struct, which carries the `on_duplicate` and `on_missing` parameters.
  - The new `write` method includes a `SELECT ... FOR UPDATE` statement to lock all tuples mentioned in the request.
  - The logic for processing deletes and writes is updated to use the results from the `SELECT` query, allowing for no-op behavior.
  - Batched `INSERT` and `DELETE` statements are used to replace individual statements, drastically reducing database round trips and improving performance.
  - The `busyRetry` mechanism is applied to the new batched operations to handle potential contention.
  - New test cases are added to validate the new idempotent behavior.

### Scope

In Scope:
 - Datastore API changes
 - SQL driver changes:
 * PostgreSQL, MySQL
 * SQLite changes ( https://github.com/openfga/openfga/pull/2662 )
 - InMemory DB changes

Out of scope:
 - API Changes ( adding the actual API endpoint parameters ) - related PR https://github.com/openfga/api/pull/233

-----

### References

  * **Closes:** https://github.com/openfga/roadmap/issues/79
  * **Related to:** https://github.com/openfga/openfga/pull/2593

-----

### Review Checklist

  - [ ] I have clicked on [["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
  - [ ] I have added documentation for new/changed functionality in this PR or in a PR to [[openfga.dev](https://github.com/openfga/openfga.dev)](https://github.com/openfga/openfga.dev)

## Summary by CodeRabbit

- Breaking Changes
  - Write interface now requires a write-options parameter, affecting storage backends and implementers.
- New Features
  - Per-call write options to control behavior for missing deletes and duplicate inserts (error or ignore).
  - Improved concurrency safety and deterministic locking for transactional writes across SQL and SQLite backends.
  - Enhanced in-memory write behavior with option-driven validation.
- Bug Fixes
  - Clearer, consistent conflict errors for insert/delete races and condition mismatches.
- Documentation
  - Changelog updated with breaking-change notice.
- Tests
  - Expanded coverage for conflict handling and option combinations.

### DIFF
--- a/pkg/storage/sqlite/sqlite.go
+++ b/pkg/storage/sqlite/sqlite.go
@@ -111,6 +111,11 @@ func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
 		return nil, fmt.Errorf("initialize sqlite connection: %w", err)
 	}
 
+	return NewWithDB(db, cfg)
+}
+
+// NewWithDB creates a new [Datastore] storage with the provided database connection.
+func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
 	var collector prometheus.Collector
 	if cfg.ExportMetrics {
 		collector = collectors.NewDBStatsCollector(db, "openfga")
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -704,7 +704,7 @@
                 "limit": {
                     "description": "if shared-iterator-enabled is enabled, this is the limit of the number of iterators that can be shared.",
                     "type": "integer",
-                    "default": "1000000",
+                    "default": 1000000,
                     "x-env-variable": "OPENFGA_SHARED_ITERATOR_LIMIT"
                 }
             }
@@ -715,6 +715,31 @@
             "format": "duration",
             "default": "3s",
             "x-env-variable": "OPENFGA_REQUEST_TIMEOUT"
+        },
+        "planner": {
+            "type": "object",
+            "properties": {
+                "initialGuess": {
+                    "description": "The initial guess for the planners estimation.",
+                    "type": "string",
+                    "format": "duration",
+                    "default": "10ms",
+                    "x-env-variable": "OPENFGA_PLANNER_INITIAL_GUESS"
+                },
+                "evictionThreshold": {
+                    "description": "How long a planner key can be unused before being evicted.",
+                    "format": "duration",
+                    "default": "0",
+                    "x-env-variable": "OPENFGA_PLANNER_EVICTION_THRESHOLD"
+                },
+                "cleanupInterval": {
+                    "description": "How often the planner checks for stale keys.",
+                    "type": "string",
+                    "format": "duration",
+                    "default": "0",
+                    "x-env-variable": "OPENFGA_PLANNER_CLEANUP_INTERVAL"
+                }
+            }
         }
     },
     "definitions": {
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,100 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "main" ]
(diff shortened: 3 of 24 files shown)


pkg/storage/sqlite/sqlite.go
```diff
@@ -341,31 +404,42 @@ func (s *Datastore) write(
...
+				// If tuple conditions are different, we throw an error.
+				return storage.TupleConditionConflictError(tk)
+			case storage.OnDuplicateInsertError:
+				fallthrough
+			default:
+				return storage.InvalidWriteInputError(
+					tk,
+					openfgav1.TupleOperation_TUPLE_OPERATION_WRITE,
+				)
+			}
```
DanCech, line 542: Thinking more about this, we're effectively treating the combination `store, object_type, object_id, relation, user_object_type, user_object_id, user_relation` as a unique key (I'm excluding `user_type` since it's derived from `user_relation`), but we have not actually defined it as a unique key in the database itself which means that we don't have any automatic protection against inserting duplicate records and are enforcing uniqueness only via the code here in the application layer.

DanCech: Nevermind, I was glossing over the primary key

pkg/server/test/list_users.go
```diff
@@ -180,6 +180,8 @@ func BenchmarkListUsers(b *testing.B, ds storage.OpenFGADatastore) {
 		bm.inputRequest.StoreId = storeID
 		ctx = typesystem.ContextWithTypesystem(ctx, typeSystem)
 
+		b.ResetTimer()
```
senojj: unnecessary

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -474,107 +494,301 @@ func NewDBInfo(db *sql.DB, stbl sq.StatementBuilderType, errorHandler errorHandl
...
+		return nil
+	}
+
+	existing := make(map[string]*openfgav1.Tuple, len(lockKeys))
+
+	// 3. If list compiled in step 2 is not empty, execute SELECT … FOR UPDATE statement
+
+	// Efficient, deterministic locking: probe existing rows using a row-constructor IN on the composite key.
+	// Batch to keep SQL size reasonable.
+	const lockBatchSize = 128
```
DanCech, line 612: I noticed that while we batch here, we are not batching in the delete, insert or changelog builders. Do you think we should look at batching there too?

adriantam: The actual execution of the builder is only once for multiple lines of insert/delete/changelog insertion (see line 807 and 827).

DanCech: Yes, that's what I mean when I say that they are not batched. Here we're limiting the lock select to a max of 128 but we do the insert/delete/changelog in one shot regardless of the number of changes.

cikasfm: Yes, that's correct. We should probably page everything.

For now we're limiting to 40 items per write, so this won't be applicable, but if we want to increase the num writes - that would make sense.

Can we create an issue & do it in a future PR?

adriantam (approved):

PR #2664, opened 2025-08-27
chore(docs): added caching  docs

aaguiarz:
## Description

Added internal documentation about how caching works for review. Afterwards, we should move this to the public docs.

There's no documentation about how caching works.

## Review Checklist

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive caching system documentation covering cache types, configuration options, cache invalidation strategies, observability, performance tuning, memory considerations, and best practices for deployment environments.

* **Chores**
  * Updated cache-related metric descriptions for improved clarity.


docs/throttling.md
```diff
@@ -0,0 +1,246 @@
...
+
+export OPENFGA_LIST_OBJECTS_DISPATCH_THROTTLING_ENABLED=true
+export OPENFGA_LIST_OBJECTS_DISPATCH_THROTTLING_THRESHOLD=50
+export OPENFGA_LIST_OBJECTS_DISPATCH_THROTTLING_MAX_THRESHOLD=150
+export OPENFGA_LIST_OBJECTS_DISPATCH_THROTTLING_FREQUENCY=100ms
+
+export OPENFGA_LIST_USERS_DISPATCH_THROTTLING_ENABLED=true
+export OPENFGA_LIST_USERS_DISPATCH_THROTTLING_THRESHOLD=50
+export OPENFGA_LIST_USERS_DISPATCH_THROTTLING_MAX_THRESHOLD=150
+export OPENFGA_LIST_USERS_DISPATCH_THROTTLING_FREQUENCY=100ms
```
cikasfm: The limits look quite low to me

- Frequency 100ms = max 10 dispatch per second.
- Frequency 50ms = max 20 dispatch per second.

I would reduce frequency to sub 20ms ( or at least no more than 50ms )

Also, thresholds are likely to be 10x larger.

Many OpenFGA users have requests with over 1000 dispatches, some number between 1000 and 10000 is more realistic.

docs/caching.md
```diff
@@ -0,0 +1,250 @@
+# Caching in OpenFGA
+
+OpenFGA provides a sophisticated multi-layered caching system designed to improve performance by reducing database queries and computation overhead. This document explains how caching works in OpenFGA and how to configure it for optimal performance in your deployment.
+
+## Overview
+
+OpenFGA implements several complementary types of caching:
+
+1. **Check Query Cache**: Caches results of Check API requests to avoid recomputing identical authorization checks
```
adriantam: It is not the caches result of check API request, but caches results of sub problem within check.

docs/caching.md
```diff
@@ -0,0 +1,250 @@
...
+2. **Check Iterator Cache**: Caches database query results (iterators) to reduce database load during Check operations
+3. **ListObjects Iterator Cache**: Caches database iterators specifically for ListObjects API calls
+4. **Cache Controller**: Manages cache invalidation based on data changes to maintain consistency. It works in background, and it's not involved in the query resolution. After each TTL period, it checks the changelog, and invalidates the cache if tuples were written.
```
adriantam: I wouldn't go into "not involved in the query resolution".  Instead, the idea is to cache the result of change request so that we don't need to ask Datastore for changes if data is already in the controller cache.

aaguiarz: @adriantam 
Can you elaborate? I don't understand what you mean. What data does the controller cache has? I thought the controller cache did not store any data.

adriantam: The cache controller has a [cache](https://github.com/openfga/openfga/blob/1e79eba0a65392fe0cc04dbd20f1ff2e193c3830/internal/cachecontroller/cache_controller.go#L211) where we store the recent read changes result.  Whenever the cache controller is called to check whether to invalidate the sub problem cache / iterator cache, it will first consult with the cache controller cache.  If the entry is there, it will use that entry's result to determine whether the subproblem/iterator cache is to be invalidated.  If the entry is not there, it will then call ReadChanges (which read from the datastore) to see whether the sub problem / iterator cache is to be invalidated.  The cache controller will then store the read changes result in the cache controller cache.

docs/caching.md
```diff
@@ -0,0 +1,181 @@
...
+
+#### First Check Staleness
+
+Because invalidation is triggered asynchronously by Check and List Objects requests, there is an accepted race condition where the first Check after a write (and subsequent Checks until the async job finishes) could return stale data if its Check response was cached before the write, even if it has been more than the cache controller TTL since the write.
+
+For example, if we look at the previous example, if Check 3 had occurred at t=100s, it would have still returned a stale result since it triggers invalidation asynchronously but returns immediately. The following Check however would see the results of the invalidation and compute a fresh result.
+
+Note that *any* Check request (if the cache controller TTL has passed since the last invalidation) or List Objects request (if list objects iterator cache is enabled) will trigger invalidation for the entire store, so this issue only occurs with very infrequent requests.
+
+#### Incompatibility with Distributed Cache / Eventually Consistent Database
```
aaguiarz: I would not mention this, as it's not relevant for OpenFGA users and adds confusion

docs/caching.md
```diff
@@ -0,0 +1,181 @@
...
+- **Benefits**: Reduces database load by caching some of its results in-memory.
+
+#### Example
+
+Consider the [Entitlements sample store](https://github.com/openfga/sample-stores/tree/79fa8c1710f12d3f0befaba77f15463bb5a97860/stores/entitlements) and OpenFGA running with check iterator cache enabled. Given the query "does user `user:charles` have relationship `can_access` with `feature:draft_prs`?", the following cache entries (simplified) are set:
+
+1. Key: What users have relationship `associated_plan` with object `feature:draft_prs`? Value: `[plan:enterprise, plan:team]`.
+2. Key: What users have relationship `subscriber` with object `plan:enterprise`? Value: `[organization:cups]`.
+3. Key: What are the tuples where user `user:charles` has relationship `member` with objects of type `organization`? Value: `[organization:cups#member@user:charles]`.
+
```
aaguiarz, line 83: We have a configurable limit on the number of records this cache can hold, right? Can we mention it?

docs/caching.md
```diff
@@ -0,0 +1,181 @@
...
+t=1s | Write: Tuple deleted | Store updated | Previous Check is now invalid |
+t=2s | Replica 2: Check | Returns `allowed: true` (stale) | Gets stale data from distributed cache so computes a stale result and caches it locally. | Triggered, but nothing to invalidate as the latest write was at t=1s and our cache entries were set later, at t=2s.
+t=12s | Replica 2: Check | " | Stale data in distributed cache has expired, but we cached stale data locally at t=2s and it wasn't invalidated nor has it expired so we use that to compute a stale result. | "
+t=300s | Replica 2: Check | " | " | "
+
+Note:
+- OpenFGA currently supports In-Memory, SQLite, PostgreSQL, and MySQL databases (all strongly consistent) without a distributed cache, so this is not an issue with a standard setup.
+- This issue is mitigated by frequent writes to the store that does cause stale data to be invalidated.
+
+## Best Practices
```
aaguiarz: Can we explain how to configure it? e.g which settings enable each cache?

docs/caching.md
```diff
@@ -0,0 +1,181 @@
...
+1. [**Check Query Cache**](#check-query-cache): Caches results of sub-problems within Check API requests to avoid recomputing them.
+2. [**Check Iterator Cache**](#check-iterator-cache): Caches database query results (iterators) to reduce database load during Check operations.
+3. [**List Objects Iterator Cache**](#list-objects-iterator-cache): Same as Check Iterator Cache, but used for List Objects requests.
+4. [**Cache Controller**](#cache-controller): Periodically invalidates cache entries in the background based on recent writes to the store.
+
+Note: For any request, if the `HIGHER_CONSISTENCY` consistency preference is specified, caching is bypassed entirely.
+
+## Cache Implementation
+
+OpenFGA uses the [Theine](https://github.com/Yiling-J/theine-go) library for in-memory caching:
```
aaguiarz: This is not very relevant for cache users, maybe we can mention it at the end. I'd move the content from the Note in this section as part of the note content of the previous section.

docs/caching.md
saad-h1: My one concern with mentioning config fields is that we could forget/not know to update this if things change, so this could become out of date. Also the config options listed in https://openfga.dev/docs/getting-started/setup-openfga/configuration should already have descriptions for each thing, and that table is automatically generated from the schema in the code. I added a generic statement here about looking at the config to enable cache but do you still think it's worth mentioning the exact config fields?

Depending on where this document lives it _could_ also make sense to automatically generate config options for caching specifically in each section... but that might get quite a bit more complicated.

docs/caching.md
saad-h1: Yes - there's `OPENFGA_CHECK_CACHE_LIMIT` for the max number of entries (of any kind) that are cached at one time. For each iterator cache entry, `OPENFGA_CHECK_ITERATOR_CACHE_MAX_RESULTS` and `OPENFGA_LIST_OBJECTS_ITERATOR_CACHE_MAX_RESULTS` are the max number of tuples to store in a single iterator cache entry.

Left a generic statement about checking the config with a similar concern as [this comment](https://github.com/openfga/openfga/pull/2664#discussion_r2666436995), but let me know if you still think it's worth explicitly mentioning the config field(s).

docs/caching.md
```diff
@@ -0,0 +1,170 @@
...
+- Controller TTL ≤ your acceptable staleness (e.g., 10s).
+- Cache TTLs as long as your memory budget and workload allow.
+
+Note: "invalidation" is done by setting new cache entries which are checked before using a regular cache entry. Thus, the number of cache removals (eviction/expiration/deletion) in the metrics may be inflated.
+
+### Example
+
+Consider a single replica OpenFGA server with a cache controller TTL of 10s and check query, check iterator, list objects iterator cache TTLs of 300s:
+
+Time | Event | Result | Notes | Invalidation
```
JeremyLoy, line 140: I took a stab at converting this to Mermaid as I think a swimlane diagram might make this more clear. I don't think its correct though as I don't think I have a good understanding of the literal entities. I also wasn't following along with the table 100% so I might have translated it incorrectly

If you think this representation is useful, I can help polish it up more.
```mermaid
sequenceDiagram
    participant C as Client
    participant CC as Cache Controller
    participant Q as Check Query Cache
    participant CI as Check Iterator Cache
    participant LI as List Objects Iterator Cache
    participant S as Service/Store
    participant I as Invalidator

    Note over C, I: t=0s: Initial Request
    C->>CC: Check 1
    CC->>Q: 1. Lookup Query Cache (Miss)
    CC->>CI: 2. Lookup Iterator Cache (Miss)
    CC->>S: 3. Fetch from Store
    S-->>CI: Set Key: [tuples] (t=0)
    S-->>Q: Set Key: true (t=0)
    CC-->>C: allowed: true
    CC->>I: Trigger Invalidation (Set TTL)

    rect rgb(240, 240, 240)
        Note right of I: t=0s + ε (Background)
        activate I
        I->>I: No recent writes to invalidate
        deactivate I
    end

    Note over C, I: t=1s: Data Change
    C->>S: Write: Tuple deleted
    Note right of S: Store updated

    Note over C, I: t=5s
    C->>CC: Check 2
    CC->>Q: Lookup Query Cache (Stale Hit)
    Q-->>C: allowed: true

    Note over C, I: t=10s: TTL Reset
    C->>CC: Check 3
    CC->>Q: Lookup Query Cache (Stale Hit)
    CC->>I: Trigger Invalidation (TTL expired)
    Q-->>C: allowed: true

    rect rgb(255, 230, 230)
        Note right of I: t=10s + ε (Background)
        activate I
        I->>Q: Purge all Query Cache
        I->>CI: Purge relevant Iterator Cache
        deactivate I
    end

    Note over C, I: t=11s: The Recovery
    C->>CC: Check 4
    CC->>Q: 1. Lookup Query Cache (Invalid)
    CC->>CI: 2. Lookup Iterator Cache (Invalid)
    CC->>S: 3. Fetch from Store
    S-->>CI: Set Key: [tuples] (t=11)
    S-->>Q: Set Key: false (t=11)
    CC-->>C: allowed: false

    Note over C, I: t=312s: Data Change
    C->>S: Write: Tuple added

    Note over C, I: t=313s
    C->>CC: List Objects 1
    CC->>LI: Lookup List Iterator (Stale Hit)
    CC->>I: Trigger Invalidation (Immediate)
    LI-->>C: Returns stale results

    rect rgb(255, 230, 230)
        Note right of I: t=313s + ε (Background)
        activate I
        I->>LI: Purge List Iterator Cache
        deactivate I
    end

    Note over C, I: t=314s
    C->>CC: Check 8
    CC->>Q: 1. Lookup Query Cache (Invalid)
    CC->>CI: 2. Lookup Iterator Cache (Invalid)
    CC->>S: 3. Fetch from Store
    S-->>CI: Set Key: [tuples] (t=314)
    S-->>Q: Set Key: true (t=314)
    CC-->>C: allowed: true
 ```

File: docs/caching.md [hunk 9]
```diff
@@ -0,0 +1,170 @@
...
+### Enablement
+
+The different caches must explicitly enabled - see [OpenFGA Configuration Options](https://openfga.dev/docs/getting-started/setup-openfga/configuration) for more info.
+
+### Production Deployment
+
+1. **Gradual rollout**: Enable caching incrementally, starting with query cache
+2. **Monitor memory usage**: Track cache memory consumption
+3. **Set up alerting**: Alert on low cache hit rates or high eviction rates
+4. **Regular tuning**: Periodically review and adjust cache settings based on metrics
```
hamilton-earthscope, line 214: It would be useful if the exact names of the cache metrics were mentioned here.

maybe even promote cache observability to its own section

docs/caching.md
```diff
@@ -0,0 +1,170 @@
...
+- **Benefits**: Eliminates computation for repeated identical Checks or different Checks that share common relationship evaluations.
+
+#### Example
+
+Consider the [Entitlements sample store](https://github.com/openfga/sample-stores/tree/79fa8c1710f12d3f0befaba77f15463bb5a97860/stores/entitlements) and OpenFGA running with check query cache enabled. Given the query "does user `user:charles` have relationship `can_access` with `feature:draft_prs`?", the following cache entries (simplified) are set:
+
+1. (Final result) Key: does user `user:charles` have relationship `can_access` with object `feature:draft_prs`? Value: `true`.
+2. (Sub-problem) Key: does user `user:charles` have relationship `subscriber_member` with object `plan:enterprise`? Value: `true`.
+
+### Check Iterator Cache
```
hamilton-earthscope, line 43: Can this section be expanded to explain cache behavior for tuples with context?

aaguiarz:
@saad-h1 based on feedback from the meeting:
- Document how cache controller eviction works (what happens in check cache and iterator cache).
- Document what's the cache key for each cache type
- Document how to configure it

docs/caching.md
saad-h1: updated

docs/caching.md
saad-h1: this has been re-written since

docs/caching.md
saad-h1: since community also wanted specific config fields mentioned, I've updated all sections to mention the relevant configs

docs/caching.md
saad-h1: since community also wanted specific config fields mentioned, I've updated all sections to mention the relevant configs

docs/caching.md
saad-h1: done! added an observability section with all caching-related metrics names/descriptions

docs/caching.md
saad-h1: Added a note about contextual/conditional tuples - is that the info you were looking for?

docs/caching.md
hamilton-earthscope: thank you! that's excellent

docs/caching.md
hamilton-earthscope: yes i think so

docs/caching.md
```diff
@@ -0,0 +1,258 @@
...
+
+### Enablement
+
+The different caches must explicitly enabled - see [OpenFGA Configuration Options](https://openfga.dev/docs/getting-started/setup-openfga/configuration) for more info.
+
+### Production Deployment
+
+1. **Gradual rollout**: Enable caching incrementally, starting with query cache
+2. **Monitor memory usage**: Track cache memory consumption
+3. **Set up alerting**: Alert on low cache hit rates or high eviction rates
```
hamilton-earthscope, line 211: > Alert on low cache hit rates

Can you please also clarify how to compute hit ratios for each cache? For example, it appears the hit ratio for the query cache should be

```
openfga_check_cache_hit_count_total / openfga_check_cache_count_total
```

while the hit ratio for the iterator cache appears to be

```
openfga_tuples_cache_hit_count_total / (openfga_tuples_cache_hit_count_total + openfga_tuples_cache_count_total)
```

but it's not very clear from the descriptions in the table above. I.e. is `openfga_tuples_cache_count_total` the iterator cache miss count?

hamilton-earthscope: based on [where the counter is incremented](https://github.com/openfga/openfga/blob/128b7013081ba9974eafa400aba4fa695eeec997/pkg/storage/storagewrappers/cached_datastore.go#L341), it appears that `openfga_tuples_cache_count_total` is the total hit+miss count, so the hit ratio for the iterator cache is simply

```
openfga_tuples_cache_hit_count_total / openfga_tuples_cache_count_total
```

docs/caching.md
saad-h1: Hey Jeremy, sorry for the delay, I've been out for a bit but just getting back to this now - I'm open to anything that folks think might aid in understanding this, this piece is definitely the most complicated 🙂 there is some info in the table that isn't translated here such as Checks 5-7 not showing or the explicit callout of why invalidation didn't run for certain calls, but I think whatever is shown in the diagram is accurate at least. If there's any more polishing up that would be great, I think it wouldn't hurt to have both the table and diagram to supplement each other.

docs/caching.md
saad-h1: Sure thing, and yep that's right! Added that info here now

adriantam:
@coderabbitai review

docs/caching.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Invalidation Details
+
+Invalidation runs asynchronously (eventually consistent) and is triggered on:
+
+- Any Check request if cache controller is enabled, but not more than once every cache controller TTL.
+- Any List Objects request if List Objects iterator cache is enabled, **irrespective** of cache controller TTL.
+- Only one invalidation can be running at a time.
+
+When the cache controller runs invalidation, it queries the database's changelog table for the most recent writes and saves the time of the most recent write. Then, it sets a cache entry of type `changelog` with key `cc.<storeID>` which contains the time of the latest write to the store. Now, invalidation is handled differentl for query vs. iterator cache entries:
```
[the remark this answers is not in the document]
saad-h1: fixed

docs/caching.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Observability
+
+OpenFGA exposes the following metrics for caching:
+
+Metric Name                                               | [Type](https://prometheus.io/docs/concepts/metric_types/) | Description
+----------------------------------------------------------|-----------|------------------------------------------------------------
+`openfga_cache_item_count`                                | Gauge     | The current number of items stored in the cache
+`openfga_cache_item_removed_count_total`                  | Counter   | The total number of items removed (evicted/expired/deleted) from the cache
+`openfga_cachecontroller_cache_count_total`               | Counter   | The total number of cache controller requests triggered by Check.
```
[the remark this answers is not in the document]
saad-h1: Ah so when I was looking in the Prometheus UI after starting an openfga server locally, I saw it renamed the metrics to comply with the [prometheus naming standards](https://prometheus.io/docs/practices/naming/), e.g., moving `_total` to the end for counter metrics:

image

and I believe I remember seeing this in the Grafana UI as well.

But now I'm checking the raw metrics and see that the name is just as set in the code:

image

Maybe I'll change this back to the original names and add a note for viewing in Prometheus/Grafana UIs.

docs/caching.md
```diff
@@ -0,0 +1,260 @@
...
+- The cache is in-memory, so different replicas of the service do not share their caches. Therefore, its effectiveness depends on the probability of repeated/similar requests hitting the same replica, which may depend on the model, tuple distribution, number of replicas, and the load balancing algorithm used.
+
+## Understanding Cache Types
+
+### Check Query Cache
+
+- **What it caches**: Authorization decisions (allow/deny) for the overall Check request as well as for any intermediate ("sub-problem") Checks.
+- **Benefits**: Eliminates computation for repeated identical Checks or different Checks that share common relationship evaluations.
+- **Cache key**: `hash("{object}#{relation}@{user} sp.{storeID}/{modelID}[/{contextual_tuples_comma_separated}]")`
+  - E.g., `hash('plan:enterprise#subscriber_member@user:charles sp.01KFB9MX3X2BK3EP2CBJVX8JDC/01KFB9MX467JS5BRQWQXQXPD8Z/organization:cups#member@user:alice,organization:cups#member@user:charles')` = `17904216596666502988`
```
aaguiarz: what about 'context' parameters?

docs/caching.md
```diff
@@ -0,0 +1,260 @@
...
+`openfga_cachecontroller_cache_invalidation_count_total`  | Counter   | The total number of invalidation requests that invalidated iterator caches.
+`openfga_cachecontroller_invalidation_duration_ms_bucket` | Histogram | The duration (in ms) required for cache controller to find changes and invalidate labeled by whether invalidation is required and buckets of changes size.
+`openfga_cachecontroller_invalidation_duration_ms_count`  | "         | "
+`openfga_cachecontroller_invalidation_duration_ms_sum`    | "         | "
+`openfga_check_cache_count_total`                         | Counter   | The total number of calls to ResolveCheck with caching enabled (including any recursive calls).
+`openfga_check_cache_hit_count_total`                     | Counter   | The total number of valid Check Query cache hits for ResolveCheck (including any recursive calls).
+`openfga_check_cache_invalid_hit_count_total`             | Counter   | The total number of Check Query cache hits for ResolveCheck (including any recursive calls) that were discarded because they were invalidated.
```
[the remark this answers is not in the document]
saad-h1: updated, see https://github.com/openfga/openfga/pull/2664#discussion_r2841444367

docs/caching.md
```diff
@@ -0,0 +1,260 @@
...
+t=310s + ε | Invalidation from Check 6 complete | Nothing to invalidate | Sees cache entries were set at t=11s which is after the latest write at t=1s, so nothing to invalidate |
+t=311s | Check 7 | Returns `allowed: false` | Local cache has finally expired, so computes result from database and caches locally | Not triggered as the cache controller TTL hasn't passed since the last invalidation from Check.
+t=312s | Write: Tuple added | Store updated | Previous Check is now invalid |
+t=313s | List Objects 1 | Returns stale results | Uses iterator cache entries from previous Check | Triggered.
+t=313s + ε | Invalidation from List Objects 1 complete | Invalidates cache | Sees cache entries were set at t=310s which is before the latest write at t=312s, so invalidates cache |
+t=314s | Check 8 | Returns `allowed: true` | Finds local cache entries but they are invalid, so computes result from database and sets new local cache entries | Not triggered as the cache controller TTL hasn't passed since the last invalidation from Check.
```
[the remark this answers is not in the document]
saad-h1: `no_leading_or_trailing` pipe style is valid based on [MD055](https://github.com/DavidAnson/markdownlint/blob/0213a0274dfeaac589af389f6c9bc9e821655810/doc/md055.md)

docs/caching.md
```diff
@@ -0,0 +1,260 @@
...
+`openfga_check_cache_hit_count_total`                     | Counter   | The total number of valid Check Query cache hits for ResolveCheck (including any recursive calls).
+`openfga_check_cache_invalid_hit_count_total`             | Counter   | The total number of Check Query cache hits for ResolveCheck (including any recursive calls) that were discarded because they were invalidated.
+`openfga_current_iterator_cache_count`                    | Gauge     | The current number of cached iterator instances.
+`openfga_tuples_cache_count_total`                        | Counter   | The total number of created cached iterator instances.
+`openfga_tuples_cache_discard_count_total`                | Counter   | The total number of discards from cached iterator instances.
+`openfga_tuples_cache_hit_count_total`                    | Counter   | The total number of cache hits from cached iterator instances.
+`openfga_tuples_cache_size_bucket`                        | Histogram | The number of tuples cached from iterator cache entries.
+`openfga_tuples_cache_size_count`                         | "         | "
+`openfga_tuples_cache_size_sum`                           | "         | "
```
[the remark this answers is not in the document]
saad-h1: see https://github.com/openfga/openfga/pull/2664#discussion_r2848054610

docs/caching.md
saad-h1: Ah good catch I missed that, they [do](https://github.com/openfga/openfga/blob/e04bde9ef17409e14a6a26dd243fa8ebc0f650b2/internal/graph/resolve_check_request.go#L107) include context parameters in the cache key. Updated the docs to mention this.

telemetry/grafana/dashboards/openfga.json
```diff
@@ -340,7 +340,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "(sum(increase(openfga_check_cache_hit_count{service=\"openfga\"}[1m]))\n  /\nsum(increase(openfga_check_cache_total_count{service=\"openfga\"}[1m])))* 100",
+          "expr": "(sum(increase(openfga_check_cache_hit_count_total{service=\"openfga\"}[1m]))\n  /\nsum(increase(openfga_check_cache_count_total{service=\"openfga\"}[1m])))* 100",
```
saad-h1, line 343: this is because the datasource is prometheus, which renames metrics to conform to its naming conventions.

before the change:

Image

after the change:

Image

docs/caching.md
saad-h1: I'll merge this PR for now since I think it might be confusing to add a diagram that only partially matches the table, but if you get around to updating this feel free to open a PR and I'll give it a review 🙂 otherwise I'll probably revisit these docs in a few months and will see if I can add it then. Ty for this!

PR #2691, opened 2025-09-08
added support for ttus with multiple parents in the intersection for LO

yissellokta:
When TTUs have more than one parent and are part of an intersection for LO, the new optimization code was not treating the TTUs as one group even if it has multiple edges when they are part of the same definition of ttu. This code groups the ttus that are part of the same definition in one group, and groups all the direct edges in one group. Find the lowest weight for the group weight, and the group weight is calculated as the maximum weight. To convert the groups back to typesystem we need to treat them as a group instead of individuals. This last issue did  not affect correctness, but it could affect performance

## Review Checklist

## Summary by CodeRabbit

- New Features
  - More consistent evaluation of complex intersections in authorization checks, improving result accuracy in edge-heavy scenarios.
- Refactor
  - Reworked intersection handling to support multiple lowest-weight edges and grouped sibling edges, with dynamic userset construction and aggregation.
  - Adjusted direct-edge evaluation to use a simplified “this” userset where applicable.
- Tests
  - Updated and reorganized test suite to align with the new edge grouping model.
  - Removed obsolete cases and refined DSL fixtures for clarity.

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand_weighted.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand_weighted.go
@@ -677,48 +677,48 @@ func (c *ReverseExpandQuery) intersectionHandler(
 	sourceUserType string,
 	resolutionMetadata *ResolutionMetadata,
 ) error {
-	intersectionEdgeComparison, err := typesystem.GetEdgesForIntersection(edges, sourceUserType)
+
+	intersectionEdges, err := typesystem.GetEdgesForIntersection(edges, sourceUserType)
 	if err != nil {
 		return fmt.Errorf("%w: operation: intersection: %s", ErrLowestWeightFail, err.Error())
 	}
 
-	if !intersectionEdgeComparison.DirectEdgesAreLeastWeight && intersectionEdgeComparison.LowestEdge == nil {
-		// no need to go further because list objects must return empty
+	// if no edges to call LO, return nill
+	if len(intersectionEdges.LowestEdges) == 0 {
 		return nil
 	}
 
-	lowestWeightEdges := []*weightedGraph.WeightedAuthorizationModelEdge{intersectionEdgeComparison.LowestEdge}
-
-	if intersectionEdgeComparison.DirectEdgesAreLeastWeight {
-		lowestWeightEdges = intersectionEdgeComparison.DirectEdges
-	}
-
 	tmpResultChan := make(chan *ReverseExpandResult, listObjectsResultChannelLength)
-
-	siblings := intersectionEdgeComparison.Siblings
-	usersets := make([]*openfgav1.Userset, 0, len(siblings)+1)
-
-	if !intersectionEdgeComparison.DirectEdgesAreLeastWeight && len(intersectionEdgeComparison.DirectEdges) > 0 {
-		// direct weight is not the lowest edge. Therefore, need to call check against directly assigned types.
-		usersets = append(usersets, typesystem.This())
-	}
-
-	for _, sibling := range siblings {
-		userset, err := c.typesystem.ConstructUserset(sibling)
+	intersectEdges := intersectionEdges.SiblingEdges
+	usersets := make([]*openfgav1.Userset, 0, len(intersectEdges))
+	for _, edges := range intersectEdges {
+		// no matter how many direct edges, for typesystem only required this
+		if edges[0].GetEdgeType() == weightedGraph.DirectEdge {
+			usersets = append(usersets, typesystem.This())
+			continue
+		}
+		// no matter how many parent types have for the same ttu rel from parent will be only one created in the typesystem
+		// for any other case, does not have more than one edge, the groupings only occur in direct edges or ttu edges
+		userset, err := c.typesystem.ConstructUserset(edges[0])
 		if err != nil {
-			// This should never happen.
 			return fmt.Errorf("%w: operation: intersection: %s", ErrConstructUsersetFail, err.Error())
 		}
 		usersets = append(usersets, userset)
 	}
-	userset := &openfgav1.Userset{
-		Userset: &openfgav1.Userset_Intersection{
-			Intersection: &openfgav1.Usersets{
-				Child: usersets,
-			}}}
+
+	var userset *openfgav1.Userset
+	if len(usersets) > 1 {
+		userset = &openfgav1.Userset{
+			Userset: &openfgav1.Userset_Intersection{
+				Intersection: &openfgav1.Usersets{
+					Child: usersets,
+				}}}
+	} else {
+		userset = usersets[0]
+	}
 
 	// Concurrently find candidates and call check on them as they are found
-	c.findCandidatesForLowestWeightEdge(pool, req, tmpResultChan, lowestWeightEdges, sourceUserType, resolutionMetadata)
+	c.findCandidatesForLowestWeightEdge(pool, req, tmpResultChan, intersectionEdges.LowestEdges, sourceUserType, resolutionMetadata)
 	c.callCheckForCandidates(pool, req, tmpResultChan, resultChan, userset, true, resolutionMetadata)
 
 	return nil
--- a/pkg/typesystem/weighted_graph.go
+++ b/pkg/typesystem/weighted_graph.go
@@ -2,8 +2,10 @@ package typesystem
 
 import (
 	"fmt"
+	"math"
 	"slices"
 
+	"github.com/google/uuid"
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 	"github.com/openfga/language/pkg/go/graph"
 
@@ -23,14 +25,8 @@ func hasPathTo(nodeOrEdge weightedGraphItem, destinationType string) bool {
 }
 
 type IntersectionEdges struct {
-	LowestEdge *graph.WeightedAuthorizationModelEdge   // nil if direct is lowest, otherwise lowest edge
-	Siblings   []*graph.WeightedAuthorizationModelEdge // all non lowest and excluding direct edges siblings
-	// We need to separate out the direct edges from the other edges so that in the case of direct edges are not the lowest weight,
-	// the check against these direct edges do not require checking against these individual directly assigned edges.
-	// Instead, these directly assigned edges should be treated as a group such that if one of these edges satisfy,
-	// the intersection satisfy.
-	DirectEdges               []*graph.WeightedAuthorizationModelEdge
-	DirectEdgesAreLeastWeight bool // whether direct edges are the lowest weight edges
+	LowestEdges  []*graph.WeightedAuthorizationModelEdge   // lowest edges to apply list objects
+	SiblingEdges [][]*graph.WeightedAuthorizationModelEdge // the rest of the edges to apply intersection
 }
 
 // GetEdgesForIntersection returns the lowest weighted edge and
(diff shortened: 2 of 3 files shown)


pkg/typesystem/weighted_graph.go
```diff
@@ -44,76 +40,64 @@ func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sour
 		return IntersectionEdges{}, fmt.Errorf("invalid edges for source type %s", sourceType)
 	}
 
-	directEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
-	var lowestEdge *graph.WeightedAuthorizationModelEdge
-	directEdgesAreLowest := false
-	lowestWeight := 0
+	// Group edges by type
+	groupedEdges := make(map[string][]*graph.WeightedAuthorizationModelEdge)
+	directEdgesKey := "direct_edges"
```
adriantam: This should be a constant.

pkg/typesystem/weighted_graph.go
```diff
@@ -44,76 +40,64 @@ func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sour
 		return IntersectionEdges{}, fmt.Errorf("invalid edges for source type %s", sourceType)
 	}
 
-	directEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
-	var lowestEdge *graph.WeightedAuthorizationModelEdge
-	directEdgesAreLowest := false
-	lowestWeight := 0
+	// Group edges by type
+	groupedEdges := make(map[string][]*graph.WeightedAuthorizationModelEdge)
```
adriantam: should we pre-allocate groupedEdges to have a maximum length of edges? i.e.,

```
groupedEdges := make(map[string][]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
```

pkg/typesystem/weighted_graph_test.go
```diff
@@ -270,81 +232,61 @@ func TestGetEdgesForIntersection(t *testing.T) {
...
-		require.Empty(t, comparator.DirectEdges)
-		require.Len(t, comparator.Siblings, 1)
-		require.Equal(t, graph.OperatorNode, comparator.Siblings[0].GetTo().GetNodeType())
-		require.Equal(t, graph.RewriteEdge, comparator.Siblings[0].GetEdgeType())
-		require.Equal(t, graph.IntersectionOperator, comparator.Siblings[0].GetTo().GetLabel())
+		_, err = GetEdgesForIntersection(actualIntersectionEdges, "user")
+		require.Error(t, err)
 	})
-	t.Run("non_directed_assignment_no_path", func(t *testing.T) {
+
```
adriantam, line 263: Can we also add a case where there are multiple TTUs in an intersection and this is the lowest weight? i.e.,

```
			model
				schema 1.1
			type user
			type subteam1
				relations
					define member: [user]
			type subteam2
				relations
					define member: [user]
			type adhoc
				relations
					define member: [subteam1#member]
			type group
				relations
					define parent: [subteam1, subteam2]
					define member: [adhoc#member] and member from parent
```

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -676,49 +676,49 @@
...
+		}
+		// no matter how many parent types have for the same ttu rel from parent will be only one created in the typesystem
+		// for any other case, does not have more than one edge, the groupings only occur in direct edges or ttu edges
+		userset, err := c.typesystem.ConstructUserset(edges[0])
 		if err != nil {
-			// This should never happen.
 			return fmt.Errorf("%w: operation: intersection: %s", ErrConstructUsersetFail, err.Error())
 		}
 		usersets = append(usersets, userset)
 	}
```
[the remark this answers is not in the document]
adriantam: There should not be possible because an intersection will always have at least 1 siblings as it is always something like `A and B`.  You should not be able to write a model such that it is `A and A`.

However, it will be good to call this out as comments.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
adriantam, line 1: Can you add the following use case in reverse_expand_weighted_test.go (TestReverseExpandWithWeightedGraph)


```
{
			name: "ttus_with_multiple_parents",
			model: `
				model
					schema 1.1
				type user
				type group
					relations
						define member: [user]
				type team
					relations
						define member: [user]
				type type1
					relations
						define member: [user]
				type type2
					relations
						define member: [type1#member]
				type doc
					relations
						define parent: [group, team]
						define owner: [type2#member] and member from parent	
		`,
			tuples: []string{
				"doc:1#parent@group:1",
				"group:1#member@user:1",
				"doc:1#parent@team:1",
				"team:1#member@user:2",
				"doc:1#owner@type2:1#member",
				"type2:1#member@type1:1#member",
				"type1:1#member@user:1",
				"type1:1#member@user:2",
			},
			objectType: "doc",
			relation:   "owner",
			user:       &UserRefObject{Object: &openfgav1.Object{Type: "user", Id: "2"}},
			expectedOptimizedObjects: []string{
				"doc:1",
			},
			expectedUnoptimizedObjects: []string{
				"doc:1",
			},
		},
```

adriantam (commented):
Can you run `make lint` to satisfy linter?  Thanks.

adriantam:
Should also update changelog.

yissellokta:
> Should also update changelog.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -677,48 +677,48 @@ func (c *ReverseExpandQuery) intersectionHandler(
...
+				Intersection: &openfgav1.Usersets{
+					Child: usersets,
+				}}}
+	} else {
+		userset = usersets[0]
+	}
 
 	// Concurrently find candidates and call check on them as they are found
-	c.findCandidatesForLowestWeightEdge(pool, req, tmpResultChan, lowestWeightEdges, sourceUserType, resolutionMetadata)
+	c.findCandidatesForLowestWeightEdge(pool, req, tmpResultChan, intersectionEdges.LowestEdges, sourceUserType, resolutionMetadata)
```
elbuo8, line 718: Minor performance improvement. This producer can be called earlier.

pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go
```diff
@@ -935,6 +935,49 @@ func TestReverseExpandWithWeightedGraph(t *testing.T) {
 			expectedOptimizedObjects:   []string{"org:a"},
 			expectedUnoptimizedObjects: []string{"org:a", "org:c"},
 		},
+		{
+			name: "ttus_with_multiple_parents",
```
elbuo8, line 939: Is there no matrix for LO? Ideally this should be covered at a high level if we make yet another strategy/refactor.

pkg/typesystem/weighted_graph.go
```diff
@@ -44,76 +41,64 @@ func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sour
...
-	directEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
-	var lowestEdge *graph.WeightedAuthorizationModelEdge
-	directEdgesAreLowest := false
-	lowestWeight := 0
-
-	// Process all edges first to separate direct from non-direct
-	nonDirectEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
+	// Group edges by type
+	groupedEdges := make(map[string][]*graph.WeightedAuthorizationModelEdge, len(edges))
+	const directEdgesKey = "direct_edges"
```
elbuo8: Top of package since makes no point to re-declare each time. Perhaps this is something the `wg` should actually expose.

pkg/typesystem/weighted_graph.go
```diff
@@ -179,10 +156,10 @@ func (t *TypeSystem) ConstructUserset(currentEdge *graph.WeightedAuthorizationMo
 				Userset: &openfgav1.Userset_TupleToUserset{
 					TupleToUserset: &openfgav1.TupleToUserset{
 						Tupleset: &openfgav1.ObjectRelation{
-							Relation: parent,
+							Relation: parent, // parent
 						},
 						ComputedUserset: &openfgav1.ObjectRelation{
-							Relation: relation,
+							Relation: relation, // rel2
```
elbuo8: 🤔

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -677,48 +677,48 @@ func (c *ReverseExpandQuery) intersectionHandler(
...
 		}
 		usersets = append(usersets, userset)
 	}
-	userset := &openfgav1.Userset{
-		Userset: &openfgav1.Userset_Intersection{
-			Intersection: &openfgav1.Usersets{
-				Child: usersets,
-			}}}
+
+	var userset *openfgav1.Userset
```
elbuo8, line 707:
```suggestion
	var userset usersets[0]
```

And eliminate the `else`?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -677,48 +677,48 @@ func (c *ReverseExpandQuery) intersectionHandler(
...
-		// direct weight is not the lowest edge. Therefore, need to call check against directly assigned types.
-		usersets = append(usersets, typesystem.This())
-	}
-
-	for _, sibling := range siblings {
-		userset, err := c.typesystem.ConstructUserset(sibling)
+	intersectEdges := intersectionEdges.SiblingEdges
+	usersets := make([]*openfgav1.Userset, 0, len(intersectEdges))
+	for _, edges := range intersectEdges {
+		// no matter how many direct edges, for typesystem only required this
```
elbuo8: Can we add a bit more context on the comment?

Also why have this case if its already handled by `typesystem.ConstructUserset`? I don't think we gain much there or I'm missing something.

pkg/typesystem/weighted_graph.go
```diff
@@ -44,76 +41,64 @@ func GetEdgesForIntersection(edges []*graph.WeightedAuthorizationModelEdge, sour
...
-	// Process all edges first to separate direct from non-direct
-	nonDirectEdges := make([]*graph.WeightedAuthorizationModelEdge, 0, len(edges))
+	// Group edges by type
+	groupedEdges := make(map[string][]*graph.WeightedAuthorizationModelEdge, len(edges))
+	const directEdgesKey = "direct_edges"
 
-	// It is assumed that direct edge always appear first.
 	for _, edge := range edges {
-		weight, ok := edge.GetWeight(sourceType)
+		var key string
```
elbuo8: Mega minor... can we move this declaration after the next `if`.

pkg/typesystem/weighted_graph.go
senojj: Note that Go does not really "re-declare" constants in the sense of allocating new memory for the value. Constants are placed into a specific section of program memory at compile time, but the scope of its value is limited to the scope in which it is declared.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
senojj: `typesystem.ConstructUserset` constructs a `Userset_ComputedUserset` in the case of a `DirectEdge`. I believe the _correct_ value in this case is `Userset_This`.

I assume changing `typesystem.ConstructUserset` was deferred to a later change due to its usage in other parts of the code.

pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go
```diff
@@ -935,6 +935,49 @@ func TestReverseExpandWithWeightedGraph(t *testing.T) {
...
+			objectType: "doc",
+			relation:   "owner",
+			user:       &UserRefObject{Object: &openfgav1.Object{Type: "user", Id: "2"}},
+			expectedOptimizedObjects: []string{
+				"doc:1",
+			},
+			expectedUnoptimizedObjects: []string{
+				"doc:1",
+			},
+		},
```
adriantam, line 980: Will be good to add a test for the exclusion case where B is a TTU with multiple parent

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -673,52 +665,57 @@ func (c *ReverseExpandQuery) intersectionHandler(
...
 	resultChan chan<- *ReverseExpandResult,
-	edges []*weightedGraph.WeightedAuthorizationModelEdge,
+	intersectionNode *weightedGraph.WeightedAuthorizationModelNode,
 	sourceUserType string,
 	resolutionMetadata *ResolutionMetadata,
 ) error {
-	intersectionEdgeComparison, err := typesystem.GetEdgesForIntersection(edges, sourceUserType)
-	if err != nil {
-		return fmt.Errorf("%w: operation: intersection: %s", ErrLowestWeightFail, err.Error())
+	if intersectionNode == nil || intersectionNode.GetNodeType() != weightedGraph.OperatorNode || intersectionNode.GetLabel() != weightedGraph.IntersectionOperator {
```
elbuo8, line 672: 🤔 what are the circumstances for this?

yissellokta: fails if the node in the arguments is not a valid interception node for the intersectionHandler

justincoh: Could we reduce this to only check `nil` and the `NodeType`? Checking the type and the label feels redundant.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -735,24 +732,33 @@ func (c *ReverseExpandQuery) exclusionHandler(
 	pool *concurrency.Pool,
 	req *ReverseExpandRequest,
 	resultChan chan<- *ReverseExpandResult,
-	edges []*weightedGraph.WeightedAuthorizationModelEdge,
+	exclusionNode *weightedGraph.WeightedAuthorizationModelNode,
 	sourceUserType string,
 	resolutionMetadata *ResolutionMetadata,
 ) error {
-	baseEdges, excludedEdge, err := typesystem.GetEdgesForExclusion(edges, sourceUserType)
+	if exclusionNode == nil || exclusionNode.GetNodeType() != weightedGraph.OperatorNode || exclusionNode.GetLabel() != weightedGraph.ExclusionOperator {
```
justincoh, line 739: Same comment from above. Can we just check `nil` and `Type()`?

senojj: Each of these conditions is an invariant of the function. Passing a node value to this function that is _not_ an Exclusion operator node is _not_ an expected or recoverable exception; it is programming error.

Therefore, I might even suggest using a `panic` instead of returning an `error`.

adriantam: In the interest of resiliency, should we simply return 500 error instead of crash and burn?

adriantam:
@coderabbitai generate docstrings

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
senojj: panic != "crash and burn" if handled appropriately. the problem is that our code is not currently written in a way that can handle it properly by propagating it to the middleware layer.

note that I'm only suggesting using a panic for violations of function invariants -- impossibilities in properly written code.

senojj:
Is this PR in a completed state at this point?

adriantam:
> Is this PR in a completed state at this point?

@senojj - Yes

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -412,12 +412,15 @@ func (c *ReverseExpandQuery) execute(
...
 			}
 
-			// we can ignore this error, if the weighted graph failed to build, req.skipWeightedGraph would
-			// have prevented us from entering this block.
-			edges, needsCheck, _ := c.typesystem.GetEdgesForListObjects(
+			edges, err := c.typesystem.GetConnectedEdges(
 				typeRel,
 				sourceUserType,
 			)
+			if err != nil {
```
elbuo8: Then why have the additional check. Use `_` and put the comment of the reason.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -673,52 +665,59 @@ func (c *ReverseExpandQuery) intersectionHandler(
...
-	tmpResultChan := make(chan *ReverseExpandResult, listObjectsResultChannelLength)
-
-	siblings := intersectionEdgeComparison.Siblings
-	usersets := make([]*openfgav1.Userset, 0, len(siblings)+1)
-
-	if !intersectionEdgeComparison.DirectEdgesAreLeastWeight && len(intersectionEdgeComparison.DirectEdges) > 0 {
-		// direct weight is not the lowest edge. Therefore, need to call check against directly assigned types.
-		usersets = append(usersets, typesystem.This())
+	// if no edges to call LO or call check, return nil
+	if len(intersectionEdges.LowestEdges) == 0 || len(intersectionEdges.SiblingEdges) == 0 {
```
elbuo8: Then why the extra check if its granted 🤔 If the WG is doing this work for you already, why are you duplicating its work at runtime which will happen for each intersection node visit. If anything leave the comment for the reader.

PR #2693, opened 2025-09-09
chore: simplify set reducers (and add guards) and additional refactor for upcoming changes

elbuo8:

Simplify resolvers with our existing functions and remove clutter and limiters were they were not needed/redundant.

## Review Checklist

## Summary by CodeRabbit

- Refactor
  - Overhauled concurrency and fan-in mechanisms for better scalability and responsiveness under load.
  - Streamlined internal resolvers to use channel-based workflows, improving cancellation and early-exit behavior.

- Bug Fixes
  - More robust panic recovery and error propagation, reducing unexpected failures during checks.
  - Improved handling of cancellation and deadline scenarios to return clearer, consistent outcomes.
  - Ensured proper cleanup of in-flight work to prevent resource leaks during complex evaluations.

### DIFF
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -326,7 +326,7 @@ func exclusion(ctx context.Context, _ int, handlers ...CheckHandlerFunc) (*Resol
 			}
 
 			// Short-circuit: If base is false, the whole expression is false.
-			if res.resp != nil && (res.resp.GetCycleDetected() || !res.resp.GetAllowed()) {
+			if res.resp.GetCycleDetected() || !res.resp.GetAllowed() {
 				return &ResolveCheckResponse{Allowed: false, ResolutionMetadata: ResolveCheckResponseMetadata{CycleDetected: res.resp.GetCycleDetected()}}, nil
 			}
 
@@ -343,7 +343,7 @@ func exclusion(ctx context.Context, _ int, handlers ...CheckHandlerFunc) (*Resol
 			}
 
 			// Short-circuit: If subtract is true, the whole expression is false.
-			if res.resp != nil && (res.resp.GetCycleDetected() || res.resp.GetAllowed()) {
+			if res.resp.GetCycleDetected() || res.resp.GetAllowed() {
 				return &ResolveCheckResponse{Allowed: false, ResolutionMetadata: ResolveCheckResponseMetadata{CycleDetected: res.resp.GetCycleDetected()}}, nil
 			}
 
--- a/internal/graph/check.go
+++ b/internal/graph/check.go
@@ -5,11 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/emirpasic/gods/sets/hashset"
-	"github.com/sourcegraph/conc"
 	"github.com/sourcegraph/conc/panics"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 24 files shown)


elbuo8:
Needs: https://github.com/openfga/openfga/pull/2691

adriantam:
Do we need to investigate the memory db test failure?

elbuo8:
The perf regression can be avoided by changing how errors and handled but at the expense of a less readable loop. I'm not sure its worth it so I left it with I considered more readable.

internal/graph/check.go
```diff
@@ -141,313 +139,226 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
+	defer cancel()
 
-	ctx, cancel := context.WithCancel(ctx)
-	resultChan := make(chan checkOutcome, len(handlers))
+	pool := concurrency.NewPool(cancellableCtx, concurrencyLimit)
+	out := make(chan checkOutcome, len(handlers))
 
-	drain := resolver(ctx, concurrencyLimit, resultChan, handlers...)
+	for _, handler := range handlers {
+		h := handler // Capture loop variable for the goroutine
```
adriantam, line 230: can you explain this?  Why do we need to assign separate variable `h`?

internal/graph/check.go
```diff
@@ -141,313 +139,226 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
+			return nil, ctx.Err()
+
+		case outcome, ok := <-out:
+			if !ok {
+				break
 			}
 
-			if result.resp.GetCycleDetected() {
-				cycleDetected = true
+			if outcome.err != nil {
```
adriantam, line 193: This is slightly different than the intersection case.  In the intersection case, we store the first error but here we store the last error.  Is this intended?

internal/graph/check.go
```diff
@@ -141,313 +139,226 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
+	if finalErr != nil {
+		return nil, finalErr
 	}
 
-	resp = &ResolveCheckResponse{
-		Allowed: true,
-	}
-
-	return
+	// If the loop completes without any "false" outcomes or errors, the result is "true".
```
adriantam, line 281: I know this is not a change in functionality, do we want to call out anywhere as comments we do not propagate metadata such as datastore query count?

internal/graph/check.go
```diff
@@ -141,313 +139,226 @@ type CheckHandlerFunc func(ctx context.Context) (*ResolveCheckResponse, error)
...
-				baseErr = baseResult.err
+		case <-ctx.Done():
+			return nil, ctx.Err()
+
+		case res, ok := <-baseChan:
+			if !ok {
+				baseChan = nil // Stop selecting this case.
 				continue
 			}
+			resultsReceived++
```
adriantam, line 321: Question - why do we want to increment this after the `if !ok` in line 317?  What will happen in the case where the channel is closed immediately (for some reason without ever receiving a valid message)?  In  this case, would we be stuck in this loop until timeout?

internal/graph/check.go
elbuo8: That is the current contract.  Not looking to change that. Also, doesn't really matter because there is no order guarantees on how the results come from the channel. So in 2 different runs, you might get 2 different errors if force them to error out.

internal/graph/check.go
elbuo8: Because we are passing it via closure. The value is being mutated on each run. Therefore we create a local one to prevent the scope from being wrong.

internal/graph/check.go
elbuo8: I'm confused, what does this have to do in this context? DSQ has _long_ been gone from these reducers.

internal/graph/check.go
elbuo8: >  why do we want to increment this after the if !ok in line 317?

Because `!ok` means no message was received. Thus this loop is just to exit out.

> What will happen in the case where the channel is closed immediately (for some reason without ever receiving a valid message)? In this case, would we be stuck in this loop until timeout?

The only case in which that can happen if it's the `ctx` is cancelled. Which we have a handle for. Otherwise you receive a message.

There are units that actually cover these questions too.

adriantam: So, if I understand correctly, the intend if we see !ok without ever receiving a valid message is to return context timeout error in line 314.  Can we add comment to call this out?

elbuo8: It doesn't need to be a timeout. Whichever cancelation reason we receive, we would exit through case of `ctx.Done`. 

Notice the producer of this channel will always attempt to write to the channel. The only reason were you would get no message, and a channel closed event (because given its a buffered channel you will receive everything in the order that it happens) is if the context is no longer active.

This is somewhat explicit here:
```
go func() {
		concurrency.TrySendThroughChannel(ctx, runHandler(ctx, handlers[0]), baseChan)
		close(baseChan)
	}()

```

PR #2658, opened 2025-08-22
add latency and query count numbers in shadow check resolver

justincoh:
#### What problem is being solved?
We use this shadow pattern to test experimental improvements to Check, but we don't currently have latency metrics built into the comparison. This PR calculates the duration of both `main` and `shadow` check resolution and adds the data as log attributes.

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

## Summary by CodeRabbit

- New Features
  - None.

- Chores
  - Added latency metrics to logs for main and shadow checks to improve observability.
  - Logs now include microsecond-level durations for both execution paths (when results match or differ).
  - No changes to behavior or public interfaces.

PR #2698, opened 2025-09-11
chore: replace shadow check resolver with shadow check query (#2658)

cikasfm:
### Summary
This PR replaces the previous shadow check resolver mechanism with a command-level shadow mode implementation. Shadowing is now encapsulated inside `NewCheckCommand`, which wraps a primary `CheckQuery` and (optionally) executes a sampled shadow `CheckQuery` asynchronously.

### Key Changes
- Introduced `NewCheckQuery` (low-level) and redefined `NewCheckCommand` (high-level dispatcher supporting shadow mode).
- Removed dependency on building a separate shadow `graph.CheckResolver`.
- Unified shadow behavior across:
  - Check API
  - Batch Check API (per unique deduplicated check)
  - List Objects internal check operations
- Shadow List Objects Query remains a distinct feature; its internal checks now inherit shadow mode automatically.

### Configuration
Existing server settings feed into:
- `NewCheckCommandConfig` (main)
- `NewCheckCommandShadowConfig` (shadow: enabled flag, pct, timeout, logger)
- Combined via `NewCheckCommandServerConfig`
APIs instantiate per-request commands with `NewCheckCommand`.

### Behavior
- Shadow queries are sampled (`rand.Intn(100) < pct`), time-bounded, and non-intrusive.
- Divergences logged (`shadow check difference`), matches logged (`shadow check match`).
- No user-visible latency impact (shadow asynchronous).
- No metric changes for shadow path (future enhancement candidate).

### Non-Goals
- Modifying Shadow List Objects Query.
- Adding new metrics for divergence.
- Changing external API contracts.

### Migration
No external API changes. Operational configs continue to work; internal wiring simplified.

### Changelog
Enhancement: Replace shadow check resolver with command-level shadow mode for Check, Batch Check, and List Objects internal evaluations.

## Review Checklist

# Design: Check Command Shadow Mode (Replacement for Shadow Check Resolver)

## Summary
Introduce shadow mode at the Check Command layer, replacing the previous shadow check resolver implementation. The shadow execution now wraps the primary `CheckQuery` and (optionally) executes a sampled secondary CheckQuery asynchronously with timeout and differential logging. This applies uniformly to:
- Check API
- Batch Check API (per unique deduplicated check)
- List Objects API (indirectly via internal check usage)
Shadow List Objects Query remains a separate, existing feature and is unaffected except that its internal use of the Check Command inherits shadow mode behavior.

## Background
Previously the server created two `graph.CheckResolver` instances (primary + shadow) when shadowing was enabled. This added complexity and leaked resolver concerns outside command boundaries. The new design moves shadowing into the command construction phase, centering behavior in `NewCheckCommand`.

## Scope of This PR
- Replace shadow check resolver (dual resolver pattern) with command-level shadow mode.
- Integrate sampling + timeout + differential logging inside the Check Command.
- Ensure Batch Check and List Objects automatically benefit via reuse of the shared `CheckCommandServerConfig`.
- Do not modify Shadow List Objects Query semantics.
- Rename:
  - Old factory `NewCheckCommand` → `NewCheckQuery` (low-level constructor producing a single non-shadow query).
  - New high-level `NewCheckCommand` now selects between plain and shadowed execution (`CheckCommand` interface).

## Terminology
- Main query: primary authorization evaluation.
- Shadow query: sampled secondary evaluation (not user-visible).
- ShadowCheckQuery: wrapper implementing `CheckCommand` that runs main + optional shadow.
- ShadowListObjectsQuery: separate mechanism that shadows full list objects operations (still independent).

## Previous Design (Removed)
- Server built:
  - `checkResolver`
  - `shadowCheckResolver` (if enabled)
- Check-related APIs selected which resolver to call.
- Shadow logic lived outside the command layer → duplication + tight coupling to resolver creation.

### Core Abstractions
- `NewCheckQuery`: constructs a concrete `*CheckQuery` (single execution path).
- `NewCheckCommand`: returns either:
  - Plain `CheckQuery` (no shadow enabled or zero sample rate).
  - `shadowedCheckQuery` (wrapping main + shadow `CheckQuery` instances).
- `ShadowCheckCommandConfig`: holds enable flag, sample percentage, timeout, logger, test sync mode.

### Server Configuration Flow (`pkg/server/server.go`)
1. Build base config:
   ```go
   checkCommandConfig := commands.NewCheckCommandConfig(datastore, checkResolver, ...options)
   ```
2. Optionally build shadow config:
   ```go
   shadowCheckCommandConfig := commands.NewCheckCommandShadowConfig(
       commands.NewCheckCommandConfig(datastore, shadowCheckResolver, ...),
       commands.WithShadowCheckQueryEnabled(s.shadowCheckResolverEnabled),
       commands.WithShadowCheckQueryPct(s.shadowCheckResolverSamplePercentage),
       commands.WithShadowCheckQueryTimeout(s.shadowCheckResolverTimeout),
       commands.WithShadowCheckQueryLogger(s.logger),
   )
   ```
3. Assemble:
   ```go
   s.checkCommandServerConfig = commands.NewCheckCommandServerConfig(
       checkCommandConfig,
       commands.WithShadowCheckCommandConfig(shadowCheckCommandConfig),
   )
   ```

### Execution Path (API Layer)
| API | Construction | Shadow Sampling Point | Notes |
|-----|--------------|-----------------------|-------|
| Check | `NewCheckCommand(s.checkCommandServerConfig, params)` | Once per request | Differential logs only |
| Batch Check | Per deduplicated unique check inside worker pool | Once per unique cache key | Shared results mapped to multiple correlation IDs |
| List Objects | Internal calls use `CheckCommand` via injected config | Each internal check | Independent from ShadowListObjectsQuery feature |

### Sampling & Timeout
- Sample triggered when: `rand.Intn(100) < shadowPct`.
- Shadow executed in goroutine with `context.WithTimeout`.
- On timeout: silently abandoned (only log).
- Panic safeguarded with `recover` + error log.

### Logging
On shadow execution completion:
- Match: log `shadow check match`
- Divergence: log `shadow check difference`
Logged fields include store id, model id, request tuple, latency, datastore query counts, cycle flags.

### Error Isolation
- Shadow errors never affect user-visible response.
- Only main query error propagates.
- Shadow panic or timeout produces log entry only.

### Metadata & Metrics
- Only main query contributes to:
  - Dispatch count
  - Datastore query count
  - Duration histograms
  - Throttle counters
- Shadow path currently does not emit separate metrics (future extension possible).

## Interaction With Shadow List Objects Query
| Aspect | Shadow Check Mode | Shadow List Objects Query |
|--------|-------------------|---------------------------|
| Applies To | Check + internal Check usages (including List Objects sub-checks) | List Objects API only |
| Sampling Unit | Per check invocation | Per list objects request |
| Purpose | Detect evaluation divergences between main and experimental check logic | Compare full object enumeration logic |
| Independence | Yes | Yes (still uses Check Command internally, which may shadow) |

Thus a List Objects request may trigger:
1. ShadowListObjectsQuery (if enabled).
2. Multiple internal Check Commands, each of which may independently trigger shadow mode.

## Batch Check Specifics
- Deduplication done via hashed cache key.
- Shadow sampling performed per unique deduplicated check, reducing overhead.
- Duplicate correlation IDs share the same outcome (and therefore same shadow sample decision).

## Error Handling Summary
| Scenario | Result |
|----------|--------|
| Main query error | Propagated (converted if needed) |
| Shadow query error | Logged, discarded |
| Shadow timeout | Logged as warning |
| Shadow panic | Recovered + error log |
| Throttling (main) | Reflected in metadata + metrics |
| Throttling (shadow) | Ignored (not surfaced) |

## Migration & Backward Compatibility
- Remove reliance on constructing a separate shadow `graph.CheckResolver` externally.
- Existing flags / config values (`shadowCheckResolverEnabled`, sample percentage, timeout) now map to shadow command config.
- No API surface changes.
- Operational behavior consistent: sampling semantics preserved.
- Reduced resource footprint (no duplicate resolver allocation if disabled).

## Testing Hooks
- `runSync` flag in `ShadowCheckCommandConfig` (test-only) waits for shadow goroutine completion.
- Differential logging assertions possible by injecting a capturing logger and forcing 100% sample.

## Risks & Mitigations
| Risk | Mitigation |
|------|------------|
| Hidden performance regression | Sampling off by default or controlled percentage |
| Log noise at high sample rate | Operator controls sample percentage |
| Missed metric parity for shadow | Intentional; future phase if needed |
| Race on shared params | Shadow uses cloned context without cancellation |

## Future Extensions (Out of Scope)
- Emit structured metrics for divergence rate.
- Attach diff metadata in traces.
- Adaptive sampling based on divergence frequency.

## Changelog Entry (Proposed)
Enhancement: Replaced shadow check resolver with command-level shadow mode. Shadow execution now occurs within `NewCheckCommand`, affecting Check, Batch Check, and internal List Objects checks. Shadow List Objects Query remains unchanged. No API changes.

## Conclusion
The redesign localizes shadow logic at the command abstraction, reduces resolver complexity, preserves existing capabilities, and standardizes behavior across all consumers of the Check Command while keeping Shadow List Objects Query independent.

### ⚠️ Breaking Change: Shadow Check Resolver Refactor

The previous dedicated "shadow check resolver" (separate shadow `CheckResolver` instance wired manually in `Server` setup) has been removed. Shadow evaluation is now implemented as an internal wrapper around the main Check command via a new shadow configuration passed to `NewCheckCommandServerConfig`.

#### What Changed
- Removed: explicit construction and storage of a separate shadow resolver.
- Added: shadow execution path encapsulated in `shadowedCheckQuery` (see `pkg/server/commands/check_command_shadow.go`).
- The shadow logic is now activated by providing a `ShadowCheckCommandConfig` via `WithShadowCheckCommandConfig(...)` when building the `CheckCommandServerConfig`.
- Sampling, timeout, and logging are configured through new option functions: `WithShadowCheckQueryEnabled`, `WithShadowCheckQueryPct`, `WithShadowCheckQueryTimeout`, `WithShadowCheckQueryLogger`.

#### Impact
- Any code relying on a second resolver or fields like `shadowCheckResolver` must be updated.
- Direct references to a "shadow resolver" should be removed.
- Shadow behavior now lives entirely inside the unified `NewCheckCommand` execution path.
- Metrics and logging still fire, but shadow differences are now emitted from the new wrapper.

#### Migration Guide
1. Remove manual construction of a second shadow resolver.
2. When building the server’s check command config, inject the shadow configuration:
   ```go
   // Go
   mainCfg := commands.NewCheckCommandConfig(datastore, mainCheckResolver,
       commands.WithCheckCommandLogger(logger),
       // other options...
   )

   shadowCfg := commands.NewCheckCommandShadowConfig(
       commands.NewCheckCommandConfig(datastore, shadowCheckResolver,
           commands.WithCheckCommandLogger(logger),
       ),
       commands.WithShadowCheckQueryEnabled(true),
       commands.WithShadowCheckQueryPct(10),          // 10% sampling
       commands.WithShadowCheckQueryTimeout(500*time.Millisecond),
       commands.WithShadowCheckQueryLogger(logger),
   )

   serverCfg := commands.NewCheckCommandServerConfig(
       mainCfg,
       commands.WithShadowCheckCommandConfig(shadowCfg),
   )
   ```
3. Use `commands.NewCheckCommand(serverCfg, params)` everywhere (no branching for shadow).
4. Remove any cleanup logic tied to a second resolver instance.
5. If you previously gated shadow mode via a custom flag, map it to:
   - enable/disable → `WithShadowCheckQueryEnabled`
   - sample rate → `WithShadowCheckQueryPct`
   - timeout → `WithShadowCheckQueryTimeout`

#### Behavioral Notes
- Shadow path never affects the primary decision.
- Divergences are logged with fields: `main`, `shadow`, latency, query counts.
- If shadow sampling pct is `0` or disabled, no overhead is incurred.

#### Action Required
Update initialization code; remove legacy shadow resolver references; adopt the new unified configuration pattern.

#### Files Introduced / Modified
- Added logic: `pkg/server/commands/check_command_shadow.go`
- Updated integration: `pkg/server/server.go` (shadow config wiring)
- Unified execution path: `pkg/server/commands/check_command.go`

---

## Summary by CodeRabbit

- New Features
  - Add shadow-mode diagnostics for Check requests: configurable sampling, timeout, and structured logging for background shadow comparisons (no impact on main responses).
- Refactor
  - Centralized command construction into a server-level config; per-call parameters (including the type system) are provided at construction, Execute now runs from that configured command.
  - Removed legacy shadow resolver/path and deprecated shadow-cache/list-objects options.
- Documentation
  - Unreleased changelog entry added noting datastore query count visibility in shadow checks.

### DIFF
--- a/pkg/server/commands/batch_check_command.go
+++ b/pkg/server/commands/batch_check_command.go
@@ -23,7 +23,7 @@ type BatchCheckQuery struct {
 	logger              logger.Logger
 	maxChecksAllowed    uint32
 	maxConcurrentChecks uint32
-	checkConfig         CheckCommandServerConfig
+	checkSettings       *CheckCommandSettings
 }
 
 type BatchCheckCommandParams struct {
@@ -82,10 +82,10 @@ func WithBatchCheckMaxChecksPerBatch(maxChecks uint32) BatchCheckQueryOption {
 	}
 }
 
-func NewBatchCheckCommand(checkConfig CheckCommandServerConfig, opts ...BatchCheckQueryOption) *BatchCheckQuery {
+func NewBatchCheckCommand(checkConfig *CheckCommandSettings, opts ...BatchCheckQueryOption) *BatchCheckQuery {
 	cmd := &BatchCheckQuery{
 		logger:              logger.NewNoopLogger(),
-		checkConfig:         checkConfig,
+		checkSettings:       checkConfig,
 		maxChecksAllowed:    config.DefaultMaxChecksPerBatchCheck,
 		maxConcurrentChecks: config.DefaultMaxConcurrentChecksPerBatchCheck,
 	}
@@ -97,11 +97,6 @@ func NewBatchCheckCommand(checkConfig CheckCommandServerConfig, opts ...BatchChe
 }
 
 func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckCommandParams) (map[CorrelationID]*BatchCheckOutcome, *BatchCheckMetadata, error) {
-	if params.Typesys == nil {
-		return nil, nil, &BatchCheckValidationError{
-			Message: "missing typesystem in batch check params",
-		}
-	}
 	if len(params.Checks) > int(bq.maxChecksAllowed) {
 		return nil, nil, &BatchCheckValidationError{
 			Message: "batchCheck received " + strconv.Itoa(len(params.Checks)) + " checks, the maximum allowed is " + strconv.Itoa(int(bq.maxChecksAllowed)),
@@ -156,7 +151,7 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
 			default:
 			}
 
-			checkParams := CheckCommandParams{
+			checkParams := &CheckCommandParams{
 				StoreID:          params.StoreID,
 				TupleKey:         check.GetTupleKey(),
 				ContextualTuples: check.GetContextualTuples(),
@@ -166,9 +161,9 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
 				Operation:        "batch-check",
 			}
 
-			checkQuery := NewCheckCommand(bq.checkConfig, checkParams)
+			checkQuery := NewCheckCommand(bq.checkSettings)
 
-			response, metadata, err := checkQuery.Execute(ctx)
+			response, metadata, err := checkQuery.Execute(ctx, checkParams)
 
 			resultMap.Store(key, &BatchCheckOutcome{
 				CheckResponse: response,
--- a/pkg/server/commands/check_command.go
+++ b/pkg/server/commands/check_command.go
@@ -29,20 +29,19 @@ const (
 
 // CheckCommand interface wraps the Execute method for performing a check operation.
 type CheckCommand interface {
-	Execute(ctx context.Context) (*graph.ResolveCheckResponse, *graph.ResolveCheckRequestMetadata, error)
+	Execute(ctx context.Context, params *CheckCommandParams) (*graph.ResolveCheckResponse, *graph.ResolveCheckRequestMetadata, error)
 }
 
 type CheckQuery struct {
-	params                     CheckCommandParams
 	logger                     logger.Logger
 	checkResolver              graph.CheckResolver
+	datastore                  storage.RelationshipTupleReader
 	sharedCheckResources       *shared.SharedDatastoreResources
 	cacheSettings              config.CacheSettings
 	maxConcurrentReads         uint32
 	shouldCacheIterators       bool
 	datastoreThrottleThreshold int
 	datastoreThrottleDuration  time.Duration
-	datastoreWithTupleCache    *storagewrappers.RequestStorageWrapper
 }
 
 var _ CheckCommand = (*CheckQuery)(nil)
@@ -65,45 +64,12 @@ type CheckCommandConfig struct {
 }
 
 // NewCheckCommandConfig creates a new CheckCommandConfig.
-func NewCheckCommandConfig(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, options ...CheckQueryOption) CheckCommandConfig {
-	return CheckCommandConfig{
+func NewCheckCommandConfig(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, options ...CheckQueryOption) *CheckCommandConfig {
+	return &CheckCommandConfig{
 		datastore:     datastore,
 		checkResolver: checkResolver,
 		options:       options,
 	}
-}
-
-// CheckCommandServerConfig holds the full server configuration including shadow mode settings.
-type CheckCommandServerConfig struct {
-	checkCfg  CheckCommandConfig
-	shadowCfg ShadowCheckCommandConfig
-}
-
-type CheckCommandServerConfigOption func(*CheckCommandServerConfig)
-
-func WithShadowCheckCommandConfig(shadowCfg ShadowCheckCommandConfig) CheckCommandServerConfigOption {
-	return func(c *CheckCommandServerConfig) {
-		c.shadowCfg = shadowCfg
-	}
-}
-
-// NewCheckCommandServerConfig creates a new CheckCommandServerConfig.
-func NewCheckCommandServerConfig(main CheckCommandConfig, opts ...CheckCommandServerConfigOption) CheckCommandServerConfig {
-	serverConfig := CheckCommandServerConfig{
-		checkCfg: main,
-	}
-
-	for _, opt := range opts {
-		opt(&serverConfig)
-	}
-	return serverConfig
-}
-
-// NewCheckCommand creates a CheckCommand that automatically handles shadow mode
-// based on the server configuration. This is the main entry point that should be used
-// throughout the codebase instead of the individual factory functions.
-func NewCheckCommand(serverConfig CheckCommandServerConfig, params CheckCommandParams) CheckCommand {
-	return newCheckCommandWithShadowConfig(serverConfig.checkCfg, serverConfig.shadowCfg, params)
 }
 
 type CheckQueryOption func(*CheckQuery)
@@ -134,10 +100,10 @@ func WithCheckDatastoreThrottler(threshold int, duration time.Duration) CheckQue
 	}
 }
 
-func NewCheckQuery(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, params CheckCommandParams, opts ...CheckQueryOption) *CheckQuery {
+func NewCheckQuery(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, opts ...CheckQueryOption) *CheckQuery {
 	cmd := &CheckQuery{
-		params:               params,
 		logger:               logger.NewNoopLogger(),
+		datastore:            datastore,
 		checkResolver:        checkResolver,
 		maxConcurrentReads:   defaultMaxConcurrentReadsForCheck,
 		shouldCacheIterators: false,
@@ -151,48 +117,30 @@ func NewCheckQuery(datastore storage.RelationshipTupleReader, checkResolver grap
 		opt(cmd)
 	}
 
-	var ctxTupleKeys []*openfgav1.TupleKey
-	if params.ContextualTuples != nil {
-		ctxTupleKeys = params.ContextualTuples.GetTupleKeys()
-	}
-	cmd.datastoreWithTupleCache = storagewrappers.NewRequestStorageWrapperWithCache(
-		datastore,
-		ctxTupleKeys,
-		&storagewrappers.Operation{
-			Method:            apimethod.Check,
-			Concurrency:       cmd.maxConcurrentReads,
-			ThrottleThreshold: cmd.datastoreThrottleThreshold,
-			ThrottleDuration:  cmd.datastoreThrottleDuration,
-		},
-		storagewrappers.DataResourceConfiguration{
-			Resources:     cmd.sharedCheckResources,
-			CacheSettings: cmd.cacheSettings,
-		},
-	)
 	return cmd
 }
 
-func (c *CheckQuery) Execute(ctx context.Context) (*graph.ResolveCheckResponse, *graph.ResolveCheckRequestMetadata, error) {
-	err := validateCheckRequest(c.params.Typesys, c.params.TupleKey, c.params.ContextualTuples)
+func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*graph.ResolveCheckResponse, *graph.ResolveCheckRequestMetadata, error) {
+	err := validateCheckRequest(params.Typesys, params.TupleKey, params.ContextualTuples)
 	if err != nil {
 		return nil, nil, err
 	}
 
 	cacheInvalidationTime := time.Time{}
 
-	if c.params.Consistency != openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
-		cacheInvalidationTime = c.sharedCheckResources.CacheController.DetermineInvalidationTime(ctx, c.params.StoreID)
+	if params.Consistency != openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
+		cacheInvalidationTime = c.sharedCheckResources.CacheController.DetermineInvalidationTime(ctx, params.StoreID)
 	}
 
 	resolveCheckRequest, err := graph.NewResolveCheckRequest(
 		graph.ResolveCheckRequestParams{
-			StoreID:                   c.params.StoreID,
-			TupleKey:                  tuple.ConvertCheckRequestTupleKeyToTupleKey(c.params.TupleKey),
-			Context:                   c.params.Context,
-			ContextualTuples:          c.params.ContextualTuples,
-			Consistency:               c.params.Consistency,
+			StoreID:                   params.StoreID,
+			TupleKey:                  tuple.ConvertCheckRequestTupleKeyToTupleKey(params.TupleKey),
+			Context:                   params.Context,
+			ContextualTuples:          params.ContextualTuples,
+			Consistency:               params.Consistency,
 			LastCacheInvalidationTime: cacheInvalidationTime,
-			AuthorizationModelID:      c.params.Typesys.GetAuthorizationModelID(),
+			AuthorizationModelID:      params.Typesys.GetAuthorizationModelID(),
 		},
 	)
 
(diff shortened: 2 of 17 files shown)


cikasfm:
@coderabbitai review

cikasfm:
@coderabbitai review

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 - Add planner for selecting check resolution strategies based on runtime statistics, behind the `enable-check-optimization` flag. [#2624](https://github.com/openfga/openfga/pull/2624)
 - Add `server.WithShadowCheckCacheEnabled` to enable creation of a separate cache for shadow check resolver. [#2683](https://github.com/openfga/openfga/pull/2683)
 - Run weight 2 optimization for cases where there are more than 1 directly assignable userset. [#2684](https://github.com/openfga/openfga/pull/2684)
+- Add datastore query count in shadow check resolver. [#2698](https://github.com/openfga/openfga/pull/2698)
```
adriantam: Please move this to unreleased section

pkg/server/commands/check_command.go
```diff
@@ -27,25 +27,87 @@ const (
...
+	checkCfg  CheckCommandConfig
+	shadowCfg ShadowCheckCommandConfig
+}
+
+type CheckCommandServerConfigOption func(*CheckCommandServerConfig) bool
+
+func WithShadowCheckCommandConfig(shadowCfg ShadowCheckCommandConfig) CheckCommandServerConfigOption {
+	return func(c *CheckCommandServerConfig) bool {
+		c.shadowCfg = shadowCfg
+		return true
```
adriantam: this is a setter function, why should it return true?

pkg/server/commands/check_command.go
```diff
@@ -27,25 +27,84 @@ const (
...
+	for _, opt := range opts {
+		opt(&serverConfig)
+	}
+	return serverConfig
+}
+
+// NewCheckCommandWithServerConfig creates a CheckCommand that automatically handles shadow mode
+// based on the server configuration. This is the main entry point that should be used
+// throughout the codebase instead of the individual factory functions.
+// TODO : replace public NewCheckCommand with this.
```
justincoh: still a TODO?

pkg/server/commands/check_command_shadow.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// only used for testing signals
+	wg      sync.WaitGroup
+	runSync bool // if true, run shadow query synchronously (used for testing)
+}
+
+type ShadowCheckCommandConfig struct {
+	enabled       bool          // A boolean indicating whether shadow mode is enabled.
+	shadowPct     int           // An integer representing the shadowPct of list_objects requests that will also trigger the shadow query. This allows for controlled rollout and data collection without impacting all requests. Value should be between 0 and 100.
+	shadowTimeout time.Duration // A time.Duration specifying the maximum amount of time to wait for the shadow list_objects query to complete. If the shadow query exceeds this shadowTimeout, it will be cancelled, and its result will be ignored, but the shadowTimeout event will be logged.
```
justincoh: nit: `list_objects` --> `check`

pkg/server/commands/check_command_test.go
```diff
@@ -293,3 +308,523 @@ func TestCheckCommandErrorToServerError(t *testing.T) {
...
+			zap.String("model_id", params.Typesys.GetAuthorizationModelID()),
+			zap.String("function", ShadowCheckQueryFunction),
+			zap.Bool("main", true), // Bug: should be true (response.GetAllowed()) but uses shadowRes.GetAllowed()
+			zap.Bool("main_cycle", false),
+			gomock.Any(),                              // main_latency - we can't predict exact timing
+			zap.Uint32("main_query_count", uint32(0)), // FIXME: mock metadata
+			zap.Bool("shadow", false),
+			zap.Bool("shadow_cycle", false),
+			gomock.Any(),                                // shadow_latency - we can't predict exact timing
+			zap.Uint32("shadow_query_count", uint32(0)), // FIXME: mock metadata
```
justincoh: there's a number of `FIXME` and `Bug` comments in this file, just tagging them to make sure they're not being forgotten

pkg/server/commands/list_objects_test.go
```diff
@@ -471,7 +487,164 @@ func TestErrorInCheckSurfacesInListObjects(t *testing.T) {
...
+		zap.String("function", ShadowCheckQueryFunction),
+		zap.Bool("main", true), // Bug: should be true (response.GetAllowed()) but uses shadowRes.GetAllowed()
+		zap.Bool("main_cycle", false),
+		gomock.AssignableToTypeOf(zap.Int64("", 0)), // main_latency - we can't predict exact timing
+		zap.Uint32("main_query_count", uint32(0)),   // FIXME: mock metadata
+		zap.Bool("shadow", false),
+		zap.Bool("shadow_cycle", false),
+		gomock.AssignableToTypeOf(zap.Int64("", 0)), // shadow_latency - we can't predict exact timing
+		zap.Uint32("shadow_query_count", uint32(0)), // FIXME: mock metadata
+	)
```
justincoh, line 581: I don't think we need to assert on the log fields again in here. It's enough to assert that the shadow checker ran, the log statement assertions are covered by the check test file

internal/shared/shared.go
```diff
@@ -31,22 +31,13 @@ func WithCacheController(cacheController cachecontroller.CacheController) Shared
...
-}
-
 // SharedDatastoreResources contains resources that can be shared across Check requests.
 type SharedDatastoreResources struct {
 	SingleflightGroup     *singleflight.Group
 	WaitGroup             *sync.WaitGroup
 	ServerCtx             context.Context
 	CheckCache            storage.InMemoryCache[any]
 	CacheController       cachecontroller.CacheController
-	ShadowCheckCache      storage.InMemoryCache[any]
```
adriantam, line 48: @elbuo8 @yissellokta @senojj - I know that we have agree to remove the in-memory cache.  Want to double confirm whether it can potentially cause any problem if the shadow resolve causes cache to store the wrong value?

pkg/server/config/cache_settings.go
```diff
@@ -65,15 +65,3 @@ func (c CacheSettings) ShouldCacheListObjectsIterators() bool {
 func (c CacheSettings) ShouldCreateShadowNewCache() bool {
```
adriantam: Do we still need this function?  (I only see it being used in test)

cikasfm: Removed

pkg/server/commands/check_command.go
```diff
@@ -177,6 +236,9 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
 }
 
 func validateCheckRequest(typesys *typesystem.TypeSystem, tupleKey *openfgav1.CheckRequestTupleKey, contextualTuples *openfgav1.ContextualTupleKeys) error {
+	if typesys == nil {
+		return errors.New("typesystem is required")
```
elbuo8: Why are we adding this? Same could be done to the other parameters.

pkg/server/commands/check_command.go
```diff
@@ -27,25 +27,83 @@ const (
...
 	checkResolver              graph.CheckResolver
-	typesys                    *typesystem.TypeSystem
-	datastore                  storage.RelationshipTupleReader
 	sharedCheckResources       *shared.SharedDatastoreResources
 	cacheSettings              config.CacheSettings
 	maxConcurrentReads         uint32
 	shouldCacheIterators       bool
 	datastoreThrottleThreshold int
 	datastoreThrottleDuration  time.Duration
+	datastoreWithTupleCache    *storagewrappers.RequestStorageWrapper
```
elbuo8: Why is this needed? If anything you can override/wrap the storage. Or not even, you are just passing it in via context to the downstream graph package.

pkg/server/commands/check_command.go
```diff
@@ -27,25 +27,83 @@ const (
...
 }
 
+var _ CheckCommand = (*CheckQuery)(nil)
```
elbuo8, line 47: Fun that _all_ methods within the command pkg are actually named `XQuery` and only Check is called Command... :)

pkg/server/commands/check_command.go
```diff
@@ -27,25 +27,83 @@ const (
...
+func NewCheckCommandConfig(datastore storage.RelationshipTupleReader, checkResolver graph.CheckResolver, options ...CheckQueryOption) CheckCommandConfig {
+	return CheckCommandConfig{
+		datastore:     datastore,
+		checkResolver: checkResolver,
+		options:       options,
+	}
+}
+
+// CheckCommandServerConfig holds the full server configuration including shadow mode settings.
+type CheckCommandServerConfig struct {
```
elbuo8: Server 🤔 ? I mean its a nit but unsure if its appropriate.

pkg/server/commands/check_command.go
```diff
@@ -27,25 +27,83 @@ const (
...
+type CheckCommandServerConfigOption func(*CheckCommandServerConfig)
+
+func WithShadowCheckCommandConfig(shadowCfg ShadowCheckCommandConfig) CheckCommandServerConfigOption {
+	return func(c *CheckCommandServerConfig) {
+		c.shadowCfg = shadowCfg
+	}
+}
+
+// NewCheckCommandServerConfig creates a new CheckCommandServerConfig.
+func NewCheckCommandServerConfig(main CheckCommandConfig, opts ...CheckCommandServerConfigOption) CheckCommandServerConfig {
```
elbuo8: Why value rather than pointer like everywhere else? If its more performance/nicer to GC, make a subsequent PR were everything changes uniformly but introduce it as the rest of the things.

pkg/server/commands/list_objects.go
```diff
@@ -178,15 +178,10 @@ func WithListObjectsOptimizationsEnabled(enabled bool) ListObjectsQueryOption {
...
-func WithListObjectsUseShadowCache(useShadowCache bool) ListObjectsQueryOption {
-	return func(d *ListObjectsQuery) {
-		d.useShadowCache = useShadowCache
-	}
-}
-
 func NewListObjectsQuery(
 	ds storage.RelationshipTupleReader,
 	checkResolver graph.CheckResolver,
+	checkCommandServerConfig CheckCommandServerConfig, // required
```
elbuo8:
```suggestion
	checkCommandServerConfig CheckCommandServerConfig,
```

pkg/server/commands/batch_check_command.go
```diff
@@ -125,6 +97,11 @@ func NewBatchCheckCommand(datastore storage.RelationshipTupleReader, checkResolv
 }
 
 func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckCommandParams) (map[CorrelationID]*BatchCheckOutcome, *BatchCheckMetadata, error) {
+	if params.Typesys == nil {
```
elbuo8: Why?

pkg/server/config/config.go
```diff
@@ -39,8 +39,6 @@ const (
 	DefaultCheckQueryCacheEnabled = false
 	DefaultCheckQueryCacheTTL     = 10 * time.Second
 
-	DefaultShadowCheckCacheEnabled = false
```
elbuo8, line 42: How are we configuring this now?

pkg/server/commands/check_command.go
cikasfm: This was existing just moved from `Execute` to `NewCommand`, so it had to be in the `Command` itself.

pkg/server/commands/check_command.go
cikasfm: Meaning server level config, that's common to all commands; vs just command config. Suggestions for a better name?

pkg/server/config/config.go
cikasfm: We're removing the "shadow check cache" config concept; i.e. the shadow command will always have it's own cache.

pkg/server/commands/batch_check_command.go
cikasfm: I think it was a suggestion by `coderabbitai`, just a nil check

pkg/server/commands/check_command.go
cikasfm: I think it was a suggestion by `coderabbitai`, just a nil check

pkg/server/config/config.go
```diff
@@ -58,10 +56,6 @@ const (
 	DefaultShadowCheckSamplePercentage = 10
 	DefaultShadowCheckResolverTimeout  = 1 * time.Second
 
-	DefaultShadowListObjectsCheckResolverEnabled  = false
```
elbuo8, line 61: So whatever % we decide will be applied for both Check and the internal Check calls for LO correct?

pkg/server/commands/batch_check_command.go
elbuo8: Remove it.

pkg/server/config/config.go
cikasfm: Yes, correct

justincoh:
closing this PR for now as this work has been deprioritized

PR #2700, opened 2025-09-11
fix: flaky TestConstructUserset unit test

adriantam:
There is no guarantee order for GetGroupedEdges.  As such, the unit test should not make any assumption in the order.

See flaky test for TestConstructUserset (operator_intersection).  Root cause is that [GetGroupedEdges](https://github.com/openfga/openfga/blob/92a663b9edddb64a0d8f1c2cc69b4785859ba1e1/pkg/typesystem/weighted_graph.go#L88) does not guarantee any order.  Hence, unit test should not make assumption in the order.

#### How is it being solved?
Update unit test to test for membership and length instead of the exact order.

Update unit test.

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Improved robustness of userset-related tests to reduce brittleness and better validate expected behavior.
  * Tests now verify outcomes without relying on exact internal structures, making them more resilient to non-functional changes.
* **User Impact**
  * No user-facing changes; functionality and behavior remain the same.
* **Chores**
  * Internal test maintenance to enhance reliability of the test suite.

### DIFF
--- a/pkg/typesystem/weighted_graph_test.go
+++ b/pkg/typesystem/weighted_graph_test.go
@@ -821,26 +821,23 @@ func TestConstructUserset(t *testing.T) {
 		require.Equal(t, graph.RewriteEdge, dut.GetEdgeType())
 		userset, err := typeSystem.ConstructUserset(dut, "user")
 		require.NoError(t, err)
-		require.Equal(t, &openfgav1.Userset{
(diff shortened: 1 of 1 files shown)


elbuo8:
Was literally bitten by this now.

PR #2701, opened 2025-09-11
chore: planner adjustments

elbuo8:

Strategies now have "weight" which help the planner to best adjust decisions. Will not make this configurable at the current time as its still experimental.

Penalize heavily when a request times out. 

Removed initial guess configuration as a global value is not enough to make accurate selections.

## Review Checklist

## Summary by CodeRabbit

- Bug Fixes
  - Refined cancellation handling so only deadline timeouts count toward over-guess metrics, improving accuracy of performance reporting.
- Refactor
  - Adjusted initial Thompson sampling parameterization for more stable initial behavior.
- Tests
  - Added test to verify eviction of stale planner keys while preserving recently accessed entries.


internal/planner/planner.go
```diff
@@ -20,17 +19,23 @@ type Planner struct {
 	stopCleanup chan struct{}
 }
 
+type KeyPlanStrategy struct {
+	Type         string
+	InitialGuess time.Duration
```
yissellokta, line 33: Add this comment:
In Thompson Sampling, the algorithm models its belief about each strategy's performance as a probability distribution (a "belief curve"). The InitialGuess is the value that the algorithm assumes is the most likely average latency for a strategy it knows nothing about. To choose a strategy, the algorithm doesn't just use the average; it takes a random sample (a "draw") from each strategy's full belief curve. Before any data is collected, these draws will be statistically centered around the InitialGuess. A strategy with a low InitialGuess will have its draws centered around a low number, making it more likely to be chosen initially. Once a strategy is chosen and its actual performance is measured, that new data point is used to update the belief curve via a Bayesian update. If a strategy with an InitialGuess of 50ms consistently performs at 10ms, the curve's center will quickly "move" from 50ms down to 10ms. This is how the algorithm learns and adapts.

internal/planner/planner.go
```diff
@@ -20,17 +19,23 @@ type Planner struct {
...
+	Lambda       float64
```
yissellokta: Add this comment: 
This represents our confidence in the InitialGuess. How many good runs do we believe we've effectively seen already? 
- A low Lambda (e.g., 1) means we have very little confidence. The model is "open-minded" and will quickly change its beliefs based on the first few real results. This encourages exploration. 
-  medium lambda: 5, means a modest trust
- A high Lambda (e.g., 10) means we are very confident. The model is "stubborn" and will require a lot of contradictory data to move away from its initial guess. This encourages exploitation.

internal/planner/planner.go
```diff
@@ -20,17 +19,23 @@ type Planner struct {
...
+	Alpha        float64
```
yissellokta: Add this comment:
These two parameters (alpha and beta) work together to define our belief about the consistency (or variance) of the strategy's performance. They describe the shape of the curve, not just its center, like the initial guess. 
Alpha is related to the number of observations about consistency.
- Higher α (for the same β) → higher expected precision → lower variance. For example, α = 20,  β = 2 create a tall, narrow distribution. This means we are very confident that the performance will be extremely consistent and tightly clustered around the average.
- Low Alpha and Beta values (e.g., α ≤ 1,α = 0.5) create a wide, flat distribution. This means we are very uncertain and expect performance to be highly variable, and this causes more exploration

internal/planner/planner.go
```diff
@@ -20,17 +19,23 @@ type Planner struct {
...
+	Beta         float64
```
yissellokta: Add this comment:
Beta is related to the amount of variation seen in those observations. 
- Higher β (for the same α) → lower precision → higher variance.

In General: 
- If we expect the strategy to have bursty behaviour, keep α small (2–3) and/or β larger.
- If we expect the strategy to have tight behaviour, push α up and/or β down.
- If we don't have an idea, α≤1 keeps you very agnostic (heavy-tailed).

internal/graph/default_resolver.go
```diff
@@ -4,19 +4,41 @@ import (
...
 	"github.com/openfga/openfga/pkg/typesystem"
 )
 
 const defaultResolver = "default"
 
+var defaultPlan = &planner.KeyPlanStrategy{
+	Type:         defaultResolver,
+	InitialGuess: 50 * time.Millisecond,
+	// Very Low Lambda: Represents zero confidence. It's a pure guess.
+	Lambda: 1,
```
yissellokta, line 26: Provide a better explanation: With α = 0.5 ≤ 1, it means maximum uncertainty about variance; with λ = 1, we also have weak confidence in the mean. These values will encourage strong exploration of other strategies. Having these values for the default weight 2 strategy helps to enforce the usage of the fast path strategy or more optimal strategies, helping out with the cold start when we don't have enough data.

internal/graph/default_resolver.go
```diff
@@ -4,19 +4,41 @@ import (
...
+	Alpha:  0.5,
+	Beta:   0.5,
+}
+
+var defaultRecursivePlan = &planner.KeyPlanStrategy{
+	Type:         defaultResolver,
+	InitialGuess: 300 * time.Millisecond,
+	// Very Low Lambda: Represents zero confidence. It's a pure guess.
+	Lambda: 1,
```
yissellokta, line 38: Same as above, but with a different mean (300)

internal/graph/recursive_resolver.go
```diff
@@ -21,6 +23,19 @@ import (
...
 
+var recursivePlan = &planner.KeyPlanStrategy{
+	Type:         recursiveResolver,
+	InitialGuess: 150 * time.Millisecond,
+	// Medium Lambda: Represents medium confidence in the initial guess. It's like
+	// starting with the belief of having already seen 5 good runs.
+	Lambda: 5.0,
+	// UNCERTAINTY ABOUT CONSISTENCY: The gap between p50 and p99 is large.
+	// Low Alpha/Beta values create a wider belief curve, telling the planner
+	// to expect and not be overly surprised by performance variations.
```
yissellokta, line 36: Add more info: 
- Low expected precision: 𝐸[𝜏]= 𝛼/𝛽 = 2.0/2.5 = 0.8
- High expected variance: E[σ2]= β/(α−1) =2.5/1 = 2.5, this will allow for relative bursty / jiterry results
- Wide tolerance for spread: 𝛼 = 2, this will allow for considerable uncertainty in how spike the latency can be.
When β > α, we expect lower precision and higher variance

internal/graph/recursive_resolver.go
```diff
@@ -21,6 +23,19 @@ import (
 
 const recursiveResolver = "recursive"
 
+var recursivePlan = &planner.KeyPlanStrategy{
```
yissellokta, line 28: In general these values tell the query planner that the recursive strategy usually performs around 150 ms but occasionally spikes. However, even when it spikes we want to keep it using it or exploring it despite variance, rather than over-penalizing single slow runs.

internal/graph/weight_two_resolver.go
```diff
@@ -23,6 +25,20 @@ const BaseIndex = 0
...
+	Type:         weightTwoResolver,
+	InitialGuess: 20 * time.Millisecond,
+	// High Lambda: Represents strong confidence in the initial guess. It's like
+	// starting with the belief of having already seen 10 good runs.
+	Lambda: 10.0,
+	// High Alpha, Low Beta: Creates a very NARROW belief about variance.
+	// This tells the planner: "I am very confident that the performance is
+	// consistently close to 10ms". A single slow run will be a huge surprise
+	// and will dramatically shift this belief.
+	Alpha: 20,
```
yissellokta, line 43: Add more info:
High expected precision: 𝐸[𝜏]= 𝛼/𝛽 = 20/2 = 10
Low expected variance: E[σ2]= β/(α−1) =2/9 = 0.105, narrow jitter
A slow sample will look like an outlier and move the posterior noticeably but overall this prior exploits

internal/graph/weight_two_resolver.go
```diff
@@ -23,6 +25,20 @@ const BaseIndex = 0
 const DifferenceIndex = 1
 const weightTwoResolver = "weight2"
 
+var weight2Plan = &planner.KeyPlanStrategy{
```
yissellokta, line 29: This strategy is configured to show that it has proven fast and consistent.

internal/graph/check.go
```diff
@@ -53,6 +53,7 @@ type checkOutcome struct {
 type LocalChecker struct {
 	delegate             CheckResolver
 	concurrencyLimit     int
+	upstreamTimeout      time.Duration
```
adriantam, line 56: Will be good to call out here that the upstreamTimeout is only used for calculating strategy plan's time in case of timeout.  Otherwise, reader may initially think the context timeout is set here..

PR #2708, opened 2025-09-19
feat: Add default OpenFGA feature Provider

justincoh:
## Review Checklist


cikasfm (commented):
@justincoh I like this, but I would prefer to have some abstraction/wrapper around the `openfeature.Client`; i.e. like we have one for cache.

So that we could use some feature flag service/provider in our code everywhere and would not depend on the `openfeature` implementation; also could replace the implementation with anything else by just reimplementing the wrapper service, instead of modifying everything in the source code.

Also, that would allow simplifying some of the funcs, like

```golang
	checkOptimizationsEnabled := s.featureClient.Boolean(
		context.Background(),
		string(ExperimentalCheckOptimizations),
		false,
		openfeature.EvaluationContext{},
	)
```

could be just 

```golang
checkOptimizationsEnabled := s.featureFlags.Boolean(ExperimentalCheckOptimizations)
```

pkg/server/server.go
```diff
@@ -960,10 +980,16 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		)
 	}
 
+	checkOptimizationsEnabled := s.featureClient.Boolean(
```
elbuo8: This should happen within the commands themselves. Not just 1 time at boot, but every single time. Otherwise its no different than a static config flag.

That will allow the feature to be toggled without having restart the service.

internal/featureflags/provider.go
```diff
@@ -0,0 +1,72 @@
+package featureflags
+
+import (
+	"context"
+
+	"github.com/open-feature/go-sdk/openfeature"
+)
+
```
elbuo8: Where is the interface? The provider being shipped here is a default, but openfga doesn't enforce any particular clients, thus an interface must be provided so that clients can be injected.

pkg/server/server.go
```diff
@@ -176,6 +177,8 @@ type Server struct {
 	AccessControl                    serverconfig.AccessControlConfig
 	AuthnMethod                      string
 	serviceName                      string
+	featureProvider                  openfeature.FeatureProvider
```
elbuo8: These should be interfaces.

justincoh: FeatureProvider is an interface ([this one](https://github.com/open-feature/go-sdk/blob/main/openfeature/provider.go#L49)). The `openfeature.Client` below is a specific type which could equally be expressed as [openfeature.IClient](https://github.com/open-feature/go-sdk/blob/main/openfeature/client.go#L52) (which is an interface).

I may end up changing both of these shortly though so hold that thought

lukas-reining:
Hey, I am part of the OpenFeature project. 
It is great to see you building on our spec!
Right now, there is no way to change the flags at runtime, correct? 

Do you consider any way to use other providers than `Provider` to enable dynamic flags?
That would be a great usage of our SDK! Maybe having optional [OFREP provider](https://github.com/open-feature/go-sdk-contrib/tree/main/providers/ofrep) support could be a great start, then you could use [`flagd`](https://github.com/open-feature/flagd) and [others](https://openfeature.dev/ecosystem?instant_search%5BrefinementList%5D%5Btype%5D%5B0%5D=OFREP%20API) by just providing an OFREP URL and optionally credentials.

You could use the [`MultiProvider`](https://github.com/open-feature/go-sdk/pull/354) (once it is merged) to consume flags from command line and OFREP together with priorities set as needed.

justincoh:
Hey @lukas-reining, thanks for reaching out!

> Right now, there is no way to change the flags at runtime, correct?

That is correct. With this initial PR we're focused on maintaining existing functionality: static runtime flags.

> Do you consider any way to use other providers than Provider to enable dynamic flags?

We are considering them indeed! For now, this PR will enable OpenFGA users to pass in their own `Provider`, which they could use to implement dynamic features. We'll keep an eye on community usage and feedback (and also on OpenFeature!) to see where we should go from here.

pkg/server/server.go
justincoh: Making that change in [a followup PR here](https://github.com/openfga/openfga/pull/2715)

internal/featureflags/client.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+type defaultClient struct {
+	flags map[string]any
+}
+
+// NewDefaultClient creates a default feature flag client which takes in a static list of enabled feature flag names
+// and stores them as keys in a map.
+func NewDefaultClient(flags []string) Client {
+	enabledFlags := make(map[string]any)
```
adriantam:
```suggestion
	enabledFlags := make(map[string]any, len(flags))
```

nitpike: preallocate size

pkg/featureflags/client_test.go
adriantam, line 1: Mind changing this to use github.com/stretchr/testify/require?

This will make some of the test easier...
for example, instead of 
```
		if !result {
			t.Errorf("expected true, got %t", result)
		}
```
You can just use `require.True(t, result)`

pkg/server/config/config.go
```diff
@@ -102,6 +104,10 @@ const (
 	DefaultPlannerInitialGuess      = 10 * time.Millisecond
 	DefaultPlannerEvictionThreshold = 0
 	DefaultPlannerCleanupInterval   = 0
+
+	ExperimentalCheckOptimizations       ExperimentalFeatureFlag = "enable-check-optimizations"
+	ExperimentalListObjectsOptimizations ExperimentalFeatureFlag = "enable-list-objects-optimizations"
+	ExperimentalAccessControlParams      ExperimentalFeatureFlag = "enable-access-control"
```
adriantam: @rhamzeh - do we still want to keep access control as experimental?  Or we should consider moving this to be a stable feature?

(Not blocking for this work)

rhamzeh: For now yes - not b/c the feature itself is experimental, but b/c we have not yet thought of a good DX around it for system maintainers.

Making the following simple and with a better DX
- Initial setup
- Updating the model
- Ongoing maintenance (granting and removing access)

adriantam:
@coderabbitai review

internal/graph/check.go
```diff
@@ -761,7 +770,7 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
 			}
 			defer iter.Stop()
 
-			if !c.optimizationsEnabled {
+			if !c.ff.Boolean(serverconfig.ExperimentalCheckOptimizations, false, nil) {
```
elbuo8: In `Check`, we should retain `optimizationsEnabled` as a struct setting.

The inspection of the flag should happen right before calling _ResolveCheck_ at the controller. The reason being, we want a stable Check resolution. Don't want some nodes being checked with optimizations and others not in the case of a flip. Could lead to unstable results (or if the FF service is down/serving stale)

pkg/featureflags/client.go
```diff
@@ -0,0 +1,44 @@
+package featureflags
+
+type Client interface {
+	Boolean(flagName string, defaultValue bool, featureCtx map[string]any) bool
```
elbuo8: Why do we have to pass in `defaultValue`? This would mean having to set it in all places and when swapping the actual default value having to search everywhere the flag is used which could lead to missed swapps. The client should know the default value on its instantiation.

pkg/featureflags/client.go
```diff
@@ -0,0 +1,44 @@
...
+type defaultClient struct {
+	flags map[string]any
+}
+
+// NewDefaultClient creates a default feature flag client which takes in a static list of enabled feature flag names
+// and stores them as keys in a map.
+func NewDefaultClient(flags []string) Client {
+	enabledFlags := make(map[string]any, len(flags))
+	for _, flag := range flags {
+		enabledFlags[flag] = struct{}{}
```
elbuo8, line 16: Rather than a struct, store the default value 🤔

pkg/featureflags/client.go
elbuo8: Same logic applies to LO.

justincoh: defaultValue here is the fallback in case a flag isn't configured, or in case a Client has an error. That would be up to users of OpenFGA to handle appropriately though, and maybe we should just leave it up to whoever is writing their own client. Also, in the case of a Boolean flag you could argue that the default value is always `false`, otherwise there would be no flag to check.

I think I'll change this to `Boolean(flagName string, featureCtx map[string]any)`

pkg/featureflags/client.go
justincoh: With the default provider we're just reading in the enabled experimental flags. So just the fact that `enable_check_optimizations` exists means that it is active; an inactive flag won't be present in the map.

This is related to [the comment above](https://github.com/openfga/openfga/pull/2708#discussion_r2379599309), where a Boolean flag could always be assumed to have a default of `false`—I think making that change will solve both problems.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Added
+- Added a default featureflag.Client to the server and ability to pass custom clients to the server via `server.WithFeatureFlagClient`. [#2708](https://github.com/openfga/openfga/pull/2708)
```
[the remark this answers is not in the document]
adriantam: Agree

pkg/server/commands/list_objects_shadow.go
```diff
@@ -114,14 +115,14 @@ func newShadowedListObjectsQuery(
...
+		slices.Concat(opts, []ListObjectsQueryOption{WithFeatureFlagClient(featureflags.NewHardcodedBooleanClient(false))})...,
 	)
 	if err != nil {
 		return nil, err
 	}
 	optimized, err := NewListObjectsQuery(ds, checkResolver,
 		// enable optimizations
-		slices.Concat(opts, []ListObjectsQueryOption{WithListObjectsUseShadowCache(true), WithListObjectsOptimizationsEnabled(true)})...,
+		slices.Concat(opts, []ListObjectsQueryOption{WithFeatureFlagClient(featureflags.NewHardcodedBooleanClient(true))})...,
 	)
```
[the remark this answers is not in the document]
justincoh: In this specific case I think clobbering is appropriate. This behavior isn't net-new in this PR, it's how it was being done before as well.

pkg/server/commands/list_objects.go
```diff
@@ -172,15 +173,9 @@ func WithListObjectsDatastoreThrottler(threshold int, duration time.Duration) Li
...
-		d.optimizationsEnabled = enabled
-	}
-}
-
-func WithListObjectsUseShadowCache(useShadowCache bool) ListObjectsQueryOption {
-	return func(d *ListObjectsQuery) {
-		d.useShadowCache = useShadowCache
+		d.ff = client
 	}
 }
```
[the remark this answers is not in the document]
adriantam: agree to improve code safety

pkg/server/list_objects.go
```diff
@@ -88,7 +88,7 @@ func (s *Server) ListObjects(ctx context.Context, req *openfgav1.ListObjectsRequ
 		commands.WithMaxConcurrentReads(s.maxConcurrentReadsForListObjects),
 		commands.WithListObjectsCache(s.sharedDatastoreResources, s.cacheSettings),
 		commands.WithListObjectsDatastoreThrottler(s.listObjectsDatastoreThrottleThreshold, s.listObjectsDatastoreThrottleDuration),
-		commands.WithListObjectsOptimizationsEnabled(s.IsExperimentallyEnabled(ExperimentalListObjectsOptimizations)),
+		commands.WithFeatureFlagClient(s.featureFlagClient),
 	)
```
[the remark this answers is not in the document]
justincoh: We are already doing this in server.go on line 914

internal/graph/check.go
```diff
@@ -761,7 +761,9 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
 			}
 			defer iter.Stop()
 
-			if !c.optimizationsEnabled {
+			// Optimizations can be enabled for a specific request OR for the checker itself (like in shadow tests),
```
Vic-Dev, line 764: Is this a bug fix? Would it be possible to extract this logic to a separate PR?

lukas-reining:
> Hey @lukas-reining, thanks for reaching out!
> 
> > Right now, there is no way to change the flags at runtime, correct?
> 
> That is correct. With this initial PR we're focused on maintaining existing functionality: static runtime flags.
> 
> > Do you consider any way to use other providers than Provider to enable dynamic flags?
> 
> We are considering them indeed! For now, this PR will enable OpenFGA users to pass in their own `Provider`, which they could use to implement dynamic features. We'll keep an eye on community usage and feedback (and also on OpenFeature!) to see where we should go from here.

Okay this makes sense @justincoh!

> We are considering them indeed! For now, this PR will enable OpenFGA users to pass in their own `Provider`, which they could use to implement dynamic features. We'll keep an eye on community usage and feedback (and also on OpenFeature!) to see where we should go from here.

Okay so for now, users would have to build from source to enable other providers right? 
If you have any questions regarding `OFREP` just reach out to us :)

internal/graph/resolve_check_request.go
```diff
@@ -204,3 +208,10 @@ func (r *ResolveCheckRequest) GetInvariantCacheKey() string {
 	}
 	return r.invariantCacheKey
 }
+
+func (r *ResolveCheckRequest) GetOptimizationsEnabled() bool {
```
elbuo8: Why define it if not going to use it? 

Also, in which case would `r` be nil? Wouldn't panic before even getting to the if 🤔

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -176,7 +176,7 @@ func WithCheckResolver(resolver graph.CheckResolver) ReverseExpandQueryOption {
 	}
 }
 
-func WithListObjectOptimizationsEnabled(enabled bool) ReverseExpandQueryOption {
+func WithListObjectsOptimizationsEnabled(enabled bool) ReverseExpandQueryOption {
```
elbuo8: Not using the same pattern as Check?

internal/graph/resolve_check_request.go
justincoh: > Why define it if not going to use it?

Is used in `.clone()` above.

> Also, in which case would r be nil? Wouldn't panic before even getting to the if

I am not sure, I just followed the existing pattern in this file

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: updated to make it consistent with check

justincoh:
Converting to draft while I get this back up to date

pkg/server/commands/list_objects_test.go
```diff
@@ -608,9 +609,9 @@ func runOneBenchmark(
 ) {
 	if optimizationsEnabled {
 		name += "_with_optimization"
-		query.optimizationsEnabled = true
+		query.ff = featureflags.NewHardcodedBooleanClient(true)
 	} else {
-		query.optimizationsEnabled = false
+		query.ff = featureflags.NewHardcodedBooleanClient(false)
 	}
```
adriantam, line 900: Nitpik (can ignore):

```suggestion

if optimizationsEnabled {
      name += "_with_optimization"
} 

    query.ff = featureflags.NewHardcodedBooleanClient(optimizationsEnabled)
}
```

adriantam:
@coderabbitai review

adriantam (dismissed):

internal/graph/check.go
```diff
@@ -67,12 +66,6 @@ func WithResolveNodeBreadthLimit(limit uint32) LocalCheckerOption {
 	}
 }
 
-func WithOptimizations(enabled bool) LocalCheckerOption {
-	return func(d *LocalChecker) {
-		d.optimizationsEnabled = enabled
-	}
-}
-
```
justincoh, line 75: this was relocated from `pkg graph` into the check command so it can be determined at request time and not at build time

pkg/server/commands/check_command.go
```diff
@@ -38,6 +38,7 @@ type CheckQuery struct {
 	shouldCacheIterators       bool
 	datastoreThrottleThreshold int
 	datastoreThrottleDuration  time.Duration
+	optimizationsEnabled       bool
```
justincoh: this was moved out of `graph` and into `commands` so it can be determined with each request instead of at build time

pkg/server/commands/list_objects.go
```diff
@@ -216,14 +223,19 @@ func NewListObjectsQuery(
...
+		ff:                   featureflags.NewNoopFeatureFlagClient(),
 	}
 
 	for _, opt := range opts {
 		opt(query)
 	}
 
+	if query.ff.Boolean(serverconfig.ExperimentalListObjectsOptimizations, nil) {
+		query.optimizationsEnabled = true
+	}
```
justincoh, line 246: passing in the `ff` instance itself, rather than calling the command `.WithOptimization` as the shadow list objects flow may need to check a _different_ feature flag name

justincoh:
going back to draft pending a rebase on the refactor work here https://github.com/openfga/openfga/pull/2742

PR #2714, opened 2025-09-22
migrate grpc dial context to NewClient

amammay:

Migrate from using using the deprecated grpc dial context/blocking dial option to using `grpc.NewClient`.  I believe the usage of `WithBlock` along with DialContext was used to validate the state of the system before proceeding. Based on the referenced article although...

> The reason these options can be a problem is that connections with a ClientConn are dynamic -- they may come and go over time. If your client successfully connects, the server could go down 1 second later, and your RPCs will fail. "Knowing you are connected" does not tell you much in this regard.
> 
> Additionally, all RPCs created on an "idle" or a "connecting" ClientConn will wait until their deadline or until a connection is established before failing. This means that you don't need to check that a ClientConn is "ready" before starting your RPCs. By default, RPCs will fail if the ClientConn enters the "transient failure" state, but setting WaitForReady(true) on a call will cause it to queue even in the "transient failure" state, and it will only ever fail due to a deadline, a server response, or a connection loss after the RPC was sent to a server.
> 
> Some users of Dial use it as a way to validate the configuration of their system. If you wish to maintain this behavior but migrate to NewClient, you can call GetState, then Connect if the state is Idle and WaitForStateChange until the channel is connected. However, if this fails, it does not mean that your configuration was bad - it could also mean the service is not reachable by the client due to connectivity reasons.

I dont think its necessarily required to have this blocking dial occur in the startup. Mainly because the handling of the gRPC server creation happens right before, any errors related to `net.Listen`/ `grpcServer.Serve(lis)`would of have been handled already or whenever the go routine gets spawned for the server. So i think its an safe assumption that if the server creation hasn't barfed, then the client creation can move forward without blocking.

I would be open to adding in gRPC mitigation reference _if_ wanted.

```go
		timeoutCtx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
		defer cancel()

		if conn.GetState() == connectivity.Idle {
			conn.Connect()
		}

		for {
			state := conn.GetState()
			if state == connectivity.Ready {
				break
			}

			if !conn.WaitForStateChange(timeoutCtx, state) {
				break
			}
		}
		if conn.GetState() != connectivity.Ready {
			return fmt.Errorf("failed to connect to gRPC server")
		}
```

#### How is it being solved?

By following the guidance from [this](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#the-wrong-way-grpcdial) and [this](https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#especially-bad-using-deprecated-dialoptions). To using the gRPC NewClient func.

- Removal of invoking deprecated gRPC dial context func
- Removal of gRPC deprecated `DialOption` of `WithBlock`

## References

- https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#how-to-properly-create-a-clientconn-grpcnewclient

## Review Checklist


adriantam:
@coderabbitai

adriantam:
@coderabbitai review

senojj (approved):

PR #2715, opened 2025-09-23
feat: add WithFeatureFlagClient options to check, list objects

justincoh:
This PR builds on #2708 to make feature flag evaluations dynamic. In #2708 we just implemented the interface and plugged it into OpenFGA server; the "feature flags" themselves remained static values. In this PR we're changing all feature flag calls to be evaluated at runtime, rather than build time.

* Added `WithFeatureFlagClient` to check, list objects, and reverse_expand
* Created `hardcodedBooleanClient` for cases where we want to force enable or disable feature flags, like in shadow resolvers or testing
* Removed extra `ExperimentalFeatureFlag string` type and just use `string`. They're already declared via `const`, the additional type seems unnecessary

## Review Checklist


justincoh:
merged this feature branch into the [main pr here](https://github.com/openfga/openfga/pull/2708)

PR #2717, opened 2025-09-24
Remove zap type conversion on request logger

amammay:

When using `pkg/middleware/logging` both of the public functions accept a `logger.Logger` (which is an interface) but within the implementation there is an unnecessary type conversion back to the zap logger type. This was causing my application to panic when i was adapting a zerolog shim to the api surface exposed.

#### How is it being solved?

By removing the type conversion and removing the call to `ctx: ctxzap.ToContext(ctx, zapLogger.Logger)`.

This can be done because the actual usage the context stored on the reporter struct is just fetching the grpc_tags stored on context via `r.ctx` and **not** the whole logger.  https://github.com/openfga/openfga/blob/v1.10.1/pkg/middleware/logging/logging.go#L66

Code removal

## References

- `ctxzap.ToContext` https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap#ToContext

> ToContext adds the zap.Logger to the context for extraction later. Returning the new context that has been created.

- `ctxzap.TagsToFields` https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap#TagsToFields

> TagsToFields transforms the Tags on the supplied context into zap fields.

## Review Checklist

## Summary by CodeRabbit

- New Features
  - Improved structured logging for gRPC requests, ensuring service name and request metadata are consistently captured.
- Refactor
  - Simplified logging context propagation to reduce overhead while preserving all logged fields and behavior.
- Tests
  - Added an integration test that verifies a single RPC emits a complete JSON log with expected fields (level, message, service, method, user agent, request/response payloads, duration, peer address, request ID, and status code).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ It helps developers easily model and enforce fine-grained access control in thei
 - 🌿 [Terraform Provider](https://github.com/openfga/terraform-provider-openfga) for configuring OpenFGA servers as code
 - 🎮 [Playground](https://openfga.dev/docs/getting-started/setup-openfga/playground) for modeling and testing
 - 🛠 Can also be embedded as a [Go library](https://pkg.go.dev/github.com/openfga/openfga/pkg/server#example-NewServerWithOpts)
-- 🤝 Adopted by [Grafana Labs](https://grafana.com/), [Canonical](https://canonical.com/), [Docker](https://docker.com),  [Agicap](https://agicap.com), [Read.AI](https://read.ai) and [others](https://github.com/openfga/community/blob/main/ADOPTERS.md)
+- 🤝 Adopted by [Auth0](https://fga.dev), [Grafana Labs](https://grafana.com/), [Canonical](https://canonical.com/), [Docker](https://docker.com),  [Agicap](https://agicap.com), [Read.AI](https://read.ai) and [others](https://github.com/openfga/community/blob/main/ADOPTERS.md)
 
 ---
 


pkg/middleware/logging/logging_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	listner := bufconn.Listen(1024 * 1024)
+	srv := grpc.NewServer(serverOpts...)
+	t.Cleanup(srv.Stop)
+
+	openfgav1.RegisterOpenFGAServiceServer(srv, &fgaServer{})
+
+	go func() {
+		err := srv.Serve(listner)
+		require.NoError(t, err)
```
cikasfm: you might want to add a wait group here to wait for the results in your test

cikasfm:
@amammay thank you for your contribution!

> By removing the type conversion and removing the call to ctx: ctxzap.ToContext(ctx, zapLogger.Logger).
> This can be done because the actual usage the context stored on the reporter struct is just fetching the grpc_tags stored on context via r.ctx and not the whole logger. https://github.com/openfga/openfga/blob/v1.10.1/pkg/middleware/logging/logging.go#L66

Good catch! The `logger` is not used and only `Tags` are extracted from the context using `ctxzap.TagsToFields`, so setting the logger in the context can be removed completely.

pkg/middleware/logging/logging_test.go
amammay: done!

cikasfm:
@amammay please update the `CHANGELOG.md` to reflect the change that it potentially can be breaking for some users ( if they expect the `zap` logger to be in the context ) & we should be good to merge.

amammay:
> @amammay please update the `CHANGELOG.md` to reflect the change that it potentially can be breaking for some users ( if they expect the `zap` logger to be in the context ) & we should be good to merge.

Updated. I dont think a note about the behavior change is needed because this `ctx` that was tweaked is different then the `ctx` that is used in the request chain. 

If you look at the signature of `interceptors.CommonReportableFunc` https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors#CommonReportableFunc.ServerReporter the context that is returned from the second position is the one that is used when traversing the chain. 

The one i touched was the one stored on `reporter` struct, https://github.com/openfga/openfga/blob/7b52513c529adbc11f589dcfaa70add774993b05/pkg/middleware/logging/logging.go#L155-L161 which would of been disconnected from the returned one. Since https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/logging/zap/ctxzap#ToContext returns a **new**  context with the zap logger on it.

PR #2718, opened 2025-09-25
fix: order of go mod commands in make deps

sergiught:
## Description

Previously, `make deps` ran:
```bash
go mod vendor && go mod tidy
```
This caused the `vendor/` directory to sometimes contain stale or unused dependencies, since `go mod tidy` was cleaning up after vendoring.

### Change
This PR updates the order to:

```bash
go mod tidy && go mod vendor
```

### Benefits
✅ Ensures go.mod and go.sum are up-to-date before vendoring
✅ Keeps vendor/ consistent with the cleaned module file
✅ Prevents unnecessary/untracked dependencies from being copied into vendor/

## How to test
Run make deps on a fresh checkout.

Confirm `go.mod` and` go.sum` contain no unexpected changes.

Confirm `vendor/` contains only the required dependencies.

## Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ clean: ## Clean project files
 
 deps: ## Download dependencies
 	${call print, "Downloading dependencies"}
-	@go mod vendor && go mod tidy
+	@go mod tidy && go mod vendor
 
 $(GO_BIN)/golangci-lint:
 	${call print, "Installing golangci-lint within ${GO_BIN}"}


Makefile
```diff
@@ -37,7 +37,7 @@ clean: ## Clean project files
 
 deps: ## Download dependencies
 	${call print, "Downloading dependencies"}
-	@go mod vendor && go mod tidy
+	@go mod tidy && go mod vendor
```
rhamzeh, line 40:
```suggestion
	@go mod tidy
```

sergiught: I'm not sure I understand the suggestion, could you please clarify @rhamzeh ? You're suggesting we stop doing go mod vendor?

rhamzeh: We already don't vendor - this is a typo in the makefile. We should remove the `go mod vendor` from here.

sergiught: We still need the packages to be installed locally in a vendor folder to be able to compile the binary, rather than using the go mod cache and bloat the user's system.

rhamzeh: > We still need the packages to be installed locally in a vendor folder to be able to compile the binary, rather than using the go mod cache and bloat the user's system.

That is an opinion not a fact.
> We still need the packages to be installed locally [...] to be able to compile the binary

True, and this works regardless whether they are in the vendor folder or cache

> using the go mod cache and bloat the user's system

I disagree with this - we're making a choice that no one has complained about. And the vendored files _are also cached_. In fact we're doing the worse - we're now bloating, as if the user has 5 projects using the same version of a library, only one copy exists in the cache. What we're doing here is having another copy in the vendor folder.

There is one reason to run go mod vendor that is helpful, and that is when debugging dependencies and iterating on them - but that is something you can run when you do that - not something we have to run every time.

Regardless, I disagree with this but I will not block it on the server as I believe the others working more on the code have the final say.

sergiught: Appreciate the input, @rhamzeh  🙏. But I think we might be mixing two conversations here.

This PR is just keeping the Makefile consistent with the repo’s current setup — which already uses a vendor setup. So the Makefile isn’t inventing that pattern, just following it and fixing an annoying issue encountered by the team. 

The bigger "should we vendor at all?" feels like a repo-wide discussion for the team (maybe via a GitHub issue, please feel free to do so), not something to resolve inside this Makefile change.

saad-h1: For what it's worth we haven't actually committed the `vendor/` directory to openfga, which is kind of the whole point when using vendoring from my understanding, so in that sense we're not using it (I can't think of much benefit to only vendoring locally over using the go mod cache).

I like the suggestion to just get rid of it entirely.

justincoh: @sergiught Bumping this back up your feed. What do you think about just removing `go mod vendor` entirely?

sergiught: Thanks for the ping @justincoh, this flew off my radar for a bit:)

My personal preference is to vendor dependencies locally (but not commit them) so that everything stays within the project folder and doesn't affect the global Go module cache. That cache is shared across all Go projects on a machine, and I'd rather keep things isolated to avoid surprises or extra bloat.

On the point that vendoring only makes sense if you commit the `vendor/` folder, I don’t fully agree. Vendoring can still be useful even if uncommitted, like for offline builds, cleaner CI environments, or just keeping things self-contained.

That said, I’m fine with removing `go mod vendor` from the `Makefile` if the team prefers. Just wanted to share where I was coming from. Appreciate the thoughtful convo here.

---

*Leveraging AI to give you an example of an issue with purely a global go mod cache...*

### 🧨 Example: Two Projects Rely on Different Versions of an Indirect Dependency (Go 1.20+)

**Scenario:**

You have two Go projects:

- `project-a/` depends on `github.com/service/a v1.2.0`
- `project-b/` depends on `github.com/service/a v1.3.0`

Each of these versions pulls in a different transitive dependency:

- `v1.2.0` pulls `github.com/shared/lib v0.9.0`
- `v1.3.0` pulls `github.com/shared/lib v1.0.0`

Due to Go 1.20+ features like **lazy module loading** and **graph pruning**, these transitive differences are easy to miss.

**What can go wrong:**

1. You build `project-a` first, caching `shared/lib@v0.9.0` globally.
2. Then you build `project-b`, expecting `shared/lib@v1.0.0`, but:
   - Go may reuse or misresolve dependencies based on the cached graph.
   - You might get build inconsistencies, version mismatches, or unexpected behavior.
   - Especially if `shared/lib` changed public APIs or build constraints between versions.

So even without any proxy issues or manual changes, you can end up with **order-dependent builds** that behave differently depending on what’s in the cache.

**✅ How vendoring helps:**

Vendoring locks the entire resolved dependency graph per project. Each project is self-contained and isolated from the global cache, avoiding accidental cross-project contamination or dependency mismatches.

senojj: That AI analysis seems... incorrect. The very purpose of the `go.mod` and `go.sum` files is to guarantee that the correct versions of each dependency are used for each build.

senojj: Honestly, my initial reaction to this was "Why is this simple command in the Makefile?"

senojj: Ah, I see... The command is doing more than just pulling Go dependencies. I did not expect it to install other things.

justincoh: > My personal preference is to vendor dependencies locally (but not commit them)

That's fine with me as well. Can you add `/vendor` to the `.gitignore` in this PR?

sergiught: https://github.com/openfga/openfga/pull/2718/changes/82a68ae5e9c980ab03c541ed61867c875e117eef

PR #2722, opened 2025-09-27
Migrate to postgresql 18

Oscmage:


adriantam (dismissed):

jpadilla (changes requested):
@Oscmage Adding something like `chown -R postgres:postgres /var/lib/postgresql` right before this line should be enough to fix the failing tests https://github.com/Oscmage/openfga/blob/postgres-18/pkg/testfixtures/storage/postgres.go#L272

pkg/testfixtures/storage/postgres.go
```diff
@@ -342,7 +345,7 @@ func (p *postgresTestContainer) getMasterContainerID(dockerClient *client.Client
 func (p *postgresTestContainer) configureMasterForReplication(t testing.TB, dockerClient *client.Client, masterContainerID string) error {
 	// Configuration for streaming replication - only pg_hba.conf and reload.
 	commands := [][]string{
-		{"sh", "-c", "echo 'host replication postgres all trust' >> /var/lib/postgresql/data/pg_hba.conf"},
+		{"sh", "-c", "echo 'host replication postgres all trust' >> $PGDATA/pg_hba.conf"},
```
Oscmage, line 348: The location was changed in postgres 18, this variant should be compatible with both.

https://github.com/docker-library/postgres/blob/master/18/bookworm/Dockerfile#L189

See example: https://aronschueler.de/blog/2025/10/30/fixing-postgres-18-docker-compose-startup/

CHANGELOG.md
```diff
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ## [Unreleased]
+### Changed
+- Upgraded PostgreSQL test containers from version 17 to version 18. The data directory structure changed in PostgreSQL 18 (from `/var/lib/postgresql/data` to `/var/lib/postgresql/18/docker`), so `pg_hba.conf` paths now use `$PGDATA` environment variable for compatibility. See [PostgreSQL 18 Docker image documentation](https://github.com/docker-library/postgres/blob/master/18/bookworm/Dockerfile#L189) for details. [#2722](https://github.com/openfga/openfga/pull/2722)
+
 ### Fixed
+- Fixed PostgreSQL replica container permissions issue in tests by ensuring postgres user owns `/var/lib/postgresql` directory before starting PostgreSQL. [#2722](https://github.com/openfga/openfga/pull/2722)
```
jpadilla, line 14: I think we'll want to make this entry more concise.

@adriantam have a suggestion of what this should read?

adriantam:
```
### Changed
- Upgraded PostgreSQL test containers from version 17 to version 18 which introduces data directory structure and updated `pg_hba.conf` paths for compatibility. See [PostgreSQL 18 Docker image documentation](https://github.com/docker-library/postgres/blob/master/18/bookworm/Dockerfile#L189) for details. [#2722](https://github.com/openfga/openfga/pull/2722)
```

How about this?  We don't want to include details of the PostgreSQL changes as it is outside the scope of this project.

PR #2727, opened 2025-10-02
bump the latest language and apply the breaking changes

yissellokta:
Latest language changes introduced the logicalttuGrouping and the logicalDirectGrouping to group direct assignments in a relation or the same ttu definition with multiple usersets. 
This PR will make the required changes in check and listobjects to traverse the graph and take decisions with these two new nodes and edges

Provide a list of any applicable references here (GitHub Issue, [OpenFGA RFC](https://github.com/openfga/rfcs), other PRs, etc..). We prefer an accompanying issue for all non-trivial PRs.
It also removes the complexity of grouping edges of ttu or direct assignment when in the presence of intersection and exclusion. Specifically, allows one to rely again on exclusion operations only having two edges.

## Review Checklist

## Summary by CodeRabbit

* New Features
  * More robust evaluation of complex permission logic involving intersections, exclusions, and recursive relations.
* Bug Fixes
  * Corrected edge cases in recursive and tuple-based paths to prevent incorrect results and cycle-related issues.
* Performance
  * Streamlined internal edge handling to reduce overhead during authorization checks.
* Reliability
  * Improved consistency when resolving logically grouped relationships across usersets.
* Chores
  * Updated an external language module dependency to the latest beta version.


pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -760,20 +758,32 @@ func (c *ReverseExpandQuery) exclusionHandler(
...
 	}
 
 	// This means the exclusion edge does not have a path to the terminal type.
 	// e.g. `B` in `A but not B` is not relevant to this query.
-	if edges.ExcludedEdges == nil {
-		newReq := req.clone()
+	if edges.ExcludedEdge == nil {
 
+		var baseEdges []*weightedGraph.WeightedAuthorizationModelEdge
+		if edges.BaseEdge.GetEdgeType() == weightedGraph.DirectLogicalEdge || edges.BaseEdge.GetEdgeType() == weightedGraph.TTULogicalEdge {
```
elbuo8: Thought there was a method in `language` that did this 🤔

pkg/typesystem/typesystem.go
```diff
@@ -558,15 +559,17 @@ func (t *TypeSystem) TTUUseRecursiveResolver(objectType, relation, userType stri
 		return false
 	}
 
+	// if we are not in the presence of a recursive relation or it is part of a tuple cycle, return false
+	if objRelNode.GetRecursiveRelation() != objRelNode.GetUniqueLabel() || objRelNode.IsPartOfTupleCycle() {
```
elbuo8, line 565:
```suggestion
	if objRelNode.IsPartOfTupleCycle() || objRelNode.GetRecursiveRelation() != objRelNode.GetUniqueLabel() ||  {
```

elbuo8: Should we move this _before_ getting the weight?

elbuo8: We should also modify the documentation of this one and the Userset one.

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
yissellokta: no, which method?

pkg/typesystem/typesystem.go
yissellokta: This will only apply when the weight is infinite

pkg/typesystem/typesystem.go
```diff
@@ -576,31 +579,26 @@ func (t *TypeSystem) TTUUseRecursiveResolver(objectType, relation, userType stri
...
+				if edge.GetEdgeType() == graph.RewriteEdge || edge.GetEdgeType() == graph.TTULogicalEdge {
+					newEdges, okEdge := t.authzWeightedGraph.GetEdgesFromNode(edge.GetTo())
+					if !okEdge {
 						return false
 					}
 					// these edges will need to be evaluated in subsequent iterations
-					innerEdges = append(innerEdges, operationalEdges...)
-					continue
-				} else if edge.GetEdgeType() == graph.TTUEdge && !recursiveTTUFound && edge.GetTuplesetRelation() == tuplesetRelation && strings.HasSuffix(edge.GetTo().GetUniqueLabel(), computedRelationSuffix) && edge.GetTo() == objRelNode {
-					// find and validate the TTUEdge which is infinite (the one being processed at the current time)
```
elbuo8, line 591: I think this block comment shouldn't be lost.

pkg/typesystem/weighted_graph.go
```diff
@@ -243,29 +169,29 @@ func (t *TypeSystem) ConstructExclusionUserset(node *graph.WeightedAuthorization
 		return nil, fmt.Errorf("incorrect exclusion node: %s", node.GetUniqueLabel())
 	}
 	exclusionEdges, err := GetEdgesForExclusion(edges, sourceUserType)
-	if err != nil || len(exclusionEdges.BaseEdges) == 0 {
+	if err != nil || exclusionEdges.BaseEdge == nil {
```
elbuo8: > exclusionEdges.BaseEdge == nil

could this actually happen in this implementation? If the target user type isn't related, shouldn't it just yield error rather than nil in base/difference?

yissellokta: not really, if a base edge don;t lead to the usertype the operation won't have a weight for it

pkg/typesystem/typesystem.go
yissellokta: done

pkg/typesystem/weighted_graph.go
yissellokta: done

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
elbuo8: There isn't, just confirmed.

pkg/typesystem/typesystem.go
elbuo8: Ah right 🤦🏽

yissellokta:
Found an issue in language, will close the PR, until the fix is merged and I will re-open it

yissellokta:
reopening the PR with the latest changes from language

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -551,6 +555,14 @@ func (c *ReverseExpandQuery) findCandidatesForLowestWeightEdge(
 		topItemStack = stack.Push(nil, topItem)
 	}
 
+	var edges []*weightedGraph.WeightedAuthorizationModelEdge
+	if edge.GetEdgeType() == weightedGraph.DirectLogicalEdge || edge.GetEdgeType() == weightedGraph.TTULogicalEdge {
+		logicalEdges, _ := c.typesystem.GetConnectedEdges(edge.GetTo().GetUniqueLabel(), sourceUserType)
```
adriantam: would it make sense to check and return error?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
```diff
@@ -551,6 +555,14 @@ func (c *ReverseExpandQuery) findCandidatesForLowestWeightEdge(
 		topItemStack = stack.Push(nil, topItem)
 	}
 
+	var edges []*weightedGraph.WeightedAuthorizationModelEdge
+	if edge.GetEdgeType() == weightedGraph.DirectLogicalEdge || edge.GetEdgeType() == weightedGraph.TTULogicalEdge {
```
adriantam: Do we want to the following encapsulate logic into a helper function
```
if edge.GetEdgeType() == weightedGraph.DirectLogicalEdge || edge.GetEdgeType() == weightedGraph.TTULogicalEdge {
   return c.typesystem.GetConnectedEdges(edge.GetTo().GetUniqueLabel(), sourceUserType)
else
   return edge, nil
} 
``` 
as I see the exclusionHandler also has similar logic below (except that it uses edges.BaseEdge except edge

pkg/typesystem/typesystem.go
```diff
@@ -558,15 +559,17 @@ func (t *TypeSystem) TTUUseRecursiveResolver(objectType, relation, userType stri
 		return false
 	}
 
+	// if we are not in the presence of a recursive relation or it is part of a tuple cycle, return false
```
adriantam, line 564: Wondering do we need to update the comment at the top of the function

pkg/typesystem/typesystem.go
adriantam, line 549: Do we need to update `TestRecursiveTTUCanFastPathV2` to test the changes in logic here?

pkg/typesystem/weighted_graph.go
```diff
@@ -123,54 +77,19 @@ func GetEdgesForExclusion(
...
-		} else if baseKey != key {
-			exclusionKey = key
-		}
-		groupedEdges[key] = append(groupedEdges[key], edge)
-	}
-
-	if len(groupedEdges) > 2 || len(baseKey) == 0 {
-		return ExclusionEdges{}, fmt.Errorf("invalid exclusion edges for source type %s", sourceType)
+	if !hasPathTo(edges[0], sourceType) {
+		return ExclusionEdges{}, fmt.Errorf("exclusion edges do not have weight for source type %s", sourceType)
```
adriantam: Do we want to update comment to indicate that it is the base edge that does not have the weight for the source type?

pkg/typesystem/weighted_graph.go
adriantam: Do we need to update this logic to assert that there must be only 2 edges?

pkg/typesystem/weighted_graph.go
```diff
@@ -182,8 +101,17 @@ func (t *TypeSystem) ConstructUserset(currentEdge *graph.WeightedAuthorizationMo
...
 		return This(), nil
+	case graph.LogicalTTUGrouping:
+		edges, err := t.GetEdgesFromNode(currentNode, sourceUserType)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get edges from node %s: %w", currentNode.GetUniqueLabel(), err)
+		}
+		if len(edges) == 0 {
+			return nil, fmt.Errorf("no edges found for node %s", currentNode.GetUniqueLabel())
+		}
+		return t.ConstructUserset(edges[0], sourceUserType)
```
adriantam, line 115: Can we call out as comment why do we only look at edges[0]?

pkg/server/commands/reverseexpand/reverse_expand_weighted.go
yissellokta: The weighted graph won't have logical grouping without edges, it will error out

pkg/typesystem/typesystem.go
yissellokta: The current tests should be sufficient; a recursive relation property will only be populated for all nodes and edges of the recursive path. What other tests are you thinking?

PR #2734, opened 2025-10-07
feat: switching postgresql to use pgxpool to allow for finer control

adriantam:
Currently, the PostgreSQL driver does not allow for finer control such as MaxConnLifetimeJitter, MinIdleConns, MinConns.  Fine tune control of these parameters allow more optimal performance for PostgreSQL

#### How is it being solved?
Switching driver to use [pgxpool](https://github.com/jackc/pgx/tree/master/pgxpool) instead of the standard pgx/sql.DB allows for more control. 

Switching to use pgxpool for PostgreSQL.  In addition, introducing new configuration parameter such as OPENFGA_DATASTORE_MIN_IDLE_CONNS and OPENFGA_DATASTORE_MIN_OPEN_CONNS.  Note that these two parameters are only applicable for PostgreSQL for now.  

## Review Checklist

### DIFF
--- a/pkg/server/config/config.go
+++ b/pkg/server/config/config.go
@@ -125,9 +125,17 @@ type DatastoreConfig struct {
 	// MaxOpenConns is the maximum number of open connections to the database.
 	MaxOpenConns int
 
+	// MinOpenConns is the minimum number of open connections to the database.
+	// This is only used for some datastore engines (PostgresSQL that uses pgxpool).
+	MinOpenConns int
+
 	// MaxIdleConns is the maximum number of connections to the datastore in the idle connection
-	// pool.
+	// pool. This is only used for some datastore engines (non-PostgresSQL that uses sql.DB).
 	MaxIdleConns int
+
+	// MinIdleConns is the minimum number of connections to the datastore in the idle connection
+	// pool. This is only used for some datastore engines (PostgresSQL that uses pgxpool).
+	MinIdleConns int
 
 	// ConnMaxIdleTime is the maximum amount of time a connection to the datastore may be idle.
 	ConnMaxIdleTime time.Duration
@@ -466,6 +474,14 @@ func (cfg *Config) VerifyServerSettings() error {
 		return errors.New("maxConditionsEvaluationCosts less than 100 can cause API compatibility problems with Conditions")
 	}
 
+	if cfg.Datastore.MaxOpenConns < cfg.Datastore.MinOpenConns {
+		return errors.New("datastore MaxOpenConns must not be less than datastore MinOpenConns")
+	}
+
+	if cfg.Datastore.MinOpenConns < cfg.Datastore.MinIdleConns {
+		return errors.New("datastore MinOpenConns must not be less than datastore MinIdleConns")
+	}
+
 	return nil
 }
 
@@ -706,7 +722,9 @@ func DefaultConfig() *Config {
 		Datastore: DatastoreConfig{
 			Engine:       "memory",
 			MaxCacheSize: DefaultMaxAuthorizationModelCacheSize,
+			MinIdleConns: 0,
 			MaxIdleConns: 10,
+			MinOpenConns: 0,
 			MaxOpenConns: 30,
 		},
 		GRPC: GRPCConfig{
--- a/pkg/storage/postgres/pgxpool_iter_query.go
+++ b/pkg/storage/postgres/pgxpool_iter_query.go
@@ -1,0 +1,94 @@
+package postgres
+
+import (
+	"context"
+
+	sq "github.com/Masterminds/squirrel"
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/openfga/openfga/pkg/storage/sqlcommon"
+)
+
+// pgxpoolIterQuery is a helper to run queries using pgxpool when used in sqlcommon iterator.
+type pgxpoolIterQuery struct {
+	db    *pgxpool.Pool
+	query string
+	args  []interface{}
+}
+
+var _ sqlcommon.SQLIteratorRowGetter = (*pgxpoolIterQuery)(nil)
+
+func newPgxPoolGetRows(db *pgxpool.Pool, sb sq.SelectBuilder) (*pgxpoolIterQuery, error) {
+	stmt, args, err := sb.ToSql()
+	if err != nil {
+		return nil, err
+	}
+	return &pgxpoolIterQuery{
+		db:    db,
+		query: stmt,
+		args:  args,
+	}, nil
+}
+
+// GetRows executes the pgxpool query and returns the sqlcommon.Rows.
+func (p *pgxpoolIterQuery) GetRows(ctx context.Context) (sqlcommon.Rows, error) {
+	rows, err := p.db.Query(ctx, p.query, p.args...)
+	if err != nil {
+		return nil, HandleSQLError(err)
+	}
+	return &pgxRowsWrapper{row: rows}, nil
+}
+
+// pgxtxnIterQuery is a helper to run queries using pgxpool when used in sqlcommon iterator.
+type pgxtxnIterQuery struct {
+	txn   pgx.Tx
+	query string
+	args  []interface{}
+}
+
+var _ sqlcommon.SQLIteratorRowGetter = (*pgxtxnIterQuery)(nil)
+
+func newPgxTxnGetRows(txn pgx.Tx, sb sq.SelectBuilder) (*pgxtxnIterQuery, error) {
+	stmt, args, err := sb.ToSql()
+	if err != nil {
+		return nil, err
+	}
+	return &pgxtxnIterQuery{
+		txn:   txn,
+		query: stmt,
+		args:  args,
+	}, nil
+}
+
+// GetRows executes the txn query and returns the sqlcommon.Rows.
+func (p *pgxtxnIterQuery) GetRows(ctx context.Context) (sqlcommon.Rows, error) {
+	rows, err := p.txn.Query(ctx, p.query, p.args...)
+	if err != nil {
+		return nil, HandleSQLError(err)
+	}
+	return &pgxRowsWrapper{row: rows}, nil
+}
+
+type pgxRowsWrapper struct {
+	row pgx.Rows
+}
+
+func (r *pgxRowsWrapper) Err() error {
+	return r.row.Err()
+}
+
+func (r *pgxRowsWrapper) Next() bool {
+	return r.row.Next()
+}
+
+func (r *pgxRowsWrapper) Scan(dest ...any) error {
+	return r.row.Scan(dest...)
+}
+
+func (r *pgxRowsWrapper) Close() error {
+	r.row.Close()
+	return nil
+}
+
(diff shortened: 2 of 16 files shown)


aaguiarz:
We should explain how to configure those new settings

adriantam:
> We should explain how to configure those new settings

Yes, we will need to update https://openfga.dev/docs/best-practices/running-in-production#database-recommendations on how to configure for PostgreSQL.

adriantam:
@coderabbitai review

adriantam:
@coderabbitai review

senojj:
As we begin to move towards supporting proprietary driver interfaces, many implementations will expose configurations that other implementations do not. Maintaining a collection of individual environment variables that provide access to all of these configurations will not be maintainable.

Perhaps we need to move towards a more generic way to set storage driver configurations.

ex: `OPENFGA_DATASTORE_CONFIG="MinIdleConns=10;MinOpenConns=10;MaxConnLifetimeJitter=15s"`

pkg/server/config/config.go
```diff
@@ -125,10 +125,18 @@ type DatastoreConfig struct {
 	// MaxOpenConns is the maximum number of open connections to the database.
 	MaxOpenConns int
 
+	// MinOpenConns is the minimum number of open connections to the database.
+	// This is only used for some datastore engines (PostgresSQL that uses pgxpool).
```
justincoh: > (PostgresSQL that uses pgxpool).

This is all FGA postgresql after this PR right? If so, could just say `This is only used for Postgresql` for clarity

pkg/storage/postgres/pgxpool_iter_query.go
```diff
@@ -0,0 +1,95 @@
...
+	rows, err := p.txn.Query(ctx, p.query, p.args...)
+	if err != nil {
+		return nil, HandleSQLError(err)
+	}
+	return &pgxRowsWrapper{row: rows}, nil
+}
+
+// pgxRowsWrapper wraps pgx.Rows to implement sqlcommon.Rows interface.
+type pgxRowsWrapper struct {
+	row pgx.Rows
```
justincoh: nit: `rows pgx.Rows` for consistency?

pkg/storage/postgres/postgres.go
```diff
@@ -54,63 +52,83 @@ type Datastore struct {
 // Ensures that Datastore implements the OpenFGADatastore interface.
 var _ storage.OpenFGADatastore = (*Datastore)(nil)
 
-// initDB initializes a new postgres database connection.
-func initDB(uri string, username string, password string, cfg *sqlcommon.Config) (*sql.DB, error) {
-	if username != "" || password != "" {
+func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Config, error) {
+	c, err := pgxpool.ParseConfig(uri)
+	if err != nil {
+		return nil, fmt.Errorf("parse postgres connection uri: %w", err)
```
justincoh: Can we modify this error message so it isn't identical to the one on line 64? If either of them hits we'd want to know exactly which line triggered

pkg/storage/postgres/postgres.go
```diff
@@ -54,63 +52,83 @@ type Datastore struct {
...
-			parsed.User = url.User(username)
 		}
+	}
 
-		uri = parsed.String()
+	if cfg.MinIdleConns != 0 {
+		c.MinIdleConns = int32(cfg.MinIdleConns)
 	}
+	// TODO: If MinIdleConns is not set, should we use the max idle conns to provide some sane value?  Or is
+	// it better to leave it to the pgx default?
```
justincoh: I think we should probably maintain some number of MinIdleConns > 0, but I don't have a good sense for what that number should be.

pkg/storage/postgres/postgres.go
```diff
@@ -303,24 +309,187 @@ func (s *Datastore) Write(
 ) error {
 	ctx, span := startTrace(ctx, "Write")
 	defer span.End()
+	return s.write(ctx, store, deletes, writes, storage.NewTupleWriteOptions(opts...), time.Now().UTC())
+}
+
+func (s *Datastore) write(
```
justincoh: this is a doozy of a method, is it possible to split it apart?

pkg/storage/postgres/postgres.go
```diff
@@ -202,26 +211,18 @@ func (s *Datastore) Close() {
...
+func (s *Datastore) getPgxPool(consistency openfgav1.ConsistencyPreference) *pgxpool.Pool {
+	if consistency == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
 		// If we are using higher consistency, we need to use the write database.
-		return s.primaryStbl
+		return s.primaryDB
 	}
 	if s.isSecondaryConfigured() {
 		// If we are using lower consistency, we can use the read database.
-		return s.secondaryStbl
+		return s.secondaryDB
```
justincoh, line 220: How is the `secondaryDB` kept up to date with the primary?

DanCech: https://openfga.dev/docs/getting-started/setup-openfga/configure-openfga#postgresql-read-replicas-configuration

pkg/storage/postgres/postgres.go
adriantam: I would defer to the pgxpool's [defaultMinIdleConns](https://github.com/jackc/pgx/blob/cf865756ec51121c099d5dd369b8b4ed8cafd12f/pgxpool/pool.go#L20) (which is 0) for now.  We can tweak this later when we have more data to support a "right number".

adriantam:
> As we begin to move towards supporting proprietary driver interfaces, many implementations will expose configurations that other implementations do not. Maintaining a collection of individual environment variables that provide access to all of these configurations will not be maintainable.
> 
> Perhaps we need to move towards a more generic way to set storage driver configurations.
> 
> ex: `OPENFGA_DATASTORE_CONFIG="MinIdleConns=10;MinOpenConns=10;MaxConnLifetimeJitter=15s"`

I have added an issue and we can address this as a separate piece of work as we should do this for all datastore and have a migration plan where we support configuration via both methods at the same time.

pkg/storage/postgres/postgres.go
```diff
@@ -312,34 +312,15 @@ func (s *Datastore) Write(
...
-		return HandleSQLError(err)
-	}
-	// Important - use the same txn (instead of via db) to ensure all works are done as a transaction
-
-	defer func() { _ = txn.Rollback(ctx) }()
-
-	// 2. Compile a SELECT … FOR UPDATE statement to read the tuples for writes and lock tuples for deletes
-	// Build a deduped, sorted list of keys to lock.
-	lockKeys := sqlcommon.MakeTupleLockKeys(deletes, writes)
+// execute SELECT … FOR UPDATE statement for all the rows indicated by the lockKeys
```
DanCech, line 325: This is good work, but we're creating drift between this implementation and the MySQL and SQLite variants, I'd suggest breaking it out into a separate PR and keeping the 3 in sync.

adriantam: Agree.  We will need to address the drift in separate PR

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 ### Added
 - Enable planner for selecting check resolution strategies based on runtime statistics. [#2751](https://github.com/openfga/openfga/pull/2751)
 
+### Changed
+- Breaking: Update PostgreSQL to use [pgxpool](https://pkg.go.dev/github.com/jackc/pgx/v5/pgxpool) instead of `database/sql` to allow for finer PostgreSQL connection control. [#2734](https://github.com/openfga/openfga/pull/2734).
```
justincoh, line 15: What are your thoughts on semver for this PR?

justincoh (approved):

CHANGELOG.md
adriantam: We likely need to use a minor version instead of patch version.  CC: @tylernix

tylernix: Minor

tylernix: But it would be nice to not have to release minor version if any datastore has a breaking change. We should discuss how to follow semver for openfga and supported datastore separately. Only increment openfga minor version if a change affects all underlying datastores.

adriantam: Agree on that.

pkg/storage/postgres/postgres.go
```diff
@@ -54,63 +52,87 @@ type Datastore struct {
...
 	}
 
-	db, err := sql.Open("pgx", uri)
-	if err != nil {
-		return nil, fmt.Errorf("initialize postgres connection: %w", err)
+	if cfg.MinIdleConns != 0 {
+		c.MinIdleConns = int32(cfg.MinIdleConns)
+	} else if cfg.MaxOpenConns != 0 {
+		// Use some sane value for defaults
+		c.MinIdleConns = int32(cfg.MaxOpenConns)
```
DanCech: With this change we'll end up holding the maximum number of allowed connections open at all times unless the config is updated to set both minimums. I'm not sure that setting it to zero is the correct solution either (since we'll end up churning connections needlessly), but for users who have configured a relatively high maximum it could result in a significant change in the number of open connections under normal usage.

Would it make sense to default these limits to some fraction of the maximum?

yissellokta (approved):

pkg/storage/postgres/postgres.go
adriantam: Agree with you.  Maybe half of the max?  @senojj ?

adriantam: Spoke with @senojj off line.  The desired behaviour is to let the 0 to use the pgxpool default behaviour.  Will update this code to reflect this.

senojj: The reasoning for setting the default values to 0 is that it matches, as closely as possible, the existing connection behavior.

The current connection pool will not open any connections until requested by the application. With this new connection pool, each connection will have a maximum lifetime, which will unavoidably churn connections more frequently than the current default configuration.

Additionally, depending upon the connection lifetime configuration, it is possible for the minimum connections in the pool to drop below the configured minimum until the health check detects that new connections are necessary: https://github.com/jackc/pgx/blob/e4a063e7cb0fe9dd022fb1171d7655f6e837c98e/pgxpool/pool.go#L180-L190

justincoh (dismissed):
Ty for remembering to update the config-schema also 🙏

pkg/storage/sqlcommon/sqlcommon.go
```diff
@@ -774,6 +799,57 @@ func Write(
...
+// Write provides the common method for writing to database across sql storage.
+func Write(
+	ctx context.Context,
+	dbInfo *DBInfo,
+	db *sql.DB,
+	store string,
+	deletes storage.Deletes,
+	writes storage.Writes,
+	opts storage.TupleWriteOptions,
+	now time.Time,
```
senojj: 8 arguments is quite a lot, which generally indicates that the function may be doing too much, or that we are missing some logical data grouping.

adriantam: group parameters together into single data struct

justincoh (approved):

PR #2742, opened 2025-10-15
refactor: build checkResolvers per-request instead of at startup

justincoh:
FGA's checkResolvers are currently built one time at server startup and remain statically-defined for the life of the service. This refactor changes it so check resolvers are built per-request, which allows for more flexibility. We have some coming FeatureFlag-related work which will make immediate use of this change.

I benchmarked the new buildCheckResolver function added in this PR and it takes around `500ns` aka `0.0005ms`, so we're not concerned with a performance regression here at all.

* removed checkResolver, checkResolverCloser, listObjectsCheckResolver and listObjectsCheckResolverCloser from the server struct
* instantiates checkResolver at the start of every `Check`, `ListObjects`, and `BatchCheck` request, and defers its `.closer()` call

## Review Checklist

## Summary by CodeRabbit

* **Refactor**
  * Check resolvers are now created per request rather than at server start, with per-request lifecycle management applied across Check, Batch Check, and List Objects operations.

* **Documentation**
  * Changelog updated (Unreleased 1.10.3) to reflect the per-request check resolver behavior.

* **Tests**
  * Updated tests to build a fresh resolver per request and ensure proper cleanup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 ### Fixed
 - Align datastore throttle configuration names with struct property names. [#2668](https://github.com/openfga/openfga/pull/2668)
+- Increase the buffer size for LO requests to be 10% of the max results returned or 10. [#2735](https://github.com/openfga/openfga/pull/2735)
 
 ## [1.10.2] - 2025-09-29
 ### Changed
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -321,7 +321,11 @@ func (q *ListObjectsQuery) evaluate(
 			}
 		}
 
-		reverseExpandResultsChan := make(chan *reverseexpand.ReverseExpandResult, 1)
+		var bufferSize uint32
+		cappedMaxResults := uint32(math.Min(float64(maxResults), 1000)) // cap max results at 1000
+		bufferSize = uint32(math.Max(float64(cappedMaxResults/10), 10)) // 10% of max results, but make it at least 10
+
+		reverseExpandResultsChan := make(chan *reverseexpand.ReverseExpandResult, bufferSize)
 		objectsFound := atomic.Uint32{}
 
 		ds := storagewrappers.NewRequestStorageWrapperWithCache(


pkg/server/list_objects.go
```diff
@@ -270,3 +283,21 @@ func (s *Server) StreamedListObjects(req *openfgav1.StreamedListObjectsRequest,
...
+	checkCacheOptions, checkDispatchThrottlingOptions := s.getCheckResolverOptions()
+
+	listObjectsCheckResolver, listObjectsCheckResolverCloser, err := graph.NewOrderedCheckResolvers([]graph.CheckResolverOrderedBuilderOpt{
+		graph.WithLocalCheckerOpts([]graph.LocalCheckerOption{
+			graph.WithResolveNodeBreadthLimit(s.resolveNodeBreadthLimit),
+			graph.WithOptimizations(s.IsExperimentallyEnabled(ExperimentalCheckOptimizations)),
+			graph.WithMaxResolutionDepth(s.resolveNodeLimit),
+		}...),
+		graph.WithCachedCheckResolverOpts(s.cacheSettings.ShouldCacheCheckQueries(), checkCacheOptions...),
+		graph.WithDispatchThrottlingCheckResolverOpts(s.checkDispatchThrottlingEnabled, checkDispatchThrottlingOptions...),
```
justincoh, line 299: removed all the shadow-check-related check logic from the list objects checker, we don't need an additional layer of shadow checking

pkg/server/list_objects.go
```diff
@@ -270,3 +283,21 @@ func (s *Server) StreamedListObjects(req *openfgav1.StreamedListObjectsRequest,
...
+	}...).Build()
+	if err != nil {
+		return nil, nil, err
+	}
+	return listObjectsCheckResolver, listObjectsCheckResolverCloser, nil
+}
```
senojj: Instead of returning 3 separate values, you could just return the "builder" instance that is currently having it's `Build()` method invoked on line 298.

pkg/server/check.go
```diff
@@ -154,3 +161,37 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+		}...),
+		graph.WithCachedCheckResolverOpts(s.cacheSettings.ShouldCacheCheckQueries(), checkCacheOptions...),
+		graph.WithDispatchThrottlingCheckResolverOpts(s.checkDispatchThrottlingEnabled, checkDispatchThrottlingOptions...),
+	}...).Build()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return checkResolver, checkResolverCloser, nil
+}
```
senojj: Instead of returning 3 separate values, you could just return the "builder" instance that is currently having it's Build() method invoked on line 191.

justincoh:
@Vic-Dev Refactor brought it up to 87%. Still not > 90% but close enough for me. The only way these builder blocks can error is if we configure a `maxCacheSize <= 0` in code, which would 💥 tests elsewhere

PR #2752, opened 2025-10-20
Feat/read support

yissellokta:
As part of the rewrite of Check we want to be more efficient on the queries that are executed against the datastore. This PR is to filter by subject type when we have different terminal types for one relation we would like to filter in the database instead of filtering in memory

## Review Checklist

## Release Notes

* **New Features**
  * Added support for user-type prefix filtering across all storage backends, enabling queries to match tuples by user category (e.g., "user:", "group:") in addition to exact user matches.

* **Tests**
  * Expanded test coverage for user-type filtering and pagination scenarios across storage implementations.


justincoh (commented):
If the filtering has moved to the DB-level, is there some go cleanup needed as part of this PR too? Or should we do that in a follow up?

yissellokta:
> Is there any opportunity to share the tests across mysql/postgres/sqlite? Ideally, these tests are shared if there are nothing specific about the datastore type.

Where do you suggest to add this shared test?

yissellokta:
> If the filtering has moved to the DB-level, is there some go cleanup needed as part of this PR too? Or should we do that in a follow up?

It won't affect the current code, as we are not calling it with those parameters, but the new code is

elbuo8 (approved):

adriantam:
> > Is there any opportunity to share the tests across mysql/postgres/sqlite? Ideally, these tests are shared if there are nothing specific about the datastore type.
> 
> Where do you suggest to add this shared test?

Would it make sense to add it to https://github.com/openfga/openfga/blob/main/pkg/storage/test/tuples.go?  In this way, all storage will benefit from the test.

PR #2757, opened 2025-10-23
Feat/read conditions support

yissellokta:
Add to Read, ReadPage, ReadStartingWithUser and ReadUsersets the ability to filter by condition names. If the model has a relation rel1: [user, user with cond1, user with cond2] we can filter by passing the []string{"", "cond1"}. To continue gathering all the data without filtering by condition, then do not populate the condition part of the filter. 
The reason to introduce this filtering is to reduce the number of unnecessary tuples loaded if stores has multiple models with tuples with different conditions, and reduce the filtering in memory 
Current code is not using this new filtering but they new algorithms will benefit from it.

## Review Checklist

## Summary by CodeRabbit

# Release Notes

* **Refactor**
  * Restructured tuple filtering mechanism across all storage backends to use a unified filter interface with support for conditional matching, improving consistency in relationship query handling.

### DIFF
--- a/internal/graph/recursive_resolver.go
+++ b/internal/graph/recursive_resolver.go
@@ -169,24 +169,18 @@ func buildRecursiveMapper(ctx context.Context, req *ResolveCheckRequest, mapping
 	consistencyOpts := storage.ConsistencyOptions{
 		Preference: req.GetConsistency(),
 	}
+
 	switch mapping.kind {
 	case storage.UsersetKind:
-		objectType := ""
-		relation := ""
-		if req.GetTupleKey() != nil {
-			objectType = req.GetTupleKey().GetObject()
-			relation = req.GetTupleKey().GetRelation()
-		}
+		objectType := req.GetTupleKey().GetObject()
+		relation := req.GetTupleKey().GetRelation()
 		iter, err = ds.ReadUsersetTuples(ctx, req.GetStoreID(), storage.ReadUsersetTuplesFilter{
 			Object:                      objectType,
 			Relation:                    relation,
 			AllowedUserTypeRestrictions: mapping.allowedUserTypeRestrictions,
 		}, storage.ReadUsersetTuplesOptions{Consistency: consistencyOpts})
 	case storage.TTUKind:
-		objectType := ""
-		if req.GetTupleKey() != nil {
-			objectType = req.GetTupleKey().GetObject()
-		}
+		objectType := req.GetTupleKey().GetObject()
 		iter, err = ds.Read(ctx, req.GetStoreID(),
 			storage.ReadFilter{Object: objectType, Relation: mapping.tuplesetRelation, User: ""},
 			storage.ReadOptions{Consistency: consistencyOpts})
--- a/pkg/server/commands/read.go
+++ b/pkg/server/commands/read.go
@@ -94,15 +94,13 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgav1.ReadRequest) (*o
 		Consistency: storage.ConsistencyOptions{Preference: req.GetConsistency()},
 	}
 
-	var filter storage.ReadFilter
+	filter := storage.ReadFilter{}
 	if tk != nil {
 		filter = storage.ReadFilter{
 			Object:   tk.GetObject(),
 			Relation: tk.GetRelation(),
 			User:     tk.GetUser(),
 		}
-	} else {
-		filter = storage.ReadFilter{}
 	}
 
 	tuples, contUlid, err := q.datastore.ReadPage(ctx, store, filter, opts)


pkg/storage/mysql/mysql.go
```diff
@@ -319,6 +327,10 @@ func (s *Datastore) ReadUsersetTuples(
 		sb = sb.Where(orConditions)
 	}
 
+	if len(filter.Conditions) > 0 {
+		sb = sb.Where(sq.Eq{"condition_name": filter.Conditions})
+	}
+
```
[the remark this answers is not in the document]
adriantam: Agree with the @coderabbitai flag above.  @yissellokta mind looking into this?

internal/graph/recursive_resolver.go
```diff
@@ -171,13 +171,24 @@ func buildRecursiveMapper(ctx context.Context, req *ResolveCheckRequest, mapping
 	}
 	switch mapping.kind {
 	case storage.UsersetKind:
+		objectType := ""
+		relation := ""
+		if req.GetTupleKey() != nil {
```
elbuo8: When would it be nil?

pkg/server/commands/read.go
```diff
@@ -94,7 +94,18 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgav1.ReadRequest) (*o
 		Consistency: storage.ConsistencyOptions{Preference: req.GetConsistency()},
 	}
 
-	tuples, contUlid, err := q.datastore.ReadPage(ctx, store, tupleUtils.ConvertReadRequestTupleKeyToTupleKey(tk), opts)
+	var filter storage.ReadFilter
+	if tk != nil {
```
elbuo8, line 98: Why not instantiate `filter` with zero value rather than nil and then populate if tk is available.

pkg/storage/mysql/mysql.go
```diff
@@ -174,24 +174,32 @@ func (s *Datastore) read(ctx context.Context, store string, tupleKey *openfgav1.
...
+	objectType, objectID := tupleUtils.SplitObject(filter.Object)
 	if objectType != "" {
 		sb = sb.Where(sq.Eq{"object_type": objectType})
 	}
 	if objectID != "" {
 		sb = sb.Where(sq.Eq{"object_id": objectID})
 	}
-	if tupleKey.GetRelation() != "" {
-		sb = sb.Where(sq.Eq{"relation": tupleKey.GetRelation()})
+	if filter.Relation != "" {
```
elbuo8, line 184: Keep using getters when dealing with protos (I don't like them either but for consistency purposes)

internal/graph/recursive_resolver.go
yissellokta: it should not be, it was AI complaining :)

DanCech: technically it's possible, if `req` is a nil pointer then `req.GetTupleKey()` will return `nil` due to the internal nil pointer receiver check. I wonder whether an `if req == nil { return nil, errors.New("nil request") }` check above would make the AI happy?

pkg/server/commands/read.go
```diff
@@ -94,7 +94,18 @@ func (q *ReadQuery) Execute(ctx context.Context, req *openfgav1.ReadRequest) (*o
...
+		filter = storage.ReadFilter{
+			Object:   tk.GetObject(),
+			Relation: tk.GetRelation(),
+			User:     tk.GetUser(),
+		}
+	} else {
+		filter = storage.ReadFilter{}
+	}
```
senojj: `var filter storage.ReadFilter` is already the "zero value" of `storage.ReadFilter`. This `else` branch is unnecessary because it simply re-assigns the zero value to `filter`.

pkg/storage/memory/memory.go
```diff
@@ -528,6 +532,10 @@ func (s *MemoryBackend) ReadUsersetTuples(
 					continue
 				}
 			}
+
+			if len(filter.Conditions) > 0 && !slices.Contains(filter.Conditions, t.ConditionName) {
```
senojj, line 536: if `filter.Conditions` is empty, `slices.Contains` will return `false`. The length check is redundant.
```suggestion
			if !slices.Contains(filter.Conditions, t.ConditionName) {
```

pkg/storage/memory/memory.go
```diff
@@ -561,6 +569,10 @@ func (s *MemoryBackend) ReadStartingWithUser(
 			continue
 		}
 
+		if len(filter.Conditions) > 0 && !slices.Contains(filter.Conditions, t.ConditionName) {
```
senojj, line 572: apropos of the above comment, the length check is redundant
```suggestion
		if !slices.Contains(filter.Conditions, t.ConditionName) {
```

pkg/storage/memory/memory.go
yissellokta: in this case it could be null and it will fail

pkg/storage/memory/memory.go
yissellokta: in this case it could be null and it will fail

pkg/storage/memory/memory.go
DanCech: makes sense, the empty slice means "match everything" so we only check `slices.Contains` if `filter.Conditions` is not empty.

senojj: totally missed that. thanks.

PR #2772, opened 2025-10-29
Rebuild Reverse Lookup Index To Support Omitted Relation

JeremyLoy:
This PR adds a different reverse lookup index for MySQL and Postgres that places `relation` last for reasons listed below. I omitted adding a SQLite index because the schema is significantly different than MySQL and Postgres. It is likely that this problem will have to be fixed by multiple indexes, and I don't feel comfortable doing that without more SQlite expertise or understanding of why that schema is so different.

An earlier iteration of this PR tried to delete the original index, but it seems that it would negatively impact other benchmarks (great test suite!) so I'm keeping it. I'm also open to helping write additional benchmarks to keep this usecase maintained, but I need a bit of guidance. It seems that there are no benchmarks presently for `Read`, so it might be best to save that for https://github.com/openfga/roadmap/issues/33

This is tangentially related to https://github.com/openfga/roadmap/issues/33 which covers adding functionality to `Read`

EDIT 10/30: An additional note - we are currently using version 1.9.X in prod, which does not have these migrations applied https://github.com/openfga/openfga/pull/2661/files yet. That actually removes the original reverse lookup index, which Read partially relies upon. I believe our performance would be even worse without this index, so when I find time this week I'm going to run a benchmark with the 1.10 migration applied, as well as this fix (and reverting the deletion of the original index)

---

EDIT 11/3: We had some time to run our own benchmarking and I have attached screenshots below. cc @senojj and @adriantam 
image
image

---

We are using the `read` API  to facilitate data cleanup for entities in our system that have been deleted. We don't care what the relationship is for a given object, we just want to delete all tuples. We use two queries to do so:

```
object=foo:bar
```
and
```
object=foo:
user=bar:baz
```

In both cases, we are omitting `relation`.

This is supported in the API, however it caused immediate and massive performance problems when we deployed it. I believe it is because the covering index isn't being used due to the column order in the original index.

Below I have attached a screenshot of our metrics from AWS RDS Performance Insights and Grafana. As you can see, a small burst of <20rps for the Read API spiked our CPU useage to 100% (from a baseline of 4%, x axis is 15 seconds)

image

image
image

## Summary by CodeRabbit

* **Chores**
  * Reorganized database indexes for reverse lookup operations across MySQL and PostgreSQL environments, optimizing query performance.

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -131,11 +131,11 @@
             "x-env-variable": "OPENFGA_CONTEXT_PROPAGATION_TO_DATASTORE"
         },
         "experimentals": {
-            "description": "a list of experimental features to enable",
+            "description": "a comma-separated list of experimental features to enable",
             "type": "array",
             "items": {
                 "type": "string",
-                "enum": ["enable-check-optimizations", "enable-list-objects-optimizations", "enable-access-control", "pipeline_list_objects"]
+                "enum": ["enable-check-optimizations", "enable-list-objects-optimizations", "enable-access-control", "pipeline_list_objects", "datastore_throttling"]
             },
             "default": [],
             "x-env-variable": "OPENFGA_EXPERIMENTALS"
--- a/cmd/run/run.go
+++ b/cmd/run/run.go
@@ -93,7 +93,7 @@ func NewRunCommand() *cobra.Command {
 	defaultConfig := serverconfig.DefaultConfig()
 	flags := cmd.Flags()
 
-	flags.StringSlice("experimentals", defaultConfig.Experimentals, fmt.Sprintf("a list of experimental features to enable. Allowed values: %s, %s, %s, %s", serverconfig.ExperimentalCheckOptimizations, serverconfig.ExperimentalCheckOptimizations, serverconfig.ExperimentalAccessControlParams, serverconfig.ExperimentalPipelineListObjects))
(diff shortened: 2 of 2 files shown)


senojj:
Hi @JeremyLoy, thank you for bringing this case to our attention. We are currently reviewing.

JeremyLoy:
thanks @senojj ! We are running some benchmarks today with this index applied manually. I am updating this post as the results come in.

We will likely need to add back the index that was removed in #2661  as well. We are going to run a test with that index deleted too

JeremyLoy:
@senojj Update - we finished our load testing and find no reason to partially revert #2661 .

Additionally, I took a stab at adding a benchmark for `Read`. let me know what you think!

Vic-Dev:
@claude

senojj:
@JeremyLoy, I apologize for the delay in resolution. Ultimately, it was concluded that `Read` supports too many filter permutations to reasonably cover all column combinations (4!) with a composite index, without negatively impacting write performance. However, users of OpenFGA are encouraged to create any indexes necessary for their specific needs.

JeremyLoy:
@senojj no worries, I'll close this then. Our application is not write performance bound, but I can sympathize with that perspective. 

I'm curious - are you able to share more about metrics thatthis index (or the additional other indexes you spoke of) impacted write performance?

senojj:
@JeremyLoy in order to provide a covering index for every `Read` filter permutation would require 4! == 24 unique indexes. Adding 24 composite indexes would most certainly have a negative impact on write/update performance, especially for larger data sets.

Adding only your proposed index would not impact performance on its own, but it would open the door to questions like, "why isn't an index covering w, x, y, z created? Or why not an index covering x, z, w, y?".

I think there are a set of emerging covering indexes that will be more useful than others, and we may choose to add those as part of the official schema in the future.

Another option would be to create 4 single column indexes, however that would not provide the same performance as the covering indexes.

PR #2792, opened 2025-11-06
fix: cache controller invalidation timestamp

justincoh:
In #2779 we pushed an incorrect fix for Cached Check resolutions, which would have drastically reduced the check cache hit rate. It was [quickly reverted](https://github.com/openfga/openfga/pull/2791) (and never released). This PR is a correct fix.

### Context
OpenFGA has multiple caches, and we like to stretch their TTLs as long as possible to maximize hit rate. To enable this, we have an optional `cache_controller`, which attempts to do smarter cache invalidations specific to each cache; it's TTL is typically much lower than the actual cache TTLs. The goal is that individual cache's TTL values can be made quite long if you have the cache controller checking on their validity more frequently.

This PR addresses an issue where the CacheController was returning an incorrect `LastInvalidationTime` to `Check`, [causing the cached_resolver](https://github.com/openfga/openfga/blob/main/internal/graph/cached_resolver.go#L165) to return cached values when it should not have.

#### How the cache controller works in the Check flow
When we receive a check request, one of the first things we do is call [cacheController.DetermineInvalidationTime](https://github.com/openfga/openfga/blob/main/pkg/server/commands/check_command.go#L111) to determine the check cache's [LastCacheInvalidationTime](https://github.com/openfga/openfga/blob/main/pkg/server/commands/check_command.go#L121). The last cache invalidation time can be described in English as `the most recent write or delete this store has committed`. That `LastInvalidationTime` is then passed into check. If check finds a cached response to the current query, it will [compare that cache record's timestamp against the LastInvalidationTime](https://github.com/openfga/openfga/blob/main/internal/graph/cached_resolver.go#L165), and if the cached response is newer it will return it.

#### What the issues were
`DetermineInvalidationTime` checks for a `ChangelogCacheEntry`, and was returning that ChangelogCacheEntry's LastModified timestamp **only if `LastModified` was within {TTL} seconds of Now()**, otherwise it would trigger a background job to see if any other caches need invalidating and return `time.Time{}`. This caused two issues:

1. Comparing the `ChangelogCacheEntry.LastModified` with the `CacheController.TTL` is incorrect. Consider the case where a store hasn't written or deleted a tuple in hours or days. Their `changelog.LastModified` will not change, and will *never* be within the CacheController's TTL; so this method would **always** return `time.Time{}` for that store.
2. When comparing cached check entries with `time.Time{}`, **all** of them are newer, so Check could have returned a cached value when it should not have.

#### How is it being solved?
To address issue 1 above:
* Track when each store's most recent invalidation occurred via `invalidationLog` sync.map. **This** is what is now being compared to the cache controller's TTL value. In this code path, the cache controller only needs to know when the changelog was last checked, **not** when the last tuple change was made. The CacheController `TTL` is not really a "TTL" but more of a "frequency to check the changelog".

To address issue 2 above:
* Changed the signature of `DetermineInvalidationTime` to return an additional boolean value. 
* When we have not invalidated a particular store's cache in more than `TTL` time, this new boolean will be `false` so `Check` knows to disregard any cached responses for this request. In this case, we also trigger background invalidations (as we were before this PR).

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect default return value in cache invalidation timing logic, ensuring the system properly uses cached timestamps even after expiration periods for improved reliability in data retrieval operations.

### DIFF
--- a/assets/tests/abac_tests.yaml
+++ b/assets/tests/abac_tests.yaml
@@ -2897,3 +2897,65 @@ tests:
               "ts": "2023-10-11T09:00:00.000Z"
             expectation:
               - user:*
+  - name: wildcard_with_condition
+    stages:
+      - model: |
+          model
+            schema 1.1
+          type user
+          type group1
+            relations
+              define member: [user, user:*]
+              define other: [user with ts_less_than]
+          condition ts_less_than(ts: timestamp) {
+            ts < timestamp("2023-10-11T10:00:00.000Z")
+          }
+        tuples:
+          - user: user:*
+            relation: member
+            object: group1:1
+        checkAssertions:
+          - tuple:
+              user: user:anne
+              relation: member
+              object: group1:1
+            expectation: true
+      - model: |
+          model
+            schema 1.1
+          type user
+          type group1
+            relations
+              define member: [user, user:* with ts_less_than]
+              define other: [user with ts_less_than]
+          condition ts_less_than(ts: timestamp) {
+            ts < timestamp("2023-10-11T10:00:00.000Z")
+          }
+        tuples:
+          - user: user:*
+            relation: member
+            object: group1:2
+            condition:
+              name: ts_less_than
+        checkAssertions:
+          - tuple:
+              user: user:anne
+              relation: member
+              object: group1:1
+            context:
+              "ts": "2023-10-11T09:00:00.000Z"
+            expectation: false
+          - tuple:
+              user: user:anne
+              relation: member
+              object: group1:2
+            context:
+              "ts": "2023-10-11T09:00:00.000Z"
+            expectation: true
+          - tuple:
+              user: user:anne
+              relation: member
+              object: group1:2
+            context:
+              "ts": "2023-10-11T11:00:00.000Z"
+            expectation: false
--- a/internal/validation/validation.go
+++ b/internal/validation/validation.go
@@ -201,6 +201,9 @@ func validateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) e
 				if directlyRelatedType.GetWildcard() != nil && !tuple.IsTypedWildcard(tk.GetUser()) {
 					continue
 				}
+			} else if tuple.IsTypedWildcard(tk.GetUser()) {
+				// This is a wildcard tuple but the directlyRelatedType tuple is not for wildcard.
+				continue
 			}
 
 			return nil
--- a/pkg/server/commands/list_objects_shadow.go
+++ b/pkg/server/commands/list_objects_shadow.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
@@ -178,7 +180,14 @@ func (q *shadowedListObjectsQuery) ExecuteStreamed(ctx context.Context, req *ope
 // If the shadow function takes longer than shadowTimeout, it will be cancelled, and its result will be ignored, but the shadowTimeout event will be logged.
 // This function is designed to be run in a separate goroutine to avoid blocking the main execution flow.
 func (q *shadowedListObjectsQuery) executeShadowModeAndCompareResults(parentCtx context.Context, req *openfgav1.ListObjectsRequest, mainResult *ListObjectsResponse, latency time.Duration) {
-	parentCtx, span := tracer.Start(parentCtx, "shadow")
+	linkedSpan := trace.SpanContextFromContext(parentCtx)
+	link := trace.Link{
+		SpanContext: linkedSpan,
+	}
+	_, span := tracer.Start(context.Background(), "shadow", trace.WithLinks(link), trace.WithAttributes(
+		attribute.String("main.trace.id", linkedSpan.TraceID().String()),
+	))
+	parentCtx = trace.ContextWithSpan(parentCtx, span)
 	defer span.End()
 
 	shadowCtx, shadowCancel := context.WithTimeout(parentCtx, q.shadowTimeout)
@@ -234,6 +243,7 @@ func (q *shadowedListObjectsQuery) executeShadowModeAndCompareResults(parentCtx 
 
 	// compare sorted string arrays - sufficient for equality check
 	if !maps.Equal(mapResultMain, mapResultShadow) {
+		span.SetAttributes(attribute.Bool("matches", false))
 		delta := calculateDelta(mapResultMain, mapResultShadow)
 		totalDelta := len(delta)
 		// Limit the delta to maxDeltaItems
@@ -253,6 +263,7 @@ func (q *shadowedListObjectsQuery) executeShadowModeAndCompareResults(parentCtx 
 			loShadowLogFields(req, fields...)...,
 		)
 	} else {
+		span.SetAttributes(attribute.Bool("matches", true))
 		fields = append(
 			fields,
 			zap.Bool("is_match", true),
--- a/pkg/server/commands/reverseexpand/pipeline.go
+++ b/pkg/server/commands/reverseexpand/pipeline.go
@@ -935,8 +935,7 @@ type Pipeline struct {
 }
 
 func (p *Pipeline) Build(ctx context.Context, source Source, target Target) iter.Seq[Item] {
-	ctxParent, span := pipelineTracer.Start(ctx, "Pipeline.Build")
-	defer span.End()
+	ctxParent, span := pipelineTracer.Start(ctx, "pipeline")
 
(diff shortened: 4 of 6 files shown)


ttrzeng (approved):

internal/cachecontroller/cache_controller.go
```diff
@@ -155,6 +159,13 @@ func (c *InMemoryCacheController) DetermineInvalidationTime(
 
 	c.InvalidateIfNeeded(ctx, storeID)
 
+	// This means the most recent Changelog entry is older than the CacheController TTL,
+	// so we trigger the invalidation workflow but also return the most recent changelog timestamp
+	// for use in the check resolver, since the Check Cache TTL is likely longer than the CacheController.
+	if cachedEntry != nil {
+		return cachedEntry.LastModified
+	}
```
senojj, line 173: When the type cast fails because `cacheResp` is not a `*storage.ChangelogCacheEntry`, `entry` will be `nil`. Additionally, if the type cast succeeds, but the value is `nil`, then `entry` will be `nil`. We can avoid some nesting here:
```
	entry, _ := cacheResp.(*storage.ChangelogCacheEntry)
	if entry != nil {
		// the TTL grace period hasn't been breached, return before triggering any invalidations.
		if entry.LastModified.Add(c.ttl).After(time.Now()) {
			cacheHitCounter.Inc()
			return entry.LastModified
		}
	}

	c.InvalidateIfNeeded(ctx, storeID)

	// This means the most recent Changelog entry is older than the CacheController TTL,
	// so we trigger the invalidation workflow but also return the most recent changelog timestamp
	// for use in the check resolver, since the Check Cache TTL is likely longer than the CacheController.
	if entry != nil {
		return entry.LastModified
	}
```

justincoh: Resolving this as I'm heavily modifying this block right now, most of this is going away

internal/utils/typed_sync_map.go
```diff
@@ -0,0 +1,37 @@
...
+	if !loaded {
+		return value, false // return zero value of V and false
+	}
+
+	// Perform the type assertion internally
+	actualValue, ok := v.(V)
+	if !ok {
+		// Handle unexpected type (though this shouldn't happen if usage is correct)
+		return value, false
+	}
```
senojj: This changes the meaning of `sync.Map::Load` entirely, because now the returned `bool`'s meaning is muddled. Was the key loaded, or was there just a type assertion issue? The difference could be crucial to the caller's logic.

internal/cachecontroller/cache_controller.go
```diff
@@ -27,6 +27,9 @@ import (
 var (
 	tracer = otel.Tracer("internal/cachecontroller")
 
+	// FallbackTime is a signal to the cached check resolver to ignore its cached result.
+	FallbackTime = time.Now()
```
senojj: What makes this `FallbackTime` more meaningful than the zero value of `time.Time`?

justincoh: The zero time is meaningful for cases where the check query cache is enabled but the cache controller **is not**. In those cases, the cached_check_resolver receives the [Zero time as a comparison](https://github.com/openfga/openfga/blob/main/internal/graph/cached_resolver.go#L165), and rightfully determines that it can use any cached result (because they are all newer).

With the cache controller, the goal is to enable the actual check_query_cache TTL to be quite long by having the cache_controller check on it periodically to ensure there have been no new writes. This `FallbackTime` is a message from the cache_controller --> cached_resolver saying "You may have a CheckResponse in cache right now, but I am not 100% certain the cache remains accurate". In this scenario the cached_resolver will ignore its cached entry and resolve the query for real (and cache the new result).

`CACHE_CONTROLLER_TTL` is a bit of a misnomer, it could be more accurately described as "the frequency with which to ensure existing caches are still valid".

internal/utils/typed_sync_map.go
justincoh: I was torn as to whether this extra generic wrapper was necessary for this PR at all. I will remove it

internal/cachecontroller/cache_controller.go
justincoh: Changed this PR to have determine invalidation time return `(time.Time, bool)` to eliminate this confusion

poovamraj:
@claude

internal/cachecontroller/cache_controller.go
```diff
@@ -105,6 +108,9 @@ type InMemoryCacheController struct {
 	inflightInvalidations sync.Map
 	logger                logger.Logger
 
+	// invalidationLog is a map of [storeId] => lastInvalidationTime.
+	invalidationLog sync.Map
```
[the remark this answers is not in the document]
justincoh: `storeID` is pretty low cardinality, don't have to worry about this here.

internal/cachecontroller/cache_controller.go
```diff
@@ -142,20 +159,33 @@ func (c *InMemoryCacheController) DetermineInvalidationTime(
...
-				cacheHitCounter.Inc()
-				span.SetAttributes(attribute.Bool("cached", true))
-				return entry.LastModified
-			}
-		}
+
+	// This should never happen, this cache is only added to within this file.
+	entry, ok := cacheResp.(*storage.ChangelogCacheEntry)
+	if !ok {
+		return lastInvalidTime, isValid
```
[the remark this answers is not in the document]
justincoh: Don't think this is necessary

internal/cachecontroller/cache_controller.go
```diff
@@ -57,9 +57,10 @@ var (
 )
 
 type CacheController interface {
-	// DetermineInvalidationTime returns the timestamp of the last write for the specified store if it was in cache,
-	// Else it returns Zero time and triggers InvalidateIfNeeded().
-	DetermineInvalidationTime(context.Context, string) time.Time
+	// DetermineInvalidationTime returns the timestamp of the last write for the specified store if it was in cache
+	// And a boolean indicating whether this value can be used.
+	// If the value cannot be used, it will trigger InvalidateIfNeeded.
```
adriantam, line 62: This comment is not quite valid - it does not trigger InvalidateIfNeeded.  Instead, it simply causes cache not be used for this specific call.

pkg/server/commands/check_command.go
adriantam, line 1: Do we need to do anything for list objects/list users call?

internal/cachecontroller/cache_controller.go
saad-h1: The `InMemoryCacheController`'s `DetermineInvalidationTime` is [doing that](https://github.com/openfga/openfga/blob/bef1cd2c6ca6f2a428deb71d801ea4461cf7e559/internal/cachecontroller/cache_controller.go#L150-L152)

I am thinking that maybe this comment should be moved to that function though instead of the interface

internal/cachecontroller/cache_controller.go
```diff
@@ -127,13 +133,24 @@ func NewCacheController(ds storage.OpenFGADatastore, cache storage.InMemoryCache
...
+// The boolean would be false in cases where a store has not had its changelog read in longer than `TTL` time.
 func (c *InMemoryCacheController) DetermineInvalidationTime(
 	ctx context.Context,
 	storeID string,
-) time.Time {
+) (lastInvalidTime time.Time, isValid bool) {
 	ctx, span := tracer.Start(ctx, "cacheController.DetermineInvalidationTime", trace.WithAttributes(attribute.Bool("cached", false)))
-	defer span.End()
+
+	defer func() {
```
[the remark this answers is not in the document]
saad-h1: There is a sentence about this:
```
		// This relies on updating the named return value of this function.
```

internal/cachecontroller/cache_controller.go
```diff
@@ -142,20 +160,33 @@ func (c *InMemoryCacheController) DetermineInvalidationTime(
...
 
-	c.InvalidateIfNeeded(ctx, storeID)
+	// Check when this store was last invalidated
+	lastInvalidationTime, ok := lastInvalidation.(time.Time)
+	if !ok {
+		return lastInvalidTime, isValid
+	}
+
+	// If it was within the CacheController TTL, return the most recent changelog timestamp
+	if lastInvalidationTime.Add(c.ttl).After(time.Now()) {
```
saad-h1, line 182: Once `c.ttl` (cache controller TTL) has elapsed since the last time we triggered invalidation, we will return `isValid` = `false` and trigger invalidation async. Since we return false, we force the `CachedCheckResolver` to go to the datastore instead of using the cache. So essentially every cache controller TTL, we will go to the datastore for the Check response instead of cache.

That seems to defeat the purpose of the cache controller, since we want to avoid going to the datastore up to the cache TTL if we can afford to, but this wouldn't allow that. It seems to basically accomplish the same as simply setting the cache TTL low and not using the cache controller at all. We may need to rethink our approach.

justincoh:
Closing this in favor of #2811

PR #2793, opened 2025-11-06
Respect context cancelation in pipeline messaging

senojj:
## Description

Currently, a pipeline can deadlock with no ability to recover when a tuple cycle is present and the number of inflight messages never reaches 0 due to some potential bug.

To ensure that such a scenario will not persist until the process is terminated, a change is made to respect context cancelation within the pipeline's messaging system.

Making this change will allow a context cancelation to trigger a pipeline shutdown through its watchdog goroutine.

In practice, a context with a specified deadline will be passed to the pipeline's `Build` function. If a pipeline's lifetime exceeds that of the context, the pipeline will begin its shutdown procedure automatically.

## Review Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved cancellation handling and context propagation throughout message streaming and pipeline operations for more graceful shutdown behavior.
  * Enhanced resource cleanup by implementing context-aware APIs across producer and consumer components.

* **Tests**
  * Added goroutine leak detection tests to verify proper resource cleanup.
  * Added cancellation behavior validation tests to ensure correct early termination handling.

### DIFF
--- a/pkg/server/commands/reverseexpand/pipeline.go
+++ b/pkg/server/commands/reverseexpand/pipeline.go
@@ -1328,6 +1328,10 @@ func (s *sender) edge() *Edge {
 }
 
 func (s *sender) loop(fn loopFunc) {
+	// A background context is used here because the sender's underlying pipe
+	// relies on a cancelation from its consuming end for a shutdown signal.
+	// This sequence does not need an additional cancelation signal for that
+	// reason.
 	for msg := range s.prod.seq(context.Background()) {
 		if !fn(msg) {
 			break


pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -1320,7 +1328,7 @@ func (s *sender) edge() *Edge {
 }
 
 func (s *sender) loop(fn loopFunc) {
-	for msg := range s.prod.seq() {
+	for msg := range s.prod.seq(context.Background()) {
```
adriantam, line 1335: Q: Do we need to use a background context + some reasonable deadline (10 minutes? 30 minutes?)

Otherwise, this means that there is possibility that the seq can hang.

senojj: We do not want an external cancelation interfering with a sender's lifetime. Each sender's lifetime is managed by the parent worker, and will be ended appropriately when its own internal context is canceled.

This prevents race conditions during pipeline shutdown, and allows workers to drain appropriately.

pkg/server/commands/reverseexpand/pipeline_test.go
```diff
@@ -2007,4 +2010,129 @@ func TestPipeline(t *testing.T) {
...
+		ctx, cancel := context.WithCancel(context.Background())
+
+		seq := pipeline.Build(ctx, source, target)
+
+		cancel()
+
+		for range seq {
+			t.Log("iteration did not stop after context cancelation")
+			t.Fail()
+		}
```
justincoh, line 2090: is this deterministic? Could `Build()` manage to find an item or two before `cancel()` stops it?

senojj: You were absolutely right to question this. I just made a change that makes this deterministic. See commit 11eaaba

PR #2811, opened 2025-11-18
fix: InMemoryCacheController invalidation timing

saad-h1:
## Description

This PR supercedes https://github.com/openfga/openfga/pull/2792. The background is the same but the solution is different. Copying the background from that PR below but reworded with Gemini:

### Context

We use an optional `CacheController` to maintain cache validity. This allows us to keep Check Cache TTLs long (for high hit rates) while using a shorter "Controller TTL" to frequently verify if data has changed.

### The Problem

Previously, the logic conflated "Data Freshness" with "Check Frequency," leading to stale reads for stores with infrequent writes.

`DetermineInvalidationTime` would [only](https://github.com/openfga/openfga/blob/26a9e7d6241651319deea080dab734e344626d59/internal/cachecontroller/cache_controller.go#L148) return the `LastModified` timestamp if that modification happened *recently* (within the `CacheControllerTTL`).

1.  **False Negative:** If a store hadn't been written to in hours (exceeding the TTL), the controller returned `time.Time{}` (Zero Time).
2.  **Stale Resolution:** The Check Resolver [compares](https://github.com/openfga/openfga/blob/26a9e7d6241651319deea080dab734e344626d59/internal/graph/cached_resolver.go#L165) `CachedCheckTime > InvalidationTime`. Since `CachedCheckTime` is always greater than `Zero Time`, the resolver assumed the cached check was valid, even if it was stale.

It was effectively saying: **"Because the data hasn't changed recently (Frequency), I will act as if I don't know when it changed at all (Freshness).**"

By returning `0` (which implies "no data/invalid"), the system threw away the valid `LastModified` timestamp just because it was old. This caused the Check Resolver to assume its cached value was always newer than `0`, resulting in the stale read.

### The Solution

This PR decouples the **invalidation frequency** from the **last modification time**.

1.  **Always Return Last Known State:** The controller now always returns the `ChangelogCacheEntry.LastModified` time if it exists in memory, regardless of how old it is. This ensures the Check Resolver always has a valid anchor to compare against.
    * **Handling Misses:** If the entry is missing, we return `Zero Time` (triggering async invalidation). This is safe because if the `ChangelogCacheEntry` is missing or expired, the corresponding Check Cache entry (which shares the same TTL) is also expired or non-existent (see point 3 below). Therefore, the resolver is forced to hit the database regardless of the `Zero Time` return.
2.  **New `LastChecked` Field:** We introduced a `LastChecked` field to the `ChangelogCacheEntry`. This tracks when we last queried the database for invalidations.
    * We now trigger background invalidation only when `LastChecked` exceeds the `CacheControllerTTL`.
3.  **Extended Lifecycle:** The `ChangelogCacheEntry` TTL is now increased to match the Check Query Cache TTL.
    * *Why?* If the Check Cache entry is alive, we need the Changelog entry to validate it. If the Check Cache entry expires, we don't care about the Changelog entry anymore (because we have to hit the DB anyway).

### Caveats: "First Check" Staleness

Because invalidation is triggered asynchronously to keep the read path fast, there is an accepted race condition where the **first check after a write** (and subsequent checks until the async job finishes) could return stale data, if its Check response was cached before the write.

**Timeline of the Race Condition:**

| Time | Event | Result | State |
| :--- | :--- | :--- | :--- |
| t=1 | **Write:** Tuple added| Store updated | |
| t=2 | **Check A:** Request received | Returns `allowed: true` | Cached & Invalidation Triggered (Async) |
| t=3 | **Async Job (from Check A):** Completes | `LastModified` updated to t=1 | Cache Controller Updated |
| t=4 | **Write:** Tuple deleted | Store updated | Check A is now technically invalid |
| t=5 | **Check B:** Request received | **Returns STALE `allowed: true`** | Controller returns t=1. <br> t=1 < t=2 (Cached Check), so Cache wins. <br> Invalidation Triggered (Async). |
| t=6 | **Async Job (from Check B):** Completes | `LastModified` updated to t=4 | Cache Controller Updated |
| t=7 | **Check C:** Request received | **Returns FRESH `allowed: false`** | Controller returns t=4. <br> t=4 > t=2 (Cached Check), so Cache Invalidated. |

**Conclusion:** The staleness at `t=5` is a known trade-off for our current distributed, in-memory architecture. It is mitigated by the expectation that Checks happen frequently (so the "staleness" would likely still be in the Cache Controller TTL window) and the Cache Controller TTL (polling frequency) is low. Also, *any* Check *or* ListObjects to the store will trigger the asynchronous invalidation work to invalidate *any* earlier cached Check responses (as long as it's been more than the cache controller TTL since the last time we triggered invalidation).

The only way this would be a problem is if a Check response was cached --> there was a write to the store --> there was not a single LO or Check across the entire store for the entire cache controller TTL duration for a given replica --> a Check comes in whose response was previously cached.

### Other Options

A potential workaround which would avoid the caveat of the "first Check" staleness was to have a perpetual running background job to query the datastore every cache controller TTL to set the `ChangelogCacheEntry` - but given this entry is per store ID and openfga may run across multiple replicas, the background task would need to keep track of and query the datastore for every store ID, across every replica, every cache controller TTL. We opted away from that, given that in real production usage, there should be regular enough Checks that the aforementioned caveat isn't really an issue.

We cannot update the `ChangelogCacheEntry` (used for invalidation) after every write since the cache is in-memory, and if openfga is running across multiple replicas, the replica that gets the write may not be the replica that gets the Check, so the latter won't be able to see the cached `ChangelogCacheEntry`. A distributed cache would have solved this, but the current code only supports in-memory caching and a distributed cache would add (potentially significant) network latency.

We cannot wait for invalidation to finish before returning the Check response because invalidation needs to query the datastore (high latency compared to cache), which defeats the purpose of caching (we may as well just go to the datastore every time to compute the result).

## References

n/a

## Testing

Using in-memory datastore, cache TTL of 60s, cache controller TTL of 5s.

### Typical Behaviour

Doing a Check (to cache the response) --> delete tuple --> multiple Checks until we no longer get stale response --> write tuple --> multiple Checks until we no longer get stale response:

```log
[10:04:02] Current State ...
{
  "allowed":true,
  "resolution":""
}

[10:04:02] Deleting tuple ...
{}

[10:04:02] Expecting false ... (iteration 1/50)
{
  "allowed":true,
  "resolution":""
}

[10:04:03] Expecting false ... (iteration 2/50)
{
  "allowed":true,
  "resolution":""
}

[10:04:04] Expecting false ... (iteration 3/50)
{
  "allowed":true,
  "resolution":""
}

[10:04:05] Expecting false ... (iteration 4/50)
{
  "allowed":true,
  "resolution":""
}

[10:04:06] Expecting false ... (iteration 5/50)
{
  "allowed":true,
  "resolution":""
}

[10:04:08] Expecting false ... (iteration 6/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:09] Expecting false ... (iteration 7/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:10] Expecting false ... (iteration 8/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:11] Expecting false ... (iteration 9/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:12] Expecting false ... (iteration 10/50)
{
  "allowed":false,
  "resolution":""
}

...


[10:04:55] Adding tuple ...
{
  "successful": [
    {
      "object":"organization:cups",
      "relation":"member",
      "user":"user:charles"
    }
  ]
}

[10:04:55] Expecting true ... (iteration 1/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:56] Expecting true ... (iteration 2/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:57] Expecting true ... (iteration 3/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:58] Expecting true ... (iteration 4/50)
{
  "allowed":false,
  "resolution":""
}

[10:04:59] Expecting true ... (iteration 5/50)
{
  "allowed":false,
  "resolution":""
}

[10:05:00] Expecting true ... (iteration 6/50)
{
  "allowed":true,
  "resolution":""
}

[10:05:01] Expecting true ... (iteration 7/50)
{
  "allowed":true,
  "resolution":""
}

[10:05:02] Expecting true ... (iteration 8/50)
{
  "allowed":true,
  "resolution":""
}

[10:05:04] Expecting true ... (iteration 9/50)
{
  "allowed":true,
  "resolution":""
}

[10:05:05] Expecting true ... (iteration 10/50)
{
  "allowed":true,
  "resolution":""
}

...
```

Note that we only get stale responses for the cache controller TTL (5s).

### Deeper Debugging

Debug logs to see what's happening under the hood more:

Write tuple:

```log
openfga         | 2025-11-18T14:07:01.396Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Write", "grpc_type": "unary", "trace_id": "854848ada8d59dc3d3df5ec58cf5a7d5", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","writes":{"tuple_keys":[{"user":"organization:brayer","relation":"subscriber","object":"plan:team","condition":null}],"on_duplicate":"ignore"},"deletes":null,"authorization_model_id":""}, "raw_response": {}, "query_duration_ms": "0", "request_id": "854848ada8d59dc3d3df5ec58cf5a7d5", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "peer.address": "127.0.0.1:45028", "grpc_code": 0}
```

First Check - goes to datastore, response `true` is cached, invalidation triggered which will asynchronously cache the last write time:

```log
openfga         | 2025-11-18T14:07:09.226Z	DEBUG	InMemoryCacheController DetermineInvalidationTime cache attempt	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "hit": false}
openfga         | 2025-11-18T14:07:09.226Z	DEBUG	🔴 DetermineInvalidationTime | changelog cache entry is nil, triggering invalidation
openfga         | 2025-11-18T14:07:09.226Z	DEBUG	🟤 CachedCheckResolver not found cache key	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "tuple_key": "user:\"user:charles\"  relation:\"can_access\"  object:\"feature:draft_prs\""}
openfga         | 2025-11-18T14:07:09.226Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.r/01KABMMM5PV8TF0CQD7SKF0BTQ/feature:draft_prs#associated_plan@"}
openfga         | 2025-11-18T14:07:09.226Z	DEBUG	InMemoryCacheController findChangesAndInvalidateIfNecessary invalidation	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "entry.LastModified": "2025-11-18T14:07:01.396Z", "timestampOfLastIteratorInvalidation": "2025-11-18T14:06:09.226Z", "invalidationType": "full"}
openfga         | 2025-11-18T14:07:09.227Z	DEBUG	🟤 CachedCheckResolver not found cache key	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "tuple_key": "user:\"user:charles\"  relation:\"subscriber_member\"  object:\"plan:enterprise\""}
openfga         | 2025-11-18T14:07:09.227Z	DEBUG	🟤 CachedCheckResolver not found cache key	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "tuple_key": "user:\"user:charles\"  relation:\"subscriber_member\"  object:\"plan:team\""}
openfga         | 2025-11-18T14:07:09.227Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.r/01KABMMM5PV8TF0CQD7SKF0BTQ/plan:enterprise#subscriber@"}
openfga         | 2025-11-18T14:07:09.227Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.r/01KABMMM5PV8TF0CQD7SKF0BTQ/plan:team#subscriber@"}
openfga         | 2025-11-18T14:07:09.227Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.rtwu/01KABMMM5PV8TF0CQD7SKF0BTQ/organization#member/user:charles"}
openfga         | 2025-11-18T14:07:09.227Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Check", "grpc_type": "unary", "trace_id": "88b1f8ac67d6ba8a4c8830f71234b60e", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","tuple_key":{"user":"user:charles","relation":"can_access","object":"feature:draft_prs"},"contextual_tuples":{"tuple_keys":[]},"authorization_model_id":"","trace":false,"context":{},"consistency":"UNSPECIFIED"}, "raw_response": {"allowed":true,"resolution":""}, "query_duration_ms": "1", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "dispatch_count": 2, "datastore_query_count": 4, "datastore_item_count": 4, "request.throttled": false, "peer.address": "127.0.0.1:45028", "request_id": "88b1f8ac67d6ba8a4c8830f71234b60e", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "grpc_code": 0}
```

Delete tuple (previous Check should return false):

```log
openfga         | 2025-11-18T14:07:14.924Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Write", "grpc_type": "unary", "trace_id": "5c915774a0566f3736077795c1003eff", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","writes":null,"deletes":{"tuple_keys":[{"user":"user:charles","relation":"member","object":"organization:cups"}],"on_missing":""},"authorization_model_id":""}, "raw_response": {}, "query_duration_ms": "0", "peer.address": "127.0.0.1:45028", "request_id": "5c915774a0566f3736077795c1003eff", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "grpc_code": 0}
```

First Check after delete - since it's more than cache controller TTL (5s) since last invalidation, trigger invalidation again async, but in the meantime cache controller returns the last write time it knows (the write before the delete) and this causes the (stale) cached response of `true` to be considered valid still:

```log
openfga         | 2025-11-18T14:07:27.716Z	DEBUG	InMemoryCacheController DetermineInvalidationTime cache attempt	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "hit": true}
openfga         | 2025-11-18T14:07:27.716Z	DEBUG	🟠 DetermineInvalidationTime | c.ttl expired since last check, triggering invalidation	{"lastChecked": "2025-11-18 14:07:09.226790467 +0000 UTC m=+16.056924883", "ttl": "5s"}
openfga         | 2025-11-18T14:07:27.716Z	DEBUG	🔵 CachedCheckResolver.ResolveCheck found cache key	{"isValid": true, "CheckResponseCacheEntry.LastModified": "2025-11-18 14:07:09.2273843 +0000 UTC m=+16.057518717", "LastCacheInvalidationTime": "2025-11-18 14:07:01.396081296 +0000 UTC"}
openfga         | 2025-11-18T14:07:27.716Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Check", "grpc_type": "unary", "trace_id": "8254b20ef5bf2681a86df75c36578c9e", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","tuple_key":{"user":"user:charles","relation":"can_access","object":"feature:draft_prs"},"contextual_tuples":{"tuple_keys":[]},"authorization_model_id":"","trace":false,"context":{},"consistency":"UNSPECIFIED"}, "raw_response": {"allowed":true,"resolution":""}, "query_duration_ms": "0", "datastore_item_count": 0, "request.throttled": false, "peer.address": "127.0.0.1:45028", "request_id": "8254b20ef5bf2681a86df75c36578c9e", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "dispatch_count": 0, "datastore_query_count": 0, "grpc_code": 0}
openfga         | 2025-11-18T14:07:27.716Z	DEBUG	InMemoryCacheController findChangesAndInvalidateIfNecessary invalidation	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "entry.LastModified": "2025-11-18T14:07:14.924Z", "timestampOfLastIteratorInvalidation": "2025-11-18T14:06:27.716Z", "invalidationType": "full"}
```

Next Check - sees the updated changelog cache entry (for the delete) triggered by the previous Check, compares against the cached response, successfully invalidates and goes to the datastore, returning `false` (also triggers invalidation async as usual):

```log
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	InMemoryCacheController DetermineInvalidationTime cache attempt	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "hit": true}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	🟠 DetermineInvalidationTime | c.ttl expired since last check, triggering invalidation	{"lastChecked": "2025-11-18 14:07:27.716636378 +0000 UTC m=+34.547404558", "ttl": "5s"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	🔵 CachedCheckResolver.ResolveCheck found cache key	{"isValid": false, "CheckResponseCacheEntry.LastModified": "2025-11-18 14:07:09.2273843 +0000 UTC m=+16.057518717", "LastCacheInvalidationTime": "2025-11-18 14:07:14.924414386 +0000 UTC"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.r/01KABMMM5PV8TF0CQD7SKF0BTQ/feature:draft_prs#associated_plan@"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	InMemoryCacheController findChangesAndInvalidateIfNecessary no invalidation as entry.LastModified before last verified	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "entry.LastModified": "2025-11-18T14:07:14.924Z", "timestampOfLastInvalidation": "2025-11-18T14:07:14.924Z"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	🔵 CachedCheckResolver.ResolveCheck found cache key	{"isValid": false, "CheckResponseCacheEntry.LastModified": "2025-11-18 14:07:09.227365008 +0000 UTC m=+16.057499425", "LastCacheInvalidationTime": "2025-11-18 14:07:14.924414386 +0000 UTC"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.r/01KABMMM5PV8TF0CQD7SKF0BTQ/plan:enterprise#subscriber@"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.rtwu/01KABMMM5PV8TF0CQD7SKF0BTQ/organization#member/user:charles"}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	🟤 CachedCheckResolver not found cache key	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "tuple_key": "user:\"user:charles\"  relation:\"subscriber_member\"  object:\"plan:team\""}
openfga         | 2025-11-18T14:07:41.928Z	DEBUG	cachedIterator flush and update cache for 	{"cacheKey": "ic.r/01KABMMM5PV8TF0CQD7SKF0BTQ/plan:team#subscriber@"}
openfga         | 2025-11-18T14:07:41.928Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Check", "grpc_type": "unary", "trace_id": "340122a301b9e64c72bd074d87e967b1", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","tuple_key":{"user":"user:charles","relation":"can_access","object":"feature:draft_prs"},"contextual_tuples":{"tuple_keys":[]},"authorization_model_id":"","trace":false,"context":{},"consistency":"UNSPECIFIED"}, "raw_response": {"allowed":false,"resolution":""}, "query_duration_ms": "0", "datastore_item_count": 4, "request.throttled": false, "peer.address": "127.0.0.1:45028", "request_id": "340122a301b9e64c72bd074d87e967b1", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "dispatch_count": 2, "datastore_query_count": 4, "grpc_code": 0}
```

Also, if we do a few Checks in quick succession, they don't trigger invalidation since we haven't exceeded the cache controller TTL:

```log
openfga         | 2025-11-18T14:07:57.820Z	DEBUG	InMemoryCacheController DetermineInvalidationTime cache attempt	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "hit": true}
openfga         | 2025-11-18T14:07:57.820Z	DEBUG	🟢 DetermineInvalidationTime | no invalidation needed - changelog cache entry found and we're within c.ttl
openfga         | 2025-11-18T14:07:57.820Z	DEBUG	🔵 CachedCheckResolver.ResolveCheck found cache key	{"isValid": true, "CheckResponseCacheEntry.LastModified": "2025-11-18 14:07:41.92888076 +0000 UTC m=+48.759648981", "LastCacheInvalidationTime": "2025-11-18 14:07:14.924414386 +0000 UTC"}
openfga         | 2025-11-18T14:07:57.820Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Check", "grpc_type": "unary", "trace_id": "303bb74beeca8ad9ce8335b2185e26ef", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","tuple_key":{"user":"user:charles","relation":"can_access","object":"feature:draft_prs"},"contextual_tuples":{"tuple_keys":[]},"authorization_model_id":"","trace":false,"context":{},"consistency":"UNSPECIFIED"}, "raw_response": {"allowed":false,"resolution":""}, "query_duration_ms": "0", "peer.address": "127.0.0.1:45028", "request_id": "303bb74beeca8ad9ce8335b2185e26ef", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "dispatch_count": 0, "datastore_query_count": 0, "datastore_item_count": 0, "request.throttled": false, "grpc_code": 0}

openfga         | 2025-11-18T14:07:59.428Z	DEBUG	InMemoryCacheController DetermineInvalidationTime cache attempt	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "hit": true}
openfga         | 2025-11-18T14:07:59.428Z	DEBUG	🟢 DetermineInvalidationTime | no invalidation needed - changelog cache entry found and we're within c.ttl
openfga         | 2025-11-18T14:07:59.429Z	DEBUG	🔵 CachedCheckResolver.ResolveCheck found cache key	{"isValid": true, "CheckResponseCacheEntry.LastModified": "2025-11-18 14:07:41.92888076 +0000 UTC m=+48.759648981", "LastCacheInvalidationTime": "2025-11-18 14:07:14.924414386 +0000 UTC"}
openfga         | 2025-11-18T14:07:59.429Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Check", "grpc_type": "unary", "trace_id": "5cd4d1f716a08f67a05c730c74ceabaf", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","tuple_key":{"user":"user:charles","relation":"can_access","object":"feature:draft_prs"},"contextual_tuples":{"tuple_keys":[]},"authorization_model_id":"","trace":false,"context":{},"consistency":"UNSPECIFIED"}, "raw_response": {"allowed":false,"resolution":""}, "query_duration_ms": "0", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "dispatch_count": 0, "datastore_query_count": 0, "datastore_item_count": 0, "request.throttled": false, "peer.address": "127.0.0.1:45028", "request_id": "5cd4d1f716a08f67a05c730c74ceabaf", "grpc_code": 0}

openfga         | 2025-11-18T14:08:01.126Z	DEBUG	InMemoryCacheController DetermineInvalidationTime cache attempt	{"store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "hit": true}
openfga         | 2025-11-18T14:08:01.126Z	DEBUG	🟢 DetermineInvalidationTime | no invalidation needed - changelog cache entry found and we're within c.ttl
openfga         | 2025-11-18T14:08:01.126Z	DEBUG	🔵 CachedCheckResolver.ResolveCheck found cache key	{"isValid": true, "CheckResponseCacheEntry.LastModified": "2025-11-18 14:07:41.92888076 +0000 UTC m=+48.759648981", "LastCacheInvalidationTime": "2025-11-18 14:07:14.924414386 +0000 UTC"}
openfga         | 2025-11-18T14:08:01.126Z	INFO	grpc_req_complete	{"grpc_service": "openfga.v1.OpenFGAService", "grpc_method": "Check", "grpc_type": "unary", "trace_id": "c6f2114792ccc1d3d2def796ad4401f6", "user_agent": "openfga-cli/0.7.5", "raw_request": {"store_id":"01KABMMM5PV8TF0CQD7SKF0BTQ","tuple_key":{"user":"user:charles","relation":"can_access","object":"feature:draft_prs"},"contextual_tuples":{"tuple_keys":[]},"authorization_model_id":"","trace":false,"context":{},"consistency":"UNSPECIFIED"}, "raw_response": {"allowed":false,"resolution":""}, "query_duration_ms": "0", "datastore_item_count": 0, "request.throttled": false, "peer.address": "127.0.0.1:45028", "request_id": "c6f2114792ccc1d3d2def796ad4401f6", "store_id": "01KABMMM5PV8TF0CQD7SKF0BTQ", "authorization_model_id": "01KABMMM608FVMWG607MBJ1P0J", "dispatch_count": 0, "datastore_query_count": 0, "grpc_code": 0}
```

### *Any* Check to the store will invalidate *any* earlier cached Check response

```sh
╭───────────────────────────────────────────
│ ~/saad/stores/entitlements
╰─> fga query check user:charles can_access feature:draft_prs
{
  "allowed":true,
  "resolution":""
}

╭───────────────────────────────────────────
│ ~/saad/stores/entitlements
╰─> fga tuple delete user:charles member organization:cups
{}

╭───────────────────────────────────────────
│ ~/saad/stores/entitlements
╰─> fga query check user:anne can_access feature:issues
{
  "allowed":true,
  "resolution":""
}

╭───────────────────────────────────────────
│ ~/saad/stores/entitlements
╰─> fga query check user:charles can_access feature:draft_prs
{
  "allowed":false,
  "resolution":""
}
```

After the delete, we did a Check on `user:anne can_access feature:issues`. That triggered invalidation which wrote the `ChangelogCacheEntry` for the storeID. Then the Check for `user:charles can_access feature:draft_prs` (whose result was previously cached) could use that `ChangelogCacheEntry.LastModified` time to properly invalidate the cache and go to the datastore, returning false.

### List Objects can instantly trigger invalidation

When list objects iterator caching is enabled, it always runs invalidation instantly. This invalidation applies to Checks as well:

```sh
[14:47:36] Current State ...
{
  "allowed":true,
  "resolution":""
}


[14:47:36] Deleting tuple ...
{}


[14:47:36] List Objects (triggers invalidation)
{
  "objects": []
}


[14:47:38] Check - expecting false
{
  "allowed":false,
  "resolution":""
}


[14:47:38] Adding tuple ...
{
  "successful": [
    {
      "object":"organization:cups",
      "relation":"member",
      "user":"user:charles"
    }
  ]
}


[14:47:38] List Objects (triggers invalidation)
{
  "objects": [
    "feature:draft_prs",
    "feature:sso",
    "feature:issues"
  ]
}


[14:47:39] Check - expecting true
{
  "allowed":true,
  "resolution":""
}
```

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a regression where cached Check responses were not properly invalidated after writes to the store; invalidation now occurs consistently in all scenarios.

* **Documentation**
  * Enhanced cache configuration help text with clarifications on TTL-based invalidation behavior and frequency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 ## [1.11.1] - 2025-11-20
 ### Added
 - Added experimental list objects pipeline algorithm with `pipeline_list_objects` flag. [#2815](https://github.com/openfga/openfga/pull/2815)
+- Differentiate between dispatch throttled operations and datastore throttled operations. [#2821](https://github.com/openfga/openfga/pull/2821)
 
 ### Fixed
 - Fixed bug where public wildcard tuples without condition were being considered for a wildcard relation which was only defined with a condition. [CVE-2025-64751](https://github.com/openfga/openfga/security/advisories/GHSA-2c64-vmv2-hgfc)
--- a/internal/graph/dispatch_throttling_check_resolver.go
+++ b/internal/graph/dispatch_throttling_check_resolver.go
@@ -104,7 +104,7 @@ func (r *DispatchThrottlingCheckResolver) ResolveCheck(ctx context.Context,
 		attribute.Bool("is_throttled", shouldThrottle))
 
 	if shouldThrottle {
-		req.GetRequestMetadata().WasThrottled.Store(true)
+		req.GetRequestMetadata().DispatchThrottled.Store(true)
 		r.throttler.Throttle(ctx)
 	}
 	return r.delegate.ResolveCheck(ctx, req)
--- a/internal/graph/dispatch_throttling_check_resolver_test.go
+++ b/internal/graph/dispatch_throttling_check_resolver_test.go
@@ -49,7 +49,7 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		_, err := dut.ResolveCheck(ctx, req)
 		require.NoError(t, err)
 
-		require.False(t, req.GetRequestMetadata().WasThrottled.Load())
+		require.False(t, req.GetRequestMetadata().DispatchThrottled.Load())
 	})
 
 	t.Run("above_threshold_should_call_throttle", func(t *testing.T) {
@@ -85,7 +85,7 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		_, err := dut.ResolveCheck(ctx, req)
 		require.NoError(t, err)
 
-		require.True(t, req.GetRequestMetadata().WasThrottled.Load())
+		require.True(t, req.GetRequestMetadata().DispatchThrottled.Load())
 	})
 
 	t.Run("zero_max_should_interpret_as_default", func(t *testing.T) {
@@ -121,7 +121,7 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		_, err := dut.ResolveCheck(ctx, req)
 		require.NoError(t, err)
 
-		require.False(t, req.GetRequestMetadata().WasThrottled.Load())
+		require.False(t, req.GetRequestMetadata().DispatchThrottled.Load())
 	})
 
 	t.Run("dispatch_should_use_request_threshold_if_available", func(t *testing.T) {
@@ -158,7 +158,7 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		_, err := dut.ResolveCheck(ctx, req)
 		require.NoError(t, err)
 
-		require.True(t, req.GetRequestMetadata().WasThrottled.Load())
+		require.True(t, req.GetRequestMetadata().DispatchThrottled.Load())
 	})
 
 	t.Run("should_respect_max_threshold", func(t *testing.T) {
@@ -195,6 +195,6 @@ func TestDispatchThrottlingCheckResolver(t *testing.T) {
 		_, err := dut.ResolveCheck(ctx, req)
 		require.NoError(t, err)
 
-		require.True(t, req.GetRequestMetadata().WasThrottled.Load())
+		require.True(t, req.GetRequestMetadata().DispatchThrottled.Load())
 	})
 }
--- a/internal/graph/resolve_check_request.go
+++ b/internal/graph/resolve_check_request.go
@@ -44,8 +44,11 @@ type ResolveCheckRequestMetadata struct {
 	// After the root problem has been solved, this value can be read.
 	DispatchCounter *atomic.Uint32
 
-	// WasThrottled indicates whether the request was throttled
-	WasThrottled *atomic.Bool
+	// DispatchThrottled indicates whether the request was throttled by the dispatch throttling check resolver
+	DispatchThrottled *atomic.Bool
+
+	// DatastoreThrottled indicates whether this request was throttled at the datastore level.
+	DatastoreThrottled *atomic.Bool
 }
 
 type ResolveCheckRequestParams struct {
@@ -60,8 +63,9 @@ type ResolveCheckRequestParams struct {
 
 func NewCheckRequestMetadata() *ResolveCheckRequestMetadata {
 	return &ResolveCheckRequestMetadata{
-		DispatchCounter: new(atomic.Uint32),
-		WasThrottled:    new(atomic.Bool),
+		DispatchCounter:    new(atomic.Uint32),
+		DispatchThrottled:  new(atomic.Bool),
+		DatastoreThrottled: new(atomic.Bool),
 	}
 }
 
@@ -110,9 +114,10 @@ func (r *ResolveCheckRequest) clone() *ResolveCheckRequest {
 	origRequestMetadata := r.GetRequestMetadata()
 	if origRequestMetadata != nil {
 		requestMetadata = &ResolveCheckRequestMetadata{
-			DispatchCounter: origRequestMetadata.DispatchCounter,
-			Depth:           origRequestMetadata.Depth,
-			WasThrottled:    origRequestMetadata.WasThrottled,
+			DispatchCounter:    origRequestMetadata.DispatchCounter,
+			Depth:              origRequestMetadata.Depth,
+			DispatchThrottled:  origRequestMetadata.DispatchThrottled,
+			DatastoreThrottled: origRequestMetadata.DatastoreThrottled,
 		}
 	}
 
--- a/internal/graph/resolve_check_request_test.go
+++ b/internal/graph/resolve_check_request_test.go
@@ -46,7 +46,8 @@ func TestCloneResolveCheckRequest(t *testing.T) {
 		require.Equal(t, contextStruct, orig.GetContext())
 		require.Equal(t, uint32(0), orig.GetRequestMetadata().Depth)
 		require.Equal(t, uint32(2), orig.GetRequestMetadata().DispatchCounter.Load())
-		require.False(t, orig.GetRequestMetadata().WasThrottled.Load())
+		require.False(t, orig.GetRequestMetadata().DispatchThrottled.Load())
+		require.False(t, orig.GetRequestMetadata().DatastoreThrottled.Load())
 		require.Equal(t, map[string]struct{}{
 			"abc": {},
 		}, orig.VisitedPaths)
@@ -58,7 +59,7 @@ func TestCloneResolveCheckRequest(t *testing.T) {
 			"abc": {},
 			"xyz": {},
 		}
-		orig.GetRequestMetadata().WasThrottled.Store(true)
+		orig.GetRequestMetadata().DispatchThrottled.Store(true)
 
 		// Assert the new values of the orig
 		require.Equal(t, "12", orig.GetStoreID())
@@ -73,7 +74,7 @@ func TestCloneResolveCheckRequest(t *testing.T) {
 		require.Equal(t, contextStruct, orig.GetContext())
 		require.Equal(t, uint32(0), orig.GetRequestMetadata().Depth)
 		require.Equal(t, uint32(7), orig.GetRequestMetadata().DispatchCounter.Load())
-		require.True(t, orig.GetRequestMetadata().WasThrottled.Load())
+		require.True(t, orig.GetRequestMetadata().DispatchThrottled.Load())
 		require.Equal(t, map[string]struct{}{
 			"abc": {},
 			"xyz": {},
@@ -91,7 +92,7 @@ func TestCloneResolveCheckRequest(t *testing.T) {
 		require.Equal(t, contextStruct, cloned.GetContext())
 		require.Equal(t, uint32(0), cloned.GetRequestMetadata().Depth)
 		require.Equal(t, uint32(7), cloned.GetRequestMetadata().DispatchCounter.Load()) // note that it is intended to have the request metadata share the same dispatch counter
-		require.True(t, cloned.GetRequestMetadata().WasThrottled.Load())                // it is intended to share the same was throttled state
+		require.True(t, cloned.GetRequestMetadata().DispatchThrottled.Load())           // it is intended to share the same was throttled state
 		require.Equal(t, map[string]struct{}{
 			"abc": {},
 		}, cloned.VisitedPaths)
@@ -160,7 +161,8 @@ func TestCloneResolveCheckRequest(t *testing.T) {
 			require.Equal(t, "document:def", cloned.GetContextualTuples()[0].GetObject())
 			require.Equal(t, contextStruct, cloned.GetContext())
 			require.Equal(t, uint32(0), cloned.GetRequestMetadata().Depth)
-			require.False(t, cloned.GetRequestMetadata().WasThrottled.Load())
+			require.False(t, cloned.GetRequestMetadata().DispatchThrottled.Load())
+			require.False(t, cloned.GetRequestMetadata().DatastoreThrottled.Load())
 			require.Equal(t, map[string]struct{}{
 				"abc": {},
 			}, cloned.VisitedPaths)
--- a/pkg/server/check.go
+++ b/pkg/server/check.go
@@ -95,12 +95,14 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 	endTime := time.Since(startTime).Milliseconds()
 
 	var (
-		wasRequestThrottled bool
-		rawDispatchCount    uint32
+		dispatchThrottled  bool
+		datastoreThrottled bool
+		rawDispatchCount   uint32
 	)
 
 	if checkRequestMetadata != nil {
-		wasRequestThrottled = checkRequestMetadata.WasThrottled.Load()
+		dispatchThrottled = checkRequestMetadata.DispatchThrottled.Load()
+		datastoreThrottled = checkRequestMetadata.DatastoreThrottled.Load()
 		rawDispatchCount = checkRequestMetadata.DispatchCounter.Load()
 		dispatchCount := float64(rawDispatchCount)
 
@@ -147,10 +149,11 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 			).Observe(float64(endTime))
 		}
 
-		if wasRequestThrottled {
+		if dispatchThrottled || datastoreThrottled {
 			throttledRequestCounter.WithLabelValues(s.serviceName, methodName).Inc()
 		}
-		grpc_ctxtags.Extract(ctx).Set("request.throttled", wasRequestThrottled)
+		grpc_ctxtags.Extract(ctx).Set("request.dispatch_throttled", dispatchThrottled)
+		grpc_ctxtags.Extract(ctx).Set("request.datastore_throttled", datastoreThrottled)
 	}
 
 	if err != nil {
--- a/pkg/server/commands/batch_check_command.go
+++ b/pkg/server/commands/batch_check_command.go
@@ -212,7 +212,7 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
 			})
 
 			if metadata != nil {
-				if metadata.WasThrottled.Load() {
+				if metadata.DispatchThrottled.Load() || metadata.DatastoreThrottled.Load() {
 					totalThrottleCount.Add(1)
 				}
 				totalDispatchCount.Add(metadata.DispatchCounter.Load())
--- a/pkg/server/commands/check_command.go
+++ b/pkg/server/commands/check_command.go
@@ -165,11 +165,13 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
 	dsMeta := datastoreWithTupleCache.GetMetadata()
 	resp.ResolutionMetadata.DatastoreQueryCount = dsMeta.DatastoreQueryCount
 	resp.ResolutionMetadata.DatastoreItemCount = dsMeta.DatastoreItemCount
-	// Until dispatch throttling is deprecated, merge the results of both
-	resolveCheckRequest.GetRequestMetadata().WasThrottled.CompareAndSwap(false, dsMeta.WasThrottled)
+
+	resolveCheckRequest.GetRequestMetadata().DatastoreThrottled.Store(dsMeta.WasThrottled)
 
 	if err != nil {
-		if errors.Is(err, context.DeadlineExceeded) && resolveCheckRequest.GetRequestMetadata().WasThrottled.Load() {
+		// There are currently two possible throttling mechanisms, we need to know if either was triggered here.
+		wasThrottled := dsMeta.WasThrottled || resolveCheckRequest.GetRequestMetadata().DispatchThrottled.Load()
+		if errors.Is(err, context.DeadlineExceeded) && wasThrottled {
 			return resp, resolveCheckRequest.GetRequestMetadata(), &ThrottledError{Cause: err}
 		}
 
--- a/pkg/server/commands/check_command_test.go
+++ b/pkg/server/commands/check_command_test.go
@@ -227,7 +227,7 @@ type doc
 			DoAndReturn(func(ctx context.Context, req *graph.ResolveCheckRequest) (*graph.ResolveCheckResponse, error) {
 				req.GetRequestMetadata().Depth++
 				req.GetRequestMetadata().DispatchCounter.Add(1)
-				req.GetRequestMetadata().WasThrottled.Store(true)
+				req.GetRequestMetadata().DispatchThrottled.Store(true)
 				ds, _ := storage.RelationshipTupleReaderFromContext(ctx)
 				_, _ = ds.Read(ctx, req.StoreID, storage.ReadFilter{}, storage.ReadOptions{})
 				return nil, context.DeadlineExceeded
@@ -243,7 +243,7 @@ type doc
 		require.Equal(t, uint32(1), checkResp.GetResolutionMetadata().DatastoreQueryCount)
 		require.Equal(t, uint32(1), checkRequestMetadata.Depth)
 		require.Equal(t, uint32(1), checkRequestMetadata.DispatchCounter.Load())
-		require.True(t, checkRequestMetadata.WasThrottled.Load())
+		require.True(t, checkRequestMetadata.DispatchThrottled.Load())
 	})
 }
 
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -100,8 +100,11 @@ type ListObjectsResolutionMetadata struct {
 	// The total number of dispatches aggregated from reverse_expand and check resolutions (if any) to complete the ListObjects request
 	DispatchCounter atomic.Uint32
 
-	// WasThrottled indicates whether the request was throttled
-	WasThrottled atomic.Bool
+	// DispatchThrottled indicates whether this request was throttled by dispatch count.
+	DispatchThrottled atomic.Bool
+
+	// DatastoreThrottled indicates whether the request was throttled by the Datastore.
+	DatastoreThrottled atomic.Bool
 
 	// WasWeightedGraphUsed indicates whether the weighted graph was used as the algorithm for the ListObjects request.
 	WasWeightedGraphUsed atomic.Bool
@@ -371,8 +374,8 @@ func (q *ListObjectsQuery) evaluate(
 				return err
 			}
 			resolutionMetadata.DispatchCounter.Add(reverseExpandResolutionMetadata.DispatchCounter.Load())
-			if !resolutionMetadata.WasThrottled.Load() && reverseExpandResolutionMetadata.WasThrottled.Load() {
-				resolutionMetadata.WasThrottled.Store(true)
+			if !resolutionMetadata.DispatchThrottled.Load() && reverseExpandResolutionMetadata.DispatchThrottled.Load() {
+				resolutionMetadata.DispatchThrottled.Store(true)
 			}
 			resolutionMetadata.CheckCounter.Add(reverseExpandResolutionMetadata.CheckCounter.Load())
 			resolutionMetadata.WasWeightedGraphUsed.Store(reverseExpandResolutionMetadata.WasWeightedGraphUsed.Load())
@@ -434,8 +437,8 @@ func (q *ListObjectsQuery) evaluate(
 					resolutionMetadata.DatastoreQueryCount.Add(resp.GetResolutionMetadata().DatastoreQueryCount)
 					resolutionMetadata.DatastoreItemCount.Add(resp.GetResolutionMetadata().DatastoreItemCount)
 					resolutionMetadata.DispatchCounter.Add(checkRequestMetadata.DispatchCounter.Load())
-					if !resolutionMetadata.WasThrottled.Load() && checkRequestMetadata.WasThrottled.Load() {
-						resolutionMetadata.WasThrottled.Store(true)
+					if !resolutionMetadata.DispatchThrottled.Load() && checkRequestMetadata.DispatchThrottled.Load() {
+						resolutionMetadata.DispatchThrottled.Store(true)
 					}
 					if resp.Allowed {
 						trySendObject(ctx, res.Object, &objectsFound, maxResults, resultsChan)
@@ -456,7 +459,7 @@ func (q *ListObjectsQuery) evaluate(
 		dsMeta := ds.GetMetadata()
 		resolutionMetadata.DatastoreQueryCount.Add(dsMeta.DatastoreQueryCount)
 		resolutionMetadata.DatastoreItemCount.Add(dsMeta.DatastoreItemCount)
-		resolutionMetadata.WasThrottled.CompareAndSwap(false, dsMeta.WasThrottled)
+		resolutionMetadata.DatastoreThrottled.Store(dsMeta.WasThrottled)
 	}
 
 	go handler()
--- a/pkg/server/commands/list_objects_test.go
+++ b/pkg/server/commands/list_objects_test.go
@@ -320,7 +320,7 @@ func TestListObjectsDispatchCount(t *testing.T) {
 			require.NoError(t, err)
 
 			require.Equal(t, test.expectedDispatchCount, resp.ResolutionMetadata.DispatchCounter.Load())
-			require.Equal(t, test.expectedThrottlingValue > 0, resp.ResolutionMetadata.WasThrottled.Load())
+			require.Equal(t, test.expectedThrottlingValue > 0, resp.ResolutionMetadata.DispatchThrottled.Load())
 		})
 	}
 }
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -224,8 +224,8 @@ type ResolutionMetadata struct {
 	// The number of times we are expanding from each node to find set of objects
 	DispatchCounter *atomic.Uint32
 
-	// WasThrottled indicates whether the request was throttled
-	WasThrottled *atomic.Bool
+	// DispatchThrottled indicates whether the request was throttled by dispatch count
+	DispatchThrottled *atomic.Bool
 
 	// WasWeightedGraphUsed indicates whether the weighted graph was used as the algorithm for the ReverseExpand request.
 	WasWeightedGraphUsed *atomic.Bool
@@ -237,7 +237,7 @@ type ResolutionMetadata struct {
 func NewResolutionMetadata() *ResolutionMetadata {
 	return &ResolutionMetadata{
 		DispatchCounter:      new(atomic.Uint32),
-		WasThrottled:         new(atomic.Bool),
+		DispatchThrottled:    new(atomic.Bool),
 		WasWeightedGraphUsed: new(atomic.Bool),
 		CheckCounter:         new(atomic.Uint32),
 	}
@@ -738,7 +738,7 @@ func (c *ReverseExpandQuery) throttle(ctx context.Context, currentNumDispatch ui
 		attribute.Bool("is_throttled", shouldThrottle))
 
 	if shouldThrottle {
-		metadata.WasThrottled.Store(true)
+		metadata.DispatchThrottled.Store(true)
 		c.dispatchThrottlerConfig.Throttler.Throttle(ctx)
 	}
 }
--- a/pkg/server/commands/reverseexpand/reverse_expand_test.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand_test.go
@@ -501,7 +501,7 @@ func TestReverseExpandThrottle(t *testing.T) {
 		metadata.DispatchCounter.Store(dispatchCountValue)
 
 		reverseExpandQuery.throttle(ctx, dispatchCountValue, metadata)
-		require.False(t, metadata.WasThrottled.Load())
+		require.False(t, metadata.DispatchThrottled.Load())
 	})
 
 	t.Run("above_threshold_should_call_throttle", func(t *testing.T) {
@@ -521,7 +521,7 @@ func TestReverseExpandThrottle(t *testing.T) {
 		metadata.DispatchCounter.Store(dispatchCountValue)
 
 		reverseExpandQuery.throttle(ctx, dispatchCountValue, metadata)
-		require.True(t, metadata.WasThrottled.Load())
+		require.True(t, metadata.DispatchThrottled.Load())
 	})
 
 	t.Run("zero_max_should_interpret_as_default", func(t *testing.T) {
@@ -541,7 +541,7 @@ func TestReverseExpandThrottle(t *testing.T) {
 		metadata.DispatchCounter.Store(dispatchCountValue)
 
 		reverseExpandQuery.throttle(ctx, dispatchCountValue, metadata)
-		require.False(t, metadata.WasThrottled.Load())
+		require.False(t, metadata.DispatchThrottled.Load())
 	})
 
 	t.Run("dispatch_should_use_request_threshold_if_available", func(t *testing.T) {
@@ -563,7 +563,7 @@ func TestReverseExpandThrottle(t *testing.T) {
 		metadata.DispatchCounter.Store(dispatchCountValue)
 
 		reverseExpandQuery.throttle(ctx, dispatchCountValue, metadata)
-		require.True(t, metadata.WasThrottled.Load())
+		require.True(t, metadata.DispatchThrottled.Load())
 	})
 
 	t.Run("should_respect_max_threshold", func(t *testing.T) {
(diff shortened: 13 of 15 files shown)


internal/cachecontroller/cache_controller.go
```diff
@@ -136,26 +151,37 @@ func (c *InMemoryCacheController) DetermineInvalidationTime(
...
+		c.InvalidateIfNeeded(ctx, storeID) // async
+
+		// Return zero time to allow caller to use cache while invalidation is in progress.
+		return time.Time{}
 	}
 
-	c.InvalidateIfNeeded(ctx, storeID)
+	// Cache hit
+	cacheHitCounter.Inc()
+	span.SetAttributes(attribute.Bool("cached", true))
```
justincoh: This is a chance for us to consider what this metric actually means. At this point in the code, we only know that there exists a cached changelog entry, not whether it remains valid. The metric description itself is actually a bit ambiguous: `The total number of cache hits from cachecontroller requests.`.

I think the original implementation was _trying_ to implement a counter for `we found a changelog cache entry AND it is within TTL`. If we want to replicate that correctly, this `Inc()` method should be in an else a few lines below:

```
// Ensure invalidation is triggered at most every c.ttl duration per store.
if time.Since(entry.LastChecked) > c.ttl {
	c.InvalidateIfNeeded(ctx, storeID) // async
} else {
	// Cache hit
	cacheHitCounter.Inc()
}
```

internal/cachecontroller/cache_controller.go
```diff
@@ -109,11 +113,19 @@ type InMemoryCacheController struct {
...
+	cache storage.InMemoryCache[any],
+	ttl time.Duration,
+	queryCacheTTL time.Duration,
+	iteratorCacheTTL time.Duration,
+	opts ...InMemoryCacheControllerOpt,
+) CacheController {
 	c := &InMemoryCacheController{
 		ds:                    ds,
 		cache:                 cache,
 		ttl:                   ttl,
```
justincoh: wdyt about renaming `ttl` here to `invalidationInterval` or something like that, which reflects its actual responsibility? Now that we have `ttl`, `queryCacheTTL`, and `iteratorCacheTTL`, it'd be good to clarify.

internal/cachecontroller/cache_controller.go
```diff
@@ -109,11 +113,19 @@ type InMemoryCacheController struct {
...
+		queryCacheTTL:         queryCacheTTL,
 		iteratorCacheTTL:      iteratorCacheTTL,
```
justincoh: Now that we have multiple TTLs in here, do you think we should add a validation here that says `TTLs must be >= invalidationInterval`?

internal/cachecontroller/cache_controller.go
```diff
@@ -251,10 +280,13 @@ func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(parentCtx
...
 		LastModified: lastChangelog.GetTimestamp().AsTime(),
+		LastChecked:  time.Now(),
 	}
 
-	defer c.cache.Set(cacheKey, entry, utils.JitterDuration(c.ttl, time.Minute)) // add buffer between checks
-
+	// The changelog cache entry is only used to compare against a cached Check
+	// response. Therefore, we only need this entry for up to the TTL of the
+	// cached Check response (queryCacheTTL).
+	c.cache.Set(cacheKey, entry, c.queryCacheTTL)
```
saad-h1, line 295: This change likely needs the most scrutiny in review. 3 things changed:

1. No longer using defer
    - I didn't see a reason to wait, I thought it's better to get this available asap, not sure if I'm missing something.

2. Setting TTL to match the Check query cache TTL (longer) instead of the cache controller TTL (shorter)
    - See the comment in the code. But just thinking about the future, is there any issue with tying this to the Check cache response TTL? I suppose we can change things later if needed.

3. No longer adding jitter to TTL
    - I didn't see a reason to add jitter, not sure if I'm missing something. There is already a bit of "randomness" as to when a Check/LO triggers this for a given store.

ListObjects doesn't use `DetermineInvalidationTime()`, but it does use this function, `InvalidateIfNeeded()`. However, the only change we're making in this function is to `ChangelogCacheEntry`, and as far as I could tell, that is mainly only used in `DetermineInvalidationTime()` (which LO doesn't use). There is some small usage of it earlier in this function as well, but I don't think these changes should affect it. So I don't think LO should be affected (might be good for someone to double check me on that since I haven't looked into LO much before).

internal/utils/jitter.go
saad-h1, line 1: see https://github.com/openfga/openfga/pull/2811/files#r2539780802

internal/utils/jitter_test.go
saad-h1, line 1: see https://github.com/openfga/openfga/pull/2811/files#r2539780802

cmd/run/run.go
```diff
@@ -263,7 +263,7 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Bool("cache-controller-enabled", defaultConfig.CacheController.Enabled, "enabling dynamic invalidation of check query cache and check iterator cache based on whether there are recent tuple writes. If enabled, cache will be invalidated when either 1) there are tuples written to the store OR 2) the check query cache or check iterator cache TTL has expired.")
 
-	flags.Duration("cache-controller-ttl", defaultConfig.CacheController.TTL, "if cache controller is enabled, control how frequent read changes are invoked internally to query for recent tuple writes to the store.")
+	flags.Duration("cache-controller-ttl", defaultConfig.CacheController.TTL, "if cache controller is enabled, this is the minimum time interval between cache invalidation attempts. The cache controller will check the database for new writes and conditionally invalidate cache entries at most once per TTL duration. Note that invalidations are only triggered on Check requests, so the actual interval may be longer if there are no requests.")
```
aaguiarz, line 267: Is this used for ListObjects or only check?

saad-h1: The cache controller is used by ListObjects, but this TTL is not. This TTL is only used during Check.

Some technical detail: the cache controller has two methods:
1. `DetermineInvalidationTime()` - called only by Check. This has logic to call `InvalidateIfNeeded()` every cache controller TTL.
2. `InvalidateIfNeeded()` - called by `DetermineInvalidationTime()`, but also called directly by ListObjects. Cache controller TTL not used here.

internal/cachecontroller/cache_controller.go
saad-h1: That makes sense to me. Maybe `minInvalidationInterval` since we could take longer to invalidate if no Check comes in?

Only thing I wonder is if it would be confusing that the `cache-controller-ttl` configuration option is what this field is after renaming. Hopefully a comment to clarify that is sufficient.

adriantam:
@claude

internal/cachecontroller/cache_controller.go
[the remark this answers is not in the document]
saad-h1: same as https://github.com/openfga/openfga/pull/2811#discussion_r2539584284

internal/cachecontroller/cache_controller.go
```diff
@@ -109,11 +113,19 @@ type InMemoryCacheController struct {
 	wg sync.WaitGroup
 }
 
-func NewCacheController(ds storage.OpenFGADatastore, cache storage.InMemoryCache[any], ttl time.Duration, iteratorCacheTTL time.Duration, opts ...InMemoryCacheControllerOpt) CacheController {
+func NewCacheController(
+	ds storage.OpenFGADatastore,
+	cache storage.InMemoryCache[any],
+	ttl time.Duration,
+	queryCacheTTL time.Duration,
```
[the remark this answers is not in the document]
saad-h1: same as https://github.com/openfga/openfga/pull/2811#discussion_r2539594694

cmd/run/run.go
adriantam: Probably not for this PR.  But I think we should investigate whether we need to set LastCacheInvalidationTime for list object's call checkResolver.  CC: @senojj, @Vic-Dev

saad-h1: Ah I guess this statement is not accurate: `invalidations are only triggered on Check requests`. They're also triggered on LO requests. Updated that.

internal/cachecontroller/cache_controller.go
saad-h1: that makes sense, although that would be a breaking change right?

FWIW, `ttl` is very decoupled from `iteratorCacheTTL` so it doesn't _need_ to be less per se, although invalidation isn't really useful if the iterator cache entries expire earlier. similar case with `queryCacheTTL`, even if `queryCacheTTL < ttl`, it just means the invalidation isn't very useful since the cached response expires before invalidation runs anyway, but it doesn't affect the expectation of what should happen I think?

justincoh: Yeah something like this would be better served as a validation check [over here](https://github.com/openfga/openfga/blob/main/pkg/server/config/config_test.go#L517). Nothing _bad_ will happen if the `CacheController.ttl` is longer than the actual caches, it's just wasted effort.

internal/cachecontroller/cache_controller.go
```diff
@@ -251,10 +280,13 @@ func (c *InMemoryCacheController) findChangesAndInvalidateIfNecessary(parentCtx
 	lastChangelog := changes[0]
 	entry := &storage.ChangelogCacheEntry{
 		LastModified: lastChangelog.GetTimestamp().AsTime(),
+		LastChecked:  time.Now(),
```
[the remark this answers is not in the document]
saad-h1: I think this is acceptable for now

internal/cachecontroller/cache_controller.go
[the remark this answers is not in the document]
saad-h1: RE point 1: `ChangelogCacheEntry` is very small (~100 bytes), is stored per-store, and store cardinality is low. Even with 10,000 stores, this would only be ~1MB of memory. We should just encourage the TTLs to be set correctly anyway.

For point 2, this is already mentioned in https://github.com/openfga/openfga/pull/2811#discussion_r2539780802

For point 3 and the recommendation, this is already mentioned in https://github.com/openfga/openfga/pull/2811#discussion_r2539594694

internal/cachecontroller/cache_controller.go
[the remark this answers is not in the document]
saad-h1: same as https://github.com/openfga/openfga/pull/2811#discussion_r2539574221

internal/cachecontroller/cache_controller.go
```diff
@@ -136,26 +151,37 @@ func (c *InMemoryCacheController) DetermineInvalidationTime(
...
-		}
+
+	// entry is nil when cacheResp is nil (cache miss) or type assertion fails
+	// (shouldn't happen since we used the unique changelog cache key prefix
+	// when getting from cache).
+	entry, _ := cacheResp.(*storage.ChangelogCacheEntry)
+	if entry == nil {
+		c.InvalidateIfNeeded(ctx, storeID) // async
+
+		// Return zero time to allow caller to use cache while invalidation is in progress.
```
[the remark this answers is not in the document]
saad-h1: okay, updated

internal/cachecontroller/cache_controller.go
```diff
@@ -126,8 +138,11 @@ func NewCacheController(ds storage.OpenFGADatastore, cache storage.InMemoryCache
 	return c
 }
 
-// DetermineInvalidationTime returns the timestamp of the last write for the specified store if it was in cache,
-// Else it returns Zero time and triggers InvalidateIfNeeded().
+// DetermineInvalidationTime returns the timestamp of the last write for the
+// specified store if it was in cache, else it returns the Zero time and
+// triggers InvalidateIfNeeded(). The last write time can be used to determine
+// whether a cached entry is stil valid - if it was cached before the last write
```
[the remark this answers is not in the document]
saad-h1: updated

internal/cachecontroller/cache_controller.go
```diff
@@ -57,8 +56,11 @@ var (
 )
 
 type CacheController interface {
-	// DetermineInvalidationTime returns the timestamp of the last write for the specified store if it was in cache,
-	// Else it returns Zero time and triggers InvalidateIfNeeded().
+	// DetermineInvalidationTime returns the timestamp of the last write for the
+	// specified store if it was in cache, else it returns the Zero time and
+	// triggers InvalidateIfNeeded(). The last write time can be used to determine
+	// whether a cached entry is stil valid - if it was cached before the last write
```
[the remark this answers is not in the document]
saad-h1: updated

internal/cachecontroller/cache_controller.go
saad-h1: Good point, especially now that the `ChangelogCacheEntry` has the same TTL as the Check query cache TTL which is usually longer, and it's refreshed on a Check/LO, I expect cache misses will be less common so this metric is probably not very useful anymore.

I wonder how useful "we found a changelog cache entry and it is within TTL" is as well though? Whether we were within the TTL or not, we don't know if the time we're returning is stale or not, it all depends when a write happens. Maybe a write came in 1s after we last checked, but we're within the TTL so we're not triggering invalidation yet.

Claude Code recommended the metrics below instead, thoughts on adding any of them? Or we could just keep it simple and go with the way you suggested for now.

---

## More Useful Metrics

Here are metrics that would provide more actionable insights:

### 1. **Invalidation Trigger Counter** (Most Useful)

```go
// On line 181-182, when we actually trigger invalidation:
if time.Since(entry.LastChecked) > c.minInvalidationInterval {
    invalidationTriggeredCounter.Inc() // NEW METRIC
    c.InvalidateIfNeeded(ctx, storeID)
}
```

**Why useful:** Shows how often you're querying the changelog database. If this is too high relative to request rate, your TTL might be too short.

### 2. **Zero Time Return Counter**

```go
// On line 172, when we return zero time:
zeroTimeReturnCounter.Inc() // NEW METRIC
return time.Time{}
```

**Why useful:** Tracks the "staleness window" - requests that might get stale cache hits. This directly measures the problem you're trying to minimize.

### 3. **LastChecked Staleness Histogram**

```go
// On line 177, after cache hit:
lastCheckedAgeSeconds := time.Since(entry.LastChecked).Seconds()
lastCheckedAgeHistogram.Observe(lastCheckedAgeSeconds)
```

**Why useful:** Shows the distribution of staleness. Helps you tune `minInvalidationInterval` - if most values are near the max, your TTL might be too high.

### 4. **Invalidation Skip Counter**

```go
// In InvalidateIfNeeded, line 188 when skipping:
if present {
    invalidationSkippedCounter.Inc() // NEW METRIC
    span.SetAttributes(attribute.Bool("cache_controller_invalidation", false))
    return
}
```

**Why useful:** Shows request concurrency - how often multiple requests try to trigger invalidation simultaneously. High values indicate hot stores.

justincoh: > I wonder how useful "we found a changelog cache entry and it is within TTL"

Same. The existing metric description is also a bit confusing: `The total number of cache hits from cachecontroller requests.` Just because it is within TTL here does not mean we subsequently found a cached response for Check.

The `cacheTotalCounter` and `cacheHitCounter` here appear to be measuring "how often does the cache_controller have to invalidate", so for now let's just keep it simple. The important part of this PR is the timestamp fix anyway, metrics improvements would be an unrelated bonus (which we could still look into, but in a separate PR)

internal/cachecontroller/cache_controller.go
saad-h1: On the off chance that someone does have their query cache TTL or iterator cache TTL greater than their cache controller TTL and we start returning a config validation error, that could be a breaking change for them right? Not sure of the typical practice for configs, is that acceptable?

internal/cachecontroller/cache_controller.go
saad-h1: updated per your suggestion

internal/cachecontroller/cache_controller.go
justincoh: hmm yes it would. I wonder what the best behavior here would be 🤔 . At the very least we should log a warning, and at the strictest we could just automatically disable the cache controller entirely in this scenario, rather than failing to build. Option 2 would be surprising behavior (even if correct). We could just log a warning at the end of `NewCacheController`? But I don't want this small bit to hold up this PR so maybe forget about it for right now; we can take it in a tiny follow up PR

justincoh (approved):

cmd/run/run.go
```diff
@@ -261,9 +261,9 @@ func NewRunCommand() *cobra.Command {
 
 	flags.Duration("check-query-cache-ttl", defaultConfig.CheckQueryCache.TTL, "if check-query-cache-enabled, this is the TTL of each value")
 
-	flags.Bool("cache-controller-enabled", defaultConfig.CacheController.Enabled, "enabling dynamic invalidation of check query cache and check iterator cache based on whether there are recent tuple writes. If enabled, cache will be invalidated when either 1) there are tuples written to the store OR 2) the check query cache or check iterator cache TTL has expired.")
+	flags.Bool("cache-controller-enabled", defaultConfig.CacheController.Enabled, "enable invalidation of check query cache and iterator cache based on recent tuple writes. Invalidation is triggered by Check and List Objects requests, which periodically check the datastore's changelog table for writes and invalidate cache entries earlier than recent writes. Invalidations from Check requests are rate-limited by cache-controller-ttl, whereas List Objects requests invalidate every time if list objects iterator cache is enabled.")
```
adriantam, line 264: Please also update wording in [.config-schema.json](https://github.com/openfga/openfga/blob/main/.config-schema.json)

internal/cachecontroller/cache_controller.go
adriantam: @saad-h1 - just want to double verify the intention of number 2 (setting the cache entry to the cache TTL instead of the controller TTL).  Does this mean that if the query Cache TTL is  1H but cache controller TTL entry is 1 minute, the cache controller's changelog entry will be valid for 1H.  is this the real intend?  What is the purpose of the cache controller TTL then?

saad-h1: @adriantam yes that's right. The cache controller TTL is used in this PR as `c.minInvalidationInterval` in `DetermineInvalidationTime()` to trigger invalidation at most once every cache controller TTL duration (assuming `DetermineInvalidationTime()` is regularly called, which is triggered by Check requests): https://github.com/openfga/openfga/blob/20d6f440b16e6a9045939d4b7409260344310094/internal/cachecontroller/cache_controller.go#L180

The `ChangelogCacheEntry` is mainly used in `DetermineInvalidationTime()` to return the time of the last known write to the store, which a Check compares against its cached Check response to see if it's later than the last write or not. If the `ChangelogCacheEntry` expires, we are returning the zero time to allow the cache to be used to favour minimizing latency, so I wanted the `ChangelogCacheEntry` to be available as long as possible in case a Check hasn't come in to this replica for a while. However, it is still being updated more regularly than its TTL - it's updated every cache controller TTL assuming regular Checks. So given your example numbers, assuming regular enough Checks, the `ChangelogCacheEntry` should be updated every 1m.

I did try setting this TTL to the cache controller TTL at some point during my testing, but that makes it so that a Check would have to regularly come in within the short cache controller TTL to get a meaningful time from `DetermineInvalidationTime()`. If a Check doesn't come within that short window, the `ChangelogCacheEntry` expires and we return the zero time, which allows any cached Check response, which moves us closer to the original problem.

PR #2816, opened 2025-11-19
refactor: check using weighted graph

elbuo8:
## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * V2 authorization check engine with pluggable strategies (including recursive and weight‑2) and model-graph resolver.
  * Support for contextual tuples in checks and list operations.
  * Experimental weighted-graph check and optional background shadow validation.

* **Tests**
  * Large expansion of unit/integration tests covering new strategies, recursive flows, iterators, and contextual tuple behavior.
  * Test framework updated to exercise contextual modes.

* **Bug Fixes / Changes**
  * Simplified invariant cache key composition (may affect existing cache keys).

### DIFF
--- a/internal/check/response.go
+++ b/internal/check/response.go
@@ -9,6 +9,9 @@ type Response struct {
 }
 
 func (r *Response) GetAllowed() bool {
+	if r == nil {
+		return false
+	}
 	return r.Allowed
 }
 


elbuo8:
@claude

internal/check/weight2.go
```diff
@@ -0,0 +1,159 @@
...
+				return &Response{Allowed: true}, nil
+			}
+
+			// Otherwise, store for future comparison
+			rightSeen[rightMsg.Value] = struct{}{}
+		}
+	}
+
+	// If we get here, no match was found
+	return &Response{Allowed: false}, lastErr
```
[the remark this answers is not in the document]
yissellokta: No, it should not fail fast. If at least there is a path to a solution, even when other paths can have errors, then it should return a solution. Only when there is not solution (false) and there were errors, we will return error

internal/check/check.go
```diff
@@ -0,0 +1,828 @@
...
+		})
+	}
+
+	var err error
+	for i := 0; i < scheduledHandlers; i++ {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case msg := <-out:
+			if msg.Err != nil || !msg.Res.GetAllowed() {
```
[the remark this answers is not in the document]
yissellokta: when msg.Err != nil then msg.Res == nil, if msg.Err == nil then msg.Res != nil

internal/check/bottom_up.go
```diff
@@ -0,0 +1,544 @@
...
+	}
+
+	compareValues := make([]string, 0, len(iters))     // these values are the head values of each iterator
+	reverseLookupIndexes := make([]int, 0, len(iters)) // these values are the iters index that are currently active, for example 0, 1, 2, 3, 4
+	for idx, iter := range iters {
+		value, err := iter.Next(ctx)
+		if err != nil {
+			if handleStreamError(ctx, err, &batch, out) {
+				return
+			}
```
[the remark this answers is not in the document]
yissellokta: it is ok to send partial results

internal/check/recursive.go
```diff
@@ -0,0 +1,325 @@
...
+	var err error
+	for leftChan != nil || rightChan != nil {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case msg, ok := <-leftChan:
+			if !ok {
+				leftChan = nil
+				// if no ids from the left side were returned then return false without error
+				if len(idsFromUser) == 0 {
```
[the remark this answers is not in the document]
yissellokta: it will return error if there was a previous error already stored

elbuo8:
@coderabbitai review

internal/iterator/filter.go
```diff
@@ -0,0 +1,133 @@
...
+				// Head() will return 1. If it is invalid, Next() will return 1 and move the pointer to end of list.
+				// Thus, Head() will return ErrIteratorDone next time being called.
+				return null, err
+			}
+			continue
+		}
+		f.onceValid = true
+		return tuple, nil
+	}
+}
```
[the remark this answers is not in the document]
elbuo8: We should delete Head completely.

internal/check/recursive.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+	iter := storage.NewTupleKeyIteratorFromTupleIterator(tIter)
+	if ctxIter != nil {
+		iter = iterator.Concat(ctxIter, iter)
+	}
+	iterFilters := make([]iterator.FilterFunc[*openfgav1.TupleKey], 0, 2)
+	iterFilters = append(iterFilters, BuildUniqueTupleKeyFilter(visited, uniqueKeyFunc))
+	if len(edge.GetConditions()) > 1 || edge.GetConditions()[0] != authzGraph.NoCond {
+		iterFilters = append(iterFilters, BuildConditionTupleKeyFilter(ctx, s.model, edge.GetConditions(), req.GetContext()))
+	}
```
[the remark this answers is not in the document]
elbuo8: It's already covered by inspecting `len` > 1.

internal/check/check.go
```diff
@@ -0,0 +1,822 @@
...
+	// excluded edge
+	if _, ok := r.model.GetEdgeWeight(edges[1], req.GetUserType()); ok {
+		scheduledHandlers++
+		subtract = make(chan ResponseMsg, 1)
+		wg.Go(func() {
+			// exclusion is never part of a cycle or recursion
+			res, err := r.ResolveEdge(ctx, req, edges[1], nil)
+			concurrency.TrySendThroughChannel(ctx, ResponseMsg{Res: res, Err: err}, subtract)
+			close(subtract)
+		})
```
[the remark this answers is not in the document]
elbuo8: Sadly, we have to do this.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,539 @@
...
+	go func() {
+		recoveredError := panics.Try(func() {
+			resolver(ctx, iterChans, out)
+		})
+
+		if recoveredError != nil {
+			concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Err: fmt.Errorf("%w: %s", ErrPanicRequest, recoveredError.AsError())}, out)
+		}
+	}()
+	return out, nil
```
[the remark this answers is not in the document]
elbuo8: This would require completely changing the control flow of weight2. Wouldn't do this.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,539 @@
...
+	}
+
+	iter := storage.NewTupleKeyIteratorFromTupleIterator(tIter)
+	if ctxTuples, ok := req.GetContextualTuplesByUserID(tuple.TypedPublicWildcard(req.GetUserType()), relation, objectType); ok {
+		iter = iterator.Merge(iter, storage.NewStaticTupleKeyIterator(ctxTuples), func(a, b *openfgav1.TupleKey) int {
+			if a.GetObject() < b.GetObject() {
+				return -1
+			}
+			return 1
+		})
```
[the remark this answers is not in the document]
yissellokta: no, this is to merge sort, and even when equal we want the same order.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,539 @@
...
+
+	iter := storage.NewTupleKeyIteratorFromTupleIterator(tIter)
+	if ctxTuples, ok := req.GetContextualTuplesByUserID(req.GetTupleKey().GetUser(), relation, objectType); ok {
+		iter = iterator.Merge(iter, storage.NewStaticTupleKeyIterator(ctxTuples), func(a, b *openfgav1.TupleKey) int {
+			if a.GetObject() < b.GetObject() {
+				return -1
+			}
+			return 1
+		})
+	}
```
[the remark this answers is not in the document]
yissellokta: No, this is to merge sort, and even when values could be equal, I want always to choose a specific order

internal/check/bottom_up.go
```diff
@@ -0,0 +1,544 @@
...
+		value, err := iters[BaseIndex].Next(ctx)
+		if err != nil {
+			handleStreamError(ctx, err, &batch, out)
+			return
+		}
+		compareValues[BaseIndex] = value
+	}
+}
+
+func skipTo(ctx context.Context, iter storage.Iterator[string], target string) (string, error) {
```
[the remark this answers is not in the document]
yissellokta: no needed, it is mandatory to iterate the whole iterator, skipTo is only a convenience function to go faster

internal/check/weight2.go
```diff
@@ -0,0 +1,159 @@
...
+// From the perspective of the model, the left hand side of a TTU is the computed relationship being expanded.
+func (s *Weight2) execute(ctx context.Context, leftChan chan *iterator.Msg, rightIter storage.TupleMapper) (*Response, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	defer rightIter.Stop()
+	defer iterator.Drain(leftChan)
+
+	// Set to store already seen values from each side
+	// We use maps for O(1) lookup complexity, consistent with hashset implementation
+	leftSeen := make(map[string]struct{})
```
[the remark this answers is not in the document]
yissellokta: It was implemented a different mechanism to avoid adding to the map unnecessarily

internal/check/bottom_up.go
```diff
@@ -0,0 +1,544 @@
...
+	childrenTotal := len(iters)
+	if childrenTotal == 0 || ctx.Err() != nil {
+		return
+	}
+
+	compareValues := make([]string, 0, len(iters)) // these values are the head values of each iterator
+	for _, iter := range iters {
+		value, err := iter.Next(ctx)
+		if err != nil {
+			handleStreamError(ctx, err, &batch, out)
```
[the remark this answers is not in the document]
yissellokta: Intersection fails fast on the first error, which is the current implementation. After each handleStreamError, there is a return

elbuo8:
@claude

elbuo8:
@coderabbitai review

internal/check/check.go
```diff
@@ -0,0 +1,825 @@
...
+				continue
+			}
+
+			if msg.Res.GetAllowed() {
+				return msg.Res, nil
+			}
+		}
+	}
+	return &Response{Allowed: false}, err
+}
```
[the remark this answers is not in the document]
elbuo8: It won't block forever given the channel is created with the size of 2, so both goroutines will be able to write to the channel regardless of the context being cancelled or not.

internal/check/check.go
```diff
@@ -0,0 +1,825 @@
...
+	if len(conditions) > 1 || conditions[0] != authzGraph.NoCond {
+		iterFilters = append(iterFilters, BuildConditionTupleKeyFilter(ctx, r.model, conditions, req.GetContext()))
+	}
+
+	if len(iterFilters) > 0 {
+		return iterator.NewFilteredIterator(tupleKeyIter, iterFilters...)
+	}
+
+	return tupleKeyIter
+}
```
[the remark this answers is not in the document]
elbuo8: `conditions` will never be empty.

internal/check/filters.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/openfga/openfga/internal/modelgraph"
+)
+
+func evaluateCondition(ctx context.Context, model *modelgraph.AuthorizationModelGraph, conditions []string, t *openfgav1.TupleKey, reqCtx *structpb.Struct) (bool, error) {
+	if !slices.Contains(conditions, t.GetCondition().GetName()) {
+		return false, nil
+	}
+
+	return eval.EvaluateTupleCondition(ctx, t, model.GetConditions()[t.GetCondition().GetName()], reqCtx)
+}
```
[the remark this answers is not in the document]
elbuo8: Conditions don't have much usage, we can consider this later.

internal/check/recursive.go
```diff
@@ -0,0 +1,326 @@
...
+
+	go func() {
+		_ = pool.Wait()
+		close(responsesChan)
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
```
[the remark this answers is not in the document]
elbuo8: Pool supports cancellation.

internal/check/recursive.go
```diff
@@ -0,0 +1,326 @@
...
+				Allowed: true,
+			}}, out)
+			return // cancel will be propagated to the remaining goroutines
+		}
+		pool.Go(func() error {
+			s.recursiveMatchResolver(ctx, req, edge, recursiveType, pool, idsFromUser, visitedIds, t, out)
+			return nil
+		})
+	}
+}
```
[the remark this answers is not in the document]
elbuo8: We will rely on throttling and timeouts.

internal/modelgraph/resolver.go
```diff
@@ -0,0 +1,102 @@
...
+			if errors.Is(err, storage.ErrNotFound) {
+				return nil, ErrModelNotFound
+			}
+
+			return nil, fmt.Errorf("failed to FindLatestAuthorizationModel: %w", err)
+		}
+		model = m
+		key = CacheKeyPrefix + storeID + modelID
+		if wg := r.cache.Get(key); wg != nil {
+			return wg.(*AuthorizationModelGraph), nil
```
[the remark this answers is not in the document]
elbuo8: Good catch.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,542 @@
...
+	if ctx.Err() != nil {
+		return
+	}
+
+	compareValues := make([]string, 0, len(iters))     // these values are the head values of each iterator
+	reverseLookupIndexes := make([]int, 0, len(iters)) // these values are the iters index that are currently active, for example 0, 1, 2, 3, 4
+	for idx, iter := range iters {
+		value, err := iter.Next(ctx)
+		if err != nil {
+			if handleStreamError(ctx, err, &batch, out) {
```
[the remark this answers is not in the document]
elbuo8: Fixed.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,542 @@
...
+		return nil, err
+	}
+
+	iter := storage.NewTupleKeyIteratorFromTupleIterator(tIter)
+	if ctxTuples, ok := req.GetContextualTuplesByUserID(req.GetTupleKey().GetUser(), relation, objectType); ok {
+		iter = iterator.Merge(iter, storage.NewStaticTupleKeyIterator(ctxTuples), func(a, b *openfgav1.TupleKey) int {
+			if a.GetObject() < b.GetObject() {
+				return -1
+			}
+			return 1
```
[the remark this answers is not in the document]
yissellokta: yes, it is intentional to always be bias to toward the second iterator (the contextual tuples win)

.golangci.yaml
```diff
@@ -28,6 +28,9 @@ linters:
     - wastedassign
     - whitespace
   settings:
+    gocritic:
+      disabled-checks:
+        - ifElseChain
```
adriantam, line 34: why do we disable this?

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -5658,96 +5659,6 @@ tests:
               relation: viewer
             expectation:
               - user:anne
-  - name: simple_userset_child_computed_userset
```
adriantam, line 5661: Can you call out why we remove this?  I guess this is redundant - but want to confirm.

adriantam:
Given the large scale changes involved, we definitely should mention this in CHANGELOG.md

pkg/server/check.go
```diff
@@ -178,6 +185,70 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+		zap.Bool("main", mainRes.GetAllowed()),
+		zap.Bool("shadow", res.GetAllowed()),
+	)
+}
+
+func (s *Server) v2Check(ctx context.Context, req *openfgav1.CheckRequest) (*openfgav1.CheckResponse, error) {
+	cacheInvalidationTime := time.Time{}
+
+	if req.GetConsistency() != openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
+		cacheInvalidationTime = s.sharedDatastoreResources.CacheController.DetermineInvalidationTime(ctx, req.GetStoreId())
```
adriantam, line 237: Question: In the case this comes from shadow check, will we invalidate the cache controller twice?

pkg/server/commands/check.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func WithCheckQueryV2UpstreamTimeout(timeout time.Duration) CheckQueryV2Option {
+	return func(cmd *CheckQueryV2) {
+		cmd.upstreamTimeout = timeout
+	}
+}
+
+func NewCheckQuery(opts ...CheckQueryV2Option) *CheckQueryV2 {
```
adriantam, line 85: Should we rename this to NewCheckQueryV2 to be more explicit that this is V2

pkg/server/commands/errors.go
```diff
@@ -49,17 +52,30 @@ func (e *ThrottledError) Error() string {
...
 
-	var invalidTupleError *InvalidTupleError
-	if errors.As(err, &invalidTupleError) {
+	var invalidTupleDeprecate *InvalidTupleError
+	if errors.As(err, &invalidTupleDeprecate) {
 		tupleError := tuple.InvalidTupleError{Cause: err}
 		return serverErrors.HandleTupleValidateError(&tupleError)
 	}
 
+	if errors.Is(err, errInvalidTuple) {
```
adriantam, line 75: How is this different than line 70-75?

pkg/server/server.go
```diff
@@ -889,6 +892,9 @@ func NewServerWithOpts(opts ...OpenFGAServiceV1Option) (*Server, error) {
 		return nil, err
 	}
 
+	// TODO: make the cache duration configurable (maybe)
+	s.authzModelGraphResolver = modelgraph.NewResolver(s.datastore, s.sharedDatastoreResources.CheckCache, 24*7*time.Hour)
```
adriantam, line 997: I remember someone raised an issue (and likely PR) that they want the cache size used for typesystem configurable.  Perhaps we also want to do this here

pkg/storage/tuple_mappers.go
```diff
@@ -78,6 +78,10 @@ func (n UsersetMapper) doMap(t *openfgav1.TupleKey) (string, error) {
 	return usersetName, nil
 }
 
+func (n UsersetMapper) doMap(t *openfgav1.TupleKey) (string, error) {
```
adriantam: Can you call out why we need a private method that calls the public function?

pkg/storage/tuple_mappers.go
```diff
@@ -107,10 +111,14 @@ func (n TTUMapper) Head(ctx context.Context) (string, error) {
 	return n.doMap(tupleRes)
 }
 
-func (n TTUMapper) doMap(t *openfgav1.TupleKey) (string, error) {
+func MapTTU(t *openfgav1.TupleKey) (string, error) {
 	return t.GetUser(), nil
 }
 
+func (n TTUMapper) doMap(t *openfgav1.TupleKey) (string, error) {
```
adriantam: Similar to https://github.com/openfga/openfga/pull/2816/files#r2599788459, can you call out why do we need to have the private method calling the public function?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+	"github.com/openfga/openfga/internal/iterator"
+	"github.com/openfga/openfga/internal/modelgraph"
+	"github.com/openfga/openfga/internal/planner"
+	"github.com/openfga/openfga/pkg/logger"
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/tuple"
+)
+
+const DELIMITER = "|"
+const CacheKeyPrefix = "c."
```
adriantam: Why are these public?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+	Cache                     storage.InMemoryCache[any]
+	CacheTTL                  time.Duration
+	LastCacheInvalidationTime time.Time
+	Planner                   planner.Manager
+	ConcurrencyLimit          int
+	UpstreamTimeout           time.Duration
+	Logger                    logger.Logger
+	Strategies                map[string]Strategy
+}
+type Resolver struct {
```
adriantam, line 52: Why is this public?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+		cacheTTL:                  cfg.CacheTTL,
+		lastCacheInvalidationTime: cfg.LastCacheInvalidationTime,
+		planner:                   cfg.Planner,
+		concurrencyLimit:          cfg.ConcurrencyLimit,
+		upstreamTimeout:           cfg.UpstreamTimeout,
+		logger:                    cfg.Logger,
+		strategies:                cfg.Strategies,
+	}
+
+	if r.strategies == nil {
```
adriantam, line 84: pattern like this suggests this is a good candidate for `WithStrategies()` pattern

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+			DefaultStrategyName:   NewDefault(cfg.Model, r, cfg.ConcurrencyLimit),
+			WeightTwoStrategyName: NewWeight2(cfg.Model, cfg.Datastore),
+			RecursiveStrategyName: NewRecursive(cfg.Model, cfg.Datastore, cfg.ConcurrencyLimit),
+		}
+	}
+
+	return r
+}
+
+func (r *Resolver) SetStrategies(strategies map[string]Strategy) {
```
adriantam: is this used anywhere?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+func (r *Resolver) ResolveUnionEdges(ctx context.Context, req *Request, edges []*authzGraph.WeightedAuthorizationModelEdge, visited *sync.Map) (*Response, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	if res, ok := r.isCached(req.GetConsistency(), req.GetCacheKey()); ok {
+		return res, nil
+	}
+
+	out := make(chan ResponseMsg, len(edges))
+	var pool errgroup.Group
```
adriantam, line 174: probably good to call out here that we don't anticipate errgroup to have any errors as the errors are returned via channel.

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+	pool.SetLimit(r.concurrencyLimit)
+	defer func() {
+		_ = pool.Wait()
+		close(out)
+	}()
+
+	ids := make([]string, 0, len(edges))
+	for _, edge := range edges {
+		id := buildEdgeCacheKey(r.model.GetModelID(), req, edge)
```
adriantam, line 186: Want to double confirm that in the case of public wildcard with non public wildcard, there will still be one edge.  Same as condition / without condition...  Otherwise, will it yield the same cache key?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+			}
+		}
+	}
+	if err != nil {
+		// we only return error in a union when all edges are exhausted and there is at least one edge with error
+		return nil, err
+	}
+	res := &Response{Allowed: false}
+	entry := &ResponseCacheEntry{Res: res, LastModified: time.Now()}
+	for _, id := range ids {
```
adriantam: I think what you are trying to say here is that all the response MUST be false if we get here (because otherwise it will be returned in line 196's block. If that is the case, it will be good to explicitly call out here as comments...

adriantam: One suggestion - will adding the cache in line 195 be better?  The case I am thinking of is the there are three edges... 
we have edge 1 return false,
edge 2 return true (and short circuited here)

Now, we will only cache edge 2.  But since we did do the work for edge 1, should we cache the result?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+		r.cache.Set(id, entry, r.cacheTTL)
+	}
+	r.cache.Set(req.GetCacheKey(), entry, r.cacheTTL)
+	return res, nil
+}
+
+// reduce as a logical union operation (exit the moment we have a single true).
+func (r *Resolver) ResolveUnion(ctx context.Context, req *Request, node *authzGraph.WeightedAuthorizationModelNode, visited *sync.Map) (*Response, error) {
+	emptyCycle := visited == nil
```
adriantam, line 239: nitpik: Should we insist that the caller call it visited sync map created?  I know that sync map does not provide a clean way to check for length...but if somehow the sync map is called with `.Clear()`, then this logic will be wrong ..  
I am good either way though

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+	ids := make([]string, 0, len(edges))
+	for _, edge := range edges {
+		id := buildEdgeCacheKey(r.model.GetModelID(), req, edge)
+		ids = append(ids, id)
+		if res, ok := r.isCached(req.GetConsistency(), id); ok {
+			concurrency.TrySendThroughChannel(ctx, ResponseMsg{ID: id, Res: res}, out)
+			continue
+		}
+		pool.Go(func() error {
+			res, err := r.ResolveEdge(ctx, req, edge, visited)
```
adriantam, line 196: Want to double confirm here - visited can be nil.  In this case, each Edge will have their own version of sync map.  
If visited != nil, each Edge will share the sync map.

internal/check/check.go
adriantam, line 239: Should we have trace in this function?
Actually, what is our tracing strategy in the new check?  When do we have a trace and when we do not?

internal/check/check.go
```diff
@@ -0,0 +1,847 @@
...
+)
+
+const DELIMITER = "|"
+const CacheKeyPrefix = "c."
+
+var tracer = otel.Tracer("internal/check")
+
+var ErrValidation = errors.New("object relation does not exist")
+var ErrUsersetInvalidRequest = errors.New("userset request cannot be resolved when exclusion operation is involved")
+var ErrPanicRequest = errors.New("invalid check request")
```
adriantam, line 37: Should we do any logging? I see that we return this in various place but how do we debug further if we see this.

For example, I see that 
```
terminalEdges, err := r.model.FlattenNode(node, req.GetUserType(), req.IsTypedWildcard(), false)
	if err != nil {
		return nil, ErrPanicRequest
	}
```

We lost all information of what is the error returned by the model's FlattenNode function.

.golangci.yaml
elbuo8: Because the linter doesn't like certain pieces of code which it deems its better to use a switch case, and it's def not better for readability.

assets/tests/consolidated_1_1_tests.yaml
elbuo8: We will not support usersets which are not explicitly defined. `member_c4` is a computed relationship, complete waste.

internal/check/check.go
elbuo8: 🤷🏽

internal/check/check.go
elbuo8: How do you suggest we go about it.

internal/check/check.go
elbuo8: 🤷🏽‍♂️  Not looking to create a massive interface for something that doesn't need it. Thus I'm exporting the direct resolver.

internal/check/check.go
elbuo8: It's not needed.

internal/check/check.go
elbuo8: Not adopting With here.

internal/check/check.go
elbuo8: https://github.com/openfga/language/blob/main/pkg/go/graph/weighted_graph_edge.go#L60

Only 1 edge, it will be grouped.

internal/check/check.go
elbuo8: How would you like it to read? The select block contains the logic about exiting on a single true. 

Those would be wasteful entries.

internal/check/check.go
elbuo8: No. Why even bother creating it if we don't need it at all. Then you have the problem that you stated, should I even be using it at all because I added entries.

elbuo8: I actually have tracing in my TODO.

pkg/server/check.go
elbuo8: No. Cache controller will realize that there is either an inflight request for the store or that a previous check was done and just return that value. This is no different that 2 concurrent check requests to the same store.

pkg/server/commands/check.go
elbuo8: Notice that the existing check is the only thing called `command`. Once we merge, everything will follow the same pattern, `Query` suffix. We can drop the V2 once there is no collision. By renaming to V2 we further add more cleanup work vs here is contained to a single file.

pkg/server/commands/errors.go
elbuo8: error.As vs error.Is. Once we are done with the previous one we can remove that which is not as performant.

pkg/server/server.go
elbuo8: In this already ridiculously massive PR? No.

And we really need to not depend on `sharedResources.CheckCache` as that makes absolutely no sense so we need to revisit the whole building caches story which is a complete mess.

pkg/storage/tuple_mappers.go
elbuo8: Changed. Tho this was already here.

elbuo8:
> Given the large scale changes involved, we definitely should mention this in CHANGELOG.md

Why if its all behind a FF? Thought we only added entries once its GA'ed.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,593 @@
...
+	}
+}
+
+func newBottomUpRecursive(model *modelgraph.AuthorizationModelGraph, ds storage.RelationshipTupleReader) *bottomUp {
+	return &bottomUp{
+		model:     model,
+		datastore: ds,
+		strategy:  recursive,
+	}
+}
```
Vic-Dev, line 49: Why do we need both of these?

internal/check/bottom_up.go
```diff
@@ -0,0 +1,593 @@
...
+
+// setOperationSetup returns a channel with a number of elements that is >= the number of children.
+// Each element is an iterator.
+// The caller must wait until the channel is closed.
```
Vic-Dev: Comment seems to have been copy pasted

internal/check/recursive.go
```diff
@@ -0,0 +1,326 @@
...
+				leftChan = nil
+				// if no ids from the left side were returned then return false without error
+				if len(idsFromUser) == 0 {
+					return &Response{Allowed: false}, err
+				}
+				break
+			}
+			if msg.Err != nil {
+				err = msg.Err
+				// if no ids from the left side were returned then return false with error, there not value to compare the right side
```
adriantam: Do you mean `there are no value to compare against the right side`?

internal/check/recursive.go
```diff
@@ -0,0 +1,326 @@
...
+				if len(idsFromUser) == 0 {
+					return &Response{Allowed: false}, err
+				}
+				continue
```
adriantam, line 114: curious continue vs. break here - or it doesn't matter

internal/check/recursive.go
```diff
@@ -0,0 +1,326 @@
...
+			ctxIter = storage.NewStaticTupleKeyIterator(ctxTuples)
+		}
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var kind storage.TupleMapperKind
+	var uniqueKeyFunc func(key *openfgav1.TupleKey) string
+	if recursiveType == RecursiveTypeTTU {
```
adriantam: I thought we already check for recursiveType in line 267.  Can we move this logic up into the same block?

internal/check/request.go
```diff
@@ -0,0 +1,405 @@
...
+		}
+
+		// Skip if duplicate found
+		if existingKey == newKey {
+			return slice
+		}
+	}
+
+	// Grow slice and insert at position
+	slice = append(slice, nil)
```
adriantam: More as a FYI - Instead of doing the insert ourself, we can use [slices.Insert](https://pkg.go.dev/slices#Insert)
Can totally go with this as well.

elbuo8: Nice. Updated.

internal/check/recursive.go
elbuo8: Updated.

internal/check/recursive.go
elbuo8: Done.

internal/check/recursive.go
elbuo8: Has to be continue.

internal/iterator/channel.go
```diff
@@ -0,0 +1,152 @@
...
+func ToChannel[T any](ctx context.Context, iter storage.Iterator[T], batchSize int) chan ValueMsg[T] {
+	out := make(chan ValueMsg[T], batchSize)
+
+	go func() {
+		defer close(out)
+		for {
+			t, err := iter.Next(ctx)
+			if err != nil {
+				if storage.IterIsDoneOrCancelled(err) {
+					return
```
senojj, line 37: Given that the only exit for the loop is here, is it possible for this goroutine to get stuck in an infinite busy loop when `iter.Next` continuously returns an error that is not "iterator done" or "canceled"?

If an iterator returns an error on a call to `Next`, is it even possible for subsequent calls to `Next` to return successfully?

senojj: I use the term "infinite" loosely here since the context will _eventually_ cancel due to a timeout.

internal/iterator/channel.go
```diff
@@ -0,0 +1,152 @@
...
+	"context"
+	"sync"
+
+	"github.com/openfga/openfga/internal/concurrency"
+	"github.com/openfga/openfga/pkg/storage"
+)
+
+func Drain(ch <-chan *Msg) *sync.WaitGroup {
+	if ch == nil {
+		return nil
```
senojj: Returning a `nil` value here creates a requirement for the caller to always check the return value for the presence of `nil` before calling `Wait` on the returned `sync.WaitGroup`.

It may improve the experience of the caller to just return a zero value `sync.WaitGroup` here instead, providing a guarantee that the return value will never be `nil`. However, this will result in an unnecessary allocation when `ch` is `nil`.

Another option that could avoid an unnecessary allocation would be to return a function like `func()` from `Drain`, and instead of returning `nil`, return an empty function here `func(){}`, and `wg.Done` in the successful condition.

internal/check/check.go
yissellokta: updated

internal/check/check.go
yissellokta: added comment

internal/check/check.go
yissellokta: weighted graph deduplicate edges and combine the conditions for the same from and to nodes

internal/check/check.go
yissellokta: @adriantam it will have value if the flattening process bypass multiple relations, having the edge as false will have value, but if the edge is connected directly to the request, having the edge store as false it is a waste. Good point

internal/iterator/channel.go
elbuo8: This is only used for testing (this was not introduced in this PR, just moved into this package). Modified it to just return the zero value for now.

internal/iterator/channel.go
elbuo8: > an infinite busy loop when iter.Next continuously returns an error that is not "iterator done" or "canceled"?

It depends on the iterator implementation. It could. 

This is also an existing piece of code (`streamedLookupUsersetFromIterator`) that was made generalized and moved into this package. 

But as you noted, in the very worst case that the iterator's implementation is sub optimal, it will inevitably exit due to context cancelation.

internal/check/bottom_up.go
elbuo8: Explicitness. We could have made it passing the strategy name at the caller.

internal/check/check.go
adriantam: Thinking of at least logging it (or chaining errors - whatever is easiest).

internal/check/weight2.go
```diff
@@ -0,0 +1,175 @@
...
+	// Convert right iterator to channel for uniform processing
+	rightChan := iterator.ToChannel[string](ctx, rightIter, IteratorMinBatchThreshold)
+
+	var lastErr error
+	lastLeftVal := ""
+	lastRightVal := ""
+
+	// Process both channels concurrently
+	for leftChan != nil || rightChan != nil {
+		select {
```
senojj, line 89: There exists a scenario where `leftChan` is exhausted after having added at least one value to `leftSeen` and `rightChan` is exhausted after having added at least one value to `rightSeen` without an intersection.

In said scenario, this select will block until `ctx` is canceled.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,590 @@
...
+	}
+
+	out := make(chan *iterator.Msg, len(iterChans))
+	go func() {
+		recoveredError := panics.Try(func() {
+			resolver(ctx, iterChans, out)
+		})
+
+		if recoveredError != nil {
+			fmt.Println(recoveredError.AsError())
```
adriantam: Instead of fmt.Println - should this be via logger?  (maybe passed in logger to bottomUp)

internal/check/bottom_up.go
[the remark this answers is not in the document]
adriantam: In this case, let's add it as comment to make it obvious of the intention

internal/check/bottom_up.go
```diff
@@ -0,0 +1,590 @@
...
+			if a.GetObject() < b.GetObject() {
+				return -1
+			}
+			return 1
+		})
+	}
+	iterFilters := make([]iterator.FilterFunc[*openfgav1.TupleKey], 0, 1)
+	if len(edge.GetConditions()) > 1 || edge.GetConditions()[0] != authzGraph.NoCond {
+		iterFilters = append(iterFilters, BuildConditionTupleKeyFilter(ctx, s.model, edge.GetConditions(), req.GetContext()))
+	}
```
adriantam: This feels a bit weird.
Instead of append to nil, should we simply create a len(1) slice with BuildConditionTupleKeyFilter(...) as the only item in the if(...) condition. 
For example
```
iterFilters := []iterator.FilterFunc[*openfgav1.TupleKey]{}
if len(edge.GetConditions()) > 1 || edge.GetConditions()[0] != authzGraph.NoCond {
		iterFilters = []iterator.FilterFunc[*openfgav1.TupleKey]{BuildConditionTupleKeyFilter(ctx, s.model, edge.GetConditions(), req.GetContext()))
	}
```

internal/check/bottom_up.go
```diff
@@ -0,0 +1,590 @@
...
+	i := iterator.NewFilteredIterator(iter, iterFilters...)
```
adriantam: This code feels so identical to line 194-206 ... wondering any opportunity to abstract it out

internal/check/bottom_up.go
elbuo8: Shouldn't have been here.

internal/check/weight2.go
elbuo8: For each channel we listen close events (exhausted) and it gets set as `nil`. Thus exiting the for loop and returning false.

senojj: a `nil` channel within a `select` is infinitely blocking, no? In other words, if I were to have the following scenario:
```go
var ch1 chan int
var ch2 chan int

select {
case _, _ = <-ch1:
case _, _ = <-ch2:
}
```
The `select` statement would block indefinitely. However, in your case, the select would _eventually_ wake up when `ctx` is canceled.

internal/iterator/filter.go
```diff
@@ -0,0 +1,96 @@
...
+	mu        *sync.Mutex
+	once      *sync.Once
+	lastErr   error
+	onceValid bool
+}
+
+func (f *filter[T]) Stop() {
+	f.once.Do(func() {
+		f.iter.Stop()
+	})
```
senojj: To avoid a potential allocation from the closure capture:
```suggestion
	f.once.Do(f.iter.Stop)
```

internal/check/bottom_up.go
```diff
@@ -0,0 +1,580 @@
...
+
+func addValueToBatch(value string, batch []string, ctx context.Context, out chan<- *iterator.Msg) []string {
+	if value == "" {
+		return batch
+	}
+	batch = append(batch, value)
+	// Flush batch if needed
+	if len(batch) >= IteratorMinBatchThreshold {
+		concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch)}, out)
+		batch = make([]string, 0, IteratorMinBatchThreshold)
```
senojj: Since the initial capacity of a batch is constant, this may benefit from using a `sync.Pool` to reduce allocations; depends upon how hot this path could get.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,580 @@
...
+	for _, it := range iters {
+		it.Stop()
+	}
+}
+
+// resolveUnion implements a merge-style algorithm for the union of sorted iterators
+// handleStreamError sends an error through the output channel, sets batch to nil,
+// and returns true if the function should terminate.
+func resolveUnion(ctx context.Context, iters []storage.Iterator[string], out chan<- *iterator.Msg) {
+	batch := make([]string, 0, IteratorMinBatchThreshold)
```
senojj: Seeing this, my previous suggestion to look into a `sync.Pool` is strengthened.

internal/check/weight2.go
elbuo8: If both channels are nil, you don't go into the select anymore due to the `for` clause not matching anymore. These are covered by `continue` statements right after each channel is exhausted (in the case there is no early exit).

senojj: Ah yes, I totally missed the `for` condition. I kept reading it as a bare `for`. 

*puts glasses on*

senojj:
I don't see any obvious or glaring issues with this PR. 

I made a couple of comments regarding an optimization using a `sync.Pool` for places where we are (re)allocating buffers regularly. Is there any appetite to add that optimization now, or defer until later? If the decision is "defer" or "won't do", I'll provide an approval on this PR.

elbuo8:
@senojj added

senojj:
@elbuo8 I see what you did there to avoid the stupid lint check. FWIW, I usually disable that specific lint check via comment, because a slice is already a pointer.

In any case, take my approval.

senojj (approved):

senojj:
I do see a group of tests that failed due to a mysql connection never getting closed. Additionally, there seems to be in incorrect result given in another test.

Is this algorithm currently hooked up to any of the tests running with the experimental flag?

internal/check/bottom_up.go
```diff
@@ -308,10 +298,11 @@ func (b *batcher) handleError(err error) bool {
...
 	// Discard the batch and return the buffer to the pool
-	if b.buf != nil {
-		b.buf.items = b.buf.items[:0]
-		batchPool.Put(b.buf)
-		b.buf = nil
+	if b.items != nil {
+		b.items = b.items[:0]
+		//nolint:staticcheck
+		batchPool.Put(b.items)
+		b.items = nil
```
senojj: It appears that `batcher.close` is always called after a call to `batcher.handleError`. Could we rely on `batcher.close` always returning the buffer to the pool when `batcher.items != nil`?

elbuo8: Not always the case, check resolveDifference for example.

senojj: In the case that `resolveDifference` calls `batcher.handleError`, if `batcher.handleError` returns `true`, `resolveDifference` will return and `batcher.close` will then be executed as a deferred function.

The only time that `batcher.handlError` puts the buffer back in the pool is when it returns true. So it that case, is _seems_ like my suggestion holds true.

internal/check/bottom_up.go
```diff
@@ -0,0 +1,606 @@
...
+		}
+	}
+}
+
+func resolveIntersection(ctx context.Context, iters []storage.Iterator[string], out chan<- *iterator.Msg) {
+	batch := newBatcher(ctx, out)
+
+	defer func() {
+		batch.close(iters, nil)
+	}()
```
senojj, line 403: This closure capture is unnecessary and likely results in an allocation. Can be avoided by simply calling:
```go
defer batch.close(iters, nil)
```

internal/check/bottom_up.go
```diff
@@ -0,0 +1,606 @@
...
+		}
+	}
+}
+
+func resolveDifference(ctx context.Context, iters []storage.Iterator[string], out chan<- *iterator.Msg) {
+	batch := newBatcher(ctx, out)
+
+	defer func() {
+		batch.close(iters, nil)
+	}()
```
senojj, line 497: This closure capture is unnecessary and likely results in an allocation. Can be avoided by simply calling:
```go
defer batch.close(iters, nil)
```

internal/check/bottom_up.go
```diff
@@ -0,0 +1,606 @@
...
+
+// resolveUnion implements a merge-style algorithm for the union of sorted iterators
+// handleStreamError sends an error through the output channel, sets batch to nil,
+// and returns true if the function should terminate.
+func resolveUnion(ctx context.Context, iters []storage.Iterator[string], out chan<- *iterator.Msg) {
+	batch := newBatcher(ctx, out)
+	lastError := error(nil)
+	defer func() {
+		batch.close(iters, lastError)
+	}()
```
senojj, line 312: Just a note to offset my other comments related to similar blocks of code.

This closure is necessary because `lastError` may be assigned a different value later on in the function. The arguments of a deferred function are evaluated at the time of declaration, so if you were not to use a closure here, `batch.close` would always receive the `error(nil)` value.

senojj:
> @elbuo8 I see what you did there to avoid the stupid lint check. FWIW, I usually disable that specific lint check via comment, because a slice is already a pointer.

Just wanted to note that I was wrong in an important way here. Given a slice value, such as:
```go
words := make([]string, 5)
```
`words` indeed has a pointer to a backing array, however, the value of `words` is actually a value type of a slice header `struct`.

Therefore, when passing a type like `[]T` to a `sync.Pool` `Put` method, a heap allocation is always created on each call due to interface boxing.

For this reason, the `sync.Pool` should be configured to create `*[]T` values when working with slices, so that a heap allocation is only created at the time of slice instantiation, and not each time the slice value is placed back into the `sync.Pool`.

If working with `*[]T` values is too cumbersome, then you can take the approach that you originally had where you were wrapping the slice in another `struct` type. Personally, I prefer to just work directly with `*[]T` types.

internal/graph/resolve_check_request.go
```diff
@@ -144,6 +157,8 @@ func (r *ResolveCheckRequest) clone() *ResolveCheckRequest {
 		LastCacheInvalidationTime: r.GetLastCacheInvalidationTime(),
 		invariantCacheKey:         r.GetInvariantCacheKey(),
 		SelectedStrategy:          r.GetSelectedStrategy(),
+		objectType:                r.objectType,
+		userType:                  r.userType,
```
senojj: According to our test `TestCloneResolveCheckRequest/empty_clone`, `(*ResolveCheckRequest).clone()` is sometimes called on a `nil` value. Attempting to access the fields of `r` directly will result in a panic in those cases.

Though I'm unsure that we actually consider cloning a `nil` request value to be correct.

internal/graph/resolve_check_request.go
```diff
@@ -224,3 +239,17 @@ func (r *ResolveCheckRequest) GetSelectedStrategy() string {
...
+	}
+	return r.objectType
+}
+
+func (r *ResolveCheckRequest) GetUserType() string {
+	if r == nil {
+		return ""
+	}
+	return r.userType
+}
```
senojj, line 255: Perhaps use these functions within the `(*ResolveCheckRequest).clone()` method to avoid panics in `nil` cases.

internal/check/check.go
```diff
@@ -0,0 +1,934 @@
...
+	relation := req.GetTupleKey().GetRelation()
+	objectType := tuple.GetType(req.GetTupleKey().GetObject())
+	objectRelation := tuple.ToObjectRelationString(objectType, relation)
+	ids := make([]string, 0, len(edges))
+	for _, edge := range edges {
+		id := buildEdgeCacheKey(r.model.GetModelID(), req, edge)
+		ids = append(ids, id)
+		if res, ok := r.isCached(req.GetConsistency(), id); ok {
+			concurrency.TrySendThroughChannel(ctx, ResponseMsg{ID: id, Res: res}, out)
+			continue
```
senojj, line 193: Could we not break here when the cached value is `true` and avoid creating any further goroutines?

PR #2821, opened 2025-11-20
fix: update throttling-related metadata & logging

justincoh:
ListObjects is currently not adding metadata to requests to indicate whether a request was throttled. We have the information available in the handler, but it looks like it was never added to the grpc context. Additionally, we now have two distinct throttling mechanisms and we need some way to differentiate between them.

#### How is it being solved?
Instead of a single `wasThrottled` metadata, this PR adds `DispatchThrottled` (our old throttling mechanism) and `DatastoreThrottled` to metadata for Check and ListObjects

**Note:** this currently only updates log fields, it does not make any changes to metrics.

## Review Checklist

## Summary by CodeRabbit

* **Refactor**
  * Refined throttling tracking to distinguish between dispatch-level and datastore-level throttling sources.
  * Updated API response metadata fields and logging tags to reflect separate throttling types.
  * Improved throttling detection logic to consider both throttling sources.

### DIFF
--- a/pkg/server/commands/check_command.go
+++ b/pkg/server/commands/check_command.go
@@ -168,9 +168,9 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
 
 	resolveCheckRequest.GetRequestMetadata().DatastoreThrottled.Store(dsMeta.WasThrottled)
 
-	// There are currently two possible throttling mechanisms, we need to know if either was triggered here.
-	wasThrottled := dsMeta.WasThrottled || resolveCheckRequest.GetRequestMetadata().DispatchThrottled.Load()
 	if err != nil {
+		// There are currently two possible throttling mechanisms, we need to know if either was triggered here.
+		wasThrottled := dsMeta.WasThrottled || resolveCheckRequest.GetRequestMetadata().DispatchThrottled.Load()
 		if errors.Is(err, context.DeadlineExceeded) && wasThrottled {
 			return resp, resolveCheckRequest.GetRequestMetadata(), &ThrottledError{Cause: err}
 		}


justincoh:
@coderabbit hit me

pkg/server/commands/reverseexpand/reverse_expand.go
```diff
@@ -224,8 +224,8 @@ type ResolutionMetadata struct {
 	// The number of times we are expanding from each node to find set of objects
 	DispatchCounter *atomic.Uint32
 
-	// WasThrottled indicates whether the request was throttled
-	WasThrottled *atomic.Bool
+	// DispatchThrottled indicates whether the request was throttled by dispatch count
+	DispatchThrottled *atomic.Bool
```
saad-h1, line 229: unrelated to this PR but just curious, do you know why reverse expand doesn't have datastore throttling configs?

pkg/server/commands/check_command.go
```diff
@@ -165,11 +165,13 @@ func (c *CheckQuery) Execute(ctx context.Context, params *CheckCommandParams) (*
 	dsMeta := datastoreWithTupleCache.GetMetadata()
 	resp.ResolutionMetadata.DatastoreQueryCount = dsMeta.DatastoreQueryCount
 	resp.ResolutionMetadata.DatastoreItemCount = dsMeta.DatastoreItemCount
-	// Until dispatch throttling is deprecated, merge the results of both
-	resolveCheckRequest.GetRequestMetadata().WasThrottled.CompareAndSwap(false, dsMeta.WasThrottled)
 
+	resolveCheckRequest.GetRequestMetadata().DatastoreThrottled.Store(dsMeta.WasThrottled)
+
+	// There are currently two possible throttling mechanisms, we need to know if either was triggered here.
+	wasThrottled := dsMeta.WasThrottled || resolveCheckRequest.GetRequestMetadata().DispatchThrottled.Load()
```
saad-h1: nit: could put this one line down, in the `if err != nil {}` block

saad-h1 (dismissed):

pkg/server/commands/reverseexpand/reverse_expand.go
justincoh: The datastore throttling info comes from the [RequestStorageWrapper here](https://github.com/openfga/openfga/blob/main/pkg/storage/storagewrappers/request.go#L47), which ReverseExpand _does_ use [via its datstore](https://github.com/openfga/openfga/blob/main/pkg/server/commands/reverseexpand/reverse_expand.go#L188), but it doesn't know anything about it. It gets surfaced at the ListObjects level instead.

ReverseExpand dispatch throttler is its [own separate thing](https://github.com/openfga/openfga/blob/main/pkg/server/commands/reverseexpand/reverse_expand.go#L726) (similar to dispatch_throttling_check_resolver), so that's why it has this extra piece of metadata.

PR #2837, opened 2025-12-03
rework the messaging between pipeline workers

senojj:
During the shadowing of the new list objects "pipeline" algorithm, metrics have been observed that indicate pipelines are occasionally remaining active beyond the lifetime of their context, indefinitely. The indicating metrics are "active goroutine count", and "memory allocation". Both increase steadily over time, as requests are issued, when the pipeline algorithm is active in shadow mode.

#### How is it being solved?
Firstly, new tests have been added that ensure the pipeline cleans up its resources appropriately in various scenarios.

Secondly, a major refactor of the pipeline code was performed. The pipeline workers, resolvers, and messaging pipes have all been overhauled to ensure that resources are cleaned up and all goroutines will end.

For the messaging pipe code, pipe buffers are now dynamically sized and can be "grown" at runtime. Additionally, the trackers have been removed from the pipe entirely and moved to a higher layer of abstraction in the pipeline code. Pipes are now more generalized and applicable in much broader scenarios.

In-flight message count tracking has been moved into the worker layer and is managed by a worker's "senders" and "listeners". In the previous implementation, a worker's messages remained "in-flight" until the subscribing worker finished processing each message. In this refactor, the "in-flight" message count is transferred to the subscribing worker immediately upon receipt of a message. This allowed the removal of a "hack" that was added to the intersection and exclusion resolvers in which a "fake" in-flight message was added to the worker's tracker for the lifetime of the resolver. That hack was initially necessary because intersection and exclusion resolvers do not immediately convert their input into output for their subscribers. Instead, their input is buffered until all senders have completed. Then action is taken on the buffered input to produce their output.

Resolvers have been reworked to increase efficiency of processing by introducing concurrent message processing within intersection and exclusion resolvers. A new `Bag` container type has been introduced that is useful for scenarios involving highly concurrent inserts that do not require ordering on iteration.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Configurable pipeline options (buffer size, chunk size, processor count)
  * New generic concurrent pipe and static receiver for message streaming
  * Introduced a thread-safe LIFO bag collection

* **Improvements**
  * Simplified context and tracing propagation in shadowed list operations
  * Refactored pipeline to a string-based flow with improved lifecycle and publish/subscribe handling
  * Exposed tracker API for broader reuse

* **Tests**
  * Added comprehensive pipeline shutdown tests; removed an old resolver test

### DIFF
--- a/internal/seq/reader.go
+++ b/internal/seq/reader.go
@@ -18,9 +18,8 @@ type SeqReader[T any] struct {
 
 // Read is a function that fills the given buffer with elements from the internal iter.Seq value.
 // A count of total items read into the buffer is returned. The count returned will be less than
-// the size of the buffer when fewer items remain in the internal iter.Seq value. When the count
-// returned is less than the length of the buffer, the sequence is complete and subsequent calls
-// to Read will return a count of 0.
+// the size of the buffer when fewer items remain in the internal iter.Seq value. In this case,
+// the sequence is complete and subsequent calls to Read will return a count of 0.
 func (r *SeqReader[T]) Read(buf []T) int {
 	var head int
 


internal/pipe/messaging.go
```diff
@@ -0,0 +1,190 @@
...
+	defer p.mu.Unlock()
+
+	replacement := make([]T, len(p.data)+n)
+	copy(replacement, p.data)
+	p.data = replacement
+
+	p.initialize()
+
+	p.full.Broadcast()
+}
```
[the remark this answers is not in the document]
senojj: This problem was addressed in commits 4703fc8 and 57a1956

pkg/server/commands/reverseexpand/containers/bag.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	for {
+		oldHead := b.head.Load()
+		tail.next = oldHead
+		if b.head.CompareAndSwap(oldHead, newHead) {
+			break
+		}
+	}
+}
```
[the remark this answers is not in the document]
senojj: This problem was addressed in commit 108b881

senojj:
@claude

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -739,145 +594,200 @@ type Item struct {
...
+	}
+
+	if l.ChunkSize <= 0 {
+		return l.send(items)
+	}
+
+	var sent bool
+
+	for len(items) > 0 {
+		ext := int(math.Min(float64(l.ChunkSize), float64(len(items))))
```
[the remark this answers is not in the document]
senojj: This logic is correct.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -344,220 +319,128 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
-		go func() {
-			defer wgRecursive.Done()
-			proc()
-		}()
+			wgStandard.Add(1)
+			go func() {
+				defer wgStandard.Done()
+				sentCount.Add(r.process(ctx, snd, lst, &outputBuffer))
+			}()
+		}
```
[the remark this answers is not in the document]
senojj: This is not an issue in our version of Go.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -739,145 +594,200 @@ type Item struct {
...
+
+	var cleanup containers.Bag[func()]
+
+	for i, snd := range senders {
+		for range snd.NumProcs {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				r.process(ctx, snd, &bags[i], &cleanup)
+			}()
```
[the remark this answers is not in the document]
senojj: This is not an issue in our version of Go.

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -344,220 +319,128 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
+func (r *baseResolver) Resolve(ctx context.Context, senders []*sender, lst pipe.Tx[[]Item]) {
+	ctx, span := pipelineTracer.Start(ctx, "baseResolver.Resolve")
+	defer span.End()
 
-	for ndx := range len(senders) {
-		r.inBuffers[ndx] = make(map[string]struct{})
-		r.errBuffers[ndx] = make(map[string]struct{})
-	}
+	var outputBuffer sync.Map
+	defer outputBuffer.Clear()
```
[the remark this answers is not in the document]
senojj: We specify go 1.24 as the project minimum.

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -344,220 +319,128 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
+		ctx, span := pipelineTracer.Start(ctx, "message.received", trace.WithAttributes(messageAttrs...))
+
+		for _, item := range msg.Value {
+			if item.Err != nil {
+				items = append(items, item)
+				continue
+			}
+
+			if isCyclical {
+				if _, loaded := inputBuffer.LoadOrStore(item.Value, struct{}{}); !loaded {
```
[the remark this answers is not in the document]
senojj: When an Item contains an error, continue is called on line 371.

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -945,64 +891,75 @@ func (p *Pipeline) Build(ctx context.Context, source Source, target Target) iter
...
-
-			runtime.Gosched()
-		}
-	}()
 
-	return func(yield func(Item) bool) {
-		defer span.End()
-		defer wg.Wait()
-		defer cancel()
+				runtime.Gosched()
```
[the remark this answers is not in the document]
senojj: Addressed in commit 64ae9bf

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -571,162 +454,134 @@ func (r *baseResolver) resolve(senders []*sender, listeners []*listener) {
...
-	status      *StatusPool
-	trk         tracker
 }
 
-func (r *exclusionResolver) resolve(senders []*sender, listeners []*listener) {
-	defer func() {
-		r.trk.Add(-1)
-		r.reporter.Report(false)
-	}()
+func (r *exclusionResolver) process(ctx context.Context, snd *sender, items *containers.Bag[Item], cleanup *containers.Bag[func()]) {
```
Vic-Dev: The intersection and exclusion process methods are the same, should we consolidate them?

internal/containers/bag.go
```diff
@@ -0,0 +1,66 @@
...
+		n := &node[T]{
+			value: i,
+		}
+		if newHead == nil {
+			newHead = n
+			tail = n
+			continue
+		}
+		n.next = newHead
+		newHead = n
```
Vic-Dev, line 41:
```suggestion
		if newHead == nil {
			tail = &node[T]{
				value: i,
			}
			newHead = tail
			continue
		}

		newHead = &node[T]{
			value: i,
			next:  newHead,
		}
```

senojj: any particular reason for this change?

Vic-Dev: just a nit, found it a little easier to read

senojj:
@claude

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -739,146 +685,153 @@ type Item struct {
...
+	bags := make([]containers.Bag[Item], len(senders))
+
+	var cleanup containers.Bag[func()]
+
+	for i, snd := range senders {
+		for range r.numProcs {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				r.process(ctx, snd, &bags[i], &cleanup)
```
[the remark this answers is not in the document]
senojj: This is not an issue in the version of Go that we are targeting.

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -739,146 +685,153 @@ type Item struct {
...
 
-	for item := range seq {
-		items = append(items, item)
+	var sentCount int
+
+	for items := range chunk(itemSeq, r.chunkSize) {
+		r.tracker.Add(1)
+		msg := NewMessage(items, r.tracker.Dec)
+		for _, lst := range listeners {
+			if !lst.Send(msg) {
```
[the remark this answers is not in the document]
senojj: This is a broadcast pattern and each message should account for its own allotment of the in-flight count.

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -336,399 +331,350 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
+			return !loaded
+		})
 
-		for _, lst := range listeners {
-			lst.send(outGroup)
+		for items := range chunk(results, r.chunkSize) {
+			for _, lst := range listeners {
+				r.tracker.Add(1)
+				msg := NewMessage(items, r.tracker.Dec)
+				if !lst.Send(msg) {
```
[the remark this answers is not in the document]
senojj: These are different values.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -336,399 +331,350 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
+func (r *baseResolver) Resolve(ctx context.Context, senders []Sender[*Edge, *Message], listeners []Listener[*Edge, *Message]) {
+	ctx, span := pipelineTracer.Start(ctx, "baseResolver.Resolve")
+	defer span.End()
 
-	for ndx := range len(senders) {
-		r.inBuffers[ndx] = make(map[string]struct{})
-		r.errBuffers[ndx] = make(map[string]struct{})
-	}
+	var outputBuffer sync.Map
+	defer outputBuffer.Clear()
```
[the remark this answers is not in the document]
senojj: The wait groups do provide this guarantee.

internal/containers/bag.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	for {
+		oldHead := b.head.Load()
+		tail.next = oldHead
```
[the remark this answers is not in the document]
senojj: This is an add-only data structure. It would not be possible to encounter the ABA problem.

internal/pipe/messaging.go
```diff
@@ -0,0 +1,202 @@
...
+	p.data[p.head] = item
+	p.head = (p.head + 1) % len(p.data)
+	p.count++
+
+	// Signal that the buffer is no longer empty.
+	p.empty.Signal()
+	return true
+}
+
+func (p *Pipe[T]) Recv(t *T) bool {
```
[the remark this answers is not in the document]
senojj: The value is always copied from `p.data` into `*t`.

internal/pipe/messaging.go
```diff
@@ -0,0 +1,202 @@
...
+	replacement := make([]T, len(p.data)+n)
+
+	for i := 0; i < p.count; i++ {
+		replacement[i] = p.data[(p.tail+i)%len(p.data)]
+	}
+	p.data = replacement
+	p.tail = 0
+	p.head = p.count
+
+	p.init.Do(p.initialize)
```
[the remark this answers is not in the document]
senojj: Grow holds a lock, so no other goroutines could initialize the pipe during that time.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -336,399 +331,350 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
-	// indicating that the worker is ready for cleanup once all messages have finished processing.
 	r.reporter.Report(false)
 
+	r.reporter.Wait(func(s bool) bool {
+		return !s
+	})
+
+	// Once the standard senders have all finished processing, wait until all messages have finished
+	// processing.
+	r.tracker.Wait(func(i int64) bool {
```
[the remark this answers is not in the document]
senojj: It uses a sync.Cond

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -1350,62 +1202,92 @@ type Target struct {
...
-
 	go func() {
 		defer w.wg.Done()
-		defer w.close()
+		defer w.Close()
+		w.Resolver.Resolve(ctx, w.senders, w.listeners)
+	}()
 
-		w.resolver.resolve(w.senders, w.listeners)
+	w.wg.Add(1)
```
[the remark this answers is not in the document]
senojj: The context is created within the worker and canceled on worker close.

Vic-Dev (dismissed):

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -916,108 +869,133 @@ func (o *omniInterpreter) interpret(ctx context.Context, edge *Edge, items []Ite
...
-	}()
 
-	return func(yield func(Item) bool) {
-		defer span.End()
-		defer wg.Wait()
-		defer cancel()
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			<-ctx.Done()
```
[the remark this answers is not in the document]
senojj: addressed in commit 644b51e

Vic-Dev (approved):

Vic-Dev (dismissed):

internal/containers/bag.go
adriantam, line 1: 1. Should this be specific to reverseexpand or it can be in a generic place?
2. Should we have unit test?  (This seems like a good candidate)

internal/pipe/messaging.go
```diff
@@ -0,0 +1,202 @@
...
+	tail  int
+	count int
+	done  bool
+	mu    sync.Mutex
+	full  *sync.Cond
+	empty *sync.Cond
+	init  sync.Once
+}
+
+func (p *Pipe[T]) Grow(n int) {
```
adriantam, line 37: Should the caller call this or this is called internally by the Pipe?  If it is internally, should we make this public?  If it is called externally, what will be the use case?

I see that the only time this is used (outside of test) is during initialization.  In this case, should this be hidden in the initialization function?

The reason I am asking this is that the caller of Pipe does not know its size (nor should it)...it feels a bit weird to be the caller's responsibility to grow the pipe's size when it doesn't know what size the pipe is to begin with.

internal/pipe/messaging.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+type TxCloser[T any] interface {
+	Tx[T]
+	io.Closer
+}
+
+const defaultPipeSize int = 1
+
+type Pipe[T any] struct {
```
adriantam, line 25: Would be good to assert Pipe satisfy the above interface in compile time

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -336,399 +342,411 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
+		if head > 0 {
+			yield(chunk{
+				Values: buffer,
+				Size:   head,
+			})
+		}
+	}
+}
+
+type mmap[K comparable, V any] struct {
```
adriantam: This seems to be a generic thing - should this be moved into a different location?

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -336,399 +342,411 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
-		}
+	numProcs int
+}
 
-		attrs = append(
-			attrs,
-			attribute.String("edge.to", edgeTo),
-			attribute.String("edge.from", edgeFrom),
-		)
+func (r *baseResolver) process(ctx context.Context, snd Sender[*Edge, *Message], listeners []Listener[*Edge, *Message], inputBuffer *mmap[string, struct{}], outputBuffer *mmap[string, struct{}]) int64 {
```
adriantam: mind adding comments what this function tries to do - so that others can quickly understand the intention without going deep inside :)

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -336,399 +342,411 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
 
-			if isRecursive {
-				recursive = append(recursive, proc)
+			for range r.numProcs {
+				wgRecursive.Add(1)
+				go func() {
+					defer wgRecursive.Done()
+					sentCount.Add(r.process(ctx, snd, listeners, &inputBuffer, &outputBuffer))
+				}()
 				continue
```
adriantam: Curious why do we need continue here?

pkg/server/commands/reverseexpand/pipeline.go
```diff
@@ -336,399 +342,411 @@ func (b *Backend) query(ctx context.Context, input queryInput) iter.Seq[Item] {
...
+		r.bufferPool.Put(values)
 
-	procExcluded = func(msg message[group]) bool {
-		attrs := []attribute.KeyValue{
-			attribute.String("resolver", "exclusion"),
-			attribute.Int("item.count", len(msg.Value.Items)),
-			attribute.Int("sender.index", 1),
+		// If there are no unseen items, skip processing.
+		if len(unseen) == 0 {
+			goto AfterInterpret
```
adriantam: Seeing this makes me wonder whether it should be the responsibility of the Interpret function to be smart and handle when len(unseen) == 0 and quickly returns nil.  In this way, the code will be cleaner and we don't need the goto

pkg/server/commands/reverseexpand/pipeline/track/reporting.go
```diff
@@ -1,35 +1,66 @@
...
 
 import (
 	"sync"
-	"sync/atomic"
 )
 
-type tracker interface {
-	Add(int64) int64
-	Load() int64
+type Tracker struct {
```
adriantam, line 10: maybe some comment on what Tracker does?

internal/pipe/messaging.go
senojj: I don't really see the benefit of that practice. If you attempt to use a Pipe as that interface, but it does not implement it, you will get a compile time error.

internal/pipe/messaging.go
senojj: Grow is a part of the Pipe's public interface so that it can be grown dynamically by the user. It is not a requirement to grow the pipe's size, but it is a feature to be used when it makes sense to do so. The entity that instantiates the pipe absolutely knows its size, so that is the audience of this function. Not necessarily the receiver of the pipe as it is passed around. If you don't want to expose all of the Pipe's methods, then use the Rx and Tx interfaces.

pkg/server/commands/reverseexpand/pipeline.go
senojj: Just because it is generic doesn't necessarily make it a candidate for a separate package. However, I am open to arguments that suggest why it might be useful in other places.

pkg/server/commands/reverseexpand/pipeline.go
senojj: I moved this logic into the interpreter.

pkg/server/commands/reverseexpand/pipeline.go
senojj: It was superfluous. Good catch.

pkg/server/commands/reverseexpand/pipeline/track/reporting.go
senojj: Added comments.

internal/containers/bag.go
senojj: Moved to an internal package. Will add unit tests.

pkg/server/commands/reverseexpand/pipeline.go
senojj: moved to an internal package.

internal/pipe/messaging.go
senojj: Additionally, Grow doesn't require knowledge of the initial or current size of the Pipe. It is simply stating that n number of slots should be added to its buffer. The interface of this function is very similar to the bytes.Buffer type's Grow function in the standard library.

internal/pipe/messaging.go
```diff
@@ -0,0 +1,202 @@
...
+	p.count++
+
+	// Signal that the buffer is no longer empty.
+	p.empty.Signal()
+	return true
+}
+
+func (p *Pipe[T]) Recv(t *T) bool {
+	p.mu.Lock()
+
```
adriantam, line 115: Curious why do we need to do individual p.mu.Unlock() vs. defer p.mu.Unlock()

internal/seq/reader.go
```diff
@@ -0,0 +1,54 @@
...
+	// stop is the function returned by iter.Pull that signals that the iter.Seq will
+	// no longer be iterated.
+	stop func()
+}
+
+// Read is a function that fills the given buffer with elements from the internal iter.Seq value.
+// A count of total items read into the buffer is returned. The count returned will be less than
+// the size of the buffer when fewer items remain in the internal iter.Seq value. When the count
+// returned is less than the length of the buffer, the sequence is complete and subsequent calls
+// to Read will return a count of 0.
```
adriantam:
```suggestion
// A count of total items read into the buffer is returned. The count returned will be less than
// the size of the buffer when fewer items remain in the internal iter.Seq value. 
// In this case, the sequence is complete and subsequent calls
// to Read will return a count of 0.
```

adriantam (approved):

internal/pipe/messaging.go
senojj: We don't. This is just an artifact of the refactor from the previous implementation.

internal/seq/reader.go
senojj: change applied

internal/pipe/messaging.go
senojj: change applied.

PR #2839, opened 2025-12-04
feat: Track datastore throttling in metrics and in BatchCheck, ListUsers metadata

justincoh:
## Description

In #2821 we split out `dispatch_throttling` and `datastore_throttling` in our request metadata for Check and ListObjects, but we forgot to do it in BatchCheck and ListUsers. This PR adds the info to those.

This PR also adds a label for the `throttledRequestCounter` metric to different between `dispatch` and `datastore` throttling types.

We also noticed ListUsers doesn't have throttling enabled - will follow up on that in a separate PR.

## Review Checklist

## Summary by CodeRabbit

* **Chores**
  * Enhanced throttling metrics tracking in batch check operations with improved visibility into datastore throttling states.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Changed
+- Datastore throttling separated from dispatch throttling in BatchCheck, ListUsers metadata. Also, `throttling_type` label added to `throttledRequestCounter` metric to differentiate between dispatch/datastore throttling. [#2839](https://github.com/openfga/openfga/pull/2839)
 
 ## [1.11.2] - 2025-12-04
 ### Fixed


pkg/server/batch_check.go
```diff
@@ -102,13 +102,14 @@ func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckReques
...
+	var dispatchThrottled bool
 
 	if metadata.ThrottleCount > 0 {
-		throttled = true
+		dispatchThrottled = true
 		throttledRequestCounter.WithLabelValues(s.serviceName, methodName).Add(float64(metadata.ThrottleCount))
 	}
-	grpc_ctxtags.Extract(ctx).Set("request.throttled", throttled)
+	grpc_ctxtags.Extract(ctx).Set("request.dispatch_throttled", dispatchThrottled)
+	grpc_ctxtags.Extract(ctx).Set("request.datastore_throttled", metadata.ThrottleCount)
```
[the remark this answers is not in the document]
justincoh: correct that was a bad autocomplete

pkg/server/batch_check.go
```diff
@@ -102,13 +102,14 @@ func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckReques
...
-	var throttled bool
+	var dispatchThrottled bool
 
 	if metadata.ThrottleCount > 0 {
-		throttled = true
+		dispatchThrottled = true
 		throttledRequestCounter.WithLabelValues(s.serviceName, methodName).Add(float64(metadata.ThrottleCount))
 	}
-	grpc_ctxtags.Extract(ctx).Set("request.throttled", throttled)
+	grpc_ctxtags.Extract(ctx).Set("request.dispatch_throttled", dispatchThrottled)
```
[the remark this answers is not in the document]
justincoh: updated

pkg/server/commands/batch_check_command.go
```diff
@@ -212,10 +214,14 @@ func (bq *BatchCheckQuery) Execute(ctx context.Context, params *BatchCheckComman
...
 
 			if metadata != nil {
-				if metadata.DispatchThrottled.Load() || metadata.DatastoreThrottled.Load() {
+				if metadata.DispatchThrottled.Load() {
 					totalThrottleCount.Add(1)
 				}
 				totalDispatchCount.Add(metadata.DispatchCounter.Load())
+
+				if metadata.DatastoreThrottled.Load() {
+					wasDatastoreThrottled.Store(true)
```
senojj: An accumulator would be more generally useful than a boolean value here. Why not have it match what was done for dispatch throttling, and use an atomic integer?

saad-h1: Updated.

I noticed the dispatch throttling count was only used for the metrics but was being mixed with the datastore throttling count in other places, so I separated that metric out. That also led me to notice ListUsers wasn't differentiating how it gets throttled, so I made that change as well. It actually looks like datastore throttling is always disabled for ListUsers (need to provide `ThrottlingEnabled` in `&storagewrappers.Operation` [here](https://github.com/openfga/openfga/blob/aff85c79dfb31b5ff7e8991d6e9d5b40f1af9092/pkg/server/commands/listusers/list_users_rpc.go#L182-L185) to enable it but that's not provided). I'll look into that separately from this PR.

pkg/server/server.go
```diff
@@ -107,10 +107,16 @@ var (
...
-		Name:      "throttled_requests_count",
-		Help:      "The total number of requests that have been throttled.",
+		Name:      "dispatch_throttled_requests_count",
+		Help:      "The total number of requests that have been dispatch throttled.",
+	}, []string{"grpc_service", "grpc_method"})
+
+	datastoreThrottledRequestCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "datastore_throttled_requests_count",
+		Help:      "The total number of requests that have been datastore throttled.",
```
senojj: Could these be kept as a single "throttledRequestCounter" metric and add a dimension like "kind" which can either be "dispatch" or "datastore"?

Or does it make more sense to have them separate for a particular reason?

saad-h1: Hm I was thinking about that too, I know the hope is to remove dispatch throttling eventually so I initially thought separating these out would make it easier to clean that up. But now that I think about it again, it shouldn't be much harder to clean up a label and maybe we should avoid changing the metric name in case people have built some tooling around it. I'll separate these through the labels then.

PR #2847, opened 2025-12-15
fix: Cache controller always using no-op logger

saad-h1:

Cache controller always uses a no-op logger:

- When creating a NewCacheController, the [default logger](https://github.com/openfga/openfga/blob/0ad13d78133fcb1045f0b9f5fee11574a0d63bb2/internal/cachecontroller/cache_controller.go#L132) is a noop.
- It can be overridden when [applying options](https://github.com/openfga/openfga/blob/0ad13d78133fcb1045f0b9f5fee11574a0d63bb2/internal/cachecontroller/cache_controller.go#L135-L137) with the [WithLogger](https://github.com/openfga/openfga/blob/0ad13d78133fcb1045f0b9f5fee11574a0d63bb2/internal/cachecontroller/cache_controller.go#L90) option.
- When an openfga server is started, it [does provide](https://github.com/openfga/openfga/blob/0ad13d78133fcb1045f0b9f5fee11574a0d63bb2/internal/shared/shared.go#L83) that option using the server’s logger (s.Logger).
- The server’s logger is also a [noop](https://github.com/openfga/openfga/blob/0ad13d78133fcb1045f0b9f5fee11574a0d63bb2/internal/shared/shared.go#L66) at this point since the option to apply a different server logger happens [after](https://github.com/openfga/openfga/blob/0ad13d78133fcb1045f0b9f5fee11574a0d63bb2/internal/shared/shared.go#L107-L109) the cache controller is created.. I.e., we’re setting the cache controller logger to be a noop logger still.

#### How is it being solved?

Pass the server's logger to it.

Apply SharedResources options before cache controller creation, but then add checks to make sure we don't overwrite a custom cache controller with the default one.

## References

n/a

## Review Checklist
- [x] (n/a) I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [x] (n/a) I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed resource initialization to prevent unintended overrides of custom cache and shadow cache controllers.
  * Custom shadow cache defaults are now properly respected during initialization.

* **Tests**
  * Added test coverage for custom logger, cache controller, and shadow cache controller configurations.


internal/shared/shared.go
```diff
@@ -80,15 +85,23 @@ func NewSharedDatastoreResources(
 	}
 
 	if settings.ShouldCreateCacheController() {
-		s.CacheController = cachecontroller.NewCacheController(ds, s.CheckCache, settings.CacheControllerTTL, settings.CheckQueryCacheTTL, settings.CheckIteratorCacheTTL, cachecontroller.WithLogger(s.Logger))
+		// Only create a cache controller if it wasn't already set via opts.
+		if _, isNoop := s.CacheController.(*cachecontroller.NoopCacheController); isNoop {
```
adriantam: This feels a bit clunky - especially requiring asserting for type.  
Maybe do something like

```
   defaultCacheController := cachecontroller.NewNoopCacheController()
s := &SharedDatastoreResources{
		WaitGroup:         &sync.WaitGroup{},
		SingleflightGroup: sharedSf,
		ServerCtx:         sharedCtx,
		CacheController:   defaultCacheController,
		Logger:            logger.NewNoopLogger(),
		SharedIteratorStorage: sharediterator.NewSharedIteratorDatastoreStorage(
			sharediterator.WithSharedIteratorDatastoreStorageLimit(
				int(settings.SharedIteratorLimit))),
	}
    for _, opt := range opts {
		opt(s)
	}
	...
	if settings.ShouldCreateCacheController() && s.CacheController == defaultCacheController {
	...
	}

```

saad-h1: updated

PR #2851, opened 2025-12-16
remove timeout causing a flaky test

senojj:
## Description
This PR removes an explicit timeout used within a list objects test.

I believe this timeout was intended to eliminate an unnecessary wait period when the test "deadline" was configured lower than the mock datastore's artificially imposed latency.

However, this timeout ultimately introduces a race condition when the mock datastore takes longer to return than 1 second, for whatever reason. In this scenario, the test will fail because the timeout triggered and no results had yet been produced.

Additionally, the existence of the timeout invalidates the results of the test scenario that it is intended to exercise, because it doesn't allow the query enough time to errantly produce results.

It is better to wait out the duration of the artificial latency and observe the result. This creates a deterministic test for all cases.

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Simplified internal test implementation for object listing streaming logic.

### DIFF
--- a/internal/cachecontroller/cache_controller.go
+++ b/internal/cachecontroller/cache_controller.go
@@ -95,7 +95,8 @@ func WithLogger(logger logger.Logger) InMemoryCacheControllerOpt {
 
 // InMemoryCacheController will invalidate cache iterator (InMemoryCache) and sub problem cache (CachedCheckResolver) entries
 // that are more recent than the last write for the specified store.
-// Note that the invalidation is done asynchronously, and only after a Check request is received.
+// Note that the invalidation is done asynchronously, triggered by Check requests,
+// or List Objects requests when list objects iterator cache is enabled.
 // It will be eventually consistent.
 type InMemoryCacheController struct {
 	ds    storage.OpenFGADatastore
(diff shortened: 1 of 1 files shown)


adriantam (commented):
Does the test `respects_max_results_when_deadline_timeout_and_returns_no_error_and_no_results` do anything now?  Or we should remove the test and remove the readTuplesDelay parameter?

senojj:
> Does the test `respects_max_results_when_deadline_timeout_and_returns_no_error_and_no_results` do anything now? Or we should remove the test and remove the readTuplesDelay parameter?

@adriantam it ensures that no results are returned when the deadline is exceeded, does it not?

adriantam:
> > Does the test `respects_max_results_when_deadline_timeout_and_returns_no_error_and_no_results` do anything now? Or we should remove the test and remove the readTuplesDelay parameter?
> 
> @adriantam it ensures that no results are returned when the deadline is exceeded, does it not?

If I understand the test correctly, the test is intended to [simulate timeout](https://github.com/openfga/openfga/blob/8ea4cf162f3572e36336fe059b81d82754beee2f/pkg/server/test/list_objects.go#L201).  Now that we remove the timeout, do we still need this test?

senojj:
> If I understand the test correctly, the test is intended to [simulate timeout](https://github.com/openfga/openfga/blob/8ea4cf162f3572e36336fe059b81d82754beee2f/pkg/server/test/list_objects.go#L201). Now that we remove the timeout, do we still need this test?

The test case is testing what happens when the list objects deadline is exceeded. However, the timeout that was removed would cause the _test itself_ to timeout, not the list objects query.

PR #2865, opened 2026-01-08
fix: do not allow model where intersection / union children is less than 2

adriantam:
Currently, typesystem allows model where the relation is an intersection or union is empty.  This can cause problem when running check / list objects / list users because the graph system expects intersection to have at least 1 child.

In reality, it is invalid to define an intersection/union with less than 2 children since it is not expressible as DSL.  Thus, we should mark these models as invalid.

#### How is it being solved?
When validating intersection and union, check to see whether the number of children is less than 2.

Update typesystem's validation code.

## References
Added similar unit test to assert in [language weighted graph](https://github.com/openfga/language/pull/523)

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for union and intersection type definitions to ensure they contain at least 2 children, with descriptive error messages.

* **Tests**
  * Added test coverage for invalid type definitions with insufficient children.

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -628,7 +628,7 @@ func (q *ListObjectsQuery) Execute(
 			return nil, serverErrors.ValidationError(fmt.Errorf("user: %s relation: %s not in graph", objectType, objectID))
 		}
 
-		seq := pl.Build(ctx, source, target)
+		seq := pl.Build(timeoutCtx, source, target)
 
 		var res ListObjectsResponse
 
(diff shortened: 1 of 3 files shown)


pkg/typesystem/typesystem.go
```diff
@@ -1045,6 +1045,9 @@ func hasEntrypoints(
 	case *openfgav1.Userset_Union:
 		// At least one type must have an entrypoint.
 		loop := false
+		if len(rw.Union.GetChild()) < 2 {
+			return false, false, fmt.Errorf("invalid type definition for '%s#%s' as union has less than 2 children", typeName, relationName)
+		}
```
senojj, line 1050: Could we create an error variable like: 
```go
var ErrInvalidModel = errors.New("invalid model")
```
And use it like:
```go
fmt.Errorf("%w: invalid type definition for '%s#%s': union has less than 2 children", ErrInvalidModel, typeName, relationName)
```
That way we could assert:
```go
if errors.Is(err, ErrInvalidModel) {
    // handle in some special way if necessary.
}
```

PR #2868, opened 2026-01-08
Fix pipeline context cancelation reporting

senojj:
Currently, a pipeline may return several context cancelation errors. This depends upon how many calls to an iterator are made before the pipeline shuts down.

The changes within this PR ensure that only a single context error is returned to the caller.

#### How is it being solved?
The `query` function used within a pipeline is altered so that it does not return context cancelation errors. There can be many that result from queries, so it is better to handle the returning of such an error to a higher-level abstraction.

The behavior of the sequence returned from `Pipeline::Build` is changed to ensure that it emits a final context error on shutdown, if the context was canceled for any reason.

## Review Checklist

## Release Notes

* **Bug Fixes**
  * Improved timeout and cancellation handling to prevent spurious error reporting when operations are interrupted by deadline or context cancellation.
  * Enhanced partial result handling during early termination of operations.

* **Tests**
  * Updated timeout and cancellation tests to verify correct error propagation behavior.

### DIFF
--- a/pkg/server/commands/reverseexpand/pipeline/pipeline.go
+++ b/pkg/server/commands/reverseexpand/pipeline/pipeline.go
@@ -1136,7 +1136,7 @@ func (pl *Pipeline) Build(ctx context.Context, source Source, target Target) ite
 
 		if !abandoned && ctx.Err() != nil {
 			// Context was canceled so there is no guarantee that all
-			// objects have been returned. An error must be signlaled
+			// objects have been returned. An error must be signaled
 			// here to indicate the possibility of a partial result.
 			yield(Item{Err: ctx.Err()})
 		}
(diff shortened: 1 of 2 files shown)


pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -1108,23 +1110,36 @@ func (pl *Pipeline) Build(ctx context.Context, source Source, target Target) ite
...
 						break
 					}
 				}
 			}
 			msg.Done()
 		}
+
+		if !abandoned && ctx.Err() != nil {
+			// Context was canceled so there is no guarantee that all
+			// objects have been returned. An error must be signlaled
```
Vic-Dev:
```suggestion
			// objects have been returned. An error must be signaled
```

PR #2870, opened 2026-01-11
Extend pipeline buffers dynamically

senojj:
## Description

This PR adds a dynamic buffer extension capability to pipes between workers within a ListObjects pipeline. This capability is enabled using two configuration parameters:
- extendAfter: When a call to a pipe's send method blocks for longer than this duration, the pipe's internal buffer capacity is doubled.
- maxExtensions: A pipe's internal buffer can be extended up to this number of times, before blocking indefinitely.

This feature is important for models that have tuple cycles or recursion as well as a high factor of tuple expansion. Using the following model:
```
type user

type group
    relations
        define member: [user, team#member]

type team
    relations
```
Within worker `team#member`, a query for `group` objects that have a `member` relationship to object `team:A` may result in a large number of returned `group` objects, and each `group` object will be added to the pipe `team#member -> group#member`.

Within worker `group#member`, each `group` object must then be queried for any `team` objects that have a `member` relation with the `group`. The output of those queries will be added to the pipe `group#member -> team#member`.

As the workers receive objects from their incoming pipes, query, and send the results to their outgoing pipes, the circular nature of this relationship may cause the pipes' internal buffers to fill to capacity, effectively blocking future calls to send on the pipe. If both the `team#member` worker and the `group#member` worker become blocked waiting to send to each other, the pipeline will become deadlocked until the pipeline's context is canceled.

This is not necessarily a "bug", because all systems have constraints. Were we to allow objects to be added to a pipe without any kind of limit, the application may attempt to allocate more memory than the host system has available, potentially resulting in OOM crashes.

#### How is it being solved?
There are a couple of solutions to this problem:
1. Set a pipe buffer size large enough to accommodate the necessary objects. This solution may allocate significantly more memory than necessary to facilitate object transfer between workers.
2. Add a feature to dynamically grow pipe buffers as necessary, up to a larger capacity threshold. This solution allows the initial pipe buffer size to be set lower, keeping pipe buffers smaller between workers that don't require the additional capacity, while accommodating the need for larger capacity.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable dynamic buffer extension for pipelines with adjustable extension timing and limits
  * Added buffer monitoring capabilities to check current size and capacity
  * Extended ListObjects operation with new pipeline extension configuration options

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -77,10 +77,12 @@ type ListObjectsQuery struct {
 	optimizationsEnabled bool // Indicates if experimental optimizations are enabled for ListObjectsResolver
 	useShadowCache       bool // Indicates that the shadow cache should be used instead of the main cache
 
-	pipelineEnabled bool // Indicates whether to run with the pipeline optimized code
-	chunkSize       int
-	bufferSize      int
-	numProcs        int
+	pipelineEnabled   bool // Indicates whether to run with the pipeline optimized code
+	chunkSize         int
+	bufferSize        int
+	numProcs          int
+	pipeExtendAfter   time.Duration
+	pipeMaxExtensions int
 }
 
 type ListObjectsResolver interface {
@@ -222,6 +224,13 @@ func WithListObjectsBufferSize(value int) ListObjectsQueryOption {
 func WithListObjectsNumProcs(value int) ListObjectsQueryOption {
 	return func(d *ListObjectsQuery) {
 		d.numProcs = value
+	}
+}
+
+func WithListObjectsPipeExtension(extendAfter time.Duration, maxExtensions int) ListObjectsQueryOption {
+	return func(d *ListObjectsQuery) {
+		d.pipeExtendAfter = extendAfter
+		d.pipeMaxExtensions = maxExtensions
 	}
 }
 
@@ -602,6 +611,10 @@ func (q *ListObjectsQuery) Execute(
 			options = append(options, pipeline.WithNumProcs(q.numProcs))
 		}
 
+		if q.pipeExtendAfter > 0 {
+			options = append(options, pipeline.WithPipeExtension(q.pipeExtendAfter, q.pipeMaxExtensions))
+		}
+
 		pl, err := pipeline.New(backend, options...)
 		if err != nil {
 			return nil, serverErrors.ValidationError(err)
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -808,6 +810,17 @@ func WithListObjectsNumProcs(value int) OpenFGAServiceV1Option {
...
+// When the ListObjects "pipeline" algorithm is enabled, this option enables extension
+// functionality within a pipeline, which dynamically extends the buffers between pipeline
+// workers, as needed. When a call to send on a buffer blocks for longer than the extendAfter
+// duration, the buffer size is doubled up to maxExtensions number of times.
+func WithListObjectsPipeExtension(extendAfter time.Duration, maxExtensions int) OpenFGAServiceV1Option {
+	return func(s *Server) {
+		s.listObjectsPipeExtendAfter = extendAfter
+		s.listObjectsPipeMaxExtensions = maxExtensions
+	}
+}
```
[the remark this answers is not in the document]
senojj: This is addressed in commit b265555

internal/pipe/messaging.go
```diff
@@ -219,11 +334,25 @@ func (p *Pipe[T]) Send(item T) bool {
 		return false
 	}
 
+	var timer *time.Timer
+	if p.full() && !p.done && p.extendAfter > 0 {
+		if p.maxExtensions < 0 || p.extendCount < p.maxExtensions {
+			currentSize := len(p.data)
+			timer = time.AfterFunc(p.extendAfter, func() {
+				p.extend(currentSize)
+			})
```
justincoh, line 343: Is there any concern with this happening in the background? It looks like `Send()` and `extend` acquire the same lock, so `extend` won't actually be able to execute until after `Send()` releases the lock. But before releasing the lock, `Send()` stops the timer. Will `time.AfterFunc` still execute, or is there a subtle race in here?

senojj: No race condition. The later call to `cond.Wait` will release the lock currently held by the call to `Send`. If/When the timer goroutine executes, it will be able to obtain the lock in order to extend the buffer.

pkg/server/commands/list_objects.go
```diff
@@ -602,6 +611,10 @@ func (q *ListObjectsQuery) Execute(
 			options = append(options, pipeline.WithNumProcs(q.numProcs))
 		}
 
+		if q.pipeExtendAfter > 0 {
```
justincoh, line 614: Do we want to verify that maxExtensions is !=0 here as well?

senojj: It isn't really necessary due to this condition:
```go
if p.maxExtensions < 0 || p.extendCount < p.maxExtensions {
    currentSize := len(p.data)
    timer = time.AfterFunc(p.extendAfter, func() {
        p.extend(currentSize)
    })
}
```
When `p.maxExtensions == 0` then `p.extendCount == p.maxExtensions` always.

PR #2874, opened 2026-01-13
fix: dial 127.0.0.1 when grpc server binds to 0.0.0.0

### LINKED ISSUE #2871: HTTP server's gRPC client should explicitly dial 127.0.0.1
### Checklist

### Description

Currently, when enabled, the HTTP server's gRPC client will dial the gRPC server using its bind address. See [here](https://github.com/openfga/openfga/blob/db7a7135f362f7a9cb3b077bc3ebe67ea52f9743/cmd/run/run.go#L626). Commonly, the gRPC server will be configured to bind to `0.0.0.0` indicating all available interfaces. However, the address `0.0.0.0` is not technically a valid address for a client to dial; on Linux it happens to route to `localhost`.

For configurations without TLS enabled for the gRPC server, this isn't a problem on most platforms. However, with TLS enabled for the gRPC server, when the HTTP server's gRPC client attempts to dial `0.0.0.0` the client will likely encounter a TLS error due to the address dialed not matching any of the addresses within the server's certificate.

Ideally, the HTTP server's gRPC client should dial `127.0.0.1` explicitly, and the server's certificate should include `127.0.0.1` within the SAN.

### OpenFGA version

latest

As a binary

In-Memory

### OpenFGA Flags

none

### Logs

_No response_

SamirZanata:
Fixes: #2871

## Description
This PR addresses an issue where the HTTP server's internal gRPC client attempts to dial `0.0.0.0` when the gRPC server is configured to bind to all interfaces. This behavior can cause TLS errors (certificate mismatch) or connectivity issues on certain environments.

## Changes
- Modified `dialGrpc` in `cmd/run/run.go`.
- Added logic to check if the configured `GRPC.Addr` host is `0.0.0.0`.
- If detected, the client now explicitly dials `127.0.0.1` instead, while preserving the configured port.
- Used `net.SplitHostPort` and `net.JoinHostPort` to ensure address parsing is robust.

## Verification
- Ran local tests using `go test -v ./cmd/run/...` and verified they pass (ignoring environment-specific Windows errors unrelated to this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GRPC connection handling by normalizing localhost addresses to ensure reliable connections to local services.


cmd/run/run.go
```diff
@@ -620,10 +620,18 @@ func (s *ServerContext) dialGrpc(config *serverconfig.Config) (*grpc.ClientConn,
 		dialOpts = append(dialOpts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	}
 
+	grpcAddr := config.GRPC.Addr
+	host, port, err := net.SplitHostPort(grpcAddr)
+	if err == nil {
```
zweizeichen: Now that you introduced a function that could return an error, you should probably bubble it up the callstack to be handled or panic depending on how the app is structured.

cmd/run/run.go
```diff
@@ -620,10 +620,18 @@ func (s *ServerContext) dialGrpc(config *serverconfig.Config) (*grpc.ClientConn,
...
+		if host == "0.0.0.0" {
```
zweizeichen: Please parse this into a proper address or add v6 support as the AI suggested (also see: https://stackoverflow.com/a/42480523, it might be a hostname). The 'cleanest' way would probably be parsing it into an address and using: https://pkg.go.dev/net/netip#Addr.IsUnspecified

zweizeichen (changes requested):
Thanks for taking the initiative! I'm not a maintainer, but a few things stood out to me here, maybe it saves someone else some time on the review.

cmd/run/run.go
```diff
@@ -620,10 +620,18 @@ func (s *ServerContext) dialGrpc(config *serverconfig.Config) (*grpc.ClientConn,
...
+			grpcAddr = net.JoinHostPort("127.0.0.1", port)
+		}
+	}
```
Vic-Dev:
```suggestion
    grpcAddr := config.GRPC.Addr
    host, port, err := net.SplitHostPort(grpcAddr)
    if err != nil {
        s.Logger.Fatal("failed to get gRPC dial target", zap.Error(err))
    }
    if host == "" {
	    grpcAddr := net.JoinHostPort("127.0.0.1", port)
    }
    if ip := net.ParseIP(host); ip != nil && ip.IsUnspecified() {
	    grpcAddr := net.JoinHostPort("127.0.0.1", port)
    }
```

Vic-Dev (changes requested):
Can you also add tests for this change? Can be easier to test if you extract the target parsing into it's own helper function

senojj:
Why not _always_ dial `127.0.0.1`? This will always be dialed from within the same process as the server. Yes, it will require the address to be added to the TLS certificate, but at least it will be predictable.

senojj:
Another, _better_ (IMHO) approach, might be to have the HTTP server communicate with the gRPC server over a unix domain socket, instead of TCP.

SamirZanata:
"Thanks for the feedback @Vic-Dev  and @senojj 

Regarding the implementation:

I will definitely add the tests and extract the parsing logic as suggested by @Vic-Dev.

@senojj, do you prefer we move strictly to Unix Domain Sockets right now, or should I stick to the 127.0.0.1 (TCP) approach for this PR to keep it simpler? Let me know which direction you prefer and I'll update the code."

zweizeichen:
> Another, _better_ (IMHO) approach, might be to have the HTTP server communicate with the gRPC server over a unix domain socket, instead of TCP.

Uh I mean at that point skipping the network/TLS entirely and going with shared memory/Go concurrency primitives might be the better option from a performance standpoint unless the HTTP server portion is code generated and it becomes a maintenance tradeoff. 

But I reckon this is something for another PR.

senojj:
> Uh I mean at that point skipping the network/TLS entirely and going with shared memory/Go concurrency primitives might be the better option from a performance standpoint unless the HTTP server portion is code generated and it becomes a maintenance tradeoff.

Using a UDS would skip the TCP stack and mitigate the need for TLS while sharing the gRPC server's implementation. This is the typical approach to exposing the same server over multiple sockets/interfaces.

> But I reckon this is something for another PR.

Yes, this would be a significantly larger change than just using `127.0.0.1` to dial in the client. However, it is something that I will put on our roadmap.

> do you prefer we move strictly to Unix Domain Sockets right now, or should I stick to the 127.0.0.1 (TCP) approach for this PR to keep it simpler?

@SamirZanata, you can move forward by using `127.0.0.1` to dial for the gRPC client.

SamirZanata:
Hi everyone, first of all, I apologize for the delay in getting back to this PR.

I've refactored the code to address all the feedback:

Refactoring: Extracted the logic into a standalone getGrpcDialTarget function to allow for unit testing, as suggested by @Vic-Dev .

Logic: The client now always dials 127.0.0.1 whenever the configured address is unspecified (e.g., :port, 0.0.0.0, or [::]), ensuring predictable behavior as requested by @senojj .

Robustness: Improved error handling using netip for modern IP parsing and ensured all errors from net.SplitHostPort are bubbled up and logged, addressing @zweizeichen  point.

Testing: Added comprehensive table-driven tests covering IPv4/IPv6 (unspecified and specific), hostnames, and malformed inputs.

Code Quality: Fixed potential variable shadowing and ensured idiomatic use of net.JoinHostPort and error wrapping (%w).

cmd/run/run_test.go.2315474557705495845
senojj, line 1: This file should not be committed.

senojj:
@SamirZanata there is an outstanding required change that must be addressed before we can merge this. Would you like to continue to move this forward?

SamirZanata:
Thanks for catching that, @senojj. The temporary test file was committed by mistake. I have cleaned up the branch and pushed the fix. Please let me know if there’s anything else needed to merge this

senojj:
> I have cleaned up the branch and pushed the fix. Please let me know if there’s anything else needed to merge this

@SamirZanata I don't see any new changes.

senojj:
This PR has become very stale, and the changes within touch code that another PR that I have submitted is changing #2937

I was attempting to delay my changes until this PR was merged, but it has been 2 weeks since the last communication on this PR.

Vic-Dev:
Thank you for your interest in contributing to OpenFGA! Unfortunately, due to inactivity we will be closing this PR.

PR #2877, opened 2026-01-14
Optimize pipeline exclusion

senojj:
## Description

This PR introduces a change to improve the efficiency of exclusion resolvers in the ListObjects pipeline algorithm.

Currently, an exclusion resolver will wait for both the "include" and "subtract" pipe branches to complete before producing any values to its listeners.

#### How is it being solved?
Once the "subtract" pipe branch has completed, values begin to stream from the exclusion resolver into its listeners as they are processed from the "include" pipe branch.

Instead of having the "include" processor collect values in a shared `containers.Bag`, a `pipe.Pipe` is provided instead.

In order to allow the "subtract" process to continue to use a shared `containers.Bag`, a `txBag` type is created as an alias to `containers.Bag` and extended with a function that satisfies the `pipe.Tx` interface.

Additional small changes are made to the Pipe type:
- A new default value of `-1` for the `DefaultExtendAfter` constant.
- A value of `-1` for `extendAfter` is now treated as the disabling value.
- A value of `0` for `extendAfter` is now treated as a "no wait" signal.
- A test was added to exercise unbounded dynamic growth.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management in pipe handling when buffers are empty.
  * Fixed extension trigger conditions to allow extension when thresholds are reached.

* **Tests**
  * Added test coverage for unbounded buffer extension scenarios.

* **Refactor**
  * Optimized internal pipeline processing to improve efficiency and reliability of item flow through the system.

### DIFF
--- a/internal/pipe/messaging.go
+++ b/internal/pipe/messaging.go
@@ -180,6 +180,7 @@ func Must[T any](n int) *Pipe[T] {
 // The buffer's size doubles after each extension.
 //
 // A negative extendAfter value will disable dynamic extension.
+// A 0 value may be provided for extendAfter to indicate immediate extension.
 //
 // A negative maxExtensions value will allow unlimited extensions.
 func (p *Pipe[T]) SetExtensionConfig(extendAfter time.Duration, maxExtensions int) {
--- a/pkg/server/commands/reverseexpand/pipeline/pipeline.go
+++ b/pkg/server/commands/reverseexpand/pipeline/pipeline.go
@@ -622,8 +622,11 @@ func (r *baseResolver) Resolve(ctx context.Context, senders []Sender[*Edge, *Mes
 
 type edgeHandler func(context.Context, *Edge, []string) iter.Seq[Item]
 
+// txBag is a type that implements the interface pipe.Tx[T]
+// for the type containers.Bag[T].
 type txBag[T any] containers.Bag[T]
 
+// Send implements the pipe.Tx[T] interface.
 func (tx *txBag[T]) Send(t T) bool {
 	bag := (*containers.Bag[T])(tx)
 	bag.Add(t)


internal/pipe/messaging.go
```diff
@@ -15,13 +15,13 @@ const (
...
-	DefaultExtendAfter time.Duration = 0
+	DefaultExtendAfter time.Duration = -1
 
 	// This value is used as the default value that indicates the maximum
 	// number of times that a pipe's internal buffer may be extended.
 	//
-	// A value of 0 will prevent extensions.
-	DefaultMaxExtensions int = 0
+	// A negative value will all unlimited extensions.
+	DefaultMaxExtensions int = -1
```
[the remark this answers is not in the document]
senojj: addressed in commit b84191c

internal/pipe/messaging.go
```diff
@@ -345,7 +349,7 @@ func (p *Pipe[T]) Send(item T) bool {
 	for p.full() && !p.done {
 		var timer *time.Timer
 
-		if p.extendAfter > 0 && (p.maxExtensions < 0 || p.extendCount < p.maxExtensions) {
+		if p.extendAfter >= 0 && (p.maxExtensions < 0 || p.extendCount < p.maxExtensions) {
 			currentSize := len(p.data)
 			timer = time.AfterFunc(p.extendAfter, func() {
```
adriantam, line 355: What does extendAfter == 0 means?

senojj: 0 is a "no wait" signal.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -622,6 +622,14 @@ func (r *baseResolver) Resolve(ctx context.Context, senders []Sender[*Edge, *Mes
 
 type edgeHandler func(context.Context, *Edge, []string) iter.Seq[Item]
 
+type txBag[T any] containers.Bag[T]
+
+func (tx *txBag[T]) Send(t T) bool {
+	bag := (*containers.Bag[T])(tx)
+	bag.Add(t)
+	return true
+}
```
adriantam, line 634: Curious why we will return a constant value (and we never check the value).  Is it because of interface consistency?

senojj: ?? The boolean result is to satisfy an interface `pipe.Tx`. `bag.Add` does not return a boolean.

senojj: I can add a comment

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -715,30 +723,38 @@ func (r *exclusionResolver) Resolve(ctx context.Context, senders []Sender[*Edge,
...
 
-	var included containers.Bag[Item]
 	var excluded containers.Bag[Item]
 
 	var cleanup containers.Bag[func()]
 
+	var wgExclude sync.WaitGroup
+
+	pipeInclude := pipe.Must[Item](1 << 7) // create a pipe with an initial capacity of 128
+	pipeInclude.SetExtensionConfig(0, -1)  // allow pipe to grow to an unbounded capacity with no wait
```
adriantam, line 737: Do we need some for of limit to protect against consuming too much memory?

senojj: We explicitly set the limit to unlimited here; it is the same behavior as previous implementation. The number of items is limited by objects for the relation.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
```diff
@@ -752,7 +768,7 @@ func (r *exclusionResolver) Resolve(ctx context.Context, senders []Sender[*Edge,
 		exclusions[item.Value] = struct{}{}
 	}
 
-	results := seq.Filter(included.Seq(), func(item Item) bool {
+	results := seq.Filter(pipeInclude.Seq(), func(item Item) bool {
```
adriantam, line 774: Is there any possibility the pipeInclude is closed in line 742 before we get here?  If so, would there be any negative effect?

senojj: If it is closed prior to here, the pipe will allow us to drain it.

pkg/server/commands/reverseexpand/pipeline/pipeline.go
adriantam: Yes, I am asking `return true`

pkg/server/commands/reverseexpand/pipeline/pipeline.go
adriantam: Maybe we should call it out here that the number of items is limited by objects for the relation - worry that for example if they have 200K items (or 1M items) for the relations, would it cause any problem in terms of memory.

senojj: Let's assume 1M items == 1MB of data, with a high enough RPS on such a query, yes we could exhaust available memory. However, note that the "subtract" branch also buffers unbounded.

senojj: intersection also works the same way.

senojj: Do we have such a limit in check?

adriantam: I think for now it should be fine.

PR #2886, opened 2026-01-21
Fix/mixed strategies

yissellokta:
## Description

Overview
This PR optimizes the planner's strategy selection logic for Weight 2 and Recursive relations. It addresses issues where high variance in latency caused the Bayesian sampler to return negative values, leading to incorrect strategy selection. It also fixes "stubbornness" in strategy adaptation and a logic bug in recursive strategy checks.

Key Issues Addressed
     1- Negative Sampling (The "Variance Explosion"):
               Problem: When the Default Strategy had high variance (e.g., mu ~1275ms, sigma ~ 866ms), the Normal distribution curve became wide enough that Sample() frequently returned negative numbers.
               Impact: The planner incorrectly interpreted negative samples as "zero latency" or "super fast," causing it to select the slower Default Strategy over the faster Weight 2 Strategy.
     2- Model "Stubbornness" (Inertia):
               Problem: High Lambda values in the Weight 2 Strategy created high inertia. If the strategy's performance degraded, the system took hundreds of rounds to adapt, potentially causing self-imposed latency spikes.
     3- Recursive Strategy Bug:
               Problem: Even with optimizations, the planner penalized the Optimized Recursive Strategy due to a bug in the check algorithm that mixed execution data with the Default Recursive Strategy.
    
Implemented Solutions
       1. Transition to Log-Normal Distribution
               To solve the negative sampling issue without folding probabilities (which doubles error rates), the mathematical model was switched from Normal to Log-Normal.
               Behavior: Log-Normal distributions cannot go below zero. When variance increases, the curve creates a heavy tail to the right (indicating risk of slowness) rather than dipping into negative numbers.
               Implementation: The initial mu is converted to log space; update and sample functions now maintain values in log space.
      2. Adaptive Lambda Capping
             To solve the "stubbornness" issue where the model refused to switch away from a degrading strategy:
             Fix: Capped lambda to a configurable maximum.
             Effect: Creates a "sliding window" effect, ensuring the planner only considers the most recent X requests, allowing it to adapt quickly when a previously "good" strategy turns bad.
      3. Recursive Parameter Tuning & Bug Fix
      Specific adjustments were made to the recursive strategy configurations:
      Parameter Updates:
           Optimized Recursive Strategy: Decreased lambda to 3 (was 5).
           Global Recursive Settings: Updated alpha = 3 and beta = 2.
           Logic Fix: Patched the check algorithm to prevent the mixing of strategies, ensuring the Optimized strategy is not penalized by the Default strategy's execution.

## Review Checklist

## Release Notes

* **Performance Improvements**
  * Implemented strategy reuse across nested resolution requests, reducing redundant computation and improving efficiency.
  * Optimized resolver parameters to enhance performance consistency and accuracy.

* **Tests**
  * Added comprehensive test coverage for strategy propagation and selection behavior.

### DIFF
--- a/internal/graph/recursive_resolver.go
+++ b/internal/graph/recursive_resolver.go
@@ -48,7 +48,7 @@ type recursiveMapping struct {
 	allowedUserTypeRestrictions []*openfgav1.RelationReference
 }
 
-func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, _ []*openfgav1.RelationReference, rightIter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
+func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, _ []*openfgav1.RelationReference, rightIter storage.TupleKeyIterator, _ string) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		typesys, _ := typesystem.TypesystemFromContext(ctx)
 
@@ -64,7 +64,7 @@ func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequ
 
 // recursiveTTU solves a union relation of the form "{operand1} OR ... {operandN} OR {recursive TTU}"
 // rightIter gives the iterator for the recursive TTU.
-func (c *LocalChecker) recursiveTTU(_ context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, rightIter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
+func (c *LocalChecker) recursiveTTU(_ context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, rightIter storage.TupleKeyIterator, _ string) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		typesys, _ := typesystem.TypesystemFromContext(ctx)
 
--- a/internal/graph/weight_two_resolver.go
+++ b/internal/graph/weight_two_resolver.go
@@ -48,7 +48,7 @@ var ErrShortCircuit = errors.New("short circuit")
 
 type fastPathSetHandler func(context.Context, *iterator.Streams, chan<- *iterator.Msg)
 
-func (c *LocalChecker) weight2Userset(_ context.Context, req *ResolveCheckRequest, usersets []*openfgav1.RelationReference, iter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
+func (c *LocalChecker) weight2Userset(_ context.Context, req *ResolveCheckRequest, usersets []*openfgav1.RelationReference, iter storage.TupleKeyIterator, _ string) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		cancellableCtx, cancel := context.WithCancel(ctx)
 		defer cancel()
@@ -68,7 +68,7 @@ func (c *LocalChecker) weight2Userset(_ context.Context, req *ResolveCheckReques
 	}
 }
 
-func (c *LocalChecker) weight2TTU(ctx context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, iter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
+func (c *LocalChecker) weight2TTU(ctx context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, iter storage.TupleKeyIterator, _ string) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
 		typesys, _ := typesystem.TypesystemFromContext(ctx)
 		objectType := tuple.GetType(req.GetTupleKey().GetObject())


internal/graph/check.go
```diff
@@ -900,6 +954,9 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 		keyPlan := c.planner.GetPlanSelector(planKey)
 		strategy := keyPlan.Select(possibleStrategies)
 
+		// Set the selected strategy on the request so child dispatches will use it
+		req.SelectedStrategy = strategy.Name
```
senojj: A data race was detected here in the tests. Two goroutines are attempting to write to `req.SelectedStrategy` at the same time.

internal/graph/check.go
```diff
@@ -683,13 +683,30 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
...
+			if selectedStrategy := req.GetSelectedStrategy(); selectedStrategy != "" {
+				if _, exists := possibleStrategies[selectedStrategy]; exists {
+					resolver := c.defaultUserset
+					if selectedStrategy == recursiveResolver {
+						resolver = c.recursiveUserset
+					}
+					return resolver(ctx, req, directlyRelatedUsersetTypes, iter)(ctx)
+				}
+				// If the selected strategy is not in the possible strategies, fall through to planner
+			}
```
adriantam: Can we under this as 

```
if  req.GetSelectedStrategy() == defaultResolver {
    return c.defaultUserset(ctx, req, directlyRelatedUsersetTypes, iter)(ctx)
}  else if req.GetSelectedStrategy() == recursiveResolver {
    return c.recursiveUserset(ctx, req, directlyRelatedUsersetTypes, iter)(ctx)
}  
```

since there are only two possibleStrategies ( selectedStrategy / recursiveResolver)?

internal/graph/check.go
```diff
@@ -714,6 +735,19 @@ func (c *LocalChecker) checkDirectUsersetTuples(ctx context.Context, req *Resolv
 			}
 			// NOTE: we collect defers given that the iterator won't be consumed until `union` resolves at the end.
 			defer iter.Stop()
+
+			// If a strategy was already selected, use it without re-planning
+			if selectedStrategy != "" {
```
adriantam, line 730: curious why do we need to check for != "" if we assert for exists in the next line

internal/graph/check.go
```diff
@@ -883,6 +920,23 @@ func (c *LocalChecker) checkTTU(parentctx context.Context, req *ResolveCheckRequ
 			return resolver(ctx, req, rewrite, filteredIter)(ctx)
 		}
 
+		// If a strategy was already selected by a parent call, use it without re-planning.
+		// This prevents the planner from being called again during recursive dispatch calls.
+		if selectedStrategy := req.GetSelectedStrategy(); selectedStrategy != "" {
```
adriantam, line 912: similarly why check for selectedStrategy != ""

internal/graph/check.go
senojj: I might suggest changing the various functions to accept a value of `ResolveCheckRequest` instead of a pointer, since it is passed recursively and modified.

Alternatively, `SelectedStrategy` would need to be an `atomic.Pointer`, or you could introduce a `sync.Mutex` into `ResolveCheckRequest` (undesirable).

internal/graph/check.go
senojj: Generally, `switch` blocks are more succinct than `else if` branches.
```go
switch req.GetSelectedStrategy() {
case defaultResolver:
    return c.defaultUserset(ctx, req, directlyRelatedUsersetTypes, iter)(ctx)
case recursiveResolver:
    return c.recursiveUserset(ctx, req, directlyRelatedUsersetTypes, iter)(ctx)
default:
}
```

internal/graph/check.go
yissellokta: one is for recursion use case, and the other one is for weight2 userset

internal/graph/check.go
yissellokta: because the selectStrategy will be empty if the planner hasn't run for that combination

adriantam: What I meant is that if selectedStrategy is not set, the next block (exists) will not satisfy anyways.

adriantam: this applies to other blocks that have the same pattern

internal/planner/thompson.go
```diff
@@ -34,15 +34,22 @@ func (ts *ThompsonStats) Sample(r *rand.Rand) float64 {
...
+	logSample := params.mu + stdNormal*math.Sqrt(variance)
 
-	return mean
+	// Convert back to Linear Time (Milliseconds)
+	// exp(x) is guaranteed to be positive for all real x.
+	val := math.Exp(logSample)
+	if val < 1.0 {
+		return 1.0 // Enforce a minimum of 1ms to avoid pathological cases
+	}
+	return val
```
adriantam, line 52: Since we are using math package, we can do something like
```suggestion
    // Enforce a minimum of 1ms to avoid pathological cases
	return math.Min(math.Exp(logSample), 1.0)
```

internal/planner/thompson.go
```diff
@@ -81,14 +88,26 @@ func (ts *ThompsonStats) fastGammaSample(r *rand.Rand, alpha, beta float64) floa
...
 func (ts *ThompsonStats) Update(duration time.Duration) {
-	x := float64(duration.Nanoseconds()) / 1e6 // Convert to milliseconds with higher precision
+	// 1 Convert to milliseconds with higher precision
+	rawMS := float64(duration.Nanoseconds()) / 1e6
+
+	// 2. Handle Edge Cases (0ms or negative duration)
+	// math.Log(0) is undefined (-Inf). We must enforce a small positive floor.
+	if rawMS <= 0.001 {
+		rawMS = 0.001
+	}
```
adriantam, line 98: Use math package
```suggestion
	// 1 Convert to milliseconds with higher precision
	// math.Log(0) is undefined (-Inf). We must enforce a small positive floor.
	rawMS := math.Min(float64(duration.Nanoseconds()) / 1e6, 0.001)
```

internal/planner/thompson.go
```diff
@@ -116,11 +135,20 @@ func (ts *ThompsonStats) Update(duration time.Duration) {
 func NewThompsonStats(initialGuess time.Duration, lambda, alpha, beta float64) *ThompsonStats {
 	initialMs := float64(initialGuess.Nanoseconds()) / 1e6
 
+	// Safety check: Log(0) is -Infinity. Clamp to a tiny value if needed.
+	if initialMs <= 0 {
```
adriantam, line 139: Want to double check whether you intend to check agains <=0 or <= 0.001

yissellokta: negative or 0 values will be set to a floor of 0.001

internal/planner/thompson.go
yissellokta: This is wrong, I am not taking the min, will be the max

internal/planner/thompson.go
yissellokta: It cannot be Min, it has to be Max. In my opinion, it is clear with the if condition, even though it has more lines of code. In terms of micro efficiency, the compiler will resolve the if faster than the function Max

internal/planner/thompson.go
yissellokta: IMO, I prefer to use the more verbose form, it is a little bit more micro-efficient than the math function and I think is clearer than one liner for the intend

internal/graph/check.go
yissellokta: We verify if the selected strategy != "" because only when the selectStrategy has a value then it is when we would run the resolver with the selected strategy. In the CheckTTU( .. ) that is the only place the selectedStrategy is being verified. If the question is why have that verification instead of the switch case to account for "" then, it is just code styling.

yissellokta: Can you clarify what you are requesting.

internal/graph/check.go
```diff
@@ -367,12 +367,13 @@ func (c *LocalChecker) Close() {
 
 // dispatch clones the parent request, modifies its metadata and tupleKey, and dispatches the new request
 // to the CheckResolver this LocalChecker was constructed with.
-func (c *LocalChecker) dispatch(_ context.Context, parentReq *ResolveCheckRequest, tk *openfgav1.TupleKey) CheckHandlerFunc {
+func (c *LocalChecker) dispatch(_ context.Context, parentReq *ResolveCheckRequest, tk *openfgav1.TupleKey, selectedStrategy string) CheckHandlerFunc {
```
senojj: Instead of passing both the `ResolveCheckRequest` and `selectedStrategy` here, and then cloning. Why not just clone and add the strategy just before passing `ResolveCheckRequest` to `dispatch`?

One other note, perhaps to do in another PR... The `clone` method of `ResolveCheckRequest` is fairly unnecessary. With a few tweaks, we could simply rely on value copy semantics for "cloning" the instance, which would be more performant.

internal/planner/benchmark_planner_test.go
```diff
@@ -0,0 +1,92 @@
...
+	b.StopTimer()
+
+	// Optional: Print the win rate to verify the algorithm is working
+	total := fastCount + slowCount
+	if total > 0 {
+		fmt.Printf("\n--- Benchmark Results ---\n")
+		fmt.Printf("Total Iterations: %d\n", total)
+		fmt.Printf("Default (Slow) Chosen: %d (%.2f%%)\n", slowCount, float64(slowCount)/float64(total)*100)
+		fmt.Printf("Recursive (Fast) Chosen: %d (%.2f%%)\n", fastCount, float64(fastCount)/float64(total)*100)
+		fmt.Printf("-------------------------\n")
```
adriantam: Please use [b.Logf](https://pkg.go.dev/testing#B.Logf) instead

internal/graph/check.go
```diff
@@ -365,16 +365,10 @@ func exclusion(ctx context.Context, _ int, handlers ...CheckHandlerFunc) (*Resol
 func (c *LocalChecker) Close() {
 }
 
-// dispatch clones the parent request, modifies its metadata and tupleKey, and dispatches the new request
-// to the CheckResolver this LocalChecker was constructed with.
-func (c *LocalChecker) dispatch(_ context.Context, parentReq *ResolveCheckRequest, tk *openfgav1.TupleKey) CheckHandlerFunc {
+// dispatch dispatches the given request to the CheckResolver this LocalChecker was constructed with.
+func (c *LocalChecker) dispatch(_ context.Context, req *ResolveCheckRequest) CheckHandlerFunc {
 	return func(ctx context.Context) (*ResolveCheckResponse, error) {
-		parentReq.GetRequestMetadata().DispatchCounter.Add(1)
```
senojj, line 372: was removing this intentional?

internal/graph/default_resolver.go
```diff
@@ -70,15 +76,25 @@ func (c *LocalChecker) defaultUserset(_ context.Context, req *ResolveCheckReques
...
 	}
 }
 
-func (c *LocalChecker) produceUsersetDispatches(ctx context.Context, req *ResolveCheckRequest, dispatches chan dispatchMsg, iter storage.TupleKeyIterator) {
+// prepareChildRequest creates a clone of the parent request and updates its fields
+// to create a child request for dispatching.
+func (c *LocalChecker) prepareChildRequest(parentReq *ResolveCheckRequest, tk *openfgav1.TupleKey, strategy string) *ResolveCheckRequest {
+	childRequest := parentReq.clone()
+	childRequest.TupleKey = tk
+	childRequest.GetRequestMetadata().Depth++
```
senojj: Is depth something that is not incremented on every dispatch? If it is, it should probably be the responsibility of the dispatch function to do. Else, ignore this comment.

internal/graph/check.go
yissellokta: yes, it was added in the default resolver before the go routine

internal/graph/default_resolver.go
yissellokta: It is incremented in every dispatch.

internal/graph/default_resolver.go
```diff
@@ -251,9 +266,27 @@ func (c *LocalChecker) processDispatches(ctx context.Context, limit int, dispatc
...
+						// Skip dispatching if parent request or tuple key is nil
+						concurrency.TrySendThroughChannel(
+							ctx,
+							checkOutcome{err: ErrPanic},
+							outcomes,
+						)
+						return
+					}
+					// Increment dispatch counter in the main thread before dispatching
+					msg.dispatchParams.parentReq.GetRequestMetadata().DispatchCounter.Add(1)
```
senojj, line 279: is there a reason that this incrementation would not be the responsibility of the dispatch function?

internal/graph/recursive_resolver.go
```diff
@@ -64,7 +64,7 @@ func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequ
 
 // recursiveTTU solves a union relation of the form "{operand1} OR ... {operandN} OR {recursive TTU}"
 // rightIter gives the iterator for the recursive TTU.
-func (c *LocalChecker) recursiveTTU(_ context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, rightIter storage.TupleKeyIterator) CheckHandlerFunc {
+func (c *LocalChecker) recursiveTTU(_ context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, rightIter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
```
adriantam: Q: Want to confirm whether we expect the selectedStrategy to be used.  If not, should we change it to

```
func (c *LocalChecker) recursiveTTU(_ context.Context, req *ResolveCheckRequest, rewrite *openfgav1.Userset, rightIter storage.TupleKeyIterator, _ string) CheckHandlerFunc {
```

internal/graph/recursive_resolver.go
```diff
@@ -48,7 +48,7 @@ type recursiveMapping struct {
 	allowedUserTypeRestrictions []*openfgav1.RelationReference
 }
 
-func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, _ []*openfgav1.RelationReference, rightIter storage.TupleKeyIterator) CheckHandlerFunc {
+func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, _ []*openfgav1.RelationReference, rightIter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
```
adriantam: Want to confirm whether we expect selectedStrategy to be used.  If not, should we change it to _

```
func (c *LocalChecker) recursiveUserset(_ context.Context, req *ResolveCheckRequest, _ []*openfgav1.RelationReference, rightIter storage.TupleKeyIterator, _ string) CheckHandlerFunc {
```

internal/graph/recursive_resolver_test.go
```diff
@@ -488,7 +488,7 @@ type group
 		checker := NewLocalChecker()
 		tupleKeys := []*openfgav1.TupleKey{{Object: "group:0", Relation: "parent", User: "group:1"}}
 
-		result, err := checker.recursiveTTU(ctx, req, typesystem.TupleToUserset("parent", "member"), storage.NewStaticTupleKeyIterator(tupleKeys))(ctx)
+		result, err := checker.recursiveTTU(ctx, req, typesystem.TupleToUserset("parent", "member"), storage.NewStaticTupleKeyIterator(tupleKeys), "recursive")(ctx)
```
adriantam, line 491: nitpick: should we use `recursiveResolver` instead of the string "recursive"?

internal/graph/weight_two_resolver.go
```diff
@@ -48,7 +48,7 @@ var ErrShortCircuit = errors.New("short circuit")
 
 type fastPathSetHandler func(context.Context, *iterator.Streams, chan<- *iterator.Msg)
 
-func (c *LocalChecker) weight2Userset(_ context.Context, req *ResolveCheckRequest, usersets []*openfgav1.RelationReference, iter storage.TupleKeyIterator) CheckHandlerFunc {
+func (c *LocalChecker) weight2Userset(_ context.Context, req *ResolveCheckRequest, usersets []*openfgav1.RelationReference, iter storage.TupleKeyIterator, selectedStrategy string) CheckHandlerFunc {
```
adriantam: again, do we expect selectedStrategy to be used in this function?

internal/graph/recursive_resolver_test.go
yissellokta: the strategy is called recursive not recursiveResolver, we are using the name of the strategy

PR #2903, opened 2026-01-29
fix: remove LOCK=NONE from MySQL migration 007 for Percona XtraDB Cluster compatibility

### LINKED ISSUE #2902: MySQL migration 007 fails on Percona XtraDB Cluster due to LOCK=NONE
### Checklist

When deploying OpenFGA with a MySQL-compatible backend (Percona XtraDB Cluster), migration  
`007_collate_object_id.sql` fails during `openfga migrate`.

This is reproducible with **OpenFGA 0.2.51 and 0.2.52**.

The migration contains:

```sql
CREATE INDEX idx_user_lookup
ON tuple (store, _user, relation, object_type, object_id)
LOCK = NONE;
```

On PXC, this fails with:

```text
ERROR 1845 (0A000): LOCK=NONE is not supported for this operation. Try LOCK=SHARED.
```

The `tuple` table uses InnoDB and MySQL-8 collations (`utf8mb4_0900_ai_ci`).  
Creating the same index without the `LOCK` clause succeeds.

This blocks fresh installs (e.g. Helm deployments with migrations enabled).

Remove `LOCK = NONE` from the migration (or downgrade to `LOCK = SHARED`) to improve compatibility with MySQL variants such as Percona XtraDB Cluster.

```sql
CREATE INDEX idx_user_lookup
ON tuple (store, _user, relation, object_type, object_id);
```

Migration 007 then needs to be manually marked as applied in `goose_db_version`.

### Additional context

- OpenFGA: 0.2.51, 0.2.52  
- Backend: Percona XtraDB Cluster (MySQL-compatible, InnoDB)

tylernix:
## Description

This PR removes `LOCK=NONE` clauses from MySQL migration `007_collate_object_id.sql` to improve compatibility with Percona XtraDB Cluster (PXC).

### Problem

When deploying OpenFGA with a MySQL-compatible backend using Percona XtraDB Cluster, migration 007 fails with:

```
ERROR 1845 (0A000): LOCK=NONE is not supported for this operation. Try LOCK=SHARED.
```

PXC has stricter requirements for DDL operations due to synchronous replication, and explicitly rejects `LOCK=NONE` for certain operations.

### Solution

Remove `LOCK=NONE` from `CREATE INDEX` and `DROP INDEX` statements. This is safe because:

1. **No behavior change for standard MySQL/InnoDB**: MySQL uses `LOCK=NONE` as the default for secondary index operations anyway, so omitting it results in the same online DDL behavior.

2. **Consistent with existing migrations**: Other OpenFGA MySQL migrations (e.g., `003_add_reverse_lookup_index.sql`) already create/drop indexes without specifying `LOCK` clauses.

3. **Improves compatibility**: Works correctly on standard MySQL, MariaDB, and Percona XtraDB Cluster.

### Changes

- Removed `LOCK = NONE` from `CREATE INDEX idx_user_lookup` (goose Up)
- Removed `LOCK = NONE` from `DROP INDEX idx_reverse_lookup_user` (goose Up)
- Removed `LOCK = NONE` from `DROP INDEX idx_user_lookup` (goose Down)
- Removed `LOCK = NONE` from `CREATE INDEX idx_reverse_lookup_user` (goose Down)

Note: `LOCK = SHARED` is retained for `ALTER TABLE` operations as this is supported by PXC.

Fixes #2902

## Summary by CodeRabbit

* **Chores**
  * Simplified database migration procedures by removing locking hints from index operations. This streamlines both index creation and removal across forward and rollback migration paths, potentially reducing deployment overhead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 ## [Unreleased]
 
+### Fixed
+- MySQL migration 007 now works with Percona XtraDB Cluster by removing unsupported `LOCK=NONE` clauses from index operations. [#2903](https://github.com/openfga/openfga/pull/2903)
+
 ## [1.11.3] - 2026-01-28
 ### Added
 - Add configuration option to limit max type system cache size. [2744](https://github.com/openfga/openfga/pull/2744)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
-
 ### Fixed
 - MySQL migration 007 now works with Percona XtraDB Cluster by removing unsupported `LOCK=NONE` clauses from index operations. [#2903](https://github.com/openfga/openfga/pull/2903)
 
--- a/assets/migrations/mysql/007_collate_object_id.sql
+++ b/assets/migrations/mysql/007_collate_object_id.sql
@@ -2,14 +2,14 @@
 ALTER TABLE tuple MODIFY COLUMN object_id VARCHAR(255) COLLATE utf8mb4_bin,
 LOCK = SHARED;
 
-CREATE INDEX idx_user_lookup ON tuple (store, _user, relation, object_type, object_id) LOCK = NONE;
+CREATE INDEX idx_user_lookup ON tuple (store, _user, relation, object_type, object_id);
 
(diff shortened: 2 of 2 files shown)


senojj:
Note that if we choose to always use a shared lock for index creation going forward, migrations that include new indexes will prevent writes to the table during index creation.

tylernix:
@senojj if no `LOCK` clause is included, what is the default for PXC? `LOCK=NONE` or `LOCK=SHARED`?

DanCech (approved):
This seems like a reasonable change, if we want to ensure that the database remains 100% available during migrations we should probably look into something like https://github.com/github/gh-ost (unfortunately MySQL-only).

senojj:
> This seems like a reasonable change, if we want to ensure that the database remains 100% available during migrations we should probably look into something like https://github.com/github/gh-ost (unfortunately MySQL-only).

@DanCech I agree with this, entirely. However, there have been internal conversations around avoiding any downtime, to the extent possible, for new migrations.

@tylernix the default is to use a shared lock. If we introduce this change, then I suggest that going forward, we hold a policy that migrations _may_ require _some_ amount of downtime.

Vic-Dev:
@tylernix do you have any response to @senojj's comment?

curfew-marathon:

jba-cloud:
Hello, can i ask for an quick update on this as we are experiencing the same issue.

Issue #1401, opened 2024-02-26
Flaky MySQL tests


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

https://github.com/openfga/openfga/actions/runs/8052034021/job/21991285655

```shell
go generate ./...
go test -race \
		-coverpkg=./... \
		-coverprofile=coverageunit.tmp.out \
		-covermode=atomic \
		-count=1 \
		-timeout=10m \
		./...
?   	github.com/openfga/openfga/assets	[no test files]
?   	github.com/openfga/openfga/cmd	[no test files]
?   	github.com/openfga/openfga/cmd/openfga	[no test files]
?   	github.com/openfga/openfga/cmd/util	[no test files]
?   	github.com/openfga/openfga/internal/authn	[no test files]
?   	github.com/openfga/openfga/internal/authn/oidc	[no test files]
?   	github.com/openfga/openfga/internal/authn/presharedkey	[no test files]
?   	github.com/openfga/openfga/internal/build	[no test files]
?   	github.com/openfga/openfga/internal/condition/metrics	[no test files]
?   	github.com/openfga/openfga/internal/middleware/authn	[no test files]
?   	github.com/openfga/openfga/internal/mocks	[no test files]
?   	github.com/openfga/openfga/internal/test/check	[no test files]
?   	github.com/openfga/openfga/internal/test/listobjects	[no test files]
ok  	github.com/openfga/openfga/cmd/migrate	42.036s	coverage: 17.7% of statements in ./...
?   	github.com/openfga/openfga/pkg/middleware/logging	[no test files]
ok  	github.com/openfga/openfga/cmd/run	40.641s	coverage: 4[5](https://github.com/openfga/openfga/actions/runs/8052034021/job/21991285655#step:4:6).3% of statements in ./...
ok  	github.com/openfga/openfga/cmd/validatemodels	31.314s	coverage: 22.9% of statements in ./...
ok  	github.com/openfga/openfga/internal/condition	1.127s	coverage: 55.9% of statements in ./...
ok  	github.com/openfga/openfga/internal/condition/eval	1.117s	coverage: 27.4% of statements in ./...
[...]

PR #2912, opened 2026-02-04
test: Limit connection idletime and lifetime to reduce flaky tests

rafanaskin:
Introduce a configuration for parallel tests with tuned DB connection `idletime` and `lifetime`.

Parallel tests using `DB` sometimes fail due to stale or closed connections in the connection pool. This results in [flakiness](https://github.com/openfga/openfga/issues/1401) and occasional `driver: bad connection` errors.

#### How is it being solved?
A dedicated test configuration is introduced with tuned connection pool settings. By limiting connection `idletime` and `lifetime`, the pool avoids keeping connections that may be killed by `DB`, ensuring stable execution even when tests run in parallel and under runner resource throttling.

Although the issue could not be reproduced, using a tuned connection pool is expected to positively impact test stability.

 - Introduced a configuration for parallel tests with tuned DB connection `idletime` and `lifetime`.
- Used this configuration in `check`, `listobjects` and `listusers` tests.
- Used constants for `config.Experimentals` instead of hard-coded values.

## References

closes https://github.com/openfga/openfga/issues/1401

## Review Checklist

## Release Notes

* **Chores**
  * Improved test infrastructure with optimized configuration handling for parallel testing scenarios and streamlined experimental feature flag management across test suites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Fixed
+- Reverted recent changes made to internal/planner/thompson.go that caused a regression in specific scenarios. [#2915](https://github.com/openfga/openfga/pull/2915)
 
 ## [1.11.3] - 2026-01-28
 ### Added
--- a/internal/planner/thompson.go
+++ b/internal/planner/thompson.go
@@ -34,22 +34,15 @@ func (ts *ThompsonStats) Sample(r *rand.Rand) float64 {
 
 	// Fast normal sampling
 	variance := 1.0 / (params.lambda * tau)
-	// Safety check for mathematical anomalies
-	if variance <= 0 || math.IsNaN(variance) {
-		return math.Exp(params.mu) // Return the geometric mean
+	if variance <= 0 {
+		return params.mu
 	}
 
 	// Use standard normal * sqrt(variance) + mean for better performance
 	stdNormal := r.NormFloat64()
-	logSample := params.mu + stdNormal*math.Sqrt(variance)
+	mean := params.mu + stdNormal*math.Sqrt(variance)
 
-	// Convert back to Linear Time (Milliseconds)
-	// exp(x) is guaranteed to be positive for all real x.
-	val := math.Exp(logSample)
-	if val < 1.0 {
-		return 1.0 // Enforce a minimum of 1ms to avoid pathological cases
-	}
-	return val
+	return mean
 }
 
 // fastGammaSample implements the highly efficient Marsaglia and Tsang acceptance-rejection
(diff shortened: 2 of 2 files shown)


senojj:
> By limiting connection idletime and lifetime, the pool avoids keeping connections that may be killed by DB, ensuring stable execution even when tests run in parallel and under runner resource throttling.

Considering that the connection pool, within its implementation, checks a connection's status before yielding it from the pool, what is the anticipated scenario that limiting the idle time and lifetime of the connections will solve? I'm only asking because I am not aware of one.

Generally speaking, unless the connection to the database is severed without even so much as a TCP RST packet, the connection pool is able to track the connection's status.

rafanaskin:
Hi @senojj,

Thanks for the review and question.

Initially, my main hypothesis was server-initiated idle cleanup via `wait_timeout`. I found an old MariaDB [issue](https://github.com/MariaDB/mariadb-docker/issues/113) where the default `wait_timeout` was unexpectedly low (`600s` instead of the documented `28800s`). I checked the MySQL images that could be used (8.0.32, 8.0.36) and they do have the expected default of `28800s`, so I ruled out a misconfigured default timeout as the primary cause.

After that, I tried to reproduce the issue directly. As you mentioned, the connection pool is quite smart and does track connection state. I explicitly set `wait_timeout` to 3 seconds, executed a query, waited for 3 seconds and then executed another query. While not deterministic, I did observe cases where a connection was taken from the pool and failed on the query with: `[4031] The client was disconnected by the server because of inactivity. See wait_timeout and interactive_timeout for configuring this behavior`.

This demonstrates that a connection can become invalid in the gap between pool validation (checkout) and the first query execution.

The intention behind limiting connection `idletime` and `lifetime` is not to eliminate failures entirely, but to reduce the probability of reusing long-lived connections. By proactively retiring connections, we narrow this reuse window and make connection behavior more predictable, which should improve test stability.

The remaining question is why this can happen during tests even with an 8h `wait_timeout`. My assumption is based on the test environment characteristics: CI is relatively unstable compared to local or production setups, 150+ tests run in parallel, the connection pool size is 30, individual tests are very short-lived.

Under these conditions, a connection can become bad during the increased gap between checkout and first query due to rare events, such as Docker network hiccups. These events are hard to reproduce locally but can surface as flaky failures in CI.

As additional context, this is also why creating new connections tends to be more resilient than reusing pooled ones in this scenario. When a new connection is created any failure typically happens during dial or handshake, before the connection is handed to the caller. In that case, `database/sql` can safely discard the connection and retry internally.

senojj:
@rafanaskin, thank you for taking the time to walk me through your analysis. I think that your observation and conclusion of the connection being closed by the database, between being yielded from the pool, and the client issuing a query, is a likely culprit of the flaky connection errors in MySQL tests.

Due to the difficulty of determining the exact cause of this phenomenon in a CI environment, I'm inclined to agree that shortening the lifetime of connections in the pool is a reasonable mitigation tactic.

senojj:
On a related note, a PostgreSQL test just failed on this branch with error: `sql error: failed to deallocate cached statement(s): timeout: read tcp 127.0.0.1:50454->127.0.0.1:32818: i/o timeout`.

I don't think it is related to these changes, but wanted to make a note of it.

rafanaskin:
> On a related note, a PostgreSQL test just failed on this branch with error: `sql error: failed to deallocate cached statement(s): timeout: read tcp 127.0.0.1:50454->127.0.0.1:32818: i/o timeout`.
> 
> I don't think it is related to these changes, but wanted to make a note of it.

I've looked into this and I agree that it's unlikely the issue is related to changes in `idletime` or `lifetime`.

One possible mitigation for the cached statement problem is to set `default_query_exec_mode = describe_exec`. According to the [documentation](https://github.com/jackc/pgx/blob/master/conn.go#L665), this mode does not use cached prepared statements, which eliminates the issue entirely. I tested this locally and all tests pass with this configuration.
Would you prefer that I open a separate issue to track this change?

During the investigation, I also noticed that we recreate database containers very frequently (10+ times during a test run). I experimented with reusing the same container and compared test execution time in `github.com/openfga/openfga/tests`:
- New container per run: `DONE 31931 tests, 832 skipped in 114.205s`
- Reused container: `DONE 31931 tests, 832 skipped in 74.182s`

Reusing containers reduces execution time and also lowers resource pressure in CI. In my opinion, the frequent container recreation is a likely root cause of the test instability we're seeing.

@senojj - what do you think?
Would it make sense for me to implement container reuse for both `MySQL` and `PostgreSQL` tests?

senojj:
> I've looked into this and I agree that it's unlikely the issue is related to changes in `idletime` or `lifetime`.
> 
> One possible mitigation for the cached statement problem is to set `default_query_exec_mode = describe_exec`. According to the [documentation](https://github.com/jackc/pgx/blob/master/conn.go#L665), this mode does not use cached prepared statements, which eliminates the issue entirely. I tested this locally and all tests pass with this configuration. Would you prefer that I open a separate issue to track this change?

My concern is less about the cached statement deallocation, and more about the root cause of the error: `read tcp 127.0.0.1:50454->127.0.0.1:32818: i/o timeout`. Without knowing exactly why there was an i/o timeout, there isn't much we can do as a solution. Since the statement cache deallocation was only a side effect of the i/o timeout, I don't believe there is anything we should change with it at this time.

> what do you think? Would it make sense for me to implement container reuse for both `MySQL` and `PostgreSQL` tests?

Reusing test containers, to the extent possible, would be a great enhancement. Perhaps a separate PR for that change, since it will likely be large. That would enable us to continue to push these current changes forward; I don't think these changes will be altered by reuse of the test containers.

rafanaskin:
> Without knowing exactly why there was an i/o timeout, there isn't much we can do as a solution.

Agreed.

> Reusing test containers, to the extent possible, would be a great enhancement.

Thanks - I'll take this on in a separate PR.

Given that, are we ready to merge this PR?

PR #2914, opened 2026-02-04
Read PGPASSFILE in BeforeConnect hook every time

### LINKED ISSUE #2913: PGPASS Regression in 1.11.X with pgxPool change
### Checklist

### Description

hey folks, we discovered an issue with PGXPool used in 1.11.x of OpenFGA that is preventing us from upgrading, I'm hoping to get some feedback on a fix.

* We use the PGPASS env var ([understood by the pgx library under the hood](https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L21)) for DB authentication
  * For us, this value is dynamic and changes every 15 minutes - it is generated AWS IAM
* In versions < 1.11, as this file is updated in-place, new connections use the newest password
  * [code location here](https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414) and [here](https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/conn.go#L135-L151)
* in contrast, pgxpool only retrieves config [once here](https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgxpool/pool.go#L212-L219) upon pool initialization.
  * this is what is causing our OpenFGA instances to error after 15 minutes of uptime.

The most immediate solution I can think of (that doesn't involve patching pgx itself) is to set the [BeforeConnect](https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgxpool/pool.go#L125-L127) hook to update the config based on PGPASS changing.

### OpenFGA version

1.10.3 upgradign to 1.11.3

As a binary

Postgres

### OpenFGA Flags

"OPENFGA_CHECK_QUERY_CACHE_ENABLED"="true"
"OPENFGA_CHECK_ITERATOR_CACHE_ENABLED"="true"
"OPENFGA_CACHE_CONTROLLER_ENABLED"="true"
"OPENFGA_SHARED_ITERATOR_ENABLED"="true"
"OPENFGA_CHECK_CACHE_LIMIT"="150000"
"OPENFGA_CHECK_ITERATOR_CACHE_MAX_RESULTS"="10000"
"OPENFGA_CHECK_QUERY_CACHE_TTL"="10m"
"OPENFGA_CHECK_ITERATOR_CACHE_TTL"="10m"
"OPENFGA_CACHE_CONTROLLER_TTL"="1m"
"OPENFGA_REQUEST_DURATION_DATASTORE_QUERY_COUNT_BUCKETS"="50,200"
"OPENFGA_AUTHN_OIDC_ISSUER"="redacted"
"OPENFGA_TRACE_ENABLED"="true"
"OPENFGA_LOG_FORMAT"="json"
"OPENFGA_DATASTORE_ENGINE"="postgres"
"OPENFGA_DATASTORE_URI"="postgres://redacted@redacted:5432/openfga"
"PROXY_ENDPOINT"="redacted"
"OPENFGA_DATASTORE_MAX_OPEN_CONNS"="1000"
"OPENFGA_DATASTORE_MIN_OPEN_CONNS"="150"
"OPENFGA_DATASTORE_MIN_IDLE_CONNS"="50"
"OPENFGA_DATASTORE_CONN_MAX_LIFETIME"="10m"
"OPENFGA_DATASTORE_CONN_MAX_IDLE_TIME"="5m"
"OPENFGA_PLAYGROUND_ENABLED"="false"
"OPENFGA_REQUEST_DURATION_DISPATCH_COUNT_BUCKETS"="50,200"
"OPENFGA_METRICS_ENABLE_RPC_HISTOGRAMS"="true"
"OPENFGA_HTTP_ADDR"="0.0.0.0:8000"
"OPENFGA_AUTHN_METHOD"="oidc"
"OPENFGA_TRACE_SAMPLE_RATIO"="0.3"
"OPENFGA_TRACE_OTLP_TLS_ENABLED"="false"
"OPENFGA_DATASTORE_METRICS_ENABLED"="true"
"OPENFGA_TRACE_OTLP_ENDPOINT"="redacted"
"OPENFGA_LOG_LEVEL"="info"
"OPENFGA_AUTHN_OIDC_SUBJECTS"="redacted"
"OPENFGA_AUTHN_OIDC_AUDIENCE"="redacted"
"OPENFGA_MAX_TUPLES_PER_WRITE"="400"
"PGPASSFILE"=".pgpass"

### Logs

RDS Proxy Log statement:
// Service deploys at 16:04
> 2026-02-04T16:06:06.284Z [INFO] [proxyEndpoint=default] [clientConnection=613066239] A new client connected from 10.8.72.81:51478.

// Service uses credentials which expired at 16:19 - 9 minutes after expiry
> 2026-02-04T16:28:06.998Z [INFO] [proxyEndpoint=default] [clientConnection=2985566783] Proxy authentication with IAM authentication failed for user "redacted" with TLS on. Reason: The proxy couldn't authenticate using IAM. The authentication token has expired. Generate a new authentication token and use it in a new client connection.

OpenFGA logs

> "error": {
                    "internal_error": "internal_error",
                    "message": "sql error: sql error: failed to connect to `user=redacted database=redacted`: 10.8.85.144:5432 (iredacted): server error: FATAL: The IAM authentication failed for the role redacted. Check the IAM token for this role and try again. (SQLSTATE 28P01)"
                }

JeremyLoy:

regression with 1.11.x - PGPASS updates are not being picked up with new conns

#### How is it being solved?

setting up the pgxpool.BeforeHook to read PGPASS updates

## References

fixes #2913

## Review Checklist

## Release Notes

* **New Features**
  * PostgreSQL connections now automatically authenticate using .pgpass files. When a pgpass file is present, credentials are securely loaded during connection initialization, enabling password management without exposing credentials in connection strings.

### DIFF
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/IBM/pgxpoolprometheus"
@@ -108,7 +107,7 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
 		c.MaxConnIdleTime = cfg.ConnMaxIdleTime
 	}
 
-	c.BeforeConnect = createBeforeConnect(cfg.Logger, &FSPassfileProvider{cfg.Logger, os.UserHomeDir, FileOpener, sync.Once{}})
+	c.BeforeConnect = createBeforeConnect(cfg.Logger, &FSPassfileProvider{cfg.Logger, os.UserHomeDir, FileOpener})
 	return c, nil
 }
 
@@ -133,7 +132,6 @@ type FSPassfileProvider struct {
 	Logger     logger.Logger
 	GetHomeDir func() (string, error)
 	OpenFile   func(name string) (FileStat, error)
-	LogOnce    sync.Once
 }
 
 func (p *FSPassfileProvider) OpenPassfile() (io.ReadCloser, error) {
@@ -141,9 +139,7 @@ func (p *FSPassfileProvider) OpenPassfile() (io.ReadCloser, error) {
 	if len(fileLocation) == 0 {
 		homeDir, err := p.GetHomeDir()
 		if err != nil {
-			p.LogOnce.Do(func() {
-				p.Logger.Info("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
-			})
+			p.Logger.Debug("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
 			return nil, ErrNoPassfile
 		}
 		fileLocation = filepath.Join(homeDir, ".pgpass")
@@ -180,7 +176,6 @@ func (p *FSPassfileProvider) OpenPassfile() (io.ReadCloser, error) {
 //
 // https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414
 func createBeforeConnect(logger logger.Logger, provider PassfileProvider) func(ctx context.Context, conn *pgx.ConnConfig) error {
-	var logOnce sync.Once
 	return func(ctx context.Context, config *pgx.ConnConfig) error {
 		file, err := provider.OpenPassfile()
 		// Like with libpq - ignore the pgpass file if the permissions are too permissive
@@ -203,9 +198,7 @@ func createBeforeConnect(logger logger.Logger, provider PassfileProvider) func(c
 
 		password := passfile.FindPassword(host, strconv.Itoa(int(config.Port)), config.Database, config.User)
 		if password == "" {
-			logOnce.Do(func() {
-				logger.Warn("no password found in pgpassfile for user/host/database combination")
-			})
+			logger.Debug("no password found in pgpassfile for user/host/database combination")
 		} else {
 			config.Password = password
 		}


adriantam:
Hello @JeremyLoy , thanks for filing the issue and create a patch.  Want to step back a bit to understand how you are running the OpenFGA.  Are you running OpenFGA as a library or as a standalone executable?  The reason I am asking this is wondering whether we can put the BeforeConnect code outside of OpenFGA and then we pass this in as a function pointer during the pgx initialization.  In this way, we can keep the customized code out of the main project.  My concern is that this use case is very specific and some other operator of OpenFGA may expect a different (i.e., pgxpool default) behaviour.  In addition, I am not a big fan of copying nearly verbatim from pgx code as it poses maintainability risk when pgxpool is updated.  Let me know what you think.  Thanks!

JeremyLoy:
Thanks for the response @adriantam!

We are running OpenFGA as an executable using the published docker image.

My intent with this PR was to preserve the pre-1.11.x behavior that we depended on, but I can understand not wanting to change the new defaults of pgxpool.

I’ll keep tinkering with this and try to pass the function pointers down do OpenFGA when used as a library. For us it would be more work, as would now maintain a (small) Golang wrapper program to inject the function. At that point we could probably nix PGPASS ourself and use AWS IAM in the BeforeConnect.

I can change this PR to go down that path - want me to forward the other hooks while I’m at it? And where do you suggest making these changes? Somewhere in `cmd`?

JeremyLoy:
@adriantam I updated this to pass the hook in from `cmd/run`. This is not suitable for merging at present and lacks tests, but I wanted to get feedback on some design obstacles first 

* the BeforeConnect function signature is inherently pgx specific. I think thats fine because `sqlcommon.Config` already has a few pgx specific options like `WithMinIdleConns`.
* There are more pgx hooks and options that could be passed through. I think passing a `struct` down would be more sustainable, and would like feedback on this - what to call it, what to have as a part of this initial PR etc

As a consumer, since we are no longer running the docker image / published binary, there are a few ergonomic hurdles
* we need to copy `cmd/openfga/main.go` and keep this in-sync
  * this is the only way to maintain the CLI options and flags  
* the types returned by everything in `cmd` are all `cobra` commands, and are pretty impenetrable
  * this is where we could improve the customization experience - have these functions return `struct`s that contain the `cobra` commands, but still allow manipulation of configuration and flags
  
  ---
  
  Also, I don't seem to have the permissions to update my PR title, so if you could do so that would be appreciated!

senojj:
I would not recommend exposing the `BeforeConnect` hook outside of the `postgres` package as it is specific to `pgxpool`. 

Encapsulated within the postgres adapter, dynamic reloading of the .pgpass file should be implemented using the `BeforeConnect` hook.

We can use the pgx related [libarary](https://github.com/jackc/pgpassfile) for parsing a .pgpass file, or avoid another dependency and just implement the simple logic ourselves.

I don't believe this to be a niche use-case and warrants a built-in solution.

JeremyLoy:
@senojj thanks for the review!

An earlier iteration of this PR did exactly that, I will bring that back right now so you can take a look.

An aside - the push to expose the hooks would allow adopters to write custom hooks (like getting the password from AWS IAM directly).

I'm fine with either approach, I just need to know which direction to take the PR. For now, I'll revert to what it was before

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,52 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+// error should fail acquiring the connection
+//
+// https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414
+func createBeforeConnect(cfg *sqlcommon.Config) func(ctx context.Context, conn *pgx.ConnConfig) error {
+	osUser, err := user.Current()
+	if err != nil {
+		cfg.Logger.Info("could not get current os user, skipping pgxpool BeforeConnect hook creation", zap.Error(err))
+		return nil
+	}
+	passfileName := cmp.Or(os.Getenv("PGPASSFILE"), filepath.Join(osUser.HomeDir, ".pgpass"))
```
senojj: Check for presence of "PGPASSFILE" before attempting to get the user directory. Failing to get the user's home directory should not short the env var path.

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,52 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+// Fixes issue 2913 where pgxpool doesn't fetch updates to PGPASSFILE upon every connection attempt.
+// This is a behavior change compared to plain pgx conn acquisition and previous versions of OpenFGA.
+//
+// The implementation is copied nearly verbatim from pgx code, but handles errors differently.
+// If a PGPASSFILE exists at program start, it is assumed you intend to read from it constantly. Therefore a read
+// error should fail acquiring the connection
+//
+// https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414
+func createBeforeConnect(cfg *sqlcommon.Config) func(ctx context.Context, conn *pgx.ConnConfig) error {
+	osUser, err := user.Current()
```
senojj: use `os.UserHomeDir()` instead as it is the most portable approach.

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,52 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+		cfg.Logger.Info("could not get current os user, skipping pgxpool BeforeConnect hook creation", zap.Error(err))
+		return nil
+	}
+	passfileName := cmp.Or(os.Getenv("PGPASSFILE"), filepath.Join(osUser.HomeDir, ".pgpass"))
+	_, err = os.Stat(passfileName)
+	if err != nil {
+		cfg.Logger.Info("no pgpassfile found, skipping pgxpool BeforeConnect hook creation", zap.Error(err))
+		return nil
+	}
+	// At this point, the passfile exists. Any failure to read from it should be considered an error in acquiring the connection
```
senojj: The file may not exist at the same directory or name upon a subsequent invocation. It is possible that the env var value has been updated.

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,52 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+	return func(ctx context.Context, config *pgx.ConnConfig) error {
+		passfile, err := pgpassfile.ReadPassfile(passfileName)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("pgxpool BeforeConnect hook - failed to read passfile: %w", err)
+		}
+		host := config.Host
+		if network, _ := pgconn.NetworkAddress(config.Host, config.Port); network == "unix" {
```
senojj, line 195: Why set `host` to "localhost" when a unix domain socket?

senojj: Nevermind, I see the comment in the library.

JeremyLoy:
@senojj I addressed all of the feedback, let me know if you have any more items!

image


Here is a snippet from our staging env where I have this fork deployed. We are rotating the PGPASSFILE every 10 min and it looks to be picking up the new settings correctly

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +105,59 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+// error should fail acquiring the connection
+//
+// https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414
+func createBeforeConnect(logger logger.Logger, getFileName func(logger.Logger) string) func(ctx context.Context, conn *pgx.ConnConfig) error {
+	pgpassFileName := getFileName(logger)
+	_, err := os.Stat(pgpassFileName)
+	if err != nil {
+		logger.Info("no pgpassfile found, skipping pgxpool BeforeConnect hook creation", zap.Error(err))
+		return nil
+	}
```
senojj: It doesn't exist now, but may in the future. Any reason to avoid looking for it when new connections are created?

JeremyLoy: yep I can make that change. So just always set the hook no matter what.

If getting the homedir fails, then what should the behavior be? In pgxconn its a silent error. We could at least log it.

https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/defaults.go#L18-L39

pkg/storage/postgres/postgres_test.go
```diff
@@ -222,17 +233,54 @@ func TestParseConfig(t *testing.T) {
...
+				MaxConnLifetime:       defaultConfig.MaxConnLifetime,
+			},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			if tt.expectedPgpassHook {
+				pgPassPath := filepath.Join(t.TempDir(), ".pgpass")
+				t.Setenv("PGPASSFILE", pgPassPath)
+				if err := os.WriteFile(pgPassPath, []byte("*:*:*:*:secondpassword"), 0o600); err != nil {
```
senojj: It's not great that this test requires writing to a filesystem. Perhaps it is possible to make the logic more "unit testable" through abstraction?

As a suggestion, add a provider interface that returns an `io.Reader` and use `pgpassfile.ParsePassfile` instead of `pgpassfile.ReadPassfile`.

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,59 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+// The implementation is copied nearly verbatim from pgx code, but handles errors differently.
+//
+// https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414
+func createBeforeConnect(logger logger.Logger, getFileName func(logger.Logger) string, getFile func(string) (io.Reader, error)) func(ctx context.Context, conn *pgx.ConnConfig) error {
+	return func(ctx context.Context, config *pgx.ConnConfig) error {
+		pgpassFileName := getFileName(logger)
+		file, err := getFile(pgpassFileName)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("pgxpool BeforeConnect hook - failed to read passfile: %w", err)
+		}
```
DanCech, line 188: we need to handle ErrNotExists here

JeremyLoy: thanks...that one got lost in a previous iteration

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,62 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
 	return c, nil
 }
 
+// Fixes issue 2913 where pgxpool doesn't fetch updates to PGPASSFILE upon every connection attempt.
+// This is a behavior change compared to plain pgx conn acquisition and previous versions of OpenFGA.
+//
+// The implementation is copied nearly verbatim from pgx code, but handles errors differently.
+//
+// https://github.com/jackc/pgx/blob/f56ca73076f3fc935a2a049cf78993bfcbba8f68/pgconn/config.go#L404-L414
+func createBeforeConnect(logger logger.Logger, getFileName func(logger.Logger) string, getFile func(string) (io.Reader, error)) func(ctx context.Context, conn *pgx.ConnConfig) error {
```
senojj: Instead of decoupling the file name fetching and the `io.Reader` producing... why not combine them behind an interface? This allows the implementation more flexibility.
```go
var ErrNoPassfile = errors.New("passfile does not exist")

type PassfileProvider interface {
    OpenPassfile() (io.Reader, error)
}

type FSPassfileProvider struct {
    Logger logger.Logger
}

func (p *FSPassfileProvider) OpenPassfile() (io.Reader, error) {
    homeDir, err := os.UserHomeDir()
	if err != nil {
		logger.Info("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
	} else {
		homeDir = filepath.Join(homeDir, ".pgpass")
	}
	fileLocation = cmp.Or(os.Getenv("PGPASSFILE"), homeDir)
    f, err := os.Open(fileLocation)
    if errors.Is(err, os.ErrNotExist) {
        return nil, ErrNoPassfile
    }
    return f, err
}

type MemoryPassfileProvider struct {
    Content string
}

func (p *MemoryPassfileProvider) OpenPassfile() (io.Reader, error) {
    return strings.NewReader(p.Content), nil
}

type MissingPassfileProvider struct {}

func (p *MissingPassfileProvider) OpenPassfile() (io.Reader, error) {
    return nil, ErrNoPassfile
}
```
Then your function becomes:
```go
func createBeforeConnect(provider PassfileProvider) func(context.Context, *pgx.ConnConfig) error { ... }
```

JeremyLoy: thanks, its been a few years since I've written some go!

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,71 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+}
+
+func (p *FSPassfileProvider) OpenPassfile() (io.Reader, error) {
+	homeDir, err := p.GetHomeDir()
+	if err != nil {
+		p.Logger.Info("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
+	} else {
+		homeDir = filepath.Join(homeDir, ".pgpass")
+	}
+	fileLocation := cmp.Or(os.Getenv("PGPASSFILE"), homeDir)
```
DanCech: The way this is structured currently is a bit hard to follow, something like this may make the logic clearer and avoid the current situation where we'll try to open `""` if the environment variable is not set and we can't get the user's home dir (as well as always trying to get the homedir even if the env var is set):
```suggestion
	fileLocation := os.Getenv("PGPASSFILE")
	if len(fileLocation) == 0 {
		homeDir, err := p.GetHomeDir()
		if err != nil {
			p.Logger.Info("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
			return nil, fmt.Errorf("could not get current user home directory")
		}
		fileLocation = filepath.Join(homeDir, ".pgpass")
	}
```

JeremyLoy: this suggestion has a behavior change where `p.GetHomeDir()` now results in an error different than `ErrNoPassfile`, so I'm going to modify it slightly. I don't think if `os.GetHomeDir()` fails (which from what I understand can happen in some containerized environments) the function should fail

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Changed
+- PGPASSFILE is read upon every connection attempt if it exists instead a single time upon application start. This restores <v1.11.x behavior. [#2913](https://github.com/openfga/openfga/pull/2913)
```
DanCech:
```suggestion
- If PGPASSFILE exists it is read upon every connection attempt instead of a single time upon application start. This restores <v1.11.x behavior. [#2913](https://github.com/openfga/openfga/pull/2913)
```

pkg/storage/postgres/postgres_test.go
```diff
@@ -250,11 +314,148 @@ func TestParseConfig(t *testing.T) {
...
+	t.Run("the hook returns an error if reading the file fails", func(t *testing.T) {
+		provider := &ReaderPassfileProvider{iotest.ErrReader(errors.New("read error")), nil}
+		hook := createBeforeConnect(noOpLogger, provider)
+		ctx := context.Background()
+		config := new(pgx.ConnConfig)
+		err := hook(ctx, config)
+		require.Error(t, err, "read error")
+		require.Empty(t, config.Password)
+	})
+	t.Run("the retrieves localhost entries for unix socket hosts", func(t *testing.T) {
```
DanCech:
```suggestion
	t.Run("retrieves localhost entries for unix socket hosts", func(t *testing.T) {
```

pkg/storage/postgres/postgres_test.go
```diff
@@ -222,14 +257,43 @@ func TestParseConfig(t *testing.T) {
...
+			name: "bad_uri",
+			uri:  "bad_uri",
+			cfg: sqlcommon.Config{
+				Logger: logger.NewNoopLogger(),
+			},
 			override:    true,
 			expectedErr: true,
 		},
+		{
+			name:     "reads password from PGPASS",
```
DanCech:
```suggestion
			name:     "reads password from PGPASSFILE",
```

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +105,73 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+		if err != nil {
+			p.Logger.Info("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
+			return nil, ErrNoPassfile
+		}
+		fileLocation = filepath.Join(homeDir, ".pgpass")
+	}
+	f, err := os.Open(fileLocation)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, ErrNoPassfile
+	}
```
senojj: Reading the [official documentation](https://www.postgresql.org/docs/current/libpq-pgpass.html), libpq wisely only reads and uses the .pgpass file's contents when the file's permissions are equal to, or more strict than 0600 on Unix-like systems.

We should follow this practice as well.

senojj: I realize that this behavior is not equivalent to what was originally there, so we can continue with this PR without that addition. However, I wanted to note this here for follow-up.

JeremyLoy: I don't mind adding it now - the implementation has already drifted significantly farther than just a simple port of the pxconn implementation.

It's been a hot minute since I've had to modify unix perm flags - would this be as simple as `perm &^ 0o600`?

JeremyLoy: also do we not check this on Windows (just like libpq)?

pkg/storage/postgres/postgres_test.go
```diff
@@ -250,11 +305,148 @@ func TestParseConfig(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Equal(t, "password", config.Password)
+	})
+}
+
+func TestFSPassfileProvider_OpenPassfile(t *testing.T) {
+	t.Run("uses home dir without PGPASSFILE env var set", func(t *testing.T) {
+		dir := t.TempDir()
+		filename := filepath.Join(dir, ".pgpass")
+		_, err := os.Create(filename)
```
senojj: The non-determinism of filesystem i/o could be avoided by accepting an `OpenFile func(string) (io.Reader, error)` into `FSPassFileProvider` as a field. (much like what you did for `GetHomeDir`).

JeremyLoy: This is a test for the FSPassfile though...feels weird to not test using the file system in the FileSystemProvider right?

Especially with the earlier ask to incorporate file perm check, I'm not sure how this would shake out.

pkg/storage/postgres/postgres.go
JeremyLoy: added

pkg/storage/postgres/postgres_test.go
JeremyLoy: https://pkg.go.dev/testing/fstest#MapFS

I could rewrite this to use `fs` instead - and `fstest` has an in-memory FS implementation. What do you think about that?

JeremyLoy: agh, I don't think that would work. $HOME/.pgpass is almost certainly going to be in a different directory than whatever PGPASSFILE points too, and we really only care about one file

JeremyLoy: OK I took a stab at it. I made an interface around the `*os.File` methods we actually use. It's a bit verbose on the testing side but it does get rid of the non-determinism

JeremyLoy: OK I took a stab at it. I made an interface around the *os.File methods we actually use. It's a bit verbose on the testing side but it does get rid of the non-determinism

JeremyLoy:
looks to be a bit of flakiness in CI related to #2912. Let me know what else you all need! Thanks for the feedback as I get back up-to-speed with go

JeremyLoy:
image

seems stable now with #2912 merged in.

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,106 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+	GetHomeDir func() (string, error)
+	OpenFile   func(name string) (FileStat, error)
+}
+
+func (p *FSPassfileProvider) OpenPassfile() (io.ReadCloser, error) {
+	fileLocation := os.Getenv("PGPASSFILE")
+	if len(fileLocation) == 0 {
+		homeDir, err := p.GetHomeDir()
+		if err != nil {
+			p.Logger.Info("could not get current user home directory for pgxpool BeforeHook creation", zap.Error(err))
```
senojj: This log could get noisy in the event that the user is not using a pgpass and the application is running as a user without a home directory within a *nix container.

pkg/storage/postgres/postgres.go
```diff
@@ -99,9 +106,106 @@ func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Con
...
+			return fmt.Errorf("pgxpool BeforeConnect hook - failed to parse passfile: %w", err)
+		}
+		host := config.Host
+		if network, _ := pgconn.NetworkAddress(config.Host, config.Port); network == "unix" {
+			host = "localhost"
+		}
+
+		password := passfile.FindPassword(host, strconv.Itoa(int(config.Port)), config.Database, config.User)
+		if password == "" {
+			logger.Warn("pgpassfile password was empty")
```
senojj: Could this be an intended configuration? As in, use the password from the pgpass file when present, otherwise use currently configured password.

JeremyLoy: I think so, if I understand your question correctly.

it could very well be the case that a valid file exists but no password is found for the current user/host combo. A single file can contain multiple rows.

in that case, I guess the log statement could use some clarification - it’s really “password not found for current host/user combination”.

pkg/storage/postgres/postgres.go
JeremyLoy: I agree, but if possible I’d like to leave it at least once just for debugging.

wrap in a sync.Once?

pkg/storage/postgres/postgres.go
senojj: That could be an intended behavior though. Meaning that I provide a default password in my connection string, but the application should use the password from the pgpass file, if it exists and contains a relevant password.

If this is the intention of the user, seeing many, many of these logs would be less than desirable.

JeremyLoy: It’s wrapped in a sync.Once. Let me know if you want it out entirely

pkg/storage/postgres/postgres.go
senojj: It would be fine as a debug log.

pkg/storage/postgres/postgres.go
senojj: This could be made into a debug log.

senojj: Wrapping logging points in a sync.Once is not ideal.

JeremyLoy: fair, done.

JeremyLoy:
Thank you @senojj @DanCech and @adriantam  for all of the feedback! 🚀

PR #2921, opened 2026-02-10
feat: enable pipeline algorithm by default

Vic-Dev:

Enabling pipeline as the default ListObjects algorithm. Can be disabled by setting the `--listObjects-pipeline-enabled` flag to `false`.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * ListObjects pipeline is now enabled by default and no longer experimental.

* **Configuration**
  * New `listObjectsPipelineEnabled` setting (default: true) controls pipeline behavior.
  * Disable via flag `--list-objects-pipeline-enabled=false` or environment variables `OPENFGA_LIST_OBJECTS_PIPELINE_ENABLED`/`OPENFGA_LISTOBJECTSPIPELINEENABLED`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,439 +2,148 @@
 
 This file provides guidance to AI coding agents when working with code in this repository.
 
-## Project Goals and Non-Goals
+## Project Overview
 
-### Goals
+OpenFGA is a high-performance ReBAC (Relationship-Based Access Control) authorization engine inspired by Google Zanzibar. It evaluates graph-based authorization queries using set operations (union, intersection, exclusion).
 
-- **ReBAC authorization engine** inspired by Google Zanzibar with graph evaluation (union, intersection, exclusion)
-- **Low-latency checks** (sub-millisecond to single-digit ms) with minimal memory allocations
-- **Bounded concurrency** with breadth/depth limits, dispatch throttling, multi-layer caching
-- **Deployment flexibility**: Service mode (HTTP/gRPC APIs) or library mode (embeddable Go library)
-- **SQL storage**: PostgreSQL (production), MySQL, SQLite (beta) with ACID properties
-- **Production-grade observability**: Prometheus metrics, OpenTelemetry tracing, structured logging
-- **Developer experience**: Playground UI, SDKs (Java, Node.js, Go, Python, .NET), YAML test matrices
-
-### Non-Goals
-
-**Storage Backends:**
-- NoSQL databases (MongoDB, DynamoDB, Cassandra, Redis as primary storage)
-
-**Identity and Authentication:**
-- Identity provider functionality (user creation, password management, user profiles)
-- User authentication and credential validation
-- LDAP or SAML integration for user authentication
-
-**Authorization Paradigms:**
-- Pure Attribute-Based Access Control (ABAC) without relationships
-
-**API Protocols:**
-- GraphQL API
-
-**Audit and Compliance:**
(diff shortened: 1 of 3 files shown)


senojj:
#2926 should unblock this work.

internal/test/listobjects/listobjects.go
```diff
@@ -7,9 +7,10 @@ import (
...
-	ContextualTuples []*openfgav1.TupleKey `json:"contextualTuples"`
-	Context          *structpb.Struct
-	Expectation      []string
-	ErrorCode        int `json:"errorCode"` // If ErrorCode is non-zero then we expect that the ListObjects call failed.
+	Request                  *openfgav1.ListObjectsRequest
+	ContextualTuples         []*openfgav1.TupleKey `json:"contextualTuples"`
+	Context                  *structpb.Struct
+	Expectation              []string
+	ErrorCode                int `json:"errorCode"`                // If ErrorCode is non-zero then we expect that the ListObjects call failed.
+	WithoutPipelineErrorCode int `json:"withoutPipelineErrorCode"` // If WithoutPipelineErrorCode is non-zero then we expect that the ListObjects call failed when pipeline is disabled.
```
senojj: Why is this necessary? Any reason that we cannot use the `ErrorCode` property and simply check if the pipeline was enabled or not to determine the context of the error?

Vic-Dev: This is about cases where when the pipeline is disabled the existing code would error but the pipeline code would _not_ error, for example with max depth:
https://github.com/openfga/openfga/pull/2921/changes#diff-1d77257dd9e1a0e76a719b659c451b109c9c9cfb2247b56f9f20b7bd9b8d5ceeR7722

senojj: Instead of asserting this:
```go
case !params.pipelineEnabled:
    require.Error(t, err, detailedInfo)
	require.True(t, ok, detailedInfo)
	require.Equal(t, assertion.WithoutPipelineErrorCode, int(e.Code()), detailedInfo)
```
You could just do this, right?
```go
case !params.pipelineEnabled:
	require.Error(t, err, detailedInfo)
	require.True(t, ok, detailedInfo)
	require.Equal(t, assertion.ErrorCode, int(e.Code()), detailedInfo)
```

senojj: I misunderstood.

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -7719,7 +7719,9 @@ tests:
               type: resource
               relation: can_view
               user: user:maria
-            errorCode: 2002
+            withoutPipelineErrorCode: 2002
+            expectation:
+              - resource:1
```
senojj: since this is behavior that we no longer want to support in ListObjects, let's just remove this assertion.

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -8306,7 +8308,9 @@ tests:
...
```
senojj: since this is behavior that we no longer want to support in ListObjects, let's just remove this assertion.

assets/tests/consolidated_1_1_tests.yaml
```diff
@@ -8436,7 +8440,9 @@ tests:
...
```
senojj: since this is behavior that we no longer want to support in ListObjects, let's just remove this assertion.

PR #2936, opened 2026-02-20
refine AGENTS.md to make it more concise with higher value.

senojj:
Currently bloated AGENTS.md file.

#### How is it being solved?
This is an attempt to refine it and increase its value by providing only information that cannot be inferred from the project files or would be required frequently as context.

## Review Checklist

## Summary by CodeRabbit

* **Documentation**
  * Reorganized and condensed project documentation for improved clarity and navigation.
  * Simplified development commands section with concise make targets and examples.
  * Restructured architecture overview with better component descriptions and resolver pattern documentation.
  * Enhanced contributor workflow references and quick-start guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Changed
+- Migrate `grpc.DialContext` to `grpc.NewClient` for grpc-gateway client [#2714](https://github.com/openfga/openfga/pull/2714)
+
 ### Security
 - Bump [`grpc-health-probe`](https://github.com/grpc-ecosystem/grpc-health-probe) to [v0.4.45](https://github.com/grpc-ecosystem/grpc-health-probe/releases/tag/v0.4.45) to address nvd.nist.gov/vuln/detail/CVE-2025-68121 affecting `grpc_health_probe`.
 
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -2,439 +2,146 @@
...
+
+### Test-Driven Development
+
+Follow TDD: write failing test → observe failure → implement → run tests → refactor.
+
+### YAML Matrix Tests
+
+Primary test format for Check, ListObjects, and ListUsers APIs. Test files embedded via `go:embed` in `assets/assets.go`:
+- `assets/tests/consolidated_1_1_tests.yaml` - Core relationship tests
+- `assets/tests/abac_tests.yaml` - Condition/CEL tests
```
adriantam, line 97: Feel that this is a bit misleading here.  The text written as it is implies the above are the matrix test while the actual matrix test is in https://github.com/openfga/openfga/blob/main/tests/check/check.go and https://github.com/openfga/openfga/blob/main/tests/listobjects/matrix.go

line 125 implies that tests/check and tests/listobjects are simply the runner of the two above set of tests while in reality there is a big suite of matrix test inside the tests/check tests/listobjects itself

senojj: Line 125 states the entry points of the tests i.e. the "runners".

These lines are simply there to inform an agent about the location of yaml test files.

Perhaps it is the word "Primary" that is throwing you off? I can remove that word and point it to the matrix tests as the "Primary" tests.

AGENTS.md
```diff
@@ -2,439 +2,146 @@
...
+| `pkg/storage/` | `OpenFGADatastore` interface + implementations (memory, postgres, mysql, sqlite) |
+| `pkg/typesystem/` | Authorization model parsing/validation, `weighted_graph.go` for query path optimization |
+| `internal/validation/` | Tuple/request validation against type system |
+| `internal/authn/`, `internal/authz/` | Authentication (None, Preshared Key, OIDC) and API authorization |
 
-- 8080: HTTP API
-- 8081: gRPC API
-- 3000: Playground UI
-- 2112: Prometheus metrics
+### Resolver Chain
```
justincoh, line 50: Should we make it clear that this is an implementation detail specific to `check`, or do you think the names of the resolvers themselves are enough of a hint?

senojj: These are bits of information that the agent decided would be helpful while providing sufficient context for other agents. The goal was to remove any context that could be inferred from the project files.

Remember, this is additional context for an agent that is parsing the files in the project. It is not context for a human.

PR #2950, opened 2026-02-26
fix: `cache_item_count` metric overcounting

### LINKED ISSUE #2944: possible double-increment in cache item count metric
### Checklist

### Description

Currently, the `InMemoryLRUCache` implementation blindly increments the cache item count on any call to `(InMemoryLRUCache).Set`. However, the Theine library used internally will always update any existing key's value in place, extending its TTL. In this scenario, the blind increment of cache item count creates an invalid count, reporting more items than actually exist within the cache.

Credit for the find of this bug goes to: @evanmay

### OpenFGA version

1.11.6

As a binary

In-Memory

### OpenFGA Flags

no

### Logs

_No response_

justincoh:
Our cache code was improperly incrementing the `cache_item_count` metric by blindly incrementing on any `cache.Set` call, even if it was for an overwrite. This caused the metric to steadily increase over time like so:

image

Example: Calling `cache.Set("key", {random_ulid})` 1000 times would make it seem like there were 1000 items in the cache, even though there is actually only one.

Rather than track cache_item_count ourselves via `.Inc()` and `.Dec()` calls, use the cache client's built in `.EstimatedSize()` method.

## References

Fixes #2944 

## Review Checklist

## Bug Fixes
- Fixed a metrics accuracy issue where cache item count was incorrectly incrementing when overwriting existing entries, which was causing the counter to drift upwards over time.


pkg/storage/cache.go
```diff
@@ -163,8 +163,17 @@ func (i InMemoryLRUCache[T]) Set(key string, value T, ttl time.Duration) {
...
+
+	// If we're overwriting a key, we don't want to increment metrics later
+	_, isOverwrite := i.client.Get(key)
+
+	// Ignore the boolean return here as we always pass cost=1 and items are always admitted
 	i.client.SetWithTTL(key, value, 1, ttl)
 
+	if isOverwrite {
+		return
+	}
```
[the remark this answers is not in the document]
justincoh: This is a valid point, but I'm not sure I want to add a lock into a path this hot just for the sake of 100% metric accuracy; I think this risk is acceptable for the sake of performance.

adriantam (commented):
As discussed offline, we should consider using EstimatedSize() instead of tracking it ourselves.  The trade off is that we will lose entity type visibility.

pkg/storage/cache.go
```diff
@@ -133,7 +133,6 @@ func NewInMemoryLRUCache[T any](opts ...InMemoryLRUCacheOpt[T]) (*InMemoryLRUCac
 			entityLabel = unspecifiedLabel
 		}
 
-		cacheItemCount.WithLabelValues(entityLabel).Dec()
```
adriantam, line 136: Do we want to update with the estimated size?

justincoh: We don't need to, and we actually couldn't do it in here. 

This RemovalListener is passed to the cache builder before the cache client is constructed, so at this point there is no cache `client` instance with which to call `client.EstimatedSize()`.

Even without a metric call here though, the new configuration calling `cacheItemCount.Set()` is a suitable replacement for both `Inc()` and `Dec()`, since it can both increase and decrease the metric. Having `cacheItemCount.Set` called inside our `cache.Set` method should keep this metric very close to real-time since cache.Set is a very hot path, and it's called far more often than the removalListener (which only hits on eviction, expiration or removal).

adriantam (dismissed):

Issue #2948, opened 2026-02-26
dialGrpc internal HTTP-to-gRPC client does not reload TLS credentials on certificate rotation


Avinash-Kamath:
### Checklist

- [x] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [x] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

When TLS certificate rotation occurs (e.g. via cert-manager), the OpenFGA HTTP gateway becomes unable to proxy requests to the internal gRPC server, causing all HTTP API calls to fail with:
connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate signed by unknown authority"
Root Cause
In cmd/run/run.go, the dialGrpc function loads TLS credentials once at startup using credentials.NewClientTLSFromFile:

``` go
func (s *ServerContext) dialGrpc(config *serverconfig.Config) *grpc.ClientConn {
    if config.GRPC.TLS.Enabled {
        creds, err := credentials.NewClientTLSFromFile(config.GRPC.TLS.CertPath, "")
        // loaded once, never refreshed
        dialOpts = append(dialOpts, grpc.WithTransportCredentials(creds))
    }
}
```
While the gRPC server correctly uses watchAndLoadCertificateWithCertWatcher (via controller-runtime's certwatcher) to dynamically reload the serving certificate, the internal HTTP gateway's gRPC client credentials are never refreshed.
When a new certificate is issued by the CA:

The gRPC server reloads and starts serving the new cert 
The HTTP gateway's internal gRPC client still holds the old cert/CA loaded at startup.
TLS handshake between HTTP gateway and gRPC server fails

I can See the Dynamic certificate watcher  was added as a part of this this [PR](https://github.com/openfga/openfga/pull/2182) But the gateway cert was missed

### OpenFGA version

v1.11.6

### How are you running OpenFGA?

[...]

PR #2951, opened 2026-02-27
fix: gateway grpc client tls cert rotation

RepoRonin:
Fix HTTP gateway gRPC client not refreshing TLS credentials after certificate rotation

When TLS certificate rotation occurs (e.g. via cert-manager), the HTTP gateway becomes unable to proxy requests to the internal gRPC server, causing all HTTP API calls to fail with a 500 and the error:

```
connection error: desc = "transport: authentication handshake failed: tls: failed to verify certificate: x509: certificate signed by unknown authority"
```

The gRPC server correctly uses [watchAndLoadCertificateWithCertWatcher](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:1060:0-1088:1) (via controller-runtime's certwatcher) to dynamically reload the serving certificate. However, the HTTP gateway's internal gRPC client loads its TLS credentials once at startup using `credentials.NewClientTLSFromFile` and never refreshes them. When the server starts presenting the new certificate, the client still holds the old cert as its trust anchor, causing the TLS handshake to fail.

#### How is it being solved?

The static `credentials.NewClientTLSFromFile` call in [dialGrpc](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:609:0-670:1) is replaced with the existing [watchAndLoadCertificateWithCertWatcher](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:1060:0-1088:1) helper, which uses controller-runtime's certwatcher to watch the certificate file for changes. A `VerifyPeerCertificate` callback is used to dynamically build a cert pool from the watcher's latest certificate on each new connection and verify the server's presented certificate against it. This ensures the gateway client trusts the new server certificate immediately after rotation without requiring a restart.

- [dialGrpc](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:609:0-670:1) is updated to accept a `context.Context` (required to run the certwatcher goroutine) and uses `VerifyPeerCertificate` with the dynamically watched certificate as a trust anchor instead of a statically loaded credential.
- The call site in [Run](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:780:0-1016:1) is updated to pass the server context to [dialGrpc](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:609:0-670:1).
- A new test [TestHTTPServerWithGRPCTLSEnabled](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run_test.go:791:0-834:1) is added, which starts the server with gRPC TLS enabled and the HTTP gateway active, then verifies that HTTP requests can be proxied through the gateway successfully.

## References

- closes https://github.com/openfga/openfga/issues/2948

## Review Checklist
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev]

## Release Notes

* **New Features**
  * Enhanced gRPC TLS security with dynamic certificate rotation, enabling certificates to be updated at runtime without requiring a service restart.

* **Tests**
  * Added test coverage for gRPC TLS-enabled scenarios to ensure secure server connections function properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 ## [Unreleased]
 ### Changed
-- HTTP gateway's internal gRPC client now uses dynamic TLS credentials that automatically update on certificate rotation via certwatcher, preventing connection failures when certificates are rotated (e.g., by cert-manager). [#XXXX](https://github.com/openfga/openfga/pull/XXXX)
+- HTTP gateway's internal gRPC client now uses dynamic TLS credentials that automatically update on certificate rotation via certwatcher, preventing connection failures when certificates are rotated (e.g., by cert-manager). [#2951](https://github.com/openfga/openfga/pull/2951)
 
 ### Fixed
 - Fixed a few bugs. Two potential index out of bounds scenarios, and one cache of an invalid result. [#2942](https://github.com/openfga/openfga/pull/2942)


senojj:
A few thoughts on this:
1. We should avoid creating a new pool on every connection. Instead, a global pool that is managed behind an `atomic.Pointer`, and updated by a watcher whenever the cert file changes, is the most ideal.
2. Only the leaf certificate verification really matters in this case, as it is a local cert with a local connection, and an exact match is the target.
3. The hostname _must_ be verified when taking this approach, as Go disables automatic hostname checks when InsecureSkipVerify is set to `true` in the TLS configuration.

RepoRonin:
I am gonna look into it, and fix this

RepoRonin:
@senojj I made the changes, do inform me on further improvements

senojj:
You should add the leaf certificate to the RootCA pool and use the pool to verify the server's certificate; essentially certificate pinning. Just comparing the bytes of the certificate is insecure.

Also you can verify using the DNS name and the pool at the same time.

This PR adds a lot of complexity to simply add behavior to update the trusted certificate. The watcher has the ability to register a callback that is triggered on file changes.

RepoRonin:
## Changes

Previously, `VerifyPeerCertificate` in [dialGrpc](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:610:0-668:1) was rebuilding an `x509.CertPool`
on every TLS handshake, allocating and re-parsing certificates per connection.

I replaced this with a single pool held behind an `atomic.Pointer[x509.CertPool]`,
managed by a [newAtomicCertPool](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:670:0-711:1) helper that:

- Creates a `certwatcher` instance and uses [RegisterCallback](cci:1://file:///home/shashank/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.1/pkg/certwatcher/certwatcher.go:91:0-100:1) to build and atomically
  swap in a new pinned pool (containing only the leaf certificate) whenever the cert
  file changes on disk — no polling goroutine needed.
- Exposes the pool as an atomic pointer, so each TLS handshake does a single [Load()](cci:1://file:///home/shashank/PROJECTS/openfga/cmd/run/run.go:1101:0-1129:1)
  with no allocation.

`VerifyPeerCertificate` now does a single `peerCert.Verify` call with both `Roots`
(the pinned pool) and `DNSName` (hostname extracted from the gRPC address), handling
certificate pinning and hostname verification together — the latter being required
since `InsecureSkipVerify: true` disables Go's automatic hostname check.

@senojj do tell me if further changes are required

senojj:
A second watcher, watching the gRPC certificate file is unnecessary. Utilize the same watcher that is currently in use.

RepoRonin:
> A second watcher, watching the gRPC certificate file is unnecessary. Utilize the same watcher that is currently in use.

@senojj did the change, anything else ?

senojj:
Every call to `dialGrpc` will overwrite the grpc watcher's callback. Currently the function is only called once, however this could change in the future. 

Creating a package-level global pool would allow an "update" callback function to be registered within the `watchAndLoadCertificateWithCertWatcher` function without needing to return the watcher and pass it around throughout the codebase. Additionally, each connection created via `dialGrpc` would use the same global pool.

RepoRonin:
> Every call to `dialGrpc` will overwrite the grpc watcher's callback. Currently the function is only called once, however this could change in the future.
> 
> Creating a package-level global pool would allow an "update" callback function to be registered within the `watchAndLoadCertificateWithCertWatcher` function without needing to return the watcher and pass it around throughout the codebase. Additionally, each connection created via `dialGrpc` would use the same global pool.

Could you help me with that, like how do I solve it ?

RepoRonin:
> Every call to `dialGrpc` will overwrite the grpc watcher's callback. Currently the function is only called once, however this could change in the future.
> 
> Creating a package-level global pool would allow an "update" callback function to be registered within the `watchAndLoadCertificateWithCertWatcher` function without needing to return the watcher and pass it around throughout the codebase. Additionally, each connection created via `dialGrpc` would use the same global pool.

check now

cmd/run/run.go
```diff
@@ -1024,10 +1063,20 @@ func watchAndLoadCertificateWithCertWatcher(ctx context.Context, certPath, keyPa
...
+		leaf, parseErr := x509.ParseCertificate(cert.Certificate[0])
+		if parseErr != nil {
+			logger.Error("failed to parse rotated certificate for cert pool", zap.Error(parseErr))
+			return
+		}
+		pool := x509.NewCertPool()
+		pool.AddCert(leaf)
+		grpcTLSCertPool.Store(pool)
+		logger.Info("gRPC client cert pool updated", zap.String("certPath", certPath))
+	})
```
senojj: `watchAndLoadCertificateWithCertWatcher` is also called by the HTTP server code. We do not want this callback to be applied for the HTTP server's certificate. The callback can be a package-level function, instead of a closure (to avoid unnecessary allocation), and should be passed into the `watchAndLoadCertificateWithCertWatcher` function as an argument, only for the grpc server setup. For the HTTP server setup, a `nil` value should be passed as the callback value.

RepoRonin: I did the changes, check now

cmd/run/run.go
```diff
@@ -82,6 +84,27 @@ const (
 	datastoreURIFlag    = "datastore-uri"
 )
 
+// grpcTLSCertPool is a package-level pinned certificate pool for verifying
+// the gRPC server's TLS certificate from the HTTP gateway client. It is
+// populated and kept up-to-date by watchAndLoadCertificateWithCertWatcher
+// via the certwatcher's RegisterCallback. Every connection created by
+// dialGrpc shares this single pool.
+var grpcTLSCertPool sync.Pointer[x509.CertPool]
```
senojj, line 92: For safety, initialize the `grpcTLSCertPool` to an empty `x509.CertPool` on package `init`.

cmd/run/run.go
```diff
@@ -82,6 +84,27 @@ const (
...
+
+// updateGRPCCertPool builds a pinned x509.CertPool from the leaf certificate
+// and stores it in the global grpcTLSCertPool. It is intended to be passed as
+// the callback argument to watchAndLoadCertificateWithCertWatcher for the gRPC
+// server setup.
+func updateGRPCCertPool(cert tls.Certificate) {
+	leaf, err := x509.ParseCertificate(cert.Certificate[0])
```
senojj, line 109: For safety, ensure that `len(cert.Certificate) > 0` _before_ indexing.

cmd/run/run_test.go
```diff
@@ -789,6 +789,51 @@ func TestGRPCServingTLS(t *testing.T) {
...
+		defer certsAndKeys.Clean()
+
+		cfg := testutils.MustDefaultConfigWithRandomPorts()
+		cfg.GRPC.TLS = &serverconfig.TLSConfig{
+			Enabled:  true,
+			CertPath: certsAndKeys.serverCertFile,
+			KeyPath:  certsAndKeys.serverKeyFile,
+		}
+		// Port for TLS cannot be 0.0.0.0
+		cfg.GRPC.Addr = strings.ReplaceAll(cfg.GRPC.Addr, "0.0.0.0", "localhost")
```
senojj: Must be "localhost" because created test cert is for that domain name.

cmd/run/run_test.go
```diff
@@ -789,6 +789,51 @@ func TestGRPCServingTLS(t *testing.T) {
 	})
 }
 
+func TestHTTPServerWithGRPCTLSEnabled(t *testing.T) {
```
senojj, line 801: This test does not exercise the certificate rotation path, so need another to do so.

cmd/run/run_test.go
```diff
@@ -789,6 +789,51 @@ func TestGRPCServingTLS(t *testing.T) {
...
+		// credentials (via certwatcher) rather than a one-time static load.
+		client := retryablehttp.NewClient()
+		t.Cleanup(client.HTTPClient.CloseIdleConnections)
+
+		resp, err := client.Get(fmt.Sprintf("http://%s/stores", cfg.HTTP.Addr))
+		require.NoError(t, err)
+		defer resp.Body.Close()
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+	})
+}
```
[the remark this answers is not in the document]
senojj: The implementation now uses the leaf certificate of the same certificate used by the server as the root CA. When the server's certificate is verified, the leaf cert is used. The intent is to pin the certificate.

RepoRonin:
> Starting to shape up much better. Some additional changes needed though.

yes, working on them.
Thanks for the feedback

RepoRonin:
> Starting to shape up much better. Some additional changes needed though.

Hi ! I did the changes. Let me know on the follow ups

cmd/run/run_test.go
senojj: Your AI prompt resulted in multiple changed comments, but does not enforce that the value here is _always_ "localhost".

senojj:
Additionally, you will need to sign the CLA before this PR can be considered for merging, and you must add a "Changed" entry to the changelog [see here](https://github.com/openfga/openfga/blob/main/CHANGELOG.md).

cmd/run/run_test.go
RepoRonin: Yes, i used AI cuz i could not understand this one more clearly, isn't it already "localhost" .
Could you guide me more on what needs to be done ?

RepoRonin:
> Additionally, you will need to sign the CLA before this PR can be considered for merging, and you must add a "Changed" entry to the changelog [see here](https://github.com/openfga/openfga/blob/main/CHANGELOG.md).

I have signed the CLA, & added the "Changed" entry

cmd/run/run_test.go
senojj: This code changes the value to "localhost" only when the current value is "0.0.0.0". If the default value of `cfg.GRPC.Addr` were something like "test.com", this test would fail.

It is fine to use AI to generate the code changes, but it is important that you be able to properly vet, and understand all of the code that it generates before submitting it. I only mention this because much of the back-and-forth in this review was due to AI generated code changes that lacked a complete understanding of the change.

cmd/run/run_test.go
```diff
@@ -703,8 +712,6 @@ func TestHTTPServingTLS(t *testing.T) {
 			CertPath: certsAndKeys.serverCertFile,
 			KeyPath:  certsAndKeys.serverKeyFile,
 		}
-		// Port for TLS cannot be 0.0.0.0
-		cfg.HTTP.Addr = strings.ReplaceAll(cfg.HTTP.Addr, "0.0.0.0", "localhost")
```
senojj, line 707: This code is unrelated to your code change. Removing it will now cause the test to fail when `cfg.HTTP.Addr` defaults to anything other than "localhost". It currently defaults to "0.0.0.0", so it will currently fail.

cmd/run/run_test.go
```diff
@@ -769,8 +776,6 @@ func TestGRPCServingTLS(t *testing.T) {
 			CertPath: certsAndKeys.serverCertFile,
 			KeyPath:  certsAndKeys.serverKeyFile,
 		}
-		// Port for TLS cannot be 0.0.0.0
-		cfg.GRPC.Addr = strings.ReplaceAll(cfg.GRPC.Addr, "0.0.0.0", "localhost")
```
senojj, line 773: This code is unrelated to your code change. Removing it will now cause the test to fail when `cfg.GRPC.Addr` defaults to anything other than "localhost". It currently defaults to "0.0.0.0", so it will currently fail.

cmd/run/run_test.go
```diff
@@ -789,6 +794,106 @@ func TestGRPCServingTLS(t *testing.T) {
...
+			if err := runServer(ctx, cfg); err != nil {
+				log.Fatal(err)
+			}
+		}()
+
+		certPool := x509.NewCertPool()
+		certPool.AddCert(certsAndKeys.caCert)
+		creds := credentials.NewClientTLSFromCert(certPool, "")
+
+		testutils.EnsureServiceHealthy(t, cfg.GRPC.Addr, cfg.HTTP.Addr, creds)
```
senojj, line 830: this will fail if `cfg.GRPC.Addr` is anything other than "localhost". It currently defaults to "0.0.0.0". My previous comment provided instruction to use the "localhost" value directly, or set `cfg.GRPC.Addr` explicitly.

So simply adding a line like `cfg.GRPC.Addr = "localhost"` would have been sufficient.

cmd/run/run_test.go
```diff
@@ -789,6 +794,106 @@ func TestGRPCServingTLS(t *testing.T) {
...
```
senojj, line 869: this will fail if cfg.GRPC.Addr is anything other than "localhost". It currently defaults to "0.0.0.0".

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Changed
+- HTTP gateway's internal gRPC client now uses dynamic TLS credentials that automatically update on certificate rotation via certwatcher, preventing connection failures when certificates are rotated (e.g., by cert-manager). [#XXXX](https://github.com/openfga/openfga/pull/XXXX)
```
senojj: replace "XXXX" with this PR's number.

RepoRonin: done

RepoRonin:
@senojj check now

senojj:
I'm honestly unsure what the latest changes are attempting to achieve. The AI tool that you are using is removing and then adding back and removing and then adding back. These recent changes still fail to always enforce "localhost" as a value for the gRPC server, explicitly.

If the ask is too unclear, please feel free to ask further clarifying questions.

Alternatively, if you are familiar with programming, I would suggest writing this minimal change by hand.

If you are not familiar with programming, then I would suggest that we close this PR, and I can implement the enhancement within another branch.

CHANGELOG.md
senojj: You did not make the change.

RepoRonin:
> I'm honestly unsure what the latest changes are attempting to achieve. The AI tool that you are using is removing and then adding back and removing and then adding back. These recent changes still fail to always enforce "localhost" as a value for the gRPC server, explicitly.
> 
> If the ask is too unclear, please feel free to ask further clarifying questions.
> 
> Alternatively, if you are familiar with programming, I would suggest writing this minimal change by hand.
> 
> If you are not familiar with programming, then I would suggest that we close this PR, and I can implement the enhancement within another branch.

I'll work on it.

RepoRonin:
@senojj 
I'd like your review on this : 
where we bind the gRPC listener if I replace this : 


listener, err := net.Listen("tcp", cfg.GRPC.Addr)


with 

```go
host, port, err := net.SplitHostPort(cfg.GRPC.Addr)
if err != nil {
    return fmt.Errorf("invalid grpc addr: %w", err)
}

_ = host // ignore whatever user passed

cfg.GRPC.Addr = fmt.Sprintf("localhost:%s", port)

listener, err := net.Listen("tcp", cfg.GRPC.Addr)

**Comment by senojj** [comment]
> @senojj I'd like your review on this : where we bind the gRPC listener if I replace this :
> 
> listener, err := net.Listen("tcp", cfg.GRPC.Addr)
> 
> with
> 
> ```go
> host, port, err := net.SplitHostPort(cfg.GRPC.Addr)
> if err != nil {
>     return fmt.Errorf("invalid grpc addr: %w", err)
> }
> 
> _ = host // ignore whatever user passed
> 
> cfg.GRPC.Addr = fmt.Sprintf("localhost:%s", port)
> 
> listener, err := net.Listen("tcp", cfg.GRPC.Addr)
> ```

Remember this is just for a test. For our needs we need only specify a valid hostname, given our TLS certificate, and a usable port.

Therefore something like this would suffice:
```go
cfg.GRPC.Addr = "localhost:8082"
```
If you wanted to get fancier with a port scanning function, feel free.

RepoRonin:
> > @senojj I'd like your review on this : where we bind the gRPC listener if I replace this :
> > listener, err := net.Listen("tcp", cfg.GRPC.Addr)
> > with
> > ```go
> > host, port, err := net.SplitHostPort(cfg.GRPC.Addr)
> > if err != nil {
> >     return fmt.Errorf("invalid grpc addr: %w", err)
> > }
> > 
> > _ = host // ignore whatever user passed
> > 
> > cfg.GRPC.Addr = fmt.Sprintf("localhost:%s", port)
> > 
> > listener, err := net.Listen("tcp", cfg.GRPC.Addr)
> > ```
> 
> Remember this is just for a test. For our needs we need only specify a valid hostname, given our TLS certificate, and a usable port.
> 
> Therefore something like this would suffice:
> 
> ```go
> cfg.GRPC.Addr = "localhost:8082"
> ```
> 
> If you wanted to get fancier with a port scanning function, feel free.

Makes sense, since this is only for testing and we just need a valid hostname that matches the TLS certificate and a free port, I’ll simplify this to explicitly set cfg.GRPC.Addr = "localhost:8082" and avoid the additional parsing logic.

RepoRonin:
@senojj 
give it a check now

CHANGELOG.md
senojj: This still need to be corrected.

RepoRonin: > This still need to be corrected.

done

cmd/run/run.go
```diff
@@ -612,10 +645,40 @@ func (s *ServerContext) dialGrpc(config *serverconfig.Config) *grpc.ClientConn {
...
+				if err != nil {
+					return fmt.Errorf("failed to parse peer certificate: %w", err)
+				}
+
+				// Verify the leaf certificate against the global pinned pool
+				// and check the hostname in a single call. The pool is
+				// updated atomically by the watcher callback on cert rotation.
+				_, err = peerCert.Verify(x509.VerifyOptions{
+					Roots:   grpcTLSCertPool.Load(),
+					DNSName: grpcHost,
```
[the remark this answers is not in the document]
senojj: The server's leaf certificate is added to the client's RootCAs pool in order to pin the certificate.

cmd/run/run.go
```diff
@@ -82,6 +84,37 @@ const (
...
+func updateGRPCCertPool(cert tls.Certificate) {
+	if len(cert.Certificate) == 0 {
+		return
+	}
+	leaf, err := x509.ParseCertificate(cert.Certificate[0])
+	if err != nil {
+		return
+	}
+	pool := x509.NewCertPool()
+	pool.AddCert(leaf)
```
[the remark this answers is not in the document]
senojj: This PR is using certificate pinning as suggested.

senojj (approved):
The code and implementation look good to me. The rest of the team will review on Monday.

RepoRonin:
HI @senojj 
Currently this codecov/patch is failing, I don't know about it, could you tell me about it ? 
How do I resolve this ?

senojj:
> Currently this codecov/patch is failing, I don't know about it, could you tell me about it ?
How do I resolve this?

This is a check for code test coverage. What it is saying is that, of the lines of code you changed, 66.67% of those lines were hit during the test runs. The lines that are not covered by unit tests are primarily error edge cases that are very unlikely to happen.

This is not a required check, however, if you wish to satisfy it, I would not oppose you doing so, so long as you keep the tests simple and clean.

RepoRonin:
> The code and implementation look good to me. The rest of the team will review on Monday.

Any updates ?

RepoRonin:
Glad I could help

PR #2961, opened 2026-03-06
fix: correct swapped format args in DecodeParameterType error message

archy-rock3t-cloud:
The error message in `DecodeParameterType` reports "requires [found]; found [required]" because the format arguments are swapped. This fixes the order so the message correctly shows the required count first and the found count second.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected parameter type validation error message formatting to display the found count and expected count in the proper sequence for improved clarity.


adriantam:
@archy-rock3t-cloud thank you for your contribution.  Please update CHANGELOG.  If you have time, can you add unit test around the code - but this is more as a +1, not blocker.  Thank you.

archy-rock3t-cloud:
@adriantam sure, updated

PR #2976, opened 2026-03-12
feat: add graceful shutdown timeout configuration

rafanaskin:
Added ShutdownTimeout configuration.

Increased the shutdown timeout in tests from 5s to 30s to prevent [server shutdown failures](https://github.com/openfga/openfga/actions/runs/22220603519/job/64274799418#step:5:334) (context deadline exceeded) that could result in "found unexpected goroutines" errors.

Currently, the server shutdown timeout is hardcoded to 5 seconds. In rare [cases](https://github.com/openfga/openfga/actions/runs/22220603519/job/64274799418#step:5:334), this is not enough for tests to gracefully shut down the server, which can result in "found unexpected goroutines" errors.

#### How is it being solved?

Added the `ShutdownTimeout` configuration:
- flag: `shutdown-timeout`.
- environment variable: `OPENFGA_SHUTDOWN_TIMEOUT`.

- Added the `ShutdownTimeout` configuration.
- Increased the shutdown timeout in tests from 5s to 30s.
- Moved `MustDefaultConfig` to the `testutils` package.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Configurable server shutdown timeout — set via CLI flag, env var, or config file (default 10s).
  * New planner cleanup interval configurable via CLI flag/env var.
* **Behavior**
  * Shutdown uses configured timeout and runs component cleanups concurrently for faster termination.
* **Bug Fixes**
  * Validation now rejects zero or negative shutdown timeout values.
* **Tests**
  * Tests updated to cover the new timeout and validation.
* **Documentation**
  * CHANGELOG updated to note the added shutdown timeout configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 ### Changed
 - Changed the ListObjects pipeline intersection algorithm to improve intersection performance. [#3031](https://github.com/openfga/openfga/pull/3031)
 - **[BREAKING]** The Playground now only supports the `none` authentication method. Running the Playground with `preshared` key authentication is no longer supported. The server will error and not start if it detects this combination.
+- The Playground is now disabled by default as a result of [GHSA-68m9-983m-f3v5](https://github.com/openfga/openfga/security/advisories/GHSA-68m9-983m-f3v5)
 
 ### Deprecated
 - The built-in OpenFGA Playground is intended for development purposes only and is deprecated. It will be removed entirely in a future release.


pkg/server/config/config.go
```diff
@@ -857,5 +865,7 @@ func MustDefaultConfig() *Config {
 	config.Playground.Enabled = false
 	config.Metrics.Enabled = false
 
+	config.ShutdownTimeout = 30 * time.Second
+
```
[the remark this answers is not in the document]
senojj: I know this function existed prior to the changes of this PR, but I'm struggling to understand its purpose based in its name/signature. Extending from that lack of understanding, I'm unsure why we would hard code the shutdown timeout to 30 seconds here; feels like a magic number assignment for an unknown reason.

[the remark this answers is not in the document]
rafanaskin: @senojj, I agree with you.

I’ve moved `MustDefaultConfig` to the `testutils` package, which feels like a more appropriate place for it.
Now, its purpose seems clearer: it provides a default configuration for tests. This config includes changes suitable for testing, such as disabled metrics and increased timeouts to handle slower test environments.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 ## [1.12.0] - 2026-03-13
 ### Added
 - Add configuration for maximum size of received gRPC message bytes. [#2952](https://github.com/openfga/openfga/pull/2952)
+- Added configuration for the server shutdown timeout. [#2976](https://github.com/openfga/openfga/pull/2976)
```
senojj: This changelog item must be added to the `[Unreleased]` section. (may have been moved after a merge from main)

rafanaskin: Sorted

senojj:
@adriantam, it appears that this code has changed significantly from when I first looked at it. I agree that the "cleanups" must be processed in LIFO order. 

Additionally, a linked-list may be a better data structure than a slice here. Appealing properties include constant time prepends -- with the potential to be atomic -- without need for resizing, dynamic memory, and the ability to traverse from the head back to the tail without any need for sort. Traversal could look like `for i := head; i != nil; i = i.prev {`. The tradeoff would be a heap allocation for each entry.

However, since the maximum number of possible cleanups can be known at compile time, it would also be possible to use a statically sized array, and add elements starting at the end of the array. Traversal could look like: `for i := len(a) - 1; i >= 0 && a[i] != nil; i-- {`. This might use a little extra memory, but would be faster and would require 0 heap allocations.

rafanaskin:
@senojj, for readability, I implemented a `LIFO` shutdown order using a linked list. Could you please take a look?

cmd/run/run.go
```diff
@@ -1058,49 +1072,29 @@ func (s *ServerContext) Run(ctx context.Context, config *serverconfig.Config) er
...
-	if playground != nil {
-		if err := playground.Shutdown(ctx); err != nil {
-			s.Logger.Info("failed to shutdown the playground", zap.Error(err))
+	for el := cleanups.Front(); el != nil; el = el.Next() {
+		clean, ok := el.Value.(cleanup)
+		if !ok {
+			s.Logger.Info("failed to convert cleanup", zap.Any("value", el.Value))
+			continue
 		}
-	}
```
senojj: This should probably exist within a deferred function that is declared after instantiation of the `cleanups` list on line 879. That way, an early return will result in any existing resources being cleaned up properly.

.config-schema.json
senojj, line 1: It would be better to only make the necessary changes, without changing the formatting of the entire JSON structure.

rafanaskin: You're absolutely right - I made this occasionally.

cmd/run/run.go
rafanaskin: Great idea! This is a significant improvement.

PR #2979, opened 2026-03-13
new: add selected stategy counter for visibility into dynamic query t…

hamilton-earthscope:
## Description

Add a new prom metric `planner_strategy_selected_count` for visibility into the dynamic query tuning behavior. It simply increments on every planner selection with a dimension `{strategy=selectedPlan.Name}`

There is currently no visibility into the dynamic query tuning behavior.

## References
- https://auth0.com/blog/self-tuning-strategy-planner-openfga/

## Review Checklist

## Summary by CodeRabbit

* **Chores**
  * Enhanced planner observability with metric tracking for strategy selections.

### DIFF
--- a/internal/planner/plan.go
+++ b/internal/planner/plan.go
@@ -76,7 +76,10 @@ func (kp *keyPlan) Select(resolvers map[string]*PlanConfig) *PlanConfig {
 	}
 
 	selected := resolvers[bestResolver]
-	strategySelectedCounter.WithLabelValues(selected.Name).Inc()
+	if selected != nil {
+		strategySelectedCounter.WithLabelValues(selected.Name).Inc()
+	}
+
 	return selected
 }
 
--- a/internal/planner/plan.go
+++ b/internal/planner/plan.go
@@ -5,7 +5,18 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/openfga/openfga/internal/build"
 )
+
+var strategySelectedCounter = promauto.NewCounterVec(prometheus.CounterOpts{
+	Namespace: build.ProjectName,
+	Name:      "planner_strategy_selected_count",
+	Help:      "The total number of times each traversal strategy was selected by the query planner.",
+}, []string{"strategy"})
 
 // keyPlan manages the statistics for a single key and makes decisions about its resolvers.
 // This struct is now entirely lock-free, using a sync.Map to manage its stats.
@@ -64,7 +75,9 @@ func (kp *keyPlan) Select(resolvers map[string]*PlanConfig) *PlanConfig {
 		}
 	}
 
-	return resolvers[bestResolver]
+	selected := resolvers[bestResolver]
+	strategySelectedCounter.WithLabelValues(selected.Name).Inc()
+	return selected
 }
 
 // UpdateStats performs the Bayesian update for the given resolver's statistics.


adriantam:
Hello @hamilton-earthscope , thank you for your contribution.  Can you sign the CLA as per CNCF requirement?

hamilton-earthscope:
Hi @adriantam. It will take some time to get the corporate CLA signed.

In the meantime, I open this PR as an avenue for discussion: is this PR something the team is interested in accepting? 

Ultimately, I am after visibility into the dynamic query tuning. I am suspicious it is to blame for some behavior we're observing, but OpenFGA is not currently emiting metrics to back up or disprove my suspicion. This PR represents a bare minimum attempt at adding such visibility. I am unsure whether or not the metric added in this PR is broadly useful, generally sufficient, or even aligns with the teams observability practices.

Can we have that conversation while the CLA is pending?

adriantam:
> Hi @adriantam. It will take some time to get the corporate CLA signed.
> 
> In the meantime, I open this PR as an avenue for discussion: is this PR something the team is interested in accepting?
> 
> Ultimately, I am after visibility into the dynamic query tuning. I am suspicious it is to blame for some behavior we're observing, but OpenFGA is not currently emiting metrics to back up or disprove my suspicion. This PR represents a bare minimum attempt at adding such visibility. I am unsure whether or not the metric added in this PR is broadly useful, generally sufficient, or even aligns with the teams observability practices.
> 
> Can we have that conversation while the CLA is pending?

We believe it is a good idea to add statistics on the planner.

senojj:
Will this metric be useful without knowledge of the plan key? Counting the number of times that a particular strategy is chosen, without knowing at what point in the authorization model graph the strategy was employed, doesn't seem particularly useful for diagnostics.

If you want to correlate a spike in latency with an increase in the selection of a particular strategy, looking only at the strategy count would often be too broad and would likely muddle/obscure anything other than a change with a massive delta. The plan key would be required to provide sufficient fidelity for the identification of meaningful changes in strategy selection that could affect latency.

Adding the plan key would be trivial, but it is a value with moderate cardinality.

Additionally, adding a counter metric in the proposed location throws (another) atomic synchronization point into a very hot path with high contention potential.

I _would_ suggest adding this information to existing spans at strategy decision points, as a trace would be the appropriate way to make use of this information, but we currently produce so many spans that it may be difficult to get at this information.

At the very least, my suggestion would be to add the plan key as a metric dimension. Additionally, don't increment the counter within the keyPlan type. Instead, it will need to be incremented at the various decision points within the Check resolution path.

Vic-Dev:
@hamilton-earthscope Do you have any response to the above comments? If we don't receive a response in a week we will be closing this PR due to inactivity.

hamilton-earthscope:
> Will this metric be useful without knowledge of the plan key? Counting the number of times that a particular strategy is chosen, without knowing at what point in the authorization model graph the strategy was employed, doesn't seem particularly useful for diagnostics.
>
> If you want to correlate a spike in latency with an increase in the selection of a particular strategy

I wish to correlate latency with the selection of a particular strategy.

The behavior we observe is bimodal rather than spiky - when containers start up, they either perform with ~5ms p90 and <10% CPU use or 250+ms p90 with 75%+ CPU use. It seems to be about 50/50 whether we get a "good" container or a "bad" one. We have not been able to correlate any rhyme or reason. We've been experiencing this for months through versions 1.11.x - 1.13.x (our production workload went live around 2025-12-28 so I don't know if the behavior existed before 1.11).

I am after observability into the dynamic query planner to see if it is somehow locking in on a suboptimal strategy in the "bad" containers.

> I would suggest adding this information to existing spans at strategy decision points, as a trace would be the appropriate way to make use of this information, but we currently produce so many spans that it may be difficult to get at this information.

Unfortunately, we are not using traces currently because they are so voluminous. A single batch check can produce thousands of spans and we're performing 500+ check/sec sustained with spikes up to 2k+checks/sec. Maybe we will revisit trace sampling in the future.

With the bimodal behavior we observe, I thought a simple "containers A and B are using planner strategy X while container C is using planner strategy Y" metric would clearly tell us if the planner is or is not to blame. That was the goal of this PR.

I think at this point we can just close this and I'll open an issue instead.

senojj:
> With the bimodal behavior we observe, I thought a simple "containers A and B are using planner strategy X while container C is using planner strategy Y" metric would clearly tell us if the planner is or is not to blame. That was the goal of this PR.

@hamilton-earthscope what you are looking for is entirely valid. However, my point was that a single graph traversal will reference the Thompson sampling multiple times at various edges to determine which algorithm/strategy to employ at a given edge.

While it would be painfully obvious if the planner were to choose the same strategy repeatedly for every edge -- like the "default" strategy vs the "weight2" strategy -- it is possible, but unlikely that every edge would qualify for the same candidate strategies.

Likely, the bimodal behavior that you are seeing is in relation to a particular edge choosing a less optimal strategy.

This can already be seen in the traces, actually. If you were to obtain a sample trace, it may point you to where it is going awry.

> I think at this point we can just close this and I'll open an issue instead.

If you open an issue, but are unable to obtain a sample trace, might I ask a few questions to answer within the issue? 

Regarding the bimodal behavior:
- is the low p99 consistent over the lifetime of the container?
- are you seeing different p99 values across containers for the exact same inputs?
- are you able to share an authorization model that is representative of the one you are using?

hamilton-earthscope:
> Regarding the bimodal behavior:
>
> - is the low p99 consistent over the lifetime of the container?
> - are you seeing different p99 values across containers for the exact same inputs?
> - are you able to share an authorization model that is representative of the one you are using?

Yes, I can provide all of that. I plan to follow https://openfga.dev/docs/getting-started/setup-openfga/reporting-runtime-issues but just haven't had time to collect all the requested info.

Another reason for closing this would be that I've not been able to make any progress on my org signing the CLA.

So I think I'll open two issues:
1. observability into the dynamic query planner
  a. effectively the ask in this PR - without knowledge of the openfga internals, it's probably best for you all to choose where to put the metric and/or span attr (for the reasons you pointed out)
3. the runtime behavior we're observing that may or may not be the query planner

Does that sound like a reasonable plan?

curfew-marathon:

PR #2984, opened 2026-03-16
perf: tuple validation performance tweaks

senojj:
Many of the widely used tuple validation and manipulation functions are written in less-than-efficient ways.

#### How is it being solved?
Introduce benchmarks to measure function performance, and refactor implementations for performance.

## Review Checklist

## Release Notes

* **Performance**
  * Optimized memory allocations and string building in tuple operations for improved efficiency.
  * Refined validation logic with updated type-checking mechanisms.

* **Tests**
  * Added comprehensive benchmark tests covering tuple string conversion, serialization utilities, and user part operations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 ## [Unreleased]
 ### Changed
 - The ListObjects "pipeline" algorithm ditches its custom Pipe implementation and replaces it with Go native channels. [#2977](https://github.com/openfga/openfga/pull/2977)
+- Refactor tuple validation and manipulation functions for optimal performance. [#2984](https://github.com/openfga/openfga/pull/2984)
 
 ### Fixed
 - Fixed `OTEL_EXPORTER_OTLP_ENDPOINT` not accepting URIs with schemes (e.g. `http://host:4317`). The scheme is now stripped before passing to the gRPC exporter, and an `https://` scheme enables TLS regardless of the `trace.otlp.tls.enabled` flag. [#2981](https://github.com/openfga/openfga/pull/2981)


pkg/tuple/tuple.go
```diff
@@ -271,14 +302,9 @@ func StringToUserProto(userKey UserString) *openfgav1.User {
...
-	case -1:
-		return "", object
-	case len(object) - 1:
-		return object[0:i], ""
-	default:
-		return object[0:i], object[i+1:]
-	}
+	ndx := strings.IndexByte(object, ':')
+	i := ndx &^ (ndx >> (64 - 1))
+	return object[0:i], object[ndx+1:]
```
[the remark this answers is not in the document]
senojj: addressed in commit a412298

pkg/tuple/tuple.go
```diff
@@ -226,15 +234,38 @@ type UserString = string
...
 func UserProtoToString(obj *openfgav1.User) UserString {
-	switch obj.GetUser().(type) {
+	switch v := obj.GetUser().(type) {
 	case *openfgav1.User_Wildcard:
-		return obj.GetWildcard().GetType() + ":*"
+		t := v.Wildcard.GetType()
+		buf := make([]byte, len(t), len(t)+2)
+		copy(buf, t)
+		buf = append(buf, ':', '*')
+		return unsafe.String(unsafe.SliceData(buf), len(buf))
```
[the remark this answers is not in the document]
senojj: the benchmarks show a significant decrease in ns/op when using unsafe here.

pkg/tuple/tuple.go
```diff
@@ -572,14 +606,20 @@ func ToUserParts(user string) (string, string, string) {
...
-		user = user + "#" + userRelation
-	}
-	return user
+	size := len(userObjectType) + len(userObjectID) + len(userRelation) + 2
+	buf := make([]byte, size)
+	w := copy(buf, userObjectType)
+	if w > 0 && size > w {
+		buf[w] = ':'
+		w += 1
+	}
```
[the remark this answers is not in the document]
senojj: The overallocation, at most, would be 1 element. This is a reasonable trade-off for a single allocation without introducing additional logic branching.

pkg/tuple/tuple.go
```diff
@@ -28,11 +29,18 @@ func (t *Tuple) GetUser() string {
 func (t *Tuple) String() string {
 	tk := (*openfgav1.TupleKey)(t)
 
-	return tk.GetObject() +
-		"#" +
-		tk.GetRelation() +
-		"@" +
-		tk.GetUser()
```
justincoh, line 35: I thought we had a linter that specifically flagged this kind of thing 🤔

pkg/tuple/tuple.go
justincoh, line 243: How much does this gain us vs just using `strings.Builder` with `Grow` like elsewhere in this PR? I'm hesitant to use anything from the `unsafe` package.

senojj: Well, if it makes you feel any better, here is the definition of function `(*Builder).String()`:
```go
func (b *Builder) String() string {
	return unsafe.String(unsafe.SliceData(b.buf), len(b.buf))
}
```

pkg/tuple/tuple.go
senojj: This is _less_ bad than `fmt.Sprintf`, but still not as good as the proposed alternative.

pkg/tuple/tuple.go
senojj: The key here is that the array that underlies the returned string is not modifiable once this function returns.

PR #3006, opened 2026-03-24
perf: remove `fmt` from cache key generation

justincoh:
## Description

OpenFGA is using `fmt.sprintf` when generating cache keys for requests with context. This PR removes fmt from cache key generation as it's unnecessary and slows things down.

For the WriteStruct method (which writes the context's cache key) benchmarks show:
* 30% improvement in sec/op
* 38% reduction in B/op,
* 57% reduction in allocs/op

Full benchstat details and full benchmark code are in the collapsible sections below.

Benchmark was run like so: `go test -bench=BenchmarkWriteStruct -benchmem -cpu=4 -count=10`.

<details>

<summary>Full benchstat output</summary>

```
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/pkg/storage
cpu: Apple M3 Max
                     │ /Users/justin.cohen/repos/openfga/openfga/bench_old.txt │ /Users/justin.cohen/repos/openfga/openfga/bench_new.txt │
                     │                         sec/op                          │              sec/op               vs base               │
WriteStruct/small-4                                                235.1n ± 1%                        149.6n ± 2%  -36.37% (p=0.001 n=7)
WriteStruct/medium-4                                               772.7n ± 1%                        518.7n ± 4%  -32.87% (p=0.001 n=7)
WriteStruct/nested-4                                               449.0n ± 0%                        371.1n ± 1%  -17.35% (p=0.001 n=7)
geomean                                                            433.7n                             306.5n       -29.32%

                     │ /Users/justin.cohen/repos/openfga/openfga/bench_old.txt │ /Users/justin.cohen/repos/openfga/openfga/bench_new.txt │
                     │                          B/op                           │               B/op                vs base               │
WriteStruct/small-4                                                104.00 ± 0%                         56.00 ± 0%  -46.15% (p=0.001 n=7)
WriteStruct/medium-4                                                544.0 ± 0%                         352.0 ± 0%  -35.29% (p=0.001 n=7)
WriteStruct/nested-4                                                176.0 ± 0%                         120.0 ± 0%  -31.82% (p=0.001 n=7)
geomean                                                             215.1                              133.2       -38.07%

                     │ /Users/justin.cohen/repos/openfga/openfga/bench_old.txt │ /Users/justin.cohen/repos/openfga/openfga/bench_new.txt │
                     │                        allocs/op                        │            allocs/op              vs base               │
WriteStruct/small-4                                                 7.000 ± 0%                         3.000 ± 0%  -57.14% (p=0.001 n=7)
WriteStruct/medium-4                                               19.000 ± 0%                         7.000 ± 0%  -63.16% (p=0.001 n=7)
WriteStruct/nested-4                                                8.000 ± 0%                         4.000 ± 0%  -50.00% (p=0.001 n=7)
geomean                                                             10.21                              4.380       -57.10%

```

</details>

### Full benchmark code
<details>

<summary>The full benchmark code</summary>

This was the actual benchmark code used
```go
func BenchmarkWriteStruct(b *testing.B) {
	small := MustNewStruct(map[string]any{
		"keyA": "valueA",
		"keyB": "valueB",
	})

	medium := MustNewStruct(map[string]any{
		"boolKey":   true,
		"stringKey": "hello",
		"numberKey": 1.2,
		"nullKey":   nil,
		"structKey": map[string]any{
			"key1": "value1",
		},
		"listKey": []any{"item1", "item2"},
	})

	nested := MustNewStruct(map[string]any{
		"level1": map[string]any{
			"level2": map[string]any{
				"level3": map[string]any{
					"key": "value",
				},
			},
		},
		"other": "data",
	})

	cases := []struct {
		name  string
		input *structpb.Struct
	}{
		{"small", small},
		{"medium", medium},
		{"nested", nested},
	}

	for _, tc := range cases {
		b.Run("Old/"+tc.name, func(b *testing.B) {
			w := &strings.Builder{}
			for n := 0; n < b.N; n++ {
				_ = writeStructOld(w, tc.input)
				w.Reset()
			}
		})
		b.Run("New/"+tc.name, func(b *testing.B) {
			w := &strings.Builder{}
			for n := 0; n < b.N; n++ {
				_ = writeStruct(w, tc.input)
				w.Reset()
			}
		})
	}
}
```

</details>

## Summary by CodeRabbit

* **Refactor**
  * Simplified cache key serialization so keys are emitted more directly and with finer error handling during writes.
* **Documentation**
  * Added a changelog entry noting the serialization change.
* **Tests**
  * Updated and expanded unit tests to reflect new write behavior and validate additional failure scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 ## [Unreleased]
 ### Changed
+- Made some minor changes in ListObjects to reduce heap allocations. Results in minor latency reduction. [#3043](https://github.com/openfga/openfga/pull/3043)
 - Improve cache key generation performance by removing `fmt` usage and extend control-character sanitization to all cache key inputs (tuples, conditions, context). [#3006](https://github.com/openfga/openfga/pull/3006)
 
 ## [1.14.0] - 2026-04-03
--- a/internal/containers/atomic_map.go
+++ b/internal/containers/atomic_map.go
@@ -54,6 +54,14 @@ func (m *AtomicMap[K, V]) Clear() {
 	m.m = nil
 }
 
+// Len returns the number of entries in the map.
+func (m *AtomicMap[K, V]) Len() int {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	return len(m.m)
+}
+
 // Unwrap returns the internal map that backs the AtomicMap.
 // A nil value may be returned if the AtomicMap has not yet
 // been initialized, or Unwrap is called after a Clear call.
@@ -61,5 +69,8 @@ func (m *AtomicMap[K, V]) Clear() {
 // Once the internal map has been unwrapped, methods on the
 // AtomicMap instance are no longer safe.
 func (m *AtomicMap[K, V]) Unwrap() map[K]V {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
 	return m.m
 }
--- a/internal/containers/atomic_map_test.go
+++ b/internal/containers/atomic_map_test.go
@@ -1,0 +1,93 @@
+package containers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAtomicMap_Store(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	m.Store("a", 1)
+
+	v, ok := m.LoadOrStore("a", 0)
+	require.True(t, ok)
+	assert.Equal(t, 1, v)
+}
+
+func TestAtomicMap_Store_OverwritesExisting(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	m.Store("a", 1)
+	m.Store("a", 2)
+
+	v, ok := m.LoadOrStore("a", 0)
+	require.True(t, ok)
+	assert.Equal(t, 2, v)
+}
+
+func TestAtomicMap_LoadOrStore_KeyAbsent(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	v, loaded := m.LoadOrStore("x", 42)
+
+	assert.False(t, loaded)
+	assert.Equal(t, 42, v)
+
+	// Subsequent load should find the stored value.
+	v, loaded = m.LoadOrStore("x", 99)
+	assert.True(t, loaded)
+	assert.Equal(t, 42, v)
+}
+
+func TestAtomicMap_LoadOrStore_KeyPresent(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	m.Store("k", 10)
+
+	v, loaded := m.LoadOrStore("k", 20)
+
+	assert.True(t, loaded)
+	assert.Equal(t, 10, v)
+}
+
+func TestAtomicMap_Clear(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	m.Store("a", 1)
+	m.Store("b", 2)
+	m.Clear()
+
+	assert.Nil(t, m.Unwrap())
+}
+
+func TestAtomicMap_Clear_ThenReuse(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	m.Store("a", 1)
+	m.Clear()
+
+	// Map should be usable again after Clear.
+	m.Store("b", 2)
+	v, loaded := m.LoadOrStore("b", 0)
+	assert.True(t, loaded)
+	assert.Equal(t, 2, v)
+}
+
+func TestAtomicMap_Unwrap_BeforeUse(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	assert.Nil(t, m.Unwrap())
+}
+
+func TestAtomicMap_Unwrap_AfterStore(t *testing.T) {
+	var m AtomicMap[string, int]
+
+	m.Store("a", 1)
+	m.Store("b", 2)
+
+	raw := m.Unwrap()
+	assert.Equal(t, map[string]int{"a": 1, "b": 2}, raw)
+}
--- a/internal/containers/bag.go
+++ b/internal/containers/bag.go
@@ -54,6 +54,9 @@ func (b *Bag[T]) Add(v ...T) {
 
 // Seq atomically removes all values from the Bag and returns an
 // iterator that yields them in LIFO (most-recently-added-first) order.
+// The returned iterator is single-use: on early break it resumes from
+// the break point rather than restarting, and a fully consumed iterator
+// yields nothing on subsequent calls.
 func (b *Bag[T]) Seq() iter.Seq[T] {
 	head := b.head.Swap(nil)
 	return func(yield func(T) bool) {
--- a/internal/containers/bag_test.go
+++ b/internal/containers/bag_test.go
@@ -26,6 +26,48 @@ func TestBag(t *testing.T) {
 	require.Equal(t, 1000, count)
 }
 
+func TestBag_Clear(t *testing.T) {
+	var b Bag[int]
+	b.Add(1, 2, 3)
+	b.Clear()
+
+	var count int
+	for range b.Seq() {
+		count++
+	}
+	require.Equal(t, 0, count)
+}
+
+func TestBag_Add_Empty(t *testing.T) {
+	var b Bag[int]
+	b.Add(1)
+	b.Add() // no-op
+
+	var count int
+	for range b.Seq() {
+		count++
+	}
+	require.Equal(t, 1, count)
+}
+
+func TestBag_Seq_DrainsTheBag(t *testing.T) {
+	var b Bag[int]
+	b.Add(1, 2, 3)
+
+	var first int
+	for range b.Seq() {
+		first++
+	}
+	require.Equal(t, 3, first)
+
+	// Second Seq on the same bag should yield nothing because Seq drains.
+	var second int
+	for range b.Seq() {
+		second++
+	}
+	require.Equal(t, 0, second)
+}
+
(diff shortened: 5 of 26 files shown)


pkg/storage/cache_test.go
```diff
@@ -172,7 +172,7 @@ func TestWriteStruct(t *testing.T) {
 				"a": "x,'b:'y",
 			}),
 			// but our cache key should identify it correctly as 1 key
-			output: "1'a:'x,'b:'y,",
+			output: "1ax,'b:'y,",
```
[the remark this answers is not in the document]
justincoh: This is a potential security risk. I'm going to update the pr

pkg/storage/cache.go
```diff
@@ -342,7 +341,12 @@ func writeStruct(w io.StringWriter, s *structpb.Struct) (err error) {
 	}
 
 	for _, key := range keys {
-		if _, err = w.WriteString(fmt.Sprintf("'%s:'", key)); err != nil {
+		if _, err = w.WriteString(key); err != nil {
```
adriantam, line 361: Do we need to wrap the key with quotation?
i.e., 
```
		if _, err = w.WriteString("'" + key + "'"); err != nil {
```

justincoh: I think yes, we do. Copilot [flagged the same](https://github.com/openfga/openfga/pull/3006/changes#r2983663799) and it's a security issue

pkg/storage/cache_test.go
```diff
@@ -202,6 +202,46 @@ func TestWriteStruct(t *testing.T) {
...
+		"fields_write_quote_two_error": {
+			writer: &ErrorStringWriter{
+				TriggerAt: 5,
+			},
+			value: MustNewStruct(map[string]any{
+				"keyA": "valueA",
+				"keyB": "valueB",
+			}),
+			error: true,
+		},
```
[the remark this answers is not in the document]
justincoh: unfortunately this pattern is necessary to pass our code coverage quality gate

internal/utils/sanitize.go
```diff
@@ -0,0 +1,33 @@
...
+
+func Sanitize(s string) string {
+	for _, c := range s {
+		if unicode.IsControl(c) {
+			return replaceDangerousCharacters(s)
+		}
+	}
+	// Most of the time we can just return the original string and avoid any allocs
+	return s
+}
```
senojj: While I understand the intent, in the worst case scenario, this is far less efficient than just eating the allocation on each call.

senojj: An O(n) operation has become O(n+n) (worst case)

justincoh: That's true, but I'd expect the overwhelming majority of calls to _not_ contain `unicode.IsControl` characters, so I think having a `O(2n)` worst case scenario is ok.

Either way, I ran some benchmarks.
1. BenchmarkSanitize/with_control_chars
The original intention, `Sanitize` strips several `\u0008` characters
2. BenchmarkSanitize/clean_string
This implementation processes a totally clean string `"document:12345#viewer@user:alice"`
3. BenchmarkSanitize/clean_string_replace
The clean string is passed directly through `removeDangerousCharacters`, which will create a `[]byte` every time.

Option 3, where a clean string is run through the actual replacement method, is more than 2x slower than option 2 (which skips `replaceDangerousCharacters` entirely), and has a small memory footprint (B/op > 0 and allocs > 0) where the short circuit in `Sanitize` has none.


Here are the results:

```
goos: darwin
goarch: arm64
pkg: github.com/openfga/openfga/internal/utils
cpu: Apple M3 Max
BenchmarkSanitize/with_control_chars-4         	34144407	        34.24 ns/op	      24 B/op	       1 allocs/op
BenchmarkSanitize/clean_string-4               	66194719	        18.01 ns/op	       0 B/op	       0 allocs/op
BenchmarkSanitize/clean_string_replace-4       	25343389	        46.43 ns/op	      32 B/op	       1 allocs/op
```

justincoh: Here are the actual benchmarks in a collapse:

<details>
<summary>Benchmark Code</summary>

```
func BenchmarkSanitize(b *testing.B) {
	b.Run("with_control_chars", func(b *testing.B) {
		a := "hello\u0008 \u0008 \u0008 world\u0008"
		var x string
		for b.Loop() {
			x = Sanitize(a)
		}
		y = x
	})

	b.Run("clean_string", func(b *testing.B) {
		a := "document:12345#viewer@user:alice"
		var x string
		for b.Loop() {
			x = Sanitize(a)
		}
		y = x
	})

	b.Run("clean_string_replace", func(b *testing.B) {
		a := "document:12345#viewer@user:alice"
		var x string
		for b.Loop() {
			x = replaceDangerousCharacters(a)
		}
		y = x
	})
}
```
</details>

senojj: Can you also show a benchmark for the worst case scenario using only `replaceDangerousCharacters`?

senojj: (this means you would make the control character the last character in the string)

justincoh: Result:
```
BenchmarkSanitize/worst_case_with_sanitize-4         	14583778	        81.58 ns/op	      96 B/op	       2 allocs/op
BenchmarkSanitize/worst_case_replace_only-4          	19660809	        60.76 ns/op	      96 B/op	       2 allocs/op
```

Benchmark:
```
	b.Run("worst_case_with_sanitize", func(b *testing.B) {
		a := "document:12345#viewer@user:alice\u0008"
		var x string
		for b.Loop() {
			x = Sanitize(a)
		}
		y = x
	})

	b.Run("worst_case_replace_only", func(b *testing.B) {
		a := "document:12345#viewer@user:alice\u0008"
		var x string
		for b.Loop() {
			x = replaceDangerousCharacters(a)
		}
		y = x
	})
```

justincoh:
converting back to draft while I refactor

internal/utils/sanitize.go
```diff
@@ -0,0 +1,15 @@
...
+
+// ContainsControlChars reports whether s contains any Unicode control characters.
+func ContainsControlChars(s string) bool {
+	for _, c := range s {
+		if unicode.IsControl(c) {
+			return true
+		}
+	}
+	return false
+}
```
senojj: `strings.ContainsFunc(s, unicode.IsControl)`

internal/utils/sanitize.go
justincoh: Resolving this thread as outdated, this PR is no longer sanitizing in this manner

justincoh:
Refactor completed and benchmark results in the description have been updated

internal/utils/sanitize.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"strings"
+	"unicode"
+)
+
+// ContainsControlChars reports whether s contains any Unicode control characters.
+func ContainsControlChars(s string) bool {
+	return strings.ContainsFunc(s, unicode.IsControl)
+}
```
senojj: is this function useful as an abstraction? does it abstract the details of `strings.ContainsFunc(s, unicode.IsControl)`?

senojj: Might be more useful to start a function like: "ContainsForbiddenChars" that we can add to over time.

pkg/storage/cache.go
```diff
@@ -342,7 +345,17 @@ func writeStruct(w io.StringWriter, s *structpb.Struct) (err error) {
 	}
 
 	for _, key := range keys {
-		if _, err = w.WriteString(fmt.Sprintf("'%s:'", key)); err != nil {
+		if _, err = w.WriteString("'"); err != nil {
+			return
+		}
+		if err = validateNoControlChars(key); err != nil {
+			return
+		}
```
senojj: better to do this before writing any data.

internal/validation/validation.go
```diff
@@ -364,3 +380,44 @@ func ValidateUser(typesys *typesystem.TypeSystem, user string) error {
 
 	return nil
 }
+
+// ValidateStruct checks that a structpb.Struct does not contain Unicode control
```
DanCech: This comment and the name of the helper function should be updated to reflect that we're now checking for forbidden characters, not just control characters.

pkg/storage/cache.go
```diff
@@ -261,6 +261,16 @@ func GetReadCacheKey(store, tuple string) string {
 // an unexpected structpb.Value kind was encountered.
 var ErrUnexpectedStructValue = errors.New("unexpected structpb value encountered")
 
+// validateNoForbiddenChars returns an error if s contains characters OpenFGA doesn't allow.
+// These characters should be rejected at validation boundaries before reaching
+// cache key generation. If this errors, it indicates a validation bug upstream.
+func validateNoForbiddenChars(s string) error {
+	if utils.ContainsForbiddenChars(s) {
+		return fmt.Errorf("invariant violation: control character in cache key input")
```
DanCech: update error message

pkg/server/commands/errors.go
```diff
@@ -49,6 +49,18 @@ func (e *ThrottledError) Error() string {
...
+type InvalidContextError struct {
+	Cause error
+}
+
+func (e *InvalidContextError) Unwrap() error {
+	return e.Cause
+}
+
+func (e *InvalidContextError) Error() string {
+	return e.Unwrap().Error()
```
senojj, line 61: This call to `(*InvalidContextError) Unwrap()` couples `(*InvalidContextError) Error()` to `Unwrap()`. In reality, they are independent functions. It would be better to call `e.Cause.Error()` explicitly here instead.

Additionally, salting the internal error with context could be helpful:
```go
func (e *InvalidContextError) Error() string {
    return "invalid context: " + e.Cause.Error()
}
```

justincoh: That's a fair point. All of the errors in this file follow this pattern though; mind if I just update them all in a separate, small PR?

PR #3018, opened 2026-03-26
feat: reuse PostgreSQL container across tests

rafanaskin:

The `RunDatastoreTestContainer` storage fixture currently creates a new container on each invocation, which can result in more than 10 containers being started during a single test run per storage. Each container initialization involves startup and running migrations, increasing execution time and resource usage in CI. This added overhead can negatively impact overall test performance.

#### How is it being solved?

I started with `PostgreSQL` by updating its test fixture to reuse a single container across tests, while maintaining isolation through per-test databases. This significantly reduces the overhead of container creation and migration execution.

As a result, the execution time of `storage/postgres/postgres_test.go` decreased by ~30%:
- Before: `github.com/openfga/openfga/pkg/storage/postgres	159.219s`
- After:    `github.com/openfga/openfga/pkg/storage/postgres	108.411s`

> [!WARNING]
> The current implementation does not remove the `PostgreSQL` container after tests complete. Instead, it is cleaned up along with the `CI` worker. As an improvement, this cleanup step could be added to the `make test` command.

- Added an improved `Docker` client in `testutils`, used by the `PostgreSQL` fixture and ready for `MySQL` reuse.
- Updated the PostgreSQL test fixture to reuse a single container.

## References

followup https://github.com/openfga/openfga/pull/2912

## Review Checklist

## Summary by CodeRabbit

* **Chores**
  * Tests now reuse a shared PostgreSQL container and create per-test databases, speeding runs and reducing resource use.
* **Tests**
  * Added a Docker client wrapper with comprehensive container lifecycle tests and updated test harness to use a parallel-ready default config.
* **Refactor**
  * Improved database readiness, retry/backoff, replica sync, and migration/version wait logic to stabilize test startup.
* **Documentation**
  * CHANGELOG updated describing the testing infrastructure changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,7 @@ test: generate-mocks ## Run all tests. To run a specific test, pass the FILTER v
 			${GO_PACKAGES}
 	@cat coverageunit.tmp.out | grep -v "mock" > coverageunit.out
 	@rm coverageunit.tmp.out
+	@docker ps -q --filter "name=openfga-test-*" | xargs -r docker rm -f >/dev/null
 	
 test-docker: ## Run tests requiring Docker
 	${call print, "Running docker tests"}
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,6 @@ test: generate-mocks ## Run all tests. To run a specific test, pass the FILTER v
 			${GO_PACKAGES}
 	@cat coverageunit.tmp.out | grep -v "mock" > coverageunit.out
 	@rm coverageunit.tmp.out
-	@docker ps -q --filter "name=openfga-test-*" | xargs -r docker rm -f >/dev/null
 	
 test-docker: ## Run tests requiring Docker
 	${call print, "Running docker tests"}
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -240,8 +240,7 @@ func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
 }
 
 func configureDB(db *pgxpool.Pool, cfg *sqlcommon.Config, dbName string) (prometheus.Collector, error) {
-	policy := backoff.NewExponentialBackOff()
-	policy.MaxElapsedTime = 1 * time.Minute
+	policy := backoff.NewExponentialBackOff(backoff.WithMaxElapsedTime(1 * time.Minute))
 	attempt := 1
 	err := backoff.Retry(func() error {
 		err := db.Ping(context.Background())
--- a/pkg/testfixtures/storage/postgres.go
+++ b/pkg/testfixtures/storage/postgres.go
@@ -283,7 +283,7 @@ exec docker-entrypoint.sh postgres -c hot_standby=on -c max_connections=200
 		ExtraHosts:      []string{"host.docker.internal:host-gateway"},
 	}
 
-	contName := "secondary-" + postgresContainer
+	contName := postgresContainer + "-replica"
 	cont, err := docker.RunContainer(t.Context(), contCfg, hostCfg, contName)
 	require.NoError(t, err, "run postgres container")
 
--- a/pkg/testfixtures/storage/postgres.go
+++ b/pkg/testfixtures/storage/postgres.go
@@ -2,202 +2,72 @@ package storage
 
 import (
 	"context"
+	"database/sql"
 	"fmt"
-	"io"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/containerd/errdefs"
-	_ "github.com/jackc/pgx/v5/stdlib" // PostgreSQL driver.
-	"github.com/moby/moby/api/types/container"
-	"github.com/moby/moby/api/types/network"
-	"github.com/moby/moby/client"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/go-connections/nat"
+	_ "github.com/jackc/pgx/v5/stdlib"
 	"github.com/oklog/ulid/v2"
 	"github.com/pressly/goose/v3"
 	"github.com/stretchr/testify/require"
 
 	"github.com/openfga/openfga/assets"
+	"github.com/openfga/openfga/pkg/testutils"
 )
 
 const (
-	postgresImage = "postgres:17"
+	postgresImage           = "postgres:17-alpine"
+	postgresContainerPrefix = "openfga-test-postgres-"
+	postgresPort            = nat.Port("5432/tcp")
+
+	postgresDBPrefix   = "openfga-test-db-"
+	postgresTemplateDB = postgresDBPrefix + "template"
+	postgresUsername   = "postgres"
+	postgresPassword   = "secret"
 )
 
-var (
-	postgresPort = network.MustParsePort("5432/tcp")
+var _ DatastoreTestContainer = (*postgresTestContainer)(nil)
+
+type (
+	postgresTestContainer struct {
+		host string
+		port string
+
+		database string
+		username string
+		password string
+
+		version int64
+		replica *postgresReplicaContainer
+	}
+
+	postgresReplicaContainer struct {
+		host string
+		port string
+
+		username string
+		password string
+	}
 )
 
-type postgresTestContainer struct {
-	addr     string
-	version  int64
-	username string
-	password string
-	replica  *postgresReplicaContainer
-}
+// GetConnectionURI returns the postgres connection uri for the running postgres test container.
+func (p *postgresTestContainer) GetConnectionURI(includeCredentials bool) string {
+	var username, password string
+	if includeCredentials {
+		username = p.username
+		password = p.password
+	}
 
-type postgresReplicaContainer struct {
-	addr     string
-	username string
-	password string
-}
-
-// NewPostgresTestContainer returns an implementation of the DatastoreTestContainer interface
-// for Postgres.
-func NewPostgresTestContainer() *postgresTestContainer {
-	return &postgresTestContainer{}
+	return postgresConnectionURI(p.host, p.port, p.database, username, password)
 }
 
 func (p *postgresTestContainer) GetDatabaseSchemaVersion() int64 {
 	return p.version
-}
-
-// RunPostgresTestContainer runs a Postgres container, connects to it, and returns a
-// bootstrapped implementation of the DatastoreTestContainer interface wired up for the
-// Postgres datastore engine.
-func (p *postgresTestContainer) RunPostgresTestContainer(t testing.TB) DatastoreTestContainer {
-	dockerClient, err := client.New(client.FromEnv)
-	require.NoError(t, err)
-	t.Cleanup(func() {
-		dockerClient.Close()
-	})
-
-	imageListResult, err := dockerClient.ImageList(context.Background(), client.ImageListOptions{
-		All: true,
-	})
-	require.NoError(t, err)
-
-	foundPostgresImage := false
-
-AllImages:
-	for _, image := range imageListResult.Items {
-		for _, tag := range image.RepoTags {
-			if strings.Contains(tag, postgresImage) {
-				foundPostgresImage = true
-				break AllImages
-			}
-		}
-	}
-
-	if !foundPostgresImage {
-		t.Logf("Pulling image %s", postgresImage)
-		reader, err := dockerClient.ImagePull(context.Background(), postgresImage, client.ImagePullOptions{})
-		require.NoError(t, err)
-		defer reader.Close()
-
-		_, err = io.Copy(io.Discard, reader) // consume the image pull output to make sure it's done
-		require.NoError(t, err)
-	}
-
-	containerCfg := container.Config{
-		Env: []string{
-			"POSTGRES_DB=defaultdb",
-			"POSTGRES_PASSWORD=secret",
-		},
-		ExposedPorts: network.PortSet{
-			postgresPort: {},
-		},
-		Image: postgresImage,
-		Cmd: []string{
-			"postgres",
-			"-c", "wal_level=replica",
-			"-c", "max_wal_senders=3",
-			"-c", "max_replication_slots=3",
-			"-c", "wal_keep_size=64MB",
-			"-c", "hot_standby=on",
-		},
-	}
-
-	hostCfg := container.HostConfig{
-		AutoRemove:      true,
-		PublishAllPorts: true,
-		ExtraHosts:      []string{"host.docker.internal:host-gateway"},
-	}
-
-	name := "postgres-" + ulid.Make().String()
-
-	cont, err := dockerClient.ContainerCreate(context.Background(), client.ContainerCreateOptions{
-		Name:       name,
-		Config:     &containerCfg,
-		HostConfig: &hostCfg,
-	})
-	require.NoError(t, err, "failed to create postgres docker container")
-
-	t.Cleanup(func() {
-		t.Logf("stopping container %s", name)
-		timeoutSec := 5
-
-		_, err := dockerClient.ContainerStop(context.Background(), cont.ID, client.ContainerStopOptions{Timeout: &timeoutSec})
-		if err != nil && !errdefs.IsNotFound(err) {
-			t.Logf("failed to stop postgres container: %v", err)
-		}
-
-		t.Logf("stopped container %s", name)
-	})
-
-	_, err = dockerClient.ContainerStart(context.Background(), cont.ID, client.ContainerStartOptions{})
-	require.NoError(t, err, "failed to start postgres container")
-
-	inspectResult, err := dockerClient.ContainerInspect(context.Background(), cont.ID, client.ContainerInspectOptions{})
-	require.NoError(t, err)
-
-	m, ok := inspectResult.Container.NetworkSettings.Ports[postgresPort]
-	if !ok || len(m) == 0 {
-		require.Fail(t, "failed to get host port mapping from postgres container")
-	}
-
-	pgTestContainer := &postgresTestContainer{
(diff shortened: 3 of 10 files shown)


pkg/testfixtures/storage/readiness.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"database/sql"
+	"fmt"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+)
+
+// waitForDatabase attempts to establish a connection to the database and ping it until it's ready or a timeout occurs.
+func waitForDatabase(driverName, uri string) error { //nolint:unparam
```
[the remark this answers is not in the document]
rafanaskin: The linter reports that driverName in `waitForDatabase` always receives "pgx" (unparam).
This will be addressed once `MySQL` support is introduced, allowing removal of the nolint.

pkg/testfixtures/storage/postgres.go
```diff
@@ -2,97 +2,182 @@ package storage
...
-		All: true,
-	})
-	require.NoError(t, err)
+	dockerCont, found, err := docker.FindRunningContainer(t.Context(), "^"+postgresContainerPrefix, postgresImage)
+	require.NoError(t, err, "find running postgres container")
 
-	foundPostgresImage := false
+	if !found {
+		dockerCont = bootstrapPostgresContainer(t, docker)
+	}
```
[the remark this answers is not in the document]
rafanaskin: Yes, this is possible, but it's rare and won't meaningfully affect CI resources. The logic remains correct, with just one additional container created.

rafanaskin:
@coderabbitai review

pkg/testutils/dockerclient_test.go
```diff
@@ -0,0 +1,406 @@
...
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
```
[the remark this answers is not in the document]
rafanaskin: This vulnerability is not related to the current PR and will be addressed separately.

pkg/testutils/dockerclient.go
```diff
@@ -0,0 +1,206 @@
...
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/image"
+	"github.com/docker/docker/client"
+	"github.com/docker/go-connections/nat"
```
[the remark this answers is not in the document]
rafanaskin: This vulnerability is not related to the current PR and will be addressed separately.

pkg/testutils/dockerclient_test.go
```diff
@@ -0,0 +1,406 @@
...
+		if step.Error != "" {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusInternalServerError)
+			_ = json.NewEncoder(w).Encode(map[string]string{
+				"message": step.Error,
+			})
+
+			return
+		}
+	})
```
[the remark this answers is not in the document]
rafanaskin: Docker client tests send requests one by one, so there is no data race. However, to prevent possible issues in the future, the `atomic.Int32` `stepIndex` was used.

rafanaskin:
@coderabbitai review

pkg/testfixtures/storage/postgres.go
```diff
@@ -262,123 +263,60 @@ echo "Master ready, creating base backup..."
...
+pg_basebackup -h %s -p %s -U postgres -D $PGDATA -Fp -Xs -P -R -c fast
 
 # Configure as replica
 echo "hot_standby = on" >> $PGDATA/postgresql.conf
 touch $PGDATA/standby.signal
 
 echo "Starting PostgreSQL replica..."
-exec docker-entrypoint.sh postgres -c hot_standby=on -c wal_level=replica
-`, masterHost, masterPort, masterHost, masterPort)},
+exec docker-entrypoint.sh postgres -c hot_standby=on -c max_connections=200
```
DanCech, line 276: I see you added max_connections, why are we removing wal_level?

DanCech (commented):
Looks good to me

pkg/testfixtures/storage/postgres.go
rafanaskin: Hi @DanCech,

I increased `max_connections` since the tests now share a single `PostgreSQL` instance, which could otherwise run out of available connections.

I also removed `wal_level` from the replica config. As I understand it, this setting only matters on the primary, which generates `WAL`. The replica just replays `WAL` records and doesn't produce its own for replication, so `wal_level` there isn't really needed.

DanCech:
Looks like you can merge in main changes and take this out of draft, then we can get it approved and merged :)

senojj:
@coderabbitai resume

pkg/testfixtures/storage/postgres.go
```diff
@@ -2,251 +2,252 @@ package storage
...
 	})
 
-	allImages, err := dockerClient.ImageList(context.Background(), image.ListOptions{
-		All: true,
-	})
+	dockerCont, found, err := docker.FindRunningContainer(t.Context(), "^"+postgresContainerPrefix, postgresImage)
+	require.NoError(t, err, "find running postgres container")
+
+	if !found {
+		dockerCont = bootstrapPostgresContainer(t, docker)
```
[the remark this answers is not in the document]
rafanaskin: That's a good suggestion, and I'll consider implementing this behavior in the next version.
Manual deletion seems like a reasonable tradeoff for now.

pkg/testutils/dockerclient.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, fmt.Errorf("create docker client: %w", err)
+	}
+
+	return &DockerClient{client: cli}, nil
+}
+
+// Close calls the underlying Docker client's Close method, releasing its transport resources.
+func (d *DockerClient) Close() error {
+	return d.client.Close()
+}
```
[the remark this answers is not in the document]
senojj: This is unnecessary.

pkg/testfixtures/storage/storage.go
```diff
@@ -71,3 +83,16 @@ func RunDatastoreTestContainer(t testing.TB, engine string) DatastoreTestContain
 		return nil
 	}
 }
+
+// latestMigrationVersion returns the latest goose migration version in migrationDir.
+func latestMigrationVersion(migrationDir string) (int64, error) {
+	migrations, err := goose.CollectMigrations(migrationDir, 0, goose.MaxVersion)
+	if err != nil {
+		return 0, err
+	}
```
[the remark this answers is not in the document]
senojj: This is unnecessary

pkg/testfixtures/storage/readiness.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	defer db.Close()
+
+	backoffPolicy := backoff.NewExponentialBackOff(
+		backoff.WithInitialInterval(10*time.Millisecond),
+		backoff.WithMaxElapsedTime(30*time.Second),
+	)
+
+	if err := backoff.Retry(db.Ping, backoffPolicy); err != nil {
+		return fmt.Errorf("ping %s database: %w", driverName, err)
```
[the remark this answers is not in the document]
rafanaskin: Sorted.

pkg/storage/postgres/postgres.go
```diff
@@ -240,8 +240,7 @@ func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
 }
 
 func configureDB(db *pgxpool.Pool, cfg *sqlcommon.Config, dbName string) (prometheus.Collector, error) {
-	policy := backoff.NewExponentialBackOff()
-	policy.MaxElapsedTime = 1 * time.Minute
+	policy := backoff.NewExponentialBackOff(backoff.WithMaxElapsedTime(1 * time.Minute))
 	attempt := 1
 	err := backoff.Retry(func() error {
 		err := db.Ping(context.Background())
```
DanCech, line 246: We should probably use the same strategy here to limit the ping timeout

pkg/testfixtures/storage/postgres.go
```diff
@@ -2,251 +2,252 @@ package storage
...
-		if err != nil && !errdefs.IsNotFound(err) {
-			t.Logf("failed to stop postgres container: %v", err)
-		}
-
-		t.Logf("stopped container %s", name)
-	})
-
-	err = dockerClient.ContainerStart(context.Background(), cont.ID, container.StartOptions{})
-	require.NoError(t, err, "failed to start postgres container")
+	containerName := postgresContainerPrefix + ulid.Make().String()
```
DanCech: Do we actually need to do this or can we just set a fixed name? If there are any existing containers that have a name starting with the prefix we're already going to run into issues, so I doubt the ulid is doing anything for us. We needed it in the past to be able to run tests simultaneously but now they'll just use separate databases in the same container.

pkg/storage/postgres/postgres.go
rafanaskin: I checked the [implementation](https://github.com/jackc/pgx/blob/master/pgxpool/pool.go#L175), and `pgxpool` has a dedicated `PingTimeout` configuration that is used by `Pool.Acquire`. Other storages (`MySQL` and `SQLite`) also use a context without a timeout.

So I'm considering introducing a new setting, `DatastoreConfig.PingTimeout`, in a separate PR.
What do you think?

DanCech: Fair enough, let's look at that in a separate PR.

pkg/testfixtures/storage/postgres.go
rafanaskin: Using unique container names acts as a safety guard in cases where two tests simultaneously determine that no container exists and both attempt to create one.

With unique names, this results in duplicate containers being created, which is not ideal but only wastes CI resources - the tests themselves continue to pass.

With a fixed container name, however, `Docker` will return a `Conflict` error when the second test tries to create the container, causing test failures.

Some discussion with AI is [here](https://github.com/openfga/openfga/pull/3018#discussion_r3025283447).

DanCech: The only concern I have with that approach is for local development, where the situation you describe would result in 2 containers running indefinitely. We can resolve the race condition by returning the existing container if the create fails with a Conflict error.

rafanaskin: Yes, that will work as well - I'll make the changes.

senojj: It would also be possible to stand up, and tear down, a container per package using the package-level TestMain function. While this solution would not result in only a single container when running the full test suite, it would still reduce the number of database containers to only what is absolutely necessary, while allowing us to tear them down appropriately after the tests complete.

Given that we will be changing our tests to run in different groups, this may be the appropriate path forward.

senojj: I'm thinking something along the lines of this within a package like `pkg/storage/postgres`:
```go
var testContainer storage.DatastoreTestContainer

func setup() {
    testContainer = storage.RunDatastoreTestContainer(t, "postgres")
}

func teardown() {
    testContainer.Stop()
}

func TestMain(m *testing.M) {
    setup()
    exitVal := m.Run()
    teardown()
    os.Exit(exitVal)
}
```

rafanaskin: I've implemented the use of a static container name.

rafanaskin: > It would also be possible to stand up, and tear down, a container per package

Yes, that approach would work, but there's one caveat.

If we use a static container name together with setup/teardown in `TestMain`, we won't be able to safely run multiple packages in parallel (e.g. via `go test ./...`). Different packages could try to create or remove the same container at the same time, leading to conflicts.

Maybe I'm misunderstanding your idea, but in that case we'd still need unique container names (per package) to keep parallel execution reliable, at least in a local environment. In CI, we could avoid this by isolating execution across separate test jobs.

DanCech: Yes, that approach would require going back to the dynamic container name (and creating a separate container per package), but would mean that we tore down the containers vs leaving them running after the tests complete. As far as I know there isn't a great way to know when all the tests are complete to be able to clean up that single container after a test run, golang only provides the package-level TestMain functionality.

Technically the single-container approach will give the best local performance since we only have to bootstrap a single container for each database, and it would get reused across multiple test runs. The approach Josh is suggesting would mean we have better isolation between packages because each package would use its own container (and clean it up afterwards).

I don't have strong feelings either way, and both will be an improvement over the current arrangement.

senojj: > If we use a static container name together with setup/teardown in TestMain, we won't be able to safely run multiple packages in parallel (e.g. via go test ./...). Different packages could try to create or remove the same container at the same time, leading to conflicts.

That is correct. What I am advocating for is keeping the dynamic name, and cleaning containers up properly after all tests depending on that container have finished. These tests will often be run locally, and leaving even a single container running after the tests complete is not ideal.

By keeping the containers per package, we still mitigate/minimize the number of containers running at one time during integration tests. There are really only a few packages that use the containers during their tests.

rafanaskin: @senojj, @DanCech thanks for the input.

I agree that both approaches are improvements over the current situation. I also agree that developer experience matters, and leaving the container running isn't ideal.

So let's consider the available options.

## Option 1
Reuse storage containers at the package level and clean them up via `TestMain`.

### Pros
- Uses a Go-native approach for managing test resources, fully compatible with the `go test` command.
- Less aggressive use of storage resources (e.g., connections), since each package has its own container.

### Cons
- Running tests across 10+ packages will result in multiple containers being started.
- Each package needs its own cleanup logic in `main_test.go`.
- Likely slower compared to Option 2.

## Option 2
Reuse a storage container at the global level and clean it up via `make test`.

### Pros
- Uses a single container per storage across all tests.
- More efficient use of CI worker resources (e.g., CPU).
- Faster compared to Option 1.
- Faster local test reruns when using the `go test` command.
- No need for additional setup in each new package that uses containers.

### Cons
- Containers will remain running after tests complete if they are not executed via `make test`.

I'd appreciate your feedback on these options - anything to skip, add, or adjust.

senojj: > Running tests across 10+ packages will result in multiple containers being started.

@rafanaskin to my knowledge we only have a few packages that actually run these containers.

Also note that we are breaking up packages into separate CI jobs within GitHub.

rafanaskin: > to my knowledge we only have a few packages that actually run these containers.

@senojj 
The following packages use `RunDatastoreTestContainer` or functions that depend on it.
```
1. cmd/run
2. cmd/validatemodels
3. pkg/server
4. pkg/storage/migrate
5. pkg/storage/mysql
6. pkg/storage/postgres
7. tests
8. tests/authzen
9. tests/check
10. tests/listobjects
11. tests/listusers
```

> Also note that we are breaking up packages into separate CI jobs within GitHub.

This should definitely make the tests more stable.

rafanaskin: @senojj, @DanCech 
If we're aligned on `Option 2`, the PR is ready to merge.
If you'd prefer `Option 1`, I can update it accordingly.

Just a note: `TestMain` doesn't apply to nested packages, so, for example, `tests` and `tests/authzen` need separate `main_test.go` files.

senojj: Correct. I think the short-term plan will be to get this merged as-is. We can follow-up with a breakout by package if that is determined to be a more desirable approach.

Makefile
```diff
@@ -96,6 +96,7 @@ test: generate-mocks ## Run all tests. To run a specific test, pass the FILTER v
 			${GO_PACKAGES}
 	@cat coverageunit.tmp.out | grep -v "mock" > coverageunit.out
 	@rm coverageunit.tmp.out
+	@docker ps -q --filter "name=openfga-test-*" | xargs -r docker rm -f >/dev/null
```
senojj: I'm not sure this is a great practice, even though the intent is desirable. 

I wouldn't want to cull unrelated running containers that happen to match this prefix on a dev/user's machine.

rafanaskin: Good point. I've reverted the changes.

senojj (dismissed):
I approve these changes, but would also like for @DanCech to sign off before merging.

pkg/testfixtures/storage/postgres.go
```diff
@@ -264,128 +266,65 @@ echo "Master ready, creating base backup..."
...
 
-	name := "postgres-replica-" + ulid.Make().String()
-
-	cont, err := dockerClient.ContainerCreate(context.Background(), client.ContainerCreateOptions{
-		Name:       name,
-		Config:     &containerCfg,
-		HostConfig: &hostCfg,
-	})
-	require.NoError(t, err, "failed to create postgres replica docker container")
+	contName := "secondary-" + postgresContainer
```
DanCech: let's switch this to use secondary or replica as a suffix instead of prefix

rafanaskin: done

senojj (approved):

DanCech (approved):
LGTM!

pkg/storage/postgres/postgres.go
rafanaskin: Hi @DanCech, I've started to work on this.

rafanaskin: https://github.com/openfga/openfga/pull/3113

PR #3042, opened 2026-04-04
feat: reuse MySQL container across tests

rafanaskin:

The `RunDatastoreTestContainer` storage fixture currently creates a new container on each invocation, which can result in more than 10 containers being started during a single test run per storage. Each container initialization involves startup and running migrations, increasing execution time and resource usage in CI. This added overhead can negatively impact overall test performance.

#### How is it being solved?

I started with [PostgreSQL](https://github.com/openfga/openfga/pull/3018) and in this PR continue with MySQL.
The approach the same - reusing a single container across tests, while maintaining isolation through per-test databases. For MySQL this much more significantly reduces the overhead of container creation and migration execution.

This PR continues the work started in #3018 and #3062￼, now applied to MySQL.

The approach is the same: we reuse a single MySQL container across tests while maintaining isolation via per-test databases. For MySQL, this change significantly reduces the overhead of container creation and migration execution.

As a result, the execution time for `storage/mysql/mysql_test.go` is reduced by ~3×:
- Before: `github.com/openfga/openfga/pkg/storage/mysql	226.366s`
- After:    `github.com/openfga/openfga/pkg/storage/mysql	71.786s`

Also was reduces execution time of SQLite storage tests by ~2x. It was achieved by adding the `idx_store_ulid` index.

- Before: `github.com/openfga/openfga/pkg/storage/sqlite	315.027s`
- After:    `github.com/openfga/openfga/pkg/storage/sqlite	150.331s`

- Updated the `MySQL` test fixture to reuse a single container.
- Simplified `MySQL` and `Postgres` test fixtures by introducing `sync.Once` for shared container lifecycle management, which also simplifies `Docker` client.
- Improved SQLite test performance by adding an index, reducing execution time by 2x.

## References

followup https://github.com/openfga/openfga/pull/3018
followup https://github.com/openfga/openfga/pull/3062

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Test suite now reuses shared MySQL and PostgreSQL containers with isolated per-test databases, using coordinated bootstrapping, stronger readiness/backoff, and improved container startup checks; Docker client utilities and test coverage for container readiness and command execution were strengthened.

* **Chores**
  * Test harnesses now perform explicit shared-container cleanup during teardown.

* **New Features**
  * Added a DB migration to add an index on (store, ulid) to improve query performance.

### DIFF
--- a/assets/migrations/sqlite/006_add_store_ulid_index.sql
+++ b/assets/migrations/sqlite/006_add_store_ulid_index.sql
@@ -1,0 +1,5 @@
+-- +goose Up
+CREATE INDEX idx_store_ulid ON tuple (store, ulid);
+
+-- +goose Down
+DROP INDEX IF EXISTS idx_store_ulid;
--- a/pkg/server/testmain_test.go
+++ b/pkg/server/testmain_test.go
@@ -10,5 +10,6 @@ import (
 func TestMain(m *testing.M) {
 	code := m.Run()
 	storagefixtures.CleanupPostgresContainer()
+	storagefixtures.CleanupMysqlContainer()
 	os.Exit(code)
 }
--- a/pkg/testfixtures/storage/mysql.go
+++ b/pkg/testfixtures/storage/mysql.go
@@ -118,9 +118,10 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 
 		mysqlCond.Wait()
 	}
+	dockerCont := mysqlDockerCont
 	mysqlCond.L.Unlock()
 
-	port, err := docker.GetHostPort(mysqlDockerCont, mysqlPort)
+	port, err := docker.GetHostPort(dockerCont, mysqlPort)
 	require.NoError(t, err)
 
 	version, err := latestMigrationVersion(assets.MySQLMigrationDir)
@@ -142,7 +143,7 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 		Cmd: []string{"mysql", "-u", testCont.username, "-e", fmt.Sprintf("CREATE DATABASE `%s`;", testCont.database)},
 		Env: []string{"MYSQL_PWD=" + testCont.password},
 	}
-	require.NoError(t, docker.ExecCommand(t.Context(), mysqlDockerCont.ID, createExec))
+	require.NoError(t, docker.ExecCommand(t.Context(), dockerCont.ID, createExec))
 
 	t.Cleanup(func() {
 		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
@@ -153,7 +154,7 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 			Cmd: []string{"mysql", "-u", testCont.username, "-e", dropQuery},
 			Env: []string{"MYSQL_PWD=" + testCont.password},
 		}
-		if err := docker.ExecCommand(ctx, mysqlDockerCont.ID, dropExec); err != nil {
+		if err := docker.ExecCommand(ctx, dockerCont.ID, dropExec); err != nil {
 			t.Errorf("drop test database in the mysql container: %v", err)
 		}
 	})
@@ -163,13 +164,15 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 		Cmd: []string{"sh", "-ec", copyShell},
 		Env: []string{"MYSQL_PWD=" + testCont.password},
 	}
-	require.NoError(t, docker.ExecCommand(t.Context(), mysqlDockerCont.ID, copyExec))
+	require.NoError(t, docker.ExecCommand(t.Context(), dockerCont.ID, copyExec))
 
 	require.NoError(t, waitForMigrationVersion("mysql", testCont.GetConnectionURI(true), testCont.version))
 
 	return testCont
 }
 
+// CleanupMysqlContainer removes the shared mysql test container.
+// It should be called from TestMain after all tests in a package have finished.
 func CleanupMysqlContainer() {
 	_ = cleanupDatastoreTestContainer(mysqlContainerName)
 }
--- a/pkg/testfixtures/storage/mysql.go
+++ b/pkg/testfixtures/storage/mysql.go
@@ -34,7 +34,9 @@ var (
 	mysqlContainerName = "openfga-test-mysql-" + ulid.Make().String()
 	mysqlPort          = network.MustParsePort("3306/tcp")
 	mysqlDockerCont    *container.InspectResponse
-	mysqlOnce          sync.Once
+
+	mysqlBootstrapping bool
+	mysqlCond          = sync.NewCond(&sync.Mutex{})
 )
 
 type mysqlTestContainer struct {
@@ -87,9 +89,36 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 		docker.Close()
 	})
 
-	mysqlOnce.Do(func() {
-		mysqlDockerCont = bootstrapMysqlContainer(t, docker)
-	})
+	// Shared MySQL container bootstrap for concurrent tests using sync.Cond.
+	// Only one test bootstraps the shared container at a time, while others wait efficiently using sync.Cond.
+	// If bootstrap fails, waiting tests are awakened so another test can retry without being affected by the failure.
+	mysqlCond.L.Lock()
+	for mysqlDockerCont == nil {
+		if !mysqlBootstrapping {
+			mysqlBootstrapping = true
+			mysqlCond.L.Unlock()
+
+			dockerCont, err := bootstrapMysqlContainer(t.Context(), docker)
+			mysqlCond.L.Lock()
+			mysqlBootstrapping = false
+			if err == nil {
+				mysqlDockerCont = dockerCont
+			}
+
+			mysqlCond.Broadcast()
+
+			if err != nil {
+				// Unlock before failing the test to allow waiting tests to proceed with bootstrapping.
+				mysqlCond.L.Unlock()
+				require.NoError(t, err)
+			}
+
+			continue
+		}
+
+		mysqlCond.Wait()
+	}
+	mysqlCond.L.Unlock()
 
 	port, err := docker.GetHostPort(mysqlDockerCont, mysqlPort)
 	require.NoError(t, err)
@@ -145,10 +174,10 @@ func CleanupMysqlContainer() {
 	_ = cleanupDatastoreTestContainer(mysqlContainerName)
 }
 
-func bootstrapMysqlContainer(t testing.TB, docker *testutils.DockerClient) *container.InspectResponse {
-	t.Logf("No running container found for %s, creating a new one", mysqlImage)
-
-	require.NoError(t, docker.PullImage(t.Context(), mysqlImage), "pull mysql image")
+func bootstrapMysqlContainer(ctx context.Context, docker *testutils.DockerClient) (*container.InspectResponse, error) {
+	if err := docker.PullImage(ctx, mysqlImage); err != nil {
+		return nil, fmt.Errorf("pull mysql image: %w", err)
+	}
 
 	contCfg := &container.Config{
 		Env: []string{
@@ -167,20 +196,40 @@ func bootstrapMysqlContainer(t testing.TB, docker *testutils.DockerClient) *cont
 		Tmpfs:           map[string]string{"/var/lib/mysql": "", "/tmp": ""},
 	}
 
-	cont, err := docker.RunContainer(t.Context(), contCfg, hostCfg, mysqlContainerName)
-	require.NoError(t, err, "run mysql container")
+	cont, err := docker.RunContainer(ctx, contCfg, hostCfg, mysqlContainerName)
+	if err != nil {
+		return nil, fmt.Errorf("run mysql container: %w", err)
+	}
+
+	needsCleanup := true
+	defer func() {
+		if needsCleanup {
+			cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
+
+			_ = docker.RemoveContainer(cleanupCtx, cont.ID)
+		}
+	}()
 
 	port, err := docker.GetHostPort(cont, mysqlPort)
-	require.NoError(t, err)
+	if err != nil {
+		return nil, fmt.Errorf("get mysql host port: %w", err)
+	}
 
 	dbURI := mysqlConnectionURI("localhost", port, mysqlTemplateDB, mysqlUsername, mysqlPassword)
-	require.NoError(t, waitForDatabase("mysql", dbURI))
+	if err := waitForDatabase("mysql", dbURI); err != nil {
+		return nil, fmt.Errorf("wait for mysql database: %w", err)
+	}
 
 	db, err := goose.OpenDBWithDriver("mysql", dbURI)
-	require.NoError(t, err)
+	if err != nil {
+		return nil, fmt.Errorf("open mysql database: %w", err)
+	}
 	defer db.Close()
 
-	require.NoError(t, goose.Up(db, assets.MySQLMigrationDir))
+	if err := goose.Up(db, assets.MySQLMigrationDir); err != nil {
+		return nil, fmt.Errorf("apply mysql migrations: %w", err)
+	}
 
 	// Append goose_db_version last so it is restored after all schema tables.
 	dumpShell := fmt.Sprintf(
@@ -192,9 +241,12 @@ func bootstrapMysqlContainer(t testing.TB, docker *testutils.DockerClient) *cont
 		Cmd: []string{"sh", "-ec", dumpShell},
 		Env: []string{"MYSQL_PWD=" + mysqlPassword},
 	}
-	require.NoError(t, docker.ExecCommand(t.Context(), cont.ID, dumpExec))
+	if err := docker.ExecCommand(ctx, cont.ID, dumpExec); err != nil {
+		return nil, fmt.Errorf("dump mysql template database: %w", err)
+	}
 
-	return cont
+	needsCleanup = false
+	return cont, nil
 }
 
 func mysqlConnectionURI(host, port, database, username, password string) string {
--- a/pkg/testfixtures/storage/mysql.go
+++ b/pkg/testfixtures/storage/mysql.go
@@ -193,7 +193,10 @@ func bootstrapMysqlContainer(ctx context.Context, docker *testutils.DockerClient
 	hostCfg := &container.HostConfig{
 		AutoRemove:      true,
 		PublishAllPorts: true,
-		Tmpfs:           map[string]string{"/var/lib/mysql": "", "/tmp": ""},
+		Tmpfs: map[string]string{
+			"/var/lib/mysql": "size=2g",
+			"/tmp":           "size=512m",
+		},
 	}
 
 	cont, err := docker.RunContainer(ctx, contCfg, hostCfg, mysqlContainerName)
--- a/pkg/testfixtures/storage/mysql.go
+++ b/pkg/testfixtures/storage/mysql.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -33,7 +34,7 @@ var (
 
 	mysqlContainerName = "openfga-test-mysql-" + ulid.Make().String()
 	mysqlPort          = network.MustParsePort("3306/tcp")
-	mysqlDockerCont    *container.InspectResponse
+	mysqlDockerCont    atomic.Pointer[container.InspectResponse]
 
 	mysqlBootstrapping bool
 	mysqlCond          = sync.NewCond(&sync.Mutex{})
@@ -93,7 +94,7 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 	// Only one test bootstraps the shared container at a time, while others wait efficiently using sync.Cond.
 	// If bootstrap fails, waiting tests are awakened so another test can retry without being affected by the failure.
 	mysqlCond.L.Lock()
-	for mysqlDockerCont == nil {
+	for mysqlDockerCont.Load() == nil {
 		if !mysqlBootstrapping {
 			mysqlBootstrapping = true
 			mysqlCond.L.Unlock()
@@ -102,7 +103,7 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 			mysqlCond.L.Lock()
 			mysqlBootstrapping = false
 			if err == nil {
-				mysqlDockerCont = dockerCont
+				mysqlDockerCont.Store(dockerCont)
 			}
 
 			mysqlCond.Broadcast()
@@ -118,9 +119,9 @@ func RunMysqlTestContainer(t testing.TB) DatastoreTestContainer {
 
 		mysqlCond.Wait()
 	}
-	dockerCont := mysqlDockerCont
 	mysqlCond.L.Unlock()
 
+	dockerCont := mysqlDockerCont.Load()
 	port, err := docker.GetHostPort(dockerCont, mysqlPort)
 	require.NoError(t, err)
 
--- a/pkg/testfixtures/storage/mysql.go
+++ b/pkg/testfixtures/storage/mysql.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/cenkalti/backoff/v4"
 	_ "github.com/go-sql-driver/mysql"
 	"github.com/moby/moby/api/types/container"
 	"github.com/moby/moby/api/types/network"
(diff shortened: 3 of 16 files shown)


pkg/testutils/dockerclient.go
```diff
@@ -0,0 +1,206 @@
...
+	"context"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/image"
+	"github.com/docker/docker/client"
```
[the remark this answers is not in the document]
rafanaskin: The latest Docker Go module version ([github.com/docker/docker v28.5.2+incompatible](https://pkg.go.dev/github.com/docker/docker@v28.5.2+incompatible)￼) still contains a known vulnerability.

pkg/testfixtures/storage/mysql.go
```diff
@@ -2,192 +2,205 @@ package storage
...
 	require.NoError(t, err)
-	t.Cleanup(func() {
-		dockerClient.Close()
-	})
 
-	allImages, err := dockerClient.ImageList(context.Background(), image.ListOptions{
-		All: true,
+	goose.SetLogger(goose.NopLogger())
+	goose.SetBaseFS(assets.EmbedMigrations)
+
```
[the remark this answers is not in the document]
rafanaskin: Move the modification of Goose global state into `init()`

pkg/testfixtures/storage/postgres.go
```diff
@@ -2,251 +2,268 @@ package storage
...
-			}
+	require.Eventually(t, func() bool {
+		version, err := goose.GetDBVersion(db)
+		if err != nil {
+			return false
 		}
-	}
+		testCont.version = version
+		return version == expVersion
+	}, 1*time.Minute, 1*time.Second, "wait for mysql database to be ready")
```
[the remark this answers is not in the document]
rafanaskin: > Two parallel callers can therefore create different "shared" containers

It's ok


> stale container can clone postgresTemplateDB before it has the current migrations

Fixed

pkg/testfixtures/storage/postgres.go
```diff
@@ -262,123 +263,60 @@ echo "Master ready, creating base backup..."
...
+pg_basebackup -h %s -p %s -U postgres -D $PGDATA -Fp -Xs -P -R -c fast
 
 # Configure as replica
 echo "hot_standby = on" >> $PGDATA/postgresql.conf
 touch $PGDATA/standby.signal
 
 echo "Starting PostgreSQL replica..."
-exec docker-entrypoint.sh postgres -c hot_standby=on -c wal_level=replica
-`, masterHost, masterPort, masterHost, masterPort)},
+exec docker-entrypoint.sh postgres -c hot_standby=on -c max_connections=200
```
DanCech, line 337: what's the reason for this change?

DanCech (commented):
This change looks great, just need to get #3018 merged first so the diffs are clean.

pkg/testfixtures/storage/postgres.go
rafanaskin: https://github.com/openfga/openfga/pull/3018#discussion_r3042015898

> I increased max_connections since the tests now share a single PostgreSQL instance, which could otherwise run out of available connections.

> I also removed wal_level from the replica config. As I understand it, this setting only matters on the primary, which generates WAL. The replica just replays WAL records and doesn't produce its own for replication, so wal_level there isn't really needed.

rafanaskin:
@coderabbitai review

senojj:
Failing benchmark looks like an issue with recreation logic not running migrations:
```
--- FAIL: BenchmarkOpenFGAServer/BenchmarkMySQLDatastore
    mysql.go:149: No running container found for mysql:8, creating a new one
    mysql.go:143: 
        	Error Trace:	/home/runner/work/openfga/openfga/pkg/testfixtures/storage/mysql.go:143
        	            				/home/runner/work/openfga/openfga/pkg/testfixtures/storage/storage.go:74
        	            				/home/runner/work/openfga/openfga/pkg/server/server_test.go:792
        	            				/opt/hostedtoolcache/go/1.26.1/x64/src/testing/benchmark.go:219
        	            				/opt/hostedtoolcache/go/1.26.1/x64/src/testing/benchmark.go:245
        	            				/opt/hostedtoolcache/go/1.26.1/x64/src/runtime/asm_amd64.s:1771
        	Error:      	Received unexpected error:
        	            	mysql migration version: database version 0 does not match expected version 7
        	Test:       	BenchmarkOpenFGAServer/BenchmarkMySQLDatastore
```

pkg/testutils/dockerclient.go
```diff
@@ -102,42 +108,49 @@ func (d *DockerClient) RunContainer(
...
 		inspectResult, err := d.client.ContainerInspect(ctx, containerName, client.ContainerInspectOptions{})
 		if err != nil {
+			if errdefs.IsNotFound(err) {
+				return err
+			}
+
 			return backoff.Permanent(fmt.Errorf("inspect %s container: %w", containerName, err))
 		}
-		cont = inspectResult.Container
+		runResult.Container = &inspectResult.Container
```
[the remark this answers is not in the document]
rafanaskin: Good suggestion!

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +15,220 @@ import (
...
+		dropQuery := fmt.Sprintf("DROP DATABASE IF EXISTS `%s`;", testCont.database)
+		dropExec := client.ExecCreateOptions{
+			Cmd: []string{"mysql", "-u", testCont.username, "-e", dropQuery},
+			Env: []string{"MYSQL_PWD=" + testCont.password},
+		}
+		if err := docker.ExecCommand(ctx, dockerCont.ID, dropExec); err != nil {
+			t.Errorf("drop test database in the mysql container: %v", err)
+		}
+	})
```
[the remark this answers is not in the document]
rafanaskin: Sorted

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +15,220 @@ import (
...
-
-	err = mysql.SetLogger(log.New(io.Discard, "", 0))
-	require.NoError(t, err)
-
-	goose.SetLogger(goose.NopLogger())
+	// Move dump to a stable path after the full dump completes successfully.
+	renameExec := client.ExecCreateOptions{
+		Cmd: []string{"mv", mysqlTemplateDBPartialDump, mysqlTemplateDBDump},
+	}
+	require.NoError(t, docker.ExecCommand(t.Context(), runResult.Container.ID, renameExec))
```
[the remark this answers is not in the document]
rafanaskin: In this case, the tests will fail with a clear error indicating an invalid migration version, which should be easy to identify and fix. For this type of error, it is better to keep the container running for investigation.

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +15,220 @@ import (
...
-	inspectResult, err := dockerClient.ContainerInspect(context.Background(), cont.ID, client.ContainerInspectOptions{})
+	db, err := goose.OpenDBWithDriver("mysql", dbURI)
 	require.NoError(t, err)
+	defer db.Close()
 
-	p, ok := inspectResult.Container.NetworkSettings.Ports[mySQLPort]
-	if !ok || len(p) == 0 {
-		require.Fail(t, "failed to get host port mapping from mysql container")
-	}
+	require.NoError(t, goose.Up(db, assets.MySQLMigrationDir))
```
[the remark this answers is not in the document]
rafanaskin: Sorted

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +15,220 @@ import (
...
+		checkExec := client.ExecCreateOptions{
+			Cmd: []string{"test", "-s", mysqlTemplateDBDump},
+		}
+		if err := docker.ExecCommand(ctx, containerID, checkExec); err != nil {
+			return fmt.Errorf("mysql template dump not ready: %w", err)
+		}
 
-	return mySQLTestContainer
+		return nil
+	}, backoffPolicy)
```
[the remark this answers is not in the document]
rafanaskin: Sorted

rafanaskin:
> Failing benchmark looks like an issue with recreation logic not running migrations:

The issue was caused by splitting the dump into multiple container exec calls. This led to which could result in both dump files being present (`openfga-test-db-template.sql` and `openfga-test-db-template.sql-partial`). Moving everything into a single exec made the dump generation deterministic and fixed the issue.

senojj:
@rafanaskin note that I had to make some changes to the PostgreSQL test container creation in PR #3062 to prevent race conditions between parallel tests and packages when creating a single container instance shared between all tests and packages.

I ended up guarding package-level container creation with a combination of a `sync.Mutex`, a `sync.Cond`, and an `atomic.Bool`.

To guard against cross-package races, I gave each package's container a unique name. This also required adding TestMain cleanup logic per package.

However, all of this was necessary due to how Go runs test packages. Each test package compiles to its own binary, thus executing in a separate process. This creates a race to create the initial container when multiple packages share the same container name, and a mutex guard is not enough to prevent the race across packages.

rafanaskin:
@senojj thanks for sharing this.

How would you like to proceed with this PR?
* We can merge it before #3062, and you can add MySQL support here
* Or you can merge #3062 first, and I'll update this PR to follow the container-per-package approach

senojj:
@rafanaskin #3062 has already been merged as I write this, so you will end up bringing those changes into this PR. 

You can copy my changes or even improve upon them, if you so desire, as I made the changes in haste and am not necessarily "married" to the solution.

rafanaskin:
> "married" to the solution.

😂

assets/migrations/sqlite/006_add_store_ulid_index.sql
```diff
@@ -0,0 +1,5 @@
+-- +goose Up
+CREATE INDEX idx_tuple_store_uild ON tuple (store, ulid);
```
DanCech: is that a typo in the index name?

rafanaskin: Yes, it is. Thank you.

I'm also not sure this index should be part of this PR, since it's related to SQLite rather than MySQL.

DanCech: Yeah, is that just cleaning up something that we added to mysql & postgres but forgot to add to sqlite?

DanCech (commented):
This is looking great!

assets/migrations/sqlite/006_add_store_ulid_index.sql
rafanaskin: This index improves SQLite test performance, reducing execution time by about 2x. The longest-running test involves pagination by a single element, which appears to be particularly expensive for SQLite to handle. Adding this index makes the query significantly faster.

I also assume this index could improve production API performance when using SQLite.

I also tested replacing the SQLite package with the https://github.com/mattn/go-sqlite3  driver, which reduced execution time dramatically. However, it relies on CGO.

DanCech: Yeah, I opted to use the non-cgo library under the assumption that it's fine for smaller installs and larger installs would choose postgres or mysql anyway, so no need to deal with all the cgo complications.

As far as the index, it will definitely help performance significantly when just paging through all the tuples in a given store, since SQLite will use that index first to filter by store, then to sort by ulid. It likely won't be as helpful when you add more filter terms, where it would be more likely to choose one of the existing indexes if the second column covers one of the specified conditions.

With that said, I don't have any objection to adding it given that it improves test performance and will presumably have a similar impact on paging through tuples in production.

DanCech:
We should probably keep the approach to container lifecycle consistent between postgres and mysql

rafanaskin:
> We should probably keep the approach to container lifecycle consistent between postgres and mysql

Yes, I am working on it.

assets/migrations/sqlite/006_add_store_ulid_index.sql
rafanaskin: Yes, I saw your [PR](https://github.com/openfga/openfga/pull/1615) and agree that avoiding `cgo` complications is a worthwhile tradeoff.

>I don't have any objection to adding it given

In that case, I left the migration in this PR. Thank you.

rafanaskin:
@coderabbitai review

rafanaskin:
@DanCech @senojj
The PR is ready for review when you have time.
Thanks again for your earlier feedback - I really appreciate it.

> [!NOTE]
> There is an issue with Codecov.

pkg/testfixtures/storage/postgres.go
```diff
@@ -141,39 +113,11 @@ func RunPostgresTestContainer(t testing.TB) DatastoreTestContainer {
...
-			}
-			cond.Wait()
-			continue
-		}
-		break
-	}
-	mu.Unlock()
+	postgresOnce.Do(func() {
+		postgresDockerCont = bootstrapPostgresContainer(t, docker)
+	})
```
senojj: @rafanaskin the only issue that I have with using a `sync.Once` here is that if a transient error occurs within `bootstrapPostgresContainer`, your only option is to fail the test, and the rest of the tests within the package won't be able to attempt to booptstrap the container.

The version that I had prior allows the current test to fail on bootstrap error -- though it could be changed to just re-enter the loop -- and always allows subsequent tests in the package to attempt to create the container. Additionally, it does so without busy waiting on any of the goroutines not currently participating in container creation.

Another benefit of using the `sync.Cond` is that it is going to be far more efficient than a busy wait and far more responsive than an exponential backoff using `time.Sleep`.

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +16,192 @@ import (
...
 	t.Cleanup(func() {
-		dockerClient.Close()
+		docker.Close()
 	})
 
-	imageListResult, err := dockerClient.ImageList(context.Background(), client.ImageListOptions{
-		All: true,
+	mysqlOnce.Do(func() {
+		mysqlDockerCont = bootstrapMysqlContainer(t, docker)
 	})
```
senojj: @rafanaskin see comment in PostgreSQL file regarding use of package-level `sync.Once`.

senojj:
@rafanaskin by the way, thank you for all of your great work on this project. It has been a pleasure working with you, and the work that you produce has been high quality.

pkg/testfixtures/storage/postgres.go
rafanaskin: Thank you for the explanation and for outlining the reasoning behind the `sync.Cond` approach.

I originally used `sync.Once` to simplify the code, but you're right that a transient bootstrap error would prevent subsequent tests from attempting container creation again.

I've replaced `sync.Once` with `sync.Cond` and also adjusted the synchronization logic. One thing I noticed is that in the [previous version](https://github.com/rafanaskin/openfga/blob/75f503d428417cff702fbd29c7a6ea82fc31467f/pkg/testfixtures/storage/postgres.go#L146), it looked like waiting test goroutines could remain blocked on the mutex and never actually reach `cond.Wait()`. Could you please take another look at that section to confirm whether my understanding is correct?

rafanaskin:
@coderabbitai review

pkg/testfixtures/storage/postgres.go
senojj: @rafanaskin while I don't see a way for goroutines to remain blocked indefinitely on the mutex in the previous version, you are correct that no goroutines would make use of the `sync.Cond`, because they would remain blocked at the `sync.Mutex` until the boostrapping finished. There were certainly a few other issues with my original version, including the unnecessary use of an `atomic.Bool` and additional unnecessary API calls.

Thank you for observing the code closely and identifying those issues. Your code is correct.

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +16,247 @@ import (
...
-		Env: []string{
-			"MYSQL_DATABASE=defaultdb",
-			"MYSQL_ROOT_PASSWORD=secret",
-		},
-		ExposedPorts: network.PortSet{
-			mySQLPort: {},
-		},
-		Image: mySQLImage,
-	}
+	port, err := docker.GetHostPort(mysqlDockerCont, mysqlPort)
```
senojj: `mysqlDockerCont` is accessed concurrently, but it doesn't look like a lock is held here. Either wrap the value in an `atomic.Pointer`, or copy the value to another variable while holding a lock.

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +16,247 @@ import (
...
 var (
-	mySQLPort = network.MustParsePort("3306/tcp")
+	_ DatastoreTestContainer = (*mysqlTestContainer)(nil)
+
+	mysqlContainerName = "openfga-test-mysql-" + ulid.Make().String()
+	mysqlPort          = network.MustParsePort("3306/tcp")
+	mysqlDockerCont    *container.InspectResponse
+
+	mysqlBootstrapping bool
+	mysqlCond          = sync.NewCond(&sync.Mutex{})
```
senojj, line 40: does the mutex and/or condition here escape to the heap? if so, it can be avoided entirely:
```go
var (
    mu sync.Mutex
    cond sync.Cond
)

func init() {
    cond.L = &mu
}
```

pkg/testfixtures/storage/mysql.go
senojj: This _may_ be ok, since the final mutation can establish a happens-before order with the other goroutines synchronizing around access of the value.

The key thing we need to ensure is that other goroutines running on other processors don't get stale data from their processor's cache.

However, since this is a mutable package-level variable, I'd feel more comfortable synchronizing through an atomic.

rafanaskin: You're right that after bootstrapping, `mysqlDockerCont` is accessed concurrently, but only for reads, with no concurrent writes, so there's no data race.

I can copy the pointer while holding the lock if that makes the intent clearer and helps avoid confusion in the future.

```go
dockerCont := mysqlDockerCont
mysqlCond.L.Unlock()
```

What do you think?

senojj: My concern isn't with a race between reads/writes. I'm more concerned about the happens-before ordering and the Go memory model.

If you can guarantee that all goroutines will read the variable across a synchronization boundary _after_ the final mutation, then this is ok as-is.

rafanaskin: You're right that the unlocked read at this line requires some non-obvious reasoning to verify. My earlier suggestion to copy the pointer while holding the lock applies here as well, since it makes the `happens-before` relationship explicit.

Regarding `atomic.Pointer`: I think it would actually be misleading here, since it signals "this variable is written concurrently by multiple goroutines", which isn't the case. `mysqlDockerCont` is written exactly once, under the `mutex`.   

The correctness of the current code follows from the `happens-before` guarantees:
`B` - bootstrapping goroutine.
`W` - waiting goroutines.

1. `B` writes `mysqlDockerCont` while holding the lock.
2. `B` calls `Broadcast()` while still holding the lock, then exits the loop and unlocks at `line 121`.
3. `W` wake up, **re-acquire the lock** and resume after `Wait()`.
4. `W` observe `mysqlDockerCont != nil` in the loop condition while still holding the lock, then exit the loop.
5. `W` unlock at `line 121` and read `mysqlDockerCont` at `line 123`.

`Steps 1–4` establish the `happens-before` relationship through mutex synchronization. 
`Step 5` is safe due to sequential execution within the same goroutine.

That said, copying the pointer while holding the lock removes the need to reason through the memory model, so I'll make that change.

pkg/testfixtures/storage/mysql.go
rafanaskin: Investigated this - neither escapes to the heap. The compiler inlines `sync.NewCond` and determines that both `sync.Cond` and `sync.Mutex` are only referenced by a package-level variable, so they are placed in the `.data segment` by the linker. There's no `runtime.newobject` call and no `GC` involvement.

Using a `var plus init()` would produce the same result.

This can be verified in the following way.:
```go
go build -gcflags='-S' -a ./pkg/testfixtures/storage/ 2>&1 | grep 'mysql\.go:39'
// no output - no heap allocation or runtime initialization code
```

```go
go build -o /tmp/storage ./pkg/testfixtures/storage/
go tool nm /tmp/storage | grep mysqlCond
// 815a3 D github.com/openfga/openfga/pkg/testfixtures/storage.mysqlCond
// The D symbol indicates the variable lives in the program data segment rather than on the heap.
```

```go
go build -gcflags='-m=1' ./pkg/testfixtures/storage/ 2>&1 | grep mysql.go
# not contains "does not escape", "moved to heap" or "escapes to heap"
```

By the way, if optimizations are disabled, the `mutex` will escape to the heap.
```go
go build -gcflags='-S -N -l' -a ./pkg/testfixtures/storage/ 2>&1 | grep 'mysql\.go:39'
// 0x00dc 00220 (pkg/testfixtures/storage/mysql.go:39)  PCDATA  $0, $-1
// 0x00dc 00220 (pkg/testfixtures/storage/mysql.go:39)  MOVD    $type:sync.Mutex(SB), R0
// 0x00e4 00228 (pkg/testfixtures/storage/mysql.go:39)  CALL    runtime.newobject(SB)
```

pkg/testfixtures/storage/postgres.go
```diff
@@ -24,9 +23,7 @@ import (
 )
 
 const (
-	postgresImage     = "postgres:17-alpine"
-	postgresContainer = "openfga-test-postgres"
-
+	postgresImage      = "postgres:17-alpine"
```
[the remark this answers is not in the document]
rafanaskin: All tests passed successfully, no issues detected.

pkg/testfixtures/storage/postgres.go
```diff
@@ -216,15 +189,15 @@ func RunPostgresTestContainer(t testing.TB) DatastoreTestContainer {
 	return testCont
 }
 
-func bootstrapPostgresContainer(t testing.TB, docker *testutils.DockerClient) (*container.InspectResponse, error) {
-	t.Logf("No running container found for %s, creating a new one", postgresImage)
+func CleanupPostgresContainer() {
+	_ = cleanupDatastoreTestContainer(postgresContainerName)
+}
```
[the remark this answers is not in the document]
rafanaskin: Added comments

pkg/testfixtures/storage/mysql.go
```diff
@@ -19,180 +16,248 @@ import (
...
+	// Shared MySQL container bootstrap for concurrent tests using sync.Cond.
+	// Only one test bootstraps the shared container at a time, while others wait efficiently using sync.Cond.
+	// If bootstrap fails, waiting tests are awakened so another test can retry without being affected by the failure.
+	mysqlCond.L.Lock()
+	for mysqlDockerCont == nil {
+		if !mysqlBootstrapping {
+			mysqlBootstrapping = true
+			mysqlCond.L.Unlock()
+
+			dockerCont, err := bootstrapMysqlContainer(t.Context(), docker)
```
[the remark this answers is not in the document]
rafanaskin: Skip

pkg/testfixtures/storage/postgres.go
```diff
@@ -357,10 +338,11 @@ exec docker-entrypoint.sh postgres -c hot_standby=on -c max_connections=200
...
 		AutoRemove:      true,
 		PublishAllPorts: true,
+		Tmpfs:           map[string]string{"/var/lib/postgresql/data": "size=2g"},
 		ExtraHosts:      []string{"host.docker.internal:host-gateway"},
 	}
 
-	contName := containerName + "-replica"
+	contName := postgresContainerName + "-replica"
 	cont, err := docker.RunContainer(t.Context(), contCfg, hostCfg, contName)
 	require.NoError(t, err, "run postgres container")
```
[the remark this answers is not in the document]
rafanaskin: CreateSecondary is used only in a single test case, for which this method was introduced. Let’s postpone this recommendation for now.

pkg/testfixtures/storage/mysql.go
senojj: I agree that the current code is safe. This is due to step 4 being a synchronization point for each goroutine, reading the value of the variable, ensuring that the processor cache holds the most recent value.

My primary concern has primarily been that this is a package-level variable. While your code today is perfectly safe, a seemingly innocuous usage of that package-level variable elsewhere in the package could cause a stale read. This may or may not be flagged by the race detector during tests.

pkg/testfixtures/storage/mysql.go
senojj: Thanks for confirming. I mostly wanted to know for my own knowledge in this case. Even if either value did escape to the heap, it would be a negligible impact.

pkg/testfixtures/storage/mysql.go
senojj: Note that use of an atomic value does not indicate concurrent writes. It indicates that the variable can be mutated concurrently with access. If a value is mutated concurrently, within a lock, and is read _after_ the write happens but without first entering a synchronization point, there is the potential for a stale read.

Therefore, an atomic value ensures that the caches of all cores are invalidated/updated appropriately, preventing stale reads.

rafanaskin: Thanks for explaining your perspective - I'm absolutely okay with using atomic.Pointer to help prevent inaccurate usage in the future. I'll implement this.

rafanaskin: @senojj all set. I've added `atomic.Pointer` to both `PostgreSQL` and `MySQL` fixtures.

pkg/server/testmain_test.go
```diff
@@ -10,5 +10,6 @@ import (
 func TestMain(m *testing.M) {
 	code := m.Run()
 	storagefixtures.CleanupPostgresContainer()
+	storagefixtures.CleanupMysqlContainer()
```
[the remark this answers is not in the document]
rafanaskin: Each package uses a unique container, ensured by: `mysqlContainerName = "openfga-test-mysql-" + ulid.Make().String()`.

This guarantees isolation between packages by assigning a distinct container name.

senojj (approved):

Issue #3022, opened 2026-03-29
OpenTelemetry logging


simongottschlag:
### Checklist

- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Right now, we only export traces using open telemetry. Not sure about metrics.

### Describe the ideal solution

Have a flag that enables open telemetry exports for logging. If not already available, the same would be sweet for metrics as well.

alethenorio:
I have myself been looking into this recently. It would be great to be able to export logs to en OpenTelemetry endpoint such as the https://opentelemetry.io/docs/collector/ to get better support for transforming the logs to match the expect underlying platform (in our case Google Cloud)

simongottschlag:
> I have myself been looking into this recently. It would be great to be able to export logs to en OpenTelemetry endpoint such as the https://opentelemetry.io/docs/collector/ to get better support for transforming the logs to match the expect underlying platform (in our case Google Cloud)

I'll treat you and Fredrik to a lunch as a thank you when this is released. 😊

PR #3048, opened 2026-04-07
feat(log): add OTLP log export via otelzap bridge

fredrikaverpil:
## Why?

OpenFGA supports exporting traces via OTLP but has no equivalent for logs. This means log-trace correlation is not possible in observability backends that support OpenTelemetry, making it harder to debug issues across distributed systems.

## What?

- Add otelzap bridge core that exports logs via OTLP to any OpenTelemetry-compatible backend (Grafana Loki, Datadog, GCP Cloud Logging, etc.)
- Logs are exported via OTLP in addition to stdout when `log.otlp.enabled` is set
- The otelzap bridge attaches span context (trace ID, span ID) to each log record, enabling log-trace correlation
- Use `*WithContext` logging methods in the gRPC logging interceptor so the bridge can extract span context from the request
- Add `contextFilterCore` that strips the context field from stdout (only meaningful to the otelzap bridge)
- New config: `OPENFGA_LOG_OTLP_ENABLED`, `OPENFGA_LOG_OTLP_ENDPOINT`, `OPENFGA_LOG_OTLP_TLS_ENABLED`, with `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` / `OTEL_EXPORTER_OTLP_ENDPOINT` fallbacks for the endpoint

## Notes

- Follows the same pattern as the existing OTLP trace export configuration: like `trace.enabled`, the explicit `log.otlp.enabled` flag controls whether logs are exported, so the standard OTel env vars only control where they are sent
- zap's production sampling is applied before the fan-out to stdout and OTLP, so both receive the same sampled stream and collector egress stays bounded during log storms
- When OTLP log export is not enabled, behavior is unchanged


fredrikaverpil:
**Related:** #3049

These two PRs are independent approaches to the same goal: enabling log-trace correlation so that OpenFGA Check/Read logs can be seen alongside the upstream RPC that triggered them (e.g. expanding a platform service request log shows the correlated OpenFGA authorization checks with query duration, request/response payloads, etc.).

- **This PR (#3048)** — cloud-agnostic OTLP log export via otelzap bridge. Works with any OTel-compatible backend but requires an OTLP collector with a logs pipeline.
- **#3049** — adds configurable `logging.googleapis.com/*` trace fields to stdout for GCP Cloud Logging correlation. Simpler, no collector needed, but GCP-specific (behind a config flag).

Both PRs are complementary and can be used together or independently.

fredrikaverpil:
**Note:** This PR exports logs via OTLP, which produces *new* log entries with trace correlation. However, the existing stdout logs (which contain the rich structured data like raw request/response payloads) still lack correlation. For environments where stdout log correlation is the goal, #3049 is the more direct solution — it adds trace context fields to the existing stdout output so the cloud logging agent can correlate them natively.

fredrikaverpil:
Related issue: #3022

fredrikaverpil:
### Required deployment changes (Cloud Run / Knative)

For the OTLP log export to work, the following is needed in addition to the fork:

**OpenFGA container env var:**
```yaml
- name: OPENFGA_LOG_OTLP_ENABLED
  value: "true"
- name: OPENFGA_LOG_OTLP_ENDPOINT
  value: "127.0.0.1:4317"

```

**OTel collector sidecar config** — add a `logs` pipeline and `default_log_name`:
```yaml
exporters:
  googlecloud:
    log:
      default_log_name: openfga

service:
  pipelines:
    logs:
      receivers: [otlp]
      processors: [resourcedetection]
      exporters: [googlecloud]
```

**IAM** — the service account needs `roles/logging.logWriter` for the otel collector to write logs via the Cloud Logging API.

holgerstolzenberg:
hey @fredrikaverpil,

Any chance to get the merge conflicts resolved and the license/security issues resolved, so that the PR can leave draft state? We would also love to see OTEL push logs to happen for all the benefits its has.

fredrikaverpil:
@holgerstolzenberg I don’t plan to pursue this PR any further right now, as it does not solve the problem I have with not seeing logs correlated in GCP Cloud Run.

This PR itself might bring a valuable change but it also pulls in new dependencies and touches the logger in a bunch of places and I don’t have the time to push this forward right now. If you like, feel free to take it over. I've rebased. 

I’d like to continue the work on this instead, which targets the specific problem I have: https://github.com/openfga/openfga/pull/3049
If not accepted, I might come back to this PR if that is an approach that is more favorable by the maintainers.

PR #3064, opened 2026-04-13
fix: add null byte delimiter in contextual tuple cache keys and validation in v2Check

Vic-Dev:
Context tuple keys should use delimiters to prevent tuples that can result in overlapping keys from returning incorrect results. Added validation in v2Check for the initial Execute.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a cache key collision in the experimental weighted_graph_check feature that could cause incorrect contextual tuple lookups.
  * Contextual tuple indexing and lookups now use explicit delimiters so similar component combinations no longer produce ambiguous keys.


internal/check/request.go
```diff
@@ -227,7 +227,9 @@ func (r *Request) buildContextualTupleMaps() {
 		// Build key for ctxTuplesByUserID: userId+relation+objectType
 		keyBuilder.Reset()
 		keyBuilder.WriteString(user)
+		keyBuilder.WriteString(ctxTupleKeyDelimiter)
```
senojj: instead of WriteString use WriteByte

internal/check/check.go
```diff
@@ -29,6 +29,7 @@ import (
 
 const cacheKeyDelimiter = "|"
 const cacheKeyPrefix = "c."
+const ctxTupleKeyDelimiter = "\x00"
```
senojj:
```go
const ctxTupleKeyDelimiter byte = '\x00'
```

senojj: Additionally, is it 100% impossible to use a null byte within a contextual tuple value?

Vic-Dev: null bytes are being excluded from here:
https://github.com/openfga/openfga/blob/main/pkg/storage/cache.go#L385
which is called here:
https://github.com/openfga/openfga/blob/main/internal/check/request.go#L157

pkg/server/commands/check.go
```diff
@@ -132,6 +135,11 @@ func NewCheckQuery(opts ...CheckQueryV2Option) *CheckQueryV2 {
 }
 
 func (q *CheckQueryV2) Execute(ctx context.Context, req *openfgav1.CheckRequest) (*openfgav1.CheckResponse, storagewrappers.Metadata, error) {
+	err := sanitizeRequest(req)
```
senojj: This is more of a validation that a sanitization. A function that sanitizes a value, usually returns a new value with invalid characters escaped, stripped or replaced.

internal/check/check.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const cacheKeyDelimiter = "|"
+const cacheKeyDelimiterByte byte = '\x00'
```
senojj: This should probably be named something like "ctxTupleMapKeyDelimiter". It really isn't used for a cache key.

senojj (approved):

PR #3070, opened 2026-04-14
fix: v2Check honours `check-query-cache-enabled` flag

saad-h1:

v2Check always does "query caching" (caching final + intermediate "allowed? true/false" results). It does not honour the `check-query-cache-enabled` flag.

#### How is it being solved?

Pass the `check-query-cache-enabled` flag into the v2Check `Resolver` and gate cache operations behind a conditional check on that flag.

An alternative approach would be to pass a no-op cache into the `Resolver` if `check-query-cache-enabled` is false, but this would cause unnecessary computations/allocations (building cache keys, function calls to Get/Set on the no-op cache, etc.).

See files changed.

## References

n/a

## Review Checklist
n/a
n/a

## Summary by CodeRabbit

* **New Features**
  * Added separate control for query caching with an independent enable/disable toggle and TTL.

* **Improvements**
  * Improved cache configuration wiring and more reliable cache invalidation behavior for check operations.

* **Tests**
  * Added tests verifying behavior with query caching enabled and disabled.
  * Simplified cache-invalidation test flow for clearer, more maintainable verification.


internal/check/check.go
```diff
@@ -182,32 +187,41 @@ func (r *Resolver) ResolveUnionEdges(ctx context.Context, req *Request, edges []
 	relation := req.GetTupleKey().GetRelation()
 	objectType := tuple.GetType(req.GetTupleKey().GetObject())
 	objectRelation := tuple.ToObjectRelationString(objectType, relation)
-	ids := make([]string, 0, len(edges))
```
saad-h1, line 185: This variable was previously collecting the edge cache keys, but I found that the values in this slice were never actually used. Only the length of this slice was used to determine how many values should be extracted from the `out` channel in the second for loop below. Converted to an `int` instead and renamed.

pkg/server/check.go
```diff
@@ -264,6 +264,7 @@ func (s *Server) v2Check(
 		commands.WithCheckQueryV2Model(mg),
 		commands.WithCheckQueryV2Cache(cache),
 		commands.WithCheckQueryV2CacheTTL(s.cacheSettings.CheckQueryCacheTTL),
+		commands.WithCheckQueryV2QueryCacheEnabled(s.cacheSettings.ShouldCacheCheckQueries()),
```
saad-h1, line 266: `WithCheckQueryV2QueryCacheEnabled` has `Query` twice in the name, but the first refers to the `CheckQueryV2` struct and the second refers to the `QueryCacheEnabled` flag. Not sure if there's a better way to name this.

pkg/server/server_test.go
```diff
@@ -2566,37 +2566,18 @@ func TestV2CheckWithIteratorCache_Invalidation(t *testing.T) {
 	})
 	require.NoError(t, err)
 
-	// Delete the stale store-level invalidation marker so we can detect
-	// when the cache controller writes a fresh one after seeing the new writes.
-	cache.Delete(storage.GetInvalidIteratorCacheKey(storeID))
-
-	// WORKAROUND: Ideally, this test should disable the query cache; however,
```
saad-h1, line 2573: The changes to this file are just a `git revert 26d409c` which reverts [this](https://github.com/openfga/openfga/pull/3016/changes/26d409c4a76b533bc29309e085fbedb1872d5207) commit. The comment here explains why that commit was needed as a workaround since we couldn't disable the query cache. Now that we can, that commit is no longer needed.

pkg/server/check_test.go
```diff
@@ -597,3 +597,58 @@ func TestV2Check_SanitizeRequest(t *testing.T) {
...
+		s, req := setupCheckServer(t, "", nil,
+			WithCheckQueryCacheEnabled(false),
+			WithCheckCacheLimit(10),
+			WithCheckQueryCacheTTL(1*time.Minute),
+			WithCacheControllerEnabled(true),
+		)
+
+		checkCache := newRecordingCache()
+		s.sharedDatastoreResources.CheckCache = checkCache
+		s.authzModelGraphResolver = modelgraph.NewResolver(s.datastore, checkCache, 24*7*time.Hour)
```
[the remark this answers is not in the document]
saad-h1: In this test, all caches are disabled at the server level so `s.sharedDatastoreResources.CheckCache` is nil. We overwrite it afterwards in this test with a `newRecordingCache()`.

internal/check/check.go
```diff
@@ -71,6 +73,7 @@ func New(cfg Config) *Resolver {
 		cache:                     cfg.Cache,
 		cacheTTL:                  cfg.CacheTTL,
 		lastCacheInvalidationTime: cfg.LastCacheInvalidationTime,
+		queryCacheEnabled:         cfg.QueryCacheEnabled,
```
senojj: Instead of propagating `queryCacheEnabled` checks everywhere, why not create a NoopCache and pass it as the query cache when disabled?

NoopCache should always return a miss on Get, and do nothing on Set.

saad-h1: Ah we could definitely do that, I was just thinking that would cause unnecessary computations/allocations (building cache keys, function calls to Get/Set on the no-op cache, etc.). Maybe that small overhead isn't a big deal though?

senojj: The cache key building is 1 alloc/op. The edges array is small from a practical sense, no more than 20, so the loop might result in 20 allocations unnecessarily, which is likely a drop in the bucket from some of the other operations involved.

A better approach is to make the cache key building lazy, so that the key is built within the cache methods themselves. This would require creating wrapper methods like `GetCacheValue(key fmt.Stringer) (value, ok)` and `SetCacheValue(key fmt.Stringer, entry any, ttl time.Duration)`

Then you would create a type for each key.
```go
type edgeCacheKey struct {
    modelID string
    req *Request
    edge *Edge
}

func (k *edgeCacheKey) String() string {
    return buildEdgeCacheKey(k.modelID, k.req, k.edge)
}
```

Building the cache key this way _shouldn't_ require an allocation when the cache method doesn't call `String()`. The question is: does boxing the `edgeCacheKey` into a `fmt.Stringer` require a heap allocation? I would hope that the compiler is smarter than that in this case.

saad-h1: In that case, what were you envisioning the wrapper functions would look like and how would they decide whether to call String() or not? I was trying a few things but couldn't seem to get something without allocs (it seemed like because the cache itself is an interface from a _separate_ package, it couldn't avoid the allocs).

I also wonder if adding all this logic on top of cache key generation is worth it 🤔

senojj: Yes, this goes back to my original statement of:

> The edges array is small from a practical sense, no more than 20, so the loop might result in 20 allocations unnecessarily, which is likely a drop in the bucket from some of the other operations involved.

While lazily building the cache keys would be a strategy to avoid unnecessary allocs, it also adds complexity unnecessarily in a location where it likely won't have much of a measurable benefit.

senojj: > what were you envisioning the wrapper functions would look like and how would they decide whether to call String() or not?

That logic would have to reside within the cache implementation's methods. However, we can add that when we identify a measurable performance issue with a less complicated approach.

saad-h1: Makes sense yeah. For now then, do you think we should just go with the no-op cache approach or the current changes in the PR to pass the flag around and do `if`s on it?

senojj: The no-op cache is preferable.

pkg/server/commands/check.go
```diff
@@ -29,7 +29,8 @@ type CheckQueryV2 struct {
 	datastore                 storage.RelationshipTupleReader
 	datastoreOp               storagewrappers.Operation
 	cache                     storage.InMemoryCache[any]
-	cacheTTL                  time.Duration
```
saad-h1, line 32: I thought this should be renamed to `queryCacheTTL` since it's not clear whether this TTL is for the query or iterator cache. The iterator cache TTL isn't on this struct, it's on the SharedDatastoreResources.

I know "query" isn't a great descriptor for these cache entries, "sub-problem" would probably be better, but I thought it's more important to stay consistent with the config so that someone can easily associate these fields/var names to the config value they passed in.

internal/check/check.go
saad-h1: Updated to use no-op cache, code changes not needed anymore in `internal/check/`

senojj (approved):

PR #3097, opened 2026-04-29
fix: v2Check EdgeCacheKey collisions

saad-h1:

When the experimental `weighted_graph_check` resolves a union with multiple branches - e.g., `define viewer: [user, user:*]` or `define viewer: (writer and admin) or (editor and contributor)` - it caches the result of each branch individually to avoid redundant work on repeated requests. The cache key for each branch included the relation that generated it (e.g. `"document#viewer"`), but this is the same for every branch within the same relation's rewrite tree. On the first request both branches resolve correctly, but the last one to finish overwrites the cache entry for the shared key. On subsequent requests, both branches return that same cached result - meaning one branch's answer is silently substituted for another's, which can produce incorrect allow or deny decisions.

#### How is it being solved?

For the edge being resolved, its type, target node, and (for TTU edges) which tupleset relation it follows are added to the edge cache key to uniquely identify this edge.

In `buildEdgeCacheKey`, we now add the following to the key from the edge:
1. `edgeType`: needed to distinguish a direct userset edge from a TTU edge when both point at the same target node (e.g. `[group#member] or member from parent`).
2. `to.uniqueLabel`: needed to distinguish edges that target different nodes: different terminal types (`[user, user:*]`), different TTU targets (viewer from org or viewer from team), and different intersection/exclusion operator nodes (`(A and B) or (C and D)`).
3. `tuplesetRelation`: needed to distinguish two TTU edges that share the same target node but follow different tupleset paths (e.g. `admin from parent or admin from owner` where both `parent` and `owner` are of the same type)

New format: `c.<modelID>|<object>|<user>|<relationDefinition>|<edgeType>|<toUniqueLabel>|<tuplesetRelation>|<invariantCacheKey>`

Also added a missing delimiter between `relationDefinition` and `invariantCacheKey`.

Added tests to cover different collision scenarios and ensure each edge has a unique key.

## References

n/a

## Review Checklist
n/a
n/a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge cache-key construction to prevent collisions across complex edge scenarios (union/intersection branches, TTU paths, direct and computed relations), yielding more accurate cached authorization results.

* **Tests**
  * Added focused tests validating edge cache-key uniqueness across varied graph and request configurations, including multiple branch and path cases.

* **Documentation**
  * Updated changelog entry describing the cache-key fix for union/complex edges.


internal/check/check.go
```diff
@@ -159,6 +160,13 @@ func buildEdgeCacheKey(modelID string, req *Request, edge *authzGraph.WeightedAu
 	keyBuilder.WriteString(req.GetTupleKey().GetUser())
 	keyBuilder.WriteString(cacheKeyDelimiter)
 	keyBuilder.WriteString(edge.GetRelationDefinition())
+	keyBuilder.WriteString(cacheKeyDelimiter)
```
senojj: The delimiter should be a single byte... use `(*strings.Builder).WriteByte()`.

If `cacheKeyDelimiter` is a `string` type, change it to `byte`.
e.g. `var cacheKeyDelimiter = "|"` should be `var cacheKeyDelimiter byte = '|'`

senojj: note that the current format is also vulnerable to manipulation.....

internal/check/check.go
```diff
@@ -159,6 +160,13 @@ func buildEdgeCacheKey(modelID string, req *Request, edge *authzGraph.WeightedAu
...
+	keyBuilder.WriteString(strconv.FormatInt(int64(edge.GetEdgeType()), 10))
+	keyBuilder.WriteString(cacheKeyDelimiter)
+	keyBuilder.WriteString(edge.GetTo().GetUniqueLabel())
+	keyBuilder.WriteString(cacheKeyDelimiter)
+	keyBuilder.WriteString(edge.GetTuplesetRelation())
+	keyBuilder.WriteString(cacheKeyDelimiter)
 	keyBuilder.WriteString(req.GetInvariantCacheKey())
 	return keyBuilder.String()
```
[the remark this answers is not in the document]
senojj: I mention this in a prior comment as well... Is this one of the areas that we will be applying the hex encoding in a follow-up PR?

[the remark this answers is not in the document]
saad-h1: Yes I think we can look into this in a separate PR which should cover all these cases, especially since this cache key is just in an experimental feature still.

internal/check/check.go
saad-h1: addressed in https://github.com/openfga/openfga/pull/3097/commits/935b004658c542af506a1c50eadea73b97d02f52

RE vulnerability: https://github.com/openfga/openfga/pull/3097#discussion_r3163324364

internal/check/check_test.go
```diff
@@ -344,7 +344,7 @@ func TestResolveUnionEdges(t *testing.T) {
 				return nil, storage.ErrNotFound
 			}).MaxTimes(2)
 
-		key := fmt.Sprintf("^c.%s|group:1|user:maria|group#admin.*$", storeID)
+		key := fmt.Sprintf("^c.%s|group:1|user:maria|group#admin|.*$", storeID)
```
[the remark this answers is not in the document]
saad-h1: addressed in https://github.com/openfga/openfga/pull/3097/commits/2e61b4cd8c592af716a9248886976d72287b6a78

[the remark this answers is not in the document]
saad-h1: addressed in https://github.com/openfga/openfga/pull/3097/commits/2e61b4cd8c592af716a9248886976d72287b6a78

internal/check/check.go
```diff
@@ -154,19 +154,19 @@ func buildEdgeCacheKey(modelID string, req *Request, edge *authzGraph.WeightedAu
...
 	keyBuilder.WriteString(modelID)
-	keyBuilder.WriteString(cacheKeyDelimiter)
+	keyBuilder.WriteByte(cacheKeyDelimiter)
 	keyBuilder.WriteString(req.GetTupleKey().GetObject())
-	keyBuilder.WriteString(cacheKeyDelimiter)
+	keyBuilder.WriteByte(cacheKeyDelimiter)
 	keyBuilder.WriteString(req.GetTupleKey().GetUser())
-	keyBuilder.WriteString(cacheKeyDelimiter)
+	keyBuilder.WriteByte(cacheKeyDelimiter)
 	keyBuilder.WriteString(edge.GetRelationDefinition())
```
senojj, line 162: Do we want the "relation definition", which is the downstream specific type and relation, or do we want the unique label of the edge's "from" node? Or is that direction even necessary?

internal/check/check.go
```diff
@@ -153,12 +154,19 @@ func buildEdgeCacheKey(modelID string, req *Request, edge *authzGraph.WeightedAu
...
+	keyBuilder.WriteByte(cacheKeyDelimiter)
+	keyBuilder.WriteString(strconv.FormatInt(int64(edge.GetEdgeType()), 10))
+	keyBuilder.WriteByte(cacheKeyDelimiter)
+	keyBuilder.WriteString(edge.GetTo().GetUniqueLabel())
+	keyBuilder.WriteByte(cacheKeyDelimiter)
+	keyBuilder.WriteString(edge.GetTuplesetRelation())
```
yissellokta, line 168: you want the relation definition, not the tupleset relation ( the tupleset is only present on the ttu)

yissellokta: or only add the tupleset if the edge is a ttu to have a different key for the edge case of 
```
 define parent: [org]
			    define owner: [org]
			    define viewer: admin from parent or admin from owner
```

saad-h1: The relation definition is still present a few lines up, that hasn't been removed. And yep I added the tupleset relation for cases like the example you just gave.

> only add the tupleset if the edge is a ttu

If it's not a TTU, wouldn't `GetTuplesetRelation()` just return `""` so it's practically the same?

internal/check/check.go
saad-h1: We would want one of them, since for something like:

```
define rel1: [user]
define rel2: [user]
```

their edges are both the exact same (`edgeType` = DirectEdge, `to.uniqueLabel` = `"user"`, etc.), except for:
- `relationDefinition`
- `from.label`
- `from.uniqueLabel`

As for whether to use `relationDefinition` or `from.uniqueLabel`, it looks like whenever two edges have the same `relationDefinition`, they will also have the same `from.uniqueLabel`.

E.g.,

- `define rel1: (rel2 and rel3) or (rel4 and rel5) or (rel6 and rel7)` - for all 3 edges:
  - `relationDefinition`: `<objectType>#rel1`
  - `from.uniqueLabel`: `union:<ULID_1>`
- `define rel1: [user, user:*]` - for both edges:
  - `relationDefinition`: `<objectType>#rel1`
  - `from.uniqueLabel`: `<objectType>#rel1`

senojj: Have you considered:
`define rel1: (rel2 and (rel3 or rel4)) or (rel5 and rel6)`

The `(rel3 or rel4)` union edge has a "from" node of the intersection, I believe. However, its `relationDefinition` would be `<objectType>#rel1`.

internal/check/check.go
senojj: I think it is likely fine to always include the tupleset relation, as it is just an empty string when the edge is not a TTU, and would be represented in the key as an empty segment (`||`). So we would keep both the relation definition _and_ the tupleset relation.

internal/check/check.go
saad-h1: Ah that's true, although from the perspective of edge caching, it looks like that is only set in `ResolveUnionEdges`, and the two edges that would get passed in to it are:

- `(rel2 and (rel3 or rel4))` --> relationDefinition = `<objectType>#rel1`
- `(rel5 and rel6)` --> relationDefinition = `<objectType>#rel1`

Each of these is `from` the same `union:<ULID>`, but `to` different `intersection:<ULID>`s, so the `to` differentiates them. And each of these would get solved individually as a `ResolveIntersection` which doesn't do edge caching. In `ResolveIntersection` for the first edge above, it would call `ResolveUnionEdges(edges=[rel2->user])` and `ResolveUnionEdges(edges=[rel3->user, rel4->user])`. I.e., `(rel3 or rel4)` would never be an individual edge in `ResolveUnionEdges`, it would always get flattened first.

Leaning on AI since I'm not an expert at this yet 🙂 :

> When relDef is the same but `from` is different (nested union case), `from` varies because `FlattenNode` descends through different union nesting levels. But `to.uniqueLabel` always differs here — each intersection/exclusion node has a unique ULID and one parent, so different from nodes necessarily target different operator nodes. `from.uniqueLabel` varies but adds nothing beyond what `to.uniqueLabel` already provides.

And further, why it's not possible that all of `relationDefinition`, `edgeType`, `to.uniqueLabel`, and `tuplesetRelation` are the same for two edges in `ResolveUnionEdges`:

```
  buildEdgeCacheKey is only called inside ResolveUnionEdges, which only receives terminal edges — whatever FlattenNode didn't recurse through. Non-terminal edges (ComputedEdge, DirectLogicalEdge, TTULogicalEdge, RewriteEdge to union/SpecificTypeAndRelation) are collapsed away before ResolveUnionEdges is ever called. They never get a cache key.  So instead of having to reason about all 6 edge types, you only need to prove uniqueness for the 3 that can actually appear.

  RewriteEdge (type 1) to intersection/exclusion

  tuplesetRelation = "" for all RewriteEdges. Same to.uniqueLabel would require two RewriteEdges pointing to the same operator node. Impossible — each operator node is created with a fresh ULID once, by one parseRewrite call, with one parent. 
  No two edges can share the same operator node as target.                                              
                                                                                                                                                                                                                                                   
  DirectEdge (type 0)                                                                                   
                                                                                                                                                                                                                                                 
  tuplesetRelation = "" for all DirectEdges. Same relDef + to.uniqueLabel means two DirectEdges to the same target within the same relation's tree. UpsertEdge deduplicates any two DirectEdges with the same (from, to) into one edge with merged 
  conditions. For two DirectEdges with different from to the same target: parseThis is called at most once per relation's visible rewrite tree (FlattenNode stops at intersections, so any [...] block nested inside an intersection is invisible;
  at most one [...] block is ever visible). Within one parseThis call, all DirectEdges use the same from node. So there is only ever one from per DirectEdge to a given target. No two distinct DirectEdges to the same target exist.              
                                                                                                        
  TTUEdge (type 2)                                                                                                                                                                                                                                 
   
  This is the only case requiring tuplesetRelation for differentiation. Same relDef + edgeType + to.uniqueLabel + tuplesetRelation means two TTUEdges in the same relation's tree targeting the same computed relation via the same tupleset path. 
  parseTupleToUserset has an explicit HasEdge(node, nodeSource, TTUEdge, typeTuplesetRelation) guard — if an edge with that (from, to, tuplesetRelation) already exists, it skips. And the only way to get two calls with different from but the
  same to and tuplesetRelation would be the same TTU rewrite appearing under two different parent nodes — but either they share the same union node as parent (and HasEdge deduplicates), or one is inside an intersection (invisible to           
  FlattenNode). Either way, two distinct terminal TTUEdges with identical (to, tuplesetRelation) within the same relDef cannot exist.
                                                                                                                                                                                                                                                 
  ---                                                                                                                                                                                                                                              
  In all three cases, either structural uniqueness (ULID), deduplication (UpsertEdge/HasEdge), or FlattenNode's intersection boundary makes it impossible. The four fields together are exhaustively sufficient.
```

From what I can tell looking at this and running various examples, this seems accurate

senojj: sounds good. Thanks for looking into it.

senojj (approved):

PR #3111, opened 2026-05-06
fix: close all channels opened thus far, before return on error

senojj:
### Problem

In `internal/graph/weight_two_resolver.go`, the function `fastPathOperationSetup` iterates over children and creates an `iterator.Stream` for each one. If `fastPathRewrite` returns an error partway through the loop, the function returned immediately without stopping the streams that had already been created. This leaked goroutines/channels associated with those streams.

### Fix

Before returning the error, the function now iterates over all previously-created `iterStreams` and calls `stream.Stop()` on each, ensuring resources are properly released on the error path.

### Changed Files

| File | Change |
|------|--------|
| `internal/graph/weight_two_resolver.go` | Added cleanup loop (+3 lines) in the error branch of `fastPathOperationSetup` |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling to ensure in-progress producer/check streams are properly cleaned up on failure, preventing resource leaks and potential deadlocks.
* **Documentation**
  * Updated changelog with a note about the fix for streams left open after errors.

### DIFF
--- a/internal/graph/weight_two_resolver.go
+++ b/internal/graph/weight_two_resolver.go
@@ -587,6 +587,9 @@ func fastPathOperationSetup(ctx context.Context, req *ResolveCheckRequest, resol
 	for idx, child := range children {
 		producerChan, err := fastPathRewrite(ctx, req, child)
 		if err != nil {
+			for _, stream := range iterStreams {
+				stream.Stop()
+			}
 			return nil, err
 		}
 		iterStreams = append(iterStreams, iterator.NewStream(idx, producerChan))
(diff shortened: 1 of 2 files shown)


saad-h1 (approved):

internal/graph/weight_two_resolver.go
saad-h1, line 1: side note: I see some other places in this file where `iterStreams` are defined for different functions, haven't looked too deeply but seems like some of those could potentially run into the same issue as well?

senojj: No other applicable locations in that file.

PR #3131, opened 2026-05-15
chore: drop grpc_health_probe binary from released Docker image

### LINKED ISSUE #2373: Drop grpc_health_probe from released docker image
### Checklist

From https://github.com/grpc-ecosystem/grpc-health-probe

> grpc_health_probe is meant to be used for health checking gRPC applications in [Kubernetes](https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/), using the [exec probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-command).

> ⚠️ [Kubernetes has now built-in gRPC health checking](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) capability as generally available. As a result, you might no longer need to use this tool and can use the native Kubernetes feature instead.

> This tool can still be useful if you are on older versions of Kubernetes, or using advanced configuration (such as custom metadata, TLS or finer timeout tuning), or not using Kubernetes at all. 

We currently copy their built binary from the docker image `ghcr.io/grpc-ecosystem/grpc-health-probe` into the openfga released image.

https://github.com/openfga/openfga/blob/73dceae92683474c2d788c9b4467abe2d2e3b3db/Dockerfile.goreleaser#L10

We then use it with openfga/helm-charts

https://github.com/openfga/helm-charts/blob/6dc5c872b5e3648c484e3c9127f08f04b575b012/charts/openfga/templates/deployment.yaml#L388

Vulnerabilities reported in the binary are then reflected on the openfga docker image

Image

- grpc probing is supported natively in kubernetes since v1.23 (i'm not sure what the eol policy for k8s is like but that was released in 2021)
- we could replace the bare of it and expose it as in a sub command `openfga probe` to fill the void, but then that's something more we need to support and maintain ourselves. 👎 
- folks can also build their own docker image on top of ours to include the binary (as a migration alternative)

1. cut a release of openfga/helm-charts that leverages `livenessProbe` instead of `grpc_health_probe`, bump major?
2. cut a release of openfga/openfga that drops the included binary.

_No response_

### Additional context

_No response_

Pulkit7070:
## Summary

Closes #2373

Kubernetes v1.23+ (released Dec 2021) has native gRPC liveness/readiness probe support via `livenessProbe.grpc`, making the bundled `grpc_health_probe` binary unnecessary. The binary is copied from an external image layer and regularly surfaces CVEs unrelated to OpenFGA itself.

**Changes:**
- Removes the `COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe` line from `Dockerfile.goreleaser`
- No application code or behaviour changes

Consumers on Kubernetes < v1.23 can layer the binary on top of the released image in their own Dockerfile.

## Test plan
- [ ] Docker image builds without the binary
- [ ] Verify `grpc_health_probe` is absent from the released image layers
- [ ] helm-charts liveness probe updated to native `livenessProbe.grpc` (separate PR in openfga/helm-charts)

## Summary by CodeRabbit

* **Refactor**
  * Simplified the container image by removing an obsolete multi-stage workaround.
  * Resulting image and build configuration are leaner and cleaner, reducing maintenance and build complexity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openfga/openfga/pull/3131?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Dockerfile.goreleaser
+++ b/Dockerfile.goreleaser
@@ -1,11 +1,4 @@
 FROM cgr.dev/chainguard/static@sha256:77d8b8925dc27970ec2f48243f44c7a260d52c49cd778288e4ee97566e0cb75b
 COPY assets /assets
 COPY openfga /
-
-# Goreleaser does not support multi-stage builds. See:
-# https://github.com/goreleaser/goreleaser/issues/609
-#
-# Dependabot is also not capable of updating inline image copies at this time, so
-# this needs to be updated manually until one of these issues is resolved.
-COPY --from=ghcr.io/grpc-ecosystem/grpc-health-probe:v0.4.50@sha256:72c25a244d947eb2b31db49cdb503c4c807d6895c67253eefac5ffde15c44fbd /ko-app/grpc-health-probe /usr/local/bin/grpc_health_probe
 ENTRYPOINT ["/openfga"]


Pulkit7070:
Gentle ping - happy to address any feedback when you get a chance.

Dockerfile.goreleaser
```diff
@@ -1,11 +1,4 @@
 FROM cgr.dev/chainguard/static@sha256:77d8b8925dc27970ec2f48243f44c7a260d52c49cd778288e4ee97566e0cb75b
 COPY assets /assets
 COPY openfga /
-
-# Goreleaser does not support multi-stage builds. See:
```
adriantam, line 5: This will involve more than just dropping the goreleaser. Do we need to update any wiring to not use the grpc probe in the image?

Pulkit7070: Good question, and the honest answer is "yes, there is more wiring to consider." Spelling out what I found so a scope decision can be made:

**Still references `grpc_health_probe` in this repo:**

1. `Dockerfile` (the dev/non-goreleaser image) — copies the binary in **and** has a `HEALTHCHECK CMD ["/usr/local/bin/grpc_health_probe", "-addr=:8081"]` directive. The released `Dockerfile.goreleaser` has no `HEALTHCHECK` directive, so dropping the binary there is a clean no-op for container runtime behaviour; dropping it from `Dockerfile` would also require dropping or replacing the `HEALTHCHECK`. The final stage is `cgr.dev/chainguard/static`, which has no shell or `wget`/`curl`/`nc`, so a CMD-SHELL replacement is not available; the realistic options for the dev image are (a) remove `HEALTHCHECK` and let the orchestrator handle liveness, or (b) keep the dev image as-is.
2. `docker-compose.yaml` — `healthcheck.test` invokes `/usr/local/bin/grpc_health_probe`. Same chainguard/static constraint, so this would either need to be removed or rely on external probing.
3. `.coderabbit.yaml` + `.github/instructions/dockerfile-goreleaser-sync.instructions.md` — they police the version sync between the two Dockerfiles. If the probe goes from the goreleaser image, the sync rule for the probe line in `Dockerfile.goreleaser` becomes stale (the rule still makes sense for the dev `Dockerfile` if we keep it there).

**Out of this repo:** `openfga/helm-charts` uses an `exec` liveness/readiness probe that runs `grpc_health_probe`. Consumers on Kubernetes 1.23+ can move to the native `grpc:` probe; older clusters or non-k8s deployments would need to layer the binary back into the image themselves. That coordination belongs in a separate PR against helm-charts.

**Scope I'd suggest:**

- This PR: drop `grpc_health_probe` from `Dockerfile.goreleaser` only, since that's the released image the issue is about and it has no in-image healthcheck to break.
- A follow-up (if you want): drop from the dev `Dockerfile` + `docker-compose.yaml` + update the sync-rule files. That's a behaviour change for `docker compose up` users.
- A coordinated change in `openfga/helm-charts` to switch to the native gRPC probe.

Happy to do any combination of the above in this PR. Let me know which scope you'd prefer and I'll iterate. The branch is already rebased on main.

curfew-marathon:

PR #3149, opened 2026-06-02
feat: add v2Check logs for resolution breaking change

Vic-Dev:
## Description
  Adds logs for cases where v2Check (weighted-graph algorithm) returns FALSE on a userset request while v1 Check would have returned TRUE. These are intentional behavioural corrections in v2. Previously, check had several bugs where it inferred TRUE
  from schema reachability alone without requiring the underlying data.

  When v2Check returns FALSE on a userset-user request, we run a cheap schema-shape check and emit an info log tagged with a reason field if the request matches one of the known divergence patterns:

  - alias_userset — target relation accepts T#R' directly, where R' is a computed_userset alias for the user's relation R. v1 followed the alias on a stored T#R' tuple; v2 doesn't.
   - self_referential_userset — check(T:id, R, T:id#R). v1 returned TRUE unconditionally if T#R was defined in the model, regardless of data.
  - computed_userset_self_object — user's object equals the target object, and the user's relation appears as a ComputedUserset leaf in the target relation's rewrite (e.g. viewer: editor or writer, query viewer @ doc:d1#writer on doc:d1).
  - ttu_userset — target relation contains a TTU whose computed relation matches the user's relation, and the user's object type is directly-related to the tupleset relation (e.g. viewer: viewer from parent, query viewer @ folder:f2#viewer).

  The schema-shape filters are necessary conditions only — they may over-report when no matching tuple is actually stored — but they will never miss a real divergence. A storage probe would be needed to fully confirm; we explicitly chose not to
  do that to keep the path off-the-hot-path-cheap. Hence the log message uses "potential v2Check resolution breaking change".

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured request IDs are consistently initialized for check-related warning logs, so request tracing is reliably present across success, fallback, and warning paths.
  * Enhanced warning logs for recognized userset-resolution breaking-change patterns by including store/model/request context and a computed divergence reason.

* **Tests**
  * Added a new unit test covering multiple authorization-model and tuple scenarios to verify breaking-change reason detection behavior across “no match” and recognized divergence cases.


pkg/server/check.go
```diff
@@ -113,10 +116,31 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
+
+			// Flag potential v2Check resolution breaking changes for userset requests.
+			// See breakingChangeReason for the scenarios we detect.
+			if !res.GetAllowed() && tuple.IsObjectRelation(req.GetTupleKey().GetUser()) {
+				if reason := breakingChangeReason(ctx, s, req, storeID); reason != "" {
+					requestID, _ := grpc_ctxtags.Extract(ctx).Values()["request_id"].(string)
+					if requestID == "" {
+						requestID = requestid.InitRequestID(ctx)
+					}
+					s.logger.InfoWithContext(ctx, "potential v2Check resolution breaking change: userset request returned false",
```
adriantam: Should this be warning or info?

pkg/server/check.go
```diff
@@ -423,6 +447,165 @@ func isV2TerminalError(err error) bool {
...
+		}
+		for _, dr := range directlyRelated {
+			if dr.GetType() == userObjectType {
+				return true
+			}
+		}
+		return nil
+	})
+	v, _ := result.(bool)
+	return v
```
adriantam: Curious why we do this instead of just returning the actual true/false itself in the above function?

pkg/server/check.go
```diff
@@ -113,10 +116,31 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
...
 			span.SetAttributes(attribute.Bool("allowed", res.GetAllowed()))
+
+			// Flag potential v2Check resolution breaking changes for userset requests.
+			// See breakingChangeReason for the scenarios we detect.
+			if !res.GetAllowed() && tuple.IsObjectRelation(req.GetTupleKey().GetUser()) {
+				if reason := breakingChangeReason(ctx, s, req, storeID); reason != "" {
+					requestID, _ := grpc_ctxtags.Extract(ctx).Values()["request_id"].(string)
+					if requestID == "" {
+						requestID = requestid.InitRequestID(ctx)
+					}
```
saad-h1, line 136: nit: this `requestID` code is duplicated a few lines below, maybe we pull it out of this and define it once?

adriantam: Wondering if we want to abstract the request ID into a function

pkg/server/check.go
```diff
@@ -423,6 +447,163 @@ func isV2TerminalError(err error) bool {
...
+	result, _ := typesystem.WalkUsersetRewrite(rewrite, func(r *openfgav1.Userset) interface{} {
+		if cu, ok := r.GetUserset().(*openfgav1.Userset_ComputedUserset); ok {
+			if cu.ComputedUserset.GetRelation() == relation {
+				return true
+			}
+		}
+		return nil
+	})
+	return result != nil && result.(bool)
+}
```
adriantam, line 513: Can we do something like this

```
func rewriteContainsComputedUserset(rewrite *openfgav1.Userset, relation string) bool {
        result, _ := typesystem.WalkUsersetRewrite(rewrite, func(r *openfgav1.Userset) interface{} {
                if cu, ok := r.GetUserset().(*openfgav1.Userset_ComputedUserset); ok && cu.ComputedUserset.GetRelation() == relation {
                        return true
                }
                return nil
        })
        return result != nil
  }
```
to simplify the logic?

Likewise for other functions

pkg/server/check.go
```diff
@@ -423,6 +447,163 @@ func isV2TerminalError(err error) bool {
...
+	tk := req.GetTupleKey()
+	if tk.GetUser() == tk.GetObject()+"#"+tk.GetRelation() {
+		return "self_referential_userset"
+	}
+	userObject, userRelation := tuple.SplitObjectRelation(tk.GetUser())
+	userObjectType := tuple.GetType(userObject)
+	targetObjectType := tuple.GetType(tk.GetObject())
+	targetRelation := tk.GetRelation()
+
+	typesys, err := s.resolveTypesystem(ctx, storeID, req.GetAuthorizationModelId())
```
saad-h1: nit: it looks like this is the only usage of `s *Server` in this function and it feels weird to pass the whole server struct for this, might be a bit cleaner to resolveTypesystem in `Check()` and pass it into this function instead

pkg/server/check.go
```diff
@@ -423,6 +447,163 @@ func isV2TerminalError(err error) bool {
...
+//     define viewer: viewer from parent
+//     With the query `document:d1, viewer, folder:f2#viewer` and the tuple `document:d1, parent, folder:f2`;
+//     v1 returned TRUE, v2 returns FALSE.
+//
+// All schema-shape filters are necessary conditions only — they may over-report when no
+// matching tuple is actually stored, but never miss a real divergence.
+func breakingChangeReason(ctx context.Context, s *Server, req *openfgav1.CheckRequest, storeID string) string {
+	tk := req.GetTupleKey()
+	if tk.GetUser() == tk.GetObject()+"#"+tk.GetRelation() {
+		return "self_referential_userset"
```
adriantam: Should these values be const so that the test will not drift if the name is changed?

pkg/server/check_test.go
```diff
@@ -939,3 +939,129 @@ func TestCheck_DoesNotFallBackOnInvalidContextualTuple(t *testing.T) {
...
+	})
+
+	tests := []struct {
+		name           string
+		modelDSL       string
+		seedTuple      *openfgav1.TupleKey
+		object         string
+		relation       string
+		user           string
+		expectedReason string
```
adriantam: Are we validating against this?

pkg/server/check_test.go
```diff
@@ -939,3 +939,129 @@ func TestCheck_DoesNotFallBackOnInvalidContextualTuple(t *testing.T) {
 	_, ok = tags[dispatchCountHistogramName]
 	require.False(t, ok, "dispatch_count must not be set — invalid contextual tuple must not fall back to v1")
 }
+
+func TestBreakingChangeReason(t *testing.T) {
+	t.Cleanup(func() {
+		goleak.VerifyNone(t)
+	})
+
+	tests := []struct {
```
adriantam, line 894: No negative test for the IsObjectRelation gate or the non-userset reasons

pkg/server/check.go
```diff
@@ -423,6 +444,163 @@ func isV2TerminalError(err error) bool {
 	return false
 }
 
+func getRequestIDFromContext(ctx context.Context) string {
+	requestID, _ := grpc_ctxtags.Extract(ctx).Values()["request_id"].(string)
+	if requestID == "" {
+		requestID = requestid.InitRequestID(ctx)
+	}
+	return requestID
+}
```
saad-h1, line 408: nit: should this be in some other file since it's not specific to check?

saad-h1 (dismissed):

pkg/server/check.go
```diff
@@ -423,6 +444,155 @@ func isV2TerminalError(err error) bool {
...
+				for _, dr := range directlyRelated {
+					if dr.GetType() == userObjectType {
+						return true
+					}
+				}
+			}
+		}
+		return nil
+	})
+	return result != nil && result.(bool)
```
adriantam:
```suggestion
        result, _ := typesystem.WalkUsersetRewrite(rewrite, func(r *openfgav1.Userset) interface{} {
                ttu, ok := r.GetUserset().(*openfgav1.Userset_TupleToUserset)
                if !ok || ttu.TupleToUserset.GetComputedUserset().GetRelation() != userRelation {
                        return nil
                }
                tuplesetRel := ttu.TupleToUserset.GetTupleset().GetRelation()
                // Loose type-only match (ignores relation/wildcard) is intentional: this is an
                // over-reporting necessary-condition filter. Do not swap in IsDirectlyRelated.
                directlyRelated, err := ts.GetDirectlyRelatedUserTypes(targetObjectType, tuplesetRel)
                if err == nil && slices.ContainsFunc(directlyRelated, func(dr *openfgav1.RelationReference) bool {
                        return dr.GetType() == userObjectType
                }) {
                        return true
                }
                return nil
        })
        return result != nil
```

(non-blocking) No need to assert for bool

adriantam (approved):

PR #3153, opened 2026-06-02
ci: make pr benchmark comparisons less fragile

justincoh:
## Description

Our PR benchmark action currently searches the action cache for `steps.get-main-branch-sha.outputs.sha`; aka the current main branch SHA's benchmark run. So, if a PR is opened shortly after `main` was updated and `main` hasn't completed running benchmarks, or if the most recent `main` branch benchmark run failed for any reason, the PR workflow will not find a benchmark run to compare against, and it will silently skip the comparison.

#### How is it being solved?
This PR changes the key used and implements github actions' `restore-keys` directive. The key is now structured as: `{OS}-{CPU}-{SHA}`. In the event a record is not available for that specific SHA, `restore-keys` will fall back and look for benchmark records which match on `{OS}-{CPU}`.

Because a `restore-keys` fallback hit sets `cache-hit` to `false`, the comparison steps now gate on `cache-matched-key` (which is populated on both exact and fallback hits) rather than `cache-hit` — otherwise the fallback baseline would be restored but the comparison would still be skipped.

When a fallback baseline is used, the job emits a `::notice::` indicating the comparison wasn't against the exact `main` commit. When *no* same-hardware baseline exists at all, the job now emits a `::warning::` and a job-summary note rather than passing silently — so a green check is never mistaken for "no regression". The job intentionally stays green in that case to avoid blocking merges on transient cache misses or brand-new runner hardware.

Note: a fallback comparison may be against an older `main` commit than the PR's actual parent. This is the intended trade to eliminate the silent skip, and the `::notice::` makes it transparent in the run logs.

## Summary by CodeRabbit

* **Chores**
  * Improved benchmark caching to prefer same-hardware baselines and reordered keys so restore fallbacks favor matching OS/CPU before commit; CI now only compares against a baseline when a same-hardware entry exists and emits a warning/summary when comparisons are skipped.

* **Documentation**
  * Updated release notes to describe the hardware-aware benchmark caching and comparison behavior.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -174,6 +174,7 @@ jobs:
         uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: ./cache/benchmark-data.json
-          # Save with commit hash to avoid "cache already exists"
-          # Save with OS & CPU info to prevent comparing against results from different CPUs
-          key: ${{ github.sha }}-${{ runner.os }}-${{ steps.system-info.outputs.cpu-model }}-go-benchmark
+          # Save with OS & CPU info first so PRs can fall back to the most recent
+          # same-hardware baseline via restore-keys; commit hash last keeps each
+          # entry unique and avoids "cache already exists".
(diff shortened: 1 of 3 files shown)


senojj:
The variable interpolation comments should be applied.

senojj (approved):

PR #3160, opened 2026-06-03
feat: Allow multiple audiences

Firgeis:
## Description
Allow multiple audiences to be configured for authentication.

It is a common use case to provide a list of audiences the service accepts (see  [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3))

#### How is it being solved?
Transform Authn.AuthnOIDCConfig to a string array. Each item in the incoming is parsed and sanitized then verified against this array. This change is backwards compatible.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * OIDC authentication accepts multiple audience values; tokens matching any configured audience are valid.
* **Behavior / CLI**
  * Server startup flag now supports a comma-separated list of OIDC audiences for configuration.
* **Compatibility**
  * Single audience strings remain accepted and are coerced to a one-element list.
* **Tests**
  * Added tests verifying scalar-to-list compatibility.

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -890,8 +890,8 @@
                     "x-env-variable": "OPENFGA_AUTHN_OIDC_ISSUER"
                 },
                 "audience": {
-                    "description": "one or more OIDC audiences that will be accepted as valid when verifying the `aud` field of the JWTs. A token is accepted if its `aud` matches any of the configured audiences.",
-                    "type": "array",
+                    "description": "one or more OIDC audiences that will be accepted as valid when verifying the `aud` field of the JWTs. A token is accepted if its `aud` matches any of the configured audiences. Accepts either a single string or a list of strings.",
+                    "type": ["string", "array"],
                     "items": {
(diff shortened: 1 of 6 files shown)


aaguiarz:
@Firgeis can you add a little more context? Why do you need more audiences?

Firgeis:
> @Firgeis can you add a little more context? Why do you need more audiences?

Our use case is the following: We have two different authentication schemes for end users and for M2M internal services. The end users audience is restricted to their client while the M2M tokens are broader, so they use different audiences.

aaguiarz:
@Firgeis interesting, are you planning to have tokens issued for end-users to authenticate to OpenFGA?

Firgeis:
> @Firgeis interesting, are you planning to have tokens issued for end-users to authenticate to OpenFGA?

In broad terms yes, they are used by a service to perform changes on behalf of the user.

curfew-marathon:

PR #3170, opened 2026-06-11
fix: reset visitedUsersetsMap and queryDedupeMap in shallowClone

tl-abar:
## Summary

`shallowClone` in `ReverseExpandQuery` was allocating a fresh `candidateObjectsMap` but leaving `visitedUsersetsMap` and `queryDedupeMap` pointing to the same maps as the original. When intersection/exclusion handlers fan out the clone and the original as concurrent goroutines, both goroutines race on those shared maps via `sync.Map.LoadOrStore`. The goroutine that loses the race sees the userset as already visited and skips branches it hasn't explored yet — producing non-deterministic, often empty, `ListObjects` results for any model using `and` or `but not`.

The fix mirrors what was already done for `candidateObjectsMap`: allocate a fresh `sync.Map` for each of the two affected fields in `shallowClone`.

**Affected versions**: present since `shallowClone` was introduced; confirmed unfixed through v1.17.1.

## Changes

- `pkg/server/commands/reverseexpand/reverse_expand.go`: allocate fresh `visitedUsersetsMap` and `queryDedupeMap` in `shallowClone`
- `pkg/server/commands/reverseexpand/reverse_expand_weighted_test.go`: add `TestShallowCloneIsolatesSharedMaps` regression test — 20 field objects × 100 iterations with `-race`, reliably reproducing the issue before the fix

## How to reproduce (before fix)

Authorization model with intersection:

```
model
  schema 1.1
type user
type workspace
  relations
    define ctx_member: [user]
type field
  relations
    define parent_workspace: [workspace]
    define can_read: [user] and ctx_member from parent_workspace
```

Repeated `ListObjects` calls for `user:alice` on `field#can_read` return the full list ~1 in 10 times and an empty list the rest of the time, with all caches disabled and a single-instance PostgreSQL backend. The `Openfga-Authorization-Model-Id` response header is identical across all calls, ruling out model-selection as the cause.

Example curl output before fix:
```
# Most calls (~9/10):
{"objects":[]}

# Occasional correct result (~1/10):
{"objects":["field:0","field:1","field:2",...,"field:19"]}
```

## Test plan

- [x] `go test -run TestShallowCloneIsolatesSharedMaps -race ./pkg/server/commands/reverseexpand/...` passes
- [x] `go vet ./pkg/server/commands/reverseexpand/...` clean
- [ ] `make test` full suite

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a concurrency issue in reverse-expand traversal so cloned queries no longer share mutable state between parallel branches, preventing nondeterministic or empty results.

* **Tests**
  * Added a regression test that repeatedly runs concurrent reverse-expand executions under weighted scenarios to verify deterministic, isolated behavior.

### DIFF
--- a/pkg/server/commands/reverseexpand/reverse_expand.go
+++ b/pkg/server/commands/reverseexpand/reverse_expand.go
@@ -250,13 +250,20 @@ func WithLogger(logger logger.Logger) ReverseExpandQueryOption {
 
 // shallowClone creates an identical copy of reverseExpandQuery except
 // candidateObjectsMap as list object candidates need to be validated
-// via check.
+// via check. visitedUsersetsMap and queryDedupeMap are also reset so that
+// concurrent intersection/exclusion traversals do not race on shared state
+// and skip branches they haven't explored yet. queryDedupeMap is
+// intentionally per-branch: sharing it across branches would cause one
+// branch to skip in-flight queries that only delivered results to the
+// other branch's channel, producing incomplete results.
 func (c *ReverseExpandQuery) shallowClone() *ReverseExpandQuery {
 	if c == nil {
 		return nil
 	}
 	copy := *c
 	copy.candidateObjectsMap = new(sync.Map)
+	copy.visitedUsersetsMap = new(sync.Map)
+	copy.queryDedupeMap = new(sync.Map)
 	return &copy
 }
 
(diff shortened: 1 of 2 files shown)


poovamraj:
We are closing this PR as we are not able to reproduce the issue mentioned here. Feel free to comment here to reopen the issue.

PR #3173, opened 2026-06-12
feat: add opt-in sorted results to Read, ReadUsersetTuples, and ReadStartingWithUser

saad-h1:

In experimental `weighted_graph_check` (v2Check), the weight2 pruning optimization requires tuples to be ordered. These tuples could come from a datastore `Read` call in the case of TTU, `ReadUsersetTuples` in the case of userset, and/or `ReadStartingWithUser` in the case of bottom_up. `ReadStartingWithUser` sorted unconditionally while `Read`/`ReadUsersetTuples` never sorted at all. We want these methods to return sorted results only when needed.

#### How is it being solved?

IsOrdered() now tells the truth end-to-end. Storage read methods gain an opt-in WithResultsSortedAscending flag; each iterator type stores and reports the actual ordering of its data. Callers in internal/check request sorted reads only when the weight2 strategy is reachable, so the pruning optimization can fire without paying for unnecessary ORDER BY on every read.

- Added WithResultsSortedAscending to ReadOptions and ReadUsersetTuplesOptions; made ReadStartingWithUser honour its existing flag (previously it always sorted)                                                                                 
- All backends (postgres, mysql, sqlite, memory) apply ORDER BY conditionally and pass the flag to the iterator constructor                                                                                                                      
- StaticTupleIterator, SQLTupleIterator, and the memory staticIterator store an ordered bool and return it from IsOrdered()                                                                                                                      
- CachingIterator skips the redundant in-memory sort when the inner iterator is already ordered; persists the ordered flag on cache entries                                                                                                      
- CombinedTupleReader explicitly sorts contextual tuples on sorted paths and uses the appropriate merge combiner                                                                                                                                 
- bottom_up.go unconditionally requests sorted reads (correctness requirement for the merge-sort algorithms it feeds)                                                                                                                            
- check.go requests sorted reads only when weight2 pruning is reachable: edge weight is 2, no contextual tuples present, and user is a concrete ID

## Review Checklist

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added opt-in result ordering for read operations to improve consistency and enable performance optimizations.

* **Performance**
  * Enabled weight2 pruning optimization when results are ordered.

* **Documentation**
  * Updated changelog documenting result ordering behavior and iterator ordering semantics.

### DIFF
--- a/internal/check/bottom_up_test.go
+++ b/internal/check/bottom_up_test.go
@@ -41,10 +41,11 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -95,10 +96,11 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -150,10 +152,11 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -207,10 +210,11 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{"xcond", ""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -266,6 +270,7 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
@@ -319,10 +324,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -373,10 +379,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -427,10 +434,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -483,10 +491,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{"xcond"},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil, false), nil)
 
 		model := testutils.MustTransformDSLToProtoWithID(`
 			model
@@ -542,6 +551,7 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			WithResultsSortedAscending: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
@@ -652,28 +662,28 @@ func TestBottomUpResolveUnion(t *testing.T) {
 		ctx := context.Background()
 
 		producer1 := make(chan *iterator.Msg, 3)
-		producer1 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:1"})}
-		producer1 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:5"})}
-		producer1 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:6"})}
+		producer1 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:1"}, false)}
+		producer1 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:5"}, false)}
+		producer1 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:6"}, false)}
 		close(producer1)
 		producers = append(producers, iterator.FromChannel(producer1))
 
 		producer2 := make(chan *iterator.Msg, 1)
-		producer2 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:2"})}
+		producer2 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:2"}, false)}
 		close(producer2)
 		producers = append(producers, iterator.FromChannel(producer2))
 
 		producer3 := make(chan *iterator.Msg, 4)
-		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:0"})}
-		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:3"})}
-		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:8"})}
-		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:9"})}
+		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:0"}, false)}
+		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:3"}, false)}
+		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:8"}, false)}
+		producer3 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:9"}, false)}
 		close(producer3)
 		producers = append(producers, iterator.FromChannel(producer3))
 
 		producer4 := make(chan *iterator.Msg, 2)
-		producer4 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:4"})}
-		producer4 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:8"})}
+		producer4 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:4"}, false)}
+		producer4 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:8"}, false)}
 		close(producer4)
 		producers = append(producers, iterator.FromChannel(producer4))
 
@@ -755,7 +765,7 @@ func TestBottomUpResolveUnion(t *testing.T) {
 
 				for _, objs := range tt.objects {
 					producer := make(chan *iterator.Msg, 1)
-					producer <- &iterator.Msg{Iter: storage.NewStaticIterator[string](objs)}
+					producer <- &iterator.Msg{Iter: storage.NewStaticIterator[string](objs, false)}
 					close(producer)
 					producers = append(producers, iterator.FromChannel(producer))
 				}
@@ -801,7 +811,7 @@ func TestBottomUpResolveUnion(t *testing.T) {
 			for j := 0; j < numItems; j++ {
 				keys = append(keys, "obj:"+strconv.Itoa(j))
 			}
-			producer <- &iterator.Msg{Iter: storage.NewStaticIterator[string](keys)}
+			producer <- &iterator.Msg{Iter: storage.NewStaticIterator[string](keys, false)}
 			close(producer)
 			producers = append(producers, iterator.FromChannel(producer))
 		}
@@ -846,7 +856,7 @@ func TestBottomUpResolveUnion(t *testing.T) {
 		producers = append(producers, iterator.FromChannel(producer1))
 
 		producer2 := make(chan *iterator.Msg, 1)
-		producer2 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:2"})}
+		producer2 <- &iterator.Msg{Iter: storage.NewStaticIterator[string]([]string{"obj:2"}, false)}
 		close(producer2)
 		producers = append(producers, iterator.FromChannel(producer2))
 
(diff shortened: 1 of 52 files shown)


pkg/storage/storagewrappers/cached_datastore.go
```diff
@@ -369,7 +369,7 @@ func (c *CachedDatastore) newCachedIterator(
 		tuplesCacheHitCounter.WithLabelValues(operation, c.method).Inc()
 		span.SetAttributes(attribute.Bool("cached", true))
 
-		staticIter := storage.NewStaticIterator[*storage.TupleRecord](cacheEntry.Tuples)
+		staticIter := storage.NewStaticIterator[*storage.TupleRecord](cacheEntry.Tuples, false)
```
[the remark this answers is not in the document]
saad-h1: `cached_datastore.go` is not used by v2Check, and v2Check is the only place where an iterator is checked to be `IsOrdered()`, so this value doesn't matter much. To fix it would require adding an `ordered` bool on the cache entry, but this doesn't seem necessary right now. Probably this code will be removed eventually anyway.

just added a comment to clarify instead

internal/graph/weight_two_resolver.go
```diff
@@ -292,7 +292,7 @@ func addNextItemInSliceStreamsToBatch(ctx context.Context, streamSlices []*itera
 		batch = append(batch, item)
 	}
 	if len(batch) > IteratorMinBatchThreshold {
-		concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch)}, outChan)
+		concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch, false)}, outChan)
 		batch = make([]string, 0)
```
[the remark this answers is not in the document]
saad-h1: will update, although this code is in the v1 Check path which never checks `IsOrdered` on iterators, so the `ordered` value doesn't matter much

internal/graph/weight_two_resolver.go
```diff
@@ -304,7 +304,7 @@ func fastPathUnion(ctx context.Context, streams *iterator.Streams, outChan chan<
 	defer func() {
 		// flush
 		if len(batch) > 0 {
-			concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch)}, outChan)
+			concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch, false)}, outChan)
 		}
```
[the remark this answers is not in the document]
saad-h1: will update, although this code is in the v1 Check path which never checks `IsOrdered` on iterators, so the `ordered` value doesn't matter much

internal/graph/weight_two_resolver.go
```diff
@@ -376,7 +376,7 @@ func fastPathIntersection(ctx context.Context, streams *iterator.Streams, outCha
...
```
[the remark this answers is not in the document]
saad-h1: will update, although this code is in the v1 Check path which never checks `IsOrdered` on iterators, so the `ordered` value doesn't matter much

internal/graph/weight_two_resolver.go
```diff
@@ -466,7 +466,7 @@ func fastPathDifference(ctx context.Context, streams *iterator.Streams, outChan
...
```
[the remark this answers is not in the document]
saad-h1: will update, although this code is in the v1 Check path which never checks `IsOrdered` on iterators, so the `ordered` value doesn't matter much

internal/graph/weight_two_resolver.go
```diff
@@ -567,7 +567,7 @@ func fastPathDifference(ctx context.Context, streams *iterator.Streams, outChan
 			}
 			batch = append(batch, items...)
 			if len(batch) > IteratorMinBatchThreshold {
-				concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch)}, outChan)
+				concurrency.TrySendThroughChannel(ctx, &iterator.Msg{Iter: storage.NewStaticIterator[string](batch, false)}, outChan)
 				batch = make([]string, 0)
```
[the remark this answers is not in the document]
saad-h1: will update, although this code is in the v1 Check path which never checks `IsOrdered` on iterators, so the `ordered` value doesn't matter much

internal/check/check_test.go
```diff
@@ -3995,9 +3995,10 @@ func TestSpecificTypeAndRelation(t *testing.T) {
 			storeID,
 			gomock.Any(),
 			storage.ReadUsersetTuplesOptions{
-				Consistency: storage.ConsistencyOptions{Preference: openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY},
+				WithResultsSortedAscending: true,
+				Consistency:                storage.ConsistencyOptions{Preference: openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY},
 			},
-		).Return(storage.NewStaticTupleIterator([]*openfgav1.Tuple{{Key: tuple.NewTupleKey("document:1", "viewer", "document:2#owner")}}), nil).Times(1)
+		).Return(storage.NewStaticTupleIterator([]*openfgav1.Tuple{{Key: tuple.NewTupleKey("document:1", "viewer", "document:2#owner")}}, false), nil).Times(1)
```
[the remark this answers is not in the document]
saad-h1: updated

pkg/storage/mysql/mysql.go
```diff
@@ -138,32 +138,41 @@ func (s *Datastore) Read(
...
 	defer span.End()
 
-	return s.read(ctx, store, filter, nil)
+	sb := s.buildTupleQuery(store, filter)
+	if options.WithResultsSortedAscending {
+		// COLLATE utf8mb4_bin forces bytewise ordering to match Go's string comparison (<).
+		// Without it, MySQL's default collation (typically case-insensitive) may produce a
+		// different sort order, breaking the weight2 pruning algorithm.
+		// Note: because _user is not defined as utf8mb4_bin at the column level, MySQL cannot
+		// use an index to satisfy this sort and will sort in memory.
```
saad-h1: todo: this will be taken care of in another in-progress effort

pkg/storage/storage.go
adriantam, line 1: Do we need new test to assert that the returned items are in ascending order when WithResultsSortedAscending == true?

saad-h1: woops thought I added tests but must have forgotten. added now in https://github.com/openfga/openfga/pull/3173/commits/572fe1eadbb9cfed3d31fe35767378eccb3266e9

Tests for Read and ReadUsersetTuples were added to pkg/storage/test/tuples.go, and ReadStartingWithUser already had tests in there. These tests get called by `RunAllTests` in pkg/storage/test/storage.go which all the storage backends call in their tests.

pkg/storage/test/tuples.go
```diff
@@ -1302,572 +1302,410 @@ func TupleWritingAndReadingTest(t *testing.T, datastore storage.OpenFGADatastore
 		require.ErrorIs(t, err, storage.ErrNotFound)
 	})
 
-	t.Run("reading_userset_tuples_that_exists_succeeds", func(t *testing.T) {
```
saad-h1, line 1305: these userset tests were under `TupleWritingAndReadingTest`, now just moved to a new `ReadUsersetTuplesTest` function below, which follows the pattern of ReadStartingWithUser & Read

look for `"enforce_order_of_tuples"` and `"assert_bytewise_ordering_of_tuples"` in this file for the newly added tests

pkg/storage/tuple_iterators.go
```diff
@@ -243,12 +234,10 @@ func (s *StaticIterator[T]) Head(ctx context.Context) (T, error) {
...
-// IsOrdered temporarily assumes the caller provided items in sorted order.
-// This will be made conditional when sources add explicit ordering guarantees.
-func (s *StaticIterator[T]) IsOrdered() bool { return true }
+func (s *StaticIterator[T]) IsOrdered() bool { return s.ordered }
 
-func NewStaticIterator[T any](items []T) Iterator[T] {
-	return &StaticIterator[T]{items: items, mu: &sync.Mutex{}}
+func NewStaticIterator[T any](items []T, ordered bool) Iterator[T] {
+	return &StaticIterator[T]{items: items, mu: &sync.Mutex{}, ordered: ordered}
 }
```
senojj: Boolean flags in a function signature are opaque and a historical source of bugs. It would be more expressive to create separate constructors for ordered and unordered iterators:
```go
func NewOrderedIterator[T any](items []T) Iterator[T] {
    return &StaticIterator[T]{items: items, ordered: true}
}

func NewUnorderedIterator[T any](items []T) Iterator[T] {
    return &StaticIterator[T]{items: items, ordered: false}
}
```

One comment that is outside the scope of this PR: It makes no sense for a *StaticIterator to have a field of *sync.Mutex (meaning there is no reason for the sync.Mutex to be a pointer since the StaticIterator is only ever used as a pointer).

senojj: Note that this also enables the use of the functions as arguments, since they share the same signature.

pkg/storage/tuple_iterators.go
```diff
@@ -191,8 +181,9 @@ func NewTupleKeyIteratorFromTupleIterator(iter TupleIterator) TupleKeyIterator {
...
 
 type StaticIterator[T any] struct {
-	items []T // GUARDED_BY(mu)
-	mu    *sync.Mutex
+	items   []T // GUARDED_BY(mu)
+	mu      *sync.Mutex
```
senojj: You touched it, so I'm going to add a comment here that I also added in a later comment:

> It makes no sense for a *StaticIterator to have a field of *sync.Mutex (meaning there is no reason for the sync.Mutex to be a pointer since the StaticIterator is only ever used as a pointer)

pkg/storage/tuple_iterators.go
```diff
@@ -130,23 +130,13 @@ func NewCombinedIterator[T any](iters ...Iterator[T]) Iterator[T] {
...
 
 // NewStaticTupleIterator returns a [TupleIterator] that iterates over the provided slice.
-func NewStaticTupleIterator(tuples []*openfgav1.Tuple) TupleIterator {
-	iter := &StaticIterator[*openfgav1.Tuple]{
-		items: tuples,
-		mu:    &sync.Mutex{},
-	}
-
-	return iter
+func NewStaticTupleIterator(tuples []*openfgav1.Tuple, ordered bool) TupleIterator {
```
senojj: The same idea that I conveyed on the `StaticIterator` applies to this constructor.

pkg/storage/tuple_iterators.go
```diff
@@ -130,23 +130,13 @@ func NewCombinedIterator[T any](iters ...Iterator[T]) Iterator[T] {
...
 
 // NewStaticTupleKeyIterator returns a [TupleKeyIterator] that iterates over the provided slice.
-func NewStaticTupleKeyIterator(tupleKeys []*openfgav1.TupleKey) TupleKeyIterator {
-	iter := &StaticIterator[*openfgav1.TupleKey]{
-		items: tupleKeys,
-		mu:    &sync.Mutex{},
-	}
-
-	return iter
+func NewStaticTupleKeyIterator(tupleKeys []*openfgav1.TupleKey, ordered bool) TupleKeyIterator {
```
senojj: The same idea that I conveyed on the `StaticIterator` applies to this constructor.

pkg/storage/storage.go
```diff
@@ -125,7 +126,8 @@ type ReadUserTupleOptions struct {
 // ReadUsersetTuplesOptions represents the options that can
 // be used with the ReadUsersetTuples method.
 type ReadUsersetTuplesOptions struct {
-	Consistency ConsistencyOptions
+	Consistency                ConsistencyOptions
+	WithResultsSortedAscending bool
```
senojj: would save us some typing if defined as `SortAsc bool`

pkg/storage/storage.go
```diff
@@ -113,7 +113,8 @@ type ConsistencyOptions struct {
 // ReadOptions represents the options that can
 // be used with the Read method.
 type ReadOptions struct {
-	Consistency ConsistencyOptions
+	Consistency                ConsistencyOptions
+	WithResultsSortedAscending bool
```
senojj: would save us some typing if defined as `SortAsc bool`

pkg/storage/tuple_iterators.go
saad-h1: updated

pkg/storage/tuple_iterators.go
saad-h1: updated

pkg/storage/tuple_iterators.go
saad-h1: updated

pkg/storage/tuple_iterators.go
saad-h1: updated

pkg/storage/storage.go
saad-h1: updated

pkg/storage/storage.go
saad-h1: updated

internal/check/check_test.go
```diff
@@ -4308,8 +4309,11 @@ func TestTTU(t *testing.T) {
...
 			},
-			storage.ReadOptions{Consistency: storage.ConsistencyOptions{Preference: openfgav1.ConsistencyPreference_UNSPECIFIED}},
-		).Return(storage.NewStaticTupleIterator([]*openfgav1.Tuple{{
+			storage.ReadOptions{
+				SortAsc:     true,
+				Consistency: storage.ConsistencyOptions{Preference: openfgav1.ConsistencyPreference_UNSPECIFIED},
+			},
+		).Return(storage.NewUnorderedStaticTupleIterator([]*openfgav1.Tuple{{
 			Key: tuple.NewTupleKey("document:1", "parent", "document:2"),
 		}}), nil).Times(1)
```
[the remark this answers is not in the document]
saad-h1: https://github.com/openfga/openfga/pull/3173/commits/a0f2cd583441b20660e0d8f549e338c4f257d71b

internal/check/bottom_up_test.go
```diff
@@ -41,10 +41,11 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			SortAsc: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewUnorderedStaticTupleIterator(nil), nil)
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -95,10 +96,11 @@ func TestBottomUpSpecificType(t *testing.T) {
...
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -150,10 +152,11 @@ func TestBottomUpSpecificType(t *testing.T) {
...
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -207,10 +210,11 @@ func TestBottomUpSpecificType(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:1"}},
 			Conditions: []string{"xcond", ""},
 		}, storage.ReadStartingWithUserOptions{
+			SortAsc: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewUnorderedStaticTupleIterator(nil), nil)
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -319,10 +324,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{""},
 		}, storage.ReadStartingWithUserOptions{
+			SortAsc: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewUnorderedStaticTupleIterator(nil), nil)
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -373,10 +379,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
...
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -427,10 +434,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
...
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
```diff
@@ -483,10 +491,11 @@ func TestBottomUpSpecificTypeWildcard(t *testing.T) {
 			UserFilter: []*openfgav1.ObjectRelation{{Object: "user:*"}},
 			Conditions: []string{"xcond"},
 		}, storage.ReadStartingWithUserOptions{
+			SortAsc: true,
 			Consistency: storage.ConsistencyOptions{
 				Preference: openfgav1.ConsistencyPreference_UNSPECIFIED,
 			}},
-		).MaxTimes(1).Return(storage.NewStaticTupleIterator(nil), nil)
+		).MaxTimes(1).Return(storage.NewUnorderedStaticTupleIterator(nil), nil)
```
Vic-Dev:
```suggestion
		).MaxTimes(1).Return(storage.NewOrderedStaticTupleIterator(nil), nil)
```

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

internal/check/bottom_up_test.go
saad-h1: updated

pkg/storage/mysql/mysql.go
saad-h1: fixed by https://github.com/openfga/openfga/commit/a2e0dbefc3e01a95c785f81a3563bc6571b08b11

pkg/storage/mysql/mysql.go
```diff
@@ -342,16 +333,19 @@ func (s *Datastore) ReadUsersetTuples(
 	if len(filter.Conditions) > 0 {
 		sb = sb.Where(sq.Eq{"COALESCE(condition_name, '')": filter.Conditions})
 	}
+	if options.SortAsc {
+		sb = sb.OrderBy("_user COLLATE utf8mb4_bin")
```
senojj: columns are now collated this way. adding collation here is unnecessary and could have a negative impact on query performance.

saad-h1: updated

pkg/storage/postgres/postgres.go
```diff
@@ -765,13 +760,16 @@ func (s *Datastore) ReadUsersetTuples(
 	if len(filter.Conditions) > 0 {
 		sb = sb.Where(sq.Eq{"COALESCE(condition_name, '')": filter.Conditions})
 	}
+	if options.SortAsc {
+		sb = sb.OrderBy("_user collate \"C\"")
```
senojj, line 764: I don't think there exists an index where `_user` is collated using "C", so sorting this way will not be efficient. Note that such an index _does_ exist for the `object_id` column.

We should consider setting the default collation of all columns to "C" -- like we did for MySQL -- or creating new indexes.

pkg/storage/storagewrappers/cached_reader.go
```diff
@@ -227,10 +227,13 @@ func (c *CachedTupleReader) ReadStartingWithUser(
...
 
 // tryGetFromCache checks for cache hit with invalidation support.
-// Returns LockFreeCachedIterator if found and not invalidated.
+// Returns LockFreeCachedIterator if found, not invalidated, and ordered-compatible.
+// Monotonic upgrade: a sorted entry satisfies unsorted requests, but an unsorted entry
+// does not satisfy a sorted request (returns nil so the caller re-queries and upgrades).
 func (c *CachedTupleReader) tryGetFromCache(
 	cacheKey keys.Key, storeID, objectType, relation, operation string,
 	invalidEntityKeys []keys.Key,
+	wantSorted bool,
```
senojj, line 236: I would rather that we built the ordering of the iterator into the cache key, than do something like this. Though I don't think that two places use the same cache key with different ordering; please confirm.

pkg/storage/sqlite/sqlite.go
```diff
@@ -152,32 +152,36 @@ func (s *Datastore) Read(
...
+	options storage.ReadOptions,
 ) (storage.TupleIterator, error) {
-	ctx, span := startTrace(ctx, "Read")
+	_, span := startTrace(ctx, "Read")
 	defer span.End()
 
-	return s.read(ctx, store, filter, nil)
+	sb := s.buildTupleQuery(store, filter)
+	if options.SortAsc {
+		sb = sb.OrderBy("user_object_type", "user_object_id", "user_relation")
```
saad-h1, line 162: Just realized this will not work because the weight2 algorithm compares the full user string, but for these 2 users:
1. `user:bob`
2. `user2:bob`

The full user string comparison will determine `user2:bob` < `user:bob` (since `2` < `:`), but this 3-column ORDER BY will determine `user:bob` < `user2:bob` (since it first compares `user_object_type`, in which case `user` < `user2`).

Need to think about this more.

saad-h1:
Decided not to move forward with this atm.

The [weight2 optimization](https://github.com/openfga/openfga/blob/171806c93b86bca29e0212ceb8b6ee9c48eb9ac3/internal/check/weight2.go#L177) is pretty much just a memory optimization, not a performance optimization. The `bottom_up` sorting is always on ReadStartingWithUser which already sorts today (on object ID). Given that it's requiring large DB migrations and code refactoring to sort on user, plus technically sorting would affect performance, this doesn't seem worth it atm.

PR #3178, opened 2026-06-19
chore: create draft release and publish after provenance succeeds for immutable tags/releases

### LINKED ISSUE #3172: Enable immutable tags/releases
### Checklist

GoReleaser currently publishes the GitHub release immediately, and SLSA provenance artifacts are attached afterwards. That creates a window where a release is public but provenance is missing; if provenance generation later fails, the release remains public and incomplete. It also prevents turning on GitHub's immutable releases setting (which locks assets and tags at publish time) because provenance uploads would fail against a locked release.

Change the release flow to create a draft GitHub release, upload artifacts and provenance to that draft, then publish (undraft) only after provenance jobs succeed. Concretely:

1) set `release.draft: true` in `.goreleaser.yaml` (optionally `use_existing_draft: true`),
2) change the `binary-provenance` job to pass `upload-tag-name: ${{ github.ref_name }}` and `draft-release: true`,
3) add an `undraft-release` job that `needs` both provenance jobs and uses the shared `undraft-release.yml`,
4) make SLSA verification depend on the undraft step (it downloads from the published release),
5) after a successful release with the new flow, enable GitHub immutable releases for the repo.

Keep `verification-with-cosign` and Snyk behavior unchanged. 

I'm happy to open a PR implementing these workflow edits.

Alternatives considered:

- Leave the current flow (status quo): releases remain briefly public before provenance and risk incomplete public releases.
- Enable immutable releases without changing the workflow: provenance uploads would fail against a locked release.
- Adopt release-please/full automation for tagging and draft pre-creation: possible future enhancement but out of scope for this change.

### Additional context

Related resources:

- CLI implementation ([reference](https://github.com/openfga/cli/blob/main/.github/workflows/main.yaml)): `.goreleaser.yaml` and workflows in `openfga/cli` follow this pattern. 
- Shared undraft workflow: [`openfga/.github/.github/workflows/undraft-release.yml`. ](https://github.com/openfga/.github/blob/main/.github/workflows/undraft-release.yml)

SoulPancake:
Mark the GitHub release as a draft in goreleaser, upload SLSA provenance to the draft, and undraft only after both provenance jobs pass. This closes the window where a release is public without provenance and unblocks enabling GitHub immutable releases.

## Review Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated the release pipeline to publish GitHub releases as drafts by default before they’re finalized.
  * Improved provenance publication by generating provenance separately and attaching it to the release for the current tag.
  * Added an explicit step to undraft/complete the release before running downstream verification.
  * Adjusted workflow permissions and execution flow to ensure verification occurs after the draft is converted to its final state.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -149,7 +149,9 @@ jobs:
     with:
       base64-subjects-as-file: "${{ needs.goreleaser.outputs.subjects-as-file }}"
       provenance-name: "openfga.intoto.jsonl"
-      upload-assets: true # upload to a new release
+      upload-assets: true # upload to the draft release created by goreleaser
+      upload-tag-name: ${{ github.ref_name }}
+      draft-release: true
 
   image-provenance:
     needs: [ goreleaser ]
@@ -167,8 +169,14 @@ jobs:
       registry-username: ${{ secrets.DOCKERHUB_USERNAME }}
       registry-password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+  undraft-release:
+    needs: [ binary-provenance, image-provenance ]
+    permissions:
+      contents: write
+    uses: openfga/.github/.github/workflows/undraft-release.yml@15d8dd0e50827f4233abbbc3c394413940ba8db4 # pin@main
+
   verification-with-slsa-verifier:
-    needs: [ goreleaser, binary-provenance ]
+    needs: [ undraft-release, binary-provenance ]
     runs-on: ubuntu-latest
     permissions: read-all
     steps:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -149,9 +149,32 @@ jobs:
     with:
       base64-subjects-as-file: "${{ needs.goreleaser.outputs.subjects-as-file }}"
       provenance-name: "openfga.intoto.jsonl"
-      upload-assets: true # upload to the draft release created by goreleaser
-      upload-tag-name: ${{ github.ref_name }}
-      draft-release: "true"
+      # Don't let the generator attach the provenance:
+      # - it uses softprops/action-gh-release@v2.2.1, whose paginated draft lookup
+      #   creates a duplicate draft once a repo has >100 releases (2 pages)
+      # - the provenance is still published as a workflow artifact, so the
+      #   release-provenance job attaches it via `gh release upload` (by tag)
+      upload-assets: false
+
+  release-provenance:
+    needs: [ binary-provenance ]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write # To add the provenance asset to the draft release.
+    steps:
+      - name: Download the provenance
+        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+        with:
+          name: ${{ needs.binary-provenance.outputs.provenance-name }}
+
+      - name: Attach provenance to the draft release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PROVENANCE: ${{ needs.binary-provenance.outputs.provenance-name }}
+        run: |
+          set -euo pipefail
+          gh release upload "$GITHUB_REF_NAME" "$PROVENANCE" \
+            --repo "$GITHUB_REPOSITORY" --clobber
 
   image-provenance:
     needs: [ goreleaser ]
@@ -170,7 +193,7 @@ jobs:
       registry-password: ${{ secrets.DOCKERHUB_TOKEN }}
 
   undraft-release:
-    needs: [ binary-provenance, image-provenance ]
+    needs: [ release-provenance, image-provenance ]
     permissions:
       contents: write
     uses: openfga/.github/.github/workflows/undraft-release.yml@15d8dd0e50827f4233abbbc3c394413940ba8db4 # pin@main
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -142,7 +142,7 @@ jobs:
     permissions:
       actions: read # To read the workflow path.
       id-token: write # To sign the provenance.
-      contents: write # To add assets to a release.
+      contents: read
 
     # Note: this _must_ be referenced by tag. See: https://github.com/slsa-framework/slsa-verifier/issues/12
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v2.1.0
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -151,7 +151,7 @@ jobs:
       provenance-name: "openfga.intoto.jsonl"
       upload-assets: true # upload to the draft release created by goreleaser
       upload-tag-name: ${{ github.ref_name }}
-      draft-release: true
+      draft-release: "true"
 
   image-provenance:
     needs: [ goreleaser ]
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -167,8 +169,14 @@ jobs:
...
       registry-password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+  undraft-release:
+    needs: [ binary-provenance, image-provenance ]
+    permissions:
+      contents: write
+    uses: openfga/.github/.github/workflows/undraft-release.yml@15d8dd0e50827f4233abbbc3c394413940ba8db4 # pin@main
+
   verification-with-slsa-verifier:
-    needs: [ goreleaser, binary-provenance ]
```
SoulPancake, line 171: This wasn't required because `binary-provenance` already "needs" `goreleaser`

.github/workflows/release.yaml
```diff
@@ -149,7 +149,9 @@ jobs:
     with:
       base64-subjects-as-file: "${{ needs.goreleaser.outputs.subjects-as-file }}"
       provenance-name: "openfga.intoto.jsonl"
-      upload-assets: true # upload to a new release
+      upload-assets: true # upload to the draft release created by goreleaser
+      upload-tag-name: ${{ github.ref_name }}
+      draft-release: true
```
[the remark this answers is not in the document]
SoulPancake: In practice it doesn't fail - the CLI passes this unquoted and has shipped releases fine. But the docs say the input type should match the declared type: string, so quoting it!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 
 ### Changed
 - Extended experimental `weighted_graph_check` to `BatchCheck`: when the flag is enabled, each item in the batch is evaluated using the weighted graph algorithm, with per-item fallback to the standard algorithm on non-terminal errors. [#3154](https://github.com/openfga/openfga/pull/3154)
+- Release flow now creates the GitHub release as a draft, attaches SLSA provenance, and publishes only after provenance jobs succeed. This avoids public releases with missing provenance and enables GitHub immutable releases. [#3178](https://github.com/openfga/openfga/pull/3178)
```
rhamzeh: We don't need a changelog item for this as it is internal.

Tag the PR with `skip-changelog`

SoulPancake: Done!

.github/workflows/release.yaml
```diff
@@ -149,7 +149,9 @@ jobs:
     with:
       base64-subjects-as-file: "${{ needs.goreleaser.outputs.subjects-as-file }}"
       provenance-name: "openfga.intoto.jsonl"
-      upload-assets: true # upload to a new release
+      upload-assets: true # upload to the draft release created by goreleaser
+      upload-tag-name: ${{ github.ref_name }}
+      draft-release: "true"
```
adriantam: I think this has a correctness problem that's specific to openfga, and it'll bite exactly on the next release rather than some edge case.

The setup: binary-provenance now passes draft-release: "true" and upload-tag-name. Inside slsa-github-generator@v2.1.0, that flows into softprops/action-gh-release@v2.2.1, which has to locate our draft so it can attach provenance to it. The catch is in how it searches:

```
let _release = undefined;
if (config.input_draft) {                  // true here
  for await (const response of releaser.allReleases({owner, repo})) {  // per_page: 100
    _release = response.data.find((r) => r.tag_name === tag);  // reassigned every page
    }
  }
  if (_release === undefined) { return createRelease(...); }  // makes a second draft
```

 _release gets overwritten on every page instead of stopping once it finds a match. So the value that survives is whatever the last page returned, not the page where the draft actually was.

Here's why that breaks for us specifically. The releases API returns newest-first by created_at, and we're over 100 releases, so the list spans two pages:

```
gh api "repos/openfga/openfga/releases?per_page=100&page=1" --jq 'length'   # 100
gh api "repos/openfga/openfga/releases?per_page=100&page=2" --jq 'length'   # 13
```

A new release makes that 114, still two pages. The fresh draft sorts to the very top (page 1, first item), but the loop keeps going to page 2, finds nothing, and resets _release to undefined. So it falls through to createRelease() and we end up with two drafts on the same tag — goreleaser's binaries and SBOMs on one, the provenance on the other. Then undraft-release runs gh release edit "$REF_NAME" --draft=false, and which draft it publishes is anyone's guess. We could ship a release missing either the binaries or the provenance, which is the exact thing this PR is trying to prevent.

The higher release number doesn't save us, it's actually the problem — the newest release is the one that lands on page 1 and gets clobbered by the page 2 pass. The only thing that would mask this is having 100 or fewer releases total, and we crossed that line a while ago.

  Source links:
  - The reassignment loop: https://github.com/softprops/action-gh-release/blob/c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda/src/github.ts#L231-L246
  - Falls through to createRelease: https://github.com/softprops/action-gh-release/blob/c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda/src/github.ts#L248
  - allReleases paginates at per_page: 100: https://github.com/softprops/action-gh-release/blob/c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda/src/github.ts#L132-L139
  - Where the generator calls this action with draft + tag_name: https://github.com/slsa-framework/slsa-github-generator/blob/v2.1.0/.github/workflows/generator_generic_slsa3.yml#L288-L292
  - API ordering (newest created_at first, drafts included for users with push access): https://docs.github.com/en/rest/releases/releases#list-releases

  One honest caveat: I traced this from the source, I didn't run it on a live tag. Worth confirming with a real v* tag on a fork before we trust it (heads up that test* tags won't trigger this workflow because of the "!test*" filter). The comment right above the loop — // you can't get an existing draft by tag — tells me the author knew drafts can't be fetched by tag and added the loop on purpose; they just used reassignment instead of a break, which is the bug.

SoulPancake: Great catch
It does seem like a nasty bug in action-gh-release
Which we might even face in CLI when we reach 100+ releases
They don't seem to bump the version in slsa-github-generator either, There was another workaround which seemed heavy but seems like the correct thing to use, Will switch to [that](https://github.com/slsa-framework/slsa-github-generator/issues/1476#issuecomment-4510380352) logic

SoulPancake:
Moving to draft until I update to the download/upload flow

.github/workflows/release.yaml
```diff
@@ -167,8 +192,14 @@ jobs:
       registry-username: ${{ secrets.DOCKERHUB_USERNAME }}
       registry-password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+  undraft-release:
+    needs: [ release-provenance, image-provenance ]
+    permissions:
+      contents: write
+    uses: openfga/.github/.github/workflows/undraft-release.yml@15d8dd0e50827f4233abbbc3c394413940ba8db4 # pin@main
```
adriantam, line 200: [non-blocking] do we intend to pin the particular release or main?

adriantam (dismissed):

.github/workflows/release.yaml
SoulPancake: Usually I've pointed it to main on other repos because we were iterating and improving the common workflows.
But I think it's a relatively stable one for this particular part of the workflow so pinning it should be okay
Will eventually pin all the other SDKs as well

SoulPancake:
@adriantam 
Sorry, one last thing, added a timeout for the release-provenance step

SoulPancake:
Known issue with it
https://github.com/slsa-framework/slsa-github-generator/issues/2044#issuecomment-1519289250

justincoh (approved):

PR #3181, opened 2026-06-23
fix: match IPv4-mapped IPv6 addresses in the in_cidr condition

kanywst:

`in_cidr` returns different results for two textual forms of the same host. An IPv4-mapped IPv6 address such as `::ffff:10.1.2.3` is not matched against an IPv4 CIDR like `10.0.0.0/8`, while the dotted form `10.1.2.3` is. Per RFC 4291 section 2.5.5.2 the two are the same host. `ParseIPAddress` kept the address as `netip.ParseAddr` returned it, so the mapped form stays a different address family from an IPv4 prefix and `netip.Prefix.Contains` returns false.

#### How is it being solved?

- Unmap the address at parse time in `ParseIPAddress` (`addr.Unmap()`), so the mapped and dotted forms compare equal against an IPv4 CIDR.
- Add two cases to `TestIPaddressCELBinaryBinding` for the mapped form (in and out of an IPv4 CIDR). The in-cidr case fails before the change and passes after, and `internal/condition/...` stays green.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `in_cidr` evaluation so IPv4-mapped IPv6 inputs (e.g., `::ffff:192.168.1.1`) and IPv4-mapped CIDRs are treated as their IPv4 equivalents, producing correct CIDR range matches.
* **Tests**
  * Expanded coverage for IPv4-mapped IPv6 addresses and CIDRs, verifying both matching and non-matching scenarios.
* **Documentation**
  * Updated the **Unreleased** changelog to document the IPv4-mapped IPv6 behavior for `in_cidr`.

### DIFF
--- a/internal/condition/types/ipaddress.go
+++ b/internal/condition/types/ipaddress.go
@@ -64,7 +64,8 @@ func ParseIPAddress(ip string) (IPAddress, error) {
 		return IPAddress{}, err
 	}
 
-	return IPAddress{addr}, nil
+	// Unmap so an IPv4-mapped IPv6 address matches an IPv4 CIDR.
+	return IPAddress{addr.Unmap()}, nil
 }
 
 // ipaddrCelType defines a CEL type for the IPAddress and registers it as a receiver type.
@@ -145,6 +146,13 @@ func ipaddressCELBinaryBinding(lhs, rhs ref.Val) ref.Val {
 	ipaddr, ok := lhs.(IPAddress)
 	if !ok {
 		return types.NewErr("an IPAddress parameter value is required for comparison")
+	}
+
+	// Unmap the CIDR too so an IPv4-mapped IPv6 CIDR matches the unmapped address.
+	addr := network.Addr().Unmap()
+	bits := network.Bits()
+	if addr.Is4() && bits >= 96 {
+		network = netip.PrefixFrom(addr, bits-96)
 	}
 
 	return types.Bool(network.Contains(ipaddr.addr))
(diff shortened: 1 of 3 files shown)


adriantam:
Hello @kanywst , thank you for the PR.  Copilot added a few comments -  can you review them?  Also, can you updated the changelog as this is a behaviour change that needs to be called out in the changelog.  Thank you!

kanywst:
@adriantam 

Thanks! Pushed an update.

internal/condition/types/ipaddress.go
```diff
@@ -147,6 +148,12 @@ func ipaddressCELBinaryBinding(lhs, rhs ref.Val) ref.Val {
 		return types.NewErr("an IPAddress parameter value is required for comparison")
 	}
 
+	// Unmap an IPv4-mapped IPv6 CIDR to its IPv4 form so it matches the unmapped
+	// address (RFC 4291 2.5.5.2); a /N IPv6 prefix maps to a /(N-96) IPv4 prefix.
+	if network.Addr().Is4In6() && network.Bits() >= 96 {
```
senojj: `(*Addr).Is4In6()` guarantees that the network bit length is greater than or equal to 96 when it returns `true`.

kanywst: @senojj 

`Is4In6()` is a property of the address, independent of the prefix length. `netip.MustParsePrefix("::ffff:1.2.3.4/64")` returns `Is4In6() == true` with `Bits() == 64`, so it doesn't imply the `Bits() >= 96` check. `cidr` is user-supplied, so dropping that check makes `Bits() - 96` negative and `netip.PrefixFrom` returns an invalid prefix.

internal/condition/types/ipaddress.go
```diff
@@ -147,6 +148,12 @@ func ipaddressCELBinaryBinding(lhs, rhs ref.Val) ref.Val {
...
+		network = netip.PrefixFrom(network.Addr().Unmap(), network.Bits()-96)
+	}
```
senojj: Currently, you call `network.Addr().Is4In6()` explicitly, and then call `network.Addr().Unmap()` which also calls `(*Addr).Is4In6()`.

You can avoid duplicate evaluations by using the following logic:
```go
addr := network.Addr().Unmap()
bits := network.Bits()
if addr.Is4() && bits >= 96 {
	network = netip.PrefixFrom(addr, bits-96)
}
```

kanywst: @senojj 
Applied, thanks.

internal/condition/types/ipaddress.go
senojj: I apologize, as my assumption was incorrect. While a mask of 64 on an IPv6 mapped IPv4 address doesn't make any sense, the `netip.ParsePrefix` function accepts it as a valid prefix, and the `(*Addr).Is4In6()` function has no context of the network bits.

senojj (approved):

PR #3184, opened 2026-06-26
chore: Consolidate caching metrics in v2Check with v1

saad-h1:

The experimental `weighted_graph_check` engine had no aggregate visibility into its edge cache effectiveness. Cache hits were recorded only as per-request OpenTelemetry span events, making it impossible to dashboard or alert on hit rate across the fleet — a gap compared to v1's `check_cache_*` Prometheus counters.

Also, the v2 iterator cache used different metric names from v1 - when v2 is enabled, this could cause downstream observability tooling to break.

#### How is it being solved?

v2Check uses the same metrics as v1 Check, so that no downstream observability tooling updates are needed.

v2's edge caching is analogous to v1's subproblem caching, and that metric is incremented in v2Check inside `Resolver.isCached` where the three cache outcomes (attempted lookup, valid hit, invalidated hit) are already distinguished by existing control flow. No new logic is introduced.

See files changed

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added weighted graph check cache metrics for total lookup attempts, valid cache hits, and stale-entry discards.
  * Enhanced cache instrumentation to attribute iterator/tuple and check cache activity by API method, including discard and cache entry size tracking.

* **Documentation**
  * Updated the changelog’s **[Unreleased]** section to document the new weighted graph check metrics.

* **Tests**
  * Updated caching iterator unit tests and benchmarks to use the new method-aware iterator construction and metric labeling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 ## [Unreleased]
 ### Added
+- Added metrics in experimental `weighted_graph_check` for weighted graph edge result caching.
 - Added diagnostic logging in experimental `weighted_graph_check` when v2 Check resolution might produce a different result than v1 for the same query. These logs surface authorization models that may be affected by a future v1 deprecation, and no operator action is required. [#3149](https://github.com/openfga/openfga/pull/3149)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 Try to keep listed changes to a concise bulleted list of simple explanations of changes. Aim for the amount of information needed so that readers can understand where they would look in the codebase to investigate the changes' implementation, or where they would look in the documentation to understand how to make use of the change in practice - better yet, link directly to the docs and provide detailed information there. Only elaborate if doing so is required to avoid breaking changes or experimental features from ruining someone's day.
 
 ## [Unreleased]
+### Changed
+- Matched experimental `weighted_graph_check` cache metrics with original Check cache metrics: iterator cache metrics renamed to `tuples_cache_total_count`, `tuples_cache_hit_count`, `tuples_cache_discard_count`, `tuples_cache_size`; query cache metrics added as `check_cache_total_count`, `check_cache_hit_count`, `check_cache_invalid_hit_count`. [#3184](https://github.com/openfga/openfga/pull/3184)
 
 ## [1.18.1] - 2026-06-29
 ### Added
-- Added metrics in experimental `weighted_graph_check` for weighted graph edge result caching. [#3184](https://github.com/openfga/openfga/pull/3184)
 - Added diagnostic logging in experimental `weighted_graph_check` when v2 Check resolution might produce a different result than v1 for the same query. These logs surface authorization models that may be affected by a future v1 deprecation, and no operator action is required. [#3149](https://github.com/openfga/openfga/pull/3149)
 - Added diagnostic logging in `Expand` and `ListUsers` when v2 resolution might produce a different result than v1 for the same query. Note that v2 has not been implemented yet for these endpoints; these logs purely add visibility for a future v1 deprecation, and no operator action is required. [#3182](https://github.com/openfga/openfga/pull/3182)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of 
 
 ## [Unreleased]
 ### Added
-- Added metrics in experimental `weighted_graph_check` for weighted graph edge result caching.
+- Added metrics in experimental `weighted_graph_check` for weighted graph edge result caching. [#3184](https://github.com/openfga/openfga/pull/3184)
 - Added diagnostic logging in experimental `weighted_graph_check` when v2 Check resolution might produce a different result than v1 for the same query. These logs surface authorization models that may be affected by a future v1 deprecation, and no operator action is required. [#3149](https://github.com/openfga/openfga/pull/3149)
 
 ### Changed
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Try to keep listed changes to a concise bulleted list of simple explanations of
 
 ## [Unreleased]
 ### Added
+- Added metrics in experimental `weighted_graph_check` for weighted graph edge result caching.
 - Added diagnostic logging in experimental `weighted_graph_check` when v2 Check resolution might produce a different result than v1 for the same query. These logs surface authorization models that may be affected by a future v1 deprecation, and no operator action is required. [#3149](https://github.com/openfga/openfga/pull/3149)
```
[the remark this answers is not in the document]
saad-h1: already added in latest commit

senojj:
Remove "v2" from the metric names.

internal/check/check.go
```diff
@@ -158,6 +181,7 @@ func (r *Resolver) isCached(consistency openfgav1.ConsistencyPreference, key key
 	if consistency == openfgav1.ConsistencyPreference_HIGHER_CONSISTENCY {
 		return nil, false
 	}
+	checkV2CacheTotalCounter.Inc()
```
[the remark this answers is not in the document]
saad-h1: I think this is okay for now, the only metric that is being incremented is the number of "cache lookups" which is technically true even with a no-op cache, and I think it's easy to understand seeing 0 hits when cache is disabled.

Eventually, the way the cache is set up in the code will be reworked and metrics will be reported more cleanly then.

internal/check/check.go
```diff
@@ -30,6 +33,26 @@ import (
 
 var tracer = otel.Tracer("internal/check")
 
+var (
+	checkV2CacheTotalCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: build.ProjectName,
+		Name:      "check_v2_edge_cache_total_count",
+		Help:      "The total number of edge cache lookups attempted by v2 check (excluding HIGHER_CONSISTENCY requests).",
+	})
```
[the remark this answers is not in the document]
saad-h1: the help message is accurate

internal/graph/cached_resolver.go
```diff
@@ -4,15 +4,13 @@ import (
...
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promauto"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
 
-	"github.com/openfga/openfga/internal/build"
+	"github.com/openfga/openfga/internal/checkmetrics"
```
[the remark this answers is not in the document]
saad-h1: woops did a `git commit -a` but that doesn't include new files, added the missing file in https://github.com/openfga/openfga/pull/3184/commits/8c67385b5de46a402d3b48f6dac7d348caa5ec50

internal/checkmetrics/checkmetrics.go
```diff
@@ -0,0 +1,30 @@
+// Package checkmetrics holds Prometheus counters shared across the v1 and v2
+// Check cache implementations so both paths report into the same metric names.
+package checkmetrics
```
saad-h1: We decided v1 and v2 Check will share the same metric names - that way, when v2 Check is enabled, any observability tooling (e.g., dashboards/alerts) that have been set up on the old metric names don't need changing. Since we don't need shadowing anymore, the two will not run together and so it's not an issue that they're sharing metric names.

I had to move these to a separate file since you can't call `promauto.NewCounter(...)` with the same metric name twice (will panic); i.e., it had to be initialized in one location that both v1 and v2 Check must then get it from.

I could have made v2 import `internal/graph/<wherever this was originally defined>` but seemed a bit messy for v2Check to import v1Check.

pkg/storage/storagewrappers/iterator_cache.go
```diff
@@ -36,33 +33,6 @@ const (
...
-	}, []string{"operation"})
-
-	v2IterCacheSize = promauto.NewHistogramVec(prometheus.HistogramOpts{
-		Namespace: build.ProjectName,
-		Name:      "v2_iterator_cache_entry_size",
-		Help:      "Number of tuples in cached iterator entries.",
-		Buckets:   []float64{1, 10, 50, 100, 250, 500, 1000},
-	}, []string{"operation"})
-)
-
```
saad-h1, line 65: We decided v1 and v2 Check will share the same metric names - that way, when v2 Check is enabled, any observability tooling (e.g., dashboards/alerts) that have been set up on the old metric names don't need changing. Since we don't need shadowing anymore, the two will not run together and so it's not an issue that they're sharing metric names.

These metrics were removed and metric reporting in the v2 iterator cache now uses the same v1 iterator cache metrics.

internal/check/response.go
```diff
@@ -27,3 +27,7 @@ type ResponseCacheEntry struct {
 	LastModified time.Time
 	Res          *Response
 }
+
+func (e *ResponseCacheEntry) CacheEntityType() string {
+	return "check_response"
+}
```
saad-h1, line 33: mimics v1:
https://github.com/openfga/openfga/blob/46b1d53b99af0e8b310fe90c3c1b1682b1a15eb9/internal/graph/cached_resolver.go#L54

required to fulfill the CacheItem interface, which is used for cache metrics:
https://github.com/openfga/openfga/blob/46b1d53b99af0e8b310fe90c3c1b1682b1a15eb9/pkg/storage/cache.go#L57-L59

internal/checkmetrics/checkmetrics.go
senojj: any reason not to make it `internal/check/metrics`?

saad-h1: ah no that sounds good as well, updated

Vic-Dev (approved):

PR #3199, opened 2026-07-08
test: deflake reducer cancellation subtests in internal/graph

### LINKED ISSUE #3197: Flaky test: TestUnionCheckFuncReducer cancellation subtest depends on wall-clock timing
### Checklist

### Describe the bug

`TestUnionCheckFuncReducer/return_allowed:true_if_truthy_handler_evaluated_before_context_cancelled` in `internal/graph` can fail when the test process is starved of CPU for more than 10ms. Observed twice locally under heavy system load:

```
--- FAIL: TestUnionCheckFuncReducer (0.10s)
    --- FAIL: TestUnionCheckFuncReducer/return_allowed:true_if_truthy_handler_evaluated_before_context_cancelled (0.01s)
```

It passes reliably on an unloaded machine (50 consecutive isolated `-race` runs with no failure), so this is a narrow race — but the kind that oversubscribed CI runners occasionally hit, and the fix is cheap.

### Root cause

The subtest (`internal/graph/check_test.go`) encodes a required ordering ("handler evaluated **before** cancel") using a wall-clock sleep, which no scheduler guarantees:

```go
go func() {
    defer wg.Done()
    time.Sleep(10 * time.Millisecond)
    cancel()
}()

resp, err := union(ctx, concurrencyLimit, trueHandler, trueHandler)
require.NoError(t, err)
require.True(t, resp.GetAllowed())
```

If the test goroutine is descheduled for more than 10ms, the cancel fires before `union`'s select loop reads the first handler outcome, and `union` returns `context.Canceled`:

```go
select {
case <-ctx.Done():
    return nil, ctx.Err()          // <- flaky path
case outcome, ok := <-out:
    ...
    if outcome.resp.Allowed {
        return outcome.resp, nil   // <- intended path
    }
}
```

There is also a deeper wrinkle: when both channels are ready simultaneously (outcome already buffered AND context already cancelled), Go's `select` chooses randomly, so `union`'s result for that ordering is nondeterministic by design. The test name asserts an ordering that the implementation does not actually prioritize.

### Possible fixes (maintainer input welcome)

1. **Test-only fix (conservative):** make the ordering deterministic — e.g. cancel only after `union` returns, or gate the cancel goroutine on the handlers having completed. Keeps `union` behavior unchanged; the subtest then genuinely tests "handler result wins when it arrives first".
2. **Behavior fix (bigger):** bias `union` to drain ready outcomes before honoring `ctx.Done()` (non-blocking read of `out` first). This would make "short-circuit success beats concurrent cancellation" a guarantee rather than a race, but changes reducer semantics and deserves its own discussion.

Happy to send a PR for option 1 (and/or option 2 if preferred).

### Expectation

The subtest passes regardless of scheduler timing.

### Environment

- OpenFGA `main` @ 69efbd95
- go1.26.4, linux/amd64 (WSL2)

kevglynn:
## Description

Closes #3197

`TestUnionCheckFuncReducer/return_allowed:true_if_truthy_handler_evaluated_before_context_cancelled` can fail when the test process is starved of CPU for more than 10ms (details and root cause in #3197). The subtest encodes a required ordering — "handler evaluated **before** cancel" — with a `time.Sleep(10 * time.Millisecond)` before `cancel()`, which no scheduler guarantees.

The same sleep-then-cancel pattern appears at **9 sites** across `TestExclusionCheckFuncReducer`, `TestIntersectionCheckFuncReducer`, and `TestUnionCheckFuncReducer` in `internal/graph/check_test.go`, so this PR deflakes the whole pattern rather than just the one subtest that was observed failing. This is the **test-only fix (option 1)** from #3197 — no reducer behavior changes.

The 9 subtests fall into two shapes, fixed with two techniques:

**1. "Handler result wins" subtests (6 sites)** — e.g. `return_allowed:true_if_truthy_handler_evaluated_before_context_cancelled`. These assert that a handler outcome resolved before cancellation is returned. The fix moves `cancel()` to after the reducer returns, which makes the ordering the test name promises *guaranteed* instead of probabilistic:

```go
resp, err := union(ctx, concurrencyLimit, trueHandler, trueHandler)
require.NoError(t, err)
require.True(t, resp.GetAllowed())

// Cancel only after the reducer has resolved: a concurrent cancel races
// the reducer's select between handler outcomes and ctx.Done() (#3197).
cancel()
```

**2. "Cancellation wins" subtests (3 sites)** — the `return_error_if_context_cancelled_before_resolution` variants. These previously relied on a 40ms margin (cancel at 10ms vs. 50ms handler sleeps). The fix removes all sleeps: handlers block on a `release` channel, and the cancel goroutine fires only once a handler is running. At the moment of cancellation the outcome channel is guaranteed empty and open, so the reducer's `select` deterministically observes `ctx.Done()`:

```go
handlerStarted := make(chan struct{})
release := make(chan struct{})
blockedTrueHandler := func(context.Context) (*ResolveCheckResponse, error) {
    close(handlerStarted)
    <-release
    return &ResolveCheckResponse{Allowed: true}, nil
}

go func() {
    <-handlerStarted
    cancel()
}()

resp, err := union(ctx, concurrencyLimit, blockedTrueHandler)
require.ErrorIs(t, err, context.Canceled)
require.Nil(t, resp)

close(release) // let the blocked handlers drain before goleak runs
```

Because the handler start happens strictly after the reducer's fast-path `ctx.Err()` check, the cancellation always lands mid-`select`, preserving each test's original intent (mid-flight cancellation, not the already-covered "parent context cancelled before call" fast path).

### Notes for reviewers

- The gated cancel deliberately avoids the interleaving where `concurrency.TrySendThroughChannel` drops an outcome on a cancelled context and the reducer's closed-channel receive races `ctx.Done()` — with the handlers still blocked at cancel time, that interleaving cannot occur.
- A handful of sibling subtests use 10ms context *deadlines* with a similar (but much narrower) timing assumption. They are left untouched to keep this diff focused on the observed flake pattern; happy to follow up if there's interest.

## References

- Fixes #3197
- Related: #3061 fixed a different flaky test with a similar timing-assumption root cause

## Review Checklist

- [x] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork).
- [x] I have added documentation for new/changed functionality in this PR — N/A (test-only change; behavior documented in test comments).
- [x] All active GitHub checks for tests, formatting, and security are passing

## Testing

- `go test ./internal/graph/ -run 'Test(Union|Intersection|Exclusion)CheckFuncReducer' -race -count=50` — pass
- `GOMAXPROCS=1 go test ./internal/graph/ -run 'Test(Union|Intersection|Exclusion)CheckFuncReducer' -race -count=100` — pass (single-P scheduling maximizes the original flake's likelihood)
- `go test ./internal/graph/ -race` (full package) — pass
- `golangci-lint run ./internal/graph/...` — 0 issues

---

*Disclosure: this change was developed with AI assistance (Cursor); all analysis and code were human-reviewed before submission.*

## Summary by CodeRabbit

* **Tests**
  * Made cancellation-related reducer tests deterministic by removing timing-based sleeps and reducing race windows.
  * Reworked cancellation scenarios to use explicit synchronization, ensuring `context.Canceled` is observed in a predictable order relative to handler execution.
  * Improved coverage for cases where handlers complete before or after cancellation, reducing flaky test behavior and strengthening reliability.

### DIFF
--- a/internal/graph/check_test.go
+++ b/internal/graph/check_test.go
@@ -393,19 +393,6 @@ func TestExclusionCheckFuncReducer(t *testing.T) {
 		cancel()
 	})
 
-	t.Run("return_allowed:false_if_subtract_handler_evaluated_before_context_cancelled", func(t *testing.T) {
-		ctx, cancel := context.WithCancel(context.Background())
-		t.Cleanup(cancel)
-
-		resp, err := exclusion(ctx, concurrencyLimit, trueHandler, trueHandler)
-		require.NoError(t, err)
-		require.False(t, resp.GetAllowed())
-
-		// Cancel only after the reducer has resolved: a concurrent cancel races
-		// the reducer's select between handler outcomes and ctx.Done() (#3197).
-		cancel()
-	})
-
 	t.Run("return_context_canceled_when_parent_context_is_canceled", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
 		cancel()
--- a/internal/graph/check_test.go
+++ b/internal/graph/check_test.go
@@ -348,47 +348,31 @@ func TestExclusionCheckFuncReducer(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
 		t.Cleanup(cancel)
 
-		var wg sync.WaitGroup
-
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			time.Sleep(10 * time.Millisecond)
-			cancel()
-		}()
-
 		resp, err := exclusion(ctx, concurrencyLimit, falseHandler, falseHandler)
 		require.NoError(t, err)
 		require.False(t, resp.GetAllowed())
 
-		wg.Wait() // just to make sure to avoid test leaks
+		// Cancel only after the reducer has resolved: a concurrent cancel races
+		// the reducer's select between handler outcomes and ctx.Done() (#3197).
+		cancel()
 	})
 
 	t.Run("return_allowed:false_if_sub_handler_evaluated_before_context_cancelled", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
 		t.Cleanup(cancel)
 
-		var wg sync.WaitGroup
-
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-			time.Sleep(10 * time.Millisecond)
-			cancel()
-		}()
-
 		resp, err := exclusion(ctx, concurrencyLimit, trueHandler, trueHandler)
 		require.NoError(t, err)
 		require.False(t, resp.GetAllowed())
 
-		wg.Wait() // just to make sure to avoid test leaks
+		// Cancel only after the reducer has resolved: a concurrent cancel races
+		// the reducer's select between handler outcomes and ctx.Done() (#3197).
+		cancel()
 	})
 
 	t.Run("return_allowed:false_if_sub_handler_evaluated_before_base_cancelled", func(t *testing.T) {
 		ctx, cancel := context.WithCancel(context.Background())
 		t.Cleanup(cancel)
-
-		var wg sync.WaitGroup
 
 		slowTrueHandler := func(context.Context) (*ResolveCheckResponse, error) {
 			time.Sleep(50 * time.Millisecond)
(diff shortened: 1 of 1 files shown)


kevglynn:
Friendly bump — bot/security checks are all green, but the main **Pull Request** workflow is still `action_required`. Could a maintainer approve workflow runs when convenient?

Updates since last push:
- Rebased onto latest `main`
- Removed duplicate `return_allowed:false_if_subtract_handler_evaluated_before_context_cancelled` subtest (identical to the `sub_handler` variant after the deflake rewrite — addresses CodeRabbit nitpick)

Local verification: `-race -count=50` on the affected reducer tests, full `internal/graph` package `-race`, scoped lint clean.

Thanks for reviewing!

PR #3213, opened 2026-07-18
fix: retry write when on_duplicate:ignore hits a first-time insert race

### LINKED ISSUE #3201: on_duplicate: "ignore" returns 409 on concurrent first-time writes of the same tuple
### Checklist

### Description

When using on_duplicate: "ignore" in the Write API, concurrent requests that write the same new tuple (one that doesn't yet exist in the database) return 409 errors for all but one of the requests.

The on_duplicate: "ignore" option works correctly when the tuple already exists and is committed in the database — only the concurrent first-time insert case fails.

Steps to Reproduce

1. Start OpenFGA with a Postgres datastore
2. Send 5 parallel Write requests for the same tuple with on_duplicate: "ignore"
3. The tuple does not exist in the store before the requests are made

Expected Behavior

All 5 requests succeed (or are silently ignored) since on_duplicate: "ignore" is set.

Actual Behavior

One request succeeds. The remaining 4 return 409 Conflict.

Reproduction

[Reproduction repo](https://github.com/rdeanmcdonald/openfga-duplicate-write-race-condition)

Environment

- OpenFGA server: v1.17.0, v1.18.0
- @openfga/sdk: 0.9.3
- Datastore: Postgres 16

Additional Context

When the same tuple is written and committed before the parallel requests are made, on_duplicate: "ignore" correctly returns success for all concurrent duplicate writes (0 failures across 10 consecutive runs). The issue only manifests when the tuple does not yet exist at the time the concurrent writes begin.

### OpenFGA version

v0.18.0

As a binary

Postgres

### OpenFGA Flags

No

### Logs

_No response_

TheRealShek:
Fixes #3201.

`on_duplicate: "ignore"` works fine once a tuple exists — but for a brand new tuple, concurrent writes race each other and all but one come back 409 instead of being ignored.

We lock existing rows with `SELECT ... FOR UPDATE` before insert, but that can't lock a row that isn't there yet. So under READ COMMITTED, concurrent first-time inserts all pass the check, all try to insert, one wins, the rest hit the unique constraint and get turned into a 409 regardless of `on_duplicate`.

Fix: retry the write (max 3x) when `ignore` is set and we hit that specific conflict. By then the winning transaction has committed, so the retry sees the row and skips it like it should have. Non-ignore writes and real condition conflicts aren't retried.

Did this in the command layer instead of all three storage backends — felt like a retry-policy call rather than a storage concern, but open to moving it if preferred.

Added two tests (retry-then-succeed, retries-then-409), and confirmed against real Postgres using the repro from the issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved writes using the “ignore duplicates” option by automatically retrying transient insert conflicts.
  * Improved deletes using the “ignore missing” option by automatically retrying transient delete conflicts.
  * Added bounded retry behavior for persistent conflicts, returning a clear aborted failure after the limit is reached.
* **Tests**
  * Added table-driven coverage for both “ignore duplicates” and “ignore missing” retry behavior.
  * Added a regression test for concurrent “ignore duplicates” to verify only one tuple and one changelog entry are created under contention.

### DIFF
--- a/pkg/server/commands/write.go
+++ b/pkg/server/commands/write.go
@@ -20,6 +20,13 @@ import (
 	"github.com/openfga/openfga/pkg/typesystem"
 )
 
+// maxWriteConflictRetries bounds how many times Execute re-attempts a write that failed with
+// storage.ErrWriteConflictOnInsert under on_duplicate:ignore. Because the conflicting transaction
+// has already committed by the time the uniqueness violation surfaces, a single retry is enough
+// to observe the committed row and ignore it; the extra attempts are a safety margin for
+// datastores with weaker insert-time blocking.
+const maxWriteConflictRetries = 3
+
 // WriteCommand is used to Write and Delete tuples. Instances may be safely shared by multiple goroutines.
 type WriteCommand struct {
 	logger                    logger.Logger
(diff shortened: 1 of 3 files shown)


pkg/server/commands/write.go
```diff
@@ -93,14 +100,31 @@ func (c *WriteCommand) Execute(ctx context.Context, req *openfgav1.WriteRequest)
...
+
+	// With on_duplicate:ignore, concurrent first-time inserts of the same tuple can still collide
+	// at INSERT time: the datastore locks tuples with SELECT ... FOR UPDATE, which cannot lock a
+	// row that does not yet exist, so every racing transaction attempts the INSERT and all but one
+	// fail the uniqueness constraint with ErrWriteConflictOnInsert. Retrying lets the losers observe
+	// the now-committed row and ignore it (or surface a condition conflict if it was written with a
+	// different condition).
+	for retries := 0; onDuplicateInsert == storage.OnDuplicateInsertIgnore &&
+		errors.Is(err, storage.ErrWriteConflictOnInsert) &&
+		retries < maxWriteConflictRetries; retries++ {
```
adriantam: If we go with this approach, we should also handle the OnEmptyDeleteIgnore case

TheRealShek: Good point. I'll extend the handling to cover OnEmptyDeleteIgnore as well for consistency.

TheRealShek:
@adriantam when you get a little time, can you review the changes.

PR #3218, opened 2026-07-21
fix: validate tuple condition against the correct type-restriction facet

poovamraj:

`validateCondition` in `internal/validation/validation.go` decides whether a conditioned relationship tuple is permitted by the authorization model. When a relation is directly assignable by the same user type under more than one facet — a concrete user (`user`), a typed wildcard (`user:*`), or a userset (`group#member`) — and only one of those facets carries a condition, the validator only compared the tuple's user type and the condition name. It did not compare the facet.

As a result, for a relation such as `define viewer: [user, user:* with cond]`, a tuple like `document:1#viewer@user:alice` carrying `cond` was accepted even though `cond` is only defined on the `user:*` facet. The no-condition branch of the same function already screens the facet; the conditioned branch did not.

Because this validation also runs when reading tuples during query resolution (`ValidateTupleForRead` / `FilterInvalidTuples`), the mismatch could also cause Check and ListObjects to treat such a tuple inconsistently. Applying the facet check makes both write-time validation and read-time filtering reject these tuples consistently.

This change:
- Makes the conditioned branch of `validateCondition` screen the wildcard/userset facet, matching the existing no-condition branch.
- Adds `TestValidateConditionFacetMismatch` covering concrete↔wildcard mismatches in both directions, the userset facet, and the valid cases.

## Review Checklist
- [x] I have clicked on "allow edits by maintainers".
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to openfga.dev

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced stricter conditioned-tuple validation by requiring the condition binding to match the facet implied by the tuple’s user shape and the model restriction tied to the condition name.
  * Facet/condition mismatches are now consistently rejected across tuple reads, authorization queries, tuple writes, and listing APIs.
* **Tests**
  * Added unit coverage for facet mismatch behavior in conditioned tuples, including aligned vs mismatched condition bindings.
  * Added an ABAC regression case: `condition_bound_to_correct_type_restriction_facet` (including error code `2027` for mismatches).
* **Documentation**
  * Updated the **[Unreleased]** changelog with the stricter validation behavior.


internal/validation/validation_test.go
```diff
@@ -767,6 +769,116 @@ func TestValidateTupleForWrite(t *testing.T) {
...
+	// only be accepted when the matching restriction's facet (concrete / typed-wildcard /
+	// userset) also matches the tuple's user shape.
+	condCtx, err := structpb.NewStruct(map[string]interface{}{"ok": true})
+	require.NoError(t, err)
+
+	tests := []struct {
+		name    string
+		model   string
+		tuple   *openfgav1.TupleKey
+		wantErr bool
```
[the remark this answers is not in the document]
poovamraj: Fixed now

poovamraj:
@coderabbitai review

internal/validation/validation.go
```diff
@@ -233,10 +233,27 @@ func validateCondition(typesys *typesystem.TypeSystem, tk *openfgav1.TupleKey) e
...
+			if directlyRelatedType.GetRelation() != "" && directlyRelatedType.GetRelation() != userRelation {
+				continue
+			}
+
+			if directlyRelatedType.GetWildcard() != nil && !tuple.IsTypedWildcard(tk.GetUser()) {
+				continue
+			}
+		} else if tuple.IsTypedWildcard(tk.GetUser()) || userRelation != "" {
+			continue
+		}
```
adriantam: This logic looks very similar to line 201 - 211

poovamraj: Moved both of them into it's own method to avoid duplication and added tests to it as well

adriantam (approved):

PR #3210, opened 2026-07-16
feat: add `weighted_graph_check` potential resolution changes logging to BatchCheck

Vic-Dev:
#### What problem is being solved?

  Extends the experimental `weighted_graph_check` diagnostic logging to `BatchCheck`,
  mirroring what's already in place for Check (#3149, #3204). WARN-level telemetry only —
  no change to authorization results, no operator action required.

  With the flag enabled, BatchCheck emits a single aggregate log per request listing the                                                                                                                                          
  distinct v1→v2 divergence shapes found across the batch, plus a `reasons_by_correlation_id`
  map so operators can pinpoint the affected items. Detection reuses the shared `v2breaking`
  package and is memoized per shape, so identical checks in a batch are walked at most once.

  Shape predicates read the model, not tuples, so they may over-report but never miss a real
  divergence.

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

## Summary by CodeRabbit

* **New Features**
  * Added additional experimental warning diagnostics for batch checks using weighted-graph evaluation.
  * When enabled, each batch emits a single aggregated warning listing distinct v2 resolution-divergence reasons and the correlation IDs they apply to (excluding non-breaking/error outcomes).

* **Tests**
  * Expanded log-observation coverage for feature-flag gating, reason deduplication/memoization, suppression rules, and behavior when weighted-graph model resolution fails.

* **Documentation**
  * Updated the unreleased changelog entry to reflect the extended experimental batch diagnostic logging behavior.

PR #3182, opened 2026-06-24
feat: add potential v2 breaking change logs for Expand and ListUsers

Vic-Dev:
#### What problem is being solved?
Adds observability for potential v1 → v2 (weighted-graph) resolution breaking changes on the ListUsers and Expand endpoints, mirroring the Check-side logs added in #3149.

Note: Expand has no difference in behaviour for `self_referential_userset ` from v1 to v2.

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

## Summary by CodeRabbit

* **New Features**
  * Added structured diagnostic warning logs for potential v2 (weighted-graph) resolution breaking changes in **Expand** and **ListUsers**, including a computed `reason`.
  * Introduced centralized logic to compute and confirm divergence reasons for `Check`, `Expand`, and `ListUsers`.
* **Bug Fixes**
  * Aligned `Check`’s breaking-change warning reason computation for userset-shaped queries.
* **Documentation**
  * Updated the changelog under **Unreleased** to reflect the new diagnostic logging.
* **Tests**
  * Added/updated end-to-end tests validating the warning message and that the logged `reason` matches expected outcomes.

PR #3223, opened 2026-07-23
feat: add logs for v2 potential resolution changes to ListObjects

Vic-Dev:
 Extends the experimental v1→v2 (weighted-graph) diagnostic logging to ListObjects, mirroring what's already in place for Check (#3149, #3204), BatchCheck (#3210), Expand and ListUsers (#3182). WARN-level telemetry only — no change to
  authorization results, no operator action required.

  When a ListObjects request shape matches a known v1→v2 divergence, the server emits a single WARN log naming the reason. The ListObjects subject plays the same "user side" role the filter plays in ListUsers: a userset subject
  (object#relation) exercises the four per-userset shapes plus userset_with_exclusion and wildcard_with_exclusion. Detection reuses the shared v2breaking package.

  The log is additionally gated on a response-confirmation check (ListObjectsResponseConfirmsReason), which suppresses shape matches whose response didn't observably exercise the divergent v1 path. This lets the tests run with the v2 pipeline
  enabled: today the pipeline punts userset/wildcard subjects to the legacy algorithm so the divergent object is present and the log fires; once the pipeline is made consistent with v2 the object drops out, the log stops firing, and the tests
  fail — signalling they can be removed.

  Shape predicates read the model, not tuples, so they may over-report but never miss a real divergence.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added warnings for `ListObjects` requests that may encounter known v1-to-v2 permission resolution differences.
  * Warnings include relevant request and model details to support investigation.
  * Requests missing required condition context now return an error without returning objects.

* **Documentation**
  * Updated the changelog with release 1.18.3 details and unreleased fixes.


pkg/server/commands/v2breaking/v2breaking.go
```diff
@@ -193,6 +193,88 @@ func ListUsersReason(typesys *typesystem.TypeSystem, object *openfgav1.Object, r
...
+// Returns true for unknown reasons, so the caller's logic stays simple.
+func ListObjectsResponseConfirmsReason(reason, subject string, objects []string) bool {
+	switch reason {
+	case ReasonSelfReferentialUserset, ReasonComputedUsersetSelfObj:
+		subjectObject, _ := tuple.SplitObjectRelation(subject)
+		return slices.Contains(objects, subjectObject)
+	case ReasonAliasUserset, ReasonTTUUserset, ReasonUsersetWithExclusion, ReasonWildcardWithExclusion:
+		return len(objects) > 0
+	}
+	return true
```
adriantam: If default is false, then you would not need to filter `reason != ""` in line 198 of [list_objects.go](https://github.com/openfga/openfga/pull/3223/changes#diff-ccbb3ac4bfa47dc7fdbb0264e2391c6d17d67af222ce2a013dc3acde9921eb91)

adriantam (dismissed):

adriantam (approved):

PR #3226, opened 2026-07-27
fix: return the requested bucket count and exact endpoints from LinearBuckets

akshita317:
## Description

`utils.LinearBuckets` accumulates the bucket width into the `float64` loop variable:

```go
for i := minValue; i <= maxValue; i += width {
    buckets = append(buckets, i)
}
```

Rounding error compounds across iterations, so when `width` is not exactly representable in binary floating point the loop can terminate one step early and never emit `maxValue`.

This is not theoretical — it affects the only production call site. `internal/condition/metrics/metrics.go:38` requests:

```go
Buckets: utils.LinearBuckets(0, config.DefaultMaxConditionEvaluationCost, 10)
```

`width` is `100/9`, and the accumulated value at the last step lands just above `100`, so the function returns:

```
len=9  [0 11.111… 22.222… 33.333… 44.444… 55.555… 66.666… 77.777… 88.888…]
```

9 buckets topping out at ~88.89, instead of 10 buckets ending at 100. **Every condition evaluation costing between ~88.89 and the 100 maximum is recorded in `+Inf` rather than in a real bucket**, so the top of the `condition_evaluation_cost` histogram has no resolution across the range operators are most likely to care about.

A non-positive `count` was also unsafe: it yields a negative `width`, `i` moves away from `maxValue`, and the loop never terminates.

## Changes

- Compute each boundary as `minValue + width*float64(i)` instead of accumulating, so error does not compound.
- Assign `buckets[0]` and `buckets[count-1]` directly, so the closed interval `[min…max]` documented on the function is exact rather than approximate.
- Return `nil` for `count <= 0` and `[]float64{minValue}` for `count == 1`, replacing the infinite loop and the division by zero.

## Tests

The existing test only covered `LinearBuckets(1, 100, 10)`, where the width is exactly `11` and no rounding occurs — which is why this went unnoticed. That case is kept as-is and still passes.

`TestLinearBuckets` is now a table that pins the bucket count, both endpoints, and monotonicity, including widths that are not exactly representable (`(0, 100, 10)` and `(0, 0.3, 4)`) plus the `count` edge cases.

Verified the new cases fail against the previous implementation: the `(0, 100, 10)` case fails with `should have 9 item(s), but has 10`, and the negative-`count` case hangs until the test binary is killed.

```
$ go test ./internal/utils/... ./internal/condition/...
ok  github.com/openfga/openfga/internal/utils
ok  github.com/openfga/openfga/internal/condition
ok  github.com/openfga/openfga/internal/condition/eval
ok  github.com/openfga/openfga/internal/condition/types
```

## Note for reviewers

This changes the emitted bucket boundaries for `condition_evaluation_cost`, since the histogram now has its intended 10 buckets and a real `100` boundary. That is the point of the fix, but it will show up as a one-time shape change in existing dashboards for that metric.

A CHANGELOG entry has been added under `Unreleased → Fixed`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved linear bucket generation for invalid and edge-case inputs, with accurate counts and endpoints.
  * Fixed potential deadlocks in user listing and security issues in streamed object listing.
  * Corrected authorization model validation behavior.

* **New Features**
  * Added an option to override the gRPC service label.
  * Added diagnostic logging for selected authorization checks and listing operations.

* **Chores**
  * Updated Go and gRPC health probe security versions.

* **Tests**
  * Expanded bucket-generation coverage for lengths, values, ordering, and endpoints.


akshita317:
Rebased onto `main` to clear the merge conflict — this is now mergeable again.

The conflict was in `CHANGELOG.md` only; the code commit applied cleanly. Worth flagging how I resolved it: my entry originally sat in the `### Fixed` list of what was then the pending release section, but `1.18.2` and `1.18.3` have both shipped since. Rather than keep it in a released section, I took upstream's entry for `1.18.2` unchanged and moved mine to `## [Unreleased]`. Happy to move it if you would rather it sit elsewhere.

No functional change since the last review pass. For reference, the behaviour this fixes:

```
OLD LinearBuckets(0,100,10): len=9  last=88.8889
NEW LinearBuckets(0,100,10): len=10 last=100.0000
```

`go build`, `go vet` and `go test ./internal/utils/...` all pass.

akshita317:
Rebased onto `main` to clear the merge conflict.

The conflict was in `CHANGELOG.md` only. Since this PR was opened, 1.19.0 and 1.20.0 have shipped, so my entry was sitting inside a section that has since been released. It is now under `## [Unreleased]` in a new `### Fixed` block, which is where it belongs for an unreleased fix.

No code changes: `internal/utils/` is byte-identical to the previous head (`94197695`), and `go test ./internal/utils/` passes on the rebased tree. The branch is now 0 commits behind `main`.

PR #3230, opened 2026-07-27
feat: add logs for v2 potential resolution changes to StreamedListObjects

Vic-Dev:
  Extends the experimental v1→v2 (weighted-graph) diagnostic logging to `StreamedListObjects`, the streaming counterpart of the unary `ListObjects` logging added in #3223 (which itself mirrors Check (#3149, #3204), BatchCheck (#3210), Expand
  and ListUsers (#3182)). WARN-level telemetry only — no change to authorization results, no operator action required.
  
  When a `StreamedListObjects` request shape matches a known v1→v2 divergence, the server emits a single WARN log naming the reason. Detection reuses the shared `v2breaking` package (`ListObjectsReason` + `ListObjectsResponseConfirmsReason`),
  so behaviour is identical to the unary endpoint.
  
  The one difference is *where* the log lives. The `StreamedListObjects` handler never sees the streamed objects — they're sent to the client from inside `ExecuteStreamed` (the command layer) — so the response-confirmation check must run there
  too. To keep the streaming path bounded (this endpoint returns all results and ignores `listObjectsMaxResults`), we don't buffer the streamed objects: the shape reason is computed once up front, then each streamed object is folded into a
  single confirmation flag as it's sent. This is provably equivalent to running the confirmation check once over the full response, since confirmation is monotonic (`slices.Contains` for the self-object reasons, `len > 0` for the rest). The log
  is emitted once, at the end of a successful stream, matching the unary handler's success-only semantics.
  
  Shape predicates read the model, not tuples, so they may over-report but never miss a real divergence.

## Review Checklist

## Summary by CodeRabbit

- **Diagnostics**
  - Added conditional Info-level logging for `ListObjects` requests whose results may differ between API versions.
  - Logs include request, store, model, and detected divergence details to support troubleshooting.
  - Streamed requests log only after the response confirms the detected condition and only when Info-level logging is enabled.

- **Bug Fixes**
  - Improved consistency in detecting and confirming potential resolution differences across unary and streamed `ListObjects` requests.


pkg/logger/logger.go
```diff
@@ -12,6 +12,11 @@ import (
 )
 
 type Logger interface {
+	// Level reports the minimum enabled log level for this logger. Callers can
+	// use it to skip expensive work that only feeds a log line which would be
+	// dropped anyway (e.g. l.Level() <= zapcore.InfoLevel).
+	Level() zapcore.Level
```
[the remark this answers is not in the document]
Vic-Dev: Why?

pkg/server/commands/v2breaking/v2breaking.go
```diff
@@ -193,6 +193,88 @@ func ListUsersReason(typesys *typesystem.TypeSystem, object *openfgav1.Object, r
...
+// Returns true for unknown reasons, so the caller's logic stays simple.
+func ListObjectsResponseConfirmsReason(reason, subject string, objects []string) bool {
+	switch reason {
+	case ReasonSelfReferentialUserset, ReasonComputedUsersetSelfObj:
+		subjectObject, _ := tuple.SplitObjectRelation(subject)
+		return slices.Contains(objects, subjectObject)
+	case ReasonAliasUserset, ReasonTTUUserset, ReasonUsersetWithExclusion, ReasonWildcardWithExclusion:
+		return len(objects) > 0
+	}
+	return true
```
adriantam: See https://github.com/openfga/openfga/pull/3223/changes#r3897401067

pkg/server/commands/list_objects.go
```diff
@@ -744,6 +747,30 @@ func (q *ListObjectsQuery) ExecuteStreamed(ctx context.Context, req *openfgav1.S
...
+	if q.logger.Level() <= zap.InfoLevel {
+		breakingChangeReason = v2breaking.ListObjectsReason(typesys, targetObjectType, targetRelation, req.GetUser())
+	}
+	var breakingChangeConfirmed bool
+	confirmBreakingChange := func(object string) {
+		if breakingChangeReason == "" || breakingChangeConfirmed {
+			return
+		}
+		breakingChangeConfirmed = v2breaking.ListObjectsResponseConfirmsReason(breakingChangeReason, req.GetUser(), []string{object})
+	}
```
adriantam: Why do you need this as a inline function rather than a top level function?

adriantam (dismissed):

adriantam (approved):

PR #3225, opened 2026-07-25
fix: only skip the targeted recursive relation when flattening v2 Check graph nodes

yxshwanth:
#### What problem is being solved?

The experimental `weighted_graph_check` (v2 Check) can return `false` for authorization models that have **two or more independently-recursive relations**, where satisfying the outer recursive relation requires resolving through an unrelated, nested recursive relation. `CheckV1` correctly returns `true` for the same request, so this is a correctness regression in v2 Check.

Example model:

```
model
  schema 1.1
type user
type group
  relations
    define admin: direct_admin or admin from parent_group
    define child_group: [group]
    define descendant_principal: member or descendant_principal from child_group
    define direct_admin: [user]
    define direct_member: [user]
    define member: direct_member or admin
    define parent_group: [group]
```

With tuples `group:g1#direct_admin@user:u1`, `group:g1#child_group@group:g2`, `group:g2#parent_group@group:g1`, checking `group:g2#descendant_principal@user:u1` should be `true` (via `g2 -> member -> admin -> admin from parent_group -> g1 -> direct_admin -> u1`), but v2 Check returned `false`.

Root cause: `AuthorizationModelGraph.FlattenNode` (`internal/modelgraph/model.go`) took a `recursivePath bool` and, when `true`, unconditionally skipped **every** edge with a non-empty `RecursiveRelation` while flattening. `ResolveRecursive`/`bottomUp` pass `true` while unwinding the outer recursive relation (`descendant_principal`), so the flatten step also dropped the unrelated, independently-recursive `admin from parent_group` edge reached along the `member` branch - discarding the only path that could satisfy the check.

Fixes #3195.

#### How is it being solved?

`FlattenNode` now takes the specific `RecursiveRelation` string to skip (`skipRecursiveRelation`) instead of a boolean. An edge is only excluded when its own `RecursiveRelation` matches the one currently being unwound; edges belonging to a *different* recursive relation are always kept, since dropping them silently discards valid resolution paths.

#### What changes are made to solve it?

- `internal/modelgraph/model.go`: Changed `FlattenNode`'s `recursivePath bool` parameter to `skipRecursiveRelation string`; an edge is skipped only if `edge.GetRecursiveRelation() == skipRecursiveRelation` (when non-empty).
- `internal/check/check.go`: `ResolveUnion` now passes `""` (nothing to skip); `ResolveRecursive` passes `edge.GetRecursiveRelation()` (the relation it is currently unwinding).
[...]

Issue #3195, opened 2026-07-08
Weighted graph check evaluation inconsistent with CheckV1


HoangViet144:
### Checklist

- [x] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [x] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.
- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Description

In checkV2 this check request return **false** while in checkV1 return **true**

docker compose:
```
  migrate_openfga:
    depends_on:
      db:
        condition: service_healthy
    image: openfga/openfga:v1.18.1
    container_name: migrate
    command: migrate
    environment:
      - OPENFGA_DATASTORE_ENGINE=mysql
      - OPENFGA_DATASTORE_URI=root:admin@tcp(db:3306)/openfga?parseTime=true
      - OPENFGA_HTTP_CORS_ALLOWED_ORIGINS=*
  openfga:
    depends_on:
      migrate_openfga:
        condition: service_completed_successfully
    image: openfga/openfga:v1.18.1
    container_name: openfga
    environment:
      - OPENFGA_DATASTORE_ENGINE=mysql
      - OPENFGA_DATASTORE_URI=root:admin@tcp(db:3306)/openfga?parseTime=true
      - OPENFGA_DATASTORE_MAX_OPEN_CONNS=100
      - OPENFGA_DATASTORE_MIN_OPEN_CONNS=20
      - OPENFGA_DATASTORE_MIN_IDLE_CONNS=20
      - OPENFGA_DATASTORE_CONN_MAX_IDLE_TIME=5m
      - OPENFGA_DATASTORE_METRICS_ENABLED=true

      - OPENFGA_LOG_FORMAT=json
      - OPENFGA_LOG_LEVEL=debug
      - OPENFGA_AUTHN_METHOD=preshared
      - OPENFGA_AUTHN_PRESHARED_KEYS=abcdabcdabcd
[...]

PR #3235, opened 2026-07-29
fix: disable recursive optimization when a relation has multiple recursive edges

### LINKED ISSUE #3233: Weighted graph check evaluation inconsistent with CheckV1 when a relation has two or more independent recursive edges
### Checklist

### Description

Follow-up to #3195, found by @saad-h1 while reviewing #3225. Filing separately at @senojj's suggestion, since it is a distinct root cause from the one #3225 fixed.

The experimental `weighted_graph_check` (v2 Check) can return `false` where CheckV1 returns `true`, for models in which a **single** relation has two or more edges that independently recurse back into that same relation.

**Model:**

```
model
  schema 1.1
type user
type group
  relations
    define parent_group: [group]
    define child_group: [group]
    define member: [user] or member from parent_group or member from child_group
```

**Tuples:**

```
USER: user:u1
RELATION: member
OBJECT: group:g1

USER: group:g1
RELATION: parent_group
OBJECT: group:g2

USER: group:g1
RELATION: child_group
OBJECT: group:g3
```

**Check requests:**

| Request | Resolves via | Expected | v2 Check actual |
|---|---|---|---|
| is `user:u1` related to `group:g2` as `member`? | `member from parent_group` | `true` | **`false`** |
| is `user:u1` related to `group:g3` as `member`? | `member from child_group` | `true` | `true` |

Only one of the two recursive branches resolves. Which one survives depends on graph edge ordering, so the branch that breaks is not necessarily `parent_group`; the point is that exactly one of them is silently dropped.

### Root cause

`canApplyRecursiveOptimization` in [`internal/modelgraph/model.go`](https://github.com/openfga/openfga/blob/main/internal/modelgraph/model.go) scans a node's edges for the edge that recurses back into the relation being evaluated:

```go
} else if edge.GetEdgeType() == authzGraph.DirectEdge || edge.GetEdgeType() == authzGraph.TTUEdge {
    recursiveEdge = edge
}
```

There is no check for a previously assigned `recursiveEdge`. In the model above, both `member from parent_group` and `member from child_group` carry `RecursiveRelation == "group#member"`, so the loop matches twice and simply overwrites, keeping whichever edge it visited last.

`CanApplyRecursion` hands that single edge to `ResolveRecursive`, which then calls:

```go
FlattenNode(edge.GetTo(), req.GetUserType(), req.IsTypedWildcard(), edge.GetRecursiveRelation())
```

Because both competing edges share the same `RecursiveRelation` string, this flatten step skips **both** of them, not just the selected one. The selected edge is then resolved on its own via `resolveRecursiveUserset`/`resolveRecursiveTTU`, while the other edge is neither flattened into the union nor resolved separately. It drops out of evaluation entirely, so any request satisfiable only through that branch returns `false`.

This function has behaved this way since the original weighted graph Check refactor in #2816, so it predates both #3195 and #3225. It is separate from the bug fixed in #3225, which concerned `FlattenNode` discarding edges belonging to an unrelated, differently named recursive relation. This issue is the inverse: multiple edges sharing the *same* recursive relation, where the ambiguity between them is never detected in the first place.

### Suggested fix

Have `canApplyRecursiveOptimization` detect the ambiguity rather than resolve it arbitrarily. When a second edge matching the target recursive relation is encountered at any level of the walk, return `(nil, false)` so `CanApplyRecursion` declines the optimization. `ResolveUnion` then falls back to the generic path (`FlattenNode` with no skip, followed by `ResolveUnionEdges`), where every recursive edge is resolved independently under the existing cycle safe `visited` map. That path is already correct today; the optimization is what short circuits it.

I have this implemented and verified locally, with a unit test at the `CanApplyRecursion` level and an end to end `ResolveCheck` test against the model above. Both fail before the change and pass after, and the existing `internal/check` and `internal/modelgraph` suites stay green. I will open a PR with the fix.

### OpenFGA version

`main` (as of `168de21`, the merge commit for #3225)

As a binary

In-Memory

### OpenFGA Flags

`OPENFGA_EXPERIMENTALS=weighted_graph_check`

### Logs

Not applicable. Reproduced directly against a mocked datastore in a unit test rather than a running server; the test is included in the PR.

yxshwanth:
### What problem is being solved?

The experimental `weighted_graph_check` (v2 Check) can return `false` where CheckV1 returns `true`, for models in which a **single** relation has two or more edges that independently recurse back into that same relation.

Example model:

```
model
  schema 1.1
type user
type group
  relations
    define parent_group: [group]
    define child_group: [group]
    define member: [user] or member from parent_group or member from child_group
```

With tuples `group:g1#member@user:u1`, `group:g2#parent_group@group:g1` and `group:g3#child_group@group:g1`:

| Request | Resolves via | Expected | v2 Check before this PR |
|---|---|---|---|
| is `user:u1` related to `group:g2` as `member`? | `member from parent_group` | `true` | **`false`** |
| is `user:u1` related to `group:g3` as `member`? | `member from child_group` | `true` | `true` |

Exactly one of the two recursive branches resolves; which one survives depends on graph edge ordering.

**Root cause:** `canApplyRecursiveOptimization` (`internal/modelgraph/model.go`) scans a node's edges for the edge that recurses back into the relation being evaluated, and assigns `recursiveEdge = edge` on every match without checking whether one was already found. Both `member from parent_group` and `member from child_group` carry `RecursiveRelation == "group#member"`, so the loop matched twice and kept only the edge it visited last.

`CanApplyRecursion` then handed that single edge to `ResolveRecursive`, which flattens the node while skipping every edge whose `RecursiveRelation` matches the one being unwound. Because the competing edges share the same relation string, that skip removed **both** of them, and only the surviving edge was resolved separately via `resolveRecursiveUserset`/`resolveRecursiveTTU`. The other recursive branch was neither flattened into the union nor resolved on its own, so it dropped out of evaluation entirely.

This behaviour dates to the original weighted graph Check refactor in #2816, so it predates both #3195 and #3225. It is a distinct root cause from the bug fixed in #3225, which concerned `FlattenNode` discarding edges belonging to an unrelated, differently named recursive relation. This one is the inverse: multiple edges sharing the *same* recursive relation, where the ambiguity between them was never detected.

Found by @saad-h1 while reviewing #3225, and filed separately as #3233 at @senojj's suggestion.

### How is it being solved?

Detect the ambiguity rather than resolving it arbitrarily. When `canApplyRecursiveOptimization` encounters a second edge matching the target recursive relation at any level of its walk, it returns `(nil, false)` so `CanApplyRecursion` declines the optimization.

`ResolveUnion` then falls back to the generic path (`FlattenNode` with no skip, followed by `ResolveUnionEdges`), where every recursive edge is resolved independently under the existing cycle safe `visited` map. That path is already correct today; the optimization was short circuiting it. The optimization still applies unchanged to the single recursive edge case, which is the common one.

### What changes are made to solve it?

- `internal/modelgraph/model.go`: `canApplyRecursiveOptimization` now bails out with `(nil, false)` when a second edge matching the recursive relation is found, both for directly matched `DirectEdge`/`TTUEdge` edges and for edges returned by the recursive descent into operator nodes.
- `internal/modelgraph/model_test.go`: added `TestCanApplyRecursionBailsOutOnAmbiguousRecursiveEdges`, asserting `CanApplyRecursion` returns `(nil, false)` for the model above rather than picking one edge.
- `internal/check/check_test.go`: added `TestCheck_MultipleRecursiveEdgesForSameRelation`, an end to end `ResolveCheck` regression test against a mocked datastore covering both branches, where `group:g2` has no `child_group` tuples and `group:g3` has no `parent_group` tuples so each is reachable only through its own recursive edge.
- `CHANGELOG.md`: added an `[Unreleased] -> Fixed` entry referencing #3233.

### Verification

Both new tests fail on `main` and pass with this change. The `allowed_through_parent_group_only` subtest is the one that reproduces the reported false negative. `internal/check`, `internal/modelgraph` and the full `tests/check` matrix suite all pass.

## References

Fixes #3233

Related: #3225, #3195, #2816

## Review Checklist

- [x] I have clicked on "allow edits by maintainers".
- [ ] I have added documentation for new/changed functionality in this PR or in a follow-up, linked here: N/A (bug fix, no user-facing behavior or API change; the experimental flag's behavior now matches its documented v1-equivalent semantics).
- [x] The correct base branch is being used, if not `main`.
- [x] I have added tests to validate that the change in functionality is working as expected (unit test in `internal/modelgraph` and integration test in `internal/check`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed experimental recursion optimization when a single relation contains multiple ambiguous recursive edges.
  - Prevented scenarios where one matching recursive edge could be selected while others were effectively ignored, ensuring correct authorization outcomes.

- **Tests**
  - Added regression coverage to confirm recursion optimization is disabled when ambiguity is detected.
  - Added additional checks to verify authorization is allowed only through the specific recursive edge that applies.


yxshwanth:
Flagging the `go-bench` failure preemptively, since I believe it is a false positive.

The alert is on `BenchmarkOpenFGAServer/BenchmarkPostgresDatastore/BenchmarkCheck/with_simple_ttu`, at 2.07x B/op and 2.06x allocs/op.

**The changed code is not reachable from that benchmark.** `BenchmarkCheck` constructs `graph.NewLocalChecker(graph.WithOptimizations(true))`, the v1 Check engine. This PR's only non-test change is in `internal/modelgraph.canApplyRecursiveOptimization`, which is reachable solely through `CanApplyRecursion`, whose only non-test caller is `internal/check/check.go:340` in v2 Check. `internal/graph` does not import `internal/modelgraph`, so the changed code cannot execute in this benchmark on any datastore.

On why the numbers moved: the alert is on B/op and allocs/op, which are usually stable. In this case the benchmark short-circuits across 999 concurrent parent lookups (`folder:x` has 999 `parent` tuples, with the matching `member` on `group:999`), so how much work completes before cancellation varies per run, and allocation counts vary with it. At `-benchtime 1s` the Postgres case only managed 7 iterations. MySQL and SQLite ran the same model through the same resolution code and were not flagged:

| Datastore | `with_simple_ttu` allocs/op | Flagged |
|---|---|---|
| Postgres | 93,591 | yes, 2.06x |
| MySQL | 56,194 | no |
| SQLite | 28,533 | no |

That combination, single-digit iteration counts on the DB-backed Check benchmarks plus a hard 2x threshold compared against a single cached baseline run, looks like a pre-existing sharp edge in the benchmark gate rather than anything specific to this PR. Not opening an issue for it here since it is unrelated to this change, but happy to write it up separately if maintainers think it is worth tracking.

Happy to re-run the job if a second data point would help.

senojj:
@yxshwanth thank you for the PR submission.

This problem is the result of a bug in the construction of the weighted graph, thus it will be handled within the [language graph package](https://github.com/openfga/language/tree/main/pkg/go/graph).

Currently, the model relation in question does not get flagged properly as being part of a tuple cycle in the weighted graph. We will be making a change to fix this behavior, which will result in the `CanApplyRecursion` function properly handling this scenario by returning a `nil` edge value.

yxshwanth:
Thanks @senojj, that makes sense. Closing this in favour of the fix in the language graph package, which addresses the root cause properly rather than working around it downstream.

Thanks again for the explanation, and to @saad-h1 for spotting the case originally.

PR #3246, opened 2026-08-05
fix(server): BatchCheck emits v1→v2 divergence telemetry (was blind)

SashaMIT:
## Summary

- `Check` warns (`potential v2 Check resolution breaking change`, with a `v2breaking` reason) when a denied userset request matches a known v1→v2 divergence shape — the rollout guardrail for the weighted-graph resolver.
- `BatchCheck` routes through the same v2 checker but emitted **no divergence telemetry at all** — only aggregate `v2_check_count`/`v2_fallback_count` counters. A batch of 50 checks with divergence-shaped denials produced zero signal during rollout.
- Mirror the Check gate in the batch path: for each check where v2 answered `false` and the user is a userset, evaluate `v2breaking.CheckReason` and warn with `reason` + `correlation_id`. Same "may over-report on fallback, never miss" semantics as the existing package contract.

## Impact if unsolved

Operators migrating stores to the weighted-graph resolver use this telemetry to find models that will break at cutover. Divergences arriving via BatchCheck — typically the bulk of production traffic — were invisible, so a store could look clean on dashboards while its batch traffic silently resolved differently under v2.

## Test plan

- New `TestBatchCheckV2BreakingChangeLog`: a self-referential-userset check logs exactly one warning with `correlation_id` and `reason=self_referential_userset`; a plain deny and an allow stay silent.
- Full `TestBatchCheck*` suite passes.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **New Features**
  - Added warning diagnostics for BatchCheck results that may resolve differently between v1 and v2.
  - Warnings include relevant request and correlation details, along with the detected reason.

- **Bug Fixes**
  - Improved visibility into denied checks involving self-referential userset resolution while avoiding warnings for ordinary denials and allowed checks.

### DIFF
--- a/pkg/server/batch_check.go
+++ b/pkg/server/batch_check.go
@@ -154,23 +154,21 @@ func (s *Server) BatchCheck(ctx context.Context, req *openfgav1.BatchCheckReques
 	// maxChecksPerBatchCheck per request). Shape filters may over-report on
 	// per-check v1 fallback but never miss a real divergence.
 	if v2Enabled && !v2GraphResolveFailed {
-		if typesys, tsErr := s.resolveTypesystem(ctx, storeID, req.GetAuthorizationModelId()); tsErr == nil {
-			requestID := requestid.GetRequestIDFromContext(ctx)
-			for _, check := range req.GetChecks() {
-				outcome, ok := result[commands.CorrelationID(check.GetCorrelationId())]
-				tk := check.GetTupleKey()
-				if !ok || outcome.Err != nil || outcome.Allowed || !tuple.IsObjectRelation(tk.GetUser()) {
-					continue
-				}
-				if reason := v2breaking.CheckReason(typesys, tk); reason != "" {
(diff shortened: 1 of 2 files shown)


poovamraj:
Hi @SashaMIT, thanks for the contribution.

Were you able to check this prior PR we have for this - https://github.com/openfga/openfga/pull/3210

SashaMIT:
Good catch — I hadn't seen #3210 when I opened this (it was still unmerged on master when I checked for existing BatchCheck divergence telemetry). The two overlap: both emit v1→v2 breaking-change logs from BatchCheck while reusing the already-resolved typesystem.

Differences I can see: #3210 aggregates distinct divergence shapes into a single per-request log via a helper and carries a changelog entry; this PR logs inline per correlation. I'm happy to close this in favor of #3210 — maintainer-authored and further along is the right call. Want me to close it, or would you prefer to take anything from here (e.g. the test) once #3210 lands?

poovamraj:
@SashaMIT Thank you for your contribution but we will close this PR for now.

PR #3247, opened 2026-08-06
feat(cmd): add built-in healthcheck command, replacing grpc_health_probe

### LINKED ISSUE #2373: Drop grpc_health_probe from released docker image

poovamraj:
Fixes #2373. Supersedes/continues #3103 (original author: @alliasgher) — rebased onto latest `main` and with CI failures fixed.

**Why:** `grpc_health_probe` introduces transitive CVE exposure on the OpenFGA image. Rather than simply dropping it (which would break the container `HEALTHCHECK` and Compose `depends_on: service_healthy`), this PR replaces it with a first-class, in-binary `openfga healthcheck` command — no external binary, no separate CVE stream, and the probe shares the server's own address/TLS configuration.

**Changes:**
- **New `openfga healthcheck` command** (`cmd/healthcheck`):
  - `--target grpc` (default) performs a gRPC Health Checking Protocol `Check`; `--target http` probes the HTTP `/healthz` endpoint.
  - Reads the same configuration as `run` — flags, config file, and `OPENFGA_*` env vars (`grpc.addr`, `grpc.tls.*`, `http.addr`, `http.tls.*`) — so a probe running in the same container needs no extra configuration to match the server.
  - `--service` selects a named gRPC health service; `--timeout` bounds the probe (default 5s).
  - When gRPC TLS is enabled, the probe verifies the server against the configured certificate (`ServerName` derived from the address host). This closes the mTLS gap the native Kubernetes `livenessProbe.grpc` cannot cover, since that probe is plaintext-only.
- Removed the `grpc_health_probe` multi-stage copy from `Dockerfile` and `Dockerfile.goreleaser`.
- Restored the container `HEALTHCHECK` in both Dockerfiles to use `["/openfga", "healthcheck"]`.
- Restored the `openfga` service `healthcheck` in `docker-compose.yaml` to use the built-in command, so `depends_on: condition: service_healthy` keeps working.
- Removed the CodeRabbit `path_instructions` rule and the `.github/instructions/dockerfile-goreleaser-sync.instructions.md` file that enforced grpc-health-probe version sync between the two Dockerfiles (both now obsolete).
- Updated the `cmd/openfga` docker test: it waits for the container to reach the running state (fast-failing on terminal states, including `removing`, with the exit code) rather than asserting on Docker's `State.Health`; service readiness is still verified via `testutils.EnsureServiceHealthy` (gRPC + HTTP `/healthz`).
- Added `cmd/healthcheck/healthcheck_test.go` covering gRPC/HTTP success and failure paths, TLS verification against the configured cert, and error handling (93.5% package coverage).
- Updated the CHANGELOG with `Added` (the new command) and `Removed` (the binary) entries, including Kubernetes/mTLS guidance.

**Migration:** No action required for users on the official image — the container `HEALTHCHECK` and Compose healthcheck now use the built-in command automatically. Kubernetes users can use the native [`livenessProbe.grpc`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) field (plaintext gRPC only); deployments requiring gRPC mTLS should use `openfga healthcheck` (as an exec probe) or probe the HTTP `/healthz` endpoint.

## Summary by CodeRabbit

* **New Features**
  * Added the built-in `openfga healthcheck` command for gRPC and HTTP health checks.
  * Added configurable targets, timeouts, TLS settings, and server configuration.
  * Container and Docker Compose health checks now use the built-in command instead of the bundled gRPC probe.
* **Bug Fixes**
  * Improved container startup handling and readiness validation.
* **Documentation**
  * Updated the unreleased changelog with health-check changes, migration guidance, deployment alternatives, and TLS limitations.


poovamraj:
This PR is built on top of  @alliasgher's work here - https://github.com/openfga/openfga/pull/3103 and would be attributed to the original author. @alliasgher thank you for your contribution!

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ## [Unreleased]
+### Removed
+- Removed the bundled `grpc_health_probe` binary and the container `HEALTHCHECK` from the released OpenFGA images (`Dockerfile`, `Dockerfile.goreleaser`), along with the `openfga` service healthcheck in `docker-compose.yaml`. This drops the transitive CVE exposure the probe binary introduced. Kubernetes users should switch to the native [`livenessProbe.grpc`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) field (enabled by default since Kubernetes 1.24, GA since 1.27); anyone who still needs the binary can layer it onto the OpenFGA image in their own Dockerfile. [#3247](https://github.com/openfga/openfga/pull/3247)
```
Siddhant-K-code: The Kubernetes migration guidance needs an mTLS qualification. Native gRPC probes do not support authentication parameters, including TLS client credentials. State that native probes are for plaintext gRPC, and direct gRPC-mTLS users to HTTP `/healthz` or custom probes. [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)

docker-compose.yaml
```diff
@@ -48,13 +48,3 @@ services:
...
-    healthcheck:
-      test:
-        [
-          "CMD",
-          "/usr/local/bin/grpc_health_probe",
-          "-addr=openfga:8081"
-        ]
-      interval: 5s
-      timeout: 30s
-      retries: 3
```
Siddhant-K-code, line 60: Removing this healthcheck removes `State.Health` and breaks consumers using `depends_on.condition: service_healthy` (we might need to document it). Those users need to remove that condition, use an external checker, or maintain their own derived-image healthcheck. [Docker Compose documentation](https://docs.docker.com/reference/compose-file/services/)

poovamraj: But do we ship the Docker compose anywhere? I can we have a doc but that doesn't touch healthcheck there - https://openfga.dev/docs/getting-started/setup-openfga/docker. So wondering who we will be breaking to document it? Even for our internal dev use case it won't be a breaking change right?

PR #3248, opened 2026-08-06
fix: make cancelled-context Check deterministic in weighted graph resolver

### LINKED ISSUE #3214: Flaky test: TestV2CheckMetadata/cancelled_context_returns_error_not_fallback races on context cancellation
### Checklist

### Description

[`TestV2CheckMetadata/cancelled_context_returns_error_not_fallback`](https://github.com/openfga/openfga/blob/1e464a742e8cfd9887ef55f5a36d271f5d1eeb40/pkg/server/check_test.go#L611) in pkg/server/check_test.go is flaky. It passes locally but intermittently fails in [CI with](https://github.com/openfga/openfga/actions/runs/29578950559/job/88226120588?pr=3202#step:6:85):

```
check_test.go:624:
    Error: An error is expected but got nil.
    Test:  TestV2CheckMetadata/cancelled_context_returns_error_not_fallback
```

The test cancels the context before calling `Check` and asserts an error is returned (and that no v1 fallback happens). This invariant is not actually guaranteed by the resolver, so the assertion races.

The 50/50 limiter race was confirmed with a standalone select over a cancelled context and a large-buffer channel (~50% proceed). The full test only fails when the edge goroutine wins both the limiter race and the send-to-out race, which is timing/scheduling dependent.

The test has been added in #3141.

yxshwanth:

`TestV2CheckMetadata/cancelled_context_returns_error_not_fallback` is flaky in CI (#3214). The test cancels the caller's context before invoking `Check` and asserts an error is returned with no v1 fallback, but as the issue notes, that invariant is not actually guaranteed by the experimental `weighted_graph_check` resolver, so the assertion races.

The root cause is that the resolver does not check the context before it starts work. A request whose context is already done still walks the graph, and whichever branch resolves first can hand back an `Allowed: true` that raced the cancellation, so `Check` can return a successful result for a cancelled request.

#### How is it being solved?

`ResolveCheck` returns the context error on an already-cancelled or deadline-exceeded request before doing any graph work. This makes the pre-cancelled case (what the flaky test exercises) deterministic, before any goroutine or concurrency limiter is involved.

This mirrors v1, whose reducers guard the same way on entry (`union` at `internal/graph/check.go:161`, `intersection` at :227, `exclusion` at :300). Cancellation *during* resolution is deliberately left alone, also matching v1: a branch that has already reached a terminal result returns it rather than discarding work that is already paid for. An earlier revision of this PR re-checked `ctx.Err()` before the `allowed=true` short-circuit in `ResolveUnionEdges` and `ResolveRecursive`; that made v2 stricter than v1 for no benefit and was dropped in review.

`context.Canceled` is already classified by `IsV2CheckTerminalError`, so the error propagates as `ErrRequestCancelled` without falling back to v1, and the datastore query-count / throttle tags are still emitted on the error path.

The fix is in the resolver rather than in the assertion, matching the analysis in the issue.

## References

closes https://github.com/openfga/openfga/issues/3214

## Testing

- `TestResolveCheck_CancelledBeforeResolution` covers a cancelled context and an expired deadline, asserting the context error is returned and never a successful result. The model reaches `viewer` both directly and through a userset, and `ReadUserTuple` / `ReadUsersetTuples` are `Times(0)`, so the test fails unless `ResolveCheck` returns before any graph work starts. Verified that removing the gate fails both subtests, one on each read, so neither expectation is vacuous.
- `./internal/...` passes, and `./pkg/server/...` passes with `-race`.
- The #3214 test passes 25 consecutive runs. Note the original flake does not reproduce locally even without this change (400 runs green), consistent with the issue's "passes locally". The guarantee here is therefore structural rather than empirical.

## Review Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Authorization checks now return the appropriate cancellation or deadline error when requests are already cancelled before resolution begins.
  - Cancellation during resolution remains cooperative, while requests that reach a terminal result continue to return that result.
- **Tests**
  - Added regression coverage for pre-cancelled requests and expired deadlines, including verification that no authorization data is read.
- **Documentation**
  - Updated the unreleased changelog with details of the cancellation-handling fix.


yxshwanth:
`2d32451` addresses both CodeRabbit comments. The changelog entry is now scoped to `ResolveUnionEdges` / `ResolveRecursive`, and no longer implies intersection and exclusion return cancellation consistently. The `check_test.go` comment was handled in `859e115`: `GetPlanSelector` is `Times(0)`, which makes `Select` and `UpdateStats` unreachable, so the test asserts the gate returns before any planner or datastore work.

The five red checks are unrelated to this diff, which touches only `CHANGELOG.md` and `internal/check/`.

Three died during Set up job, before any code ran:

```
markdown-link-check:  Unable to resolve action `actions/checkout@3d3c42e5...`, unable to find version
tests-matrix:         Failed to resolve action download info. Error: Service Unavailable
govulncheck:          Bad Gateway -> Service Unavailable -> Internal Server Error
```

Actions could not resolve pinned action SHAs between 15:25 and 15:30Z. The same workflows and SHAs passed on the three previous runs on this branch.

`tests-unit` failed for real, in a package this PR does not touch:

```
--- FAIL: TestNewSharedIteratorDatastore_iter/single_item_head_first
    shared_iterator_datastore.go:750: Unexpected call to MockIterator.Stop() ...
    expected call at shared_iterator_datastore_test.go:1375 doesn't have a prerequisite call satisfied
```

That is a gomock `InOrder` prerequisite race in `pkg/storage/storagewrappers/sharediterator`. It passes 25/25 locally under `-race`. The last functional change to that package was #3148.

`tests` is the aggregate job. It cascades from the two above.

On `2d32451`, `make test-matrix` is green locally and codecov is green on both checks.

Can a maintainer re-run the failed jobs? I can't trigger one from a fork.

internal/check/check.go
```diff
@@ -295,6 +302,11 @@ func (r *Resolver) ResolveUnionEdges(ctx context.Context, req *Request, edges []
 			}
 
 			if msg.Res.GetAllowed() {
+				// select may choose a ready message over ctx.Done(); re-check cancellation
+				// so we do not return success when the request was already cancelled.
+				if err := ctx.Err(); err != nil {
+					return nil, err
+				}
```
senojj: If we receive a `true` resolution in the presence of a union, we should not discard it just because of a canceled context. We've done the work, and reached a terminal resolution.

internal/check/check.go
```diff
@@ -582,6 +594,11 @@ func (r *Resolver) ResolveRecursive(ctx context.Context, req *Request, edge *aut
...
```
senojj: Why abandon the result here? If any downstream resolution is I/O bound, the context cancelation will prevent it from executing. Otherwise, downstream evaluation is entirely CPU bound and thus would make no sense to abandon the result here.

internal/check/check.go
yxshwanth: Agreed, removed in be1380c. A union branch returning true is terminal, the work is already done, and the caller's cancellation does not make that result wrong.

v1 agrees too: `union` guards on entry at `internal/graph/check.go:161` and then returns its `Allowed` short-circuit at :206 without rechecking the context. The guard here was making v2 stricter than v1 for no benefit, and it did nothing for #3214 either, since that request is cancelled before `Check` is invoked and never reaches a union.

internal/check/check.go
yxshwanth: Same, removed in be1380c. Downstream I/O already respects the context, so the only result the guard could discard is one that was reached without it. `ResolveRecursive` now returns the terminal true like the other operators.

yxshwanth:
be1380c drops both short-circuit guards per @senojj's review. What remains is the `ctx.Err()` gate at the top of `ResolveCheck`, which matches what v1 already does at the top of `union` / `intersection` / `exclusion` (`internal/graph/check.go:161`, :227, :300). v1 also returns its `Allowed` short-circuit without rechecking the context, so v2 now has the same cancellation semantics rather than stricter ones.

This removes the intersection/exclusion inconsistency I flagged under "Known gap", so there is no follow-up PR to file. No operator discards a computed result now, and only the pre-cancelled case returns early. I have edited that section out of the description, along with the union test that only existed to pin the removed guard.

`TestResolveCheck_CancelledBeforeResolution` now covers both a cancelled context and an expired deadline, over a model where `viewer` is reachable directly and through a userset. With the gate removed, the two subtests fail on `ReadUserTuple` and `ReadUsersetTuples` respectively, so neither `Times(0)` is vacuous. `./internal/...` is green, and `pkg/server` plus `pkg/server/commands` are green under `-race`.

pacocartones:
Took a proper look at this since the fix is clean and the PR has been sitting a bit — sharing my analysis in case it helps nudge it along.

**The guard looks spot-on.** `ctx.Err()` at `internal/check/check.go:117-119` sits before `GetNodeByID` and any graph work, so a pre-cancelled request bails out deterministically. It mirrors exactly what v1 already does in the reducers (`internal/graph/check.go:161/227/300`), so v2 is just catching up to established behavior rather than inventing anything.

**On senojj's two comments** (discarding a `true` result on cancellation in `ResolveUnionEdges`/`ResolveRecursive`) — both look fully addressed in `be1380c`: the mid-resolution `ctx.Err()` re-checks were removed, which is the right call. A branch that already reached a terminal `Allowed: true` did real work; throwing it away buys nothing, and v1 doesn't do it either. Cancellation *during* resolution stays cooperative via the existing `select` on `ctx.Done()`, unchanged. CodeRabbit's two nits (CHANGELOG scope, `Times(0)`) also read as resolved.

**The `dirty` state is trivial** — the only conflict is `CHANGELOG.md`, and the two entries live independently in `### Fixed` with no semantic overlap. A rebase or squash-merge makes it vanish.

**Test looks solid:** `TestResolveCheck_CancelledBeforeResolution` covers both `cancelled` and `deadline_exceeded`, and the `Times(0)` on both `ReadUserTuple` and `ReadUsersetTuples` isn't vacuous — with two reachable paths to `viewer`, without the guard it'd hit one `Read*` or the other depending on traversal order. Nicely deterministic.

One tiny non-blocking nit: since the `defer` also records `ctx.Err()`, a cancelled request ends up recording the error twice on the span (guard + defer). Harmless, just a little trace noise — definitely not worth a change.

Only thing I couldn't check myself: the red CI. From the earlier note they look like infra flakes (Bad Gateway / Service Unavailable) plus one failure in `TestNewSharedIteratorDatastore` which is in a package this PR doesn't touch. Worth a re-run to confirm.

Anyway — reads as ready to me. Thanks @yxshwanth for the careful iteration on this one.

PR #3249, opened 2026-08-07
test: add regression tests for the connection pool starvation deadlock fixed by #3255

tjb9dc:
## Description

Regression tests for the connection pool starvation deadlock fixed by #3255. Originally opened as intentionally-failing tests reproducing the bug; now rebased onto `main` (post-#3255) and reframed as regression coverage, since #3255 merged without a test reproducing the deadlock. Both tests pass on `main` and fail if the `context.WithoutCancel` scoping regresses.

### The bug these tests guard against

`SQLTupleIterator.fetchBuffer` stripped cancellation (`context.WithoutCancel`, introduced in #2508) from the entire row-fetch path, so waiting for a pool connection was un-cancellable. ListObjects reverse expansion holds an open iterator (pinning a connection) while dispatching child reads that each need their own connection — with a saturated pool this formed a circular wait that no request deadline could unwind. The connections were never returned, and every subsequent datastore request timed out until the server was restarted.

Observed in production on v1.11.5 (`datastore-max-open-conns=10`) under concurrent ListObjects load: `pgxpool_acquired_conns` pinned at max, hundreds of goroutines parked in `sync.WaitGroup.Wait` / `pgxpool.Acquire`, all connections idle on Postgres yet unusable, recoverable only by restart. A version bisection from the field independently pointed at the same boundary: the behavior was fine on v1.8.6 and broken from v1.8.16 (the first release containing #2508) onward — see the comment thread for details.

### Tests

- `TestIteratorNextRespectsContextDuringPoolAcquire` — the primitive: with `WithMaxOpenConns(1)` and the connection held by another iterator, `Next()` with a 500ms-deadline context must fail promptly instead of blocking indefinitely.
- `TestReverseExpandPoolStarvationDeadlock` — the consequence: a single reverse expansion (user in 20 groups, above the default `resolveNodeBreadthLimit` of 10) against a single-connection pool must return by its 2s context deadline, and the pool must be usable afterwards.

Because the tests assert the desired behavior (return by deadline, pool recovers) rather than the presence of the bug, they required no logic changes after #3255 merged — only comment/framing updates.

## References

- Fixes validated: #3255
- Origin of the stripped context: #2508
- Related cancellation-leak fixes that did not cover this path: #3106, #3111

## Review Checklist

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for database connection pool starvation scenarios.
  * Verified that canceled operations stop waiting for connections and return promptly.
  * Added checks ensuring reverse expansion respects deadlines and the connection pool remains usable afterward.

### DIFF
--- a/pkg/storage/postgres/pool_starvation_test.go
+++ b/pkg/storage/postgres/pool_starvation_test.go
@@ -1,0 +1,212 @@
+package postgres_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/oklog/ulid/v2"
+	"github.com/stretchr/testify/require"
+
+	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+
+	"github.com/openfga/openfga/pkg/server/commands/reverseexpand"
+	"github.com/openfga/openfga/pkg/storage"
+	"github.com/openfga/openfga/pkg/storage/postgres"
+	"github.com/openfga/openfga/pkg/storage/sqlcommon"
+	storagefixtures "github.com/openfga/openfga/pkg/testfixtures/storage"
+	"github.com/openfga/openfga/pkg/testutils"
+	"github.com/openfga/openfga/pkg/tuple"
+	"github.com/openfga/openfga/pkg/typesystem"
+)
+
+// closeWithTimeout closes the datastore in the background. Closing a pgxpool
+// blocks until every acquired connection is released, which never happens if a
+// regression reintroduces the connection starvation these tests guard against,
+// so a plain ds.Close() would hang test teardown.
+func closeWithTimeout(t *testing.T, ds storage.OpenFGADatastore, d time.Duration) {
+	t.Helper()
+	done := make(chan struct{})
+	go func() {
+		ds.Close()
+		close(done)
+	}()
+	select {
+	case <-done:
+	case <-time.After(d):
+		t.Log("datastore.Close() timed out; connections are still held")
+	}
+}
+
+// TestIteratorNextRespectsContextDuringPoolAcquire verifies that a tuple
+// iterator's lazy query execution honors context cancellation while waiting
+// for a pool connection.
+//
+// Regression test for https://github.com/openfga/openfga/pull/3255. Before
+// that fix, SQLTupleIterator.fetchBuffer called context.WithoutCancel(ctx)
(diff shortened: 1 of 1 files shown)


tjb9dc:
Additional corroborating evidence for the bisection to #2508: a user reported this behavior working on v1.8.6 but broken on v1.9.0.

That boundary matches exactly. The `context.WithoutCancel` in `fetchBuffer` (dbfe47cb, 2025-06-16) first shipped in **v1.8.16** (2025-06-17), so v1.8.6 predates it and v1.9.0 contains it.

The failure-mode change across that boundary is what makes it present as a regression:

- **Pre-v1.8.16**: the same pool starvation could form transiently under ListObjects fan-out (the parent-iterator-held-while-children-dispatch structure existed then too), but children waited for connections with the real request context. When the deadline fired they errored out, the parent unwound, iterators were stopped, and connections returned to the pool — failed requests under load, then self-healing.
- **v1.8.16+**: the waits are un-cancellable, so the same congestion converts into the permanent wedge described above — connections are never returned and the server is unrecoverable without a restart.

(v1.8.x still used `database/sql` rather than pgxpool — the migration came later in the 1.10.x line — but `WithoutCancel` has the identical effect on both drivers: `QueryContext`'s internal connection wait also blocks indefinitely on a cancel-stripped context. The boundary is the commit, not the driver switch.)

Vic-Dev:
Hi @tjb9dc, in light of #3255 going in, can you please update this PR so that we can use it as regression tests?

tjb9dc:
@Vic-Dev done! Rebased onto current `main`, reworded the tests as regression coverage for #3255 (no logic changes were needed — the tests assert the desired behavior, so they pass on `main` as of #3255), and marked the PR ready for review. Verified locally against `main`:

```
--- PASS: TestIteratorNextRespectsContextDuringPoolAcquire (2.42s)
--- PASS: TestReverseExpandPoolStarvationDeadlock (2.10s)
```

Siddhant-K-code (commented):
@tjb9dc - can you sign the CLA?

tjb9dc:
@Siddhant-K-code done!

Siddhant-K-code (approved):
Overall, it looks good. Someone from @openfga/be-storage-maintainers / @openfga/backend  might be able to give a final sign-off!

PR #3250, opened 2026-08-08
fix: disable pgx prepared statements for postgres migrations

### LINKED ISSUE #1832: `migrate` returns 1 when migration already done
### Checklist

- [X] I have looked into the [README](https://github.com/openfga/openfga/blob/main/README.md) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://openfga.dev/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/openfga/openfga/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/openfga/openfga/releases/latest) of OpenFGA and the issue still persists.
- [X] I have searched the [OpenFGA Community](https://openfga.dev/community) and have not found a suitable solution or answer.

### Description

As part of our deployments we run a `migrate` every time before we start a `run` container. Now we discovered that, because we had to re-deploy the same version of openfga twice, the second time the migration errors.

`Error: failed to get db version: ERROR: relation "goose_db_version" already exists (SQLSTATE 42P07)`

We assume this might be unintended and something like a return `0` with a no migration necessary log should have happened.

### Expectation

When migration is already applied return code would be `0`

### Reproduction

1. Run migrate
2. Run migrate

### Store data

```yaml
model_file: |
  model
    schema 1.1
  type user
  type organization
    relations
      define member: [user]
tuples:
  - user: user:anne
    relation: member
    object: organization:openfga
  - user: user:bob
    relation: member
    object: organization:openfga
tests: # remove this if not a bug in Check or ListObjects API
  - name: test-1
    check:
      - user: user:anne
        object: organization:openfga
        assertions:
          member: true
    list_objects:
      - user: user:anne
        type: organization
        assertions:
          member:
            - organization:openfga
```

### OpenFGA version

latest

In Docker

Postgres

### OpenFGA Flags

-

### Logs

_No response_

pujitha24:
Motivation:
Running `openfga migrate` a second time against the same postgres
database can fail with:

  Error: failed to get db version: ERROR: relation "goose_db_version"
  already exists (SQLSTATE 42P07)

Multiple users report this only happens when the connection goes
through a connection pooler (PgBouncer/PgCat) running in transaction
pooling mode; a maintainer could not reproduce it against plain
postgres. pgx's own documentation states that its default of
server-side prepared statements is incompatible with PgBouncer:
a statement pgx believes is already prepared on a pooled connection
may not exist on the backend session that actually serves a later
query, since the pooler can swap backends between statements on the
same client connection. Separately, goose's EnsureDBVersion treats
*any* error from its initial version-check query as evidence that
the goose_db_version table is missing, and unconditionally issues a
raw CREATE TABLE (not IF NOT EXISTS) to recreate it. Chained
together: a stale-prepared-statement error from the pooler gets
misread by goose as "table missing," and the resulting CREATE TABLE
fails because the table was there all along - producing exactly the
reported error.

Approach:
For postgres only, append `default_query_exec_mode=simple_protocol`
to the connection URI used to open the migration database connection
(unless the caller already set that parameter explicitly). This
disables pgx's server-side prepared statement caching for the
migrate command, matching pgx's own documented recommendation for
PgBouncer compatibility. Migrations run infrequently and are not
performance sensitive, so this has no meaningful cost. The
username/password override logic that lived inline in RunMigrations
was extracted into buildPostgresMigrationURI unchanged, so the new
logic could be unit tested directly.

Validation:
- `go build ./...` passes.
- `go vet ./pkg/storage/migrate/... ./cmd/migrate/...` is clean.
- `golangci-lint run -c .golangci.yaml ./pkg/storage/migrate/...
  ./cmd/migrate/...` reports 0 issues.
- `go test -race -run TestBuildPostgresMigrationURI -v
  ./pkg/storage/migrate/...` passes a new targeted unit test
  covering: default_query_exec_mode is added when absent, an
  explicit caller-supplied value is not overridden, username/password
  overrides still work, existing credentials are preserved, and
  invalid URIs return an error.
- Not run: this sandbox has no Docker available, so the package's
  existing container-backed integration test
  (TestMigrateCommandRollbacks in migrate_test.go) and a live
  PgBouncer-in-transaction-mode reproduction were not executed here.
  The root-cause chain above was instead confirmed by reading the
  vendored pgx v5.10.0 and goose v3.27.3 source directly (pgx's
  documented PgBouncer incompatibility, its recognized
  default_query_exec_mode DSN parameter, and goose's
  EnsureDBVersionContext/createVersionTable behavior). User-visible
  behavior for non-pooled postgres connections and for mysql/sqlite
  is unchanged; the fix only removes a failure mode specific to
  transaction-pooling proxies.

Report: https://github.com/openfga/openfga/issues/1832
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #1832

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL migration connection handling by reliably applying migration-compatible query settings.
  * Preserved existing connection credentials and settings while correctly applying overrides.
  * Invalid migration connection URIs now return clear errors instead of being processed incorrectly.

* **Tests**
  * Added coverage for connection settings, credential handling, and invalid URI scenarios.


pujitha24:
This is rebased on main and all checks are green — happy to make any changes that would help move review along.

PR #3254, opened 2026-08-13
fix: update potential resolution changes logs to only log on InfoLevel

Vic-Dev:
#### What changed?
- Updates `Check` and `ListUsers` to only log potential resolution changes on `InfoLevel`
- Removes potential resolution changes for `Expand` -> logs are not needed

## Review Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced unnecessary v2 breaking-change diagnostics by honoring the configured logging level.
  - Updated Check and ListUsers breaking-change messages to informational logs instead of warnings.
  - Removed breaking-change warnings from Expand operations while preserving normal Expand processing and responses.
- **Logging**
  - Added support for checking the minimum enabled log level before potentially expensive logging work.
- **Documentation**
  - Updated the changelog to describe the revised v2 diagnostic logging behavior.


adriantam (approved):

pkg/server/check.go
```diff
@@ -285,7 +286,7 @@ func (s *Server) Check(ctx context.Context, req *openfgav1.CheckRequest) (*openf
 	//     failure) and we must inspect the schema ourselves: exclusion shapes
 	//     log unconditionally, per-userset shapes gate on !allowed to match
 	//     the v2-success log path above.
-	if isV2Fallback {
+	if s.logger.Level() <= zapcore.InfoLevel && isV2Fallback {
```
saad-h1, line 289: Wondering why we need this - my understanding is that when you set a log level, it should only cause that level and above to be produced. So if the log level was > info (e.g., warn/error), that should mean that no info logs are produced, no? I.e., seems like we don't need this check and can instead just change from warn to info and that's it?

Vic-Dev: The check is to prevent doing any work if the log level is higher than Info

PR #3264, opened 2026-08-19
fix(postgres): apply secondary datastore credentials

### LINKED ISSUE #3256: PostgreSQL secondary datastore credentials from environment variables are not applied
### Checklist

### Description

I found that the credentials configured through `OPENFGA_DATASTORE_SECONDARY_USERNAME` and `OPENFGA_DATASTORE_SECONDARY_PASSWORD` are not applied when OpenFGA creates the PostgreSQL secondary datastore connection.

The primary datastore uses:

- Username: `primary_user`
- Password: `primary_password`

The secondary datastore is configured with:

- Username: `secondary_user`
- Password: `secondary_password`

However, OpenFGA attempts to connect to the secondary datastore as `primary_user`.

### Reproduction

A minimal Docker Compose reproduction is available in my fork:

https://github.com/cxconj/openfga/tree/repro/secondary-datastore-credentials/repro/secondary-datastore-auth

The reproduction uses OpenFGA `v1.18.3`, PostgreSQL 17, and a PostgreSQL streaming read replica. All credentials are provided through environment variables, and neither datastore URI contains credentials.

To make the issue observable, the replica's local `pg_hba.conf` deliberately rejects connections from `primary_user` and allows connections from `secondary_user`. This is a reproduction-specific access rule and is not a default PostgreSQL read-replica restriction.

Run the reproduction with:

```bash
docker compose -f repro/secondary-datastore-auth/compose.yaml up -d
docker compose -f repro/secondary-datastore-auth/compose.yaml logs -f openfga
```
OpenFGA repeatedly retries the database connection and eventually exits:

```text
openfga-1 | ... INFO waiting for database {"attempt": 11}
openfga-1 | ... panic: initialize postgres datastore: configure secondary db: ping db: failed to connect to user=primary_user database=openfga: ... FATAL: pg_hba.conf rejects connection ... user "primary_user" ...
```

The corresponding error in the secondary PostgreSQL container is:

```
secondary-db-1 | ... FATAL: pg_hba.conf rejects connection ... user "primary_user", database "openfga", no encryption
```

The full container logs are included in the Logs section below.

### Expected behavior

The secondary datastore connection should use the credentials configured through OPENFGA_DATASTORE_SECONDARY_USERNAME and OPENFGA_DATASTORE_SECONDARY_PASSWORD.

### Actual behavior

The secondary datastore connection uses the primary datastore credentials instead.

### OpenFGA version

v1.18.3

In Docker

Postgres

### OpenFGA Flags

No command-line flags; configured via OPENFGA_DATASTORE_* environment variables.

### Logs

OpenFGA container logs:

```
openfga-1 | 2026-08-15T13:52:15.921Z INFO waiting for database {"attempt": 1}
openfga-1 | 2026-08-15T13:52:16.258Z INFO waiting for database {"attempt": 2}
openfga-1 | 2026-08-15T13:52:17.195Z INFO waiting for database {"attempt": 3}
...
openfga-1 | 2026-08-15T13:53:07.496Z INFO waiting for database {"attempt": 11}
openfga-1 | 2026-08-15T13:53:07.496Z INFO attempting to shutdown gracefully...
openfga-1 | 2026-08-15T13:53:07.496Z INFO graceful shutdown completed successfully
openfga-1 | 2026-08-15T13:53:07.499Z panic: initialize postgres datastore: configure secondary db: ping db:
failed to connect to `user=primary_user database=openfga`: 172.30.0.3:5432 (secondary-db): server error: FATAL:
pg_hba.conf rejects connection for host "172.30.0.4", user "primary_user", database "openfga", no encryption
(SQLSTATE 28000)
```

Secondary PostgreSQL container logs:

```
secondary-db-1 | 2026-08-15 13:52:15.921 UTC [54] FATAL: pg_hba.conf rejects connection for host "172.30.0.4",
user "primary_user", database "openfga", no encryption
secondary-db-1 | 2026-08-15 13:52:16.258 UTC [55] FATAL: pg_hba.conf rejects connection for host "172.30.0.4",
user "primary_user", database "openfga", no encryption
```

aayushbaluni:
## Description

`parseConfig` always read `cfg.Username` / `cfg.Password`, regardless of which datastore it was building. The secondary pool is created from the same `*sqlcommon.Config`, so it inherited the primary's credentials and `OPENFGA_DATASTORE_SECONDARY_USERNAME` / `_PASSWORD` had no effect.

`postgres.go` before this change:

```go
primaryDB, err := initDB(uri, cfg.Username != "" || cfg.Password != "", cfg)
...
secondaryDB, err = initDB(cfg.SecondaryURI, cfg.SecondaryUsername != "" || cfg.SecondaryPassword != "", cfg)
```

Both call sites correctly *detected* their own credentials, but only passed a `bool`. `parseConfig` then applied the primary pair either way.

The fix passes the credentials themselves instead of a flag. The `override` parameter becomes redundant, since both call sites derived it from exactly the values now being passed.

## Verification

Reproduced first, against `parseConfig` as it is actually called for the secondary:

```
secondary connection built with user="primary_user" password="primary_password"
    secondary user = "primary_user", want "secondary_user"
    secondary password = "primary_password", want "secondary_password"
--- FAIL: TestSecondaryCredentialsAreNotApplied
--- PASS: TestPrimaryCredentialsAreApplied     <- control, primary was never broken
```

After the change, all of `TestParseConfig`'s existing 5 subtests still pass, plus 4 new ones covering: secondary uses secondary credentials, primary still uses primary credentials, secondary with no explicit credentials keeps the ones embedded in its URI, and password-only leaves the URI's username intact.

I also added `TestPostgresSecondaryUsesSecondaryCredentials`, which runs `New()` against a real primary + streaming replica via the existing test containers, with **both URIs supplied without credentials** so they can only come from config. That is the shape described in the issue. It passes (~31s locally with Docker).

## Notes for reviewers

Two things I want to flag rather than leave for you to find:

1. **The integration test cannot distinguish this bug on its own.** The replica is a physical `pg_basebackup` copy, so it shares the primary's roles. Swapping the credentials is therefore invisible to it. I verified this: reverting the secondary call site alone keeps that test green. The unit tests are what actually pin the behaviour. Genuinely covering the call site would need a fixture where the replica has a distinct role, which felt out of scope here — happy to add it if you'd prefer.

2. **`else if parsed.User != nil` is dead code**, both before and after this change. `pgxpool.ParseConfig` has already populated `User` and `Password` from the URI by that point, so the branch never changes anything. I confirmed it directly rather than assuming. I left it alone to keep the diff focused, but it could be removed separately.

Closes #3256

---

Generative AI was used to help investigate and prepare this change, under the Linux Foundation Generative AI policy referenced in the contributing guide. All results quoted above were executed locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PostgreSQL connections so primary and secondary databases use their configured credentials independently.
  * Preserved credentials embedded in database URIs when explicit values are not provided.
  * Improved handling of partial credential configuration, such as password-only overrides.

* **Tests**
  * Added coverage verifying credential selection for primary and secondary database connections.
  * Added regression testing to confirm both database pools become ready with separate credentials.

### DIFF
--- a/pkg/storage/postgres/postgres.go
+++ b/pkg/storage/postgres/postgres.go
@@ -59,30 +59,33 @@ type Datastore struct {
 // Ensures that Datastore implements the OpenFGADatastore interface.
 var _ storage.OpenFGADatastore = (*Datastore)(nil)
 
-func parseConfig(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Config, error) {
+// parseConfig builds a pool config for uri, applying username and password when
+// they are set. Each datastore passes its own credentials: the secondary has a
+// separate pair and must not inherit the primary's.
+func parseConfig(uri string, username, password string, cfg *sqlcommon.Config) (*pgxpool.Config, error) {
 	c, err := pgxpool.ParseConfig(uri)
 	if err != nil {
 		return nil, fmt.Errorf("pgxpool parse postgres connection uri: %w", err)
 	}
 
-	if override {
+	if username != "" || password != "" {
 		parsed, err := url.Parse(uri)
 		if err != nil {
 			return nil, fmt.Errorf("url parse postgres connection uri: %w", err)
 		}
 
-		if cfg.Username != "" {
-			c.ConnConfig.User = cfg.Username
+		if username != "" {
+			c.ConnConfig.User = username
 		} else if parsed.User != nil {
 			c.ConnConfig.User = parsed.User.Username()
 		}
 
 		switch {
-		case cfg.Password != "":
-			c.ConnConfig.Password = cfg.Password
+		case password != "":
+			c.ConnConfig.Password = password
 		case parsed.User != nil:
-			if password, ok := parsed.User.Password(); ok {
-				c.ConnConfig.Password = password
+			if pw, ok := parsed.User.Password(); ok {
+				c.ConnConfig.Password = pw
 			}
 		}
 	}
@@ -210,9 +213,9 @@ func createBeforeConnect(logger logger.Logger, provider PassfileProvider) func(c
 	}
 }
 
-// initDB initializes a new postgres database connection.
-func initDB(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Pool, error) {
-	c, err := parseConfig(uri, override, cfg)
+// initDB initializes a new postgres database connection with the given credentials.
+func initDB(uri string, username, password string, cfg *sqlcommon.Config) (*pgxpool.Pool, error) {
+	c, err := parseConfig(uri, username, password, cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -227,14 +230,14 @@ func initDB(uri string, override bool, cfg *sqlcommon.Config) (*pgxpool.Pool, er
 
 // New creates a new [Datastore] storage.
 func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
-	primaryDB, err := initDB(uri, cfg.Username != "" || cfg.Password != "", cfg)
+	primaryDB, err := initDB(uri, cfg.Username, cfg.Password, cfg)
 	if err != nil {
 		return nil, fmt.Errorf("initialize postgres connection: %w", err)
 	}
 
 	var secondaryDB *pgxpool.Pool
 	if cfg.SecondaryURI != "" {
-		secondaryDB, err = initDB(cfg.SecondaryURI, cfg.SecondaryUsername != "" || cfg.SecondaryPassword != "", cfg)
+		secondaryDB, err = initDB(cfg.SecondaryURI, cfg.SecondaryUsername, cfg.SecondaryPassword, cfg)
 		if err != nil {
 			return nil, fmt.Errorf("initialize postgres connection: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


aayushbaluni:
Good catch, and correct. The replica is a physical `pg_basebackup` copy, so it shares the primary's roles, which meant identical credentials made the test pass before the fix too. I had flagged that limitation in the PR description but should have just fixed it.

Now the test creates a dedicated `openfga_secondary` role on the **primary before the base backup**, so it replicates to the standby, and asserts what the secondary pool actually authenticated as:

```go
require.NoError(t, ds.secondaryDB.QueryRow(
    context.Background(), "SELECT current_user").Scan(&connectedAs))
require.Equal(t, secondaryUser, connectedAs)
```

Verified it now catches the regression by reverting the fix and re-running:

```
baseline (fix applied):            PASS
secondary uses primary creds:      FAIL  <- caught
    secondary pool must authenticate as the configured secondary user
```

Worth noting the readiness check alone was not enough. With the fix reverted it still reports `primary: ready, secondary: ready`, because the primary role can connect to the replica perfectly well. Only `current_user` distinguishes the two.

The role also needs `USAGE ON SCHEMA public` and `SELECT ON ALL TABLES`, otherwise readiness fails on `goose_db_version` before reaching the assertion.

PR #3277, opened 2026-08-31
feat(authn): add optional JWT token cache for authenticated requests

victorvarza:

Every authenticated request in OpenFGA re-parses and re-verifies the incoming JWT bearer token from scratch. While the JWK signing keys are already cached (refreshed every 24-48h), the RSA signature verification and claims validation still runs on every single call, costing roughly 0.1–0.5 ms per request. 
At high QPS this becomes a measurable share of the authentication path, even though the same tokens are re-presented repeatedly by the same clients within a short window.

#### How is it being solved?

Introduce an opt-in in-memory cache in front of the Authenticator interface that stores the successfully-parsed `AuthClaims` keyed by the raw bearer token. On a cache hit, we skip signature verification and claims parsing entirely and return the previously validated claims. On a cache miss or on any authentication error, we fall through to the underlying authenticator; errors are never cached so revocation/expiry paths continue to work as before.

Trade-off: while the cache is warm, a token remains valid up to TTL past any external revocation. Since JWTs are already non-revocable without a denylist, this only widens the pre-existing revocation lag by at most TTL, operators should size TTL against their acceptable revocation window.

- `internal/authn/cached_authenticator.go` (new): `CachedAuthenticator` wraps any `Authenticator `(works transparently with oidc, preshared, or noop). Successful claims are cached under the key AT + <bearer token>; errors bypass the cache.
- `internal/authn/cached_authenticator_test.go` (new): Unit tests covering cache hits, distinct-token isolation, error non-caching, and missing bearer token.
- `pkg/server/config/config.go`: New `AuthnTokenCacheConfig` struct (Enabled, TTL, MaxSize) on `AuthnConfig`, defaults (false, 5m, 10000), and validation (TTL > 0 when enabled) in `verifyCacheConfig`.
- `cmd/run/run.go`: New flags -`-authn-token-cache-enabled`, `--authn-token-cache-ttl`, `--authn-token-cache-max-size`; `authenticatorConfig` now wraps the constructed authenticator with `CachedAuthenticator` when the flag is enabled.
- `cmd/run/flags.go`: Bound the three new flags to config keys and environment variables (`OPENFGA_AUTHN_TOKEN_CACHE_ENABLED`, `OPENFGA_AUTHN_TOKEN_CACHE_TTL`, `OPENFGA_AUTHN_TOKEN_CACHE_MAX_SIZE`).
- `docs/caching.md`: New "Auth Token Cache" section with description, cache key, trade-offs, and the standard configuration table; added to the top-level overview list.

Co-authored-by: Claude noreply@anthropic.com

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional authentication token caching for OIDC authentication.
  * Added configuration options, command-line flags, and environment variables for enabling caching, setting duration, and limiting cache size.

* **Bug Fixes**
  * Cache duration now respects each token’s expiration time.
  * Expired tokens and authentication failures are not cached.
  * Improved handling of requests without bearer tokens and protected cached credentials.

* **Documentation**
  * Documented token caching behavior, configuration, bypass conditions, and expiration rules.

### DIFF
--- a/.config-schema.json
+++ b/.config-schema.json
@@ -357,6 +357,31 @@
                 "oidc": {
                     "description": "The OIDC provider specific settings. This must be set if 'authn.method=oidc'.",
                     "$ref": "#/definitions/oidc"
+                },
+                "tokenCache": {
+                    "type": "object",
+                    "description": "Configuration for caching authenticated JWT tokens to avoid repeated signature verification. Only applies when authn.method=oidc.",
+                    "properties": {
+                        "enabled": {
+                            "description": "Enable caching of authenticated JWT tokens.",
+                            "type": "boolean",
+                            "default": false,
+                            "x-env-variable": "OPENFGA_AUTHN_TOKEN_CACHE_ENABLED"
+                        },
+                        "ttl": {
+                            "description": "The TTL of cached authentication tokens. The effective TTL per entry is min(this value, token exp - now).",
+                            "type": "string",
+                            "default": "5m",
+                            "x-env-variable": "OPENFGA_AUTHN_TOKEN_CACHE_TTL"
+                        },
+                        "maxSize": {
+                            "description": "The maximum number of tokens to cache.",
+                            "type": "integer",
+                            "default": 10000,
+                            "minimum": 1,
+                            "x-env-variable": "OPENFGA_AUTHN_TOKEN_CACHE_MAX_SIZE"
+                        }
+                    }
                 }
 
             }
--- a/cmd/run/flags.go
+++ b/cmd/run/flags.go
@@ -92,6 +92,15 @@ func bindRunFlagsFunc(flags *pflag.FlagSet) func(*cobra.Command, []string) {
 		util.MustBindPFlag("authn.oidc.clientIdClaims", flags.Lookup("authn-oidc-client-id-claims"))
 		util.MustBindEnv("authn.oidc.clientIdClaims", "OPENFGA_AUTHN_OIDC_CLIENT_ID_CLAIMS")
 
+		util.MustBindPFlag("authn.tokenCache.enabled", flags.Lookup("authn-token-cache-enabled"))
+		util.MustBindEnv("authn.tokenCache.enabled", "OPENFGA_AUTHN_TOKEN_CACHE_ENABLED")
+
+		util.MustBindPFlag("authn.tokenCache.ttl", flags.Lookup("authn-token-cache-ttl"))
+		util.MustBindEnv("authn.tokenCache.ttl", "OPENFGA_AUTHN_TOKEN_CACHE_TTL")
+
+		util.MustBindPFlag("authn.tokenCache.maxSize", flags.Lookup("authn-token-cache-max-size"))
+		util.MustBindEnv("authn.tokenCache.maxSize", "OPENFGA_AUTHN_TOKEN_CACHE_MAX_SIZE")
+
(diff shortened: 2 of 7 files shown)


Santoshkumarpuppala:
The exp-capping behaviour added in `5939712`/`fb2c1e4` isn't reachable by the current tests. Mutation result first, since that's the part that isn't already on the thread:

**Change the call site and nothing fails.** `cached_authenticator.go:61` is `ttl := c.effectiveTTL(authHeader)`. Replace it with `ttl := c.ttl` — removing the cap entirely — and all 12 tests in `internal/authn` still pass:

```
ok  github.com/openfga/openfga/internal/authn            0.312s
ok  github.com/openfga/openfga/internal/authn/oidc       (cached)
ok  github.com/openfga/openfga/internal/authn/presharedkey (cached)
```

**Control, so that zero means something.** Mutating `effectiveTTL` itself instead — the expired branch at `:89-91`, `return 0` → `return c.ttl` — does fail: `TestEffectiveTTL_JWTAlreadyExpired` at `cached_authenticator_test.go:184`, expected `0s`, actual `5m0s`. So the suite can fail and the function is genuinely covered. It's the wiring that isn't.

(My first attempt at a control was invalid — I deleted the function body and got a build failure from the now-unused `jwt` import, which only proves the compiler works. Worth stating so the result above can be read for what it is.)

**Why the new tests don't reach it.** The four `TestEffectiveTTL_*` cases construct `&CachedAuthenticator{ttl: 5 * time.Minute}` with no cache and no delegate and call `cached.effectiveTTL(token)` directly — 4 direct calls, 0 calls to `Authenticate` across their bodies. They pin the pure function, so they stay green no matter what the call site passes to `cache.Set`.

**A test that discriminates**, verified in both directions — passes on HEAD, fails on the `ttl := c.ttl` mutant with *"An error is expected but got nil: expired JWT must not be served from cache"*:

```go
func TestCachedAuthenticator_DoesNotServeTokenPastItsExp(t *testing.T) {
	cache := newTestCache(t)
	delegate := &mockAuthenticator{claims: &authclaims.AuthClaims{Subject: "user:alice"}}

	// Configured TTL is 5 minutes; the token itself expires in 1 second.
	cached := NewCachedAuthenticator(delegate, cache, 5*time.Minute)
	token := buildUnsignedJWT(t, time.Now().Add(1*time.Second))
	ctx := contextWithBearer(token)

	claims, err := cached.Authenticate(ctx)
	require.NoError(t, err)
	require.Equal(t, "user:alice", claims.Subject)
	require.Equal(t, 1, delegate.callCount)

	time.Sleep(1500 * time.Millisecond)

	delegate.claims = nil
	delegate.err = ErrUnauthenticated

	_, err = cached.Authenticate(ctx)
	require.Error(t, err, "expired JWT must not be served from cache")
	require.Equal(t, 2, delegate.callCount, "delegate must be re-consulted after the token's exp")
}
```

No change to `mockAuthenticator` is needed — `claims` and `err` are plain fields, so reassigning them between the two `Authenticate` calls is enough. That transition, delegate-succeeds then delegate-rejects, is the one a TTL bound exists to permit, and no existing test performs it: every case sets the pair once at construction.

It costs 1.5s of wall clock. If that's unwelcome in the unit suite, the same discrimination works by injecting a clock, but the sleep keeps it to one file and no production change.

Reusable as-is if it's useful — happy to open it as a PR against this branch instead if you'd prefer.

PR #3286, opened 2026-09-04
feat(authz): support client ids containing tuple-reserved characters

nraufiero:
Sanitize claim-sourced client ids so they form valid tuple users, unblocking Kubernetes ServiceAccount subjects as access-control clients.

The [access control](https://openfga.dev/docs/getting-started/setup-openfga/access-control#04-grant-access-to-a-store) feature grants an OIDC client access by writing an `application:<client_id>` tuple, where `<client_id>` comes straight from a claim (`azp`/`client_id` by default, or any claim named via `--authn-oidc-client-id-claims`). `tuple.IsValidUser` allows at most one `:` in a tuple user, but `ClientIDType.String()` builds `application:<client_id>` with no validation or transformation of the raw claim. A client id claim with more than one colon makes every access-control check for that client fail with "invalid tuple 'user' field format".

This is exactly the shape of a Kubernetes ServiceAccount token subject, the standard identity for an in-cluster workload calling another in-cluster service: `system:serviceaccount:<namespace>:<name>`, three colons. A cluster's own service identities can't be used as access-control clients today.

#### How is it being solved?

`ClientIDType.String()` now sanitizes the client id before building the tuple user, replacing every character `tuple.IsValidUser` rejects (`:`, `#`, space, or a Unicode control character) with `_`. `system:serviceaccount:openfga:my-service` becomes `system_serviceaccount_openfga_my-service`, so `application:system_serviceaccount_openfga_my-service` is a valid tuple user.

This can only change behavior for client ids that don't work today. Every consumer of `ClientIDType.String()`'s output is a `Check` request's `User` field (`ListAuthorizedStores` and `individualAuthorize`), so any claim value already containing one of these characters was already producing an invalid tuple and failing every check.

- `pkg/tuple/tuple.go`: new exported `SanitizeUserID(s string) string`, next to the `IsValidUserID` it's built against.
- `pkg/tuple/tuple_test.go`: `FuzzSanitizeUserID`, mirroring `FuzzIsValidUserID`'s style, asserting `IsValidUserID(SanitizeUserID(s))` holds for any non-empty `s` and that sanitizing an already-valid id is a no-op; `BenchmarkSanitizeUserID`.
- `internal/authz/authz.go`: `ClientIDType.String()` calls `tuple.SanitizeUserID`.
- `internal/authz/authz_test.go`: `TestClientIDType_String`, covering a plain client id (unchanged) and a Kubernetes ServiceAccount subject (sanitized).

#### Notes for reviewers

Sanitization only runs where the client id is read (`ClientIDType.String()`), not where access is granted. Any tuple written to grant a client access, most importantly the [bootstrap admin tuple](https://openfga.dev/docs/modeling/access-control#bootstrapping-access-control), must use the same sanitized id (e.g. `application:system_serviceaccount_openfga_my-service`), not the raw claim value. `tuple.SanitizeUserID` is exported so operator tooling can compute the exact sanitized form before writing that tuple.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for sanitizing client IDs before using them in access-control tuples.
  * Invalid user-ID characters are replaced with underscores, including support for Unicode and invalid UTF-8 input.

* **Bug Fixes**
  * Client IDs such as Kubernetes service-account subjects now produce valid, consistently formatted application tuples.

* **Tests**
  * Added coverage for plain, Unicode, invalid, and already-valid client IDs.

### DIFF
--- a/internal/authz/authz.go
+++ b/internal/authz/authz.go
@@ -68,8 +68,10 @@ func (s StoreIDType) String() string {
 
 type ClientIDType string
 
+// String sanitizes c with tuple.SanitizeUserID before building the tuple
+// user. Any tuple granting a client access must use the same sanitized id.
 func (c ClientIDType) String() string {
-	return fmt.Sprintf("%s:%s", ApplicationType, string(c))
+	return fmt.Sprintf("%s:%s", ApplicationType, tuple.SanitizeUserID(string(c)))
 }
 
 type ModuleIDType string
--- a/internal/authz/authz_test.go
+++ b/internal/authz/authz_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openfga/openfga/pkg/authclaims"
 	"github.com/openfga/openfga/pkg/logger"
 	"github.com/openfga/openfga/pkg/testutils"
+	"github.com/openfga/openfga/pkg/tuple"
 	"github.com/openfga/openfga/pkg/typesystem"
 )
 
@@ -85,6 +86,18 @@ func TestGetStoreID(t *testing.T) {
 
 		authorizer := NewAuthorizer(nil, mockServer, logger.NewNoopLogger())
 		require.Empty(t, authorizer.AccessControlStoreID())
+	})
+}
+
+func TestClientIDType_String(t *testing.T) {
+	t.Run("plain_client_id_is_unchanged", func(t *testing.T) {
+		require.Equal(t, "application:test-client", ClientIDType("test-client").String())
+	})
+	t.Run("kubernetes_serviceaccount_subject_is_sanitized", func(t *testing.T) {
+		clientID := ClientIDType("system:serviceaccount:openfga:permissions-deployer")
+		result := clientID.String()
+		require.Equal(t, "application:system_serviceaccount_openfga_permissions-deployer", result)
+		require.True(t, tuple.IsValidUser(result))
 	})
 }
 
--- a/pkg/tuple/tuple.go
+++ b/pkg/tuple/tuple.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 	"unsafe"
 
 	"google.golang.org/protobuf/types/known/structpb"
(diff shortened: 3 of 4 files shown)


rhamzeh:
Thanks for the idea and the PR @nraufiero!

In general sanitizing would be good, but two problems I forsee:
* `system:serviceaccount:openfga:permissions-deployer` being granted and converted to `system_serviceaccount_openfga_permissions-deployer` makes a client ID that is literally `system_serviceaccount_openfga_permissions-deployer` also pass as true. This is probably highly unlikely. But if even a single user in the world has this use-case, this becomes a CVE.
* Maybe we do something like we always base64() the clientID (based on an opt in config) instead, but that would also increase the length of the ID
* Maybe we allow the user to pass in/configure a sanitization function that would work for them, but that would also open security issues if they misconfigure it
* I know we would like to be even more strict in what we allow as characters, so maybe we do need to keep that in mind as we solve this.

@senojj - you are our go to guru for all things regarding character validation and sanitization, what do you think?

--

Side-note: Whatever the result of this, OpenFGA users, who are using OpenFGA like us, will face the same problem, so our solution for this should lead to documentation/guidance on how people should resolve it. (cc @aaguiarz)

senojj:
@rhamzeh is correct, converting an invalid character to a valid character will create the opportunity for collisions between different values.

The most appropriate solution to supporting reserved characters within user-supplied tuple components would be something akin to percent-encoding. Doing so would preserve the value's uniqueness while preventing a component value character from being confused with a reserved component delimiter character.

However, like with percent-encoding, we would need a byte sequence that indicates the beginning of a hexadecimal sequence. Just using the percent character ('%') could be a candidate, but it is currently an unreserved character in tuple component values.

It would be possible to implement percent-encoding in a non-breaking way; if the two characters that proceed a percent character are not valid characters in the hexadecimal alphabet, do not attempt to decode that sequence.

Additionally, we would need to consider the impact of storing encoded sequences in the database, because this fundamentally changes the string value that is persisted. As long as values are always encoded before using them in comparisons, such as within a query filter, this shouldn't cause an issue.

As an added benefit, percent-encoded values are safe in most string contexts. Therefore, we can simply encode potentially dangerous characters or byte sequences instead of rejecting them outright with an error.

senojj:
Thinking about it more, regardless of how we implement the percent-encoding, a data migration would be necessary to encode existing values in the database. Otherwise there would be mismatches between the encoded input and the stored value(s).

While this would be a good way to handle this scenario, I'm primarily concerned with how this would impact users who have implemented their own custom datastore adapters.

nraufiero:
Thanks @rhamzeh and @senojj for the comments! Totally agree on the possibility for clash. The percent-encoding makes sense to me. I like the opt-in flag @rhamzeh mentioned as a way to gate this new functionality. So users must knowingly enable that flag to enable percent-encoding, and that puts the burden on them to understand any conflicts that encoding might cause. And fwiw this change is specifically for clientID on the access control feature, which is still marked as [experimental](https://openfga.dev/docs/getting-started/setup-openfga/access-control). Not sure if that gives us more wiggle room here, of course without breaking anyone's current implementation.

Issue #3282, opened 2026-09-02
Secondary DB Support For MySQL Datastore


chiragsoni-eternal:
### Checklist

- [x] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

Currently MySQL datastore only support a single primary DB, which restricts users to scale for read operations traffic separately which is the majorly used scenario as permissions don't change too frequent but check authorization endpoints will have most of the traffic on anyone's openfga server.

### Describe the ideal solution

Adding secondary db support in mysql datastore in the same way we have it for postgresql, making all the write operations go to primary db and all read specific operation which does not require high consistency can go to secondary db which will be a reader. this implementation will have to support a consistency based DB choice logic similar to what we have in PostgreSQL datastore.

### Alternatives and current workarounds

Current workaround is just to vertically scale up the writer itself in mysql, or have a sharded mysql cluster where we can have multiple writers to distribute load, but this is not ideal as traffic will always be read heavy for an authorization server.

blev8824-ai:
Read-scaling the datastore is reasonable for a read-heavy authorization server, but the tricky part is which reads can tolerate a secondary. A Check that reads from a replication-lagged secondary can return allow for a relationship that was just revoked on the primary, which is a stale-permit -> false allow, and for an authorization store that is a security hole, not just a consistency annoyance.

The important principle to encode in the consistency-selection logic: revocations and grant deletions must always hit the synchronized/primary path (or the implementation must wait for caught-up LSN), even if ordinary read checks are routed to a secondary. Stale allow after revoke is the fail-open case you cannot accept. A safe default is: write-path reads (the decision that immediately follows a mutation, or any check that must reflect a recent revocation) go to the primary; warm long-lived relationship checks can use a secondary with the understanding that re-validating at enforcement time still happens.

This is exactly the decision-vs-execution distinction: the stored relationship may be served from a replica, but the authorization decision has to be both correct and evidenced at the moment of enforcement. AgentKey keeps that decision with a verifiable record regardless of which datastore served the tuple: https://agentkey.us

PR #3290, opened 2026-09-07
feat[mysql-storage]: added secondary db support in mysql to use read replicas for operations which require low consistency

chiragsoni-eternal:
feat[mysql-storage]: added secondary db support in mysql to use

The MySQL datastore backend had no support for read/write splitting via a secondary (read-replica) database. All queries — reads and writes — were sent to a single connection pool pointing at the primary. The `--datastore-secondary-uri` flag was silently accepted but had no effect on MySQL, and its help text explicitly stated "for postgres only".

#### How is it being solved?

The MySQL datastore now maintains two `*sql.DB` connection pools:
- **`primaryDB`** — all writes (`Write`, `WriteAuthorizationModel`, `CreateStore`, `DeleteStore`)
- **`secondaryDB`** — all reads (`Check`, `ReadTuples`, `ListObjects`, `ListUsers`, `ReadAuthorizationModel`, `ReadChanges`) when `--datastore-secondary-uri` is set

A `getSQLDB()` helper encapsulates the routing decision, falling back to `primaryDB` when no secondary is configured. This mirrors the existing PostgreSQL implementation exactly.

When the secondary goes down, `IsReady()` detects the failure and reports `NOT_SERVING`, signalling load balancers to remove the pod. The process stays alive and health recovers automatically when the replica reconnects.

- **`pkg/storage/mysql/mysql.go`** — dual pools (`primaryDB`, `secondaryDB`), extracted `configureDB()`, added `getSQLDB()` routing helper, updated `IsReady()` to ping both pools, updated `Close()` to release both pools and their Prometheus collectors
- **`pkg/storage/mysql/mysql_test.go`** — renamed 18× `ds.db` → `ds.primaryDB`; added `TestGetSQLDB` (routing unit test, no Docker), `TestNew/bad_secondary_uri` (error path), `TestMySQLDatastoreStatusWithSecondaryDB` (integration test with real secondary pool)
- **`pkg/testfixtures/storage/mysql.go`** — implemented `CreateSecondary()` / `GetSecondaryConnectionURI()`: creates a sibling database inside the running MySQL container and runs migrations on it, providing a genuine second pool for integration tests
- **`pkg/storage/sqlcommon/sqlcommon.go`** — removed dead `ReadAuthorizationModel` and `FindLatestAuthorizationModel` functions (MySQL backend now inlines its own routed copies, same as PostgreSQL)
- **`cmd/run/run.go`** and **`.config-schema.json`** — updated `--datastore-secondary-uri` description from "for postgres only" to "for postgres and mysql"
- **`CHANGELOG.md`** — added entry under `## [Unreleased] → ### Added`

## References

* https://github.com/openfga/openfga/issues/3282

## Review Checklist
- [x] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev)
  - `TestGetSQLDB` — unit test for read routing logic (no Docker)
  - `TestNew/bad_secondary_uri` — error path when secondary URI is invalid
  - `TestMySQLDatastoreStatusWithSecondaryDB` — integration test with real secondary pool
  - All existing MySQL datastore tests pass with the `primaryDB` field rename

## Summary by CodeRabbit

- **New Features**
  - Added MySQL read-replica support through an optional secondary datastore connection.
  - Routes eligible reads to the replica while honoring consistency preferences.
  - Keeps writes and higher-consistency reads on the primary datastore.
  - Readiness checks now include both primary and secondary datastores.

- **Documentation**
  - Clarified that secondary datastore URIs support both PostgreSQL and MySQL.
  - Added configuration, routing, read-after-write considerations, and compatibility notes for MySQL read replicas.

### DIFF
--- a/pkg/storage/mysql/mysql.go
+++ b/pkg/storage/mysql/mysql.go
@@ -36,32 +36,33 @@ func startTrace(ctx context.Context, name string) (context.Context, trace.Span) 
 
 // Datastore provides a MySQL based implementation of [storage.OpenFGADatastore].
 type Datastore struct {
-	stbl                   sq.StatementBuilderType
-	db                     *sql.DB
-	dbInfo                 *sqlcommon.DBInfo
-	logger                 logger.Logger
-	dbStatsCollector       prometheus.Collector
-	maxTuplesPerWriteField int
-	maxTypesPerModelField  int
-	versionReady           bool
+	stbl                      sq.StatementBuilderType // bound to primaryDB; used for writes
+	primaryDB                 *sql.DB
+	secondaryDB               *sql.DB // nil if not configured
+	dbInfo                    *sqlcommon.DBInfo
+	logger                    logger.Logger
+	primaryDBStatsCollector   prometheus.Collector
+	secondaryDBStatsCollector prometheus.Collector
+	maxTuplesPerWriteField    int
+	maxTypesPerModelField     int
+	versionReady              bool
 }
 
 // Ensures that Datastore implements the OpenFGADatastore interface.
 var _ storage.OpenFGADatastore = (*Datastore)(nil)
 
-// New creates a new [Datastore] storage.
-func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
-	if cfg.Username != "" || cfg.Password != "" {
+// initDB opens a MySQL connection, applying optional DSN credential overrides.
+func initDB(uri, username, password string) (*sql.DB, error) {
+	if username != "" || password != "" {
 		dsnCfg, err := mysql.ParseDSN(uri)
 		if err != nil {
 			return nil, fmt.Errorf("parse mysql connection dsn: %w", err)
 		}
-
-		if cfg.Username != "" {
-			dsnCfg.User = cfg.Username
+		if username != "" {
+			dsnCfg.User = username
 		}
-		if cfg.Password != "" {
-			dsnCfg.Passwd = cfg.Password
+		if password != "" {
+			dsnCfg.Passwd = password
 		}
 		dsnCfg.AllowNativePasswords = true
 		uri = dsnCfg.FormatDSN()
@@ -71,15 +72,15 @@ func New(uri string, cfg *sqlcommon.Config) (*Datastore, error) {
 	if err != nil {
 		return nil, fmt.Errorf("initialize mysql connection: %w", err)
 	}
-	return NewWithDB(db, cfg)
+	return db, nil
 }
 
-// NewWithDB creates a new [Datastore] storage with the provided database connection.
-func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
+// configureDB applies connection-pool settings, performs a ping with exponential
+// backoff, and optionally registers a Prometheus collector for the given db.
+func configureDB(db *sql.DB, cfg *sqlcommon.Config, dbName string) (prometheus.Collector, error) {
 	if cfg.MaxIdleConns != 0 {
-		db.SetMaxIdleConns(cfg.MaxIdleConns) // default is 2, not retaining connections(0) would be detrimental for performance
+		db.SetMaxIdleConns(cfg.MaxIdleConns) // default is 2; retaining connections is important for performance
 	}
-
 	db.SetMaxOpenConns(cfg.MaxOpenConns)
 	db.SetConnMaxIdleTime(cfg.ConnMaxIdleTime)
 	db.SetConnMaxLifetime(cfg.ConnMaxLifetime)
@@ -90,9 +91,8 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
 		ctx, cancel := context.WithTimeout(context.Background(), cfg.PingTimeout)
 		defer cancel()
 
-		err := db.PingContext(ctx)
-		if err != nil {
-			cfg.Logger.Info("waiting for database", zap.Int("attempt", attempt))
+		if err := db.PingContext(ctx); err != nil {
+			cfg.Logger.Info("waiting for database", zap.String("db", dbName), zap.Int("attempt", attempt))
 			attempt++
 			return err
 		}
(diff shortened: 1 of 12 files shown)


DanCech:
This is looking great, just a few things that need attention:
- the changelog entry should be under the Unreleased section, and the PR link should be on the same line. It also needs to mention the breaking change to `mysql.NewWithDB` and that `sqlcommon.FindLatestAuthorizationModel` and `sqlcommon.ReadAuthorizationModel` were removed.
- The changelog describes the routing rule incorrectly. It says reads with `MINIMIZE_LATENCY` are routed to the replica, but `getSQLDB` sends everything except `HIGHER_CONSISTENCY` there, so requests that never set a consistency preference now go to the replica. That matches postgres, but the note should say "all reads except those requesting `HIGHER_CONSISTENCY`". It should also mention that GetStore, ListStores, ReadAssertions, ReadAuthorizationModel, ReadAuthorizationModels, FindLatestAuthorizationModel, and ReadChanges are hardcoded to the replica regardless of what the caller asked for. Read-after-write on any of those can return stale or missing data against a lagging replica, and operators need to know that before turning the flag on.
- close the primary connection leak instead of just commenting it, the addition is small:
```
		if primaryCollector != nil {
			prometheus.Unregister(primaryCollector)
		}
		_ = primaryDB.Close()
```
- TestNew/bad_secondary_uri burns 20 seconds in ping backoff. Set a short PingRetryMaxElapsedTime in that test's config or use a DSN that fails ParseDSN.
- initDB wraps with "initialize mysql connection" and New wraps again, producing `initialize mysql primary connection: initialize mysql connection: ....` Drop the inner wrap.
- read() builds its query with sq.StatementBuilder.RunWith(db) while ReadUsersetTuples and ReadStartingWithUser use a bare sq.StatementBuilder. Both work, because NewRowGetter only calls builder.ToSql() and executes through the connector, so the runner is never used -- but the inconsistency will tempt someone to "fix" the wrong one later.
- nothing yet proves a read actually lands on the secondary. Since CreateSecondary gives you a real second database with the same schema, you can insert a distinguishable row directly into it and assert that a default-consistency Read returns it while HIGHER_CONSISTENCY does not.
- a test to verify the correct behavior for secondary username/password handling would be good too, since that's an open bug with the postgres implementation

chiragsoni-eternal:
> This is looking great, just a few things that need attention:
> 
> * the changelog entry should be under the Unreleased section, and the PR link should be on the same line. It also needs to mention the breaking change to `mysql.NewWithDB` and that `sqlcommon.FindLatestAuthorizationModel` and `sqlcommon.ReadAuthorizationModel` were removed.
> * The changelog describes the routing rule incorrectly. It says reads with `MINIMIZE_LATENCY` are routed to the replica, but `getSQLDB` sends everything except `HIGHER_CONSISTENCY` there, so requests that never set a consistency preference now go to the replica. That matches postgres, but the note should say "all reads except those requesting `HIGHER_CONSISTENCY`". It should also mention that GetStore, ListStores, ReadAssertions, ReadAuthorizationModel, ReadAuthorizationModels, FindLatestAuthorizationModel, and ReadChanges are hardcoded to the replica regardless of what the caller asked for. Read-after-write on any of those can return stale or missing data against a lagging replica, and operators need to know that before turning the flag on.
> * close the primary connection leak instead of just commenting it, the addition is small:
> 
> ```
> 		if primaryCollector != nil {
> 			prometheus.Unregister(primaryCollector)
> 		}
> 		_ = primaryDB.Close()
> ```
> 
> * TestNew/bad_secondary_uri burns 20 seconds in ping backoff. Set a short PingRetryMaxElapsedTime in that test's config or use a DSN that fails ParseDSN.
> * initDB wraps with "initialize mysql connection" and New wraps again, producing `initialize mysql primary connection: initialize mysql connection: ....` Drop the inner wrap.
> * read() builds its query with sq.StatementBuilder.RunWith(db) while ReadUsersetTuples and ReadStartingWithUser use a bare sq.StatementBuilder. Both work, because NewRowGetter only calls builder.ToSql() and executes through the connector, so the runner is never used -- but the inconsistency will tempt someone to "fix" the wrong one later.
> * nothing yet proves a read actually lands on the secondary. Since CreateSecondary gives you a real second database with the same schema, you can insert a distinguishable row directly into it and assert that a default-consistency Read returns it while HIGHER_CONSISTENCY does not.
> * a test to verify the correct behavior for secondary username/password handling would be good too, since that's an open bug with the postgres implementation

Hi @DanCech thanks for the review, I have addressed all the comments into this [commit 56c94f4](https://github.com/openfga/openfga/pull/3290/commits/56c94f415cfd7c6f55b33b3e39a8a333d90f05c8), please have a look once you have some time

pkg/storage/mysql/mysql.go
```diff
@@ -104,54 +104,123 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
...
+
+	return NewWithDB(primaryDB, secondaryDB, cfg)
+}
+
+// NewWithDB creates a new [Datastore] storage with the provided database connections.
+// SecondaryDB may be nil if no read replica is configured.
+func NewWithDB(primaryDB, secondaryDB *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
+	primaryCollector, err := configureDB(primaryDB, cfg, "openfga")
+	if err != nil {
+		return nil, fmt.Errorf("configure primary db: %w", err)
```
DanCech, line 147: there is an inconsistency between the handling here (leave primaryDB connected) and below where we `Close` it. We should make a decision about where connection handling is done and make sure we're consistent.

Since the connections are being passed in to this function, it seems like closing them here might be an overreach. It might be better to change New so that it checks the return value of NewWithDB and calls Close on the connection(s) if this function returns an error, with the internal handling here just taking care of unregistering the primary collector if needed.

pkg/storage/mysql/mysql.go
```diff
@@ -104,54 +104,123 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
...
+		versionReady:              false,
 	}, nil
 }
 
+func (s *Datastore) isSecondaryConfigured() bool {
+	return s.secondaryDB != nil
+}
+
+// getSQLDB returns the *sql.DB to use based on the requested consistency preference.
+// HIGHER_CONSISTENCY routes to primaryDB; MINIMIZE_LATENCY routes to secondaryDB when configured.
```
DanCech: this comment should match the other update, everything except HIGHER_CONSISTENCY goes to the replica

pkg/storage/mysql/mysql_test.go
```diff
@@ -1132,6 +1132,18 @@ func TestNew(t *testing.T) {
 			wantErr: true,
 		},
 	}
+	// bad_secondary_uri is tested separately because it requires a running primary container.
```
DanCech: this is no longer true, because we reject at ParseDSN, so we can avoid spinning up the containers

pkg/storage/mysql/mysql.go
chiragsoni-eternal: Fixed in [8bcd0df](https://github.com/openfga/openfga/pull/3290/commits/8bcd0dfd08bee2734c25198f2add66b2635df9c5)

pkg/storage/mysql/mysql.go
chiragsoni-eternal: Fixed in [8bcd0df](https://github.com/openfga/openfga/pull/3290/commits/8bcd0dfd08bee2734c25198f2add66b2635df9c5)

pkg/storage/mysql/mysql_test.go
chiragsoni-eternal: Fixed in [8bcd0df](https://github.com/openfga/openfga/pull/3290/commits/8bcd0dfd08bee2734c25198f2add66b2635df9c5)

pkg/storage/mysql/mysql.go
```diff
@@ -104,54 +104,129 @@ func NewWithDB(db *sql.DB, cfg *sqlcommon.Config) (*Datastore, error) {
...
 	filter storage.ReadFilter,
-	_ storage.ReadOptions,
+	options storage.ReadOptions,
 ) (storage.TupleIterator, error) {
 	ctx, span := startTrace(ctx, "Read")
 	defer span.End()
 
-	return s.read(ctx, store, filter, nil)
+	db := s.getSQLDB(options.Consistency.Preference)
+	return s.read(ctx, store, filter, nil, db)
```
DanCech, line 220: every call site seems to use this pattern, would it make sense to move the `getSQLDB` call into `read` and pass a `storage.ReadPageOptions` with the requested consistency instead of `nil`?

chiragsoni-eternal: Hi @DanCech , `nil` value of options is currently used to drive the [logic inside read function](https://github.com/openfga/openfga/blob/73591ef16ce508623920d5b706286ffcdfb6841b/pkg/storage/postgres/postgres.go#L363-L418) if we change it here then this logic needs to be touched and we will deviate from the pattern we follow in postgresql. So I think we can keep it same, also there are  just two occurrence of this pattren.

DanCech: that works, it seems like something we can clean up separately for both implementations (the order-by addition could be triggered by the user specifying a page size and/or "from" specifier rather than just non-nil options).

DanCech (dismissed):

DanCech (approved):

Vic-Dev (approved):

Vic-Dev:
@chiragsoni-eternal can you update against main?

chiragsoni-eternal:
> @chiragsoni-eternal can you update against main?

Hi @Vic-Dev , Thanks for the approval, done the update with main.

Issue #1961, opened 2024-09-24
When ListObjects & ListUsers hit the deadline, inform the user that this has happened


miparnisari:
### Checklist

- [X] I agree to the terms within the [OpenFGA Code of Conduct](https://github.com/openfga/.github/blob/main/CODE_OF_CONDUCT.md).

### Describe the problem you'd like to have solved

When ListObjects and ListUsers API have to scan a large number of tuples, it's possible that they hit the `deadline` setting in the process. The APIs will return all the objects they have found until then, but we are giving no indication to the user that the `deadline` was hit, and they could be missing out on results.

This is also important on tests - when a test fails, e.g.

```
        --- FAIL: TestServerWithSQLiteDatastore/TestListObjects/max_results_with_model_that_uses_exclusion (0.09s)
            --- FAIL: TestServerWithSQLiteDatastore/TestListObjects/max_results_with_model_that_uses_exclusion/streaming_endpoint (0.01s)
                list_objects.go:567: 
                    	Error Trace:	/home/runner/work/openfga/openfga/pkg/server/test/list_objects.go:567
                    	Error:      	"1" is not greater than or equal to "2"
                    	Test:       	TestServerWithSQLiteDatastore/TestListObjects/max_results_with_model_that_uses_exclusion/streaming_endpoint
FAIL
```

it's likely that the cause of this failure is that the `deadline` was hit. But we currently don't know.

### Describe the ideal solution

Some kind of indication that the `deadline` was hit: a new field in the response, or a response header, or <insert ideas here>.

### Alternatives and current workarounds

None :(

edwin-Marrima:
Hi, I would like to work on this issue

leefernandes:
It also seems odd that ListObjects doesn't have a continuation token implementation when it has a max results configuration.

rzmeditzyt:
Hi there, any updates on this? We recently ran into this issue and our service returned incorrect results on list calls causing even worse issues downstream... I think in these cases, it should return an error instead of producing a partial result, or at least provide some option to control this behavior.

PR #3293, opened 2026-09-07
Warn the caller when ListObjects results are truncated by the deadline

### LINKED ISSUE #3276: Bug:  ListObjects silent truncation and missing response header on Context Deadline Exceeded
### Checklist

### Issue Title
**Bug: `ListObjects` silent truncation and missing response header on Context Deadline Exceeded**

### Description
In `pkg/server/commands/list_objects.go`, when the fallback `evaluate` function hits the configured context deadline timeout (`--listObjects-deadline`), the internal `pool.Wait()` worker group returns a `context.DeadlineExceeded` error.

The function intentionally swallows `context.DeadlineExceeded` to gracefully return whatever partial results were gathered before the timeout (returning `res, nil` up to the handler). However, the `// TODO set header to indicate "deadline exceeded"` directly below it was never implemented. 

Because no specific gRPC metadata is attached, the HTTP/gRPC protocol boundaries intercept the `nil` error and successfully return an HTTP `200 OK` with an empty or partial `"objects":[]` payload, giving the API consumer absolutely zero indication that the traversal was forcefully truncated.

### Steps to Reproduce
1. Start an OpenFGA server with an artificially low ListObjects deadline to guarantee truncation before a complex resolution can finish:
   ```bash
   openfga run --listObjects-pipeline-enabled=false --listObjects-deadline 10us
   ```
2. Create a basic authorization model (e.g. `user` can `view` a `document`).
3. Write test tuples (e.g. `user:anne` is `viewer` for `document:1`, `document:2`).
4. Execute a `ListObjects` request for `document` objects viewable by `user:anne`.
5. Observe the response:
   ```http
   HTTP/1.1 200 OK
   Content-Type: application/json
   
   {"objects":[]}
   ```

The consumer receives an empty array and a `200 OK` rather than an error or an `X-Openfga-Warning` header.

### Expected Behavior
If an evaluation gracefully aborts early to salvage partial results, the server must append metadata notifying the API consumer so they do not falsely assume the object list is exhaustive. 

Applying `grpc.SetHeader(ctx, metadata.Pairs("x-openfga-warning", "deadline-exceeded"))` at the site of the `TODO` correctly propagates this context up to the `grpc-gateway`, resulting in the expected behavior:

```http
HTTP/1.1 200 OK
X-Openfga-Warning: deadline-exceeded
   
{"objects":["document:1"]} // Partial objects
```

### OpenFGA version

latest

As a binary

In-Memory

### OpenFGA Flags

--listObjects-pipeline-enabled=false --listObjects-deadline 10us

### Logs

_No response_

arpitjain099:
Closes #3276.

When resolution hits `--listObjects-deadline`, `evaluate` deliberately swallows `context.DeadlineExceeded` so the partial results survive, and the request returns 200 with whatever was found. There has been a `// TODO set header to indicate "deadline exceeded"` sitting next to that for a while. Without it a truncated list is indistinguishable from an exhaustive one, so a consumer treats a short answer as the complete set.

Rather than reach for `grpc.SetHeader` inside the command, which would put transport concerns into `pkg/server/commands`, the fact travels back the way the throttling flags already do. `ListObjectsResolutionMetadata` gains `DeadlineExceeded atomic.Bool` alongside `DispatchThrottled` and `DatastoreThrottled`, it is set at each of the three places the deadline error is elided (the legacy `evaluate` handler, the pipeline path, and the streamed pipeline path), and `pkg/server/list_objects.go` sets the header through `s.transport`, which is what the other response headers here already use. It also lands in `grpc_ctxtags` as `request.deadline_exceeded`, so it shows up in request logs next to the throttling tags.

The header is `X-Openfga-Warning: deadline-exceeded`, the name the issue proposes.

Verified by extending `TestServer_ThrottleUntilDeadline`, which already forces the deadline reliably. The server is given a recording transport, and the ListObjects subtest asserts the header. Dropping just the `SetHeader` call gives:

    expected: "deadline-exceeded"
    actual  : ""

`./pkg/server/commands/...` is green. `./pkg/server/` has six failures, all of them `TestServerWithPostgresDatastore`, `TestServerWithMySQLDatastore`, `TestServerNotReadyDueToDatastoreRevision` and `TestReleasesConnections`, which want database containers I do not have here. They fail identically on `main` without this branch; I checked both ways rather than assuming.

Two things I deliberately left alone, happy to take direction on either. `StreamedListObjects` now records the flag but does not emit a header, since on a stream the headers go out with the first message and by the time the deadline trips objects have usually been sent already. And the header is a warning rather than an error because the current contract is to return partial results with a 200; changing that is a bigger decision than this issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - ListObjects responses now indicate when resolution stops at the configured deadline.
  - Partial results include a warning header so clients can distinguish them from complete result sets.
  - Deadline-related status is recorded consistently across supported ListObjects execution paths.

### DIFF
--- a/pkg/server/commands/list_objects.go
+++ b/pkg/server/commands/list_objects.go
@@ -110,6 +110,12 @@ type ListObjectsResolutionMetadata struct {
 	// DatastoreThrottled indicates whether the request was throttled by the Datastore.
 	DatastoreThrottled atomic.Bool
 
+	// DeadlineExceeded indicates that resolution stopped early because the
+	// ListObjects deadline was reached, so Objects holds whatever was found up
+	// to that point rather than the full set. Callers that surface results to a
+	// client should tell it the list is partial.
+	DeadlineExceeded atomic.Bool
+
 	// WasWeightedGraphUsed indicates whether the weighted graph was used as the algorithm for the ListObjects request.
 	WasWeightedGraphUsed atomic.Bool
 
@@ -486,7 +492,9 @@ func (q *ListObjectsQuery) evaluate(
 			if !errors.Is(err, context.DeadlineExceeded) && !errors.Is(err, context.Canceled) {
 				resultsChan <- ListObjectsResult{Err: err}
 			}
-			// TODO set header to indicate "deadline exceeded"
+			if errors.Is(err, context.DeadlineExceeded) {
+				resolutionMetadata.DeadlineExceeded.Store(true)
+			}
 		}
 		close(resultsChan)
 		dsMeta := ds.GetMetadata()
@@ -644,6 +652,9 @@ func (q *ListObjectsQuery) Execute(
 			// current list objects behavior is to elide context cancelation errors
 			if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
 				return nil, serverErrors.HandleError("", err)
+			}
+			if errors.Is(err, context.DeadlineExceeded) {
+				res.ResolutionMetadata.DeadlineExceeded.Store(true)
 			}
 		}
 
@@ -841,6 +852,10 @@ func (q *ListObjectsQuery) ExecuteStreamed(ctx context.Context, req *openfgav1.S
 			}
 		}
 		p.Close() // ensure that the pipeline is closed after any early loop exits
+
+		if errors.Is(errRx, context.DeadlineExceeded) {
+			resolutionMetadata.DeadlineExceeded.Store(true)
+		}
 
 		if errRx != nil && !errors.Is(errRx, context.Canceled) && !errors.Is(errRx, context.DeadlineExceeded) {
 			if errors.Is(errRx, condition.ErrEvaluationFailed) {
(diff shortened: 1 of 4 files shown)


yxshwanth (commented):
Thanks for this. The metadata -> transport pattern matches how throttling is already surfaced, and the throttle-until-deadline test is the right place to lock the header in.

Two small follow-ups that would help this land cleanly:

1. **StreamedListObjects telemetry** (CodeRabbit already flagged this): `ExecuteStreamed` sets `DeadlineExceeded`, but `StreamedListObjects` never writes `request.deadline_exceeded` into `grpc_ctxtags`. Unary does. Worth adding the tag before `request.weighted_graph` even if we still skip the response header on the stream (headers go out with the first message).

2. **ListUsers gap**: the same silent truncation exists in `pkg/server/commands/listusers/list_users_rpc.go` (the TODO about response headers next to the deadline swallow). `#1961` / the ListUsers half of this story is still open. I am going to file that as a sibling issue and open a follow-up PR that reuses `WarningHeader` / `deadline-exceeded` once this lands, or include it on top of your branch with clear attribution if that is easier for maintainers.

Happy to send a tiny PR against your branch for (1) if you want.

yxshwanth:
Follow-up is up as https://github.com/openfga/openfga/pull/3311. It rebases your ListObjects commit onto current `main`, adds the StreamedListObjects `request.deadline_exceeded` tag, and extends the same warning to ListUsers ([#3310](https://github.com/openfga/openfga/issues/3310)).

Your authorship is preserved on the first commit. If you prefer to land this PR as-is first, I can trim #3311 down to ListUsers-only after merge. Otherwise #3311 can supersede this one.

PR #3308, opened 2026-09-15
feat(scripts): add "make upgrade-go" to automate Go toolchain and image upgrades

poovamraj:
## Description

Upgrading Go in OpenFGA has been a manual playbook: find the latest stable Go, pull the `chainguard/go` and `chainguard/static` images to read their digests, grab the latest `grpc-health-probe` release, then hand-edit `go.mod`, both Dockerfiles, and the `CHANGELOG`. It's easy to fat-finger a digest or forget to keep the two Dockerfiles in sync.

This adds `scripts/upgrade-go.sh` and a `make upgrade-go` target that does all of it in one step. The script:

- reads the latest stable Go version from go.dev
- resolves the `chainguard/go`, `chainguard/static`, and `grpc-health-probe` digests straight from the registry with `docker buildx imagetools inspect`, so it never downloads an image just to read a digest
- runs `chainguard/go:latest` to confirm it really is the target Go version, and aborts if Chainguard hasn't published it yet (go.dev is sometimes ahead)
- rewrites only the `toolchain` line in `go.mod` (never the `go 1.X` language directive), updates both Dockerfiles, and inserts a Security bullet in the `CHANGELOG` with a `#PLACEHOLDER` for the PR number

`make upgrade-go ARGS="--dry-run"` prints the planned changes without touching any files. The edits are idempotent, so a second run produces no diff once you're on the latest.

No Go version bump ships in this PR. This is the tooling only; running the actual upgrade is a separate follow-up.

## References

N/A. Replaces the manual upgrade steps that previously lived only in maintainers' heads.

## Review Checklist


scripts/upgrade-go.sh
```diff
@@ -0,0 +1,244 @@
...
+  docker buildx version >/dev/null 2>&1 \
+    || die "docker buildx is required (used to read image digests) but is not available."
+}
+
+# Echo the latest STABLE Go version, e.g. "1.26.8".
+latest_go_version() {
+  curl -fsSL "https://go.dev/dl/?mode=json" \
+    | jq -r 'map(select(.stable == true)) | .[0].version' \
+    | sed 's/^go//'
+}
```
saad-h1, line 67: I wonder if we should get the latest stable version for the major release prior to the latest. E.g., if 1.27 is out, 1.26.X. E.g. for https://pkg.go.dev/vuln/GO-2026-5026, 1.27 was out but the fix wasn't in a stable 1.27 release, it was in `1.27.0-rc.3`. It was in a stable 1.26 version though, 1.26.6.

Would be nice if the script could also accept an optional go version.

scripts/upgrade-go.sh
```diff
@@ -0,0 +1,244 @@
...
+# target Go version. Chainguard's public registry serves only :latest (versioned
+# tags need a paid tier) and the image carries NO version label, so the only
+# ground truth is running it. Aborts if Chainguard's latest lags the target.
+# $1 = target Go version (e.g. "1.26.8"). Echoes the digest on success.
+resolve_chainguard_go() {
+  local ver="$1" latest_digest actual_version
+  latest_digest="$(image_digest "cgr.dev/chainguard/go:latest")"
+  [[ -n "$latest_digest" ]] || die "Could not resolve cgr.dev/chainguard/go:latest digest."
+  # Pin to the digest we just resolved so the version we read is the version we pin.
+  actual_version="$(docker run --rm --entrypoint go "cgr.dev/chainguard/go:latest@${latest_digest}" version 2>/dev/null \
```
saad-h1, line 86: Instead of running an image, could we do something similar to this?
```sh
# Get the value of the latest published go from Chainguard (https://images.chainguard.dev/directory/image/go/versions)
GOVERSION="1.25.7"
# Pull the base image to get its latest digest
digest=$(docker pull cgr.dev/chainguard/go | grep Digest | cut -d ' ' -f 2)
# Attempt to pull the image pinned with both tag AND digest
docker pull cgr.dev/chainguard/go:${GOVERSION}@${digest} > /dev/null 2>&1 \
  && echo "for builder use: cgr.dev/chainguard/go:${GOVERSION}@${digest}" \
  || echo "WARNING: Failed to pull image cgr.dev/chainguard/go:${GOVERSION}@${digest} - ${GOVERSION} digest doesn't match latest"
# !!Do not replace cgr.dev/chainguard/static with cgr.dev/chainguard/go - instead bump up static to latest:
static_digest=$(docker pull cgr.dev/chainguard/static | grep Digest | cut -d ' ' -f 2) && echo "for static use: cgr.dev/chainguard/static@${static_digest}"
```

scripts/upgrade-go.sh
```diff
@@ -0,0 +1,244 @@
...
+    /^## \[Unreleased\]/ { in_unreleased=1; print; next }
+    /^## \[/ && in_unreleased==1 {
+      if (!done) { print "### Security"; print bullet; print ""; done=1 }
+      in_unreleased=0
+    }
+    in_unreleased==1 && /^### Security/ && done==0 {
+      print; print bullet; done=1; next
+    }
+    { print }
+    END { if (in_unreleased==1 && done==0) { print "### Security"; print bullet } }
```
saad-h1, line 153: I think CI enforces an ordering to the changelog as well, maybe we should run `npx keep-a-changelog` at the end of this to auto reformat

scripts/upgrade-go.sh
poovamraj: If I am understanding right, not let the script do major version upgrades and only minor or patches?

Apart from it being optional?

scripts/upgrade-go.sh
poovamraj: Gotcha, can do 👍

scripts/upgrade-go.sh
poovamraj: Yep, would be quicker this way 👍

scripts/upgrade-go.sh
saad-h1: Oh right, actually that might cause more trouble than it's worth.

Maybe we can just leave it as is (update to latest stable version by default), but also allow someone to optionally provide a version in case there's a situation like https://pkg.go.dev/vuln/GO-2026-5026 where the latest stable version might not have the fix and we have to use a different version.

